PR #12892, opened 2019-04-21
bpo-36679: rename duplicate test_class_getitem function

Windsooon:
rename duplicate `test_class_getitem` to `test_class_getitem_format`, I'm not sure should I add news for it?

https://bugs.python.org/issue36679


ilevkivskyi:
https://github.com/python/cpython/pull/12978 is the backport to 3.7

PR #12894, opened 2019-04-21
bpo-36681: remove duplicate test_regression_29220 function

Windsooon:
Remove duplicate `test_regression_29220` function

https://bugs.python.org/issue36681


eamanu (approved):
LGTM

PR #12895, opened 2019-04-21
bpo-36682: rename duplicate functions

Windsooon:
Rename duplicate functions

https://bugs.python.org/issue36682


eamanu (changes requested):
Hello! IMO This aren't a duplicate functions issue. The first method is about there are not arguments.
The second method is for invalid arguments. 

Furthermore, the name are not equal ```test_unbound_method_no_args != test_unbound_method_invalid_args```

Windsooon:
@eamanu You may have a look at test_sys_setprofile.py, there are two `test_unbound_method_no_args` functions and two `test_unbound_method_invalid_args` functions.

eamanu:
Oh yes,  I see that.  Good catch!

PR #12896, opened 2019-04-21
bpo-36683: rename duplicate function test_io_encoding to test_pyio_encoding

Windsooon:
rename duplicate function test_io_encoding to test_pyio_encoding

https://bugs.python.org/issue36683


eamanu (approved):
LGTM. What about join both check on one test?

zooba:
Better to keep the tests separate so that we get more accurate reporting of which one failed.

PR #12900, opened 2019-04-22
bpo-35184: Fix XML_POOR_ENTROPY option that breaks makesetup parsing of pyexpat line in Setup.dist

aaronpaulhurst:
When the line is uncommented, the equals character causes it to be incorrectly interpreted
as a macro definition by makesetup.  This results in an invalid Makefile output.

The expat code only requires XML_POOR_ENTROPY to be defined; the value is unnecessary.

https://bugs.python.org/issue35184

### DIFF
--- a/Modules/Setup.dist
+++ b/Modules/Setup.dist
@@ -368,7 +368,7 @@ _symtable symtablemodule.c
 # Interface to the Expat XML parser
 # More information on Expat can be found at www.libexpat.org.
 #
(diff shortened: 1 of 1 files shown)


the-knights-who-say-ni:

ned-deily:
Thanks for the PR.  See the comments on the bug issue.  (Note that the 3.6 branch is now in security-fix-only mode so a change like this would not meet the criteria for including in a future 3.6 security release which is why PRs like this should be based against the master branch.)

PR #12907, opened 2019-04-22
gh-72300: Convert Win32{Symlink,Junction}Tests to use support.TESTFN

berkerpeksag:
https://bugs.python.org/issue28113

* Issue: gh-72300

### DIFF
--- a/Lib/test/test_os.py
+++ b/Lib/test/test_os.py
@@ -2148,26 +2148,25 @@ class ReadlinkTests(unittest.TestCase):
 @unittest.skipUnless(sys.platform == "win32", "Win32 specific tests")
 @support.skip_unless_symlink
 class Win32SymlinkTests(unittest.TestCase):
-    filelink = 'filelinktest'
-    filelink_target = os.path.abspath(__file__)
-    dirlink = 'dirlinktest'
-    dirlink_target = os.path.dirname(filelink_target)
-    missing_link = 'missing link'
+    dirlink_target = os.path.join(support.TESTFN, 'win32_symlink_tests')
+    filelink_target = os.path.join(dirlink_target, support.TESTFN)
+    dirlink = os.path.join(support.TESTFN, 'dirlinktest')
+    filelink = os.path.join(support.TESTFN, 'filelinktest')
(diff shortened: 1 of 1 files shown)


Lib/test/test_os.py
```diff
@@ -2148,26 +2148,25 @@ def test_bytes(self):
 @unittest.skipUnless(sys.platform == "win32", "Win32 specific tests")
 @support.skip_unless_symlink
 class Win32SymlinkTests(unittest.TestCase):
-    filelink = 'filelinktest'
-    filelink_target = os.path.abspath(__file__)
-    dirlink = 'dirlinktest'
-    dirlink_target = os.path.dirname(filelink_target)
-    missing_link = 'missing link'
+    dirlink_target = os.path.join(support.TESTFN, 'win32_symlink_tests')
```
taleinat, line 2151: `support.TESTFN` is just a path, but not an existing directory.

Perhaps add something the following to `setUp()`?
```python3
        tempdir = tempfile.TemporaryDirectory(prefix=support.TESTFN)
        self.addCleanup(tempdir.cleanup)

        self.dirlink_target = os.path.join(tempdir , 'win32_symlink_tests')
        self.filelink_target = os.path.join(self.dirlink_target, support.TESTFN)
        self.dirlink = os.path.join(tempdir , 'dirlinktest')
        self.filelink = os.path.join(tempdir , 'filelinktest')
        self.missing_link = os.path.join(tempdir , 'missinglinktest')
```

With this all of the other existence checks and cleanups may be removed.

Lib/test/test_os.py
```diff
@@ -2200,16 +2199,12 @@ def test_remove_directory_link_to_missing_target(self):
         #  was created with target_is_dir==True.
         os.remove(self.missing_link)
 
-    @unittest.skip("currently fails; consider for improvement")
```
taleinat, line 2203: Why is this no longer skipped: has this behavior been implemented?

Likewise for the next test below.

taleinat (changes requested):
Note that `Win32JunctionTests` also uses `os.path.dirname(os.path.abspath(__file__))`; that should be amended as well.

PR #11774, opened 2019-02-06
bpo-35920: Windows 10 ARM32 platform support

paulmon:
At a minimum this code needs to be refactored to take into account removing libffi_msvc and moving to the current libffi implementation based on https://github.com/python/cpython/pull/3806.

https://bugs.python.org/issue35920

PR #12912, opened 2019-04-22
Fixes platform.win32_ver on non-Windows platforms

zooba:


zooba:
I spotted this issue on #11774 which adds some extra calls to this function in tests. We shouldn't be letting the `ImportError` escape regardless.

PR #12914, opened 2019-04-22
bpo-29183: Fix double exceptions in wsgiref.handlers.BaseHandler

berkerpeksag:
https://bugs.python.org/issue29183

### DIFF
--- a/Lib/wsgiref/handlers.py
+++ b/Lib/wsgiref/handlers.py
@@ -183,7 +183,16 @@ class BaseHandler:
                 for data in self.result:
                     self.write(data)
                 self.finish_content()
-        finally:
+        except:
+            # Call close() on the iterable returned by the WSGI application
+            # in case of an exception.
+            if hasattr(self.result, 'close'):
+                self.result.close()
+            raise
+        else:
+            # We only call close() when no exception is raised, because it
+            # will set status, result, headers, and environ fields to None.
+            # See bpo-29183 for more details.
             self.close()
 
 
--- a/Lib/test/test_wsgiref.py
+++ b/Lib/test/test_wsgiref.py
@@ -806,6 +806,31 @@ class HandlerTests(TestCase):
 
                 self.assertFalse(stderr.getvalue())
 
+    def testDontResetInternalStateOnException(self):
+        class CustomException(ValueError):
+            pass
+
+        # We are raising CustomException here to trigger an exception
+        # during the execution of SimpleHandler.finish_response(), so
+        # we can easily test that the internal state of the handler is
+        # preserved in case of an exception.
+        class AbortingWriter:
+            def write(self, b):
+                raise CustomException
+
+        stderr = StringIO()
+        environ = {"SERVER_PROTOCOL": "HTTP/1.0"}
+        h = SimpleHandler(BytesIO(), AbortingWriter(), stderr, environ)
+        h.run(hello_app)
+
+        self.assertIn("CustomException", stderr.getvalue())
+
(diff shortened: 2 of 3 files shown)


Lib/wsgiref/handlers.py
```diff
@@ -179,8 +179,14 @@ def finish_response(self):
                 for data in self.result:
                     self.write(data)
                 self.finish_content()
-        finally:
-            self.close()
+        except:
+            if hasattr(self.result, 'close'):
+                self.result.close()
+            raise
+        # We only call close() when no exception is raised, because it
```
jleclanche: Shouldn't this be in an `else` block then?

berkerpeksag: What's the advantage of using an `else` clause in this specific case? Would it make code more readable? Would it make diff shorter? I like using `try...else` clauses, I just want to know why do you prefer the `try...else` way.

vadmium:
Revision da0ca97 looks okay to me. I presume without the fix, the test would fail because the _h.run_ call will raise the double exception.

berkerpeksag:
@vadmium thanks for the review! Yes, the test will fail without the fix:

```
ERROR: testDontResetInternalStateOnException (test.test_wsgiref.HandlerTests)
  File "/Users/berkerpeksag/projects/cpython/Lib/wsgiref/handlers.py", line 138, in run
    self.finish_response()
  File "/Users/berkerpeksag/projects/cpython/Lib/wsgiref/handlers.py", line 184, in finish_response
    self.write(data)
  File "/Users/berkerpeksag/projects/cpython/Lib/wsgiref/handlers.py", line 289, in write
    self.send_headers()
  File "/Users/berkerpeksag/projects/cpython/Lib/wsgiref/handlers.py", line 347, in send_headers
    self.send_preamble()
  File "/Users/berkerpeksag/projects/cpython/Lib/wsgiref/handlers.py", line 267, in send_preamble
    self._write(('HTTP/%s %s\r\n' % (self.http_version,self.status)).encode('iso-8859-1'))
  File "/Users/berkerpeksag/projects/cpython/Lib/wsgiref/handlers.py", line 468, in _write
    result = self.stdout.write(data)
  File "/Users/berkerpeksag/projects/cpython/Lib/test/test_wsgiref.py", line 819, in write
    raise CustomException
test.test_wsgiref.HandlerTests.testDontResetInternalStateOnException.<locals>.CustomException

  File "/Users/berkerpeksag/projects/cpython/Lib/test/test_wsgiref.py", line 824, in testDontResetInternalStateOnException
    h.run(hello_app)
  File "/Users/berkerpeksag/projects/cpython/Lib/wsgiref/handlers.py", line 145, in run
    self.handle_error()
  File "/Users/berkerpeksag/projects/cpython/Lib/wsgiref/handlers.py", line 383, in handle_error
    self.finish_response()
  File "/Users/berkerpeksag/projects/cpython/Lib/wsgiref/handlers.py", line 184, in finish_response
    self.write(data)
  File "/Users/berkerpeksag/projects/cpython/Lib/wsgiref/handlers.py", line 289, in write
    self.send_headers()
  File "/Users/berkerpeksag/projects/cpython/Lib/wsgiref/handlers.py", line 346, in send_headers
    if not self.origin_server or self.client_is_modern():
  File "/Users/berkerpeksag/projects/cpython/Lib/wsgiref/handlers.py", line 359, in client_is_modern
    return self.environ['SERVER_PROTOCOL'].upper() != 'HTTP/0.9'
TypeError: 'NoneType' object is not subscriptable

```

PR #12915, opened 2019-04-22
bpo-29940: Add follow_wrapped option to help()

eamanu:
When print the docstring using help builtin for an wrapper function
it doesn't correspond to the function wrapped. For avoid this
this PR add the follow_wrapped option to help() function.

https://bugs.python.org/issue29940


chrahunt:
There are no tests. Can we find useful ones to add that will also demonstrate the reason for this change? e.g. showing correct help for functions wrapped with class-based decorators

I think that `help` may be better with `follow_wrapped=True` by default. It is meant to be used interactively as stated in the docs so the backwards-incompatibility may not be as big an issue.

Lib/pydoc.py
```diff
@@ -1669,9 +1669,11 @@ def render_doc(thing, title='Python Library Documentation: %s', forceload=0,
     return title % desc + '\n\n' + renderer.document(object, name)
 
 def doc(thing, title='Python Library Documentation: %s', forceload=0,
-        output=None):
+        output=None, follow_wrapped=True):
     """Display text documentation, given an object or a path to an object."""
     try:
+        if not follow_wrapped:
```
chrahunt: Should be just `follow_wrapped`, right?

Also can we move it outside the `try`, since it should not raise any exception?

Lib/pydoc.py
```diff
@@ -1892,9 +1894,9 @@ def __repr__(self):
                                      self.__class__.__qualname__)
 
     _GoInteractive = object()
-    def __call__(self, request=_GoInteractive):
+    def __call__(self, request=_GoInteractive, follow_wrapped=False):
```
chrahunt: Can you explain why this should be `False` by default?

chrahunt (changes requested):
Can we add some tests?

Lib/pydoc.py
```diff
@@ -1669,9 +1669,11 @@ def render_doc(thing, title='Python Library Documentation: %s', forceload=0,
     return title % desc + '\n\n' + renderer.document(object, name)
 
 def doc(thing, title='Python Library Documentation: %s', forceload=0,
-        output=None):
+        output=None, follow_wrapped=True):
     """Display text documentation, given an object or a path to an object."""
     try:
+        if not follow_wrapped:
+            thing = inspect.unwrap(thing)
```
taleinat: Perhaps this should stop when finding an object  with a `__doc__` attribute?
```suggestion
            thing = inspect.unwrap(thing, stop=lambda func: getattr(func, '__doc__', None)
```

taleinat:
Ping, @eamanu?

eamanu:
@taleinat I will update the PR today

Lib/pydoc.py
eamanu: @chrahunt If I'm not wrong unwrap raise a ValueError. I don't know if is correct move outiside `try` this line. 
Also, I add the `ValueError` to except, what about it?

Lib/pydoc.py
eamanu: Yes, thinking better, here must be a `follow_wrapped=True`

Lib/pydoc.py
eamanu: @chrahunt I use `if follow_wrapped:`

eamanu:
@taleinat @chrahunt I think is ready for a new review.

the-knights-who-say-ni:

## Recognized GitHub username

@nosamproductions

eamanu:
oops. sorry!!!!!

eamanu:
i am preparing a new pr for this... sorry for the nnoise.

taleinat:
@eamanu, no problem and no need to apologize, it happens all too often with git.

Just in case you haven't, I recommend reading the ["Lifecycles of a Pull Requst" section on the devguide](https://devguide.python.org/pullrequest/), which may be helpful. You're also welcome to ask anything on the Python IRC and Zulip chat channels and get immediate help.

vsajip:
I was under the impression that this happened only when rebasing/fast-forwarding with `master` - does it also happen when a merge is done?

taleinat:
> I was under the impression that this happened only when rebasing/fast-forwarding with master - does it also happen when a merge is done?

In this case it appears the @eamanu accidentally merged in another branch, which may have been similary re-based or based on a different commit... I don't that the details are very important. It happens. The right thing to do afterwards is indeed to close the PR and open a new one.

eamanu:
Hi I make this PR: https://github.com/python/cpython/pull/22390


> I was under the impression that this happened only when rebasing/fast-forwarding with `master` - does it also happen when a merge is done?

I need practice my git skills. I don't like the rebase (or I don't understand yet) and I did an rebase from master, and then I did (and I think that was my error) `git pull bpo-29940` and that merge my old github's branch

PR #12917, opened 2019-04-23
bpo-36511: Windows arm32 buildbot changes

paulmon:
These changes are to enable a buildbot for Windows ARM32 builds on Windows IoT Core
Suggestions on better ways to implement or refactor the changes are greatly appreciated.
My plan is for this change to work with this worker: https://github.com/python/buildmaster-config/pull/89

@zooba @zware 


zooba (approved):
I'm okay with this, but @zware traditionally has stronger opinions about the buildbot scripts than I do :)

PR #12922, opened 2019-04-23
bpo-36635, bpo-36696: Fix setup.py on AIX

vstinner:
xlc compiler doesn't support "-D define" flag only "-Ddefine".

https://bugs.python.org/issue36635


vstinner:
@aixtools: Does this PR fix https://bugs.python.org/issue36696 for you?

aixtools:
Yes, fixes issue.

vstinner:
> Yes, fixes issue.

Thanks for testing. I merged the PR.

PR #12929, opened 2019-04-23
bpo-36454: Fix test_time.test_monotonic()

vstinner:
Fix test_time.test_monotonic() on slow buildbots: only test the
minimum time, not the maximum time.

https://bugs.python.org/issue36454

### DIFF
--- a/Misc/NEWS.d/next/Tests/2019-04-23-17-48-11.bpo-36454.0q4lQz.rst
+++ b/Misc/NEWS.d/next/Tests/2019-04-23-17-48-11.bpo-36454.0q4lQz.rst
@@ -1,0 +1,2 @@
+Fix test_time.test_monotonic() on slow buildbots: only test the minimum
+time, not the maximum time.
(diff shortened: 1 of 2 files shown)


Misc/NEWS.d/next/Tests/2019-04-23-17-48-11.bpo-36454.0q4lQz.rst
```diff
@@ -0,0 +1,2 @@
+Fix test_time.test_monotonic() on slow buildbots: only test the minimum
```
pganssle: This doesn't make a lot of sense without any context. Maybe instead:

"Change ``test_time.test_monotonic()`` to test only the lower bound of elapsed time after a sleep command rather than the upper bound. This prevents unnecessary test failures on slow buildbots."

Also, is "Patch by Victor Stinner." preferred, or is that something that is considered up to the submitter?

pganssle (approved):
I have left one comment about the news entry, but it is not important to be a blocker. This change makes sense to me.

Misc/NEWS.d/next/Tests/2019-04-23-17-48-11.bpo-36454.0q4lQz.rst
vstinner: > Also, is "Patch by Victor Stinner." preferred, or is that something that is considered up to the submitter?

I never know what's the policy. In case of doubt, I followed all your proposals :-)

vstinner:
Thanks for the review @pganssle.

PR #12839, opened 2019-04-15
bpo-36616: optimize handling of thread state in function call code

jdemeyer:
Thanks to @markshannon for the idea.

This does various micro-optimizations for function calls in the bytecode interpreter to give a small increase in performance.

https://bugs.python.org/issue36616

PR #12934, opened 2019-04-24
bpo-36710: Add 'ceval' local variable to ceval.c

vstinner:
Add "struct _ceval_runtime_state *ceval = &_PyRuntime.ceval;" local
variables to function to better highlight the dependency on the
global variable _PyRuntime and to point directly to _PyRuntime.ceval
field rather than on the larger _PyRuntime.

Changes:

* Add _PyRuntimeState_GetThreadState()
* _PyThreadState_GET() now calls _PyRuntimeState_GetThreadState()
* Add an explicit 'ceval' parameter to macros like
  COMPUTE_EVAL_BREAKER() or SIGNAL_PENDING_SIGNALS().
* Add 'tstate' parameter to call_function(), do_call_core() and
  do_raise().
* Add 'runtime' parameter to _Py_CURRENTLY_FINALIZING().
* Declare 'runtime', 'tstate', 'ceval' and 'eval_breaker' variables
  as constant.

### DIFF
--- a/Include/ceval.h
+++ b/Include/ceval.h
@@ -58,7 +58,6 @@ PyAPI_FUNC(int) PyEval_MergeCompilerFlags(PyCompilerFlags *cf);
 #endif
 
 PyAPI_FUNC(int) Py_AddPendingCall(int (*func)(void *), void *arg);
-PyAPI_FUNC(void) _PyEval_SignalReceived(void);
 PyAPI_FUNC(int) Py_MakePendingCalls(void);
 
 /* Protection against deeply nested recursive calls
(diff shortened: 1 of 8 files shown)


vstinner:
This change adds a "tstate" parameter to call_function(), do_call_core() and do_raise(): @jdemeyer proposed a similar change in PR #12839. Here my intent is not optimization, but more correctness: ensure that runtime, interp and tstate are consistent in ceval.c.

vstinner:
@jdemeyer: Would you mind to review my change?

jdemeyer:
> @jdemeyer: Would you mind to review my change?

As discussed on bpo-36710 (let's keep that discussion there), I'm missing the **why**: which problem does this PR solve? Some of these changes do make sense independently to clean things up (for example, reducing the number of `_PyThreadState_GET()` calls like #12839). But some of the changes also make things more complicated, so they need to be justified.

vstinner:
cc @ericsnowcurrently: I was talking about this change with you.

vstinner:
I rebased my PR on top of the master branch.

PR #12701, opened 2019-04-06
bpo-36540: PEP 570 -- Implementation

pablogsal:
implementation of PEP 570 -- Python Positional-Only Parameters:

:warning: **This Pull Request contains only the implementation of PEP570, the documentation will be done in a different PR**.

https://www.python.org/dev/peps/pep-0570/

https://bugs.python.org/issue36540

CC: @mariocj89

PR #12942, opened 2019-04-24
bpo-36540 What's New section for PEP 570

gvanrossum:
@pablogsal You can either copy this into GH-12701 or land it after that has landed.

https://bugs.python.org/issue36540

### DIFF
--- a/Doc/whatsnew/3.8.rst
+++ b/Doc/whatsnew/3.8.rst
@@ -83,6 +83,31 @@ See :pep:`572` for a full description.
 .. TODO: Emily will sprint on docs at PyCon US 2019.
 
 
+Positional-only parameters
+--------------------------
+
+There is new syntax (``/``) to indicate that some function parameters
+must be specified positionally (i.e., cannot be used as keyword
+arguments).  This is the same notation as shown by ``help()`` for
+functions implemented in C (produced by Larry Hastings' "Argument
+Clinic" tool).  Example::
+
+  def pow(x, y, z=None, /):
+      r = x**y
+      if z is not None:
+          r %= z
+      return r
+
+Now ``pow(2, 10)`` and ``pow(2, 10, 17)`` are valid calls, but
+``pow(x=2, y=10)`` and ``pow(2, 10, z=17)`` are invalid.
+
+See :pep:`570` for a full description.
+
+(Contributed by Pablo Galindo in :issue:`36540`.)
+
+.. TODO: Pablo will sprint on docs at PyCon US 2019.
+
+
 Parallel filesystem cache for compiled bytecode files
 -----------------------------------------------------
 


vstinner (approved):
LGTM.

pablogsal:
Thanks for doing this! I will land this after GH-12701 as this is more likely to have merge conflicts (as it has as the time of writing this) and I prefer not to rebase multiple times a PR that is already very big. Also, I will make a separate PR with all the doc changes that are more likely to need native speaker reviews.

gvanrossum:
OK, fixed the merge conflict (of my own making). Are you gonna make alpha 4?

Doc/whatsnew/3.8.rst
```diff
@@ -83,6 +83,29 @@ See :pep:`572` for a full description.
 .. TODO: Emily will sprint on docs at PyCon US 2019.
 
 
+Positional-only parameters
+--------------------------
+
+There is new syntax (``/``) to indicate that some function parameters
```
pganssle:
```suggestion
New syntax (``/``) has been added to indicate that some function parameters
```

"There is new syntax" also is fine grammatically, but usually the voice of these "What's new" entries seems to assume the context is that you are explaining what this PEP changes. "There is new syntax" feels every so slightly off because it de-emphasizes the change aspect of it.

I'll note that to me this is a very subtle difference and if I were not reading this specifically with giving notes on the wording in mind I would probably not have batted an eye, so if you prefer the original wording that's fine.

pganssle, line 89: Maybe "PEP 570 adds new syntax..." instead of "There is new syntax"?

"There is new syntax" is also fine grammatically, but usually the voice of these "What's new" entries seems to assume the context is that you are explaining what this PEP changes. "There is new syntax" feels every so slightly off because it de-emphasizes the change aspect of it.

I'll note that to me this is a very subtle difference and if I were not reading this specifically with giving notes on the wording in mind I would probably not have batted an eye, so if you prefer the original wording that's fine.

pablogsal:
> OK, fixed the merge conflict (of my own making). Are you gonna make alpha 4?

That's my intention but I am still waiting for a review from @serhiy-storchaka (Victor already approved). If we are very close to the date, I would like to merge it (maybe without the changes in inspect as Serhiy suggested) and then we can fix particular details afterwards.

pganssle (approved):
Other than my one *optional* style note, I see no problems with the grammar and style of this.

gvanrossum:
I'll leave it up to Pablo.

Doc/whatsnew/3.8.rst
pablogsal: I will leave  `There is new syntax` for symmetry with the assignment expressions paragraph.

PR #12943, opened 2019-04-24
bpo-36688: Adding an implementation of RLock in _dummy_thread

vlabakje:
Implementing an extremely simple RLock that extends Lock. It accepts multiple acquire() calls and just as many release() calls but no more. This should be sufficient for the _dummy_thread module and fixes issue 36688 where python -c 'import dummy_threading' would cause an ImportError.
I also added tests for RLock in Lib/test/test_dummy_thread and made it executable.

https://bugs.python.org/issue36688

### DIFF
--- a/Lib/_dummy_thread.py
+++ b/Lib/_dummy_thread.py
@@ -14,7 +14,7 @@ Suggested usage is::
 # Exports only things specified by thread documentation;
 # skipping obsolete synonyms allocate(), start_new(), exit_thread().
 __all__ = ['error', 'start_new_thread', 'exit', 'get_ident', 'allocate_lock',
-           'interrupt_main', 'LockType']
+           'interrupt_main', 'LockType', 'RLock']
 
 # A dummy value
 TIMEOUT_MAX = 2**31
@@ -150,17 +150,28 @@ class LockType(object):
 
 
 class RLock(LockType):
+    """Dummy implementation of threading._RLock.
+
+    Re-entrant lock can be aquired multiple times and needs to be released
+    just as many times. This dummy implemention does not check wheter the
+    current thread actually owns the lock, but does accounting on the call
+    counts.
+    """
     def __init__(self):
         super().__init__()
         self._levels = 0
 
     def acquire(self, waitflag=None, timeout=-1):
+        """Aquire the lock, can be called multiple times in succession.
+        """
         locked = super().acquire(waitflag, timeout)
         if locked:
             self._levels += 1
         return locked
 
     def release(self):
+        """Release needs to be called once for every call to acquire().
+        """
         if self._levels == 0:
             raise error
         if self._levels == 1:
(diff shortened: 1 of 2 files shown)


the-knights-who-say-ni:

auvipy (commented):
probably a new entry needed?

Lib/_dummy_thread.py
```diff
@@ -148,6 +148,25 @@ def __repr__(self):
             hex(id(self))
         )
 
+
+class RLock(LockType):
+    def __init__(self):
```
skoslowski, line 160: Maybe consider adding a docstring for this class similar to `LockType`

Lib/test/test_dummy_thread.py
```diff
@@ -102,6 +102,23 @@ def test_lock_representation(self):
...
+    def test_multiple_acquire(self):
+        self.assertIn("unlocked", repr(self.rlock))
+        self.rlock.acquire()
+        self.rlock.acquire()
+        self.assertIn("locked", repr(self.rlock))
+        self.rlock.release()
+        self.assertIn("locked", repr(self.rlock))
+        self.rlock.release()
+        self.assertIn("unlocked", repr(self.rlock))
+
```
skoslowski, line 121: Consider adding another call to release() confirming that a RuntimeError error is raised.

Lib/test/test_dummy_thread.py
```diff
@@ -253,3 +270,6 @@ def test_RaiseException(self, mock_print_exc):
         func = mock.Mock(side_effect=Exception)
         _thread.start_new_thread(func, tuple())
         self.assertTrue(mock_print_exc.called)
+
+if __name__ == '__main__':
+    unittest.main()
```
skoslowski, line 276: Just as a sidenote, while many test modules have these lines, they are not required to run, see https://devguide.python.org/runtests/

Lib/_dummy_thread.py
```diff
@@ -148,6 +148,25 @@ def __repr__(self):
...
+        super().__init__()
+        self._levels = 0
+
+    def acquire(self, waitflag=None, timeout=-1):
+        locked = super().acquire(waitflag, timeout)
```
skoslowski, line 167: Again, consider adding a docstring that documents the dummy nature of this lock

skoslowski (changes requested):
Thanks for your contribution. I ran your tests and double-checked with the documentation

Lib/test/test_dummy_thread.py
vlabakje: Thanks! I'm not sure if this means I should get rid of these lines. Could you clarify that?

skoslowski: No need to remove this - lot of the test modules have it. While it is not required to run the test(s) it is sometimes convenient to have it.

PR #12946, opened 2019-04-25
bpo-21536: C extensions are no longer linked to libpython

vstinner:
On Unix, C extensions are no longer linked to libpython.

It is now possible to load a C extension built using a shared library
Python with a statically linked Python.

When Python is embedded, libpython must not be loaded with
RTLD_LOCAL, but RTDL_GLOBAL instead. Previously, using RTLD_LOCAL, it
was already not possible to load C extensions which were not linked
to libpython, like C extensions of the standard library built by the
"*shared*" section of Modules/Setup.

distutils, python-config and python-config.py have been modified.


vstinner:
This PR allows to load a C extension built by a shared libpython with a statically linked Python: https://bugs.python.org/issue21536#msg340819

It also allows to load C extension built in release mode with a Python built in debug mode: https://bugs.python.org/issue21536#msg340821

vstinner:
One potential issue: linking an application to libpython to embed Python, see discussion at https://bugs.python.org/issue21536#msg340837

pitrou (approved):
LGTM. Thank you!

Doc/whatsnew/3.8.rst
```diff
@@ -845,16 +845,19 @@ Changes in the Python API
 Changes in the C API
 --------------------
 
+* On Unix, C extensions are no longer linked to libpython. When Python is
+  embedded, ``libpython`` must not be loaded with ``RTLD_LOCAL``, but
+  ``RTDL_GLOBAL`` instead. Previously, using ``RTLD_LOCAL``, it was already not
```
skoslowski: minor typo ``RTLD_GLOBAL``

vstinner: Oops, fixed :-)

froody:
This change causes strange bugs on Mac OSX due to the two-level namespace symbol resolution for dynamic libraries. A C extension that links against libpython will resolve symbols from libpython instead of the python executable (for example, https://github.com/PDAL/python/issues/75 ). I believe linking C extensions on Mac OS X with "-flat_namespace" would be a reliable fix, as it would allow the python executable symbols to take precedence over those from libpython.

vstinner:
Please don't comment closed PRs, but open a new issue at https://bugs.python.org/ Otherwise, your comment will simply be ignored.

PR #12947, opened 2019-04-25
bpo-36654: add example to generate token from another file

Windsooon:
Add example to generate token from another file.

https://bugs.python.org/issue36654

### DIFF
--- a/Doc/library/tokenize.rst
+++ b/Doc/library/tokenize.rst
@@ -278,3 +278,15 @@ The exact token type names can be displayed using the :option:`-e` option:
     4,10-4,11:          RPAR           ')'
     4,11-4,12:          NEWLINE        '\n'
     5,0-5,0:            ENDMARKER      ''
+
+Example of tokenizing from another file::
+
+    import tokenize
+    f = open('example.py', 'rb')
+    token_gen = tokenize.tokenize(f.readline)
+
+    for token in token_gen:
+        # TokenInfo(type=59 (ENCODING), string='utf-8', start=(0, 0), end=(0, 0), line='')
+        # TokenInfo(type=1 (NAME), string='def', start=(1, 0), end=(1, 3), line='def foo:\n')
+        # ...
+        print(token)
--- a/Doc/library/tokenize.rst
+++ b/Doc/library/tokenize.rst
@@ -279,10 +279,10 @@ The exact token type names can be displayed using the :option:`-e` option:
     4,11-4,12:          NEWLINE        '\n'
     5,0-5,0:            ENDMARKER      ''
 
-Example of tokenizing from another file::
+Example of tokenizing a file programmatically::
 
     import tokenize
-    f = open('example.py', 'rb')
+    f = open('hello.py', 'rb')
     token_gen = tokenize.tokenize(f.readline)
 
     for token in token_gen:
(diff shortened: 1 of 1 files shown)


eamanu (commented):
This example confuse me, because the file ```example.py``` does not  exist. Maybe you need to do something like ```hello.py```

Doc/library/tokenize.rst
```diff
@@ -278,3 +278,15 @@ The exact token type names can be displayed using the :option:`-e` option:
     4,10-4,11:          RPAR           ')'
     4,11-4,12:          NEWLINE        '\n'
     5,0-5,0:            ENDMARKER      ''
+
+Example of tokenizing from another file::
```
berkerpeksag: Changing this sentence to something like "Example of tokenizing a file programmatically::" or "Example of using :func:\`tokenize.tokenize\` programmatically::" would be better.

See for a similar example:

https://github.com/python/cpython/blob/5c403b203510549a3f89d138d3265c5cc0cc12af/Lib/trace.py#L32-L48

Doc/library/tokenize.rst
```diff
@@ -278,3 +278,15 @@ The exact token type names can be displayed using the :option:`-e` option:
...
+
+    import tokenize
+    f = open('example.py', 'rb')
```
berkerpeksag: Using `tokenize.open()` would be better as it automatically detect the encoding of the file:

https://github.com/python/cpython/blob/5d90954856cc1d999dd5d153b807055fde0d53ea/Lib/tokenize.py#L449

berkerpeksag: Also, I agree with Emmanuel that using the existing `hello.py` as an example would make the example easier to follow.

berkerpeksag (changes requested):

Doc/library/tokenize.rst
Windsooon: I found the [tokenize function](https://github.com/python/cpython/blob/master/Lib/tokenize.py#L404) already call `detect_encoding`.

berkerpeksag: I just realized that `tokenize.open()` returns a text stream, so passing it to `tokenize.tokenize()` wouldn't work.

We can use `tokenize.generate_tokens()` to demonstrate how the `str` API works:

```py
import tokenize

with tokenize.open('hello.py') as f:
    token_gen = tokenize.generate_tokens(f.readline)
    for token in token_gen:
        print(token)
```

Then fallback to your example to show the usage of the `bytes` API:

```py
import tokenize

with open('hello.py', 'rb') as f:
    token_gen = tokenize.tokenize(f.readline)
    for token in token_gen:
        print(token)
```

matrixise:
Hi @Windsooon 

Could you update your PR with the last comment of @berkerpeksag and with the `with` statement?

Thank you

matrixise (dismissed):
Thanks for your contribution but you have to update your PR with the recommendations of @berkerpeksag

Windsooon:
Hello, @matrixise, I would love to update the example if needed. However, Serhiy Storchaka didn't agree with it

> I do not think a new example is needed. The existing example already demonstrates the use of file's readline method.

https://bugs.python.org/issue36654

csabella:
@Windsooon, any updates?  Thanks!

Windsooon:
Sure, I will fix it in two days. Thank you for reminding.

Windsooon:
Hi, @csabella. I just updated the PR base on @berkerpeksag 's suggestion.

berkerpeksag (approved):
Looks pretty good to me, thank you!

PR #12952, opened 2019-04-25
bpo-36722: Debug build loads libraries built in release mode

vstinner:
In debug build, import now also looks for C extensions compiled in
release mode and for C extensions compiled in the stable ABI.


pitrou (approved):
+1 from me.

vstinner:
Thanks for the review @pitrou!

PR #12953, opened 2019-04-25
bpo-36461: Added the total_time parameter to the timeit.autorange() function.

mangrisano:
* Add target_time parameter in the __init__.
* Add some tests to test when *repeat* is 1 or more.

https://bugs.python.org/issue36461

### DIFF
--- a/Doc/library/timeit.rst
+++ b/Doc/library/timeit.rst
@@ -141,20 +141,22 @@ The module defines three convenience functions and a public class:
             timeit.Timer('for i in range(10): oct(i)', 'gc.enable()').timeit()
 
 
-   .. method:: Timer.autorange(callback=None)
+   .. method:: Timer.autorange(callback=None, total_time=0.2)
 
       Automatically determine how many times to call :meth:`.timeit`.
 
       This is a convenience function that calls :meth:`.timeit` repeatedly
-      so that the total time >= 0.2 second, returning the eventual
+      so that the total time is greater than *total_time*, returning the eventual
       (number of loops, time taken for that number of loops). It calls
       :meth:`.timeit` with increasing numbers from the sequence 1, 2, 5,
-      10, 20, 50, ... until the time taken is at least 0.2 second.
+      10, 20, 50, ... until the time taken is at least *total_time*.
 
       If *callback* is given and is not ``None``, it will be called after
       each trial with two arguments: ``callback(number, time_taken)``.
 
       .. versionadded:: 3.6
+      .. versionchanged:: 3.8
+         The optional parameter *total_time* was added.
 
 
    .. method:: Timer.repeat(repeat=5, number=1000000)
--- a/Doc/library/timeit.rst
+++ b/Doc/library/timeit.rst
@@ -146,7 +146,7 @@ The module defines three convenience functions and a public class:
       Automatically determine how many times to call :meth:`.timeit`.
 
       This is a convenience function that calls :meth:`.timeit` repeatedly
-      so that the total time is greater than *total_time*, returning the eventual
+      so that the total time is greater than or equal to *total_time*, returning the eventual
       (number of loops, time taken for that number of loops). It calls
       :meth:`.timeit` with increasing numbers from the sequence 1, 2, 5,
       10, 20, 50, ... until the time taken is at least *total_time*.
--- a/Doc/library/timeit.rst
+++ b/Doc/library/timeit.rst
@@ -94,7 +94,8 @@ The module defines three convenience functions and a public class:
       :func:`time.perf_counter` is now the default timer.
 
 
-.. class:: Timer(stmt='pass', setup='pass', timer=<timer function>, globals=None)
+.. class:: Timer(stmt='pass', setup='pass', timer=<timer function>, globals=None,
+                    target_time=0.2)
 
    Class for timing execution speed of small code snippets.
 
@@ -119,6 +120,8 @@ The module defines three convenience functions and a public class:
 
    .. versionchanged:: 3.5
       The optional *globals* parameter was added.
+   .. versionchanged: 3.8
+      The optional *target_time* parameter was added.
 
    .. method:: Timer.timeit(number=1000000)
 
@@ -141,22 +144,22 @@ The module defines three convenience functions and a public class:
             timeit.Timer('for i in range(10): oct(i)', 'gc.enable()').timeit()
 
 
-   .. method:: Timer.autorange(callback=None, total_time=0.2)
+   .. method:: Timer.autorange(callback=None, target_time=default_target_time)
 
       Automatically determine how many times to call :meth:`.timeit`.
 
       This is a convenience function that calls :meth:`.timeit` repeatedly
-      so that the total time is greater than or equal to *total_time*, returning the eventual
+      so that the total time is greater than or equal to *target_time*, returning the eventual
       (number of loops, time taken for that number of loops). It calls
       :meth:`.timeit` with increasing numbers from the sequence 1, 2, 5,
-      10, 20, 50, ... until the time taken is at least *total_time*.
+      10, 20, 50, ... until the time taken is at least *target_time*.
 
       If *callback* is given and is not ``None``, it will be called after
       each trial with two arguments: ``callback(number, time_taken)``.
 
       .. versionadded:: 3.6
       .. versionchanged:: 3.8
-         The optional parameter *total_time* was added.
+         The optional *target_time* parameter was added.
 
 
    .. method:: Timer.repeat(repeat=5, number=1000000)
--- a/Doc/library/timeit.rst
+++ b/Doc/library/timeit.rst
@@ -119,7 +119,7 @@ The module defines three convenience functions and a public class:
 
    .. versionchanged:: 3.5
       The optional *globals* parameter was added.
-   .. versionchanged: 3.8
+   .. versionchanged:: 3.8
       The optional *target_time* parameter was added.
 
    .. method:: Timer.timeit(number=1000000)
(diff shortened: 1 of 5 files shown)


the-knights-who-say-ni:

Lib/test/test_timeit.py
```diff
@@ -359,11 +359,27 @@ def autorange(self, seconds_per_increment=1/1024, callback=None):
         t = timeit.Timer(stmt=self.fake_stmt, setup=self.fake_setup, timer=timer)
         return t.autorange(callback)
 
+    def autorange_with_argument(self, total_time, seconds_per_increment=1/1024,
+            callback=None):
```
ezio-melotti: Instead of adding a new method, I would simply add the *total_time* to the `autorange` method defined above.

Lib/test/test_timeit.py
```diff
@@ -359,11 +359,27 @@ def autorange(self, seconds_per_increment=1/1024, callback=None):
...
         num_loops, time_taken = self.autorange()
         self.assertEqual(num_loops, 500)
         self.assertEqual(time_taken, 500/1024)
 
+    def test_autorange_with_argument(self):
+        num_loops, time_taken = self.autorange_with_argument(total_time=0.6)
+        self.assertEqual(num_loops, 1000)
+        self.assertEqual(time_taken, 1000/1024)
+
+    def test_autorange_with_argument(self):
```
ezio-melotti: The name of this test seems to be duplicate, so the previous test will be overwritten.
Instead of using `with_argument`, I think it would be better to say what argument you are testing, e.g. `test_autorange_with_total_time`.
You can also merge the two tests, since they are testing the same thing.

Lib/test/test_timeit.py
```diff
@@ -387,6 +403,44 @@ def callback(a, b):
...
+                    '10 0.010\n'
+                    '20 0.020\n'
+                    '50 0.049\n'
+                    '100 0.098\n'
+                    '200 0.195\n'
+                    '500 0.488\n'
+                    '1000 0.977\n')
+        self.assertEqual(s.getvalue(), expected)
+
+    def test_autorange_with_callback_with_argument(self):
```
ezio-melotti: Same as above regarding the name.
I'm also not sure it's worth duplicating everything just to test 2 different values (0.4 and 0.6).
If testing different values makes sense and doesn't take too much time, you could create an helper method or use a loop (possibly using `subTest`).
Perhaps a test with negative values should be added as well.

ezio-melotti (changes requested):
I left a couple of comments about the tests.

mangrisano:
I've just made the requested changes. Please review.

tirkarthi:
There is also another PR that seems to add some additional changes along with the autorange changes in this PR : https://github.com/python/cpython/pull/12954/

Lib/test/test_timeit.py
```diff
@@ -354,16 +354,22 @@ def test_main_exception_fixed_reps(self):
             s = self.run_main(switches=['-n1', '1/0'])
         self.assert_exc_string(error_stringio.getvalue(), 'ZeroDivisionError')
 
-    def autorange(self, seconds_per_increment=1/1024, callback=None):
+    def autorange(self, seconds_per_increment=1/1024, callback=None,
+            total_time=0.2):
```
tirkarthi: Please indent `total_time` with `self`.

Lib/test/test_timeit.py
```diff
@@ -387,6 +393,26 @@ def callback(a, b):
                     '500 0.488\n')
         self.assertEqual(s.getvalue(), expected)
 
+    def test_autorange_with_callback_and_total_time(self):
+        def callback(a, b):
+            print("{} {:.3f}".format(a, b))
+        with captured_stdout() as s:
+            num_loops, time_taken = self.autorange(callback=callback,
+                    total_time=0.6)
```
tirkarthi: Please indent `total_time` with `callback`.

Lib/test/test_timeit.py
```diff
@@ -387,6 +393,26 @@ def callback(a, b):
...
+        self.assertEqual(num_loops, 1000)
+        self.assertEqual(time_taken, 1000/1024)
+        expected = ('1 0.001\n'
```
tirkarthi: This test imports `dedent` so that might help?

```python
expected = dedent('''\
             1 0.001
             2 0.002
             5 0.005
             10 0.010
             20 0.020
             50 0.049
             100 0.098
             200 0.195
             500 0.488
             1000 0.977
''')
```

Doc/library/timeit.rst
```diff
@@ -137,20 +137,22 @@ The module defines three convenience functions and a public class:
...
 
 
-   .. method:: Timer.autorange(callback=None)
+   .. method:: Timer.autorange(callback=None, total_time=0.2)
 
       Automatically determine how many times to call :meth:`.timeit`.
 
       This is a convenience function that calls :meth:`.timeit` repeatedly
-      so that the total time >= 0.2 second, returning the eventual
+      so that the total time is greater than *total_time*, returning the eventual
```
tirkarthi: "greater than or equal to" as per the code and previous docs.

Lib/test/test_timeit.py
mangrisano: Thanks for the hint.

mangrisano:
I've just read that there's another PR for this issue and it seems to be more complete than mine. What am I suppose to do in this case?

eamanu (commented):
Hello, like say @tirkarthi this PR is related to https://github.com/python/cpython/pull/12954 
and seems more complete that this. 

But, this PR is older (one PR older) and receive more feedback.

mangrisano:
I've just made the changes suggested by Steven. Please, review. Thank you.

mangrisano:

Lib/test/test_timeit.py
```diff
@@ -165,9 +168,13 @@ def repeat(self, stmt, setup, repeat=None, number=None):
...
-        self.assertEqual(self.fake_timer.setup_calls, repeat)
-        self.assertEqual(self.fake_timer.count, repeat * number)
-        self.assertEqual(delta_times, repeat * [float(number)])
+        self.assertEqual(self.fake_timer.setup_calls, repeat if number > 0
+                else 1)
+        # self.assertEqual(self.fake_timer.setup_calls, )
+        self.assertEqual(self.fake_timer.count, (repeat * number) if number
+                else 1)
+        self.assertEqual(delta_times, repeat * [float(number)] if number
+                else (1, 1.0))
```
ezio-melotti: These should be broken after the comma, not before the else, e.g.:
```
self.assertEqual(self.fake_timer.setup_calls,
                 repeat if number > 0 else 1)
```

Lib/test/test_timeit.py
```diff
@@ -208,7 +218,12 @@ def test_repeat_function_zero_reps(self):
     def test_repeat_function_zero_iters(self):
         delta_times = timeit.repeat(self.fake_stmt, self.fake_setup, number=0,
                 timer=FakeTimer())
-        self.assertEqual(delta_times, DEFAULT_REPEAT * [0.0])
+        self.assertEqual(delta_times, (1, 1.0))
+
+    def test_repeat_function_with_iters(self):
+        delta_times = timeit.repeat(self.fake_stmt, self.fake_setup, number=1,
+                timer=FakeTimer())
```
ezio-melotti: `timer` should be indented under `self.fake_stmt`

Lib/test/test_timeit.py
```diff
@@ -354,16 +369,22 @@ def test_main_exception_fixed_reps(self):
             s = self.run_main(switches=['-n1', '1/0'])
         self.assert_exc_string(error_stringio.getvalue(), 'ZeroDivisionError')
 
-    def autorange(self, seconds_per_increment=1/1024, callback=None):
+    def autorange(self, seconds_per_increment=1/1024, callback=None,
+                    target_time=0.2):
```
ezio-melotti: This should be indented under the `self`.

Lib/test/test_timeit.py
```diff
@@ -376,15 +397,39 @@ def callback(a, b):
...
+                    2 0.002
+                    5 0.005
+                    10 0.010
+                    20 0.020
+                    50 0.049
+                    100 0.098
+                    200 0.195
+                    500 0.488
+                    1000 0.977
+        ''')
```
ezio-melotti, line 430: I think it would be better to factor out common code instead of duplicating the previous test method.

Lib/timeit.py
```diff
@@ -169,6 +171,8 @@ def timeit(self, number=default_number):
         to one million.  The main statement, the setup statement and
         the timer function to be used are passed to the constructor.
         """
+        if (number == 0) or number is None:
```
ezio-melotti: No `()` here.

Lib/timeit.py
```diff
@@ -199,14 +203,17 @@ def repeat(self, repeat=default_repeat, number=default_number):
         interested in.  After that, you should look at the entire
         vector and apply common sense rather than statistics.
         """
+        if (number == 0) or number is None:
```
ezio-melotti: No `()` here.

ezio-melotti (changes requested):

Lib/test/test_timeit.py
mangrisano: Ok, I'm going to delete that test.

mangrisano:

mangrisano:
I'm trying to figure out how to write a good test for testing the `autorange` with `callback` and `target_time`.

mangrisano:

mangrisano:
@stevendaprano Could you review this pull request when you're ready? Thank you :)

PR #12954, opened 2019-04-25
gh-80642: timeit - Additional changes for autorange

alessandrocucci:
- make the 0.2s time configurable;
- have `timeit` and `repeat` methods (and functions) fall back 
  on `autorange` if the number is set to 0 or None.

https://bugs.python.org/issue36461

* Issue: gh-80642

### DIFF
--- a/Doc/library/timeit.rst
+++ b/Doc/library/timeit.rst
@@ -58,29 +58,36 @@ Python Interface
 The module defines three convenience functions and a public class:
 
 
-.. function:: timeit(stmt='pass', setup='pass', timer=<default timer>, number=1000000, globals=None)
+.. function:: timeit(stmt='pass', setup='pass', timer=<default timer>, number=1000000, globals=None, max_time_taken=0.2)
 
    Create a :class:`Timer` instance with the given statement, *setup* code and
    *timer* function and run its :meth:`.timeit` method with *number* executions.
    The optional *globals* argument specifies a namespace in which to execute the
-   code.
+   code. If *number* is 0, :meth:`.autorange` method is executed, a convenience
+   function that calls :meth:`.timeit` repeatedly so that the total time >= *max_time_taken* second.
 
    .. versionchanged:: 3.5
       The optional *globals* parameter was added.
 
+   .. versionchanged:: 3.7
+      The optional *max_time_taken* parameter was added.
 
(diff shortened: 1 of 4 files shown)


Doc/library/timeit.rst
```diff
@@ -233,6 +243,12 @@ Where the following options are understood:
 
    .. versionadded:: 3.5
 
+.. cmdoption:: -m, --max_time_taken=M
+
+    calls :meth:`.timeit` repeatedly so that the total time >= *max_time_taken* second
+
+   .. versionadded:: 3.7
```
auvipy: 3.8 as 3.7 is already feature freeze

auvipy (changes requested):

Doc/library/timeit.rst
alessandrocucci: Hello and thanks for the review. 
I wrote 3.7 because it was the lowest version setted in https://bugs.python.org/issue36461 by @csabella 
As you said correctly (like also PEP-537 states) 3.7 is now frozen. I already made the changes.

bsolomon1124:
Was there consideration given to adding `autorange()` as a module-level function, in addition to being an instance method of `Timer` as is added here?

Since `timeit` already uses a ["prebound method pattern"](https://github.com/python/cpython/blob/34ed40f2e56703de04241cbacb306113b59a84f9/Lib/timeit.py#L229) to put `timeit()` and `repeat()` there, maybe `autorange()` belongs with them?

alessandrocucci:
that was not discussed in [.](https://bugs.python.org/issue36461) 
I personally don't think it's necessary, do you have an use case?

bsolomon1124:
Actually, looking at things more carefully, I don't think it's necessary, since it looks like `autorange()` is really just meant to be used in `timeit()`/`repeat()`, which are already available as module-level functions.  So please just disregard my previous comment.

Carreau:
if i understand this correctly `max_time_taken` is an ill-chosen name, it should be `min_time_taken`, shouldn't it ?

alessandrocucci:
There is a proposal to rename it to "target_time", as Steven wrote in https://bugs.python.org/issue36461

rhettinger:
ISTM that almost the entire point of having autorange() was to avoid the need for configuration.

iritkatriel (changes requested):
This has merge conflicts now.

StanFromIreland:
This PR has been stale for over three years, closing in favour of #140283

PR #12955, opened 2019-04-25
gh-62299: Don't set Content-Length for 1xx, 204 and 304 requests

berkerpeksag:
https://bugs.python.org/issue18099

* Issue: gh-62299

### DIFF
--- a/Lib/test/test_wsgiref.py
+++ b/Lib/test/test_wsgiref.py
@@ -299,6 +299,21 @@ class IntegrationTests(TestCase):
         background.join()
         self.assertEqual(received, support.SOCK_MAX_SIZE - 100)
 
+    def test_no_content_length_header_on_304(self):
+        def app(environ, start_response):
+            start_response(
+                '304 Not Modified',
+                [('Date', 'Wed, 24 Dec 2008 13:29:32 GMT')]
+            )
+            return []
+
+        stdout, _ = run_amock(validator(app))
+        self.assertNotIn(b'Content-Length: 0', stdout)
+
+    # TODO: Test POST request + Content-Length: 0.
+    # TODO: Test 204 No Content.
+    # TODO: Test HEAD request.
+
 
 class UtilityTests(TestCase):
 
--- a/Lib/test/test_wsgiref.py
+++ b/Lib/test/test_wsgiref.py
@@ -299,21 +299,6 @@ class IntegrationTests(TestCase):
         background.join()
         self.assertEqual(received, support.SOCK_MAX_SIZE - 100)
 
-    def test_no_content_length_header_on_304(self):
-        def app(environ, start_response):
-            start_response(
-                '304 Not Modified',
-                [('Date', 'Wed, 24 Dec 2008 13:29:32 GMT')]
-            )
-            return []
-
-        stdout, _ = run_amock(validator(app))
-        self.assertNotIn(b'Content-Length: 0', stdout)
-
-    # TODO: Test POST request + Content-Length: 0.
-    # TODO: Test 204 No Content.
-    # TODO: Test HEAD request.
-
 
 class UtilityTests(TestCase):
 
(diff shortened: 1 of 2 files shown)


Lib/test/test_wsgiref.py
```diff
@@ -299,6 +299,21 @@ def run_client():
...
+    def test_no_content_length_header_on_304(self):
+        def app(environ, start_response):
+            start_response(
+                '304 Not Modified',
+                [('Date', 'Wed, 24 Dec 2008 13:29:32 GMT')]
+            )
+            return []
+
+        stdout, _ = run_amock(validator(app))
+        self.assertNotIn(b'Content-Length: 0', stdout)
```
vadmium: Better to check for 'Content-Length:' without the zero. No Content-Length should be sent at all, unless the application sends the true value.

Lib/wsgiref/handlers.py
```diff
@@ -189,15 +189,29 @@ def get_scheme(self):
...
+            return
         try:
             blocks = len(self.result)
         except (TypeError,AttributeError,NotImplementedError):
             pass
         else:
-            if blocks==1:
+            if blocks == 0:
+                # Only zero Content-Length if not set by the application and
+                # have appropriote HTTP status so that HEAD requests can be
```
vadmium: Spelling: appropriate

Lib/wsgiref/handlers.py
```diff
@@ -189,15 +189,29 @@ def get_scheme(self):
 
 
     def set_content_length(self):
-        """Compute Content-Length or switch to chunked encoding if possible"""
+        """Compute and set Content-Length header"""
+        # See 3.3.2 of RFC 7230 for more details.
+        cannot_have_content_length = self.status.startswith((
+            '100',  # Continue
+            '101',  # Switching Protocols
+            '102',  # Processing
```
vadmium, line 197: Christian’s original patch made more sense to me, just checking for the first digit being 1, rather than limiting to these three codes. In the HTTP 1.1 RFC <https://tools.ietf.org/html/rfc7230#page-31>, it says “A server must not send a Content-Length header field in any response with a status code of 1xx (Informational)”.

Lib/wsgiref/handlers.py
```diff
@@ -305,9 +320,6 @@ def sendfile(self):
     def finish_content(self):
         """Ensure headers and content have both been sent"""
         if not self.headers_sent:
-            # Only zero Content-Length if not set by the application (so
-            # that HEAD requests can be satisfied properly, see #3839)
-            self.headers.setdefault('Content-Length', "0")
```
vadmium, line 310: Perhaps if you set _self.result_ to an empty sequence here, then “Content-Length: 0” would continue to be added for empty iterables that don’t support _len_.

Lib/wsgiref/handlers.py
```diff
@@ -189,15 +189,29 @@ def get_scheme(self):
...
             blocks = len(self.result)
         except (TypeError,AttributeError,NotImplementedError):
             pass
         else:
-            if blocks==1:
+            if blocks == 0:
+                # Only zero Content-Length if not set by the application and
+                # have appropriote HTTP status so that HEAD requests can be
+                # satisfied properly. (bpo-3839)
+                self.headers.setdefault('Content-Length', "0")
```
vadmium: _set_content_length_ is only called when the application didn’t set Content-Length itself. [Edit: In other words, using _setdefault_ is redundant.] Maybe I could come up with a better suggestion with more time, but not atm.

vadmium (commented):
Hi @berkerpeksag, I wrote some comments on Sunday, but I didn’t have time to finish looking at this. Posting what I got so far in case it’s useful.

Lib/wsgiref/handlers.py
berkerpeksag: I tried that first but I think I forgot to drop the `if blocks == 1` check in `set_content_length`. I think setting `self.result = []` should work and it would even help to simplify `set_content_length`.

Lib/wsgiref/handlers.py
berkerpeksag: Ironically, I decided to list all three codes explicitly after I read that section because I had to Google them :) Both styles fine for me.

Lib/wsgiref/handlers.py
berkerpeksag: Done.

berkerpeksag:
@vadmium do you have any further comments? I'd like to get this merged this weekend if possible.

PR #12957, opened 2019-04-25
bpo-36722: Add What's New entry for debug ABI

vstinner:

### DIFF
--- a/Doc/whatsnew/3.8.rst
+++ b/Doc/whatsnew/3.8.rst
@@ -98,6 +98,30 @@ subdirectories).
 
 (Contributed by Carl Meyer in :issue:`33499`.)
 
+Debug build uses the same ABI than release build
+------------------------------------------------
+
+Python now uses the same ABI when built in release or debug mode. On Unix, when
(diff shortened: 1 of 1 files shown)


vstinner:
@elprans @pganssle: Would you mind to review this change? I'm not sure about the length of the text. I chose to document the 3 changes in 3 paragraph since each change is very subtle and it might help readers to have a link to the bpo issue.

Doc/whatsnew/3.8.rst
```diff
@@ -98,6 +98,30 @@ subdirectories).
 
 (Contributed by Carl Meyer in :issue:`33499`.)
 
+Debug build uses the same ABI than release build
```
auscompgeek: s/than/as/?

vstinner: Right, fixed.

vstinner:
Oh. I didn't know that there is a new Python 3.8 release (alpha4) next monday. I prefer to merge this What's New In Python 3.8 soon and rework the text later, than not document this change. So I merged my change.

PR #12960, opened 2019-04-26
bpo-36725: regrtest: add TestResult type

vstinner:
* Add TestResult and MultiprocessResult types to ensure that results
  always have the same fields.
* runtest() now handles KeyboardInterrupt
* accumulate_result() and format_test_result() now takes a TestResult
* cleanup_test_droppings() is now called by runtest() and mark the
  test as ENV_CHANGED if the test leaks support.TESTFN file.
* runtest() now includes code "around" the test in the test timing
* Add print_warning() in test.libregrtest.utils to standardize how
  libregrtest logs warnings to ease parsing the test output.
* support.unload() is now called with abstest rather than test_name
* Rename 'test' variable/parameter to 'test_name'
* dash_R(): remove unused the_module parameter
* Remove unused imports

https://bugs.python.org/issue36725

### DIFF
--- a/Lib/test/libregrtest/main.py
+++ b/Lib/test/libregrtest/main.py
@@ -105,26 +105,30 @@ class Regrtest:
         # used by --junit-xml
         self.testsuite_xml = None
 
-    def accumulate_result(self, test, result):
-        ok, test_time, xml_data = result
+    def accumulate_result(self, result):
+        test_name = result.test_name
+        ok = result.result
+
         if ok not in (CHILD_ERROR, INTERRUPTED):
-            self.test_times.append((test_time, test))
+            self.test_times.append((result.test_time, test_name))
+
         if ok == PASSED:
-            self.good.append(test)
+            self.good.append(test_name)
         elif ok in (FAILED, CHILD_ERROR):
-            self.bad.append(test)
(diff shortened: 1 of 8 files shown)


db3l:
In looking at this I don't think build 2469 on the buildbot should have been marked failed, but rather been a warning, as the re-run of test_threading seems to have been successful.  But it looks like with this change, both runs were considered failures.  It seems to be affecting other builders too (e.g., the last two builds (118 and 71) on cstratak-fedora).

It looks like there's a discrepancy in field order in the new TestResult namedtuple and the test re-run logic.  The re-run logic assumes [0] is the result, but it's at [1] in the tuple.  However, the new runtest function does documents the result at [0], but it's missing the test_name field.

vstinner:
@db3l: Oh, your're right. I tried to write a much tests as possible in test_regrtest, but this case wasn't properly tested. I wrote PR #12964 to fix the regression but also to add a new unit test for it.

vstinner:
> https://buildbot.python.org/all/#builders/3/builds/2469

```
test_default_timeout (test.test_threading.BarrierTests)
```

That's https://bugs.python.org/issue11871 : someone should have a look :-)

PR #12885, opened 2019-04-20
bpo-36676: Namespace prefix aware parsing support for the ET.XMLParser target

scoder:
This is based on [bpo-36673](https://bugs.python.org/issue36673) (GH-12883), thus the overlap of the first commit.

https://bugs.python.org/issue36676

PR #12966, opened 2019-04-26
bpo-13611: C14N 2.0 implementation for ElementTree

scoder:
This is based on GH-12883 and GH-12885.
The only changes that are relevant to this PR are in `ElementTree.py` and the C14N test case in `test_xml_etree.py`, plus the added C14N 2.0 test files.

Documentation missing, pending a review of the overall API.

https://bugs.python.org/issue13611


Lib/xml/etree/ElementTree.py
```diff
@@ -1633,14 +1712,347 @@ def close(self):
...
+    if xml_data is not None:
+        parser.feed(xml_data)
+        parser.close()
+    elif from_file is not None:
+        parse(from_file, parser=parser)
+
+    return sio.getvalue() if sio is not None else None
+
+
+_looks_like_prefix_name = re.compile('^\w+:\w+$', re.UNICODE).match
```
tirkarthi: This might generate a `SyntaxWarning`. I think prefixing 'r' helps here : `r'^\w+:\w+$'`

```
➜  cpython git:(pr_12966) ✗ ./python.exe -Wall -c 'import xml.etree.ElementTree'
/Users/karthikeyansingaravelan/stuff/python/cpython/Lib/xml/etree/ElementTree.py:1748: SyntaxWarning: invalid escape sequence \w
  _looks_like_prefix_name = re.compile('^\w+:\w+$', re.UNICODE).match
```

tirkarthi (changes requested):

Lib/xml/etree/ElementTree.py
scoder: Fixed, thanks.

ZackerySpytz:
I tested this PR locally, and there seems to be reference leaks. The following is the output of `./python -m test -R 3:3 test_xml_etree_c`:

```
Run tests sequentially
0:00:00 load avg: 1.09 [1/1] test_xml_etree_c
beginning 6 repetitions
123456
......
test_xml_etree_c leaked [208, 208, 208] references, sum=624
test_xml_etree_c leaked [51, 51, 51] memory blocks, sum=153
test_xml_etree_c failed

== Tests result: FAILURE ==

1 test failed:
    test_xml_etree_c

Total duration: 2 sec 801 ms
Tests result: FAILURE
```

scoder:
@ZackerySpytz, thanks for testing (and providing the command :) )
I'll look into it.

Modules/_elementtree.c
```diff
@@ -3783,6 +3787,8 @@ xmlparser_gc_traverse(XMLParserObject *self, visitproc visit, void *arg)
     Py_VISIT(self->handle_data);
     Py_VISIT(self->handle_start);
     Py_VISIT(self->handle_start_ns);
+    Py_VISIT(self->handle_end_ns);
+    Py_VISIT(self->handle_doctype);
```
scoder: This was actually missing before. Since it refers to an arbitrary user provided (callable) object, it must be visible to the GC.

PR #12969, opened 2019-04-26
bpo-36733: Fix PYTHONPATH for make regen-add

vstinner:
Add PYTHONPATH=$(srcdir) to run $(PYTHON_FOR_REGEN) -m Parser.pgen,
so it's possible to build Python from a different directory.

https://bugs.python.org/issue36733

### DIFF
--- a/Makefile.pre.in
+++ b/Makefile.pre.in
@@ -791,7 +791,7 @@ regen-grammar: regen-token
 	# Regenerate Include/graminit.h and Python/graminit.c
 	# from Grammar/Grammar using pgen
 	@$(MKDIR_P) Include
-	$(PYTHON_FOR_REGEN) -m Parser.pgen $(srcdir)/Grammar/Grammar \
+	PYTHONPATH=$(srcdir) $(PYTHON_FOR_REGEN) -m Parser.pgen $(srcdir)/Grammar/Grammar \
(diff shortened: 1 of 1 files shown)


hroncok:
This is more likely a workaround. Let me see what changed the behavior before we jump to this solution.

vstinner:
> This is more likely a workaround.

Why? For me, it's the right fix.

>  Let me see what changed the behavior before we jump to this solution.

Previously (1 month ago?), make regen-all didn't use Parser.pgen: https://bugs.python.org/issue36733#msg340903

hroncok:
I take my statement back. This is the a proper solution, we don't need to copy around pure Python code just to be able to import it.

What keeps me confused is the empty Parser directory created by configure. Should it no longer be created?

hroncok (approved):
At the end, something is indeed compiled to that directory. So I guess it should remain there.

This fixes the issue.

vstinner:
> I take my statement back. This is the a proper solution, we don't need to copy around pure Python code just to be able to import it.

Thanks for the review.

> What keeps me confused is the empty Parser directory created by configure. Should it no longer be created?

configure creates the empty mybuild/Parser subdirectory. I don't know why. At least, it doesn't prevent to build Python :-)

PR #12971, opened 2019-04-26
bpo-21536: Revert Makefile change on python-config

vstinner:
Misc/python-config.sh lives in the build directory, not in the source
directory.


vstinner:
When I build Python of the source tree, I can reproduce the bug without my change:

```
make: *** No rule to make target '/path_to_build_dir/Misc/python-config.sh', needed by 'python-config'.  Stop
```

I confirm with manual test that my change fix this use case.

PR #12973, opened 2019-04-26
bpo-36722: Don't define ALT_SOABI for Py_TRACE_REFS build

vstinner:
Py_TRACE_REFS ABI is incompatible with release and debug (Py_DEBUG)
ABI.

### DIFF
--- a/Include/pyport.h
+++ b/Include/pyport.h
@@ -797,9 +797,9 @@ extern _invalid_parameter_handler _Py_silent_invalid_parameter_handler;
 #endif /* Py_BUILD_CORE */
 
 #ifdef __ANDROID__
-/* The Android langinfo.h header is not used. */
-#undef HAVE_LANGINFO_H
-#undef CODESET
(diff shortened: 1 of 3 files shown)


vstinner:
I checked manually pyconfig.h in these 3 configurations. It now works as expected. Moreover, pyport.h ensures that ALT_SOABI is not defined when Py_TRACE_REFS is defined, just in case.

./configure:

```
/* #undef ALT_SOABI */
/* #undef Py_DEBUG */
/* #undef Py_TRACE_REFS */
```

./configure --with-pydebug:

```
#define ALT_SOABI "cpython-38-x86_64-linux-gnu"
#define Py_DEBUG 1
/* #undef Py_TRACE_REFS */
```

./configure --with-trace-refs:

```
/* #undef ALT_SOABI */
/* #undef Py_DEBUG */
#define Py_TRACE_REFS 1
```

ALT_SOABI is only defined when Py_DEBUG is defined, but not when Py_TRACE_REFS is defined: as expected.

PR #12974, opened 2019-04-26
bpo-36715: Add usage note for dict.fromkeys()

rhettinger:
https://bugs.python.org/issue36715

### DIFF
--- a/Doc/library/stdtypes.rst
+++ b/Doc/library/stdtypes.rst
@@ -4254,7 +4254,10 @@ pairs within braces, for example: ``{'jack': 4098, 'sjoerd': 4127}`` or ``{4098:
       Create a new dictionary with keys from *iterable* and values set to *value*.
 
       :meth:`fromkeys` is a class method that returns a new dictionary. *value*
-      defaults to ``None``.
+      defaults to ``None``.  The same instance of *value* is used for every
+      key, so it generally doesn't make sense for *value* to be a mutable object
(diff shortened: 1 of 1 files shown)


eamanu (approved):
LGTM

pganssle (approved):
Hm. This seems a bit obvious to me, since the only other way this could work is if `fromkeys` were making a full (possibly deep) copy of `value` for each key.

However, I suspect people like me are not the target audience for this, and honestly this would only be a problem if every method were getting tons of warnings about unexpected language features cluttering up the documentation. That clearly hasn't happened in this case, and if it ever does happen, the documentation can probably just be rearranged to be more accessible.

Doc/library/stdtypes.rst
```diff
@@ -4254,7 +4254,10 @@ pairs within braces, for example: ``{'jack': 4098, 'sjoerd': 4127}`` or ``{4098:
       Create a new dictionary with keys from *iterable* and values set to *value*.
 
       :meth:`fromkeys` is a class method that returns a new dictionary. *value*
-      defaults to ``None``.
+      defaults to ``None``.  The same instance of *value* is used for every
```
terryjreedy: Since value is not a class, 'instance of value' does not make sense.
```suggestion
      defaults to ``None``.  The one *value* object is used for every
```

terryjreedy (changes requested):
Approve except for minor suggestion.

terryjreedy (approved):

PR #12976, opened 2019-04-26
bpo-22640: Add silent mode to py_compile

nanjekyejoannah:
I have added silent mode to py_compile.

https://bugs.python.org/issue22640

### DIFF
--- a/Doc/library/py_compile.rst
+++ b/Doc/library/py_compile.rst
@@ -42,6 +42,13 @@ byte-code cache files in the directory containing the source code.
    is raised.  This function returns the path to byte-compiled file, i.e.
    whatever *cfile* value was used.
 
+   The *doraise* and *quiet* arguments determine how errors are handled while
+   compiling file. If *quiet* is 0 or 1, and *doraise* is false, the default
+   behaviour is enabled: an error string is written to ``sys.stderr``, and the
+   function returns ``None`` instead of a path. If *doraise* is true,
+   a :exc:`PyCompileError` is raised instead. However if *quiet* is 2,
+   no message is written, and *doraise* has no effect.
+
    If the path that *cfile* becomes (either explicitly specified or computed)
    is a symlink or non-regular file, :exc:`FileExistsError` will be raised.
    This is to act as a warning that import will turn those paths into regular
@@ -81,6 +88,10 @@ byte-code cache files in the directory containing the source code.
       The :envvar:`SOURCE_DATE_EPOCH` environment variable no longer
       overrides the value of the *invalidation_mode* argument, and determines
       its default value instead.
+
+   .. versionchanged:: 3.8
+      The *quiet* parameter was added to support silent mode. It returns full
+      output with false or 0, errors only with 1, and no output with 2.
 
 
 .. class:: PycInvalidationMode
--- a/Doc/library/py_compile.rst
+++ b/Doc/library/py_compile.rst
@@ -90,8 +90,7 @@ byte-code cache files in the directory containing the source code.
       its default value instead.
 
    .. versionchanged:: 3.8
-      The *quiet* parameter was added to support silent mode. It returns full
-      output with false or 0, errors only with 1, and no output with 2.
+      The *quiet* parameter was added.
 
 
 .. class:: PycInvalidationMode
--- a/Doc/whatsnew/3.8.rst
+++ b/Doc/whatsnew/3.8.rst
@@ -231,6 +231,22 @@ New Modules
 Improved Modules
 ================
 
+* The :meth:`_asdict()` method for :func:`collections.namedtuple` now returns
+  a :class:`dict` instead of a :class:`collections.OrderedDict`.  This works because
+  regular dicts have guaranteed ordering in since Python 3.7.  If the extra
+  features of :class:`OrderedDict` are required, the suggested remediation is
+  to cast the result to the desired type: ``OrderedDict(nt._asdict())``.
+  (Contributed by Raymond Hettinger in :issue:`35864`.)
+
+* The :mod:`unicodedata` module has been upgraded to use the `Unicode 12.0.0
+  <http://blog.unicode.org/2019/03/announcing-unicode-standard-version-120.html>`_
+  release.
+
+* The :meth:`py_compile.compile` method now supports silent mode. (Contributed by
+  Joannah Nanjekye in :issue:`22640`.)
+
+
+=======
 
 asyncio
 -------
--- a/Doc/whatsnew/3.8.rst
+++ b/Doc/whatsnew/3.8.rst
@@ -231,17 +231,6 @@ New Modules
 Improved Modules
 ================
 
-* The :meth:`_asdict()` method for :func:`collections.namedtuple` now returns
-  a :class:`dict` instead of a :class:`collections.OrderedDict`.  This works because
-  regular dicts have guaranteed ordering in since Python 3.7.  If the extra
-  features of :class:`OrderedDict` are required, the suggested remediation is
-  to cast the result to the desired type: ``OrderedDict(nt._asdict())``.
-  (Contributed by Raymond Hettinger in :issue:`35864`.)
-
-* The :mod:`unicodedata` module has been upgraded to use the `Unicode 12.0.0
-  <http://blog.unicode.org/2019/03/announcing-unicode-standard-version-120.html>`_
-  release.
-
 * The :meth:`py_compile.compile` method now supports silent mode. (Contributed by
   Joannah Nanjekye in :issue:`22640`.)
 
--- a/Doc/whatsnew/3.8.rst
+++ b/Doc/whatsnew/3.8.rst
@@ -231,9 +231,6 @@ New Modules
 Improved Modules
 ================
 
-* The :meth:`py_compile.compile` method now supports silent mode. (Contributed by
-  Joannah Nanjekye in :issue:`22640`.)
-
 
 =======
 
@@ -405,6 +402,13 @@ plistlib
 Added new :class:`plistlib.UID` and enabled support for reading and writing
 NSKeyedArchiver-encoded binary plists.
 (Contributed by Jon Janzen in :issue:`26707`.)
+
+
+py_compile
+----------
+
+* The :meth:`py_compile.compile` method now supports silent mode.
+  (Contributed by Joannah Nanjekye in :issue:`22640`.)
 
 
 socket
--- a/Doc/whatsnew/3.8.rst
+++ b/Doc/whatsnew/3.8.rst
@@ -405,7 +405,7 @@ NSKeyedArchiver-encoded binary plists.
 py_compile
 ----------
 
-The :meth:`py_compile.compile` method now supports silent mode.
+The :func:`py_compile.compile` method now supports silent mode.
 (Contributed by Joannah Nanjekye in :issue:`22640`.)
 
 
--- a/Doc/whatsnew/3.8.rst
+++ b/Doc/whatsnew/3.8.rst
@@ -405,7 +405,7 @@ NSKeyedArchiver-encoded binary plists.
 py_compile
 ----------
 
-:meth:`py_compile.compile` method now supports silent mode.
+The :meth:`py_compile.compile` method now supports silent mode.
 (Contributed by Joannah Nanjekye in :issue:`22640`.)
 
 
--- a/Doc/whatsnew/3.8.rst
+++ b/Doc/whatsnew/3.8.rst
@@ -407,8 +407,8 @@ NSKeyedArchiver-encoded binary plists.
 py_compile
 ----------
 
-* The :meth:`py_compile.compile` method now supports silent mode.
-  (Contributed by Joannah Nanjekye in :issue:`22640`.)
+The :meth:`py_compile.compile` method now supports silent mode.
+(Contributed by Joannah Nanjekye in :issue:`22640`.)
 
 
 socket
--- a/Doc/whatsnew/3.8.rst
+++ b/Doc/whatsnew/3.8.rst
@@ -405,7 +405,7 @@ NSKeyedArchiver-encoded binary plists.
 py_compile
 ----------
 
-The :func:`py_compile.compile` now supports silent mode.
+:func:`py_compile.compile` now supports silent mode.
 (Contributed by Joannah Nanjekye in :issue:`22640`.)
 
 
--- a/Doc/whatsnew/3.8.rst
+++ b/Doc/whatsnew/3.8.rst
@@ -232,8 +232,6 @@ Improved Modules
 ================
 
 
-=======
-
 asyncio
 -------
 
--- a/Doc/whatsnew/3.8.rst
+++ b/Doc/whatsnew/3.8.rst
@@ -405,7 +405,7 @@ NSKeyedArchiver-encoded binary plists.
 py_compile
 ----------
 
-The :func:`py_compile.compile` method now supports silent mode.
+The :func:`py_compile.compile` now supports silent mode.
 (Contributed by Joannah Nanjekye in :issue:`22640`.)
 
 
--- a/Doc/whatsnew/3.8.rst
+++ b/Doc/whatsnew/3.8.rst
@@ -407,7 +407,7 @@ NSKeyedArchiver-encoded binary plists.
 py_compile
 ----------
 
-The :meth:`py_compile.compile` method now supports silent mode.
+:meth:`py_compile.compile` method now supports silent mode.
 (Contributed by Joannah Nanjekye in :issue:`22640`.)
 
 
--- a/Lib/py_compile.py
+++ b/Lib/py_compile.py
@@ -77,7 +77,7 @@ def _get_default_invalidation_mode():
 
 
 def compile(file, cfile=None, dfile=None, doraise=False, optimize=-1,
-            invalidation_mode=None):
+            invalidation_mode=None, quiet=0):
     """Byte-compile one Python source file to Python bytecode.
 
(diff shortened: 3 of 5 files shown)


Lib/py_compile.py
```diff
@@ -76,7 +76,7 @@ def _get_default_invalidation_mode():
         return PycInvalidationMode.TIMESTAMP
 
 
-def compile(file, cfile=None, dfile=None, doraise=False, optimize=-1,
+def compile(file, cfile=None, dfile=None, doraise=False, quiet=0, optimize=-1,
```
berkerpeksag: Since these aren't keyword-only arguments, users can invoke `compile()` like `compile('...', '...', '...', True, 1)`, so if we add `quiet` after `doraise`, we may break their code. It would be better if we could add `quiet` after `invalidation_mode`.

Lib/test/test_py_compile.py
```diff
@@ -192,6 +192,14 @@ def test_invalidation_mode(self):
                 fp.read(), 'test', {})
         self.assertEqual(flags, 0b1)
 
+    def test_quiet(self):
+        bad_coding = os.path.join(os.path.dirname(__file__), 'bad_coding2.py')
+        with support.captured_stderr():
+            self.assertIsNone(py_compile.compile(bad_coding, doraise=False, quiet=2))
```
berkerpeksag, line 198: We can drop `assertIsNone`s since the return value of `py_compile.compile()` isn't important.

Lib/test/test_py_compile.py
```diff
@@ -192,6 +192,14 @@ def test_invalidation_mode(self):
                 fp.read(), 'test', {})
         self.assertEqual(flags, 0b1)
 
+    def test_quiet(self):
+        bad_coding = os.path.join(os.path.dirname(__file__), 'bad_coding2.py')
+        with support.captured_stderr():
```
berkerpeksag:
```suggestion
        with support.captured_stderr() as stderr:
```

Then we can add something like

```py
self.assertEqual(stderr.getvalue(), '')
```

to test the `quiet` option.

Misc/NEWS.d/next/Library/2019-04-26-22-13-26.bpo-22640.p3rheW.rst
```diff
@@ -0,0 +1 @@
+`py_compile` now supports silent mode.
```
berkerpeksag: :func:\`py_compile.compile\`

berkerpeksag: And please add "Patch by Your Name.".

Doc/whatsnew/3.8.rst
```diff
@@ -769,6 +769,8 @@ Changes in the Python API
 * The :meth:`imap.IMAP4.logout` method no longer ignores silently arbitrary
   exceptions.
 
+* The :meth:`py_compile.compile` now supports silent mode.
```
berkerpeksag: Please add:

```rst
(Contributed by Your Name in :issue:`22640`.)
```

Doc/library/py_compile.rst
```diff
@@ -82,6 +83,10 @@ byte-code cache files in the directory containing the source code.
       overrides the value of the *invalidation_mode* argument, and determines
       its default value instead.
 
+   .. versionchanged:: 3.8.0
```
berkerpeksag: Nit: 3.8.0 -> 3.8

Doc/library/py_compile.rst
```diff
@@ -82,6 +83,10 @@ byte-code cache files in the directory containing the source code.
...
+      The *quiet* parameter was added to support silent mode. The function
```
berkerpeksag: The part that starts with "The function" can be deleted.

Doc/library/py_compile.rst
```diff
@@ -40,7 +40,8 @@ byte-code cache files in the directory containing the source code.
    when an error is encountered while compiling *file*. If *doraise* is false
    (the default), an error string is written to ``sys.stderr``, but no exception
    is raised.  This function returns the path to byte-compiled file, i.e.
-   whatever *cfile* value was used.
+   whatever *cfile* value was used. If *quite* is False or 0, a full output will
```
berkerpeksag: Nit: False -> false

Doc/library/py_compile.rst
```diff
@@ -40,7 +40,8 @@ byte-code cache files in the directory containing the source code.
...
+   be returned, onlly errors when 1 and nothing when 2.
```
berkerpeksag: Typo: onlly -> only

berkerpeksag (changes requested):
Please wait for @brettcannon's feedback before addressing my review comments. I opened bpo-22640 years ago and unfortunately I forgot to add Brett to nosy list.

Doc/library/py_compile.rst
```diff
@@ -27,7 +27,7 @@ byte-code cache files in the directory containing the source code.
    Exception raised when an error occurs while attempting to compile the file.
 
 
-.. function:: compile(file, cfile=None, dfile=None, doraise=False, optimize=-1, invalidation_mode=PycInvalidationMode.TIMESTAMP)
+.. function:: compile(file, cfile=None, dfile=None, doraise=False, quite=0, optimize=-1, invalidation_mode=PycInvalidationMode.TIMESTAMP)
```
ZackerySpytz: It should be "quiet", not "quite".

Doc/library/py_compile.rst
ZackerySpytz: It should be "quiet", not "quite".

ZackerySpytz (commented):
Do not use "quite" instead of "quiet", please.

Lib/py_compile.py
```diff
@@ -191,21 +194,24 @@ def main(args=None):
...
     else:
         for filename in args:
             try:
-                compile(filename, doraise=True)
+                compile(filename, doraise=True, quiet=0)
             except PyCompileError as error:
                 # return value to indicate at least one failure
                 rv = 1
-                sys.stderr.write("%s\n" % error.msg)
+                if args.quiet < 2:
```
tirkarthi: `args` is a list of filenames and `py_compile` doesn't seem to use argparse. Does this throw an `AttributeError`?

Lib/py_compile.py
```diff
@@ -191,21 +194,24 @@ def main(args=None):
...
             except OSError as error:
                 rv = 1
-                sys.stderr.write("%s\n" % error)
+                if args.quiet < 2:
+                    sys.stderr.write("%s\n" % error)
     else:
         for filename in args:
             try:
-                compile(filename, doraise=True)
+                compile(filename, doraise=True, quiet=0)
```
tirkarthi: `quiet` defaults to zero in the function definition so does this needs to be passed?

Lib/py_compile.py
```diff
@@ -143,10 +145,11 @@ def compile(file, cfile=None, dfile=None, doraise=False, optimize=-1,
                                      _optimize=optimize)
     except Exception as err:
         py_exc = PyCompileError(err.__class__, err, dfile or file)
-        if doraise:
+        if doraise and quiet < 2:
             raise py_exc
         else:
-            sys.stderr.write(py_exc.msg + '\n')
+            if quiet < 2:
```
vadmium: It might make more sense to check _quiet_ once, before checking _doraise_

Lib/test/test_py_compile.py
vadmium: I think it is relevant, because the current documentation says the return value should be a path, not None. In reality, returning None is a special case when the compilation fails but does not raise an exception.

Doc/library/py_compile.rst
vadmium: The _quiet_ mode should not affect the function’s return value. And there is no difference between “full output” and “only errors”. Based on your implementation, I would move the existing text for _doraise_ to a new paragraph, something like:

The _doraise_ and _quiet_ arguments determine how errors are handled while compiling *file*. If _quiet_ is 0 or 1, and *doraise* is false, the default behaviour is enabled: an error string is written to ``sys.stderr``, and the function returns None instead of a path. If *doraise* is true, a ``:exc:`PyCompileError` `` is raised instead. However if _quiet_ is 2, no message is written, and _doraise_ has no effect.

Doc/whatsnew/3.8.rst
vadmium: Assuming we fix up the compatibility problem, I think this belongs under “Improved” Modules, not the Porting section.

The sentence needs a noun; I suggest

The ``:func:`py_compile.compile` `` function . . .

nanjekyejoannah:
> Please wait for @brettcannon's feedback before addressing my review comments. I opened bpo-22640 years ago and unfortunately I forgot to add Brett to nosy list.

Noted.

brettcannon:
I'm fine with the idea.

nanjekyejoannah:
Great, I will finish then address the reviews on this PR as soon as I can.

Lib/py_compile.py
nanjekyejoannah: I know what you mean. But when I tried to do this:

```
if quiet < 2:
    if doraise:
        raise py_exc
    else:
        sys.stderr.write(py_exc.msg + '\n')
        return
```
I got an  UnboundLocalError error.

Lib/py_compile.py
nanjekyejoannah: Nope, you are right on this after second thought.

nanjekyejoannah:

Doc/library/py_compile.rst
```diff
@@ -42,6 +42,13 @@ byte-code cache files in the directory containing the source code.
    is raised.  This function returns the path to byte-compiled file, i.e.
    whatever *cfile* value was used.
 
+   The doraise and quiet arguments determine how errors are handled while
```
berkerpeksag: doraise -> \*doraise\*

quiet -> \*quiet\*

Same for other usages of doraise and quiet.

Doc/library/py_compile.rst
```diff
@@ -42,6 +42,13 @@ byte-code cache files in the directory containing the source code.
...
+   compiling file. If quiet is 0 or 1, and doraise is false, the default
+   behaviour is enabled: an error string is written to sys.stderr, and the
```
berkerpeksag: sys.stderr -> \`\`sys.stderr\`\` to be consistent with rest of the documentation.

Doc/library/py_compile.rst
```diff
@@ -42,6 +42,13 @@ byte-code cache files in the directory containing the source code.
...
+   function returns None instead of a path. If doraise is true,
```
berkerpeksag: None -> \`\`None\`\`

Doc/library/py_compile.rst
```diff
@@ -82,6 +89,9 @@ byte-code cache files in the directory containing the source code.
       overrides the value of the *invalidation_mode* argument, and determines
       its default value instead.
 
+   .. versionchanged:: 3.8
+      The *quiet* parameter was added to support silent mode. It returns full
+      output with False or 0, errors only with 1, and no output with 2.
```
berkerpeksag: False -> false

Doc/whatsnew/3.8.rst
```diff
@@ -230,6 +230,21 @@ New Modules
 Improved Modules
 ================
 
+* The :meth:`_asdict()` method for :func:`collections.namedtuple` now returns
```
berkerpeksag: These two items shouldn't be in this PR. Perhaps you need to rebase your branch again?

Doc/whatsnew/3.8.rst
```diff
@@ -230,6 +230,21 @@ New Modules
...
+  features of :class:`OrderedDict` are required, the suggested remediation is
+  to cast the result to the desired type: ``OrderedDict(nt._asdict())``.
+  (Contributed by Raymond Hettinger in :issue:`35864`.)
+
+* The :mod:`unicodedata` module has been upgraded to use the `Unicode 12.0.0
+  <http://blog.unicode.org/2019/03/announcing-unicode-standard-version-120.html>`_
+  release.
+
+* The :meth:`py_compile.compile` method now supports silent mode. (Contributed by
+  Joannah Nanjekye in :issue:`22640`)
```
berkerpeksag: Nit: Missing a full stop:

```rst
(Contributed by Joannah Nanjekye in :issue:`22640`.)
```

Lib/py_compile.py
berkerpeksag: I think Martin meant something like:

```diff
diff --git a/Lib/py_compile.py b/Lib/py_compile.py
index a3278a4713..21736896af 100644
--- a/Lib/py_compile.py
+++ b/Lib/py_compile.py
@@ -145,12 +145,12 @@ def compile(file, cfile=None, dfile=None, doraise=False, optimize=-1,
                                      _optimize=optimize)
     except Exception as err:
         py_exc = PyCompileError(err.__class__, err, dfile or file)
-        if doraise and quiet < 2:
-            raise py_exc
-        else:
-            if quiet < 2:
+        if quiet < 2:
+            if doraise:
+                raise py_exc
+            else:
                 sys.stderr.write(py_exc.msg + '\n')
-            return
+        return
     try:
         dirname = os.path.dirname(cfile)
         if dirname:
```

Could you share the traceback?

berkerpeksag (changes requested):
This looks good to me, thank you! I just left mostly trivial comments.

You may also want to check documentation changes with a native English speaker. I'm not a native speaker, so they may help us to simply documentation changes :)

the-knights-who-say-ni:

nanjekyejoannah:
Yo, this was a git mess. Am working on bringing sanity back.

Lib/py_compile.py
nanjekyejoannah: Got it, I was returning wrongly. It is now fine.

nanjekyejoannah:
I think this is okay. I have made the requested changes; please review again.

nanjekyejoannah:
@berkerpeksag  PTAL when you can.

Doc/whatsnew/3.8.rst
```diff
@@ -231,6 +231,11 @@ New Modules
 Improved Modules
 ================
 
+* The :meth:`py_compile.compile` method now supports silent mode. (Contributed by
```
berkerpeksag: Can you move this to a new "py_compile" sub-section?

Example:

```rst
.. This needs to be placed after the plistlib section.

py_compile
----------

* The :meth:`py_compile.compile` method now supports silent mode.
  (Contributed by Joannah Nanjekye in :issue:`22640`.)
```

Doc/library/py_compile.rst
```diff
@@ -82,6 +89,10 @@ byte-code cache files in the directory containing the source code.
       overrides the value of the *invalidation_mode* argument, and determines
       its default value instead.
 
+   .. versionchanged:: 3.8
+      The *quiet* parameter was added to support silent mode. It returns full
```
berkerpeksag: We've already described the new parameter in the main documentation. We can just say "\*quiet\* parameter was added." here.

Lib/test/test_py_compile.py
berkerpeksag: Why did you mark this resolved? It still needs to be addressed.

Misc/NEWS.d/next/Library/2019-04-26-22-13-26.bpo-22640.p3rheW.rst
berkerpeksag: This comment is still needs to be addressed as well.

berkerpeksag (changes requested):

Lib/test/test_py_compile.py
nanjekyejoannah: Sorry, I had confused it with @vadmium 's comment .

Doc/whatsnew/3.8.rst
```diff
@@ -402,6 +402,13 @@ NSKeyedArchiver-encoded binary plists.
 (Contributed by Jon Janzen in :issue:`26707`.)
 
 
+py_compile
+----------
+
+The :meth:`py_compile.compile` method now supports silent mode.
```
ZackerySpytz: ``py_compile.compile`` is not a method.

ZackerySpytz:
@nanjekyejoannah I seems that you marked my comment as "resolved", but the "What's New" entry still falsely states that ``py_compile.compile`` is a method. Please fix that.

ZackerySpytz:
@nanjekyejoannah It seems that you marked my comment as "resolved", but the "What's New" entry still falsely states that ``py_compile.compile`` is a method. Please fix that.

nanjekyejoannah:
@berkerpeksag  PTAL .

nanjekyejoannah:

berkerpeksag (approved):

berkerpeksag:
Thanks!

vstinner:
Well done @nanjekyejoannah!

merwok:
Follow-up in #17134 to fix the NameErrors in `main`.
As this is already released, 3.7 should get a simple fix and 3.8 the missing part of the feature :slightly_smiling_face:

vstinner:
It seems like this change introduced a regression in main(): see https://bugs.python.org/issue40456

merwok:
It seems that you missed my previous message Victor! :slightly_smiling_face:

vstinner:
> It seems that you missed my previous message Victor! slightly_smiling_face

Oh, right. Well, the main() changes of this PR should be reverted in 3.8. #17134 is new feature which should not into 3.8.

PR #12979, opened 2019-04-27
bpo-36722: Style and grammar edits for ABI entry

pganssle:
@vstinner expedited the merge for this ABI entry to make sure it's in before the alpha4 release. This is a cleanup PR with my copy edits.

The one thing I wasn't sure about is that I changed the word "define" to "macro" in a few places, referring to preprocessor directives like `Py_TRACE_REFS`. Macro sounds more right to me than "define", though I did always think there was a qualitative difference between expansion macros and macros you define for `#ifdef` guards. Anyway, `man gcc` says "`[-Dmacro[=defn]...] [-Umacro]`, so I think "macro" is correct.

CC: @elprans

### DIFF
--- a/Doc/whatsnew/3.8.rst
+++ b/Doc/whatsnew/3.8.rst
@@ -101,20 +101,21 @@ subdirectories).
 Debug build uses the same ABI as release build
 -----------------------------------------------
 
-Python now uses the same ABI when built in release and in debug mode. On Unix,
-when Python is build in debug mode, it is now possible to load C extensions
-built in release mode and C extensions built using the stable ABI.
+Python now uses the same ABI whether it built in release or debug mode. On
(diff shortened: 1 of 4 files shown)


vstinner (commented):
Oh, I am using the "define" term whereas you are right: I should use the "macro" term.

Would you mind to check if Misc/SpecialBuilds.txt uses the right term?

And maybe also update the NEWS entries of the 3 issues?

pganssle:
@vstinner Updated. I went with "preprocessor directives" in the `SpecialBuilds.txt`, since it already said "preprocessor" and I think "preprocessor directive" is the broadest term for these. If you want I can change it to macro.

vstinner:
Thanks Paul!

PR #12985, opened 2019-04-27
Corrected pydocs brackets

borisbutenko:
Corrected brackets in the desc of "_randbelow_with_getrandbits" and "_randbelow_without_getrandbits" methods

### DIFF
--- a/Lib/random.py
+++ b/Lib/random.py
@@ -248,7 +248,7 @@ class Random(_random.Random):
         return self.randrange(a, b+1)
 
     def _randbelow_with_getrandbits(self, n):
-        "Return a random int in the range [0,n).  Raises ValueError if n==0."
+        "Return a random int in the range [0, n].  Raises ValueError if n==0."
 
         getrandbits = self.getrandbits
         k = n.bit_length()  # don't use (n-1) here because n can be 1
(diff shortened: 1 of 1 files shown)


the-knights-who-say-ni:

Lib/random.py
```diff
@@ -248,7 +248,7 @@ def randint(self, a, b):
         return self.randrange(a, b+1)
 
     def _randbelow_with_getrandbits(self, n):
-        "Return a random int in the range [0,n).  Raises ValueError if n==0."
+        "Return a random int in the range [0, n].  Raises ValueError if n==0."
```
tirkarthi, line 251: I don't think this is a typo. The docstring is using the mathematical notation to indicate the fact that resulting random integer can be of the boundary 0 <= result < n. This change would cause the docstring to mean 0 <= result <= n and the function never seems to return a value where result == n.


```
>>> import random
>>> list(filter(lambda x: x == 2, [random.Random()._randbelow_with_getrandbits(2) for _ in range(10000)]))
[]
```

rhettinger:
The [half-open interval notation](https://en.wikipedia.org/wiki/Interval_(mathematics)#Notations_for_intervals) [0, n] means  0 <= x < n.

PR #12987, opened 2019-04-27
bpo-2091: Fix typo in exception message

berkerpeksag:
https://bugs.python.org/issue2091


tirkarthi (approved):
LGTM. I am not sure of backporting this to 3.7. There was a similar typo fixed in Lib/random.py error message and was not backported to 3.7 as noted in https://bugs.python.org/issue35782#msg334364.

Thanks.

berkerpeksag:
We don't backport changes to existing error messages only when we significantly changed their formats. 2433a2ab705e93f9a44f01c260d351b205a73e9d and this one are completely OK to backport.

PR #12989, opened 2019-04-27
 bpo-21536: On Android, C extensions are linked to libpython

xdegaye:


auvipy (approved):

vstinner:
LGTM. Thanks @xdegaye for this fix!

I tested the PR on Linux: C extensions are still not linked to libpython, as expected.

Xavier tested the cross-compilation to Android.

embray:
Note: This is a follow-up to #12946.

configure.ac
```diff
@@ -4647,6 +4647,14 @@ AC_MSG_CHECKING(LDVERSION)
 LDVERSION='$(VERSION)$(ABIFLAGS)'
 AC_MSG_RESULT($LDVERSION)
 
+# On Android the shared libraries must be linked with libpython.
+AC_SUBST(LIBPYTHON)
+if test -z "$ANDROID_API_LEVEL"; then
```
embray, line 4652: Out of curiosity, being that I'm unfamiliar with build issues on Android why the specific test for `$ANDROID_API_LEVEL` and not something more generic like:

```
case $host in
    *-linux-android*) ....
```
?

Reason being, I need to add additional cases where libpython should be linked explicitly (e.g. Cygwin, MinGW, etc.) and it would be easier to build off the host_os for this than something very Android-specific.

xdegaye: Using a test based on ``$ANDROID_API_LEVEL`` or on ``$host`` as you suggest is equivalent.

PR #12990, opened 2019-04-28
bpo-26978: Implement pathlib.Path.link_to (Using os.link)

nanjekyejoannah:
I have implemented `pathlib.Path.link_to`.

https://bugs.python.org/issue26978

### DIFF
--- a/Doc/library/pathlib.rst
+++ b/Doc/library/pathlib.rst
@@ -1054,6 +1054,12 @@ call fails (for example because the path doesn't exist).
    use :func:`Path.rmdir` instead.
 
 
+.. method:: Path.link(target)
+
+   Create a hard link pointing to a path named *target*. If the path is a directory,
+   use :func:`Path.mkdir` instead.
+
+
 .. method:: Path.write_bytes(data)
 
    Open the file pointed to in bytes mode, write *data* to it, and close the
--- a/Doc/library/pathlib.rst
+++ b/Doc/library/pathlib.rst
@@ -1054,10 +1054,11 @@ call fails (for example because the path doesn't exist).
    use :func:`Path.rmdir` instead.
 
 
-.. method:: Path.link(target)
+.. method:: Path.link_to(target)
 
-   Create a hard link pointing to a path named *target*. If the path is a directory,
-   use :func:`Path.mkdir` instead.
+   Create a hard link pointing to a path named *target*.
+
+   .. versionchanged:: 3.8
 
 
 .. method:: Path.write_bytes(data)
--- a/Doc/whatsnew/3.8.rst
+++ b/Doc/whatsnew/3.8.rst
@@ -334,7 +334,8 @@ pathlib
 contain characters unrepresentable at the OS level.
 (Contributed by Serhiy Storchaka in :issue:`33721`.)
 
-Added :meth:`pathlib.Path.link()`.
+Added :meth:`pathlib.Path.link_to()` which creates a hard link pointing
+to a path.
 (Contributed by Joannah Nanjekye in :issue:`26978`)
 
 
--- a/Doc/whatsnew/3.8.rst
+++ b/Doc/whatsnew/3.8.rst
@@ -333,8 +333,11 @@ pathlib
 :exc:`ValueError` or its subclass :exc:`UnicodeEncodeError` for paths that
 contain characters unrepresentable at the OS level.
 (Contributed by Serhiy Storchaka in :issue:`33721`.)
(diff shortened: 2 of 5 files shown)


auvipy (approved):

nanjekyejoannah:
cc @pitrou

Doc/library/pathlib.rst
```diff
@@ -1054,6 +1054,12 @@ call fails (for example because the path doesn't exist).
    use :func:`Path.rmdir` instead.
 
 
+.. method:: Path.link(target)
```
pitrou: Since we already have `symlink_to`, this method should probably be called `link_to`.

Doc/library/pathlib.rst
```diff
@@ -1054,6 +1054,12 @@ call fails (for example because the path doesn't exist).
...
+
+   Create a hard link pointing to a path named *target*. If the path is a directory,
+   use :func:`Path.mkdir` instead.
+
+
```
pitrou, line 1063: You should add a `versionadded` marker for this method.

Lib/pathlib.py
```diff
@@ -1303,6 +1305,15 @@ def lstat(self):
             self._raise_closed()
         return self._accessor.lstat(self)
 
+    def link(self, target):
+        """
+        Create a hard link pointing to a path named target.
+        If the path is a directory, use mkdir() instead.
```
pitrou: I'm not sure what this mean? `mkdir` is not the same thing as creating a link...

Lib/test/test_pathlib.py
```diff
@@ -1643,6 +1643,24 @@ def test_rmdir(self):
...
+        P = self.cls(BASE)
+        p = P / 'fileA'
+        size = p.stat().st_size
+        # linking to another path.
+        q = P / 'dirA' / 'fileAA'
+        try:
+            p.link(q)
+        except PermissionError as e:
+            self.skipTest('os.link(): %s' % e)
+        self.assertEqual(q.stat().st_size, size)
```
pitrou, line 1656: To test that the target is really a link, you should probably use [os.path.samefile](https://docs.python.org/3/library/os.path.html#os.path.samefile).

pitrou (changes requested):
Thanks for submitting this PR. You'll find some comments below.

Lib/pathlib.py
nanjekyejoannah: I just removed that statement.

nanjekyejoannah:

Doc/whatsnew/3.8.rst
```diff
@@ -334,6 +334,9 @@ pathlib
 contain characters unrepresentable at the OS level.
 (Contributed by Serhiy Storchaka in :issue:`33721`.)
 
+Added :meth:`pathlib.Path.link()`.
```
tirkarthi: Should it be `pathlib.Path.link_to` ? 

Minor nit: Could add a short sentence that it creates a hardlink here and in NEWS entry. I will let it to reviewers on the wording though.

pitrou (approved):

pitrou:
Thank you @nanjekyejoannah. I will wait for CI to pass and then merge if green.

Doc/whatsnew/3.8.rst
```diff
@@ -334,6 +334,10 @@ pathlib
 contain characters unrepresentable at the OS level.
 (Contributed by Serhiy Storchaka in :issue:`33721`.)
 
+Added :meth:`pathlib.Path.link_to()` which creates a hard link pointing
+to a path.
+(Contributed by Joannah Nanjekye in :issue:`26978`)
```
ZackerySpytz, line 339:
```suggestion
(Contributed by Joannah Nanjekye in :issue:`26978`.)
```

Misc/NEWS.d/next/Library/2019-04-28-01-52-39.bpo-26978.Lpm-SI.rst
```diff
@@ -0,0 +1,2 @@
+`pathlib.path.link_to()` is now implemented. It creates a hard link pointing
```
ZackerySpytz, line 1: It should be `pathlib.Path.link_to()`, not `pathlib.path.link_to()`.

Also, double backqoutes should be used here instead of single backquotes (as per https://devguide.python.org/documenting/#inline-markup). The following could also be used:

```rst
:meth:`pathlib.Path.link_to()`
```

PR #12991, opened 2019-04-28
bpo-24758: Improve the error msg for unittest.mock.Mock()'s unsafe mode

ZackerySpytz:
https://bugs.python.org/issue24758


Lib/unittest/mock.py
```diff
@@ -595,7 +595,7 @@ def __getattr__(self, name):
             raise AttributeError(name)
         if not self._mock_unsafe:
             if name.startswith(('assert', 'assret')):
-                raise AttributeError(name)
+                raise AttributeError("Attributes cannot start with 'assert'")
```
tirkarthi: `startswith` checks both "assert" and "assret" so hardcoding "assert" might lead to confusing error message when the attribute starts with assret instead. Perhaps change this to "Attributes cannot start with 'assert' or 'assret'" ?

```
./python.exe
Python 3.8.0a3+ (heads/master:55d035113d, Apr 28 2019, 01:37:48)
>>> from unittest import mock
>>> mock.Mock().assret_mock_call
  File "/Users/karthikeyansingaravelan/stuff/python/cpython/Lib/unittest/mock.py", line 598, in __getattr__
    raise AttributeError("Attributes cannot start with 'assert'")
AttributeError: Attributes cannot start with 'assert'
```

kushaldas (changes requested):
`"Attributes cannot start with 'assert' or 'assret'"` is a better error message in this. Also, please check for the right error message in `test_mock_unsafe` as we are checking the assertion there.

ZackerySpytz:
Thank you, @kushaldas.

tirkarthi (approved):
LGTM. The attribute can start with assert or assret in unsafe mode but adding it to error message looks intrusive to me and this seems like a good improvement.

Thank you for the PR.

kushaldas (approved):
This looks good now, thank you for the PR. Approved.

PR #12992, opened 2019-04-28
bpo-36743: __get__ is sometimes called without the owner argument

rhettinger:
https://bugs.python.org/issue36743


jdufresne:
Would it make sense to update the signature line to?

```
.. method:: object.__get__(self, instance, owner=None)
```

jdemeyer (changes requested):
As explained on bpo-36743, the text "Note that callers are allowed to omit the *owner* argument when it isn't needed." is not actually true in general, or trivially true depending on how you interpret "isn't needed".

PR #12993, opened 2019-04-28
bpo-36745: Fix a possible reference leak in PyObject_SetAttr()

ZackerySpytz:
https://bugs.python.org/issue36745


asvetlov:
Good catch!

asvetlov (approved):

PR #12995, opened 2019-04-28
bpo-32424: Deprecate xml.etree.ElementTree.Element.copy() in favor of copy.copy()

GPHemsley:
* Deprecate `Element.copy()` in the Python implementation of `xml.etree.ElementTree`.
* Test that its use throws `DeprecationWarning` when present.
* Test that it still copies the element as expected (based on changes made in #12891).

https://bugs.python.org/issue32424

### DIFF
--- a/Lib/test/test_xml_etree.py
+++ b/Lib/test/test_xml_etree.py
@@ -1958,6 +1958,35 @@ class BasicElementTest(ElementTestCase, unittest.TestCase):
         self.assertIsNot(element_foo.attrib, attrib)
         self.assertNotEqual(element_foo.attrib, attrib)
 
+    def test_copy(self):
+        # Only run this test if Element.copy() is defined.
+        if "copy" not in dir(ET.Element):
+            raise unittest.SkipTest("Element.copy() not present")
+
+        element_foo = ET.Element("foo", { "zix": "wyp" })
+        element_foo.append(ET.Element("bar", { "baz": "qix" }))
+
+        with self.assertWarns(DeprecationWarning):
+            element_foo2 = element_foo.copy()
+
+        # elements are not the same
+        self.assertIsNot(element_foo2, element_foo)
+
+        # string attributes are equal
+        self.assertEqual(element_foo2.tag, element_foo.tag)
+        self.assertEqual(element_foo2.text, element_foo.text)
+        self.assertEqual(element_foo2.tail, element_foo.tail)
+
+        # number of children is the same
+        self.assertEqual(len(element_foo2), len(element_foo))
+
+        # children are the same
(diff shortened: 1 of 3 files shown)


GPHemsley:
@scoder @serhiy-storchaka Does this match expectations?

Lib/test/test_xml_etree.py
```diff
@@ -1958,6 +1958,35 @@ def test___init__(self):
...
+
+        # number of children is the same
+        self.assertEqual(len(element_foo2), len(element_foo))
+
+        # children are the same
+        for (child1, child2) in itertools.zip_longest(element_foo, element_foo2):
+            self.assertIs(child1, child2)
+
+        # attrib is a copy
+        self.assertEqual(element_foo2.attrib, element_foo.attrib)
```
scoder, line 1988: `attrib is a copy` is not quite what this is testing, but it should.

Lib/xml/etree/ElementTree.py
```diff
@@ -194,6 +194,13 @@ def copy(self):
         original tree.
 
         """
+        warnings.warn(
+            "elem.copy() is deprecated. Use copy.copy(elem) instead.",
+            DeprecationWarning
```
scoder, line 199: Not sure if we should issue a `DeprecationWarning` straight away (or a `PendingDeprecationWarning` first), although I personally wouldn't mind, given that the C implementation has shadowed this method for years (in CPython, that is).
@serhiy-storchaka, what do you think?

Lib/test/test_xml_etree.py
GPHemsley: The actual copy test will fail at the moment. (Note that it is also missing from the `__copy__` test that has already been merged.) It'll be included in the patch for bpo-36685, when it will pass.

Lib/xml/etree/ElementTree.py
GPHemsley: FWIW, `copy()` does not appear in the documentation, either.

GPHemsley: Also, I can't find any record of `copy` ever having been available in the C implementation; it looks like it was `__copy__` from the beginning.

The first appearance of `copy` in the Python implementation was in f15351d938e76c4b2a42a638825d67f108685140 (first released in Python 3.1 and 2.7.4), which deemed it "experimental". 84fae785ce4b0a15172fb4590d684431def9dea9 (first released in Python 3.4) removed that designation when switching from comments to docstrings.

GPHemsley:
@serhiy-storchaka Any thoughts on this? I think your opinion is the only thing outstanding here.

PR #12996, opened 2019-04-28
Don't use the LHS/RHS acronym in Simple statements

andresdelfino:
Prefer the full wording instead, as it is more meaningful for someone not familiar with the terms.

Also, LFS/RHS is not used anywhere else in the documentation, while left/right-hand side mentions are common.


alessandrocucci (approved):
+1

asvetlov (approved):

PR #12999, opened 2019-04-28
bpo-36746: Create test for fcntl.lockf()

nanjekyejoannah:
I have created a test for` fcntl.lockf()`.

https://bugs.python.org/issue36746

### DIFF
--- a/Lib/test/test_fcntl.py
+++ b/Lib/test/test_fcntl.py
@@ -12,8 +12,6 @@ from test.support import (verbose, TESTFN, unlink, run_unittest, import_module,
 fcntl = import_module('fcntl')
 
 
-# TODO - Write tests for flock() and lockf().
-
 def get_lockdata():
     try:
         os.O_LARGEFILE
@@ -137,6 +135,39 @@ class TestFcntl(unittest.TestCase):
 
         self.assertRaises(ValueError, fcntl.flock, -1, fcntl.LOCK_SH)
         self.assertRaises(TypeError, fcntl.flock, 'spam', fcntl.LOCK_SH)
+
+    def test_lockf(self):
+
+        self.f = open(TESTFN, 'wb+')
+
+        self.assertRaises(TypeError, fcntl.lockf, self.f, "foo")
+        self.assertRaises(TypeError, fcntl.lockf, self.f, fcntl.LOCK_UN, "foo")
+        self.assertRaises(ValueError, fcntl.lockf, self.f, -256)
+        self.assertRaises(ValueError, fcntl.lockf, self.f, 256)
+
+        fcntl.lockf(self.f, fcntl.LOCK_EX | fcntl.LOCK_NB)
+
+        pid = os.fork()
+        if pid == 0:
+            rval = 2
+            try:
+                fcntl.lockf(open(self.f.name, self.f.mode), fcntl.LOCK_EX | fcntl.LOCK_NB)
+            except IOError as e:
+                if e.errno not in (errno.EACCES, errno.EAGAIN):
+                    raise
+                rval = 0
+            else:
+                rval = 1
+            finally:
+                os._exit(rval)
+
+        assert pid > 0
+        (pid, status) = os.waitpid(pid, 0)
+        self.assertEqual(os.WIFEXITED(status), True)
+
+        fcntl.lockf(self.f, fcntl.LOCK_UN)
+
+        self.f.close()
 
     @cpython_only
     def test_flock_overflow(self):
--- a/Lib/test/test_fcntl.py
+++ b/Lib/test/test_fcntl.py
@@ -5,6 +5,8 @@ import os
 import struct
 import sys
 import unittest
+import textwrap
+import subprocess
 from test.support import (verbose, TESTFN, unlink, run_unittest, import_module,
                           cpython_only)
 
@@ -140,34 +142,36 @@ class TestFcntl(unittest.TestCase):
 
         self.f = open(TESTFN, 'wb+')
 
+        fcntl.lockf(self.f, fcntl.LOCK_EX | fcntl.LOCK_NB)
+
+        code = textwrap.dedent('''
+            try:
+                fcntl.lockf(open(self.f.name, self.f.mode), fcntl.LOCK_EX | fcntl.LOCK_NB)
+            except OSError as e:
+                if e.errno not in (errno.EACCES, errno.EAGAIN):
+                    raise
+                os.EX_OSERR
+            else:
+                os.EX_OK
+            finally:
+                os._exit()
+        ''')
+
+        args = (sys.executable, '-c', code)
+        proc = subprocess.run(args,
+                              stdout=subprocess.PIPE,
+                              stderr=subprocess.STDOUT)
+
+        fcntl.lockf(self.f, fcntl.LOCK_UN)
+
+    def test_lockf_errors(self):
+
+        self.f = open(TESTFN, 'wb+')
+
         self.assertRaises(TypeError, fcntl.lockf, self.f, "foo")
         self.assertRaises(TypeError, fcntl.lockf, self.f, fcntl.LOCK_UN, "foo")
         self.assertRaises(ValueError, fcntl.lockf, self.f, -256)
         self.assertRaises(ValueError, fcntl.lockf, self.f, 256)
-
-        fcntl.lockf(self.f, fcntl.LOCK_EX | fcntl.LOCK_NB)
-
-        pid = os.fork()
-        if pid == 0:
-            rval = 2
-            try:
-                fcntl.lockf(open(self.f.name, self.f.mode), fcntl.LOCK_EX | fcntl.LOCK_NB)
-            except IOError as e:
-                if e.errno not in (errno.EACCES, errno.EAGAIN):
-                    raise
-                rval = 0
-            else:
-                rval = 1
-            finally:
-                os._exit(rval)
-
-        assert pid > 0
-        (pid, status) = os.waitpid(pid, 0)
-        self.assertEqual(os.WIFEXITED(status), True)
-
-        fcntl.lockf(self.f, fcntl.LOCK_UN)
-
-        self.f.close()
 
     @cpython_only
     def test_flock_overflow(self):
@@ -176,8 +180,5 @@ class TestFcntl(unittest.TestCase):
                           fcntl.LOCK_SH)
 
 
-def test_main():
-    run_unittest(TestFcntl)
-
 if __name__ == '__main__':
-    test_main()
+    unittest.main()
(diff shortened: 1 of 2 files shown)


Misc/NEWS.d/next/Tests/2019-04-28-18-36-54.bpo-36746.aqYNtS.rst
```diff
@@ -0,0 +1 @@
+A test for `fcntl.lockf()` is now implemented.
```
berkerpeksag: NEWS entry isn't need for trivial PRs. You can submit your work without adding a NEWS entry. I'm working on a PR that sets the "skip news" label automatically for documentation and test only PRs.

nanjekyejoannah: ohh great...At this point I dont know how to draw a line between trivial and non trivial but my judgement will get better with time.

berkerpeksag: I usually ask myself questions like "would this change be important to users? can it break user code?" when I'm not sure about adding a NEWS entry. New features and bug fixes are always need a NEWS entry.

In PR #12999's case, while increasing test coverage of fcntl module is nice, it doesn't have any effect on end users, so we can merge it without a NEWS entry.

berkerpeksag: Also, don't worry about NEWS entries too much, we can always add or delete them later :)

nanjekyejoannah: Cool. Also I wanted to change the test for posix.lockf() to look as detailed as the one in this PR. Does this implementation look better ?

nanjekyejoannah:

Lib/test/test_fcntl.py
```diff
@@ -138,6 +136,39 @@ def test_flock(self):
         self.assertRaises(ValueError, fcntl.flock, -1, fcntl.LOCK_SH)
         self.assertRaises(TypeError, fcntl.flock, 'spam', fcntl.LOCK_SH)
 
+    def test_lockf(self):
+
+        self.f = open(TESTFN, 'wb+')
+
+        self.assertRaises(TypeError, fcntl.lockf, self.f, "foo")
```
berkerpeksag: It would be nice to move these into a separate test.

Lib/test/test_fcntl.py
```diff
@@ -138,6 +136,39 @@ def test_flock(self):
...
+        self.assertRaises(TypeError, fcntl.lockf, self.f, fcntl.LOCK_UN, "foo")
+        self.assertRaises(ValueError, fcntl.lockf, self.f, -256)
+        self.assertRaises(ValueError, fcntl.lockf, self.f, 256)
+
+        fcntl.lockf(self.f, fcntl.LOCK_EX | fcntl.LOCK_NB)
+
+        pid = os.fork()
+        if pid == 0:
+            rval = 2
```
berkerpeksag: It would be nice to explain these return values. For example, if I understand the test correctly, `rval` should never return `2`, so it would be nice to document when the value of `rval` will be `2`.

Lib/test/test_fcntl.py
```diff
@@ -138,6 +136,39 @@ def test_flock(self):
...
+            try:
+                fcntl.lockf(open(self.f.name, self.f.mode), fcntl.LOCK_EX | fcntl.LOCK_NB)
+            except IOError as e:
```
berkerpeksag: `IOError` is deprecated and an alias of `OSError`. It would be better if we just use `OSError` directly.

Lib/test/test_fcntl.py
```diff
@@ -138,6 +136,39 @@ def test_flock(self):
...
+                if e.errno not in (errno.EACCES, errno.EAGAIN):
+                    raise
+                rval = 0
+            else:
+                rval = 1
+            finally:
+                os._exit(rval)
+
+        assert pid > 0
```
berkerpeksag: We can replace this assert with an else clause.

Lib/test/test_fcntl.py
```diff
@@ -138,6 +136,39 @@ def test_flock(self):
...
+        if pid == 0:
+            rval = 2
+            try:
+                fcntl.lockf(open(self.f.name, self.f.mode), fcntl.LOCK_EX | fcntl.LOCK_NB)
+            except IOError as e:
+                if e.errno not in (errno.EACCES, errno.EAGAIN):
+                    raise
+                rval = 0
+            else:
+                rval = 1
```
berkerpeksag: What do you think about returning `os.EX_OK`  here? It seems like we can use predefined values at https://docs.python.org/3/library/os.html#os._exit

Lib/test/test_fcntl.py
```diff
@@ -138,6 +136,39 @@ def test_flock(self):
...
+            finally:
+                os._exit(rval)
+
+        assert pid > 0
+        (pid, status) = os.waitpid(pid, 0)
+        self.assertEqual(os.WIFEXITED(status), True)
```
berkerpeksag: Nit: assetEqual -> assertTrue

Lib/test/test_fcntl.py
```diff
@@ -138,6 +136,39 @@ def test_flock(self):
...
+
+        fcntl.lockf(self.f, fcntl.LOCK_UN)
+
+        self.f.close()
```
berkerpeksag: `self.f` will be closed by the `tearDown` method, so we can drop this line.

Lib/test/test_fcntl.py
```diff
@@ -138,6 +136,39 @@ def test_flock(self):
...
+        self.assertRaises(ValueError, fcntl.lockf, self.f, -256)
+        self.assertRaises(ValueError, fcntl.lockf, self.f, 256)
+
+        fcntl.lockf(self.f, fcntl.LOCK_EX | fcntl.LOCK_NB)
+
+        pid = os.fork()
+        if pid == 0:
+            rval = 2
+            try:
+                fcntl.lockf(open(self.f.name, self.f.mode), fcntl.LOCK_EX | fcntl.LOCK_NB)
```
tirkarthi, line 149: This line causes a `ResourceWarning` since the open file handle is not closed. Perhaps assign it to a variable and make sure it's closed in the test.

```
➜  cpython git:(master) ✗ ./python.exe -m unittest -v test.test_fcntl.TestFcntl.test_lockf
struct.pack:  b'\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x03\x00\x00\x00'
test_lockf (test.test_fcntl.TestFcntl) ... /Users/karthikeyansingaravelan/stuff/python/cpython/Lib/test/test_fcntl.py:154: ResourceWarning: unclosed file <_io.BufferedRandom name='@test_27262_tmp'>
  fcntl.lockf(open(self.f.name, self.f.mode), fcntl.LOCK_EX | fcntl.LOCK_NB)
ok

Ran 1 test in 0.008s

OK
```

Lib/test/test_fcntl.py
tirkarthi: I feel this is redundant since `tearDown` takes care of closing this. I am not sure some of the other tests close it too. Is it intentional?

Lib/test/test_fcntl.py
tirkarthi: `self.assertGreater(pid, 0)` could be useful in error messages where pid value is shown for debugging assertion failure.

berkerpeksag (changes requested):
I think your approach looks pretty good.

Have you considered using `subprocess.run()` instead of `os.fork()`? I think the former might be a better alternative especially if you are planning to increase test coverage of fcntl since it would make code reusing easier.

Also, while you are editing this file, could you please replace the following snippet

```py
def test_main():
    run_unittest(TestFcntl)

if __name__ == '__main__':
    test_main()
```

with

```py
if __name__ == '__main__':
    unittest.main()
```

nanjekyejoannah:
@berkerpeksag I have made some changes to use subprocess.run(). I have not used subprocess.run() before so your thoughts on this change are highly appreciated.

nanjekyejoannah:

Lib/test/test_fcntl.py
```diff
@@ -138,15 +138,47 @@ def test_flock(self):
         self.assertRaises(ValueError, fcntl.flock, -1, fcntl.LOCK_SH)
         self.assertRaises(TypeError, fcntl.flock, 'spam', fcntl.LOCK_SH)
 
+    def test_lockf(self):
+
+        self.f = open(TESTFN, 'wb+')
+
+        fcntl.lockf(self.f, fcntl.LOCK_EX | fcntl.LOCK_NB)
+
+        code = textwrap.dedent('''
```
vstinner, line 147: fcntl module should be imported, no? Moreover, you should pass self.f.name into the string. For example, use an f-string and replace self.f.name with {TESTFN!r}. Replace self.f.mode with 'wb+'.

Lib/test/test_fcntl.py
```diff
@@ -138,15 +138,47 @@ def test_flock(self):
...
+            try:
+                fcntl.lockf(open(self.f.name, self.f.mode), fcntl.LOCK_EX | fcntl.LOCK_NB)
+            except OSError as e:
+                if e.errno not in (errno.EACCES, errno.EAGAIN):
+                    raise
+                os.EX_OSERR
+            else:
+                os.EX_OK
+            finally:
+                os._exit()
```
vstinner, line 157: os._exit() requires an argument.

Lib/test/test_fcntl.py
```diff
@@ -138,15 +138,47 @@ def test_flock(self):
...
+
+        fcntl.lockf(self.f, fcntl.LOCK_EX | fcntl.LOCK_NB)
+
+        code = textwrap.dedent('''
+            try:
+                fcntl.lockf(open(self.f.name, self.f.mode), fcntl.LOCK_EX | fcntl.LOCK_NB)
+            except OSError as e:
+                if e.errno not in (errno.EACCES, errno.EAGAIN):
+                    raise
+                os.EX_OSERR
```
vstinner, line 153: what's the purpose of this statement?

Lib/test/test_fcntl.py
```diff
@@ -138,15 +138,47 @@ def test_flock(self):
...
+
+        args = (sys.executable, '-c', code)
+        proc = subprocess.run(args,
+                              stdout=subprocess.PIPE,
+                              stderr=subprocess.STDOUT)
+
+        fcntl.lockf(self.f, fcntl.LOCK_UN)
+
+    def test_lockf_errors(self):
+
```
vstinner, line 168: remove this empty line

vstinner (changes requested):

vstinner:
PR #17010 added tests for fcntl.lockf(). I close this PR. Thanks @nanjekyejoannah anyway!

PR #13001, opened 2019-04-29
bpo-36546: Fix typo quaatile to quantile

tirkarthi:
https://bugs.python.org/issue36546

### DIFF
--- a/Lib/test/test_statistics.py
+++ b/Lib/test/test_statistics.py
@@ -2178,7 +2178,7 @@ class TestQuantiles(unittest.TestCase):
     def test_specific_cases_inclusive(self):
         # Match results computed by hand and cross-checked
         # against the PERCENTILE.INC function in MS Excel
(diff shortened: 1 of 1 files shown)


Lib/test/test_statistics.py
```diff
@@ -2178,7 +2178,7 @@ def f(x):
     def test_specific_cases_inclusive(self):
         # Match results computed by hand and cross-checked
         # against the PERCENTILE.INC function in MS Excel
-        # and against the quaatile() function in SciPy.
+        # and against the quantile() function in SciPy.
```
rougeth, line 2181: Wouldn't be better to reference NumPy instead of SciPy?
ref: https://docs.scipy.org/doc/numpy/reference/generated/numpy.quantile.html

tirkarthi: I will leave it to @rhettinger on changing scipy to numpy

rougeth (approved):

PR #13002, opened 2019-04-29
bpo-36748: optimize TextIOWrapper.write() for ASCII string

methane:
https://bugs.python.org/issue36748


ZackerySpytz:
I tested this PR locally, and it seems that there are reference leaks (according to `./python -m test -R 3:3 test_io`).
```
test_io leaked [4, 4, 4] references, sum=12
test_io failed in 4 min 12 sec

== Tests result: FAILURE ==

1 test failed:
    test_io

Total duration: 4 min 12 sec
Tests result: FAILURE
```

PR #13004, opened 2019-04-29
bpo-36749: AIX dynamic linker requires to link to libpython

vstinner:
On AIX, the dynamic linker requires that C extensions are linked to
libpython.

https://bugs.python.org/issue36749


vstinner:
PPC64 AIX 3.x buildbot worker is back to green. It seems like my PR is useless, I close it.

PR #13007, opened 2019-04-29
bpo-35952: Fix pythoninfo when the compiler is missing

xdegaye:
https://bugs.python.org/issue35952

### DIFF
--- a/Lib/test/pythoninfo.py
+++ b/Lib/test/pythoninfo.py
@@ -571,13 +571,23 @@ def collect_cc(info_add):
     except ImportError:
         args = CC.split()
     args.append('--version')
-    proc = subprocess.Popen(args,
-                            stdout=subprocess.PIPE,
-                            stderr=subprocess.STDOUT,
-                            universal_newlines=True)
-    stdout = proc.communicate()[0]
-    if proc.returncode:
-        # CC --version failed: ignore error
+    proc = None
+    try:
+        proc = subprocess.Popen(args,
+                                stdout=subprocess.PIPE,
+                                stderr=subprocess.STDOUT,
+                                universal_newlines=True)
+    except (FileNotFoundError, PermissionError):
+        # Cannot run the compiler, for example when Python has been
+        # cross-compiled and installed on the target platform where the
+        # compiler is missing.
+        pass
+    else:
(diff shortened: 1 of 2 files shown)


Lib/test/pythoninfo.py
```diff
@@ -571,13 +571,23 @@ def collect_cc(info_add):
...
+        # Cannot run the compiler, for example when Python has been
+        # cross-compiled and installed on the target platform where the
+        # compiler is missing.
+        pass
+    else:
+        stdout = proc.communicate()[0]
+
+    # Failed to run the compiler or 'CC --version' failed.
+    if proc is None or proc.returncode:
+        info_add('CC.name', normalize_text(CC))
```
vstinner: I'm not sure of the purpose of this change: CC is already logged as "sysconfig[CC]: gcc -pthread", no?

Lib/test/pythoninfo.py
```diff
@@ -571,13 +571,23 @@ def collect_cc(info_add):
...
-    stdout = proc.communicate()[0]
-    if proc.returncode:
-        # CC --version failed: ignore error
+    proc = None
+    try:
+        proc = subprocess.Popen(args,
+                                stdout=subprocess.PIPE,
+                                stderr=subprocess.STDOUT,
+                                universal_newlines=True)
+    except (FileNotFoundError, PermissionError):
```
vstinner: This part is perfectly fine ;-)

vstinner: Hum, honestly, ignoring any OSError would be just fine. There is no need to be specific here.

Lib/test/pythoninfo.py
xdegaye: I missed that point and will change that.

Lib/test/pythoninfo.py
xdegaye: Ok.

Lib/test/pythoninfo.py
vstinner: > I missed that point and will change that.

No problem.

vstinner:
Thanks for the fix Xavier. I backported latest pythoninfo changes (not only this one) to 2.7 and 3.7 branches.

PR #13011, opened 2019-04-29
bpo-34602: Avoid failures setting macOS stack resource limit

ned-deily:
Under some conditions the earlier fix for [bpo-18075](https://bugs.python.org/issue18075), "Infinite recursion
tests triggering a segfault on Mac OS X", now causes failures on macOS
when attempting to change stack limit with resource.setrlimit
resource.RLIMIT_STACK, like regrtest does when running the test suite.
The reverted change had specified a non-default stack size when linking
the python executable on macOS.  As of macOS 10.14.4, the previous
code causes a hard failure when running tests, although similar
failures had been seen under some conditions under some earlier
systems.  For now, revert the original change and resume using
the default stack size when linking the interpreter.

https://bugs.python.org/issue34602


ned-deily:
Note that this change and its backports have been reverted and a different approach taken.  See the bug tracker issue for more details.

PR #13016, opened 2019-04-29
bpo-36751: Deprecate getfullargspec and report positional-only args as regular args

pablogsal:
https://bugs.python.org/issue36751

### DIFF
--- a/Doc/library/inspect.rst
+++ b/Doc/library/inspect.rst
@@ -948,6 +948,11 @@ Classes and functions
    APIs. This function is retained primarily for use in code that needs to
    maintain compatibility with the Python 2 ``inspect`` module API.
 
+   .. deprecated:: 3.8
+      Use :func:`signature` and
+      :ref:`Signature Object <inspect-signature-object>`, which provide a
+      better introspecting API for callables.
+
    .. versionchanged:: 3.4
       This function is now based on :func:`signature`, but still ignores
       ``__wrapped__`` attributes and includes the already bound first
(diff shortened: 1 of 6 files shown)


zooba (approved):
Looks good to me. Thanks, Pablo! 👍

tirkarthi (changes requested):
This generates `DeprecationWarning` in two tests in `Lib/unittest/test/testmock/testhelpers.py`. Please wrap it with `self.assertWarns(DeprecationWarning)` or better use `inspect.signature` in the two assert calls to remove the `DeprecationWarning`.

```
➜  cpython git:(pr_13016) ./python.exe Lib/unittest/test/testmock/testhelpers.py
...................................................../Users/karthikeyansingaravelan/stuff/python/cpython/Lib/inspect.py:1114: DeprecationWarning: Use inspect.signature() instead of inspect.getfullargspec()
  warnings.warn("Use inspect.signature() instead of inspect.getfullargspec()",
.............
Ran 66 tests in 0.832s

OK
```

pablogsal:
Done in 74639b3. Thanks @tirkarthi, good catch!

tirkarthi (approved):
LGTM. Thanks @pablogsal .

ncoghlan:
Please revert this change - this API was deliberately undeprecated, as per https://bugs.python.org/issue27172 and PEP 570 isn't sufficient justification to deprecate it again.

ncoghlan:
Slight correction: the API fixes to restore compatibility with 3.7 are a good thing, the problem is the reintroduction of the programmatic deprecation warnings, even though the reasons for leaving them out still hold.

pablogsal:
@ncoghlan I have opened #13245 to handle the undeprecation. Sorry for the inconvenience.

PR #13017, opened 2019-04-29
bpo-36742: Fixes handling of pre-normalization characters in urlsplit()

zooba:


eamanu (approved):
LGTM. 

I test it in a Debian 4.9.168-1 x86_64 GNU/Linux and the bugs is fixed. 

Tests run ok.

PR #13025, opened 2019-04-30
[2.7] bpo-36742: Fixes handling of pre-normalization characters in urlsplit() (GH-13017)

zooba:

### DIFF
--- a/Lib/test/test_urlparse.py
+++ b/Lib/test/test_urlparse.py
@@ -641,6 +641,12 @@ class UrlParseTestCase(unittest.TestCase):
         self.assertIn(u'\u2100', denorm_chars)
         self.assertIn(u'\uFF03', denorm_chars)
 
+        # bpo-36742: Verify port separators are ignored when they
+        # existed prior to decomposition
+        urllib.parse.urlsplit('http://\u30d5\u309a:80')
+        with self.assertRaises(ValueError):
+            urllib.parse.urlsplit('http://\u30d5\u309a\ufe1380')
+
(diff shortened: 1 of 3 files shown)


Lib/test/test_urlparse.py
```diff
@@ -641,6 +641,12 @@ def test_urlsplit_normalization(self):
         self.assertIn(u'\u2100', denorm_chars)
         self.assertIn(u'\uFF03', denorm_chars)
 
+        # bpo-36742: Verify port separators are ignored when they
+        # existed prior to decomposition
+        urllib.parse.urlsplit('http://\u30d5\u309a:80')
```
tirkarthi: I guess the Python 3 patch applied here. `urlunparse.urlsplit` could be the correct function for this test.

zooba: Yep, you're right. Thanks!

tirkarthi: You're welcome. There is some test error on CI :

```
test test_urlparse failed -- Traceback (most recent call last):
  File "/home/travis/build/python/cpython/Lib/test/test_urlparse.py", line 648, in test_urlsplit_normalization
    urlparse.urlsplit('http://\u30d5\u309a\ufe1380')
AssertionError: ValueError not raised
```

tirkarthi: Ah I guess u"" prefix is missing since it's Python 2 ? Adding it makes the tests to pass locally for the PR.

tirkarthi (approved):
LGTM. Thanks.

PR #13032, opened 2019-05-01
bpo-30458: Disable https related urllib tests on a build without ssl

tirkarthi:
These tests require an SSL enabled build. Skip these tests when python is built without SSL to fix test failures.


auvipy (approved):

gpshead (approved):
thanks!

PR #13033, opened 2019-05-01
remove jython support from unittest.mock

cjw296:
Hmm, looks like we can no longer push to master or review our own PRs. I guess that's a good thing ;-)

If someone could ack this, I'd be very grateful.

### DIFF
--- a/Lib/unittest/mock.py
+++ b/Lib/unittest/mock.py
@@ -36,13 +36,6 @@ from functools import wraps, partial
 
 _builtins = {name for name in dir(builtins) if not name.startswith('_')}
 
-BaseExceptions = (BaseException,)
-if 'java' in sys.platform:
-    # jython
-    import java
-    BaseExceptions = (BaseException, java.lang.Throwable)
-
-
 FILTER_DIR = True
 
 # Workaround for issue #12370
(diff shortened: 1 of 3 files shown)


cjw296:
See https://github.com/testing-cabal/mock/issues/453 for discussion.

cjw296:
Oh, nevermind, looks like I can merge once tests have passed.

vstinner:
I'm not sure that removing features just to get 100% coverage is a good practice. It's not like the removed 3 lines caused a huge maintenance burden.

I'm not sure that Jython is fully dead. We well see, @cjw296 asked on pytrhon-dev, email forwarded to jython-users by Jeff Allen:
https://mail.python.org/pipermail/python-dev/2019-April/157255.html

cjw296:
I'm not sure this code ever belonged in the cpython repo; it's code that was inherited when mock was moved into the core, and my feeling it that it probably shouldn't have been.

Maybe I'm missing something? Are there other examples of Jython/IronPython/pypy conditionals inside the cpython codebase?

The Jython mailing list looks quite dead, and my understanding is that there's not Python 3 support in Jython, which means it'll not be supported come the end of the year.

berkerpeksag:
By the way, the recommended form for #NNNN PR numbers is GH-NNNN, not GH#NNNN :) See https://devguide.python.org/pullrequest/#making-good-commits for more details.

cjw296:
Thanks, it's such a shame we couldn't have used GH's defaults, I'm forever getting this wrong, thought I'd finally cracked it here since the bot didn't moan, but I shall try harder.

PR #13037, opened 2019-05-01
bpo-36763: _PyCoreConfig_SetPyArgv() preinitializes Python

vstinner:
_PyCoreConfig_SetPyArgv() now pre-initializes Python if needed to
ensure that the locale encoding is properly configured.

* Add _Py_PreInitializeFromPyArgv() internal function.
* Add 'args' parameter to _Py_PreInitializeFromCoreConfig()


vstinner:
I'm not sure about this change since it can lead to mojibake:
https://github.com/python/peps/pull/1018#issuecomment-488286085

PR #13039, opened 2019-05-01
bpo-36769: Document that fnmatch.filter supports any kind of iterable

andresdelfino:
https://bugs.python.org/issue36769

### DIFF
--- a/Doc/library/fnmatch.rst
+++ b/Doc/library/fnmatch.rst
@@ -75,7 +75,7 @@ patterns.
 
 .. function:: filter(names, pattern)
 
-   Return the subset of the list of *names* that match *pattern*. It is the same as
+   Return a list of items in iterable *names* that match *pattern*. It is the same as
    ``[n for n in names if fnmatch(n, pattern)]``, but implemented more efficiently.
 
 
--- a/Doc/library/fnmatch.rst
+++ b/Doc/library/fnmatch.rst
@@ -75,7 +75,7 @@ patterns.
 
 .. function:: filter(names, pattern)
 
-   Return a list of items in iterable *names* that match *pattern*. It is the same as
+   Construct a list from those elements of the iterable *names* that match *pattern*. It is the same as
    ``[n for n in names if fnmatch(n, pattern)]``, but implemented more efficiently.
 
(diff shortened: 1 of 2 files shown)


auvipy (approved):

JulienPalard:
@csabella would you please review the proper order of words in the proposed english sentence? As a french speaker I would have written ```Return a list of items in the *names* iterable``` but I have no idea if it's worse or better :]

JulienPalard:
(Closing and re-opening to give another chance to Travis, please ignore)

andresdelfino:
@JulienPalard Cheryl isn't active on GitHub since July. Perhaps someone else can help us here?

andresdelfino:
HI @csabella :) Could you review the proper order of words in the proposed English sentence, please?

Doc/library/fnmatch.rst
```diff
@@ -75,7 +75,7 @@ patterns.
 
 .. function:: filter(names, pattern)
 
-   Return the subset of the list of *names* that match *pattern*. It is the same as
+   Return a list of items in iterable *names* that match *pattern*. It is the same as
```
csabella: I would be more inclined to say `Return a list of items from *names* that match *pattern*.`  I don't think the usage of the word `list` in the original necessary meant a Python list type, just that names was a sequence of one or more items.  The next line with the equivalency is enough to show that `names` can be an iterable.

andresdelfino: Hmm, I'd say that whenever it is simple enough, it's better to document what's is expected and whas is returned in the API description, instead of examples. IMHO, examples should just illustrate not give actual information.

andresdelfino: @csabella Any thought regarding my last comment?

iritkatriel: How about borrowing the language from the doc of builtin filter:

"Construct a list from those elements of the iterable names for which.. "

andresdelfino: @iritkatriel I like it. Just updated the PR with that change.

JulienPalard (approved):

PR #13040, opened 2019-05-01
bpo-29512: Change bisect to bisect_cmd in docstring

tirkarthi:
`bisect` was renamed to `bisect_cmd` https://bugs.python.org/issue29512#msg318191. This changes the docstrings as per the update.

https://bugs.python.org/issue29512


tirkarthi:
cc: @vstinner 

Thanks for this tool :)

ZackerySpytz (approved):

PR #13044, opened 2019-05-01
bpo-30458: Use InvalidURL instead of ValueError.

gpshead:
Change the fix to raise InvalidURL for consistency with other http.client APIs instead of ValueError.


hemberger:
After this change, the documentation for the `InvalidURL` exception is no longer accurate (it only mentions the port case): https://docs.python.org/3/library/http.client.html#http.client.InvalidURL

vstinner:
Please open an issue at bugs.python.org to suggest updating InvalidURL documentation.

PR #13058, opened 2019-05-02
bpo-36776: Add @support.skip_unless_symlink to test_lll.py

ZackerySpytz:
https://bugs.python.org/issue36776


ZackerySpytz:
@vstinner Please take a look.

serhiy-storchaka (approved):

PR #12667, opened 2019-04-02
bpo-36475: Finalize PyEval_AcquireLock() and PyEval_AcquireThread() properly

nanjekyejoannah:
I have added changes to finalize finalize `PyEval_AcquireLock() `and `PyEval_AcquireThread() `properly.

https://bugs.python.org/issue36475

PR #13068, opened 2019-05-02
bpo-36475: Make PyThread_exit_thread with _Py_NO_RETURN

vstinner:
https://bugs.python.org/issue36475


vstinner:
@ncoghlan @nanjekyejoannah: Would you mind to review this change?

It's a follow-up of PR #12667.

ncoghlan (approved):

vstinner:
Thanks for the review Nick!

PR #13069, opened 2019-05-02
bpo-36778: fix test_startup_imports if default codepage is UTF-8

paulmon:
Windows desktop skus have a default ANSI codepage (returned by GetACP()) of 1252 (Western European).  Windows IoT Core and Windows Nano Server have a default codepage of 65001 (UTF-8). 

This causes test_site.StartupImportTests.test_startup_imports to fail on Windows IoT Core and Windows Nano Server because cp65001.py is loaded instead of the frozen cp1252.py at startup.

I tried changing the default codepage to 65001 on my dev machine and rebuilding Python and it had no effect that I could tell on the generated frozen importlibs.

The simplest solutions would be for the test_startup_imports test to be skipped or changed to pass when the locale.getpreferredencoding() returns 'cp65001'

@zooba @zware

### DIFF
--- a/Lib/test/test_site.py
+++ b/Lib/test/test_site.py
@@ -23,6 +23,9 @@ import tempfile
 from unittest import mock
 from copy import copy
 
+if sys.platform == 'win32':
+    import locale
+
 # These tests are not particularly useful if Python was invoked with -S.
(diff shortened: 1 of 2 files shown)


paulmon:
Sorry I clicked the wrong button in the branch view on my fork.

PR #13073, opened 2019-05-03
bpo-36779: time.tzname returns empty string on Windows if default cod…

paulmon:
Calling setlocale(LC_CTYPE, "") on a system where GetACP() returns CP_UTF8 results in empty strings in _tzname[].

This causes time.tzname to be an empty string.
I have reported the bug to the UCRT team and will follow up, but it will take some time get a fix into production.

In the meantime one possible workaround is to temporarily change the locale by calling setlocale(LC_CTYPE, "C") before calling _tzset and restore the current locale after if the GetACP() == CP_UTF8 or CP_UTF7

@zooba 

https://bugs.python.org/issue36779


paulmon:
This workaround enables time.tzname to work on Windows ARM32.
Alternatively we could skip the test and leave time.tzname broken on Windows ARM32.
The CRT error which causes the empty string has been fixed in ucrtbase.dll but will likely take months to service.

matrixise:
Hi @paulmon I have edited the PR because that was the wrong BPO issue. Also, I have added this PR to the right bpo issue and removed it from the other one.

paulmon:
Thank you!

zooba:
Definitely needs the tabs converted to spaces, and I wonder if there's a better (Windows-specific) way to get this value without changing the locale? If we're going to have an ifdef, we may as well use the proper APIs, provided the values come back the same (I have no idea what values to expect here).

@pganssle may also have an opinion. Or perhaps not :)

paulmon:
I'll try to get the timezone name without changing the locale.  
And I'll fix the tabs of course.

pganssle:
Regardless of the resolution, I think this needs a test.

What is this supposed to return on Windows? What does it return with the workaround?

I think you can definitely get the time zone and alt zone names without changing the locale. One way to get it is from the registry, which is how we do it [in `tzwinlocal`](https://github.com/dateutil/dateutil/blob/40924bd0392308914b01815f56d311c8478217b8/dateutil/tz/win.py#L276). I think this returns things like "Eastern Standard Time" and "Eastern Daylight Time" rather than "EST"/"EDT".

There may be a more convenient way as well.

paulmon:
This is the fix for a failing test.  It's supposed to return 'UTC-08:00' and it returns an empty string.

FAIL: test_strptime (test.datetimetester.TestSubclassDateTime_Fast)

  File "C:\python\lib\test\datetimetester.py", line 2523, in test_strptime
    self.assertEqual(dt.tzname(), tzname)
AssertionError: 'UTC-08:00' != ''
- UTC-08:00
+

pganssle:
> Traceback (most recent call last):
> File "C:\python\lib\test\datetimetester.py", line 2523, in test_strptime
> self.assertEqual(dt.tzname(), tzname)
> AssertionError: 'UTC-08:00' != ''

This is very confusing to me. [This is the relevant test](https://github.com/python/cpython/blob/b82e17e626f7b1cd98aada0b1ebb65cb9f8fb184/Lib/test/datetimetester.py#L2511-L2523), and it seems like what it's doing is creating three strings: `"+0000 UTC"`, `"-0000 GMT"` and a locale-specific one.  If `time.tzname` returns an empty string, I would expect [this line](https://github.com/python/cpython/blob/b82e17e626f7b1cd98aada0b1ebb65cb9f8fb184/Lib/test/datetimetester.py#L2521) to raise an Exception, since the locale-specific string would be something like `"-0800 "`, which doesn't match `%z %Z`.

You can see what it's testing by running this script:

```python
import time
from datetime import datetime

for tzseconds, tzname in ((0, 'UTC'), (0, 'GMT'),
                         (-time.timezone, time.tzname[0])):
    print("-----")
    print(f"tzseconds: {tzseconds}; tzname: {tzname}")
    if tzseconds < 0:
        sign = '-'
        seconds = -tzseconds
    else:
        sign ='+'
        seconds = tzseconds
    hours, minutes = divmod(seconds//60, 60)
    dtstr = "{}{:02d}{:02d} {}".format(sign, hours, minutes, tzname)
    print(dtstr)
```

On my system, this returns:

```
-----
tzseconds: 0; tzname: UTC
+0000 UTC
-----
tzseconds: 0; tzname: GMT
+0000 GMT
-----
tzseconds: -18000; tzname: EST
-0500 EST
```

I am not sure where `UTC-0800` is coming from. Can you run my script on your system so we can see what's going on?

paulmon:
-----
tzseconds: 0; tzname: UTC
+0000 UTC
-----
tzseconds: 0; tzname: GMT
+0000 GMT
-----
tzseconds: -28800; tzname:
-0800

paulmon:
UTC-0800 is my local time zone
```
import time
time.tzname
```
This returns  `('Pacific Standard Time', 'Pacific Daylight Time')` on desktop windows
On Windows IoT Core it return `('', '')`

I tracked this this down to a bug in tzset() that sets _tzname[0] and _tzname[1] to empty strings when the default ANSI code pageis set to CP_UTF8, or CP_UTF7.  For more on the default ANSI code page see , see [GetACP( )](https://docs.microsoft.com/en-us/windows/desktop/api/winnls/nf-winnls-getacp)

The default ANSI code page on most Windows installs is not unicode.  On my dev machine for example it is 1252 ([Code Page Identifiers](https://docs.microsoft.com/en-us/windows/desktop/Intl/code-page-identifiers)).

On Windows IoT Core the default ANSI code page is 65001 or CP_UTF8.

So in line 2511 of datetimetester.py _time.tzname[0] is empty causing the test to fail.

```
        for tzseconds, tzname in ((0, 'UTC'), (0, 'GMT'),
                                 (-_time.timezone, _time.tzname[0])):
```

paulmon:
> I would expect [this line](https://github.com/python/cpython/blob/b82e17e626f7b1cd98aada0b1ebb65cb9f8fb184/Lib/test/datetimetester.py#L2521) to raise an Exception

I think it doesn't raise an exception because it's valid to use "" as parameter to a format string.

pganssle:
> This returns `('Pacific Standard Time', 'Pacific Daylight Time')` on desktop windows
> On Windows IoT Core it returns `('', '')`

This is what I expected, but I don't understand why the test was working in the first place, on Linux:

```python
>>> datetime.strptime("-0800 ", '%z %Z')
...
ValueError: time data '-0800 ' does not match format '%z %Z'

>>> datetime.strptime("-0800 Pacific Standard Time", "%z %Z")
...
ValueError: time data '-0800 Pacific Standard Time' does not match format '%z %Z'
```

Is allowing empty format specifiers a Windows-specific thing?

@paulmon If the two commands above don't fail for you, what do they return? What is `.tzname()` on the datetimes they return?

paulmon:
Closing and re-opening to force azure pipelines to run again.
Not sure why it failed.  It looked like the MacOS tests were cancelled for some reason.

paulmon:
> If the two commands above don't fail for you, what do they return? What is `.tzname()` on the datetimes they return?

There seems to be a connection between the value of time.tzname and the results of those two commands.

Without the current GetTimeZoneInformation fix

```
>>> datetime.strptime("-0800 ", "%z %Z")
datetime.datetime(1900, 1, 1, 0, 0, tzinfo=datetime.timezone(datetime.timedelta(days=-1, seconds=57600)))
>>> datetime.strptime("-0800 Pacific Standard Time", "%z %Z")
  File "C:\python\lib\_strptime.py", line 568, in _strptime_datetime
    tt, fraction, gmtoff_fraction = _strptime(data_string, format)
  File "C:\python\lib\_strptime.py", line 352, in _strptime
    raise ValueError("unconverted data remains: %s" %
ValueError: unconverted data remains: Pacific Standard Time
```

With the GetTimeZoneInformation fix 

```
>>> datetime.strptime("-0800 ", "%z %Z")
  File "C:\python\lib\_strptime.py", line 568, in _strptime_datetime
    tt, fraction, gmtoff_fraction = _strptime(data_string, format)
  File "C:\python\lib\_strptime.py", line 349, in _strptime
    raise ValueError("time data %r does not match format %r" %
ValueError: time data '-0800 ' does not match format '%z %Z'
>>> datetime.strptime("-0800 Pacific Standard Time", "%z %Z")
datetime.datetime(1900, 1, 1, 0, 0, tzinfo=datetime.timezone(datetime.timedelta(days=-1, seconds=57600), 'Pacific Standard Time'))
>>> datetime.strptime("-0800 Pacific Standard Time", "%z %Z").tzname()
'Pacific Standard Time'
```

paulmon:
The PR has been updated so that it doesn't change locale.  
Does it look good, or do you think it needs more changes?
Thanks!

@zooba @pganssle

zooba:
@paulmon I think this one deserves a NEWS item - could you add it?

zooba (approved):

pganssle:
@paulmon Hey sorry I never followed up on this - this is belated, but for the record I agree with this change. Thank you for for doing this - particularly because I learned something about how `strptime` works (I didn't realize that it checks `tzname` if available, though it makes a lot of sense).

vstinner:
> https://buildbot.python.org/all/#builders/211/builds/47

I reported this bug as: https://bugs.python.org/issue37314

paulmon:
> > https://buildbot.python.org/all/#builders/211/builds/47
> 
> I reported this bug as: https://bugs.python.org/issue37314

It looks like a different commit causing this buildbot failure.   The only file changed in this commit is timemodule.c, and the build failures are all in _iomodule.c

vstinner:
Welcome to the exciting universe of buildbots!

PR #13078, opened 2019-05-03
bpo-36786: Run compileall in parallel during "make install"

pitrou:
https://bugs.python.org/issue36786


pitrou:
@brettcannon @vadmium @berkerpeksag

berkerpeksag (approved):

PR #13079, opened 2019-05-03
bpo-34155: Dont parse domains containing @

jpic:
Before:
    
        >>> email.message_from_string('From: a@malicious.org@important.com', policy=email.policy.default)['from'].addresses
        (Address(display_name='', username='a', domain='malicious.org'),)
    
        >>> parseaddr('a@malicious.org@important.com')
        ('', 'a@malicious.org')
    
    After:
    
        >>> email.message_from_string('From: a@malicious.org@important.com', policy=email.policy.default)['from'].addresses
        (Address(display_name='', username='', domain=''),)
    
        >>> parseaddr('a@malicious.org@important.com')
        ('', 'a@')

https://bugs.python.org/issue34155

Automerge-Triggered-By: @warsaw

### DIFF
--- a/Lib/email/_header_value_parser.py
+++ b/Lib/email/_header_value_parser.py
@@ -1587,6 +1587,8 @@ def get_domain(value):
         token, value = get_dot_atom(value)
     except errors.HeaderParseError:
         token, value = get_atom(value)
+    if value and value[0] == '@':
+        raise errors.HeaderParseError('Invalid Domain')
     if leader is not None:
         token[:0] = [leader]
     domain.append(token)
--- a/Lib/email/_parseaddr.py
+++ b/Lib/email/_parseaddr.py
@@ -379,7 +379,12 @@ class AddrlistClass:
         aslist.append('@')
         self.pos += 1
         self.gotonext()
-        return EMPTYSTRING.join(aslist) + self.getdomain()
+        domain = self.getdomain()
+        if not domain:
+            # Invalid domain, return an empty address instead of returning a
+            # local part to denote failed parsing.
+            return EMPTYSTRING
+        return EMPTYSTRING.join(aslist) + domain
 
     def getdomain(self):
         """Get the complete domain name from an address."""
@@ -394,6 +399,10 @@ class AddrlistClass:
             elif self.field[self.pos] == '.':
                 self.pos += 1
                 sdlist.append('.')
+            elif self.field[self.pos] == '@':
+                # bpo-34155: Don't parse domains with two `@` like
+                # `a@malicious.org@important.com`.
+                return EMPTYSTRING
             elif self.field[self.pos] in self.atomends:
                 break
             else:
--- a/Lib/test/test_email/test_email.py
+++ b/Lib/test/test_email/test_email.py
@@ -3041,6 +3041,20 @@ class TestMiscellaneous(TestEmailBase):
         self.assertEqual(utils.parseaddr('<>'), ('', ''))
         self.assertEqual(utils.formataddr(utils.parseaddr('<>')), '')
 
+    def test_parseaddr_multiple_domains(self):
+        self.assertEqual(
+            utils.parseaddr('a@b@c'),
+            ('', '')
+        )
+        self.assertEqual(
+            utils.parseaddr('a@b.c@c'),
+            ('', '')
+        )
+        self.assertEqual(
+            utils.parseaddr('a@172.17.0.1@c'),
+            ('', '')
+        )
+
     def test_noquote_dump(self):
         self.assertEqual(
             utils.formataddr(('A Silly Person', 'person@dom.ain')),
--- a/Misc/NEWS.d/next/Security/2019-05-04-13-33-37.bpo-34155.MJll68.rst
+++ b/Misc/NEWS.d/next/Security/2019-05-04-13-33-37.bpo-34155.MJll68.rst
@@ -1,0 +1,1 @@
+Fix parsing of invalid email addresses with more than one ``@`` (e.g. a@b@c.com.) to not return the part before 2nd ``@`` as valid email address. Patch by maxking & jpic.
--- a/Lib/test/test_email/test__header_value_parser.py
+++ b/Lib/test/test_email/test__header_value_parser.py
@@ -1444,6 +1444,16 @@ class TestParser(TestParserMixin, TestEmailBase):
         self.assertEqual(addr_spec.domain, 'example.com')
         self.assertEqual(addr_spec.addr_spec, 'star.a.star@example.com')
 
+    def test_get_addr_spec_multiple_domains(self):
+        with self.assertRaises(errors.HeaderParseError):
+            parser.get_addr_spec('star@a.star@example.com')
+
+        with self.assertRaises(errors.HeaderParseError):
+            parser.get_addr_spec('star@a@example.com')
+
+        with self.assertRaises(errors.HeaderParseError):
+            parser.get_addr_spec('star@172.17.0.1@example.com')
+
     # get_obs_route
 
     def test_get_obs_route_simple(self):


the-knights-who-say-ni:

Lib/email/_header_value_parser.py
```diff
@@ -1559,6 +1559,8 @@ def get_domain(value):
         token, value = get_dot_atom(value)
     except errors.HeaderParseError:
         token, value = get_atom(value)
+    if value and value[0] == '@':
+        raise errors.HeaderParseError('Multiple domains')
```
maxking: I'd rename the error as `Invalid Domain`.

Lib/test/test_email/test_email.py
```diff
@@ -3032,6 +3032,20 @@ def test_parseaddr_empty(self):
         self.assertEqual(utils.parseaddr('<>'), ('', ''))
         self.assertEqual(utils.formataddr(utils.parseaddr('<>')), '')
 
+    def test_parseaddr_multiple_domains(self):
+        self.assertEqual(
+            utils.parseaddr('a@b@c'),
+            ('', 'a@')
```
maxking: According to the docs:
> email.utils.parseaddr(address)
>    Parse address – which should be the value of some address-containing field such as To or Cc – into its constituent realname and email address parts. Returns a tuple of that information, unless the parse fails, in which case a 2-tuple of ('', '') is returned.

Can we expect the same in this case?

I made this change to make that happen:
```
@@ -379,7 +379,12 @@ class AddrlistClass:
         aslist.append('@')
         self.pos += 1
         self.gotonext()
-        return EMPTYSTRING.join(aslist) + self.getdomain()
+        domain = self.getdomain()
+        if not domain:
+            # Invalid domain, return an empty address instead of returning a
+            # local part to denote failed parsing.
+            return EMPTYSTRING
+        return EMPTYSTRING.join(aslist) + domain
```

Misc/NEWS.d/next/Security/2019-05-04-13-33-37.bpo-34155.MJll68.rst
```diff
@@ -0,0 +1 @@
+Don't parse email domain containing an at, ie. a@malicious.org@important.com
```
maxking: I would suggest a slightly different wording for this:

```
Fix parsing of invalid email addresses with more than one ``@`` (e.g. a@b@c.com.) to not return the part before 2nd ``@`` as valid email address. Patch by jpic.
```

maxking (commented):
Thanks for the patch!

I have some different ideas on how to do error recovery in this case, please see [my comment](https://bugs.python.org/msg344030) on bpo-34155.

Lib/test/test_email/test_email.py
jpic: Sure, please confirm that I got it right:

You mean that `utils.parseaddr('a@b@c')` should return `('', '')`, or just `''` ?

Just one question about consistency, would that change also implicate a change of the following : 

```
>>> parseaddr('foo@a.') 
('', 'foo@a.')
```

Should it also return same as with erroneous caracter @ in domain ? (an empty string or tuple of empty strings)

If so, should that also be in the security patch or do you want a new issue and a new patch for breaking backward compatibility ?

To be completely honest with you, I was convinced this patch would pass because I didn't see any other implementation that I could propose on my own in the context of closing the security issue. I see that there is more work to keep on doing in the email library and am looking forward to do more, in this patch request or another, I'm all yours for homework assignments @maxking.

Thank you for your review !

jpic:
Pushed a couple of fixes in new commits in my branch, left a question for you.

Thank you very much for your review @maxking ! You're max kind ;)

Lib/test/test_email/test_email.py
maxking: Yes, I meant return value should be `('', '')`.

Coming to the behavior of `parseaddr`, here is my thinking on this this issue. Right now, we return either "whatever we can parse" (which is why we return the invalid domains as-is without opinions) and `('', '')` when there is an exception raised during parsing.

So, following this line of reasoning, with domains like `a@b@c.com` should actually return `('', ,'a@b@c.com' )`. I think this is the right thing to do. *If* you think this behavior can cause security issue for the downstreams who might have similar bug and would take in `a@b@c.com` as `a@b` and think we should consider this, then the next right thing to do would be to return an empty string denoting unparseable address.

Does that line of reasoning made sense like it did in my head?

jpic: Yes of course it is expected to break BC to fix security issue. In this case do you want this patch to also change parseaddr('a@b.') to return a tuple of two empty strings in this patch for consistency with parseaddr('a@b@c') which is now going to return an empty strings tuple ? Sorry you make sense but there i am not sure if this consistency point is important or not for the time being

jpic: I think we can deal with this later, so I added your patch, thanks @maxking ! But I didn't know how to add credit

maxking: You can add credits in commit message, I think.

You may need to update your tests to reflect the patch too, they seem to be failing right now with new updates.

warsaw:
As I mentioned in [this comment on bpo](https://bugs.python.org/issue34155?@ok_message=msg%20347157%20created%0Aissue%2034155%20message_count%2C%20messages%2C%20versions%20edited%20ok&@template=item#msg347157) I think that `parseaddr('a@b@c')` has to return `('', '')`.  To me, it's the only sane return value for illegal addresses.

Lib/email/_header_value_parser.py
```diff
@@ -1584,6 +1584,8 @@ def get_domain(value):
         token, value = get_dot_atom(value)
     except errors.HeaderParseError:
         token, value = get_atom(value)
+    if value and value[0] == '@':
+        raise errors.HeaderParseError('Invalid Domain')
```
warsaw, line 1591: I wonder if this should add a defect instead of raising a `HeaderParseError`?  I can't decide.  I suppose the exception is okay as long as it doesn't get exposed in the higher level email parsing APIs (e.g. `message_from_string()`, since those are not allowed to raise exceptions.

Lib/email/_parseaddr.py
```diff
@@ -394,6 +400,9 @@ def getdomain(self):
             elif self.field[self.pos] == '.':
                 self.pos += 1
                 sdlist.append('.')
+            elif self.field[self.pos] == '@':
+                #34155 Don't parse domain for a@malicious.org@important.com
```
warsaw: Please label this `# bpo-34155` since it has to be unique in case we eventually do change bug trackers.

maxking:
I totally agree on the fact that we should return tuple of two empty strings on "failed to parse", whether on not we decide to be opinionated about what we fail to parse.

jpic:
Understood, it was decided to deal with that later if it had to be dealt with at all anyway.

I have fixed my own tests, but the new implementation introduces a bunch of regressions that I'm digging into.

```
FAIL: test_escape_backslashes (test.test_email.test_email.TestMiscellaneous)
  File "/home/travis/build/python/cpython/Lib/test/test_email/test_email.py", line 3080, in test_escape_backslashes
    self.assertEqual(utils.parseaddr(utils.formataddr((a, b))), (a, b))
AssertionError: Tuples differ: ('Arthur \\Backslash\\ Foobar', 'person@') != ('Arthur \\Backslash\\ Foobar', 'person@dom.ain')
First differing element 1:
'person@'
'person@dom.ain'
- ('Arthur \\Backslash\\ Foobar', 'person@')
+ ('Arthur \\Backslash\\ Foobar', 'person@dom.ain')
?                                         +++++++
FAIL: test_escape_dump (test.test_email.test_email.TestMiscellaneous)
  File "/home/travis/build/python/cpython/Lib/test/test_email/test_email.py", line 3067, in test_escape_dump
    self.assertEqual(
AssertionError: Tuples differ: ('A (Very) Silly Person', 'person@') != ('A (Very) Silly Person', 'person@dom.ain')
First differing element 1:
'person@'
'person@dom.ain'
- ('A (Very) Silly Person', 'person@')
+ ('A (Very) Silly Person', 'person@dom.ain')
?                                   +++++++
FAIL: test_formataddr_does_not_quote_parens_in_quoted_string (test.test_email.test_email.TestMiscellaneous)
  File "/home/travis/build/python/cpython/Lib/test/test_email/test_email.py", line 3172, in test_formataddr_does_not_quote_parens_in_quoted_string
    self.assertEqual(utils.parseaddr(addrstr), addr)
AssertionError: Tuples differ: ("'foo@example.com' (foo@example.com)", 'foo@') != ("'foo@example.com' (foo@example.com)", 'foo@example.com')
First differing element 1:
'foo@'
'foo@example.com'
- ("'foo@example.com' (foo@example.com)", 'foo@')
+ ("'foo@example.com' (foo@example.com)", 'foo@example.com')
?                                              +++++++++++
FAIL: test_getaddresses (test.test_email.test_email.TestMiscellaneous)
  File "/home/travis/build/python/cpython/Lib/test/test_email/test_email.py", line 3209, in test_getaddresses
    eq(utils.getaddresses(['aperson@dom.ain (Al Person)',
AssertionError: Lists differ: [('Al Person', 'aperson@'), ('Bud Person', 'bperson@')] != [('Al Person', 'aperson@dom.ain'), ('Bud Person', 'bperson@dom.ain')]
First differing element 0:
('Al Person', 'aperson@')
('Al Person', 'aperson@dom.ain')
- [('Al Person', 'aperson@'), ('Bud Person', 'bperson@')]
+ [('Al Person', 'aperson@dom.ain'), ('Bud Person', 'bperson@dom.ain')]
?                         +++++++                            +++++++
FAIL: test_getaddresses_embedded_comment (test.test_email.test_email.TestMiscellaneous)
Test proper handling of a nested comment
  File "/home/travis/build/python/cpython/Lib/test/test_email/test_email.py", line 3228, in test_getaddresses_embedded_comment
    eq(addrs[0][1], 'foo@bar.com')
AssertionError: 'foo@' != 'foo@bar.com'
- foo@
+ foo@bar.com
FAIL: test_getaddresses_nasty (test.test_email.test_email.TestMiscellaneous)
  File "/home/travis/build/python/cpython/Lib/test/test_email/test_email.py", line 3220, in test_getaddresses_nasty
    eq(utils.getaddresses(
AssertionError: Lists differ: [('', ''), ('Jason R. Mastaler', 'jason@')] != [('', ''), ('Jason R. Mastaler', 'jason@dom.ain')]
First differing element 1:
('Jason R. Mastaler', 'jason@')
('Jason R. Mastaler', 'jason@dom.ain')
- [('', ''), ('Jason R. Mastaler', 'jason@')]
+ [('', ''), ('Jason R. Mastaler', 'jason@dom.ain')]
?                                         +++++++
FAIL: test_multiline_from_comment (test.test_email.test_email.TestMiscellaneous)
  File "/home/travis/build/python/cpython/Lib/test/test_email/test_email.py", line 3180, in test_multiline_from_comment
    self.assertEqual(utils.parseaddr(x), ('Foo Bar', 'foo@example.com'))
AssertionError: Tuples differ: ('Foo Bar', 'foo@') != ('Foo Bar', 'foo@example.com')
First differing element 1:
'foo@'
'foo@example.com'
- ('Foo Bar', 'foo@')
+ ('Foo Bar', 'foo@example.com')
?                  +++++++++++
FAIL: test_name_with_dot (test.test_email.test_email.TestMiscellaneous)
  File "/home/travis/build/python/cpython/Lib/test/test_email/test_email.py", line 3126, in test_name_with_dot
    self.assertEqual(utils.parseaddr(x), (a, b))
AssertionError: Tuples differ: ('John X. Doe', 'jxd@') != ('John X. Doe', 'jxd@example.com')
First differing element 1:
'jxd@'
'jxd@example.com'
- ('John X. Doe', 'jxd@')
+ ('John X. Doe', 'jxd@example.com')
?                      +++++++++++
FAIL: test_parseaddr_multiple_domains (test.test_email.test_email.TestMiscellaneous)
  File "/home/travis/build/python/cpython/Lib/test/test_email/test_email.py", line 3045, in test_parseaddr_multiple_domains
    self.assertEqual(
AssertionError: Tuples differ: ('', '') != ('', 'a@')
First differing element 1:
''
'a@'
- ('', '')
+ ('', 'a@')
?       ++
FAIL: test_parseaddr_preserves_quoted_pairs_in_addresses (test.test_email.test_email.TestMiscellaneous)
  File "/home/travis/build/python/cpython/Lib/test/test_email/test_email.py", line 3142, in test_parseaddr_preserves_quoted_pairs_in_addresses
    eq(utils.parseaddr('""example" example"@example.com'),
AssertionError: Tuples differ: ('', '""example" example"@') != ('', '""example" example"@example.com')
First differing element 1:
'""example" example"@'
'""example" example"@example.com'
- ('', '""example" example"@')
+ ('', '""example" example"@example.com')
?                           +++++++++++
FAIL: test_parseaddr_preserves_spaces_in_local_part (test.test_email.test_email.TestMiscellaneous)
  File "/home/travis/build/python/cpython/Lib/test/test_email/test_email.py", line 3156, in test_parseaddr_preserves_spaces_in_local_part
    self.assertEqual(('', "merwok wok@xample.com"),
AssertionError: Tuples differ: ('', 'merwok wok@xample.com') != ('', 'merwok wok@')
First differing element 1:
'merwok wok@xample.com'
'merwok wok@'
- ('', 'merwok wok@xample.com')
?                  ----------
+ ('', 'merwok wok@')
```

jpic:
Just one last question please, currently with this patch we get:

```
>>> parseaddr('a@b@c')                
('', '')
```

But when there is an invalid domain such as `b.`, it doesn't return an empty string:

```
>>> parseaddr('a@b.')
('', 'a@b.')
```

Do you also want this patch to make `parseaddr('a@b.')` to return a tuple of empty strings ?

Or is it out of the scope of this patch ?

Thanks a heap for your support ! Deeply appreciated ;)

maxking:
> Do you also want this patch to make parseaddr('a@b.') to return a tuple of empty strings ?
> Or is it out of the scope of this patch ?

That seems to be a change of behaviour, even though it is going to align with what we say in docs. Perhaps for the the short-term, we should make the docs align a bit more with the reality. 

In the long term, we can decide if we want to return empty string where the email address is technically invalid. As it stands today, "failed to parse" means there is no way to read the value at all, so that includes weird encodings, random chars etc. I am not sure if we do RFC level validation for what is and isn't a valid address (maybe we do, haven't read those parts of the code too closely, but that is my current impression) so it may or may not be tricky to weed out all invalid addresses.

That discussion can perhaps start on a new issue and follow the usual change-of-behavior process and should be fixed in a different patch. This being a security fix will go back to 3.6 as well, so I guess we shouldn't do that here.

> Or is a@b. not an illegal address at all ? I thought it was but I'm having a last minute doubt now (sorry about that)

It is an invalid address AFAIK, I can't quote the right RFC page to corroborate that right now, will need some more time for that.

jpic:
Thanks for your explanation.  Please let me know if there's anything else you want me to do. If you want me to open the other issue I will start studying the corresponding contribution docs and try my best to match Python's standard expectations.

Lib/email/_parseaddr.py
```diff
@@ -394,6 +399,9 @@ def getdomain(self):
...
```
maxking: Can you please make this
```
# bpo-34155: Don't parse domains with two `@` like `a@malicious.org@important.com`.
```

We prefer to use `bpo-` to link to issues in the code, which stands for bugs.python.org :)

maxking (commented):
The patch looks good to me, just a very teeny tiny nitpick that I have added inline comment for.

@warsaw I think after the fix for the comment, this patch should be good to go! 
Needs backport to 3.6, 3.7, 3.8 since this has security implications.

maxking:
@jpic I think it would be good to create a new issue to discuss if we want to change behaviour of `parseaddr` to validate valid addresses.

Lib/email/_parseaddr.py
jpic: The comment was made to fit without exceeding 80 caracters :joy:

Would this work for you ?

```
                # bpo-34155: Don't parse domains with two `@` like
                # `a@malicious.org@important.com`.
```

Or

```
# bpo-34155: Don't parse two @ like a@malicious@ok.com
```

jpic:
There goes the [new issue](https://bugs.python.org/issue37492), hope I did it right, otherwise let me know of course ;)

Lib/email/_parseaddr.py
maxking: I think two lines are fine, we don't really need to fit it into a single line.

jpic: Thanks for your reply it's pushed !

jpic:
Tests passed except MacOS job which failed on Azure, but it doesn't seem related to this patch.

2019-07-04-183254_1478x905_scrot

jpic:
Do you also want PR against 3.6, 3.7, 3.8 since this has security implications ?

maxking:
I am not sure about the failing tests, maybe we can rebase this against master and hope that it is fixed there? :) 

I am currently travelling, so won't be able to help out with debugging much for this week.

And no, you don't need to manually create backports, I added the comment for whoever merges this to add right "needs backport to xxx" labels, which would then trigger the bot to create backport PRs.

jpic:
Rebasing worked.
Thanks @maxking for your support and review and pedagogy :tophat:

maxking:
This LGTM! 

Thanks to *you* @jpic for your patience and contribution :)

Lib/email/_parseaddr.py
jpic: Is this version ok ?

https://github.com/python/cpython/pull/13079/files#diff-18392c73bfdad73b43ab16d2adfd5b1eR403

Lib/email/_header_value_parser.py
jpic: The [HeaderParseError is expected by the higher level API which adds a defect](https://github.com/python/cpython/blob/02257012f6d3821d816cb6a7e8461a88a05b9a08/Lib/email/_header_value_parser.py#L1956) so you don't have to take any decision. As a user you can still know about that parse problem using the defects attribute:
```
>>> email.message_from_string('From: a@malicious.org@example.com',
policy=email.policy.default)['from'].defects[0]
InvalidHeaderDefect('invalid address in address-list')
```
https://bugs.python.org/msg341381

jpic: Updated the above comment with a link to the expectation of HeaderParseError: https://github.com/python/cpython/blob/02257012f6d3821d816cb6a7e8461a88a05b9a08/Lib/email/_header_value_parser.py#L1956

warsaw (approved):

ned-deily:
@warsaw, do you still want the backports of this PR to be merged?  They are still awaiting review and merging.

maxking:
@ned-deily It would be good to merge the backports for 3.6, I can merge the ones for 3.7 and 3.8.

PR #13098, opened 2019-05-05
bpo-36801: Fix waiting in StreamWriter.drain for closing SSL transport

asvetlov:
https://bugs.python.org/issue36801

### DIFF
--- a/Lib/asyncio/streams.py
+++ b/Lib/asyncio/streams.py
@@ -179,6 +179,9 @@ class FlowControlMixin(protocols.Protocol):
         self._drain_waiter = waiter
         await waiter
 
+    def _get_close_waiter(self, stream):
+        raise NotImplementedError
+
 
 class StreamReaderProtocol(FlowControlMixin, protocols.Protocol):
     """Helper class to adapt between Protocol and StreamReader.
@@ -293,6 +296,9 @@ class StreamReaderProtocol(FlowControlMixin, protocols.Protocol):
             return False
         return True
 
+    def _get_close_waiter(self, stream):
+        return self._closed
+
     def __del__(self):
         # Prevent reports about unhandled exceptions.
         # Better than self._closed._log_traceback = False hack
@@ -348,7 +354,7 @@ class StreamWriter:
         return self._transport.is_closing()
 
     async def wait_closed(self):
-        await self._protocol._closed
+        await self._protocol._get_close_waiter(self)
 
     def get_extra_info(self, name, default=None):
         return self._transport.get_extra_info(name, default)
@@ -369,7 +375,8 @@ class StreamWriter:
             # Wait for protocol.connection_lost() call
             # Raise connection closing error if any,
             # ConnectionResetError otherwise
-            await self._protocol._closed
+            fut = self._protocol._get_close_waiter(self)
+            await fut
             raise ConnectionResetError('Connection lost')
         await self._protocol._drain_helper()
 
--- a/Lib/asyncio/streams.py
+++ b/Lib/asyncio/streams.py
@@ -366,13 +366,11 @@ class StreamWriter:
             if exc is not None:
                 raise exc
         if self._transport.is_closing():
-            # Yield to the event loop so connection_lost() may be
-            # called.  Without this, _drain_helper() would return
-            # immediately, and code that calls
-            #     write(...); await drain()
-            # in a loop would never call connection_lost(), so it
-            # would not see an error when the socket is closed.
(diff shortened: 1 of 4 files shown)


asvetlov:
The backport to 3.7 is desired.
The problem not in SSL connections only.
Working on the patch I've found that `await subproc.stdin.wait_closed()` is broken on 3.7
Probably nobody executes this call though, nobody blames.

Lib/asyncio/streams.py
```diff
@@ -366,13 +372,12 @@ def get_extra_info(self, name, default=None):
...
-            #     write(...); await drain()
-            # in a loop would never call connection_lost(), so it
-            # would not see an error when the socket is closed.
-            await sleep(0, loop=self._loop)
+            # Wait for protocol.connection_lost() call
+            # Raise connection closing error if any,
+            # ConnectionResetError otherwise
+            fut = self._protocol._get_close_waiter(self)
+            await fut
+            raise ConnectionResetError('Connection lost')
```
1st1, line 380: Do we want to raise an error from drain()?  Maybe it should just return?  What's the point of this error?

asvetlov: That's how `protocol._drain_helper()` works now.
The future is set to exception only if the connection is closed with a failure.
But `await writer.drain()` raises `ConnectionResetError`. 
I believe this is good: `await writer.write(b'data')` should fail loudly if the socket is closed.

asvetlov: Maybe the code could be simplified by setting ConnectionResetError by `connection_lost()` callback handler but I'd like to keep the PR small.
Future improvement worth another pull request.

1st1: OK, got it.

1st1 (approved):

nullstd:
Hi @asvetlov  , I ran into a `BrokenPipeError` error when testing the following code:

```
#!/usr/bin/env python

import sys
import os
import asyncio





async def main_task():

    print(sys.version)


    xxx = "x" * 196428
    incoming_original_data = xxx.encode()

    print(f"stdin len: {len(incoming_original_data)}")

    proc = await asyncio.create_subprocess_shell(
        f"sleep 2 && xxd -l 1",
        stdin=asyncio.subprocess.PIPE,
        stdout=asyncio.subprocess.PIPE,
        stderr=asyncio.subprocess.PIPE)
    
    await proc.communicate(input = incoming_original_data)

    print("subprocess ended")

def main():

    opts = asyncio.run(main_task())

    print(f"main END")

    return opts




if __name__ == '__main__':

    main()

    
```

Note the subprocess ends very soon, and it seems `_stdin_closed` isn't properly handled because I see the following output:

```
3.11.3 (main, Apr  7 2023, 19:25:52) [Clang 14.0.0 (clang-1400.0.29.202)]
stdin len: 196428
subprocess ended
main END
Future exception was never retrieved
future: <Future finished exception=BrokenPipeError(32, 'Broken pipe')>
  File "/usr/local/Cellar/python@3.11/3.11.3/Frameworks/Python.framework/Versions/3.11/lib/python3.11/asyncio/subprocess.py", line 152, in _feed_stdin
    await self.stdin.drain()
  File "/usr/local/Cellar/python@3.11/3.11.3/Frameworks/Python.framework/Versions/3.11/lib/python3.11/asyncio/streams.py", line 378, in drain
    await self._protocol._drain_helper()
  File "/usr/local/Cellar/python@3.11/3.11.3/Frameworks/Python.framework/Versions/3.11/lib/python3.11/asyncio/streams.py", line 173, in _drain_helper
    await waiter
  File "/usr/local/Cellar/python@3.11/3.11.3/Frameworks/Python.framework/Versions/3.11/lib/python3.11/asyncio/unix_events.py", line 709, in _write_ready
    n = os.write(self._fileno, self._buffer)
        ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
BrokenPipeError: [Errno 32] Broken pipe
```

nullstd:
BTW, here are some links I think might be helpful:

https://stackoverflow.com/questions/14207708/ioerror-errno-32-broken-pipe-when-piping-prog-py-othercmd

https://stackoverflow.com/questions/23688492/oserror-errno-22-invalid-argument-in-subprocess#28020096

nullstd:
cc @1st1

gpshead:
please do not post questions and cc people on long closed/merged PRs.  Please use the discuss.python.org Help category - and only when it seems an actual CPython issue hasbeen  found, file a new issue in the issue tracker.

PR #13099, opened 2019-05-05
bpo-36802: Drop awrite()/aclose(), support await write() and await close() instead

asvetlov:
`writelines()` supports await syntax as well

https://bugs.python.org/issue36802


asvetlov:
The PR requires #13098 
That's why I've merged that code to this branch.
I'm trying to keep PRs isolated but it is not always possible, sorry.

Doc/library/asyncio-stream.rst
```diff
@@ -226,23 +226,48 @@ StreamWriter
...
    instead.
 
-   .. coroutinemethod:: awrite(data)
+   .. method:: write(data)
 
       Write *data* to the stream.
 
-      The method respects flow control, execution is paused if the write
-      buffer reaches the high watermark.
+      If the method is called using ``await stream.write(data)`` syntax it
```
1st1: I'd replace with the following:

```
The method attempts to write the data to the underlying socket immediately.
If that fails, the data is queued in an internal write buffer until it can be
sent.

Starting with Python 3.8, it is possible to directly await on the `write()` method::

     await stream.write(data)

The "await" pauses the current coroutine until the data is written to the socket.

Below is an equivalent code that works with Python <= 3.7::

    stream.write(data)
    await stream.drain()
```

Doc/library/asyncio-stream.rst
```diff
@@ -226,23 +226,48 @@ StreamWriter
...
 
-      Wait until all closing actions are complete, e.g. SSL shutdown for
-      secure sockets.
+      If the method is called using ``await stream.close()`` syntax it waits until all
+      closing actions are complete, e.g. SSL shutdown for secure sockets *(recommended
+      API)*.
 
-      .. versionadded:: 3.8
+      If the method is called using *sync* ``stream.close()`` form it *does not* wait
+      for actual socket closing. The call should be followed by :meth:`wait_closed`.
```
1st1: I'd replace with the following:

```
The method closes the stream and the underlying socket.

Starting with Python 3.8, it is possible to directly await on the `close()` method::

     await stream.close()

The "await" pauses the current coroutine until the stream and the underlying socket
are closed (and SSL shutdown is performed for a secure connection).

Below is an equivalent code that works with Python <= 3.7::

    stream.close()
    await stream.wait_closed()
```

Lib/asyncio/streams.py
```diff
@@ -331,9 +339,31 @@ def transport(self):
...
+                fut.set_exception(exc)
+                return fut
+        if not self._transport.is_closing():
+            if self._protocol._connection_lost:
+                fut = self._loop.create_future()
+                fut.set_exception(ConnectionResetError('Connection lost'))
+                return fut
+            if not self._protocol._paused:
+                return self._complete_fut
+        return self._loop.create_task(self.drain())
```
1st1, line 396: The idea is to avoid creating task in majority of the cases, e.g.:

In the stream `__init__`:

```python
   self._nop = self.loop.create_future()
   self._nop.set_result(None)
```

and then in `_fast_drain`:

```python
   if not self._paused:
        return self._nop
```

Doc/library/asyncio-stream.rst
asvetlov: Done

Doc/library/asyncio-stream.rst
asvetlov: Done

Lib/asyncio/streams.py
asvetlov: But I do almost the same a few lines above.
There is no `writer._paused` attribute but `writer._protocol._paused` one (which has undefined state if the protocol is closing).
That's why I have slightly more comlex logic that you mentioned in a comment but the functionality is the same.

1st1: Ah, I see, never mind.  Could you please add a comment before `return self._complete_fut` that it's a "fast path"?

asvetlov: Done

asvetlov:
Is the PR good now?

1st1 (approved):

PR #13100, opened 2019-05-05
bpo-34556: Add --upgrade-deps to venv module

cooperlees:
- This allows for pip + setuptools to be automatically upgraded to the latest version on PyPI
- Update documentation to represent this change

[bpo-34556](https://bugs.python.org/issue34556): Add --upgrade to venv module

https://bugs.python.org/issue34556


Doc/library/venv.rst
```diff
@@ -123,12 +123,17 @@ creation according to their needs, the :class:`EnvBuilder` class.
       (defaults to ``None`` which means directory name of the environment would
       be used).
 
+    * ``upgrade_deps`` - Update the base venv modules to the latest on PyPI
```
matrixise: here, you need to use two `-` and not one

Doc/using/venv-create.inc
```diff
@@ -60,10 +60,18 @@ The command, if run with ``-h``, will show the available options::
...
+      --upgrade-deps        Upgrade core dependencies: pip setuptools to the
+                            latest version in PyPI
 
     Once an environment has been created, you may wish to activate it, e.g. by
     sourcing an activate script in its bin directory.
 
+.. versionchanged:: 3.8
+   Add ``--upgrade-deps`` option to upgrade pip + setuptools to the latest on PyPI
+
+.. versionchanged:: 3.6
```
matrixise: this change is out of the scope of this PR, there is another PR (https://github.com/python/cpython/commit/3921b1cc34c2fc8b8b480c19a95ec306de710fdd) 

Maybe you could create another PR for that point.

Lib/venv/__init__.py
```diff
@@ -250,7 +256,7 @@ def setup_python(self, context):
 
             if sysconfig.is_python_build(True):
                 # copy init.tcl
-                for root, dirs, files in os.walk(context.python_dir):
+                for root, _dirs, files in os.walk(context.python_dir):
```
matrixise: why this change?

Misc/NEWS.d/next/Core and Builtins/2019-05-05-18-09-40.bpo-34556.o9kfpu.rst
```diff
@@ -0,0 +1 @@
+Add --upgrade-deps to venv module
```
matrixise: Use the ReST markup like `--upgrade-deps`. Add "Patch by ..."

matrixise (changes requested):
Hi,

Thank you so much for your contribution but I am going to ask you to remove the unnecessary blank lines and formatting. We have to be focused on the main modification, your new feature and not on a new line or a new space in a comment. It's useful in some cases, but not really here. just go to the essential thing but for your new code, you can format it, no problem.

I hope you will understand.

Thank you again for your contrib, I think it could be really useful with `./python -m venv --upgrade-deps`

PS: I have not checked but you could add yourself into the `Misc/ACKS` file

Misc/NEWS.d/next/Core and Builtins/2019-05-05-18-09-40.bpo-34556.o9kfpu.rst
cooperlees: Sure thing.

Lib/venv/__init__.py
cooperlees: dirs is not used in the loop, this is a flake recommended way to indicate that. Will revert so I get my feature landed :)

Doc/using/venv-create.inc
cooperlees: Ahh ok. I just thought it wouldn't hurt to fix things. I will revert.

Doc/library/venv.rst
cooperlees: Thanks.

cooperlees:
No problems, thanks for the review. 
I will revert the style things and apply your suggestions. I would love to come and talk to you about reformatting a lot of this module. Is there a goal that the stdlib is PEP8?

p.s. I am already in ACKs for a previous change. Thanks for asking.

matrixise:
in fact, there is no plan for the stdlib PEP8 compliant excepted for the new code, not for the former and there is a reason for that. if we migrate the stdlib to PEP8, we will have a lot of conflicts for a backport :/

for the new Python code, you can be PEP8 compliant.

Lib/venv/__init__.py
```diff
@@ -363,13 +369,25 @@ def install_scripts(self, context, path):
                         f.write(data)
                     shutil.copymode(srcfile, dstfile)
 
+    def upgrade_dependencies(self, context):
+        logger.debug(
+            'Upgrading {} packages in {}'.format(CORE_VENV_DEPS, context.bin_path)
```
matrixise: You could use the f-string here. it's python 3.8+

matrixise (changes requested):
small change with the f-string. thank you

cooperlees:
Do you all need anything else from me? Would love to see this merge :)

cooperlees:
Rebased my fork and force pushed the update so merge conflicts should go away.

ambv:
Thanks! ✨ 🍰 ✨

wimglenn:
It looks like this didn't land in 3.8.0.  I was hoping the implementation could be improved before being released in 3.8.x: rather than install the bundled pip, and then use that pip installation to uninstall itself and install latest version of pip, wouldn't it be better just to plumb through to ensurepip directly to install the latest in the first place?

Upsides:
- cleaner
- faster
- less convoluted

Downsides:
?

Thoughts?

wimglenn:
FWIW, this is the kind of shell function I had in mind:

```
function ve() {
    if [ ! -d ./.venv ]; then
        echo "creating venv..."
        if ! python3.8 -m venv .venv --prompt=$(basename $PWD) --without-pip; then
            echo "ERROR: Problem creating venv" >&2
            return 1
        else
            source .venv/bin/activate
            python /usr/local/lib/python3.8/ensurepip/_bundled/pip-19.2.3-py2.py3-none-any.whl/pip install --upgrade pip setuptools
        fi
    else
        source .venv/bin/activate
    fi
}
```

Obviously you don't have to hardcode the path to the whl file, you can build it off of `ensurepip._bundled.__path__` if you want (for now, `importlib.resources` doesn't make this easy because there is no `__init__.py` in the `_bundled` subdirectory for 3.8.0)

cooperlees:
Thanks for looking into this. All for the more efficient way to avoid the double install and passing this into ensurepip is definitely the way, but I was not sure if it could handle remote fetches to PyPI if it was needed.

I admit I went the easier route that introduced less change so I had something as performance of the creation was not a high priority for me (and it's optional) and in scripts I was already effectively doing it.

PR #13105, opened 2019-05-06
bpo-36684: travis: split coverage for Python and C

methane:
https://bugs.python.org/issue36684


GPHemsley:
Oh, hmm, I just filed #13146 for the same thing.

PR #13106, opened 2019-05-06
bpo-20092: Make __int__ defaults to __index__

remilapeyre:
https://bugs.python.org/issue33039

### DIFF
--- a/Objects/typeobject.c
+++ b/Objects/typeobject.c
@@ -71,6 +71,8 @@ _Py_IDENTIFIER(__new__);
 _Py_IDENTIFIER(__set_name__);
 _Py_IDENTIFIER(__setitem__);
 _Py_IDENTIFIER(builtins);
+_Py_IDENTIFIER(__index__);
+_Py_IDENTIFIER(__int__);
 
 static PyObject *
 slot_tp_new(PyTypeObject *type, PyObject *args, PyObject *kwds);
@@ -5393,8 +5395,9 @@ PyType_Ready(PyTypeObject *type)
         }
     }
 
-    /* Hack for tp_hash and __hash__.
-       If after all that, tp_hash is still NULL, and __hash__ is not in
+    /* Set default methods */
+
+    /* If after all that, tp_hash is still NULL, and __hash__ is not in
        tp_dict, set tp_hash to PyObject_HashNotImplemented and
        tp_dict['__hash__'] equal to None.
        This signals that __hash__ is not inherited.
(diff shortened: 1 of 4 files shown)


Objects/typeobject.c
```diff
@@ -5410,6 +5413,21 @@ PyType_Ready(PyTypeObject *type)
         }
     }
 
+    /* If __index__ is defined but not __int__, make it default to __index__.
+       Don't touch __float__ and __complex__ as there could be some loss of
+       precision.
+    */
+    PyObject* index = _PyDict_GetItemIdWithError(type->tp_dict, &PyId___index__);
+    if (PyErr_Occurred()) {
```
jdemeyer: You only need to check `PyErr_Occurred()` if `index == NULL`.

jdemeyer (changes requested):
Instead of handling this on the level of the Python `__dict__`, wouldn't it be better to handle this in `nb_int`? That is, set `nb_int` to `nb_index`. That way, it's also guaranteed to work for extension types defining only `nb_index`.

Objects/typeobject.c
remilapeyre: Thanks, I made the change

remilapeyre:
@jdemeyer, I was not absolutly clear with how the methods that have a slot interact with `__dict__`, I tried to read the code source and to look were `__dict__` or the slot were preferred over each other and I'm still not sure about the details. I couldn't find where `__getattr__` would look in the slots.

I will push a commit to update the slots instead of `__dict__` tonight.

jdemeyer:
> I was not absolutly clear with how the methods that have a slot interact with `__dict__`

It's complicated, I would have to look up the details myself.

But basically, the slots are used to put entries in the `__dict__` (instances of `wrapper_descriptor`) and the presence of a special method (but not a `wrapper_descriptor`) in the `__dict__` adds an entry (for example `slot_nb_index`) in the slot. It's especially this two-way interaction that makes things more complicated.

remilapeyre:
Thanks @jdemeyer, I think the last change should be good to make `nb_int` defaults to `nb_index`

jdemeyer:
What I meant is that should *only* do
```
type->tp_as_number->nb_int = type->tp_as_number->nb_index;
```
and then the wrapper descriptor `__int__` should be added automatically.

remilapeyre:
This doesn't seem to work, with this:
```

    /* If __index__ is defined but not __int__, make it default to __index__.
       Don't touch __float__ and __complex__ as there could be some loss of
       precision.
    */
    if (type->tp_as_number != NULL && type->tp_as_number->nb_int == NULL) {
        type->tp_as_number->nb_int = type->tp_as_number->nb_index;
    }
```
I get:
```
✗ ./python.exe -m test.test_index
...E....................................................
ERROR: test_int_defaults_to_index (__main__.BaseTestCase)
  File "/Users/remi/src/cpython/Lib/test/test_index.py", line 97, in test_int_defaults_to_index
    self.assertEqual(int(Test()), 4)
TypeError: int() argument must be a string, a bytes-like object or a number, not 'Test'
```

jdemeyer:
Please post your failing branch, otherwise it's hard to guess what the problem is.

remilapeyre:
@jdemeyer, thanks, I pushed my last commit

remilapeyre:
#13108 has been merged so the issue is now resolved.

PR #13113, opened 2019-05-06
bpo-36813: Fix QueueListener to call task_done() upon termination.

bharel:
Fixed QueueListener in order to avoid random deadlocks.
Unable to add regression tests atm due to time constraints, will add it in a bit.
Regarding implementation, although it's nested, it does not cause performance issues whatsoever, and does not call task_done() in case of an exception (which is the right thing to do IMHO).

https://bugs.python.org/issue36813


bharel:
Alright, regression tests added, all tests pass. Ready for upload!

auvipy (approved):

asvetlov:
thanks!

asvetlov (approved):

PR #13117, opened 2019-05-06
bpo-28866: Invalidate type cache on every setattr

JulienPalard:
https://bugs.python.org/issue28866


JulienPalard:
Marked as do not merge, as there's an alternative I prefer. Will close if we merge the other.

JulienPalard:
Closing, as https://github.com/python/cpython/pull/13157 is merged.

PR #13127, opened 2019-05-06
bpo-34946: mark inspect.getcallargs as deprecated as per documentation

tonybaloney:
https://bugs.python.org/issue34946

### DIFF
--- a/Lib/inspect.py
+++ b/Lib/inspect.py
@@ -1341,6 +1341,9 @@ def getcallargs(*func_and_positional, **named):
     A dict is returned, with keys the function argument names (including the
     names of the * and ** arguments, if any), and values the respective bound
     values from 'positional' and 'named'."""
+    warnings.warn("`getcallargs` is deprecated since Python 3.5. Use"
+                  " `Signature.bind` and `Signature.bind_partial` instead.",
+                  DeprecationWarning, stacklevel=2)
     func = func_and_positional[0]
     positional = func_and_positional[1:]
     spec = getfullargspec(func)
--- a/Lib/inspect.py
+++ b/Lib/inspect.py
@@ -1110,9 +1110,10 @@ def getfullargspec(func):
       - the "self" parameter is always reported, even for bound methods
       - wrapper chains defined by __wrapped__ *not* unwrapped automatically
     """
-
     warnings.warn("Use inspect.signature() instead of inspect.getfullargspec()",
                   DeprecationWarning, stacklevel=2)
+
+def _getfullargspec(func):
     try:
         # Re: `skip_bound_arg=False`
         #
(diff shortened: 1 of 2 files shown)


Lib/inspect.py
```diff
@@ -1341,6 +1341,9 @@ def getcallargs(*func_and_positional, **named):
     A dict is returned, with keys the function argument names (including the
     names of the * and ** arguments, if any), and values the respective bound
     values from 'positional' and 'named'."""
+    warnings.warn("`getcallargs` is deprecated since Python 3.5. Use"
+                  " `Signature.bind` and `Signature.bind_partial` instead.",
+                  DeprecationWarning, stacklevel=2)
     func = func_and_positional[0]
     positional = func_and_positional[1:]
     spec = getfullargspec(func)
```
tirkarthi: By the time I commented in the issue I don't think there were any code level warnings. But `getfullargspec` used in `getcallargs` was recently changed to generate a `DeprecationWarning` from d5d2b4546939b98244708e5bb0cfccd55b99d244. Generating two warnings seems little confusing to me.

cc: @pablogsal 

```
./python.exe -Wall
Python 3.8.0a3+ (heads/master:678bb9d1b5, May  6 2019, 23:38:09)
>>> import inspect
>>> def foo(a, b): pass
...
>>> inspect.getcallargs(foo, 1, 2)
/Users/karthikeyansingaravelan/stuff/python/cpython/Lib/inspect.py:1346: DeprecationWarning: Use inspect.signature() instead of inspect.getfullargspec()
  spec = getfullargspec(func)
{'a': 1, 'b': 2}
```

With PR : 

```
./python.exe -Wall
Python 3.8.0a3+ (heads/master:678bb9d1b5, May  6 2019, 23:38:09)
>>> import inspect
>>> def foo(a, b): pass
...
>>> inspect.getcallargs(foo, 1, 2)
<stdin>:1: DeprecationWarning: `getcallargs` is deprecated since Python 3.5. Use `Signature.bind` and `Signature.bind_partial` instead.
/Users/karthikeyansingaravelan/stuff/python/cpython/Lib/inspect.py:1349: DeprecationWarning: Use inspect.signature() instead of inspect.getfullargspec()
  spec = getfullargspec(func)
{'a': 1, 'b': 2}
```

pablogsal: I concur with @tirkarthi, having two warnings is undesirable and leaks the fact that one function is using another. We could just silence one of the warnings or make `getfullargspec` a thing wrapper over the current implementation (renaming it to `_ getfullargspec`) to avoid the duplicate warning.

tonybaloney:
@tirkarthi @pablogsal resolved in https://github.com/python/cpython/pull/13127/commits/cf6d7c7ff4e5e6f0a4bdacf36942d20238eb502e I also found an extra import for `warn`.

Lib/inspect.py
tirkarthi: Silencing warnings seems like a better idea to me instead of having `getfullargspec` call a wrapper function `_getfullargspec` and using `_getfullargspec` in `getcallargs` . As noted in the issue there was a discussion at https://bugs.python.org/issue27172#msg282174 about some changes around functionality difference between `getcallargs` and `Signature.bind`. 

I would request @ncoghlan to have a review over code level deprecation to see if he has any concerns.

tonybaloney:
@pablogsal I saw some other PRs de-deprecating this function, should this PR be closed?

pablogsal:
@tonybaloney Very likely :( Buy let's wait until we merge the other one in case there is more discussion. Sorry for the inconvenience.

ncoghlan:
Similar rationale applies here as for `getfullargspec` - actually migrating is a pain, and if anyone ends up following through on the idea of making `getfullargspec` a fast low level building block for the higher level Signature API, there could be a comparable lower-level-but-faster trade-off between this and the Signature methods.

tonybaloney:
Looks like one of the core devs changed the code in another PR and merged it anyway. This is now redundant

PR #13128, opened 2019-05-06
bpo-36754: Remove smart quotes in pydoc text

GiantRobato:
Resolves [bpo-36754](https://bugs.python.org/issue36754).

Removed all instances of smart quotes so on windows:

```
help("BASICMETHODS")
```

shows something like this:

```
   ...
   invoking the superclass's "__new__()" method using
   ...
```

instead of this:

```
   ...
   invoking the superclassÆs "__new__()" method using
   ...
```

https://bugs.python.org/issue36754


auvipy (changes requested):
good work please re base

csabella:
@penguindustin, thank you for your work on this issue.  As Steve said on the bug tracker, this file is autogenerated, so the text shouldn't be edited directly.  The generator function or the help() rendering is what needs to change.  I'm going to close this for now, so if you choose to revisit the issue, please open a new PR.  Thanks!

PR #13129, opened 2019-05-06
bpo-36818: Add PyInterpreterState.runtime field.

ericsnowcurrently:
https://bugs.python.org/issue36818

### DIFF
--- a/Include/internal/pycore_object.h
+++ b/Include/internal/pycore_object.h
@@ -19,9 +19,10 @@ PyAPI_FUNC(int) _PyDict_CheckConsistency(PyObject *mp, int check_content);
  * NB: While the object is tracked by the collector, it must be safe to call the
  * ob_traverse method.
  *
- * Internal note: _PyRuntime.gc.generation0->_gc_prev doesn't have any bit flags
- * because it's not object header.  So we don't use _PyGCHead_PREV() and
- * _PyGCHead_SET_PREV() for it to avoid unnecessary bitwise operations.
+ * Internal note: _PyRuntimeState.gc.generation0->_gc_prev doesn't have
+ * any bit flags because it's not object header.  So we don't use
+ * _PyGCHead_PREV() and _PyGCHead_SET_PREV() for it to avoid unnecessary
+ * bitwise operations.
  *
  * The PyObject_GC_Track() function is the public version of this macro.
  */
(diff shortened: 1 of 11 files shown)


Include/internal/pycore_object.h
```diff
@@ -38,11 +38,12 @@ static inline void _PyObject_GC_TRACK_impl(const char *filename, int lineno,
                           "object is in generation which is garbage collected",
                           filename, lineno, "_PyObject_GC_TRACK");
 
-    PyGC_Head *last = (PyGC_Head*)(_PyRuntime.gc.generation0->_gc_prev);
+    _PyRuntimeState *runtime = _PyRuntimeState_Get();
+    PyGC_Head *last = (PyGC_Head*)(runtime->gc.generation0->_gc_prev);
```
vstinner: IMHO it's wrong to put the gc structure into the runtime: it should be moved to PyInterpreterState instead, and so this function should get the current interpreter, not the runtime.

Include/internal/pycore_pystate.h
```diff
@@ -131,11 +135,13 @@ struct _gilstate_runtime_state {
 };
 
 /* hook for PyEval_GetFrame(), requested for Psyco */
-#define _PyThreadState_GetFrame _PyRuntime.gilstate.getframe
+#define _PyThreadState_GetFrame \
+    _PyRuntimeState_Get()->gilstate.getframe
```
vstinner: IMHO gilstate should be moved into PyInterpreterState.

vstinner (changes requested):
I disagree with the overall approach. Maybe I misunderstood you.
https://bugs.python.org/issue36818#msg342043

Include/internal/pycore_object.h
ericsnowcurrently: Agreed. :)  See gh-13219.

Include/internal/pycore_pystate.h
ericsnowcurrently: Also agreed.

vstinner:
Eric is working on a different approach: https://bugs.python.org/issue36854

ericsnowcurrently:
The GC stuff is orthogonal to adding `PyInterpreterState.runtime`.  I still plan on merging this after resolving merge conflicts and any feedback.

ericsnowcurrently:
FYI, I've trimmed the PR down.  The changed I've removed can wait until later.

PR #13132, opened 2019-05-06
bpo-26903: Limit ProcessPoolExecutor to 61 workers on Windows

brianquinlan:
https://bugs.python.org/issue26903

### DIFF
--- a/Lib/concurrent/futures/process.py
+++ b/Lib/concurrent/futures/process.py
@@ -57,6 +57,7 @@ import threading
 import weakref
 from functools import partial
 import itertools
+import sys
 import traceback
 
 # Workers are created as daemon threads and processes. This is done to allow the
@@ -108,6 +109,12 @@ def _python_exit():
 # (Futures in the call queue cannot be cancelled).
 EXTRA_QUEUED_CALLS = 1
 
+
+# On Windows, WaitForMultipleObjects is used to wait for processes to finish.
+# It can wait on, at most, 63 objects. There is an overhead of two objects:
+# - the result queue reader
+# - the thread wakeup reader
+_MAX_WINDOWS_WORKERS = 63 - 2
 
 # Hack to embed stringification of remote traceback in local traceback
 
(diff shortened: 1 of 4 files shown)


Lib/test/test_concurrent_futures.py
```diff
@@ -755,6 +755,13 @@ def test_default_workers(self):
 
 
 class ProcessPoolExecutorTest(ExecutorTest):
+
+    @unittest .skipUnless(sys.platform=='win32', 'Windows-only process limit')
```
zooba:
```suggestion
    @unittest.skipUnless(sys.platform=='win32', 'Windows-only process limit')
```

zooba (approved):

Lib/concurrent/futures/process.py
```diff
@@ -505,9 +512,16 @@ def __init__(self, max_workers=None, mp_context=None,
...
             self._max_workers = os.cpu_count() or 1
+            if sys.platform == 'win32':
+                self._max_workers = min(_MAX_WINDOWS_WORKERS,
+                                        self._max_workers)
         else:
             if max_workers <= 0:
                 raise ValueError("max_workers must be greater than 0")
+            elif (sys.platform == 'win32' and
+                max_workers > _MAX_WINDOWS_WORKERS):
+                raise ValueError(
```
JelleZijlstra, line 523: Should we consider automatically changing max_workers to 61 instead? That would make for a nicer API for people using `max_workers=os.cpu_count()`, and it would make code like what was in Black before https://github.com/python/black/pull/838/files work unchanged.

On the other hand, it may be unexpected that we silently use fewer workers than the user asked for.

brianquinlan: I thought silently using fewer processes could be surprising. I also think that it's pretty unusual in the standard library to silently fix bad argument values.

zooba: I like raising when too many workers are requested, and doing it quietly when it was "automatic".

zooba:
Thanks, Brian!

PR #13135, opened 2019-05-06
bpo-36820: Break unnecessary cycle in socket.py, codeop.py and dyld.py

mariocj89:
Break cycle generated when saving an exception in socket.py, codeop.py and dyld.py as they keep alive not only the exception but user objects through the ``__traceback__`` attribute.

https://bugs.python.org/issue36820

### DIFF
--- a/Lib/codeop.py
+++ b/Lib/codeop.py
@@ -93,10 +93,13 @@ def _maybe_compile(compiler, source, filename, symbol):
     except SyntaxError as e:
         err2 = e
 
-    if code:
-        return code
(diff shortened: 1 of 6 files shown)


auvipy (approved):

mariocj89:
From offline conversation @pablogsal, this was waiting on trying to get a generic way to detect these kinds of cycles. It seemed to be much harder than initially thought.
As suggested, do you want me to remove the tests here and land this without tests?

pablogsal (approved):
LGTM 🚀

PR #13137, opened 2019-05-06
bpo-24939: move stringlib/unicode_format.h to unicodeobject.c

tonybaloney:
https://bugs.python.org/issue24939


tonybaloney:
@ericsnowcurrently

tonybaloney:
Suggest no-news-required as well

PR #13143, opened 2019-05-06
bpo-34975: Add start_tls() method to streams API

icgood:
The existing event loop `start_tls()` method is not sufficient for
connections using the streams API. The existing StreamReader works
because the new transport passes received data to the original protocol.
The StreamWriter must then write data to the new transport, and the
StreamReaderProtocol must be updated to close the new transport
correctly.

The new StreamWriter `start_tls()` updates itself and the reader
protocol to the new SSL transport.

https://bugs.python.org/issue34975

### DIFF
--- a/Doc/library/asyncio-stream.rst
+++ b/Doc/library/asyncio-stream.rst
@@ -325,6 +325,24 @@ StreamWriter
       be resumed.  When there is nothing to wait for, the :meth:`drain`
       returns immediately.
 
+   .. coroutinemethod:: start_tls(sslcontext, \*, server_hostname=None, \
+                          ssl_handshake_timeout=None)
+
+      Upgrade an existing stream-based connection to TLS.
+
+      Parameters:
+
+      * *sslcontext*: a configured instance of :class:`~ssl.SSLContext`.
+
+      * *server_hostname*: sets or overrides the host name that the target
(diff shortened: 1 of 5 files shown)


asvetlov:
Fixed by #13251

arhadthedev:
@asvetlov Can this PR be reopened? #13251 got refused and was reverted by #16482 with the following explainer in bpo-38242:

> The asyncio.Stream class we have now is overloaded with functionality. It's not composable. It's internal buffer and APIs are designed to parsing text protocols (i.e. parsing a complex binary protocol requires an entirely different buffer implementation).

I'm bringing this PR up because I need to port `test.test_poplib` and `test.test_ftplib` from to-be-removed `asynchat` and they use on-demand switching from plain text to SSL and back.

@icgood Could you also add `stop_tls()` to match `start_tls()`, please?

icgood:
> @icgood Could you also add `stop_tls()` to match `start_tls()`, please?

I'd be happy to, if we get the go-ahead to reopen and proceed. I've been using undocumented API in my library for years now, in lieu of a fix in CPython: https://github.com/icgood/pymap/blob/73c52f805d5fabf91a28f89ff81dd34efcb36486/pymap/imap/__init__.py#L275

PR #13144, opened 2019-05-06
bpo-36814: ensure os.posix_spawnp handles Py_None

tonybaloney:
https://bugs.python.org/issue36814

### DIFF
--- a/Lib/test/test_posix.py
+++ b/Lib/test/test_posix.py
@@ -1828,22 +1828,23 @@ class TestPosixSpawn(unittest.TestCase, _PosixSpawnMixin):
 class TestPosixSpawnP(unittest.TestCase, _PosixSpawnMixin):
     spawn_func = getattr(posix, 'posix_spawnp', None)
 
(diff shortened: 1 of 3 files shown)


tonybaloney:
Requesting review from @nanjekyejoannah

ZackerySpytz:
I believe this change should have a test.

tonybaloney:
@ZackerySpytz agreed, added tests

nanjekyejoannah:
Looks fine for me. @vstinner any thoughts?

vstinner (changes requested):
You should not add tests to TestPosixSpawnP but _PosixSpawnMixin, otherwise only os.posix_spawnp() is tested, not os.posix_spawn().

Well, honestly, I would go with a PR without tests. We already have plenty of tests for posix_spawn().

If you want to add tests, please avoid setUp() and avoid hardcoding "posix_spawnp" in the test, since the tests should be run on posix_spawn and posix_spawnp.

tonybaloney:

PR #13146, opened 2019-05-07
bpo-36684: Split out gcc and test coverage builds for better results

GPHemsley:
The optional test coverage builds shouldn't be relied upon to certify
that gcc builds don't fail.

The combined Python and C coverage test runs now exceed Travis's
50-minute time limit. Splitting them into separate runs gives more
leeway.

Also, adding branch coverage to Python testing and ensure that
coverage is reported even if tests fail. (The primary builds are
for tracking test failures.)

https://bugs.python.org/issue36684

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -35,14 +35,9 @@ matrix:
     - os: linux
       language: c
       compiler: clang
-      env: TESTING=cpython
-      addons:
-        apt:
-          packages:
-            - xvfb
-    - os: linux
-      language: c
-      compiler: gcc
+      # gcc also works, but to keep the # of concurrent builds down, we use one C
+      # compiler here and the other to run the coverage build. Clang is preferred
+      # in this instance for its better error messages.
       env: TESTING=cpython
       addons:
         apt:
--- a/.travis.yml
+++ b/.travis.yml
@@ -35,9 +35,14 @@ matrix:
     - os: linux
       language: c
       compiler: clang
-      # gcc also works, but to keep the # of concurrent builds down, we use one C
-      # compiler here and the other to run the coverage build. Clang is preferred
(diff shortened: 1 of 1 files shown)


GPHemsley:
@ammaraskar @methane @brettcannon @zware This might interest you.

GPHemsley:
For historical reference: I did a bunch of testing in #13084.

.travis.yml
```diff
@@ -35,9 +35,14 @@ matrix:
...
-      # compiler here and the other to run the coverage build. Clang is preferred
-      # in this instance for its better error messages.
+      env: TESTING=cpython
+      addons:
+        apt:
+          packages:
+            - xvfb
+    - os: linux
+      language: c
+      compiler: gcc
```
methane: Why you added this?

GPHemsley: As I mentioned:
> The optional test coverage builds shouldn't be relied upon to certify
that gcc builds don't fail.

It's possible I'm missing something, but adding additional builds is cheap; there's not really a reason _not_ to have them.

methane: > The optional test coverage builds shouldn't be relied upon to certify
that gcc builds don't fail.

Note that not including gcc in travis job doesn't mean we don't test gcc.
For example, we use gcc in Ubuntu on Azure Pipelines.  (see [this build](https://dev.azure.com/Python/cpython/_build/results?buildId=41875))
And many buildbot works after merge.  gcc is tested well.


> It's possible I'm missing something, but adding additional builds is cheap; there's not really a reason _not_ to have them.

Additional build is not cheap always.   We experienced long queue waiting many times.

At least, it is not relating to bpo-36684.  Please remove it.

GPHemsley: Done.

PR #13148, opened 2019-05-07
bpo-34616: add flags to allow top-level-await 

Carreau:
Based on a work with @1st1 during PyCon 2019; and can be seen used in https://github.com/ipython/ipython/pull/11713 (which would dramatically reduce thee complexity and shave an extra few hundreds lines of IPython if we were supporting 3.8+ only).

https://bugs.python.org/issue34616

### DIFF
--- a/Doc/library/functions.rst
+++ b/Doc/library/functions.rst
@@ -257,6 +257,12 @@ are always available.  They are listed here in alphabetical order.
    can be found as the :attr:`~__future__._Feature.compiler_flag` attribute on
    the :class:`~__future__._Feature` instance in the :mod:`__future__` module.
 
+   The optional argument *flags* also control wether the compiled source is
+   allowed to contain top-level ``await``, ``async for`` and ``async with``.
+   When the bit ``ast.PyCF_ALLOW_TOP_LEVEL_AWAIT`` is set, the the return code
+   object has ``CO_COROUTINE`` set in ``co_code``, and can be interacitvely
+   executed via ``yield from eval(code_object)``.
+
    The argument *optimize* specifies the optimization level of the compiler; the
    default value of ``-1`` selects the optimization level of the interpreter as
    given by :option:`-O` options.  Explicit levels are ``0`` (no optimization;
@@ -289,6 +295,10 @@ are always available.  They are listed here in alphabetical order.
    .. versionchanged:: 3.5
       Previously, :exc:`TypeError` was raised when null bytes were encountered
       in *source*.
+
+   .. versionadded:: 3.8
+      *flags* can now be used to accept source that would contain a top-level
+      ``await``, ``async for`` or ``async with``.
 
 
 .. class:: complex([real[, imag]])
--- a/Doc/library/functions.rst
+++ b/Doc/library/functions.rst
@@ -257,10 +257,10 @@ are always available.  They are listed here in alphabetical order.
    can be found as the :attr:`~__future__._Feature.compiler_flag` attribute on
    the :class:`~__future__._Feature` instance in the :mod:`__future__` module.
 
-   The optional argument *flags* also control wether the compiled source is
+   The optional argument *flags* also controls whether the compiled source is
    allowed to contain top-level ``await``, ``async for`` and ``async with``.
    When the bit ``ast.PyCF_ALLOW_TOP_LEVEL_AWAIT`` is set, the the return code
-   object has ``CO_COROUTINE`` set in ``co_code``, and can be interacitvely
+   object has ``CO_COROUTINE`` set in ``co_code``, and can be interactively
    executed via ``yield from eval(code_object)``.
 
    The argument *optimize* specifies the optimization level of the compiler; the
--- a/Doc/library/functions.rst
+++ b/Doc/library/functions.rst
@@ -297,8 +297,8 @@ are always available.  They are listed here in alphabetical order.
       in *source*.
 
    .. versionadded:: 3.8
-      *flags* can now be used to accept source that would contain a top-level
-      ``await``, ``async for`` or ``async with``.
+      ``ast.PyCF_ALLOW_TOP_LEVEL_AWAIT`` can now be passed in flags to enable
+      support for top-level await, async for, and async with.
 
 
 .. class:: complex([real[, imag]])
--- a/Doc/library/functions.rst
+++ b/Doc/library/functions.rst
@@ -259,7 +259,7 @@ are always available.  They are listed here in alphabetical order.
 
    The optional argument *flags* also controls whether the compiled source is
    allowed to contain top-level ``await``, ``async for`` and ``async with``.
-   When the bit ``ast.PyCF_ALLOW_TOP_LEVEL_AWAIT`` is set, the the return code
+   When the bit ``ast.PyCF_ALLOW_TOP_LEVEL_AWAIT`` is set, the return code
    object has ``CO_COROUTINE`` set in ``co_code``, and can be interactively
    executed via ``await eval(code_object)``.
 
@@ -298,7 +298,7 @@ are always available.  They are listed here in alphabetical order.
 
    .. versionadded:: 3.8
       ``ast.PyCF_ALLOW_TOP_LEVEL_AWAIT`` can now be passed in flags to enable
-      support for top-level await, async for, and async with.
+      support for top-level ``await``, ``async for``, and ``async with``.
 
 
 .. class:: complex([real[, imag]])
--- a/Doc/library/functions.rst
+++ b/Doc/library/functions.rst
@@ -261,7 +261,7 @@ are always available.  They are listed here in alphabetical order.
    allowed to contain top-level ``await``, ``async for`` and ``async with``.
    When the bit ``ast.PyCF_ALLOW_TOP_LEVEL_AWAIT`` is set, the the return code
    object has ``CO_COROUTINE`` set in ``co_code``, and can be interactively
-   executed via ``yield from eval(code_object)``.
+   executed via ``await eval(code_object)``.
 
    The argument *optimize* specifies the optimization level of the compiler; the
    default value of ``-1`` selects the optimization level of the interpreter as
--- a/Lib/test/test_builtin.py
+++ b/Lib/test/test_builtin.py
@@ -19,6 +19,7 @@ import unittest
 import warnings
 from contextlib import ExitStack
 from inspect import CO_COROUTINE
+from itertools import product
 from textwrap import dedent
 from types import AsyncGeneratorType
 from operator import neg
@@ -362,20 +363,27 @@ class BuiltinTest(unittest.TestCase):
                 self.assertEqual(rv, tuple(expected))
 
     def test_compile_top_level_await(self):
-        for mode in ('single', 'exec'):
-            for num,code_sample in enumerate(['''await sleep(0)''',
-                                '''async for i in range(10):
-                                       print(i)''',
-                                '''async with asyncio.Lock() as l:
-                                       pass''']):
-                source = dedent(code_sample)
-                with self.assertRaises(SyntaxError,
-                                  msg='source={!r} mode={!r})'.format(source, mode)):
-                    compile(source, '?' , mode)
+        """Test whether code some top level await can be compiled.
 
-                co = compile(source, '?', mode, flags=ast.PyCF_ALLOW_TOP_LEVEL_AWAIT)
-                self.assertEqual(co.co_flags & CO_COROUTINE, CO_COROUTINE,
-                                 msg='source={!r} mode={!r})'.format(source, mode))
+        Make sure it compiles only with the PyCF_ALLOW_TOP_LEVEL_AWAIT flag set,
+        and make sure the generated code object has the CO_COROUTINE flag set in
+        order to execute it with `yield from eval(.....)` instead of exec.
+        """
+        modes = ('single', 'exec')
+        code_samples = ['''await sleep(0)''',
+        '''async for i in range(10):
+               print(i)''',
+        '''async with asyncio.Lock() as l:
+               pass''']
+        for mode, code_sample in product(modes,code_samples):
+            source = dedent(code_sample)
+            with self.assertRaises(SyntaxError,
+                                   msg='source={!r} mode={!r})'.format(source, mode)):
+                compile(source, '?' , mode)
+
+            co = compile(source, '?', mode, flags=ast.PyCF_ALLOW_TOP_LEVEL_AWAIT)
+            self.assertEqual(co.co_flags & CO_COROUTINE, CO_COROUTINE,
+                             msg='source={!r} mode={!r})'.format(source, mode))
 
     def test_compile_async_generator(self):
         co = compile(dedent("""async def ticker():
--- a/Lib/test/test_builtin.py
+++ b/Lib/test/test_builtin.py
@@ -386,6 +386,10 @@ class BuiltinTest(unittest.TestCase):
                              msg='source={!r} mode={!r})'.format(source, mode))
 
     def test_compile_async_generator(self):
+        """
+        With the PyCF_ALLOW_TOP_LEVEL_AWAIT flag added in 3.8, we want to
+        make sure AsyncGenerators are still properly not marked with CO_COROUTINE
+        """
         co = compile(dedent("""async def ticker():
                 for i in range(10):
                     yield i
--- a/Lib/test/test_builtin.py
+++ b/Lib/test/test_builtin.py
@@ -18,6 +18,8 @@ import types
 import unittest
 import warnings
 from contextlib import ExitStack
+from inspect import CO_COROUTINE
+from textwrap import dedent
 from operator import neg
 from test.support import (
     EnvironmentVarGuard, TESTFN, check_warnings, swap_attr, unlink)
@@ -357,6 +359,33 @@ class BuiltinTest(unittest.TestCase):
                 exec(code, ns)
                 rv = ns['f']()
                 self.assertEqual(rv, tuple(expected))
+
+    def test_compile_top_level_await(self):
+        for mode in ('single', 'exec'):
+            for num,code_sample in enumerate(['''await sleep(0)''',
+                                '''async for i in range(10):
+                                       print(i)''',
+                                '''async with asyncio.Lock() as l:
+                                       pass''']):
+                source = dedent(code_sample)
+                with self.assertRaises(SyntaxError,
+                                  msg='source={!r} mode={!r})'.format(source, mode)):
+                    compile(source, '?' , mode)
+                co = compile(source, '?', mode, flags=ast.PyCF_ALLOW_TOP_LEVEL_AWAIT)
+                self.assertEqual(co.co_flags & CO_COROUTINE, CO_COROUTINE)
+
+    def test_compile_async_generator(self):
+        co = compile(dedent("""async def ticker():
+                for i in range(10):
+                    yield i
+                    await asyncio.sleep(0)"""), '?', 'exec', flags=ast.PyCF_ALLOW_TOP_LEVEL_AWAIT)
+        glob = {}
+        exec(co, glob)
+
+        from types import AsyncGeneratorType
+
+        self.assertEqual(type(glob['ticker']()), AsyncGeneratorType)
+
 
     def test_delattr(self):
         sys.spam = 1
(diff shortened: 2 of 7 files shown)


Carreau:
If You wish to try it, here is a fake-repl session script that work as you would expect; 

```
import asyncio
from ast import PyCF_ALLOW_TOP_LEVEL_AWAIT
from inspect import CO_COROUTINE
from textwrap import dedent

is_async = (lambda co: co.co_flags & CO_COROUTINE == CO_COROUTINE)
g = {}
l = {}

@asyncio.coroutine
def f():
    for inp in ['import asyncio',
                'print("before")',
                dedent('''
                print("multi statement in eval...")
                res = await asyncio.sleep(0, result="slept")
                print(res)
                print(res[::-1])
                '''),
                'print("finished")']:
        co = compile(inp, '', 'exec', flags=PyCF_ALLOW_TOP_LEVEL_AWAIT)
        if is_async(co):
            yield from eval(co, l, g)
        else:
            exec(co, l,g)

import asyncio
asyncio.run(f())
```

The weird-ish think is the need to use `yield from eval(...)` when running async-code.. but close enough; 

Works as well with trio (cc @njsmith) if you use 
```
def _trio_runner(async_fn):
    import trio

    async def loc(coro):
        """
        We need the dummy no-op async def to protect from
        trio's internal. See https://github.com/python-trio/trio/issues/89
        """
        return await coro

    return trio.run(loc, async_fn)

_trio_runner(f())
```

njsmith:
CC @oremanj

asyncex
```diff
@@ -0,0 +1,20 @@
+import asyncio
```
1st1: This file shouldn't be part of the PR.  We'll need to add a couple of unit tests though.

trycomp
```diff
@@ -0,0 +1,16 @@
+#!/bin/bash
```
1st1: This file shouldn't be part of the PR.

Python/compile.c
```diff
@@ -4759,12 +4765,20 @@ compiler_visit_expr1(struct compiler *c, expr_ty e)
...
+                return compiler_error(c, "'await' outside function");
+            }
 
-        if (c->u->u_scope_type != COMPILER_SCOPE_ASYNC_FUNCTION &&
-                c->u->u_scope_type != COMPILER_SCOPE_COMPREHENSION)
-            return compiler_error(c, "'await' outside async function");
+            c->u->u_ste->ste_coroutine = 1;
+        }
+
+        if (!(c->c_flags->cf_flags & PyCF_ALLOW_TOP_LEVEL_AWAIT)) {
```
1st1: This check should be the top-level if; then nest the other two checks inside.

Python/compile.c
```diff
@@ -4759,12 +4765,20 @@ compiler_visit_expr1(struct compiler *c, expr_ty e)
...
-            return compiler_error(c, "'await' outside function");
+        if (c->u->u_ste->ste_type != FunctionBlock) {
+            if (!(c->c_flags->cf_flags & PyCF_ALLOW_TOP_LEVEL_AWAIT)) {
+                return compiler_error(c, "'await' outside function");
+            }
 
-        if (c->u->u_scope_type != COMPILER_SCOPE_ASYNC_FUNCTION &&
-                c->u->u_scope_type != COMPILER_SCOPE_COMPREHENSION)
-            return compiler_error(c, "'await' outside async function");
+            c->u->u_ste->ste_coroutine = 1;
```
1st1: I'm not sure if this is actually necessary, try to remove this line and see if the tests pass (and add tests :)).

1st1 (changes requested):

Python/compile.c
```diff
@@ -4759,12 +4763,17 @@ compiler_visit_expr1(struct compiler *c, expr_ty e)
...
 
-        if (c->u->u_scope_type != COMPILER_SCOPE_ASYNC_FUNCTION &&
-                c->u->u_scope_type != COMPILER_SCOPE_COMPREHENSION)
-            return compiler_error(c, "'await' outside async function");
+            if (c->u->u_scope_type != COMPILER_SCOPE_ASYNC_FUNCTION && 
+                    c->u->u_scope_type != COMPILER_SCOPE_COMPREHENSION){
+                return compiler_error(c, "'await' outside async function");
+            }
+        }
+        c->u->u_ste->ste_coroutine = 1;
```
1st1: I'm still not sure if this line is needed, but if it is, it should only be set when `(c->c_flags->cf_flags & PyCF_ALLOW_TOP_LEVEL_AWAIT)` is true.

Python/compile.c
```diff
@@ -5699,6 +5708,10 @@ compute_code_flags(struct compiler *c)
     /* (Only) inherit compilerflags in PyCF_MASK */
     flags |= (c->c_flags->cf_flags & PyCF_MASK);
 
+    if ((c->c_flags->cf_flags & PyCF_ALLOW_TOP_LEVEL_AWAIT) && c->u->u_ste->ste_coroutine) {
+        flags |= CO_COROUTINE;
```
Carreau, line 5726: This thing has some weird effect and turn async_generators into coroutine;  I'm going to change that to `!ste->ste_generator`

Python/compile.c
Carreau: ... well it actually is or I get 
```
unhandled exception during asyncio.run() shutdown
task: <Task finished name='Task-2' coro=<<async_generator_athrow without __name__>()> exception=RuntimeError("can't send non-None value to a just-started coroutine")>
RuntimeError: can't send non-None value to a just-started coroutine
```

Not quite sure I understand exactly why...

Carreau:
You were right that the ste->ste_coroutine was not necessary in AwaitKind, though it is necessary in `async_with` and `async_for`. I've removed some of the unnecessary files, and shuffled the `if`s to be a bit more readable.


Here is a short asyncio-repl with some example inputs: 
https://gist.github.com/Carreau/9d8626ab0cf356b63d3481f28a394562

Carreau:

Carreau:
I've pushed again to restart the failing test on travis; all seem to work. Let me know what else I can do; I'd love to get that in 3.8.

Doc/library/functions.rst
```diff
@@ -257,6 +257,12 @@ are always available.  They are listed here in alphabetical order.
    can be found as the :attr:`~__future__._Feature.compiler_flag` attribute on
    the :class:`~__future__._Feature` instance in the :mod:`__future__` module.
 
+   The optional argument *flags* also control wether the compiled source is
+   allowed to contain top-level ``await``, ``async for`` and ``async with``.
+   When the bit ``ast.PyCF_ALLOW_TOP_LEVEL_AWAIT`` is set, the the return code
+   object has ``CO_COROUTINE`` set in ``co_code``, and can be interacitvely
```
tirkarthi:
```suggestion
   object has ``CO_COROUTINE`` set in ``co_code``, and can be interactively
```

Doc/library/functions.rst
```diff
@@ -257,6 +257,12 @@ are always available.  They are listed here in alphabetical order.
    can be found as the :attr:`~__future__._Feature.compiler_flag` attribute on
    the :class:`~__future__._Feature` instance in the :mod:`__future__` module.
 
+   The optional argument *flags* also control wether the compiled source is
```
tirkarthi:
```suggestion
   The optional argument *flags* also controls whether the compiled source is
```

Lib/test/test_builtin.py
```diff
@@ -358,6 +361,32 @@ def f(): """doc"""
                 rv = ns['f']()
                 self.assertEqual(rv, tuple(expected))
 
+    def test_compile_top_level_await(self):
+        for mode in ('single', 'exec'):
+            for num,code_sample in enumerate(['''await sleep(0)''',
```
tirkarthi: Minor nit : num is not used in test so `enumerate` seems to be not needed . Perhaps the list used in enumerate can be defined as a variable outside of for loop and just iterated?

Lib/test/test_builtin.py
```diff
@@ -358,6 +361,32 @@ def f(): """doc"""
...
+                                '''async for i in range(10):
+                                       print(i)''',
+                                '''async with asyncio.Lock() as l:
+                                       pass''']):
+                source = dedent(code_sample)
+                with self.assertRaises(SyntaxError,
+                                  msg='source={!r} mode={!r})'.format(source, mode)):
```
tirkarthi: Align msg with `SyntaxError` ? PEP 8 enforces 80 character limit per line but I will leave it to reviewer on the same.

Lib/test/test_builtin.py
Carreau: Good point, I used it to debug which sample(s) were wrong and did not removed it, will do.

Lib/test/test_builtin.py
Carreau: I've done that and used `itertools.product` to remove one level of indent. I'm still slightly over.

tirkarthi: Perhaps have `msg` as a variable outside of context manager.

```python
msg = 'source={!r} mode={!r})'.format(source, mode)

with self.assertRaises(SyntaxError, msg):
    compile(source, '?' , mode)
```

or even better master has f-string debugging merged into master where `f"{source=}"` expands into `f"source={source!r}"`. So you can rebase onto master and write below : 

```python
with self.assertRaises(SyntaxError, msg=f'{source=} {mode=})'): # 75 characters
    compile(source, '?' , mode)
```

Carreau: Yep good idea. Done that and rebased on master.

Doc/library/functions.rst
```diff
@@ -257,6 +257,12 @@ are always available.  They are listed here in alphabetical order.
    can be found as the :attr:`~__future__._Feature.compiler_flag` attribute on
    the :class:`~__future__._Feature` instance in the :mod:`__future__` module.
 
+   The optional argument *flags* also controls whether the compiled source is
+   allowed to contain top-level ``await``, ``async for`` and ``async with``.
+   When the bit ``ast.PyCF_ALLOW_TOP_LEVEL_AWAIT`` is set, the the return code
+   object has ``CO_COROUTINE`` set in ``co_code``, and can be interactively
+   executed via ``yield from eval(code_object)``.
```
1st1:
```suggestion
   executed via ``await eval(code_object)``.
```

Doc/library/functions.rst
```diff
@@ -290,6 +296,10 @@ are always available.  They are listed here in alphabetical order.
       Previously, :exc:`TypeError` was raised when null bytes were encountered
       in *source*.
 
+   .. versionadded:: 3.8
+      *flags* can now be used to accept source that would contain a top-level
```
1st1: ``ast.PyCF_ALLOW_TOP_LEVEL_AWAIT`` can now be passed in *flags* to enable support
for top-level await, async for, and async with.

Lib/test/test_builtin.py
```diff
@@ -358,6 +362,43 @@ def f(): """doc"""
...
+
+    def test_compile_async_generator(self):
+        """
+        With the PyCF_ALLOW_TOP_LEVEL_AWAIT flag added in 3.8, we want to
+        make sure AsyncGenerators are still properly not marked with CO_COROUTINE
+        """
+        co = compile(dedent("""async def ticker():
+                for i in range(10):
+                    yield i
+                    await asyncio.sleep(0)"""), '?', 'exec', flags=ast.PyCF_ALLOW_TOP_LEVEL_AWAIT)
```
1st1: Please make sure all new lines fit the 79 characters limit.

Python/compile.c
1st1: Please fix to fit 79 characters

Python/compile.c
```diff
@@ -4564,7 +4566,9 @@ compiler_async_with(struct compiler *c, stmt_ty s, int pos)
     withitem_ty item = asdl_seq_GET(s->v.AsyncWith.items, pos);
 
     assert(s->kind == AsyncWith_kind);
-    if (c->u->u_scope_type != COMPILER_SCOPE_ASYNC_FUNCTION) {
+    if (c->c_flags->cf_flags & PyCF_ALLOW_TOP_LEVEL_AWAIT){
+        c->u->u_ste->ste_coroutine = 1;
```
1st1, line 4570: Have you checked if this `c->u->u_ste->ste_coroutine = 1;` line is needed?  Does it work without it?

Lib/test/test_builtin.py
```diff
@@ -358,6 +362,43 @@ def f(): """doc"""
...
+               pass''']
+        for mode, code_sample in product(modes,code_samples):
+            source = dedent(code_sample)
+            with self.assertRaises(SyntaxError, msg=f"{source=} {mode=}"):
+                compile(source, '?' , mode)
+
+            co = compile(source, '?', mode, flags=ast.PyCF_ALLOW_TOP_LEVEL_AWAIT)
+
+            self.assertEqual(co.co_flags & CO_COROUTINE, CO_COROUTINE,
+                             msg=f"{source=} {mode=}")
```
1st1: Please add a functional test, that wraps the created code object in a FunctionType and then manually iterates the coroutine (or with asyncio) and checks the result.

Python/compile.c
Carreau: yes it is necessary, otherwise: 

```
>>> async with acm() as delta:
>>>     count += delta

  File "minirepl.py", line 97, in f
    co = compile(inp, '', 'single', flags=PyCF_ALLOW_TOP_LEVEL_AWAIT)
  File "<string>", line 2
SyntaxError: 'async with' outside async function
```

1st1: OK, cool, let's keep it then :)

Carreau: And there is be a test in `Lib/test/test_builtin.py` for that  in case you ask.

Python/compile.c
Carreau: Done.

Lib/test/test_builtin.py
```diff
@@ -358,6 +364,72 @@ def f(): """doc"""
...
+                globals_ = {'asyncio': asyncio, 'a':0, 'arange': arange}
+                async_f = FunctionType(co, globals_)
+                asyncio.run(async_f())
+                self.assertEqual(globals_['a'], 1)
+
+                # test we can await-eval,
+                globals_ = {'asyncio': asyncio, 'a':0, 'arange': arange}
+                asyncio.run(eval(co, globals_))
+                self.assertEqual(globals_['a'], 1)
+        except Exception:
```
Carreau: I rewrote most of this, to also run via a `FunctionType` – though I believe creating FunctionTypes directly is not supposed to be done by usercode. and via `run(eval())`

Doc/library/functions.rst
Carreau: Done.

Lib/test/test_builtin.py
Carreau: Done.

Lib/test/test_builtin.py
Carreau: I've added 2 with `FunctionTypes` and `eval()`

Doc/library/functions.rst
```diff
@@ -257,6 +257,12 @@ are always available.  They are listed here in alphabetical order.
    can be found as the :attr:`~__future__._Feature.compiler_flag` attribute on
    the :class:`~__future__._Feature` instance in the :mod:`__future__` module.
 
+   The optional argument *flags* also controls whether the compiled source is
+   allowed to contain top-level ``await``, ``async for`` and ``async with``.
+   When the bit ``ast.PyCF_ALLOW_TOP_LEVEL_AWAIT`` is set, the the return code
```
Carreau: Note to self. Double "the" here.

Doc/library/functions.rst
```diff
@@ -290,6 +296,10 @@ are always available.  They are listed here in alphabetical order.
       Previously, :exc:`TypeError` was raised when null bytes were encountered
       in *source*.
 
+   .. versionadded:: 3.8
+      ``ast.PyCF_ALLOW_TOP_LEVEL_AWAIT`` can now be passed in flags to enable
+      support for top-level await, async for, and async with.
```
Carreau: note to self. Put `async ...` between backquotes.

Carreau:
All is up to date, and test are passing. Ready for one next iteration of the review.

1st1:
> All is up to date, and test are passing. Ready for one next iteration of the review.

I think there's no need for that! Thank so much for working on this and pushing this through, Matthias.  Would be great to see more contributions from you!  We need help with asyncio and many other CPython modules.

1st1:
@Carreau looks like `python -m test.test_tools` fails with this merged.  Do you have time to take a look?

ned-deily:
Please double check the bpo issue number. This PR appears to be unrelated.

1st1:
> Please double check the bpo issue number. This PR appears to be unrelated.

@ned-deily It's related -- the bpo is about adding an async exec and this PR is the way to do that.  I'm working on unbreaking buildbots in https://github.com/python/cpython/pull/13473

ned-deily:
Sorry, I should have been more specific.  For some reason, PR #13148 is also showing up as a PR under https://bugs.python.org/issue25234.  I'm not yet sure why.

Carreau:
Yes I can have a look... I'll try to catch up on what happen in the last few hours.

ned-deily:
(It looks like the spurious bpo entry was caused by something in one of the intermediate commits. I'll just delete the PR entry from bpo-25234.)

1st1:
@Carreau 

> Yes I can have a look... I'll try to catch up on what happen in the last few hours.

Never mind, I think I've fixed it already.

Carreau:
> Never mind, I think I've fixed it already.

Ok, Thanks  apologies for the breakage...

1st1:
> Ok, Thanks apologies for the breakage...

NP, it's not your fault at all.  Strangely, `test_tools` isn't run by our main CI, therefore the "unexpected" breakage.

Carreau:
> I think there's no need for that! Thank you so much for working on this and pushing this through, Matthias. 

Thanks for guiding me, it was way easier with an expert on my side. And looking forward to contribute more.




Would be great to see more contributions from you! We need help with asyncio and many other CPython modules.

> NP, it's not your fault at all. Strangely, `test_tools` isn't run by our main CI, therefore the "unexpected" breakage.

Is there an issue for that, and the fact the f-debug is not supported by test_unparse ?

1st1:
> Thanks for guiding me, it was way easier with an expert on my side. And looking forward to contribute more.

Great! 🙏

> Is there an issue for that, and the fact the f-debug is not supported by test_unparse ?

I'm not sure, but would appreciate if you could double check that.

gvanrossum:
This is great, but I wish there was also a documented way to generate an AST for such code. Currently the feature of letting `compile()` generate an AST is undocumented, and the documented way to do this is `ast.parse()`. Currently the latter has no way to pass arbitrary flags, so we would have to resort to undocumented(*) behavior if we wanted an AST for code with toplevel `await`. Could you add a flag to `ast.parse()` that also passes this flag? (Alternatively, rewrite the docs for `compile()` to document `ast.PyCF_ONLY_AST` and `ast.PyCF_TYPE_COMMENTS`.)

(*) Well, it's not completely undocumented -- the ast module's docs reference this. But `compile()`'s docs don't.

Carreau:
I believe that since 3.7, the SyntaxError is raised only by `compile()` when requesting a code-object, `ast.parse()` should behave fine with top-level await without any specific flags: 

```
$ python
Python 3.7.1 (default, Dec 14 2018, 13:28:58)
[Clang 4.0.1 (tags/RELEASE_401/final)] :: Anaconda custom (64-bit) on darwin
>>> import ast
>>>mod =  ast.parse('await sleep(0)')
>>> mod.body[0].value
<_ast.Await object at 0x107df50b8>
```

Do you want clarification in the docs; or still add a flag to ast-parse to prevent top-level-await unless opted-in ?

gvanrossum:
Huh, I didn't realize that some of these "syntax" errors come from
compile.c (not just that one, also about e.g. return outside function).

Given this, I don't think it makes sense to tweak ast.parse to *prevent*
this. I am dithering about whether to document all the flags supported by
compile() with the compile() function itself -- IIRC they are intentionally
not documented there currently because we don't want to place the burden of
supporting all of this on all other Python implementations.

So now I'm not sure whether there's any action item at all. What do you
think? What do others think?

Carreau:
> Huh, I didn't realize that some of these "syntax" errors come from compile.c (not just that one, also about e.g. return outside function).

To be fair, some of these used to happen both in making the ast and compile.c.

I'm inclined toward leaving this as-is for 3.8, and potentially refine for 3.9. It is a pretty narrow use case; and seeing what people are using this for we can refine the API(s)/documentation. 

This is still [quite tricky to use if you look at Yuri Asyncio REPL.](https://github.com/python/cpython/pull/13472/files#diff-b857c21ce7ca59e68e77b8624808d349R50), so maybe once this stabilise we can actually recommend a specific way of using it.

gvanrossum:
OK, let's leave it alone.

PR #1338, opened 2017-04-28
bpo-24263: Fix unittest can not load unicode pattern test

louisom:
Base on `sih4sing5hong5` patch, add unittest for unicode pattern.

PR #13149, opened 2019-05-07
gh-68451: Fix unittest to discover tests named with non-ascii characters

abadger:
This is a continuation of https://github.com/python/cpython/pull/1338 from @louisom which is itself based on a patch in the bpo issue tracker from```sih4sing5hong5```.

In addition to rebasing louisom's work to current master, I've done some work to address comments on the earlier code:

* I've added some comments to explain both the test data and why the AttributeError catching is present (it is to enable backporting to the externally maintained unittest package) which is something that @serhiy-storchaka was wondering about in an earlier review: https://bugs.python.org/review/24263/diff/15147/Lib/unittest/loader.py#newcode425

* @rbtcollins reviewed the earlier attempt and said:
> it looks for isidentifier on $thing.py, but not on just $thing (which we need to do to handle packages, vs modules).

I've addressed that by adding a test using isidentifier() in the isdirectory section of the conditional: https://github.com/python/cpython/pull/13149/files#diff-c1c31e722f730423a5e6dcc3eb6eef67R471

and adding directory names which are not valid identifiers (but have ```__init__.py``` inside of them) to the test cases.

https://bugs.python.org/issue24263

* Issue: gh-68451

### DIFF
--- a/Lib/unittest/loader.py
+++ b/Lib/unittest/loader.py
@@ -13,11 +13,6 @@ from . import case, suite, util
 
 __unittest = True
 
-# what about .pyc (etc)
-# we would need to avoid loading the same tests multiple times
-# from '.py', *and* '.pyc'
-VALID_MODULE_NAME = re.compile(r'[_a-z]\w*\.py$', re.IGNORECASE)
-
 
 class _FailedTest(case.TestCase):
     _testMethodName = None
@@ -372,23 +367,19 @@ class TestLoader(object):
         basename = os.path.basename(full_path)
         if os.path.isfile(full_path):
             root, ext = os.path.splitext(basename)
-            try:
-                if not (ext == '.py' and root.isidentifier()):
-                    # valid Python identifiers only
-                    return None, False
-            except AttributeError:
-                if not VALID_MODULE_NAME.match(basename):
-                    # valid Python identifiers only
-                    return None, False
+            if not (ext == '.py' and root.isidentifier()):
+                # valid Python identifiers only
+                return None, False
             if not self._match_path(basename, full_path, pattern):
                 return None, False
+
             # if the test file matches, load it
             name = self._get_name_from_path(full_path)
             try:
                 module = self._get_module_from_name(name)
             except case.SkipTest as e:
                 return _make_skipped_test(name, e, self.suiteClass), False
-            except:
+            except Exception:
                 error_case, error_message = \
                     _make_failed_import_test(name, self.suiteClass)
                 self.errors.append(error_message)
@@ -417,6 +408,10 @@ class TestLoader(object):
             load_tests = None
             tests = None
             name = self._get_name_from_path(full_path)
+
+            if not name.isidentifier():
+                return None, False
+
             try:
                 package = self._get_module_from_name(name)
             except case.SkipTest as e:
--- a/Lib/unittest/loader.py
+++ b/Lib/unittest/loader.py
@@ -29,6 +29,12 @@ class _FailedTest(case.TestCase):
         return testFailure
 
 
+def _is_valid_module_filename(module_filename):
+    root, ext = os.path.splitext(module_filename)
+    if not (ext == '.py' and root.isidentifier()):
+        return False
+    return True
+
 def _make_failed_import_test(name, suiteClass):
     message = 'Failed to import test module: %s\n%s' % (
         name, traceback.format_exc())
@@ -366,8 +372,7 @@ class TestLoader(object):
         """
         basename = os.path.basename(full_path)
         if os.path.isfile(full_path):
-            root, ext = os.path.splitext(basename)
-            if not (ext == '.py' and root.isidentifier()):
+            if not _is_valid_module_filename(basename):
                 # valid Python identifiers only
                 return None, False
             if not self._match_path(basename, full_path, pattern):
--- a/Lib/unittest/loader.py
+++ b/Lib/unittest/loader.py
@@ -371,9 +371,15 @@ class TestLoader(object):
         """
         basename = os.path.basename(full_path)
         if os.path.isfile(full_path):
-            if not VALID_MODULE_NAME.match(basename):
-                # valid Python identifiers only
-                return None, False
+            root, ext = os.path.splitext(basename)
+            try:
+                if not (ext == '.py' and root.isidentifier()):
+                    # valid Python identifiers only
+                    return None, False
+            except AttributeError:
+                if not VALID_MODULE_NAME.match(basename):
+                    # valid Python identifiers only
+                    return None, False
             if not self._match_path(basename, full_path, pattern):
                 return None, False
             # if the test file matches, load it
--- a/Lib/test/test_unittest/test_discovery.py
+++ b/Lib/test/test_unittest/test_discovery.py
@@ -92,6 +92,57 @@ class TestDiscovery(unittest.TestCase):
                     ('test3', 'test4')])
         self.assertEqual(suite, expected)
 
+    def test_find_tests_with_unicode(self):
+        loader = unittest.TestLoader()
+
+        original_listdir = os.listdir
+        def restore_listdir():
+            os.listdir = original_listdir
+        original_isfile = os.path.isfile
+        def restore_isfile():
+            os.path.isfile = original_isfile
+        original_isdir = os.path.isdir
+        def restore_isdir():
+            os.path.isdir = original_isdir
+
+        path_lists = [['測試2.py', '測試1.py', '不是測試.py', '測試_資料夾',
+                       '測試.付歐歐', '測試-不是-一個-模組.py', '另外的_資料夾'],
+                      ['測試4.py', '測試3.py', ]]
+        os.listdir = lambda path: path_lists.pop(0)
+        self.addCleanup(restore_listdir)
+
+        def isdir(path):
+            return path.endswith('資料夾')
+        os.path.isdir = isdir
+        self.addCleanup(restore_isdir)
+
+        def isfile(path):
+            # another_dir is not a package and so shouldn't be recursed into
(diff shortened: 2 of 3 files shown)


Lib/unittest/loader.py
```diff
@@ -425,11 +425,20 @@ def _find_test_path(self, full_path, pattern, namespace=False):
...
-            if not VALID_MODULE_NAME.match(basename):
-                # valid Python identifiers only
-                return None, False
+            root, ext = os.path.splitext(basename)
+            try:
+                if not (ext == '.py' and root.isidentifier()):
+                    # valid Python identifiers only
+                    return None, False
+            except AttributeError:
+                # Python 2 does not have isidentifier() and also has simpler
```
tonybaloney: Note for core reviewers: Would this be backported to Python 2? I doubt it, in which case this change can be dramatically simplified.

Lib/unittest/loader.py
```diff
@@ -425,11 +425,20 @@ def _find_test_path(self, full_path, pattern, namespace=False):
...
         if os.path.isfile(full_path):
-            if not VALID_MODULE_NAME.match(basename):
-                # valid Python identifiers only
-                return None, False
+            root, ext = os.path.splitext(basename)
+            try:
+                if not (ext == '.py' and root.isidentifier()):
+                    # valid Python identifiers only
+                    return None, False
+            except AttributeError:
```
ezio-melotti: I don't think we need to keep this branch (and the regex) on Python 3.
If the change is backported to 2.7, this branch can be used instead.

Lib/unittest/loader.py
```diff
@@ -466,6 +475,10 @@ def _find_test_path(self, full_path, pattern, namespace=False):
             load_tests = None
             tests = None
             name = self._get_name_from_path(full_path)
+
+            if not name.isidentifier():
+                return None, False
```
ezio-melotti, line 418: If this is backported to Python 2, it will fail because isidentifier is not available there.

Lib/unittest/test/test_discovery.py
```diff
@@ -53,8 +53,13 @@ def restore_isfile():
         def restore_isdir():
             os.path.isdir = original_isdir
 
-        path_lists = [['test2.py', 'test1.py', 'not_a_test.py', 'test_dir',
-                       'test.foo', 'test-not-a-module.py', 'another_dir'],
+        path_lists = [  # Valid module and package names
+                      ['test2.py', 'test1.py', 'not_a_test.py', 'test_dir',
```
ezio-melotti: I would move the `[` before the comment, otherwise it seems like the comment applies to the whole list.

Lib/unittest/test/test_discovery.py
```diff
@@ -91,6 +96,62 @@ def loadTestsFromModule(module, pattern=None):
...
+            os.listdir = original_listdir
+        original_isfile = os.path.isfile
+        def restore_isfile():
+            os.path.isfile = original_isfile
+        original_isdir = os.path.isdir
+        def restore_isdir():
+            os.path.isdir = original_isdir
+
+        path_lists = [  # Valid module and package names
+                      ['測試2.py', '測試1.py', '不是測試.py', '測試_資料夾',
```
ezio-melotti: Same as above.

Lib/unittest/test/test_discovery.py
```diff
@@ -91,6 +96,62 @@ def loadTestsFromModule(module, pattern=None):
...
+                       # Invalid names
+                       '測試.付歐歐', '測試-不是-一個-模組.py',
+                       '測試-不是-一個-模組_資料夾', '另外的_資料夾'],
+                       # Valid names; test case tests that these work from inside
+                       # of a package directory
+                      ['測試4.py', '測試3.py', ]]
+        os.listdir = lambda path: path_lists.pop(0)
+        self.addCleanup(restore_listdir)
```
ezio-melotti: Instead of defining all the `restore_*` functions, I would just do something like `self.addCleanup(setattr, os, 'listdir', original_listdir)`.
Another option is using `mock.patch`, but maybe here it's overkill.

Lib/unittest/test/test_discovery.py
```diff
@@ -91,6 +96,62 @@ def loadTestsFromModule(module, pattern=None):
...
+
+        def isdir(path):
+            return path.endswith('資料夾')
+        os.path.isdir = isdir
+        self.addCleanup(restore_isdir)
+
+        def isfile(path):
+            # '另外的_資料夾' is not a package and so shouldn't be recursed into
+            return not path.endswith('資料夾') and not '另外的_資料夾' in path
```
ezio-melotti: `return not path.endswith('資料夾') and '另外的_資料夾' not in path`

Lib/unittest/test/test_discovery.py
```diff
@@ -91,6 +96,62 @@ def loadTestsFromModule(module, pattern=None):
...
+                       '測試-不是-一個-模組_資料夾', '另外的_資料夾'],
+                       # Valid names; test case tests that these work from inside
+                       # of a package directory
+                      ['測試4.py', '測試3.py', ]]
+        os.listdir = lambda path: path_lists.pop(0)
+        self.addCleanup(restore_listdir)
+
+        def isdir(path):
+            return path.endswith('資料夾')
+        os.path.isdir = isdir
```
ezio-melotti: Any reason why here you defined an `isdir` function instead of using a `lambda`?
(FWIW I think for the sake of brevity `lambda`s are fine in this context.)

Lib/unittest/test/test_discovery.py
```diff
@@ -91,6 +96,62 @@ def loadTestsFromModule(module, pattern=None):
...
+
+        top_level = os.path.abspath('/foo')
+        loader._top_level_dir = top_level
+        suite = list(loader._find_tests(top_level, '測試*.py'))
+
+        # The test suites found should be sorted alphabetically for reliable
+        # execution order.
+        expected = [[name + ' module tests'] for name in
+                    ('測試1', '測試2', '測試_資料夾')]
+        expected.extend([[('測試_資料夾.%s' % name) + ' module tests'] for name in
```
ezio-melotti: `expected.extend([['測試_資料夾.%s module tests' % name] for name in`

Lib/unittest/loader.py
vstinner: Wait, why do you care of Python 2 in the master branch? You should remove VALID_MODULE_NAME and the except AttributeError.

abadger: @rbtcollins wanted it.  I believe to make it easy to backport to https://pypi.org/project/unittest2/

I can perform the manual work of either cleaning the code for Python3 or making it easily backportable but I don't feel qualified to make the decision as to which of those strategies is correct.

ezio-melotti (changes requested):

abadger:
@rbtcollins, it seems like the people presently interested in this PR are leaning towards removing the Python2 compatible code so I'm going to go ahead and remove that.

Lib/unittest/loader.py
abadger: @rbtcollins hasn't replied so I'll go ahead and remove the Python2 compatibility

Lib/unittest/loader.py
abadger: @rbtcollins hasn't replied so I'll go ahead and remove the Python2 compatibility

Lib/unittest/loader.py
abadger: Note, AFAICT, Robert Collins just wanted to make this fix available in the unittest2 package, not backport it to the Pyhton2 branch.  Since he hasn't replied, though, I'll go ahead and remove the python2-isms.

Lib/unittest/test/test_discovery.py
abadger: Done

Lib/unittest/test/test_discovery.py
abadger: Done

Lib/unittest/test/test_discovery.py
abadger: So, for the remainder of these comments.... It appears that this test_find_tests_with_unicode() function is a copy of the pre-existing test, test_find_tests(), adjusted to use new data.

If you'd like me to make these changes, I can, but I think I'd pair it with merging the two tests.

abadger:
@ezio-melotti Updated.  I believe with all of the changes that you and @vstinner asked for.

abadger:

Lib/unittest/loader.py
```diff
@@ -14,11 +14,6 @@
 
 __unittest = True
 
-# what about .pyc (etc)
-# we would need to avoid loading the same tests multiple times
-# from '.py', *and* '.pyc'
-VALID_MODULE_NAME = re.compile(r'[_a-z]\w*\.py$', re.IGNORECASE)
```
ezio-melotti, line 19: I think it would be better to keep this global and update it to use `r'[^\W\d]\w*\.py$'`, but as @serhiy-storchaka pointed out in https://github.com/python/cpython/issues/68451#issuecomment-1093683815 this doesn't match all identifiers.

Perhaps a `is_valid_module_name()` global function could be added instead -- this will still provide a way to control what is valid, instead of hardcoding the check in `_find_test_path()`.

abadger: I've added this as a private function in loader: `_is_valid_module_filename()`.  The code that makes use of the function makes use of it to tell if a filename is a valid python module name.  A different code path operates on directory names.  I think I put the amount of logic into the function that you were thinking (the equivalent of what the regex replaced).  Let me know if you were thinking of something with a larger scope.

abadger:

ziima:
@ezio-melotti please review the changes.

I found the issue and PR as part of EuroPython 24 sprints.

PR #13157, opened 2019-05-07
bpo-28866: No type cache for types with specialized mro, invalidation is hard.

JulienPalard:
This is an alternative implementation of https://github.com/python/cpython/pull/13117 based on @markshannon idea (https://bugs.python.org/issue28866#msg341581).

https://bugs.python.org/issue28866

### DIFF
--- a/Objects/typeobject.c
+++ b/Objects/typeobject.c
@@ -78,6 +78,9 @@ slot_tp_new(PyTypeObject *type, PyObject *args, PyObject *kwds);
 static void
 clear_slotdefs(void);
 
+static PyObject *
+lookup_method(PyObject *self, _Py_Identifier *attrid, int *unbound);
+
 /*
  * finds the beginning of the docstring's introspection signature.
  * if present, returns a pointer pointing to the first '('.
@@ -300,9 +303,11 @@ type_mro_modified(PyTypeObject *type, PyObject *bases) {
         return;
 
     if (custom) {
+        _Py_IDENTIFIER(mro);
+        int unbound;
         PyObject *mro_meth = lookup_method((PyObject *)type, &PyId_mro,
                                            &unbound);
-        PyObject *type_mro_meth = lookup_method(&PyType_Type, &PyId_mro,
+        PyObject *type_mro_meth = lookup_method((PyObject *)&PyType_Type, &PyId_mro,
                                            &unbound);
         if (mro_meth == NULL || type_mro_meth == NULL ||
             mro_meth != type_mro_meth)
--- a/Objects/typeobject.c
+++ b/Objects/typeobject.c
@@ -1925,6 +1925,8 @@ mro_invoke(PyTypeObject *type)
                                            &unbound);
         if (mro_meth == NULL)
             return NULL;
+        type->tp_flags &= ~(Py_TPFLAGS_HAVE_VERSION_TAG|
+                            Py_TPFLAGS_VALID_VERSION_TAG);
         mro_result = call_unbound_noarg(unbound, mro_meth, (PyObject *)type);
         Py_DECREF(mro_meth);
     }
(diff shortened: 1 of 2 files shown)


Objects/typeobject.c
```diff
@@ -1925,6 +1925,8 @@ mro_invoke(PyTypeObject *type)
                                            &unbound);
         if (mro_meth == NULL)
             return NULL;
+        type->tp_flags &= ~(Py_TPFLAGS_HAVE_VERSION_TAG|
+                            Py_TPFLAGS_VALID_VERSION_TAG);
```
markshannon: This will disable caching for any class with a metaclass that is not `type`, even if it does not override `mro()`.
Changing `custom` to detect whether `mro()` has been overridden should fix that.

JulienPalard: Right after the `if (mro_meth == NULL) return;` ? I though it would returns if there's no mro_meth.

JulienPalard: Oh, you're right, it finds the mro method of the type object.

Objects/typeobject.c
```diff
@@ -287,32 +290,45 @@ type_mro_modified(PyTypeObject *type, PyObject *bases) {
...
         return;
 
-    n = PyTuple_GET_SIZE(bases);
-    for (i = 0; i < n; i++) {
-        PyObject *b = PyTuple_GET_ITEM(bases, i);
-        PyTypeObject *cls;
+    if (custom) {
+        _Py_IDENTIFIER(mro);
+        int unbound;
+        PyObject *mro_meth = lookup_method((PyObject *)type, &PyId_mro,
```
ZackerySpytz: This first `lookup_method()` call should be checked for failure before the second call. If the first call fails, an assertion failure may occur during the second call in `_PyType_Lookup()` (as `_PyType_Lookup()` performs `assert(!PyErr_Occurred());`).

JulienPalard: Thanks for the review @ZackerySpytz :) It's fixed, I think, by avoiding to call lookup_method which is the one setting the error.

Objects/typeobject.c
```diff
@@ -287,32 +290,51 @@ type_mro_modified(PyTypeObject *type, PyObject *bases) {
...
-        assert(PyType_Check(b));
-        cls = (PyTypeObject *)b;
-
-        if (!PyType_HasFeature(cls, Py_TPFLAGS_HAVE_VERSION_TAG) ||
-            !PyType_IsSubtype(type, cls)) {
+    if (custom) {
+        _Py_IDENTIFIER(mro);
+        int unbound;
+        PyObject *mro_meth = lookup_maybe_method(
+            (PyObject *)type, &PyId_mro, &unbound);
```
markshannon, line 310: `mro_meth` needs to be checked for NULL before `lookup_maybe_method` is called again.

JulienPalard: OK :-] is using a goto legit here? It'll avoid a bunch of `if (clear == 0)` (see current impl).

Objects/typeobject.c
```diff
@@ -308,14 +327,15 @@ type_mro_modified(PyTypeObject *type, PyObject *bases) {
...
-            clear = 1;
-            break;
+            goto clear;
         }
     }
-
-    if (clear)
-        type->tp_flags &= ~(Py_TPFLAGS_HAVE_VERSION_TAG|
-                            Py_TPFLAGS_VALID_VERSION_TAG);
+    return;
```
markshannon, line 335: Doesn't this leak a reference to `mro_meth` and `type_mro_meth`?

JulienPalard: Thanks for noticing! Fixed :)

markshannon:
LGTM

PR #13160, opened 2019-05-07
fix broken link in unicode docs

redshiftzero:


tirkarthi (approved):
LGTM and I verified the rendering locally. Thanks for the catch.

deronnax:
would you mind backporting it to 3.7 ?

PR #13168, opened 2019-05-07
subinterpreters: FIX a memory leak

JulienPalard:
Was playing with:

    $ ./configure --with-assertions --with-lto --with-pydebug --with-address-sanitizer --disable-ipv6
    $ make -j4 profile-opt CFLAGS='-D __INSURE__'

(triple-checking while trying to close https://bugs.python.org/issue31200)

### DIFF
--- a/Include/internal/pycore_pylifecycle.h
+++ b/Include/internal/pycore_pylifecycle.h
@@ -20,6 +20,7 @@ extern int _Py_SetFileSystemEncoding(
     const char *errors);
 extern void _Py_ClearFileSystemEncoding(void);
 extern PyStatus _PyUnicode_InitEncodings(PyThreadState *tstate);
+extern void _PyUnicode_FiniEncodings(PyThreadState *tstate);
 #ifdef MS_WINDOWS
 extern int _PyUnicode_EnableLegacyWindowsFSEncoding(void);
 #endif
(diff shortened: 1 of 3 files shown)


JulienPalard:
@ericsnowcurrently Hi, could you review this PR / tell me if it looks useful?

ericsnowcurrently:
This should have a BPO issue (or refer to bpo-31200 if that applies).

ericsnowcurrently (commented):
Mostly LGTM. :)

Is there a test you could add?

ericsnowcurrently:
@JulienPalard ping

JulienPalard:
Problem is: the leak I fixed has no refcount, so I can't spot it using classical refcount checks. Is there a way to still write a test to test it or should we just delegate this to a build bot running address sanitizer?

JulienPalard:
@ericsnowcurrently spoke with Victor, there's a way (`./python -m test -v test__xxsubinterpreters -R 3:3`) to find `PyMem_Malloc` leaks, but there's no way to find `PyMem_RawMalloc` except by using the memory sanitizer. I'll check with Zach if we have a build bot running it, and add it as needed.

I think this can be merged as is now, can you take a look?

vstinner:
Oh, I wasn't aware of this PR. I added _PyUnicode_FiniEncodings() in commit bf305cc6f05948f264349a6a6c6fd7d49c1839d3. It's called indirectly by Py_EndInterpreter(): finalize_interp_types() calls _PyUnicode_Fini(). This PR became redundant, I close it.

Thanks anyway @JulienPalard  ;-)

Note: the main issue to enhance Py_FinalizeEx() is https://bugs.python.org/issue1635741

PR #13172, opened 2019-05-07
gh-81005: Refactor str tests to reflect that str and unicode are merged in Python 3

asqui:
Various cleanup and refactoring of unit tests for string-like types.
See https://bugs.python.org/issue36824 and/or individual commit messages for more details.

I haven't done the "Promote tests from the current `UnicodeTest` class to `CommonTest`" portion yet, but I'm running out of sprint time so I wanted to submit what I have.

I believe these changes are suitable for merge to master, but keen to hear feedback :-)

https://bugs.python.org/issue36824

* Issue: gh-81005


asqui:
As this is a test-only change with no external visibility, I don' think it needs a NEWS entry. Should the "skip news" label be applied to this PR?

jdemeyer (approved):
The changes look good to me. I agree with the `skip news`.

iritkatriel (dismissed):
This has merge conflicts now,.

furkanonder:
@asqui Could you resolve the conflicts?

asqui:
Sure! I'm AFK for the next few days but I'll take a look when I'm back next week.

asqui:
Changes to the renamed file merged in cleanly with absolutely no conflicts. (I was surprised given the age of this PR so did some sanity checks and it all looks good.)

CLA signed (again; was there a change to the CLA that required everyone to re-sign?)

I think this is good to merge now? (Needs the _Skip News_ label.)

arhadthedev:
> (Needs the Skip News label.)

I think that something like <code>String tests are modified to reflect that \`\`str\`\` and \`\`unicode\`\` are merged in Python 3. Patch by Daniel Fortunov.</code> is indeed required here. The list of tests got changed so whoever runs them periodically to build comparison graphs needs to be aware.

/cc @AlexWaygood (to check whether I apply https://github.com/python/cpython/pull/104570#issuecomment-1551964864 correctly)

AlexWaygood:
> > (Needs the Skip News label.)
> 
> I think that something like `` String tests are modified to reflect that ``str`` and ``unicode`` are merged in Python 3. Patch by Daniel Fortunov. `` is indeed required here. The list of tests got changed so whoever runs them periodically to build comparison graphs needs to be aware.
> 
> /cc @AlexWaygood (to check whether I apply [#104570 (comment)](https://github.com/python/cpython/pull/104570#issuecomment-1551964864) correctly)

Good call @arhadthedev. Changes to our test suite often don't need NEWS entries, as they don't have much impact on end users. But something as significant as renaming a whole test file probably _does_ need a NEWS entry, as it will affect the invocation people will need to use to run the tests.

asqui:
Added the suggested NEWS entry.

(This is my first time using `blurb` so hopefully I did it correctly!)

asqui:
Hmm looks like I should have used the GH issue number for the bpo-imported issue that this PR is fixing (I mistakenly put the number for this PR itself)

I'm AFK again for the next few days so won't be able to fix this until next week.

I guess all is needed is to rename the news file to feature the correct number - I don't know of any way to do that through GitHub.com without my laptop.

hugovk:
I've renamed the NEWS file to use the GitHub issue number.

hugovk (approved):

hugovk:
And I've opened https://github.com/python/core-workflow/issues/504 to suggest more validation in blurb.

hugovk:
Thank you for the PR and for your patience!

asqui:
Yay, merged at last! Thanks for the reviews and help along the way 😄

PR #13175, opened 2019-05-07
bpo-36829: Add a -X option to abort in PyErr_WriteUnraisable()

ZackerySpytz:
Add a -X command-line option to abort the current process
if PyErr_WriteUnraisable() is called.

### DIFF
--- a/Doc/c-api/exceptions.rst
+++ b/Doc/c-api/exceptions.rst
@@ -81,6 +81,10 @@ Printing and clearing
    in which the unraisable exception occurred. If possible,
    the repr of *obj* will be printed in the warning message.
 
+   .. versionchanged:: 3.8
+      The :option:`-X` ``abortunraisable`` command-line option causes this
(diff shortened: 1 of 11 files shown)


matrixise:
Hi @ZackerySpytz 

Do you think you could add a test?

vstinner:
I am not sure I like this feature. Would it make sense to add a callback instead, like sys.excepthook? The callback could log the error, do nothing, abort the process, call the police, etc. It might be more general, no?

graingert:
@vstinner PyErr_WriteUnraisable can get called during `__del__` or GC so most of the python environment could have already been deleted by the time the hook is called,

Or the hook could be called after the hook is removed by GC and then we're back to the same problem

vstinner:
I merged my PR #13187 instead, so I reject this PR. See https://bugs.python.org/issue36829#msg343201 for the rationale.

PR #13177, opened 2019-05-07
bpo-36839: Support the buffer protocol in code objects

DinoV:
[bpo-36839](https://bugs.python.org/issue36839): Support the buffer protocol in code objects

Adds support for code objects to be backed by any objects which implement the buffer protocol.  This relatively small core interpreter change makes it possible to build importer/loaders which can store their code objects in memory mapped files.  When used in an environment with fork/exec'd processes this can allow the memory to be successfully shared between all of the processes.

https://bugs.python.org/issue36839

### DIFF
--- a/Lib/dis.py
+++ b/Lib/dis.py
@@ -77,13 +77,18 @@ def dis(x=None, *, file=None, depth=None):
                 print(file=file)
     elif hasattr(x, 'co_code'): # Code object
         _disassemble_recursive(x, file=file, depth=depth)
-    elif isinstance(x, (bytes, bytearray)): # Raw bytecode
+    elif isinstance(x, (bytes, bytearray, memoryview)): # Raw bytecode
         _disassemble_bytes(x, file=file)
     elif isinstance(x, str):    # Source code
         _disassemble_str(x, file=file, depth=depth)
     else:
-        raise TypeError("don't know how to disassemble %s objects" %
-                        type(x).__name__)
+        try:
+            x = memoryview(x)
+        except TypeError:
+            raise TypeError("don't know how to disassemble %s objects" %
+                            type(x).__name__)
+        _disassemble_bytes(x, file=file)
+
 
 def distb(tb=None, *, file=None):
     """Disassemble a traceback (default: last traceback)."""
--- a/Lib/dis.py
+++ b/Lib/dis.py
@@ -84,7 +84,7 @@ def dis(x=None, *, file=None, depth=None):
     else:
         try:
             x = memoryview(x)
-        except TypeError:
+        except Exception:
             raise TypeError("don't know how to disassemble %s objects" %
                             type(x).__name__)
         _disassemble_bytes(x, file=file)
--- a/Lib/test/test_code.py
+++ b/Lib/test/test_code.py
@@ -125,8 +125,8 @@ consts: ('None',)
 
 """
 
-import array
 import inspect
+import mmap
 import sys
 import threading
 import unittest
@@ -184,15 +184,49 @@ implementing the buffer protocol and that they can successfully be evaluated'''
 
         code = f.__code__
         ct = type(f.__code__)
-        c = ct(code.co_argcount, code.co_posonlyargcount,
-               code.co_kwonlyargcount, code.co_nlocals, code.co_stacksize,
-               code.co_flags, array.array('B', code.co_code),
-               code.co_consts, code.co_names, code.co_varnames,
-               code.co_filename, code.co_name, code.co_firstlineno,
-               code.co_lnotab, code.co_freevars, code.co_cellvars)
-        nf = type(f)(c, f.__globals__, 'nf', f.__defaults__, f.__closure__)
-        self.assertEqual(f(), nf())
-        self.assertEqual(type(nf.__code__.co_code), array.array)
+        from tempfile import NamedTemporaryFile
+
+        def make_mapped_func(map):
+            c = ct(code.co_argcount, code.co_posonlyargcount,
+                   code.co_kwonlyargcount, code.co_nlocals, code.co_stacksize,
+                   code.co_flags, map,
+                   code.co_consts, code.co_names, code.co_varnames,
+                   code.co_filename, code.co_name, code.co_firstlineno,
+                   code.co_lnotab, code.co_freevars, code.co_cellvars)
+            nf = type(f)(c, f.__globals__, 'nf', f.__defaults__, f.__closure__)
+            return nf
+
+        with NamedTemporaryFile('wb', suffix='.code') as tmp_code:
+            tmp_code.write(f.__code__.co_code)
+            tmp_code.flush()
+            with mmap.mmap(tmp_code.fileno(),
+                            length=0, access=mmap.ACCESS_READ) as map:
+                nf = make_mapped_func(map)
+                self.assertEqual(f(), nf())
+                self.assertEqual(type(nf.__code__.co_code), mmap.mmap)
+
+        # Closed buffer will raise
+        self.assertRaises(BufferError, nf)
+        buf_cannot_close = "cannot close exported pointers exist"
+        with NamedTemporaryFile('wb', suffix='.code') as tmp_code:
+            tmp_code.write(f.__code__.co_code)
+            tmp_code.flush()
+            with mmap.mmap(tmp_code.fileno(),
+                            length=0, access=mmap.ACCESS_READ) as map:
+                nf = make_mapped_func(map)
+                def trace_func(frame, event, arg):
+                    if event == 'call':
+                        return trace_func
+                    elif event == 'line':
+                        # eval loop should keep buffer open
+                        self.assertRaisesRegex(BufferError, buf_cannot_close,
+                                               map.close)
+                prev_trace = sys.gettrace()
+                try:
+                    sys.settrace(trace_func)
+                    nf()
+                finally:
+                    sys.settrace(prev_trace)
 
     @cpython_only
     def test_closure_injection(self):
--- a/Lib/test/test_code.py
+++ b/Lib/test/test_code.py
@@ -125,6 +125,7 @@ consts: ('None',)
 
 """
 
+import array
 import inspect
 import sys
 import threading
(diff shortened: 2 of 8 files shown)


Objects/frameobject.c
```diff
@@ -316,10 +324,17 @@ frame_setlineno(PyFrameObject *f, PyObject* p_new_lineno, void *Py_UNUSED(ignore
         delta--;
     }
 
+    if (view.obj != NULL)
```
1st1: nit: The latest C code style recommends using `{}` always, even for single line clauses.

1st1 (approved):

Objects/genobject.c
```diff
@@ -345,7 +354,11 @@ _PyGen_yf(PyGenObject *gen)
             return NULL;
         }
 
-        if (code[f->f_lasti + sizeof(_Py_CODEUNIT)] != YIELD_FROM)
+        char opcode = code[f->f_lasti + sizeof(_Py_CODEUNIT)];
+        if (view.obj != NULL)
```
markshannon: Also needs braces

Objects/codeobject.c
```diff
@@ -94,6 +94,21 @@ intern_string_constants(PyObject *tuple)
...
 }
 
+static int
+code_check_buffer(PyObject *code) {
+    Py_buffer codebuffer = {};
+    if (!PyBytes_Check(code)) {
+        if (PyObject_GetBuffer(code, &codebuffer, PyBUF_SIMPLE))
+            return 0;
+
+        int contiguous = PyBuffer_IsContiguous(&codebuffer, 'C');
```
markshannon: This doesn't seem right. 
The buffer should be read-only and one dimensional, not just contiguous.

Lib/dis.py
```diff
@@ -77,13 +77,18 @@ def dis(x=None, *, file=None, depth=None):
...
+    elif isinstance(x, (bytes, bytearray, memoryview)): # Raw bytecode
         _disassemble_bytes(x, file=file)
     elif isinstance(x, str):    # Source code
         _disassemble_str(x, file=file, depth=depth)
     else:
-        raise TypeError("don't know how to disassemble %s objects" %
-                        type(x).__name__)
+        try:
+            x = memoryview(x)
+        except TypeError:
```
ericsnowcurrently: I expect that `TypeError` isn't the only way that `memoryview()` can fail.  For instance, there's a good chance that there are code paths that result in `ValueError`.  So it may be worth catching `Exception` instead, just to ensure that the behavior remains the same. :)

Lib/test/test_code.py
```diff
@@ -170,6 +171,28 @@ def test_newempty(self):
...
+implementing the buffer protocol and that they can successfully be evaluated'''
+        def f():
+            x = 0
+            for i in range(10):
+                x += i
+            return x
+
+        code = f.__code__
+        ct = type(f.__code__)
+        c = ct(code.co_argcount, code.co_posonlyargcount,
```
ericsnowcurrently: much respect! :)

Lib/test/test_code.py
```diff
@@ -170,6 +171,28 @@ def test_newempty(self):
...
+               code.co_kwonlyargcount, code.co_nlocals, code.co_stacksize,
+               code.co_flags, array.array('B', code.co_code),
```
ericsnowcurrently: FWIW, inlining the array hides it.  I knew to look for it due to the context of this PR.  So it would be worth making a separate variable for it above.

Lib/test/test_dis.py
```diff
@@ -453,6 +453,23 @@ def test_widths(self):
...
+    def test_dis_memoryview(self):
+        self.do_disassembly_test(memoryview(_f.__code__.co_code), dis_f_co_code)
+
+    def test_dis_mmap(self):
+        try:
+            import mmap
+        except ImportError:
+            return
+        from tempfile import NamedTemporaryFile
+        with NamedTemporaryFile('wb', suffix='.code') as f:
```
ericsnowcurrently: FWIW, "f" as a name is a bit ambiguous in this file. :)  However, it isn't confusing here.

Objects/codeobject.c
```diff
@@ -94,6 +94,21 @@ intern_string_constants(PyObject *tuple)
     return modified;
 }
 
+static int
+code_check_buffer(PyObject *code) {
+    Py_buffer codebuffer = {};
+    if (!PyBytes_Check(code)) {
+        if (PyObject_GetBuffer(code, &codebuffer, PyBUF_SIMPLE))
```
ericsnowcurrently: missing braces (see PEP 7)

Objects/codeobject.c
```diff
@@ -367,7 +382,8 @@ code_new(PyTypeObject *type, PyObject *args, PyObject *kw)
     int firstlineno;
     PyObject *lnotab;
 
-    if (!PyArg_ParseTuple(args, "iiiiiiSO!O!O!UUiS|O!O!:code",
+
+    if (!PyArg_ParseTuple(args, "iiiiiiOO!O!O!UUiS|O!O!:code",
```
ericsnowcurrently, line 390: Just to be sure I understood, you only changed the first "S" to "O" and nothing else, right?

Objects/frameobject.c
```diff
@@ -199,15 +199,23 @@ frame_setlineno(PyFrameObject *f, PyObject* p_new_lineno, void *Py_UNUSED(ignore
...
     /* We're now ready to look at the bytecode. */
-    PyBytes_AsStringAndSize(f->f_code->co_code, (char **)&code, &code_len);
+    Py_buffer view = {};
+    if (PyBytes_Check(f->f_code->co_code)) {
+        PyBytes_AsStringAndSize(f->f_code->co_code, (char **)&code, &code_len);
+    } else if (!PyObject_GetBuffer(f->f_code->co_code, &view, PyBUF_SIMPLE)) {
+        code = (unsigned char *)view.buf;
+        code_len = view.len;
+    } else {
+        return -1;
```
ericsnowcurrently, line 211: You need to set an error here, no?

ericsnowcurrently: Which test is checking the cases you've added here?

Objects/frameobject.c
```diff
@@ -199,15 +199,23 @@ frame_setlineno(PyFrameObject *f, PyObject* p_new_lineno, void *Py_UNUSED(ignore
     }
 
     /* We're now ready to look at the bytecode. */
-    PyBytes_AsStringAndSize(f->f_code->co_code, (char **)&code, &code_len);
+    Py_buffer view = {};
```
ericsnowcurrently: Would it make sense to factor this out to something like a `unsigned char * _PyCode_GetBytes(PyObject *)` helper?  We could use it here and in genobject.c, etc.

Python/ceval.c
```diff
@@ -978,11 +979,20 @@ _PyEval_EvalFrameDefault(PyFrameObject *f, int throwflag)
...
+        assert(PyBytes_GET_SIZE(co->co_code) % sizeof(_Py_CODEUNIT) == 0);
+        assert(_Py_IS_ALIGNED(PyBytes_AS_STRING(co->co_code), sizeof(_Py_CODEUNIT)));
+        first_instr = (_Py_CODEUNIT *) PyBytes_AS_STRING(co->co_code);
+    } else if (!PyObject_GetBuffer(co->co_code, &code_view, PyBUF_SIMPLE)) {
+        assert(code_view.len <= INT_MAX);
+        assert(code_view.len % sizeof(_Py_CODEUNIT) == 0);
+        assert(_Py_IS_ALIGNED(code_view.buf, sizeof(_Py_CODEUNIT)));
+        first_instr = (_Py_CODEUNIT *) code_view.buf;
+    } else {
+        goto exit_eval_frame;
```
ericsnowcurrently, line 1044: As before, don't we need to set an error here?  Where are the two new code paths tested?

ericsnowcurrently: Do we even need this case?  In the original code this was covered by an assert...  I'm not opposed to an exception instead of an insert though if you think this give us a better experience. :)

Python/ceval.c
```diff
@@ -978,11 +979,20 @@ _PyEval_EvalFrameDefault(PyFrameObject *f, int throwflag)
...
-    assert(PyBytes_GET_SIZE(co->co_code) <= INT_MAX);
-    assert(PyBytes_GET_SIZE(co->co_code) % sizeof(_Py_CODEUNIT) == 0);
-    assert(_Py_IS_ALIGNED(PyBytes_AS_STRING(co->co_code), sizeof(_Py_CODEUNIT)));
-    first_instr = (_Py_CODEUNIT *) PyBytes_AS_STRING(co->co_code);
+    if (PyBytes_Check(co->co_code)) {
+        assert(PyBytes_GET_SIZE(co->co_code) <= INT_MAX);
+        assert(PyBytes_GET_SIZE(co->co_code) % sizeof(_Py_CODEUNIT) == 0);
+        assert(_Py_IS_ALIGNED(PyBytes_AS_STRING(co->co_code), sizeof(_Py_CODEUNIT)));
+        first_instr = (_Py_CODEUNIT *) PyBytes_AS_STRING(co->co_code);
+    } else if (!PyObject_GetBuffer(co->co_code, &code_view, PyBUF_SIMPLE)) {
```
ericsnowcurrently, line 1036: I realize this isn't your fault, but I keep getting thrown off by the return value of `PyObject_GetBuffer()`. :)

ericsnowcurrently (approved):
Overall, LGTM.  I've left a number of comments about things that seemed to be missing.

I'm going to approve the PR, but with the understanding that you will address them (or dismiss them as misunderstandings). :)

Objects/codeobject.c
DinoV: Correct, I wish github had some inline highlighting here

Objects/frameobject.c
DinoV: This should really be an unreachable condition - the construction of code objects checks and makes sure we have a reasonable buffer.  It should definitely be setting an error, I'll see if I can poke around and produce an error through extra evil measures.

DinoV: Oh, another possible way to hit this would be in low memory conditions, but there's not great infrastructure to test that either.

Python/ceval.c
DinoV: I think we should handle the case where PyObject_GetBuffer fails as I think it could require allocating, so will definitely add the missing error.

Objects/frameobject.c
DinoV: Let me think about this...  I think the sig needs to actually be unsigned char* _PyCode_GetBytes(PyObject *, Py_ssize_t *len, Py_buffer *view) which I feel like is getting complicated enough that it's not going to save much.  But let me see if it actually helps much.

Objects/codeobject.c
DinoV: Well tighten this up to:

codebuffer.readonly &&
                 codebuffer.ndim == 1 &&
                 codebuffer.strides == NULL;

I think that should satisfy the readonly check because per the documentation it needs to be consistent for all consumers: https://docs.python.org/3/c-api/buffer.html#c.PyBUF_WRITABLE

the-knights-who-say-ni:

Python/ceval.c
DinoV: And I was able to hit this case, by closing the memory mapped file and then re-running the function.

Objects/frameobject.c
DinoV: And this one seems to be generally unreachable by a well-behaved buffer protocol object.

Objects/codeobject.c
skrah: For a well-behaved exporter, `PyBUF_SIMPLE` implies `C`, `ndim==1`, format `"B"` and `readonly`.

I know the code base has these (redundant) checks in other places, but at some point we may want to use `assert()`.  I'm not suggesting to eliminate the extra safety check here and now, just clarifying.

Objects/frameobject.c
gpshead: I usually add an `assert(condition)` to the code in these cases just so it shows up if we ever manage to trigger such a condition.  _(Our non-release builds at work for unittests use an assertion enabled interpreter - so it isn't just the python.org buildbots that would notice)_

matrixise:
Hi @DinoV 

Do you want to fix the conflicts? Thanks

methane (changes requested):
Please don't merge this until the benefit is proofed. (see b.p.o)
Currently, the idea is just armchair philosophy.

PR #13180, opened 2019-05-07
bpo-36674: Stops skipped tests from running in debug mode.

lisroach:
TestCase.run() will skip tests that have skip/skipIf/skipWhatever applied to them, this PR applies the same skipped functionality when tests are run using TestCase.debug().

https://bugs.python.org/issue36674

https://bugs.python.org/issue36674

### DIFF
--- a/Lib/unittest/case.py
+++ b/Lib/unittest/case.py
@@ -745,7 +745,11 @@ class TestCase(object):
     def debug(self):
         """Run the test without collecting errors in a TestResult"""
         self.setUp()
-        getattr(self, self._testMethodName)()
+        testMethod = getattr(self, self._testMethodName)
+        if (not getattr(self.__class__, "__unittest_skip__", False) and
+            not getattr(testMethod, "__unittest_skip__", False)):
+            # If the class or method was not skipped run it
+            testMethod()
         self.tearDown()
         while self._cleanups:
             function, args, kwargs = self._cleanups.pop(-1)
--- a/Lib/unittest/test/test_case.py
+++ b/Lib/unittest/test/test_case.py
@@ -1777,6 +1777,31 @@ test case
             self.assertEqual(len(result.skipped), 1)
             self.assertEqual(result.testsRun, 1)
 
+    def testSkippingDebugMode(self):
+        def _skip(self=None):
+            raise unittest.SkipTest('some reason')
+        def nothing(self):
+            pass
+
+        class Test1(unittest.TestCase):
+            test_something = _skip
+
+        class Test2(unittest.TestCase):
+            setUp = _skip
+            test_something = nothing
+
+        class Test3(unittest.TestCase):
+            test_something = nothing
+            tearDown = _skip
+
+        class Test4(unittest.TestCase):
+            def test_something(self):
+                self.addCleanup(_skip)
+
+        for klass in (Test1, Test2, Test3, Test4):
+            with self.assertRaises(unittest.SkipTest):
+                klass('test_something').debug()
+
     def testSystemExit(self):
         def _raise(self=None):
             raise SystemExit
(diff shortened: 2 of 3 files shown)


Lib/unittest/case.py
```diff
@@ -745,7 +745,11 @@ def __call__(self, *args, **kwds):
     def debug(self):
         """Run the test without collecting errors in a TestResult"""
         self.setUp()
-        getattr(self, self._testMethodName)()
+        testMethod = getattr(self, self._testMethodName)
+        if (not getattr(self.__class__, "__unittest_skip__", False) and
```
mangrisano, line 749: I think it's not necessary use the round brackets for the condition.

Lib/unittest/test/test_case.py
```diff
@@ -1777,6 +1777,31 @@ def test_something(self):
...
+
+        class Test3(unittest.TestCase):
+            test_something = nothing
+            tearDown = _skip
+
+        class Test4(unittest.TestCase):
+            def test_something(self):
+                self.addCleanup(_skip)
+
+        for klass in (Test1, Test2, Test3, Test4):
```
mangrisano, line 1801: What did you mean with "klass"?

lisroach: This is a common variable name used throughout this file (see line 1774 for example), it's a variable representing a class.

Lib/unittest/case.py
lisroach: Do you mean the parens around the if statement? It is not necessary, but encouraged by the style guide to wrap multi-line if statements in parens: https://www.python.org/dev/peps/pep-0008/#multiline-if-statements

lisroach:
How do I kick the CI to rerun their tests? I think these failures were transient.

asvetlov:
Close and Reopen -- as I did just now

lisroach:
Thanks @asvetlov!

mangrisano (approved):

Lib/unittest/case.py
potomak: Maybe it's out of the scope of this PR, but I'd like to suggest a small refactoring to make this condition a little bit clearer.

1. Define a new `_is_skipped(self, testMethod)` method:

```python
def _is_skipped(self, testMethod):
    return any(
        getattr(t, "__unittest_skip__", False)
        for t in (self.__class__, testMethod)
    )
```

2. Update the new condition to `not self._is_skipped(testMethod)` and the existing condition around line 665 to `self._is_skipped(testMethod)`

If it doesn't make sense to include this small update in this PR, but you think it makes sense in general, I'd be happy to work on it.

asfaltboy:
It seems that `TestCase.run()` implementation uses the `testPartExecutor ` context manager. It does so to wrap exceptions (and skips) raised in either setUp, the test method or tearDown calls:

https://github.com/python/cpython/blob/e4c431ecf50def40eb93c3969c1e4eeaf7bf32f1/Lib/unittest/case.py#L654-L662

With this in mind, and [considering the docs](https://docs.python.org/3/library/unittest.html#skipping-tests-and-expected-failures) would it make more sense to wrap in try/except like this:

```python
def debug(self):
    """Run the test without collecting errors in a TestResult"""
    try:
        self.setUp()
        getattr(self, self._testMethodName)()
        self.tearDown()
    except SkipError:
        self.success = False
        pass
    while self._cleanups:
        function, args, kwargs = self._cleanups.pop(-1)
        function(*args, **kwargs)
```

Note: users and developers of pytest have been struggling with understanding the `TestCase.debug()` interface and use case for a while (with regards to launching postmortem, teardown and skipping). Here's one of the relevant ones: https://github.com/pytest-dev/pytest-django/issues/772

serhiy-storchaka (commented):
`setUp()` should not be run for skipped tests. Perhaps a `SkipTest` should be raised for skipped tests.

If `SkipTest` is raised it the test method, it should be ignored. If it is raised in `setUp()` -- I am not sure whether `tearDown()` and `doCleanup()` should still be called.

iritkatriel:
https://bugs.python.org/issue36674 is closed. What is the status of this PR?

iritkatriel:
Closing as this seems abandoned (see my previous comment).

PR #13182, opened 2019-05-08
bpo-36869: Avoid warning of unused variables

eamanu:
When run ```./configure && make -j4``` there are two warnings on
Object/dictobject.c file about ix and hash variable are not used.

Zachary Ware make me note that when _PyObject_ASSERT is call and NDEBUG
is defined that expand to ((void)0).

So this PR add a #ifndef check to avoid the warnings.

https://bugs.python.org/issue36869

### DIFF
--- a/Objects/dictobject.c
+++ b/Objects/dictobject.c
@@ -459,6 +459,7 @@ static PyObject *empty_values[1] = { NULL };
 int
 _PyDict_CheckConsistency(PyObject *op, int check_content)
 {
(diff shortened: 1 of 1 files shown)


methane:
Note that this function does nothing when NDEBUG is defined.
It doesn't make sense adding `#ifndef NDEBUG` for specific part.

Add it in top of the function, and add `#endif` just before `return 1`.

eamanu:
I will open a issue on bpo

PR #13185, opened 2019-05-08
bpo-36974: PEP 590

jdemeyer:
Implementation of PEP 590 by Mark Shannon and Jeroen Demeyer.

CC @encukou @markshannon 


jdemeyer:
Rebased to latest master, needs review.

encukou:
I went through the code, and made a few small changes. Do they look OK to you?

encukou (approved):

jdemeyer:
I don't agree with 7536e69231fc94ab9c43b9ea84645c0745b45c86 for two reasons:

1. I tried to use `Py_EnterRecursiveCall` exactly where the old code also did that. The old implementation of `_PyObject_FastCallDict` didn't call `Py_EnterRecursiveCall` in the fastcall case and I don't see why it should do that now.

2. I want that `_PyObject_FastCallDict` behaves exactly the same as `_PyObject_Vectorcall` except for the way for keyword arguments are passed. And `_PyObject_Vectorcall` doesn't call `Py_EnterRecursiveCall`.

Lib/test/test_capi.py
```diff
@@ -479,6 +484,47 @@ class MethodDescriptorHeap(_testcapi.MethodDescriptorBase):
...
+                 ([].append, (0,), {}, None),
+                 (sum, ([36],), {"start":6}, 42),
+                 (testfunction, (42,), {}, 42),
+                 (testfunction_kw, (42,), {"kw":None}, 42)]
+
+        from _testcapi import pyobject_vectorcall, pyvectorcall_call
+        from types import MethodType
+        from functools import partial
+
+        def vectorcall(func, args, kwargs=None):
```
jdemeyer: If you want to set a default for `kwargs`, it should be `kwargs={}`.

Lib/test/test_capi.py
```diff
@@ -479,6 +484,47 @@ class MethodDescriptorHeap(_testcapi.MethodDescriptorBase):
             pass
         self.assertFalse(MethodDescriptorHeap.__flags__ & Py_TPFLAGS_METHOD_DESCRIPTOR)
 
+    def test_vectorcall(self):
+        # Test a bunch of different ways to call objects:
+        # 1. normal call
+        # 2. vectorcall
```
jdemeyer: It would be good to fix this comment, as you're adding a new way of testing the call.

jdemeyer:
No further comments.

encukou:
I'm not entirely convinced on the recursivecall, but it's probably OK to leave out.
I'm out for a couple of hours. Could you push a revert and your suggested changes to get the CI going?

On May 29, 2019 5:17:12 PM GMT+02:00, Jeroen Demeyer <notifications@github.com> wrote:
>No further comments.
>
>-- 
>You are receiving this because you were assigned.
>Reply to this email directly or view it on GitHub:
>https://github.com/python/cpython/pull/13185#issuecomment-496981190

-- 
Sent from my Android device with K-9 Mail. Please excuse my brevity.

encukou:
All buildbots are green now.

Thank you, @jdemeyer and @markshannon, for putting this together!
There are follow-up PRs linked from the issue. Not sure if I can get them into Python 3.8, unfortunately – but the main part, the external API, is there.

Docs should come in early beta.

jdemeyer:
Petr, thank you for reviewing and merging and in particular fixing the buildbot failure.

About follow-ups:

- #13498 would be good to merge, as it's part of the specification.
- #13460 is pretty trivial, so why not? It speeds up `functools.partial` (which is often used as a building block for other classes).
- #13464 should get a decision either way (merge or close), it makes no sense to wait until 3.9

The others are less important and can wait.

PR #13187, opened 2019-05-08
bpo-36829: Add sys.unraisablehook()

vstinner:
Add new sys.unraisablehook() function which can be overriden to
control how "unraisable exceptions" are handled. It is called when an
exception has occurred but there is no way for Python to handle it.
For example, when a destructor raises an exception or during garbage
collection (gc.collect()).


serhiy-storchaka:
The original proposition was about conditionally aborting the current process. Raising an exception (in particularly by `os.exit()`) does not work. Should not we add os.abort()? Or there are better alternatives for post-mortem investigation?

Python/errors.c
```diff
@@ -944,90 +944,132 @@ PyErr_NewExceptionWithDoc(const char *name, const char *doc,
...
+    _Py_IDENTIFIER(unraisablehook);
+    PyObject *hook = _PySys_GetObjectId(&PyId_unraisablehook);
+    PyObject *res;
+    if (hook != NULL && hook != Py_None) {
+        PyObject *args[4];
+        args[0] = (exc_type ? exc_type : Py_None);
+        args[1] = (exc_value ? exc_value : Py_None);
+        args[2] = (exc_tb ? exc_tb : Py_None);
+        args[3] = (obj ? obj : Py_None);
+        res = _PyObject_FastCall(hook, args, Py_ARRAY_LENGTH(args));
```
serhiy-storchaka: Would not be better to call `_PyErr_WriteUnraisable()` if the hook set an exception instead of silently ignore it?

Python/sysmodule.c
```diff
@@ -374,6 +374,27 @@ sys_exc_info_impl(PyObject *module)
...
+
+Print an unraisable exception.
+[clinic start generated code]*/
+
+static PyObject *
+sys_unraisablehook_impl(PyObject *module, PyObject *exc_type,
+                        PyObject *exc_value, PyObject *exc_tb, PyObject *obj)
+/*[clinic end generated code: output=f83f725d576ea3d3 input=2aba571133d31a0f]*/
+{
+    return _PyErr_WriteUnraisable(exc_type, exc_value, exc_tb, obj);
```
serhiy-storchaka: Why not set `sys.unraisablehook` to None by default? Or not set it at all? What is the use case for exposing it at Python level?

Python/errors.c
vstinner: Hum, it's hard to know if the hook started to log something or not.

Python/sysmodule.c
vstinner: I tried to use a design similar to sys.breakpointhook and sys.excepthook. It seems useful to be able to call `sys.__unraisablehook__` from a custom hook for example.

Python/errors.c
vstinner: Maybe we could do something like logging: log a special message when the custom hook fails? We should just prevent recursive calls 😁

Python/sysmodule.c
serhiy-storchaka: There is a difference between `sys.breakpointhook` and `sys.unraisablehook`. The former is only called when you explicitly call `breakpoint()`. The latter can be called implicitly at arbitrary place and time. It is out of your control. `sys.excepthook` is also called implicitly, but at known circumstances -- when an uncaught exception achieves the top level. I afraid that calling `sys.unraisablehook` at arbitrary place may be unsafe.

Python/errors.c
serhiy-storchaka: There is no recursion. I propose to call `_PyErr_WriteUnraisable()`, not `PyErr_WriteUnraisable()`. And pass `hook` as the obj argument.

Python/errors.c
```diff
@@ -944,90 +944,135 @@ PyErr_NewExceptionWithDoc(const char *name, const char *doc,
 }
 
 
-/* Call when an exception has occurred but there is no way for Python
-   to handle it.  Examples: exception in __del__ or during GC. */
-void
-PyErr_WriteUnraisable(PyObject *obj)
+PyObject*
+_PyErr_WriteUnraisable(PyObject *exc_type, PyObject *exc_value,
```
serhiy-storchaka: Make it `static`.

vstinner: It is called by sys_unraisablehook_impl() in sysmodule.c.

Python/errors.c
vstinner: PyErr_WriteUnraisable() now logs exception in custom unraisablehook using _PyErr_WriteUnraisable().

Python/sysmodule.c
vstinner: sys.unraisablehook takes (exc_type, exc_value, exc_tb, obj) arguments, it doesn't rely on the current exception. Why would it be called to call it?

vstinner:
I modified my PR to log exception if custom sys.unraisablehook itself raises an exception: see the new dedicated unit test.

Include/internal/pycore_pylifecycle.h
```diff
@@ -98,6 +98,12 @@ PyAPI_FUNC(_PyInitError) _Py_PreInitializeFromCoreConfig(
     const _PyCoreConfig *coreconfig,
     const _PyArgv *args);
 
+PyAPI_FUNC(PyObject*) _PyErr_WriteUnraisable(
```
serhiy-storchaka: Why it is exposed in the header? Should not it be the `static` function?

Python/errors.c
```diff
@@ -944,90 +944,167 @@ PyErr_NewExceptionWithDoc(const char *name, const char *doc,
...
+    if (file == NULL || file == Py_None) {
+        return 0;
+    }
 
-    if (obj) {
-        if (PyFile_WriteString("Exception ignored in: ", f) < 0)
-            goto done;
-        if (PyFile_WriteObject(obj, f, 0) < 0) {
+    if (hook_failed) {
+        if (PyFile_WriteString("sys.unraisablehook failed:", file) < 0) {
```
serhiy-storchaka: Why not show what exception was raised?

Python/errors.c
```diff
@@ -944,90 +944,167 @@ PyErr_NewExceptionWithDoc(const char *name, const char *doc,
...
+    if (hook != NULL && hook != Py_None) {
+        PyObject *args[4];
+        args[0] = (exc_type ? exc_type : Py_None);
+        args[1] = (exc_value ? exc_value : Py_None);
+        args[2] = (exc_tb ? exc_tb : Py_None);
+        args[3] = (obj ? obj : Py_None);
+        PyObject *res = _PyObject_FastCall(hook, args, Py_ARRAY_LENGTH(args));
+        Py_XDECREF(res);
+        if (res == NULL) {
+            unraisable_hook_failed();
```
serhiy-storchaka: I suggest to just fall back to the standard handler called below.
```c
if (hook != NULL && hook != Py_None) {
    PyErr_Fetch(&exc_type, &exc_value, &exc_tb);
    ... // call the hook
    Py_XDECREF(exc_type);
    Py_XDECREF(exc_value);
    Py_XDECREF(exc_tb);
    if (res != NULL) {
        goto done;
    }
    obj = hook;
}
write_unraisable(...);
done:
```

jdemeyer:
I'm not sure that it's safe to run arbitrary Python code any time that `PyErr_WriteUnraisable` is called. There might be occasions where `PyErr_WriteUnraisable` is called but where CPython is internally in an inconsistent state where calling Python code would crash the interpreter.

Include/internal/pycore_pylifecycle.h
vstinner: sys.unraisablehook() calls _PyErr_WriteUnraisable().

vstinner:
When a custom hook fails, the exception is logged. For example, test_custom_unraisablehook_fail() checks for "Exception: hook_func failed" in the output.

Another example:

```
import _testcapi
import sys

def hook_func(*args):
    raise Exception("hook_func failed")

sys.unraisablehook = hook_func
_testcapi.write_unraisable(ValueError(42), None)
```

Output:

```
sys.unraisablehook failed:
  File "x.py", line 5, in hook_func
    raise Exception("hook_func failed")
Exception: hook_func failed
```

Python/errors.c
vstinner: It is logged (see my comment below).

Python/errors.c
vstinner: I inlined unraisable_hook_failed() to make the fallback more explicit.

graingert:
If the hook fails, will it abort the process with a non-zero exit code?

vstinner:
> I'm not sure that it's safe to run arbitrary Python code any time that PyErr_WriteUnraisable is called. There might be occasions where PyErr_WriteUnraisable is called but where CPython is internally in an inconsistent state where calling Python code would crash the interpreter.

Do you have an example? I'm not aware of such issue.

Python must always remain consistent. Usually, Py_FatalError() is called when an inconstancy is detected.

vstinner:
> If the hook fails, will it abort the process with a non-zero exit code?

No, the second new error is logged. Python should not be aborted when something goes wrong: it should be possible to continue the execution.

Another complete solution for https://bugs.python.org/issue36829 would be to add a new -X command line option to use a specific exit code if at least one "unraisable" exception is logged. So far, nobody showed an example where my hook cannot be used to fix https://bugs.python.org/issue36829

vstinner:
See https://bugs.python.org/issue36829#msg342000 for PyErr_WriteUnraisable() called very late Python shutdown.

Python/sysmodule.c
vstinner: Sorry: Why would it be **unsafe** to call it?

serhiy-storchaka: For example because it can be called deep in the C stack.

Python/errors.c
serhiy-storchaka: Did you forget to push your changes?

Python/errors.c
serhiy-storchaka: I often wish `PyErr_WriteUnraisable()` take an optional C string argument for describing the context. In many case there is no appropriate callable which can be passed as obj, sou you pass NULL and lost the context.

Python/errors.c
vstinner: I didn't write exactly what you proposed, but what I wrote is basically the same without goto. I dislike goto and "obj = hook" here. I prefer a more explicit fallback.

vstinner: Well. I modified my PR to write an explicit fallback as you asked me for :-)

Python/errors.c
vstinner: I added a new _PyErr_WriteUnraisableMsg() function exactly for that. I agree that PyErr_WriteUnraisable(NULL) is hard to debug. I only used it twice to show how it can be used. We can bikeshed later on converting existing PyErr_WriteUnraisable(NULL) to _PyErr_WriteUnraisableMsg() :-)

vstinner:
I added "msg" to sys.unraisablehook.

Ok, so I rebased my PR on top of master to reduce the risk of conflicts, I squashed my commits to be able to update the main commit message, and I added a new "msg" parameter to sys.unraisablehook.

The new "msg" parameter allows to pass an arbitrary error message rather than the hardcoded "Exception ignored in: ". I added _PyErr_WriteUnraisableMsg() to log an unraisable exception with an error message.

I only used _PyErr_WriteUnraisableMsg() twice where it was appropriate to show how it can be used. If the PR is accepted, a following PR can be written to add a more accurate error message (than the hardcoded error message). I prefer to keep this PR as small as possible.

I also cleaned up the implementation and added more comments.

vstinner:
cc @methane @pablogsal

vstinner:
@serhiy-storchaka: Would you mind to review the updated PR? Do you like the new API?

vstinner:
I plan to merge this PR at the end of the week. I addressed all remarks/requests.

The only remaining complain about this PR is the fact that it doesn't allow to catch exceptions raised very late during Python shutdown: https://bugs.python.org/issue36829#msg342001

I confirm that it's a known limitation, but I am not comfortable with killing (SIGABRT) Python when PyErr_WriteUnraisable() is called late during Python finalization: there is no "easy" way to debug such issue. Only a low-level debugger like gdb can maybe provide some info about what happened. Or maybe not since Python finalization already destroyed too many things: modules, thread state, etc.

Python finalization code is very fragile and it's a work-in-progress for 5 years. I would love to see enhancements, but multiple attempts failed badly with adding new corner cases and triggering new crashes, and so had to be reverted.

My PR is a compromise for the most common cases. IMHO it's a way more generic solution than only allow to kill the process with SIGABRT. You are free to log "unraisable exceptions" into a file, raise a signal, send exceptions into a network socket, etc. Basically, whatever you want ;-)

jdemeyer:
> Do you have an example? I'm not aware of such issue.

I don't know any concrete example, it was just a general comment.

> Python must always remain consistent.

I meant temporary inconsistency, when you are in the middle of updating some data structure.

vstinner:
> I meant temporary inconsistency, when you are in the middle of updating some data structure.

Ok, I see what you mean. If we have such code, it's a bug that should be fixed.

The default hook implementation, current PyErr_WriteUnraisable() implementation, already executes "arbitrary" Python code:

* call repr(exception_value) which can be an arbitrary Python function
* get exception_type.__module__ attribute
* call sys.stderr.write() whereas sys.stderr can be a custom object
* displaying the traceback calls io.open(filename, "rb"), parse the encoding cookie, create a io.TextIOWrapper,  call fp.readline() and then fp.close()
* etc.

Moreover, the GC is not disabled and can be triggered anytime. So it's important that everything remains consistent when PyErr_WriteUnraisable() is called.

jdemeyer:
OK, I agree. Like I said, it was just a comment, not a complaint.

vstinner:
> OK, I agree. Like I said, it was just a comment, not a complaint.

Thanks for sharing your concerns. This issue is very tricky, I wouldn't be surprised if we discover a corner case tomorrow ;-) So having more people looking into the code and how it's used is very helpful!

vstinner:
New version of my API. sys.unraisablehook now gets a single argument which has 4 fields:

* exc_type
* exc_value
* exc_tb
* obj

It becomes possible to extend this object later to add new attribute without breaking the backward compatibility (existing hooks used in the wild).

I also reverted unrelated changes.

I removed _PyErr_WriteUnraisableMsg() and the "err_msg" parameter / field. I tried to write a PR as small as possible. Once this PR will be merged, I will work on a second PR to add a new err_msg field to the hook and add back _PyErr_WriteUnraisableMsg() function.

Python/errors.c
```diff
@@ -944,90 +944,246 @@ PyErr_NewExceptionWithDoc(const char *name, const char *doc,
...
+    if (write_unraisable_exc(exc_type, exc_value, exc_tb, obj) < 0) {
+        return NULL;
+    }
+    Py_RETURN_NONE;
+}
+
+
+/* Call sys.unraisablehook().
+
+   This functon can be used when an exception has occurred but there is no way
```
zooba:
```suggestion
   This function can be used when an exception has occurred but there is no way
```

Python/sysmodule.c
```diff
@@ -374,6 +374,23 @@ sys_exc_info_impl(PyObject *module)
 }
 
 
+/*[clinic input]
+sys.unraisablehook
+
+    args: object
+    /
+
+Print an unraisable exception.
```
zooba: Can you explain what `args` should be here? (Apparently people read these docs :) )

Python/errors.c
```diff
@@ -944,90 +944,246 @@ PyErr_NewExceptionWithDoc(const char *name, const char *doc,
...
-            goto done;
-        if (PyFile_WriteObject(obj, f, 0) < 0) {
+   Do nothing if sys.stderr attribute doesn't exist or is set to None. */
+static int
+write_unraisable_exc(PyObject *exc_type, PyObject *exc_value,
+                     PyObject *exc_tb, PyObject *obj)
+{
+    PyObject *file = _PySys_GetObjectId(&PyId_stderr);
+    if (file == NULL || file == Py_None) {
+        return 0;
```
zooba: `Py_XDECREF(file)`?

zooba (commented):
Minor comments here, I put my bigger thoughts on python-dev :)

Python/sysmodule.c
vstinner: Sure, done.

Python/errors.c
vstinner: _PySys_GetObjectId() returns a borrowed reference.

vstinner:
> _PySys_GetObjectId() returns a borrowed reference.

To be clear, I trust no one, especially myself: since I started to work on this PR, I'm running frequently "./python -m test test_sys -R 3:3" to make sure that nothing leaks :-) I added multiple tests in test_sys, and it happened multiple times that I introduced a giant leak because of an obvious bug in my code :-D

Python/errors.c
```diff
@@ -944,90 +944,246 @@ PyErr_NewExceptionWithDoc(const char *name, const char *doc,
...
+    assert(PyExceptionClass_Check(exc_type));
+    const char *className = PyExceptionClass_Name(exc_type);
     if (className != NULL) {
         const char *dot = strrchr(className, '.');
         if (dot != NULL)
             className = dot+1;
     }
 
-    moduleName = _PyObject_GetAttrId(t, &PyId___module__);
+    _Py_IDENTIFIER(__module__);
```
pablogsal, line 1063: Maybe move this to global scope to be aligned with Eric's efforts here: https://github.com/python/cpython/pull/13372/files

Misc/NEWS.d/next/Library/2019-05-08-12-51-37.bpo-36829.8enFMA.rst
```diff
@@ -0,0 +1,5 @@
+Add new :func:`sys.unraisablehook` function which can be overriden to
```
pablogsal: s/overriden/overridden/g

Python/errors.c
```diff
@@ -944,90 +944,246 @@ PyErr_NewExceptionWithDoc(const char *name, const char *doc,
...
+write_unraisable_exc(PyObject *exc_type, PyObject *exc_value,
+                     PyObject *exc_tb, PyObject *obj)
+{
+    PyObject *file = _PySys_GetObjectId(&PyId_stderr);
+    if (file == NULL || file == Py_None) {
+        return 0;
+    }
+
+    if (obj != NULL && obj != Py_None) {
+        if (PyFile_WriteString("Exception ignored in: ", file) < 0) {
```
pablogsal, line 1029: I may be wrong here but as `_PySys_GetObjectId` returns borrow references, it is possible that `file` is killed while you are calling `PyFile_WriteString` multiple times? Can the GIL be released in this block? Imagine a (very unlikely) situation in which someone is calling this in a daemon thread or something while the sys module is getting destroyed in the main module.

This may not be a problem, but I was thinking about this while I was reviewing this block. 

TLDR: Should we Py_INCREF(file)?

Doc/whatsnew/3.8.rst
```diff
@@ -478,6 +478,16 @@ and manipulating normal distributions of a random variable.
     [7.672102882379219, 12.000027119750287, 4.647488369766392]
 
 
+sys
+---
+
+Add new :func:`sys.unraisablehook` function which can be overriden to control
```
pablogsal: s/overriden/overridden/g

Doc/library/sys.rst
```diff
@@ -1487,6 +1488,23 @@ always available.
    is suppressed and only the exception type and value are printed.
 
 
+.. function:: unraisablehook(unraisable, /)
+
+   Display an unraisable exception.
```
pablogsal: Maybe `Handle an unraisable exception`?

Python/errors.c
vstinner: You're right, using a borrowed reference is a risk of getting a dangling pointer. I modified my PR to hold a strong reference.

Note: this bug isn't new and should maybe also be fixed in Python 2.7 and 3.7.

Python/errors.c
vstinner: I disagree with Eric, see my comments on his PR :-)

vstinner:
@pablogsal: I addressed your comments.

vstinner:
I squashed my commits, rebased them on master and fix a merge conflict.

Python/errors.c
vstinner: Sorry, I wanted to write: "I **disagree** with Eric, ..."

vstinner:
@serhiy-storchaka @pablogsal @njsmith: Would you mind to review my PR adding sys.unraisablehook?

I renamed UnraisableHookArgs fields to (exc_type, exc_value, exc_traceback, object). I also fixed a few more bugs in the implementation to fix a few more corner cases like exc_type == NULL.

I also completed the documentation.

vstinner:
I plan to merge this change next Wednesday (in 2 days). I didn't see any strong opinion to the API, and I'm not convinced by other proposed APIs.

Once this PR will be merged, I'm interested to work on follow-up (I plan to write 1 PR to bullet):

* Add an optional error message to UnraisableHookArgs to override default "Exception ignored in:"
* Try to normalize the exception
* Try to create a traceback object and attack it to the exception
* Add a helper function to test.support to catch unraisable exception and modify unit tests generating such unraisable exception
* Maybe also experiment to modify regrtest to log these unraisable exceptions to display them again in the summary at the end

auvipy (approved):

vstinner:
Thanks for reviews, I merged my PR!

vstinner:
Follow-up:

* PR #13487 backports enhancement and bugfix to Python 3.7
* PR #13488 adds _PyErr_WriteUnraisableMsg() and adds 'err_msg' field to sys.unraisablehook
* PR #13490 adds test.support.catch_unraisable_exception()

PR #13191, opened 2019-05-08
bpo-36851: Clean the frame stack if the execution ends with a return and the stack is not empty

pablogsal:
https://bugs.python.org/issue36851

When evaluating a frame object, it is possible to exit the execution with some variables in the stack. The frame deallocator (frame_dealloc) only cleans the stack if f_stacktop is not NULL, but this only happens for generators and when trace functions are set. The eval loop does this cleanup already if an exception is being raised, but not if a RETURN_VALUE is set.

In the PyconUS sprints, @DinoV  and I have been working on this and we have decided that the cleanest approach is shared the same goto label with the path for exception handling, effectively cleaning up

### DIFF
--- a/Lib/test/test_code.py
+++ b/Lib/test/test_code.py
@@ -130,6 +130,7 @@ import sys
 import threading
 import unittest
 import weakref
+import opcode
 try:
     import ctypes
 except ImportError:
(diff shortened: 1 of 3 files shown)


DinoV:
This looks good to me!

PR #13192, opened 2019-05-08
bpo-36816: Update the self-signed.pythontest.net cert

gpshead:
We updated the server, our testsuite must match.

https://bugs.python.org/issue36816

✈️ CLE -> DEN ✈️ #pycon2019


pitrou (approved):
+1

ericvsmith (approved):

PR #13193, opened 2019-05-08
[Update process.py]: allow user to choose daemonic processes or not

zuoxingdong:
It'll be very convenient to allow users to choose whether they want the processes are daemonic or not, by default `daemon=True`. Because some use cases might need to use ProcessPoolExecutor inside ProcessPoolExecutor.


the-knights-who-say-ni:

csabella:
Thank you for your interest in improving CPython. This is not a trivial change and would require an issue to be created on bugs.python.org. If you are interested in pursuing this, please open an issue on the bug tracker and reopen this pull request with the bpo number in the title as per the devguide. Also, this change would also need tests and documentation changes.  Thanks!

PR #13201, opened 2019-05-08
bpo-36831: Do not apply default namespace to unprefixed attributes in ElementPath

scoder:
Also provide better grouping of the tokenizer tests.

https://bugs.python.org/issue36831


Lib/xml/etree/ElementPath.py
```diff
@@ -72,24 +72,27 @@
 
 def xpath_tokenizer(pattern, namespaces=None):
     default_namespace = namespaces.get('') if namespaces else None
+    not_parsing_attribute = True
```
zooba: Personally I'd use `parsing_attribute` here (and `not parsing_attribute` below), just because it's easier to read. But the logic looks fine to me

zooba (approved):

PR #13202, opened 2019-05-08
bpo-36540: Documentation for PEP570 - Python positional only arguments

pablogsal:
https://bugs.python.org/issue36540

### DIFF
--- a/Doc/c-api/code.rst
+++ b/Doc/c-api/code.rst
@@ -33,7 +33,7 @@ bound into a function.
 
    Return the number of free variables in *co*.
 
-.. c:function:: PyCodeObject* PyCode_New(int argcount, int kwonlyargcount, int nlocals, int stacksize, int flags, PyObject *code, PyObject *consts, PyObject *names, PyObject *varnames, PyObject *freevars, PyObject *cellvars, PyObject *filename, PyObject *name, int firstlineno, PyObject *lnotab)
+.. c:function:: PyCodeObject* PyCode_New(int argcount, int posonlyargcount, int kwonlyargcount, int nlocals, int stacksize, int flags, PyObject *code, PyObject *consts, PyObject *names, PyObject *varnames, PyObject *freevars, PyObject *cellvars, PyObject *filename, PyObject *name, int firstlineno, PyObject *lnotab)
 
    Return a new code object.  If you need a dummy code object to
    create a frame, use :c:func:`PyCode_NewEmpty` instead.  Calling
--- a/Doc/tutorial/controlflow.rst
+++ b/Doc/tutorial/controlflow.rst
@@ -519,6 +519,169 @@ and of course it would print:
 Note that the order in which the keyword arguments are printed is guaranteed
 to match the order in which they were provided in the function call.
 
+Special parameters
+------------------
+
+By default, arguments may be passed to a Python function either by position
+or explicitly by keyword. For readability and performance, it makes sense to
+restrict the way arguments can be passed so that a developer need only look
+at the function definition to determine if items are passed by position, by
+position or keyword, or by keyword.
+
+A function definition may look like::
+
+   def f(pos1, pos2, /, pos_or_kwd, *, kwd1, kwd2):
+         -----------    ----------     ----------
+           |             |                  |
+           |        Positional or keyword   |
+           |                                - Keyword only
+            -- Positional only
+
+where ``/`` and ``*`` are optional. If used, these symbols indicate the kind of
+parameter by how the arguments may be passed to the function:
+positional-only, positional-or-keyword, and keyword-only. Keyword parameters
+are also referred to as named parameters.
+
+-------------------------------
+Positional-or-Keyword Arguments
+-------------------------------
+
+If ``/`` and ``*`` are not present in the function definition, arguments may
+be passed to a function by position or by keyword.
+
+--------------------------
+Positional-Only Parameters
+--------------------------
+
+Looking at this in a bit more detail, it is possible to mark certain parameters
+as *positional-only*. If *positional-only*, the parameters' order matters, and
+the parameters cannot be passed by keyword. Positional-only parameters would
+be placed before a ``/`` (forward-slash). The ``/`` is used to logically
+separate the positional-only parameters from the rest of the parameters.
+If there is no ``/`` in the function definition, there are no positional-only
+parameters.
+
+Parameters following the ``/`` may be *positional-or-keyword* or *keyword-only*.
+
+----------------------
+Keyword-Only Arguments
+----------------------
+
+To mark parameters as *keyword-only*, indicating the parameters must be passed
+by keyword argument, place an ``*`` in the arguments list just before the first
+*keyword-only* parameter.
+
+-----------------
+Function Examples
+-----------------
+
+Consider the following example function definitions paying close attention to the
+markers ``/`` and ``*``::
+
+   >>> def standard_arg(arg):
+   ...     print(arg)
+   ...
+   >>> def pos_only_arg(arg, /):
+   ...     print(arg)
+   ...
+   >>> def kwd_only_arg(*, arg):
+   ...     print(arg)
+   ...
+   >>> def combined_example(pos_only, /, standard, *, kwd_only):
+   ...     print(pos_only, standard, kwd_only)
+
+
+The first function definition ``standard_arg``, the most familiar form,
+places no restrictions on the calling convention and arguments may be
+passed by position or keyword::
+
+   >>> standard_arg(2)
+   2
+
+   >>> standard_arg(arg=2)
+   2
+
+The second function ``pos_only_arg`` is restricted to only use positional
+parameters as there is a ``/`` in the function definition::
+
+   >>> pos_only_arg(1)
+   1
+
+   >>> pos_only_arg(arg=1)
+   Traceback (most recent call last):
+     File "<stdin>", line 1, in <module>
+   TypeError: pos_only_arg() got an unexpected keyword argument 'arg'
+
+The third function ``kwd_only_args`` only allows keyword arguments as indicated
+by a ``*`` in the function definition::
+
+   >>> kwd_only_arg(3)
+   Traceback (most recent call last):
+     File "<stdin>", line 1, in <module>
+   TypeError: kwd_only_arg() takes 0 positional arguments but 1 was given
+
+   >>> kwd_only_arg(arg=3)
+   3
+
+And the last uses all three calling conventions in the same function
+definition::
+
+   >>> combined_example(1, 2, 3)
+   Traceback (most recent call last):
+     File "<stdin>", line 1, in <module>
+   TypeError: combined_example() takes 2 positional arguments but 3 were given
+
+   >>> combined_example(1, 2, kwd_only=3)
+   1 2 3
+
+   >>> combined_example(1, standard=2, kwd_only=3)
+   1 2 3
+
+   >>> combined_example(pos_only=1, standard=2, kwd_only=3)
+   Traceback (most recent call last):
+     File "<stdin>", line 1, in <module>
+   TypeError: combined_example() got an unexpected keyword argument 'pos_only'
+
+
+Finally, consider this function definition::
+
+    def foo(name, **kwds):
+        return 'name' in kwds
+
+There is no possible call that will make it return ``True`` as the keyword 'name'
+will always to bind to the first parameter. For example::
+
+    >>> foo(1, **{'name': 2})
+    Traceback (most recent call last):
+      File "<stdin>", line 1, in <module>
+    TypeError: foo() got multiple values for argument 'name'
+    >>>
+
+But using ``/`` (positional only arguments), this is possible::
+
+    def foo(name, /, **kwds):
+        return 'name' in kwds
+    >>> foo(1, **{'name': 2})
+    True
+
+In other words, the names of positional-only parameters can be used in
+``**kwds`` without ambiguity.
+
+-----
+Recap
+-----
+
+The use case will determine which parameters to use in the function definition::
+
+   def f(pos1, pos2, /, pos_or_kwd, *, kwd1, kwd2):
+
+As guidance:
+
+* Use positional-only if names do not matter or have no meaning, and there are
+  only a few arguments which will always be passed in the same order or in case
+  when you need arbitrary keyword parameters and some positional ones.
+* Use keyword-only when names have meaning and the function definition is
+  more understandable by being explicit with names.
 
 .. _tut-arbitraryargs:
 
--- a/Doc/tutorial/controlflow.rst
+++ b/Doc/tutorial/controlflow.rst
@@ -528,7 +528,9 @@ restrict the way arguments can be passed so that a developer need only look
 at the function definition to determine if items are passed by position, by
 position or keyword, or by keyword.
 
-A function definition may look like::
+A function definition may look like:
+
+.. code-block:: none
 
    def f(pos1, pos2, /, pos_or_kwd, *, kwd1, kwd2):
          -----------    ----------     ----------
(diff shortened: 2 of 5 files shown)


pablogsal:
CC: @pganssle

pablogsal:
CC: @pganssle @mariocj89

Doc/c-api/code.rst
```diff
@@ -33,7 +33,7 @@ bound into a function.
 
    Return the number of free variables in *co*.
 
-.. c:function:: PyCodeObject* PyCode_New(int argcount, int kwonlyargcount, int nlocals, int stacksize, int flags, PyObject *code, PyObject *consts, PyObject *names, PyObject *varnames, PyObject *freevars, PyObject *cellvars, PyObject *filename, PyObject *name, int firstlineno, PyObject *lnotab)
+.. c:function:: PyCodeObject* PyCode_New(int argcount, int posonlyargcount, int kwonlyargcount, int nlocals, int stacksize, int flags, PyObject *code, PyObject *consts, PyObject *names, PyObject *varnames, PyObject *freevars, PyObject *cellvars, PyObject *filename, PyObject *name, int firstlineno, PyObject *lnotab)
```
pganssle, line 36: I think you need a corresponding change in "Doc/data/refcounts.dat"

Doc/tutorial/controlflow.rst
```diff
@@ -519,6 +519,169 @@ and of course it would print:
...
+By default, arguments may be passed to a Python function either by position
+or explicitly by keyword. For readability and performance, it makes sense to
+restrict the way arguments can be passed so that a developer need only look
+at the function definition to determine if items are passed by position, by
+position or keyword, or by keyword.
+
+A function definition may look like::
+
+   def f(pos1, pos2, /, pos_or_kwd, *, kwd1, kwd2):
+         -----------    ----------     ----------
```
pganssle, line 536: Distinct lack of [Box-drawing characters](https://en.wikipedia.org/wiki/Box-drawing_character), tut tut!

(For those who don't know that Pablo and I are friends, this is gentle ribbing, not me being a jerk 😉)

Doc/tutorial/controlflow.rst
```diff
@@ -519,6 +519,169 @@ and of course it would print:
...
+   ...     print(arg)
+   ...
+   >>> def kwd_only_arg(*, arg):
+   ...     print(arg)
+   ...
+   >>> def combined_example(pos_only, /, standard, *, kwd_only):
+   ...     print(pos_only, standard, kwd_only)
+
+
+The first function definition ``standard_arg``, the most familiar form,
```
pganssle: Hm, I would put a comma before ``standard_arg`` here, but I think it's acceptable to leave the comma out. Up to you.

Same comment with the others.

pganssle (changes requested):
Grammar and style looks pretty much fine. Mainly the refcounts.dat file needs to be updated.

I will say that in the example of the four functions, I would probably group the "explanation of the function" parts a bit differently, because the explanation of the syntax is separated from the demonstration of the syntax, which means you have to either remember it perfectly or scroll up.

I'd rather see something like:

"A standard function definition does not restrict how arguments are passed, so they can be passed either by position or keyword:

```python
>>> def standard_arg(arg):
 ...        print(arg)
>>> standard_arg(2)
2
>>> standard_arg(arg=2)
2
```

If an argument is specified to be positional only, it is an error to pass it by keyword:

```python
>>> def pos_only_arg(arg, /):
...        print(arg)
>>> pos_only_arg(2)
2
>>> pos_only_arg(arg=2)
...
```
"

I understand that it's a trade-off, though. In your version, the contrast between the different syntaxes is highlighted, since all the definitions are right next to each other, whereas in my version the syntax is kept close to the explanation.

Doc/tutorial/controlflow.rst
```diff
@@ -519,6 +519,169 @@ and of course it would print:
...
+
+If ``/`` and ``*`` are not present in the function definition, arguments may
+be passed to a function by position or by keyword.
+
+--------------------------
+Positional-Only Parameters
+--------------------------
+
+Looking at this in a bit more detail, it is possible to mark certain parameters
+as *positional-only*. If *positional-only*, the parameters' order matters, and
```
mariocj89, line 559: Not sure "the parameters' order matters" adds anything. It is more than when calling by keyword order does not matter. on pos or keyword order can also matter if called via positional.

Doc/tutorial/controlflow.rst
```diff
@@ -519,6 +519,169 @@ and of course it would print:
...
+the parameters cannot be passed by keyword. Positional-only parameters would
+be placed before a ``/`` (forward-slash). The ``/`` is used to logically
+separate the positional-only parameters from the rest of the parameters.
+If there is no ``/`` in the function definition, there are no positional-only
+parameters.
+
+Parameters following the ``/`` may be *positional-or-keyword* or *keyword-only*.
```
mariocj89, line 566: Parameters following the ``/`` are  *positional-or-keyword* unless specified otherwise by another marker.

Doc/tutorial/controlflow.rst
```diff
@@ -519,6 +519,169 @@ and of course it would print:
...
+The use case will determine which parameters to use in the function definition::
+
+   def f(pos1, pos2, /, pos_or_kwd, *, kwd1, kwd2):
+
+As guidance:
+
+* Use positional-only if names do not matter or have no meaning, and there are
+  only a few arguments which will always be passed in the same order or in case
+  when you need arbitrary keyword parameters and some positional ones.
+* Use keyword-only when names have meaning and the function definition is
```
mariocj89, line 687: more understandable by being explicit with names or you want to prevent users relying on the position of the argument being passed.

Doc/tutorial/controlflow.rst
```diff
@@ -519,6 +519,169 @@ and of course it would print:
...
+Recap
+-----
+
+The use case will determine which parameters to use in the function definition::
+
+   def f(pos1, pos2, /, pos_or_kwd, *, kwd1, kwd2):
+
+As guidance:
+
+* Use positional-only if names do not matter or have no meaning, and there are
```
mariocj89: I  would remove the "do not matter" and try to clarify it a little further. Maybe something like:

Use positional-only if you want the name of the parameters to not be available to the user. This is useful when parameter names have no real meaning, if you want to enforce the order of the arguments when the function is called or if you need to take arbitrary keywords.

mariocj89 (approved):
Looking great, some small optional nitpicks though I am not a native speaker :).

Doc/tutorial/controlflow.rst
pablogsal: This is for the case when we use the `range()` parameter order in the PEP. This part is also from the `How to teach this` from the PEP itself.

>It is more than when calling by keyword order does not matter. on pos or keyword order can also matter if called via positional.

I do not understand what you mean :S

Doc/tutorial/controlflow.rst
pablogsal: I prefer the original wording (it's from Carol's section in the PEP) :)

On the other hand, your sentence is a bit more concrete towards `positional-or-keyword`, but as there is only one other marker apart from "/", is not much more general to say "specified otherwise by another marker" IMHO

pablogsal:
@pganssle I think I prefer to highlight the contrast between the different syntaxes (and having the definition all together) more than how they are used (although is a very fine-grained detail). But I would like more opinions on this. @mariocj89 @willingc What do you think?

Doc/tutorial/controlflow.rst
mariocj89: I mean that if a function is declared with pos or keyword, the order matters in the same way as for postional only if called without keywords. Isn’t it?

Doc/tutorial/controlflow.rst
pablogsal: Sphinx does something fancy with those characters and breaks the whole diagram :(

pablogsal: @pganssle Do you know a code block in sphinx that does not break box-drawing characters?

Doc/tutorial/controlflow.rst
pablogsal: Yep, but here we are saying that if you use positional-only there is no way of calling with an altered order.

Doc/tutorial/controlflow.rst
```diff
@@ -519,6 +519,169 @@ and of course it would print:
...
+If ``/`` and ``*`` are not present in the function definition, arguments may
+be passed to a function by position or by keyword.
+
+--------------------------
+Positional-Only Parameters
+--------------------------
+
+Looking at this in a bit more detail, it is possible to mark certain parameters
+as *positional-only*. If *positional-only*, the parameters' order matters, and
+the parameters cannot be passed by keyword. Positional-only parameters would
```
pganssle: I think change "would be" to "are".

Doc/tutorial/controlflow.rst
pganssle: I don't see anything fancy, am I missing something?

box-chars

Looks like they work just fine (other than the fact that the vertical bars don't connect perfectly, but that's already true).

pablogsal: The problem happens when you change all of the characters. Try it out and meanwhile I will try to add a screenshot.

pganssle: box-chars

Still looks fine to me with all box characters.

pablogsal: Can you copy here the text? I may be doing something wrong....

pganssle:
```
.. code-block:: none

   def f(pos1, pos2, /, pos_or_kwd, *, kwd1, kwd2):
         ──┬───────     ─┬────────     ─────┬────
           │             │                  │
           │        Positional or keyword   │
           │                                └ Keyword only
           └── Positional only
```

This is what I have.

It renders weird for PDFs in older versions of Sphinx, but renders fine in both HTML and PDF with version 2.0.1.

pablogsal: This is how it renders for me (copy-pasting you code):

Screenshot from 2019-05-12 20-53-00

Notice that in the editor it renders nicely:

Screenshot from 2019-05-12 20-53-46

pganssle: Is this a problem with the font you are using in the browser? I tried this with two different browsers and it's rendering correctly.

pablogsal: I changed the font a couple of times and I have the same problem :S

pablogsal: On the other hand, firefox does it correctly. Can you try with Chrome?

Doc/tutorial/controlflow.rst
```diff
@@ -519,6 +519,174 @@ and of course it would print:
...
+
+The use case will determine which parameters to use in the function definition::
+
+   def f(pos1, pos2, /, pos_or_kwd, *, kwd1, kwd2):
+
+As guidance:
+
+* Use positional-only if you want the name of the parameters to not be
+  available to the user. This is useful when parameter names have no real
+  meaning, if you want to enforce the order of the arguments when the function
```
Carreau, line 684: `Or if you believe the name of the parameter may change in the future, and a change of parameter name can be considered a change in API` ? 

I've seen a couple of time `def foo(filename)` -> `def foo(filename_or_buffer)`, which people don't realise is an api change.

Carreau (approved):

Doc/tutorial/controlflow.rst
pganssle: @pablogsal It works fine on Chrome for me. Maybe we should leave it as is and open a new BPO issue about box characters not working correctly in documentation?

Seems like something that can probably be fixed with a webfont or something, but no need to hold up this PR on fixing it.

vstinner:
What is the status of this PR? Is it ready to be merged?

Doc/tutorial/controlflow.rst
willingc: Good point @mariocj89 if the sentence was read by itself. I'm going to resolve this since the next section treats the other marker `*` and the code block before has a full example with `/` and `*`.

Doc/tutorial/controlflow.rst
```diff
@@ -519,6 +519,174 @@ and of course it would print:
...
+   >>> combined_example(1, standard=2, kwd_only=3)
+   1 2 3
+
+   >>> combined_example(pos_only=1, standard=2, kwd_only=3)
+   Traceback (most recent call last):
+     File "<stdin>", line 1, in <module>
+   TypeError: combined_example() got an unexpected keyword argument 'pos_only'
+
+
+Finally, consider this function definition::
```
willingc:
```suggestion
Finally, consider this function definition which has a potential collision between the positional argument ``name``  and ``**kwds`` which has ``name`` as a key::
```

Doc/tutorial/controlflow.rst
```diff
@@ -519,6 +519,174 @@ and of course it would print:
...
+
+    def foo(name, **kwds):
+        return 'name' in kwds
+
+There is no possible call that will make it return ``True`` as the keyword 'name'
```
willingc:
```suggestion
There is no possible call that will make it return ``True`` as the keyword ``'name'``
```

Doc/tutorial/controlflow.rst
```diff
@@ -519,6 +519,174 @@ and of course it would print:
...
+will always to bind to the first parameter. For example::
+
+    >>> foo(1, **{'name': 2})
+    Traceback (most recent call last):
+      File "<stdin>", line 1, in <module>
+    TypeError: foo() got multiple values for argument 'name'
+    >>>
+
+But using ``/`` (positional only arguments), this is possible::
```
willingc:
```suggestion
But using ``/`` (positional only arguments), it is possible since it allows ``name`` as a positional argument and ``'name'`` as a key in the keyword arguments.
```

Doc/tutorial/controlflow.rst
```diff
@@ -519,6 +519,174 @@ and of course it would print:
...
+The use case will determine which parameters to use in the function definition::
+
+   def f(pos1, pos2, /, pos_or_kwd, *, kwd1, kwd2):
+
+As guidance:
+
+* Use positional-only if you want the name of the parameters to not be
+  available to the user. This is useful when parameter names have no real
+  meaning, if you want to enforce the order of the arguments when the function
+  is called or if you need to take some positinal parameters and arbitrary
```
willingc:
```suggestion
  is called or if you need to take some positional parameters and arbitrary
```

Doc/tutorial/controlflow.rst
willingc: Perhaps at end of these bullets, add a bullet. For an API, use positional only to prevent breaking API changes if the parameter's name changes.

willingc (approved):
A few minor corrections/comments. Thanks @pablogsal 👍

gvanrossum:
What is this waiting for? It's documentation -- it's easy to iterate on.

Doc/tutorial/controlflow.rst
```diff
@@ -519,6 +519,174 @@ and of course it would print:
...
+There is no possible call that will make it return ``True`` as the keyword ``'name'``
+will always to bind to the first parameter. For example::
+
+    >>> foo(1, **{'name': 2})
+    Traceback (most recent call last):
+      File "<stdin>", line 1, in <module>
+    TypeError: foo() got multiple values for argument 'name'
+    >>>
+
+But using ``/`` (positional only arguments), it is possible since it allows ``name`` as a positional argument and ``'name'`` as a key in the keyword arguments.
```
willingc:
```suggestion
But using ``/`` (positional only arguments), it is possible since it allows ``name`` as a positional argument and ``'name'`` as a key in the keyword arguments::
```

willingc: @pablogsal I think the suggestion will correct the build error.

pablogsal: Thanks, @willingc  I was accepting the suggestions from the phone (best present ever) and could not investigate manually :)

willingc:
Thanks all for moving this forward. If there are additional doc changes, please file another PR. :tada:

pablogsal:
Thank you all for the review and the helpful comments and insights! :tada:

PR #13169, opened 2019-05-07
bpo-27639: Correct return type for UserList slicing operation

mblahay:
Added logic to `__getitem__` magic method for UserList to ensure that the return
type matches that of self.

https://bugs.python.org/issue27639

PR #13203, opened 2019-05-08
[3.7] bpo-27639: Correct return type for UserList slicing operation

mblahay:
Added logic to __getitem__ magic method for UserList to ensure that the return
type matches that of self.

https://bugs.python.org/issue27639


mblahay:
This needs to be merged.

matrixise:
@markshannon because you merged the #13169 maybe you could merge this one.

mblahay:
@markshannon it has been another week. This is the same change as was made in 3.8. Please provide any feedback that you believe is necessary. I would like it if this didn't hang out there too much longer.

markshannon:
Sorry about the delay. LGTM.

PR #12755, opened 2019-04-10
bpo-30458: Disallow control chars in http URLs.

gpshead:
[bpo-36276](https://bugs.python.org/issue36276) and [bpo-30458](https://bugs.python.org/issue30458):

disallow control characters such as whitespace in urls put into the raw http protocol within http.client.

https://bugs.python.org/issue36276

https://bugs.python.org/issue30458

PR #13200, opened 2019-05-08
[3.5] bpo-36816: Update the self-signed.pythontest.net cert (GH-13192)

gpshead:
We updated the server, our testsuite must match.

https://bugs.python.org/issue36816

✈️ CLE -> DEN -> SFO ✈️ #pycon2019 #smoothjazz
(cherry picked from commit 6bd81734de0b73f1431880d6a75fb71bcbc65fa1)

Authored-by: Gregory P. Smith <greg@krypto.org>

https://bugs.python.org/issue36816

PR #13207, opened 2019-05-08
[3.5] bpo-30458: Disallow control chars in http URLs. (GH-12755)

hroncok:
Disallow control chars in http URLs in urllib.urlopen.  This addresses a potential security problem for applications that do not sanity check their URLs where http request headers could be injected.

Disable https related urllib tests on a build without ssl (GH-13032)
These tests require an SSL enabled build. Skip these tests when python is built without SSL to fix test failures.

Use http.client.InvalidURL instead of ValueError as the new error case's exception. (GH-13044)

Co-Authored-By: Miro Hrončok <miro@hroncok.cz>


vstinner:
AppVeyor failed with:

vstinner:
> It seems like these failures are related to https://bugs.python.org/issue36816: PR #13200 isn't merged yet.

Travis CI basically has the same failures and so it's likely the same issue.

vstinner (approved):
LGTM. Straightforward backport (cherry-pick). The patch went fine in all other branches.

hroncok:
Compared to 3.6, I've removed the f-strings.

larryhastings:
Thanks for the 3.5 love!

PR #13209, opened 2019-05-08
bpo-35924: Document a workaround for a curses bug

abadger:
The ncurses library has a bug which can provoke a segfault when a window
is resized.  The bug is provoked when a string with embedded newlines is
added via addstr().  This commit documents that problem in the curses
python library documentation and relates how to workaround the problem
in the calling code.

It will likely be years before every vendor upgrades their ncurses library even once upstream fixes the bug so it seems like a good idea to document the issue and a workaround.

Related to https://bugs.python.org/issue35924

Note that merging this change should not close the bug report as it only documents a workaround.

https://bugs.python.org/issue35924

### DIFF
--- a/Doc/library/curses.rst
+++ b/Doc/library/curses.rst
@@ -708,9 +708,16 @@ the following methods and attributes:
 
    .. note::
 
-      Writing outside the window, subwindow, or pad raises :exc:`curses.error`.
-      Attempting to write to the lower right corner of a window, subwindow,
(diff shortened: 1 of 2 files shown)


abadger:
The bug has now been fixed in upstream ncurses.  I think merging this documentation change would be sufficient to close the bug on our side.

lisroach (approved):
Thanks for all the work you did for this!

abadger:
I'm just happy to help :-)

PR #13211, opened 2019-05-08
bpo-36778: Avoid functools in encodings.cp65001

paulmon:
based on https://github.com/python/cpython/pull/13110

Replace functools.partial() with lambda to reduce the number of
imports at startup. Avoid the following imports at startup:

_collections
_functools
_heapq
_operator
collections
functools
heapq
itertools
keyword
operator
reprlib

### DIFF
--- a/Lib/encodings/cp65001.py
+++ b/Lib/encodings/cp65001.py
@@ -3,15 +3,17 @@ Code page 65001: Windows UTF-8 (CP_UTF8).
 """
 
 import codecs
-import functools
 
 if not hasattr(codecs, 'code_page_encode'):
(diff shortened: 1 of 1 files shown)


paulmon:
Still seeing one failure in my local test pass

FAIL: test_incremental_surrogatepass (test.test_codecs.CP65001Test)
  File "C:\master\pythond\lib\test\test_codecs.py", line 436, in test_incremental_surrogatepass
    self.assertEqual(dec.decode(data[i:], True), '\uD901')
AssertionError: '' != '\ud901'
+ \ud901

vstinner:
I merged my PR #13230 instead.

PR #13215, opened 2019-05-09
bpo-36862: Add VS2019 to env.bat

tonybaloney:
https://bugs.python.org/issue36862

### DIFF
--- a/PCbuild/env.bat
+++ b/PCbuild/env.bat
@@ -10,6 +10,7 @@ rem 'v110', 'v120' or 'v140') to the build script.
 echo Build environments: x86, amd64, x86_amd64
 echo.
 set VSTOOLS=%VS140COMNTOOLS%
+if "%VSTOOLS%"=="" set VSTOOLS=%VS160COMNTOOLS%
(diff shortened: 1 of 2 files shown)


tonybaloney:
Requesting review from @zooba and @zware

auvipy (approved):

PCbuild/env.bat
```diff
@@ -10,6 +10,7 @@ rem 'v110', 'v120' or 'v140') to the build script.
 echo Build environments: x86, amd64, x86_amd64
 echo.
 set VSTOOLS=%VS140COMNTOOLS%
+if "%VSTOOLS%"=="" set VSTOOLS=%VS160COMNTOOLS%
```
zooba, line 13: I don't think this is correct - we'll want some detection more like what's in `find_msbuild.bat` (using the `vswhere.exe` app - and if that's not there, then neither VS 2017 or VS 2019 are installed!)

tonybaloney:
So, I've discovered that the scripts don't work with prerelease versions of Visual Studio. You can use the `-prerelease` flag for vswhere so that would discover MSBuild, but as VS2019 is now released, this seems like an edge-case

PR #13217, opened 2019-05-09
bpo-12800: 'tarfile.StreamError: seeking backwards is not allowed' when extract symlink

websurfer5:
Extracting a symlink from a tarball using the tarfile
module in stream mode ('r|') fails with an exception when a file
or symlink of the same name already exists. The fix is to remove
the existing file or symlink before extraction. Tests are included.

https://bugs.python.org/issue12800

### DIFF
--- a/Lib/test/test_tarfile.py
+++ b/Lib/test/test_tarfile.py
@@ -2498,6 +2498,70 @@ class Bz2PartialReadTest(Bz2Test, unittest.TestCase):
         self._test_partial_input("r:bz2")
 
 
+class SymlinkOverwriteTest(unittest.TestCase):
+    # The testcase checks for correct overwriting of an
+    # existing symlink (issue #12800)
+
+    @support.skip_unless_symlink
+    def test_overwrite_symlink(self):
+        tmpdir = support.temp_cwd('overwrite_symlink')
+        source = 'source'
+        link = 'link'
+        try:
+            with open(source, 'wb'):
+                pass
+            os.symlink(source, link)
+            with tarfile.open(tmpname, 'w') as tar:
+                tar.add(source, arcname=os.path.basename(source))
+                tar.add(link, arcname=os.path.basename(link))
+
+            with open(tmpname, 'rb') as fileobj:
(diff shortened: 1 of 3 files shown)


auvipy (approved):

catlee:
Is there a way I can help get this landed? I just ran into this today and wasted a lot of time trying to figure out why my tar file opened in streaming mode was trying to seek backwards.

Lib/test/test_tarfile.py
```diff
@@ -2498,6 +2498,70 @@ def test_partial_input_bz2(self):
         self._test_partial_input("r:bz2")
 
 
+class SymlinkOverwriteTest(unittest.TestCase):
+    # The testcase checks for correct overwriting of an
+    # existing symlink (issue #12800)
+
+    @support.skip_unless_symlink
+    def test_overwrite_symlink(self):
+        tmpdir = support.temp_cwd('overwrite_symlink')
```
JulienPalard, line 2507: temp_cwd is a context manager, please use it in a with statement (also you're not using the variable at all which is strange).

Lib/test/test_tarfile.py
```diff
@@ -2498,6 +2498,70 @@ def test_partial_input_bz2(self):
...
+        finally:
+            try:
+                os.unlink(link)
+            except:
+                pass
+
+            try:
+                os.unlink(source)
+            except:
+                pass
```
JulienPalard, line 2530: If using a temporary directory for your test (temp_cwd in a context manager for example), you should not need this at all.

JulienPalard (changes requested):
The idea of removing the file / symlink before creating the symlink looks good to me, GNU tar behave the same way.

Please note that GNU tar also removes an empty directory if it exists (but keeps a non-empty directory, and give an error), we may want to copy this too?

catlee:
I've submitted a new PR with the suggested changes: https://github.com/python/cpython/pull/20972

taleinat:
Thanks for the PR, @websurfer5, but I'm closing this in favor of PR GH-21409, which restores the original fix for this issue which was lost in a bad merge.

PR #13219, opened 2019-05-09
bpo-36854: Move GC runtime state from _PyRuntimeState to PyInterpreterState.

ericsnowcurrently:
Note: as part of this PR, I've made sure that Python threads (created by the threading module) are marked as "deleted" earlier in finalization.  That marker is how joining threads get unblocked.  It was happening in `PyThreadState_Delete()` but now it will happen in `PyThreadState_Clear()`.  This is necessary to ensure that the callback gets called *before* much interpreter/runtime finalization happens.  (It could impact daemon threads, but we already can't guarantee behavior for those once finalization starts.) 

https://bugs.python.org/issue36854

### DIFF
--- a/Include/internal/pycore_object.h
+++ b/Include/internal/pycore_object.h
@@ -19,7 +19,7 @@ PyAPI_FUNC(int) _PyDict_CheckConsistency(PyObject *mp, int check_content);
  * NB: While the object is tracked by the collector, it must be safe to call the
  * ob_traverse method.
  *
- * Internal note: _PyRuntime.gc.generation0->_gc_prev doesn't have any bit flags
+ * Internal note: PyInterpreterState.gc.generation0->_gc_prev doesn't have any bit flags
  * because it's not object header.  So we don't use _PyGCHead_PREV() and
  * _PyGCHead_SET_PREV() for it to avoid unnecessary bitwise operations.
  *
@@ -38,11 +38,12 @@ static inline void _PyObject_GC_TRACK_impl(const char *filename, int lineno,
                           "object is in generation which is garbage collected",
                           filename, lineno, "_PyObject_GC_TRACK");
 
-    PyGC_Head *last = (PyGC_Head*)(_PyRuntime.gc.generation0->_gc_prev);
+    PyInterpreterState *interp = _PyInterpreterState_GET_UNSAFE();
+    PyGC_Head *last = (PyGC_Head*)(interp->gc.generation0->_gc_prev);
     _PyGCHead_SET_NEXT(last, gc);
     _PyGCHead_SET_PREV(gc, last);
-    _PyGCHead_SET_NEXT(gc, _PyRuntime.gc.generation0);
-    _PyRuntime.gc.generation0->_gc_prev = (uintptr_t)gc;
+    _PyGCHead_SET_NEXT(gc, interp->gc.generation0);
+    interp->gc.generation0->_gc_prev = (uintptr_t)gc;
 }
 
 #define _PyObject_GC_TRACK(op) \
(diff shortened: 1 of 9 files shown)


Python/pystate.c
```diff
@@ -782,6 +781,13 @@ PyThreadState_Clear(PyThreadState *tstate)
     Py_CLEAR(tstate->async_gen_finalizer);
 
     Py_CLEAR(tstate->context);
+
+    if (tstate->on_delete != NULL) {
+        // This will unblock any joining threads.
+        tstate->on_delete(tstate->on_delete_data);
+        tstate->on_delete = NULL;
+        tstate->on_delete_data = NULL;
```
vstinner, line 789: This change looks to be unrelated. If it's related, I would prefer to see it as a separated PR to prepare that one.

Same comment for other thread changes:

```
    // All threading module threads are marked as "done" later
    // in PyThreadState_Clear().
```

and

```
        // This will unblock any joining threads.
        // We also do this in PyThreadState_Clear(), but do it here to be sure.
```

Include/internal/pycore_object.h
```diff
@@ -38,11 +38,12 @@ static inline void _PyObject_GC_TRACK_impl(const char *filename, int lineno,
                           "object is in generation which is garbage collected",
                           filename, lineno, "_PyObject_GC_TRACK");
 
-    PyGC_Head *last = (PyGC_Head*)(_PyRuntime.gc.generation0->_gc_prev);
+    PyInterpreterState *interp = _PyInterpreterState_GET_UNSAFE();
```
vstinner, line 41: Would it be possible to use a "struct _gc_runtime_state *state" variable instead?

```
struct _gc_runtime_state *state = &_PyInterpreterState_GET_UNSAFE()->gc;
```

Modules/gcmodule.c
```diff
@@ -1257,7 +1257,8 @@ static PyObject *
 gc_enable_impl(PyObject *module)
 /*[clinic end generated code: output=45a427e9dce9155c input=81ac4940ca579707]*/
 {
-    _PyRuntime.gc.enabled = 1;
+    PyInterpreterState *interp = _PyInterpreterState_GET_UNSAFE();
```
vstinner, line 1260: Would it be possible to use a "struct _gc_runtime_state *state" variable instead?

```
struct _gc_runtime_state *state = &_PyInterpreterState_GET_UNSAFE()->gc;
```

If the line is too lone, maybe add a GET_STATE() macro which returns ``&_PyInterpreterState_GET_UNSAFE()->gc``.

Same comment for the whole file. I would prefer to avoid introducing "PyInterpreterState *interp = _PyInterpreterState_GET_UNSAFE(); " lines if it's only used to get the GC state.

Objects/object.c
```diff
@@ -2074,11 +2075,12 @@ _PyTrash_thread_deposit_object(PyObject *op)
 void
 _PyTrash_destroy_chain(void)
 {
-    while (_PyRuntime.gc.trash_delete_later) {
-        PyObject *op = _PyRuntime.gc.trash_delete_later;
+    PyInterpreterState *interp = _PyInterpreterState_GET_UNSAFE();
```
vstinner, line 2078: Again, a "state" variable would be more appropriate here.

Python/pylifecycle.c
```diff
@@ -511,6 +511,8 @@ pycore_create_interpreter(_PyRuntimeState *runtime,
     }
     *interp_p = interp;
 
+    _PyGC_Initialize(&interp->gc);
```
vstinner, line 514: Hum, maybe _PyGC_Initialize() should take interp argument instead, to be consistent with other _PyGC functions.

vstinner (changes requested):
I like what I see: -) But I have some comments on the actual implementation.

csabella:
@ericsnowcurrently, were you wanting to land this for 3.8?

vstinner:
I wrote a different change, PR #17287, that I just merged. So I close this old PR. Thanks anyway Eric for working on this, my work is partially based on yours ;-)

ericsnowcurrently:
Thanks for doing it!!!

vstinner:
> Thanks for doing it!!!

You're welcome.

PR #13221, opened 2019-05-09
bpo-36865: allow 'rt' mode in FileInput

natgaertner:
The FileInput class (https://github.com/python/cpython/blob/master/Lib/fileinput.py) delegates to the open function, gzip, or bz2. Each of these delegates supports the 'rt' mode flag, so FileInput can probably support it as well.

This might clear up some confusion as in https://bugs.python.org/issue5758, which, in accordance with the FileInput docs, assumes all files opened with the 'r' flag will be opened in text mode. For gzip and bz2 this is not the case, they will be opened in binary 'rb' mode.

This pull request adds the 'rt' mode to the allowed modes, expands the unit test for FileInput to test opening with all allowed modes, and makes the documentation in FileInput about the behavior of the 'r' flag more accurate.

https://bugs.python.org/issue36865

### DIFF
--- a/Doc/library/fileinput.rst
+++ b/Doc/library/fileinput.rst
@@ -27,8 +27,8 @@ replaced by ``sys.stdin`` and the optional arguments *mode* and *openhook*
 are ignored.  To specify an alternative list of filenames, pass it as the
 first argument to :func:`.input`.  A single file name is also allowed.
 
-All files are opened in text mode by default, but you can override this by
-specifying the *mode* parameter in the call to :func:`.input` or
+All files are opened in their default read mode by default, but you can override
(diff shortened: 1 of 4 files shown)


the-knights-who-say-ni:

auvipy (changes requested):
please re base

natgaertner:
@auvipy I rebased, but that appears to have made it so I can't see your suggested changes? I apologize, I may have misunderstood the github workflow for rebasing pull requests. Would it be possible to make your suggestions again? Sorry for the hassle.

methane:
As I said in b.p.o., we should open compressed files for "r" mode.
So no need to add 't' mode.

PR #12483, opened 2019-03-21
bpo-36368: Ignore SIGINT in SharedMemoryManager servers.

pierreglaser:
When starting a `SharedMemoryManager` in an interactive session, any `KeyboardInterrupt` event will be transmitted to the (sub)process running the shared memory server, which causes the `Manager` to be unusable thereafter.

This PR catches `SIGINT` signals in server processes.

https://bugs.python.org/issue36368

PR #13222, opened 2019-05-09
bpo-36867: Make semaphore_tracker track other system resources.

pierreglaser:
## Primary Purpose
This PR proposes to extend the `semaphore_tracker` range of action to other system resources.  
Currently in `python3.8`, if a process that created a  `shared_memory` object gets violently terminated (with `kill` for example), the memory segment will not be properly released until the next machine reboot. This will be problematic for long-running clusters used by many different users.

The primary reason of this PR is thus to extend the `semaphore_tracker` to track also  `shared_memory` segments. The `semaphore_tracker` module gets consequently renamed  `resource_tracker`.

## Extension to a potentially public API
Additionally, the design of the `resource_tracker` is more generic, and possibly suggests a new **public API** to enable the tracking other named-resources. For now, the private `_CLEANUP_FUNCS` dict references all the types that can possibly be tracked. 

To start tracking a new type, the `resource_tracker` only needs to know which cleanup routine to apply to each leaked instance of this type after the python session ends.  The public API would thus consist of a function like `resource_tracker.register_resource_type(resource_type, cleanup_func)`, where: 
* resource_type is a `string`
* `cleanup_func` must be a callable taking a single string as an argument. 

Under the hood, `register_resource_type` would populate the `CLEANUP_FUNCS` with the new (`type, cleanup_func`) record.

## Operating System Availability
To ease the review process of this PR.  the resource_tracker is now POSIX only. I however have a functional windows implementation that I will provide upon demand :)

pinging @pitrou  @ogrisel  @tomMoral 

@pablogsal this may interest you as well :)

https://bugs.python.org/issue36867

### DIFF
--- a/Lib/multiprocessing/resource_tracker.py
+++ b/Lib/multiprocessing/resource_tracker.py
@@ -33,7 +33,7 @@ _HAVE_SIGMASK = hasattr(signal, 'pthread_sigmask')
 _IGNORED_SIGNALS = (signal.SIGINT, signal.SIGTERM)
 
 _CLEANUP_FUNCS = {
-    'folder': shutil.rmtree,
+    'noop': lambda: None,
     'semaphore': _multiprocessing.sem_unlink,
     'shared_memory': _posixshmem.shm_unlink
 }
@@ -119,7 +119,7 @@ class ResourceTracker(object):
         try:
             # We cannot use send here as it calls ensure_running, creating
             # a cycle.
-            os.write(self._fd, b'PROBE:0:folder\n')
+            os.write(self._fd, b'PROBE:0:noop\n')
         except OSError:
             return False
         else:
--- a/Lib/multiprocessing/resource_tracker.py
+++ b/Lib/multiprocessing/resource_tracker.py
@@ -1,3 +1,6 @@
+###############################################################################
+# Server process to keep track of unlinked resources (like shared memory
+# segments, semaphores etc.) and clean them.
 #
 # On Unix we run a server process which keeps track of unlinked
 # resources. The server ignores SIGINT and SIGTERM and reads from a
@@ -5,11 +8,12 @@
 # end of the pipe, so we get EOF when all other processes have exited.
 # Then the server process unlinks any remaining resource names.
 #
-# This is important because the system only supports a limited number
-# of named resources, and they will not be automatically removed till
-# the next reboot.  Without this resource tracker process, "killall
-# python" would probably leave unlinked resources.
-#
+# This is important because there may be system limits for such resources: for
+# instance, the system only supports a limited number of named semaphores, and
+# shared-memory segments live in the RAM. If a python process leaks such a
+# resoucre, this resource will not be removed till the next reboot.  Without
+# this resource tracker process, "killall python" would probably leave unlinked
+# resources.
 
 import os
 import signal
--- a/Lib/multiprocessing/resource_tracker.py
+++ b/Lib/multiprocessing/resource_tracker.py
@@ -1,14 +1,14 @@
 #
 # On Unix we run a server process which keeps track of unlinked
-# semaphores. The server ignores SIGINT and SIGTERM and reads from a
+# resources. The server ignores SIGINT and SIGTERM and reads from a
 # pipe.  Every other process of the program has a copy of the writable
 # end of the pipe, so we get EOF when all other processes have exited.
-# Then the server process unlinks any remaining semaphore names.
+# Then the server process unlinks any remaining resource names.
 #
 # This is important because the system only supports a limited number
-# of named semaphores, and they will not be automatically removed till
-# the next reboot.  Without this semaphore tracker process, "killall
-# python" would probably leave unlinked semaphores.
+# of named resources, and they will not be automatically removed till
+# the next reboot.  Without this resource tracker process, "killall
+# python" would probably leave unlinked resources.
 #
 
 import os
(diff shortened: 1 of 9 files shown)


pitrou:
You may want to take a look at the CI failures.

pierreglaser:
Yes: turns out we don't want a `Finalizer` for `SharedMemory` ATM (as opposed to `SemLock`)

Lib/multiprocessing/resource_tracker.py
```diff
@@ -1,22 +1,28 @@
...
-# This is important because the system only supports a limited number
-# of named semaphores, and they will not be automatically removed till
-# the next reboot.  Without this semaphore tracker process, "killall
-# python" would probably leave unlinked semaphores.
+# Then the server process unlinks any remaining resource names.
 #
+# This is important because there may be system limits for such resources: for
+# instance, the system only supports a limited number of named semaphores, and
+# shared-memory segments live in the RAM. If a python process leaks such a
+# resoucre, this resource will not be removed till the next reboot.  Without
```
pitrou: Typo: "resoucre".

Lib/multiprocessing/resource_tracker.py
```diff
@@ -26,8 +32,14 @@
 _HAVE_SIGMASK = hasattr(signal, 'pthread_sigmask')
 _IGNORED_SIGNALS = (signal.SIGINT, signal.SIGTERM)
 
+_CLEANUP_FUNCS = {
+    'folder': shutil.rmtree,
```
pitrou: Is this one only used in tests?

pitrou (commented):
This PR looks mostly good to me. I have just one comment and a question.

Lib/multiprocessing/resource_tracker.py
pierreglaser: For now yes. This refers to the "public API" discussion that I evoked here https://github.com/python/cpython/pull/13222#issue-277494385. By default, folders (created using `os.mkdir` for example) are not tracked by the resource tracker. But some users may want to track the lifecycle of folders created during python sessions. If `resource_tracker.register` is made public, we can imagine users writing code such as:

```python
>>> import os
>>> from multiprocessing import resource_tracker
>>> fname = '/tmp/foo'
>>> os.mkdir(fname)
>>> resource_tracker.register(fname, "folder")
```

But the public API discussion should maybe not be done in this PR. This first PR is mostly for not leaking `shared_memory` objects, so I am fine with removing this line for the sake of issue separation.

pitrou: Yes, I think it's better to remove it. For the purposes of the PROBE command, you may add a "noop" resource type.

pierreglaser:
Some more comment about OS availability: On windows, it looks that even with no `resource_tracker`,  abruptly terminating process does not leak `shared_memory`. So as long as the `resource_tracker` only tracks `SemLocks` and `shared_memory` objects, it is useless for windows.

pitrou:
Indeed, it seems that shared memory on Windows is more cleverly implemented than on Unix.

pitrou:
cc @applio just in case he wants to confirm.

Lib/test/_test_multiprocessing.py
```diff
@@ -4803,57 +4810,92 @@ def test_preload_resources(self):
...
             mp.set_start_method("spawn")
-            lock1 = mp.Lock()
-            lock2 = mp.Lock()
-            os.write(%d, lock1._semlock.name.encode("ascii") + b"\\n")
-            os.write(%d, lock2._semlock.name.encode("ascii") + b"\\n")
+            rand = tempfile._RandomNameSequence()
+
+
+            def create_and_register_resource(rtype):
+                if rtype == "folder":
```
pitrou: No need to keep this `if` case around.

Lib/multiprocessing/resource_tracker.py
```diff
@@ -153,18 +165,24 @@ def main(fd):
...
-                        cache.add(name)
-                    elif cmd == b'UNREGISTER':
-                        cache.remove(name)
-                    elif cmd == b'PROBE':
+                    cmd, name, rtype = line.strip().decode('ascii').split(':')
+                    cleanup_func = _CLEANUP_FUNCS.get(rtype, None)
+                    if cleanup_func is None:
+                        raise ValueError('Cannot register for automatic '
+                                         'cleanup: unknown resource type {}'
+                                         .format(name, rtype))
```
pitrou: You're passing two arguments to format but there's only one placeholder in the format string.
By the way, using f-strings is less error-prone ;-)

Lib/multiprocessing/resource_tracker.py
```diff
@@ -1,22 +1,28 @@
...
 #
+# This is important because there may be system limits for such resources: for
+# instance, the system only supports a limited number of named semaphores, and
+# shared-memory segments live in the RAM. If a python process leaks such a
+# resource, this resource will not be removed till the next reboot.  Without
+# this resource tracker process, "killall python" would probably leave unlinked
+# resources.
 
 import os
+import shutil
```
pitrou: Can remove this?

pitrou (changes requested):
A couple more nits :-)

pierreglaser:
(just making the bot happy)

pitrou (approved):

pierreglaser:
Also @pitrou, before potentially merging, any opinion on pierreglaser#3? It adds shared-memory leaking tests for all platforms,  independently of whether the `resource_tracker` exist for this platform or not.

pierreglaser:
Note also that there is still work to do to properly release shared memory segments created with a `SharedMemoryManager` after an abrupt interruption of a python session, but this is independent of this PR (but not independent of https://github.com/python/cpython/pull/12483)

pitrou:
Thanks @pierreglaser !

pierreglaser:
Thanks @pitrou for the infallible reviews ;)

PR #13226, opened 2019-05-09
bpo-36676: Update "what's new" document.

scoder:
https://bugs.python.org/issue36676


eamanu (commented):
LGTM. just a little question, why don't generate two separate entries,
one for each issues?

scoder:
I don't mind either way, so I'd like to leave this to the final document rewrite. I kept them together because they both describe changes to the parser `target` object.

PR #13228, opened 2019-05-09
Asyncio test case

asvetlov:
A draft implementation for asyncio test case


Lib/unittest/test/test_async_case.py
```diff
@@ -0,0 +1,151 @@
...
+                self.assertEqual(events, [])
+                events.append('setUp')
+
+            async def test_func(self):
+                self.assertEqual(events, ['setUp'])
+                events.append('test')
+                self.addCleanup(self.on_cleanup)
+
+            async def tearDown(self):
+                self.assertEqual(events, ['setUp', 'test'])
```
1st1, line 24: One of the problems I can see happening here is multiple inheritance of tests cases with both "async" and "sync" versions of "tearDown". 
* How are you supposed to call `super().tearDown()` in such case?  
* Is this a serious problem?  I think it might be, essentially we're breaking the cooperative nature of `super()`.
* Should we prefix all "async" hooks, e.g. "async def asyncTearDown"? Or is that too ugly?

PR #13229, opened 2019-05-09
bpo-33303: ElementTree Comment text isn't escaped

websurfer5:
properly escape XML comment text, plus test

https://bugs.python.org/issue33303


Lib/xml/etree/ElementTree.py
```diff
@@ -897,7 +897,7 @@ def _serialize_xml(write, elem, qnames, namespaces,
     tag = elem.tag
     text = elem.text
     if tag is Comment:
-        write("<!--%s-->" % text)
+        write("<!--%s-->" % _escape_cdata(text))
     elif tag is ProcessingInstruction:
         write("<?%s?>" % text)
```
scoder, line 902: Please also include the text of processing instructions.

Misc/NEWS.d/next/Library/2019-05-09-22-36-41.bpo-33303.B7_POx.rst
```diff
@@ -0,0 +1 @@
+properly escape XML comment text
```
scoder: This is meant for humans to read, so please write a complete sentence, with a dot at the end.

Lib/xml/etree/ElementTree.py
websurfer5: It now escapes "?>" with "&gt;>".

Misc/NEWS.d/next/Library/2019-05-09-22-36-41.bpo-33303.B7_POx.rst
websurfer5: Done

Lib/xml/etree/ElementTree.py
scoder: Sorry for this, I was wrong. PIs are not escaped. Please revert the commit where you added it.

scoder:
Also, since this is a backwards incompatible change, this also needs to be mentioned in the "What's new" document: https://github.com/python/cpython/blob/master/Doc/whatsnew/3.8.rst#xml

scoder:
Thank you for your work and a perfect PR. I'm really sorry that I have to reject this feature.

PR #13232, opened 2019-05-10
[doc] Mention with and except clauses in globals()

andresdelfino:
Also changed "loop control target" to "target list", as that's the actual term used by for's grammar.

The long line is intentional. If the PR is approved I'll add the line breaks.

### DIFF
--- a/Doc/reference/simple_stmts.rst
+++ b/Doc/reference/simple_stmts.rst
@@ -944,7 +944,7 @@ Names listed in a :keyword:`global` statement must not be used in the same code
 block textually preceding that :keyword:`!global` statement.
 
 Names listed in a :keyword:`global` statement must not be defined as formal
-parameters or in a :keyword:`for` loop control target, :keyword:`class`
+parameters, or as targets in :keyword:`with` statements or :keyword:`except` clauses, or in a :keyword:`for` target list, :keyword:`class`
 definition, function definition, :keyword:`import` statement, or variable
 annotation.
 


auvipy (approved):

JulienPalard:
I just tested:
```python
>>> bar = 42
>>> def foo():
...     global bar
...     with open("/etc/hosts") as bar:
...         ...
... 
>>> foo()
>>> bar
<_io.TextIOWrapper name='/etc/hosts' mode='r' encoding='UTF-8'>
```

and it works, so I'd say the proposed modification is wrong?

andresdelfino:
@JulienPalard I believe for/class/def behave exactly the same, that part of the documentation refers to global written used *after* those statements. Compare this:

```
def for_global_before():
    global bar
    for bar in []:
        ...

def for_global_after():
    for bar in []:
        ...
    global bar

for_global_before()
for_global_after()

# ----

def class_global_before():
    global bar
    class bar:
        ...

def class_global_after():
    class bar:
        ...
    global bar

class_global_before()
class_global_after()

# ----

def def_global_before():
    global bar
    def bar():
        ...

def def_global_after():
    def bar():
        ...
    global bar

def_global_before()
def_global_after()
```
In all cases, _before works, but _after fails. So does with "with" and "except".

andresdelfino:
@JulienPalard I have verified for/class/def/with/except behave in the same way. These are the examples I have run:

for/class/def:

**for**

`
# Doesn't fail
def for_global_before():
    global bar
    for bar in []:
        ...

# Fails
def for_global_after():
    for bar in []:
        ...
    global bar
`

**class**

`
# Doesn't fail
def class_global_before():
    global bar
    class bar:
        ...

# Fails
def class_global_after():
    class bar:
        ...
    global bar
`

**def**

`
# Doesn't fail
def def_global_before():
    global bar
    def bar():
        ...

# Fails
def def_global_after():
    def bar():
        ...
    global bar
`

----------------------------

with/except:

**with**

`
# Doesn't fail
def with_global_before():
    global bar
    with open('/etc/hosts') as bar:
        ...

# Fails
def with_global_after():
    with open('/etc/hosts') as bar:
        ...
    global bar
`


**except**

`
# Doesn't fail
def except_global_before():
    global bar
    try:
        pass
    except Exception as bar:
        pass

# Fails
def except_global_after():
    try:
        pass
    except Exception as bar:
        pass
    global bar
`

andresdelfino:
@JulienPalard any chance you can take a look at this? If so, please see my previous comment.

JulienPalard:
Thanks @andresdelfino ! TIL global behave differently if placed after þ

andresdelfino:
@JulienPalard What's "þ"? My guess is that you wrote something that GitHub decided to parse :P

JulienPalard:
@andresdelfino i's m'y way to write :P 

(I do it using a compose key)

PR #13233, opened 2019-05-10
gh-81022: Supporting customization of float encoding in JSON

Lee-W:
Add an `encode_float` argument to JSONEncoder for supporting customization float encoding

* Issue: gh-81022

### DIFF
--- a/Lib/json/__init__.py
+++ b/Lib/json/__init__.py
@@ -157,6 +157,9 @@ def dump(obj, fp, *, skipkeys=False, ensure_ascii=True, check_circular=True,
     If *sort_keys* is true (default: ``False``), then the output of
     dictionaries will be sorted by key.
 
+    If specified, ``encode_float`` is a function that will be called when encoding
+    float into string. It should return a string.
+
     To use a custom ``JSONEncoder`` subclass (e.g. one that overrides the
     ``.default()`` method to serialize additional types), specify it with
     the ``cls`` kwarg; otherwise ``JSONEncoder`` is used.
@@ -219,6 +222,9 @@ def dumps(obj, *, skipkeys=False, ensure_ascii=True, check_circular=True,
 
     If *sort_keys* is true (default: ``False``), then the output of
     dictionaries will be sorted by key.
+
+    If specified, ``encode_float`` is a function that will be called when encoding
+    float into string. It should return a string.
 
     To use a custom ``JSONEncoder`` subclass (e.g. one that overrides the
     ``.default()`` method to serialize additional types), specify it with
--- a/Lib/json/__init__.py
+++ b/Lib/json/__init__.py
@@ -115,11 +115,12 @@ _default_encoder = JSONEncoder(
     indent=None,
     separators=None,
     default=None,
+    encode_float=None
 )
 
 def dump(obj, fp, *, skipkeys=False, ensure_ascii=True, check_circular=True,
         allow_nan=True, cls=None, indent=None, separators=None,
-        default=None, sort_keys=False, **kw):
+        default=None, sort_keys=False, encode_float=None, **kw):
     """Serialize ``obj`` as a JSON formatted stream to ``fp`` (a
     ``.write()``-supporting file-like object).
 
@@ -165,7 +166,7 @@ def dump(obj, fp, *, skipkeys=False, ensure_ascii=True, check_circular=True,
     if (not skipkeys and ensure_ascii and
         check_circular and allow_nan and
         cls is None and indent is None and separators is None and
-        default is None and not sort_keys and not kw):
+        default is None and not sort_keys and not encode_float and not kw):
         iterable = _default_encoder.iterencode(obj)
     else:
         if cls is None:
@@ -173,7 +174,8 @@ def dump(obj, fp, *, skipkeys=False, ensure_ascii=True, check_circular=True,
         iterable = cls(skipkeys=skipkeys, ensure_ascii=ensure_ascii,
             check_circular=check_circular, allow_nan=allow_nan, indent=indent,
             separators=separators,
-            default=default, sort_keys=sort_keys, **kw).iterencode(obj)
+            default=default, sort_keys=sort_keys, encode_float=encode_float,
+            **kw).iterencode(obj)
     # could accelerate with writelines in some versions of Python, at
     # a debuggability cost
     for chunk in iterable:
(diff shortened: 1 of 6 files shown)


the-knights-who-say-ni:

auvipy (approved):

Lib/json/__init__.py
```diff
@@ -115,11 +115,12 @@
     indent=None,
     separators=None,
     default=None,
+    encode_float=None
```
mitar: Probably comma at the end of the line?

mitar (commented):
I think also some tests would be useful?

Lib/json/__init__.py
Lee-W: I've added it. Thanks!

Lee-W:
@mitar I've added a test for the newly added `encode_float` argument.

cmnord:
Rather than add another argument to these functions, why not un-nest `floatstr` from `iterencode` so that it can be overridden?

mitar:
I think the idea is that one can also use C code-path for this? Why it is an argument for decoding and not a method?

cmnord:
@mitar I don't fully understand your comment. Could you clarify what you mean?

What I meant is that it would be nice to be able to do the following:

```python
import json
import numpy as np


class MyJSONEncoder(json.JSONEncoder):
    def floatstr(o, _repr=float.__repr__, _inf=INFINITY, _neginf=-INFINITY):
        if o != o:
            text = "None"
        elif o == _inf:
            text = "Infinity"
        elif o == _neginf:
            text = "-Infinity"
        else:
            return _repr(o)

        if not self.allow_nan:
            raise ValueError(
                "Out of range float values are not JSON compliant: " + repr(o)
            )

        return text


assert json.dumps(np.nan) == "None"
```

So rather than adding a new argument to many functions, we could achieve the same goal of custom float encoding this way. What do you mean by "C code-path" and "decoding and not a method"?

mitar:
* In [load](https://github.com/python/cpython/blob/9158a6f90dd028b13a8fecbd8f573e7dcc0ebff8/Lib/json/__init__.py#L282-L283) we already have `parse_float` and `parse_int`. Why do you think this is like that, and not simply a method on `cls`?
* I think the reason why is because allows a [fast code path](https://github.com/python/cpython/blob/9158a6f90dd028b13a8fecbd8f573e7dcc0ebff8/Modules/_json.c#L1021-L1026) in optimized C version of the parser.
* So, we should probably do a similar thing in encoding as well, no? So while your suggestion works, I think the idea is that it is faster if you just have to change those two. But we should probably measure that and see if truly is. But currently, the API here tries to match decoding. It might be confusing that for decoding you have function callbacks, but for encoding you have methods?

cmnord:
@mitar thanks for clarifying, I think I understand now. 🙂

mitar:
If you do have time, you could try to measure how much does this really benefit. You could try moving parsing of floats to a class in decoding. And see if that really slows down things. Personally I also do not like extra functions, especially if we already have a nice class to put methods on.

Lee-W:
@mitar Could you please explain more on what should I measure? It seems you already explain it. Thanks 🙂

mitar:
I would propose that you measure and compare two cases a) having methods on the class b) having functions directly provided.

Lee-W:
I've tested on `Python 3.8.0a4+` using the code below.
The result is as follows.

```text
Encode
Default:  0.19853970199999998
Argument:  0.31803161799999996
Class:  0.282146192
```

```python
import json
import timeit


INFINITY = json.encoder.INFINITY


class MyJSONEncoder(json.JSONEncoder):
    def floatstr(self, obj,  _repr=float.__repr__, _inf=INFINITY, _neginf=-INFINITY):
        if obj != obj:
            text = "None"
        elif obj == json.encoder.INFINITY:
            text = "Infinity"
        elif obj == -json.encoder.INFINITY:
            text = "-Infinity"
        else:
            return float.__repr__(obj)

        if not self.allow_nan:
            raise ValueError(
                "Out of range float values are not JSON compliant: " + repr(obj)
            )

        return text


def floatstr(obj, allow_nan=True, _repr=float.__repr__, _inf=INFINITY, _neginf=-INFINITY):
    if obj != obj:
        text = "None"
    elif obj == json.encoder.INFINITY:
        text = "Infinity"
    elif obj == -json.encoder.INFINITY:
        text = "-Infinity"
    else:
        return float.__repr__(obj)

    if not allow_nan:
        raise ValueError(
            "Out of range float values are not JSON compliant: " + repr(obj)
        )
    return text


obj = {
    'inf': float('inf'),
    '-inf': -float('-inf'),
    'nan': float('nan'),
    '2': 2.23,
}

print('Encode')
print('Default: ', timeit.timeit(lambda: json.dumps(obj), number=10000))
print('Argument: ', timeit.timeit(lambda: json.dumps(obj, encode_float=floatstr), number=10000))
print('Class: ', timeit.timeit(lambda: json.dumps(obj, cls=MyJSONEncoder), number=10000))
```

Lee-W:
@mitar Do I need to do other experiments on it? Or, would the above one be sufficient? 🙂

mitar:
I will leave to somebody else from the Python team to way in here.

mitar:
@Lee-W Please update the PR, there is now a merge conflict.

I really like this PR, could somebody from Python core team review/merge this?

mitar (approved):

Lee-W:
@mitar Thanks for reminding. I just fix the conflict.

Lib/json/encoder.py
```diff
@@ -220,42 +225,46 @@ def iterencode(self, o, _one_shot=False):
...
+                        text = '-Infinity'
+                    else:
+                        return _repr(o)
+
+                    if not allow_nan:
+                        raise ValueError(
+                            "Out of range float values are not JSON compliant: " +
+                            repr(o))
+
+                    return text
```
merwok, line 258: Minor comment: why move the floatstr function definition inline here?  it could stay as a regular method in the class, then the line below would be `self.encode_float = self.floatstr` and still work.

mRemyLynceus:
@Python's team : What's going on ?

Modules/_json.c
```diff
@@ -1240,7 +1242,7 @@ encoder_new(PyTypeObject *type, PyObject *args, PyObject *kwds)
     }
 
     return (PyObject *)s;
-}
+
```
arhadthedev: This removal seems to be done by mistake and to cause a pile of compilation errors. Should we return the brace back?

```suggestion
}
```

Lee-W: Thanks! I just updated it. Let's see how it works

merwok:
As I noted on the ticket, this should be discussed to reach agreement on the need and the shape of the feature first.

Lee-W:
Got it. I just wanted to resolve the previous conflict. As this is not yet agreed, let me close this PR

PR #13238, opened 2019-05-10
bpo-36878: Allow extra text after `# type: ignore` comments

msullivan:
In the parser, when using the type_comments=True option, recognize
a TYPE_IGNORE as anything containing `# type: ignore` followed by
a non-alphanumeric character. This is to allow ignores such as
`# type: ignore[E1000]`.

This doesn't provide the text in the syntax tree yet, which might be worth also doing.
But this is a sufficient start, since tools can pull it out themselves based on line numbers.

This may also want a PEP 484 update, but I think that might want to wait on some consensus/discussion about specific details.

### DIFF
--- a/Lib/test/test_type_comments.py
+++ b/Lib/test/test_type_comments.py
@@ -76,6 +76,10 @@ def foo():
 
 def bar():
     x = 1  # type: ignore
+
+def baz():
+    pass  # type: ignore[excuse]
+    x = 1  # type: ignore whatever
 """
 
 # Test for long-form type-comments in arguments.  A test function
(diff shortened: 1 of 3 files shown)


msullivan:
I didn't make a news entry because it is a small update to something that will already be new in 3.8, but I can add one if that's the right thing to do.

Lib/test/test_type_comments.py
```diff
@@ -76,6 +76,10 @@ def foo():
 
 def bar():
     x = 1  # type: ignore
+
+def baz():
+    pass  # type: ignore[excuse]
+    x = 1  # type: ignore whatever
```
ilevkivskyi, line 84: Not that this is super-important, but I would add two other typical scenarios:
```python
# type: ignore=excuse  <- Pylint style
# type: ignore [excuse]  <- just in case
```

ilevkivskyi (approved):
The idea (allow anything non-alphanumeric as first char after `ignore`) looks good to me. I have one random tests suggestion.

msullivan:
It occurs to me there is one awkward thing about how the current isalnum check operates, which is that things like `type: ignoreé` are accepted.

ilevkivskyi:
@msullivan Well, at least `# type: ignored` will be caught. I suppose there may be better functions, but I am not a big unicode expert, so I am happy with this simple solution. Plus, `type: ignoreé` can definitely be caught by a type checker.

ilevkivskyi:
@gvanrossum If you don't have any other suggestion, I am going to merge this soon.

PR #13239, opened 2019-05-10
bpo-32587: Make winreg.REG_MULTI_SZ support zero-length strings

ZackerySpytz:
https://bugs.python.org/issue32587

### DIFF
--- a/PC/winreg.c
+++ b/PC/winreg.c
@@ -765,13 +765,6 @@ Reg2Py(BYTE *retDataBuf, DWORD retDataSize, DWORD typ)
                 for (index = 0; index < s; index++)
                 {
                     size_t slen = wcsnlen(str[index], len);
-                    if (slen > INT_MAX) {
-                        PyErr_SetString(PyExc_OverflowError,
-                            "registry string is too long for a Python string");
-                        Py_DECREF(obData);
-                        PyMem_Free(str);
-                        return NULL;
-                    }
                     PyObject *uni = PyUnicode_FromWideChar(str[index], slen);
                     if (uni == NULL) {
                         Py_DECREF(obData);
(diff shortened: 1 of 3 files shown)


ZackerySpytz:
This PR currently lacks a unit test, mainly because using MoveFileExA() with MOVEFILE_DELAY_UNTIL_REBOOT requires administrative rights.  @eryksun, your insight would be appreciated here.
I've tested this patch locally with a modified version of the rtlqueryreg.py script on the BPO issue.
Also, this PR should fix the issue with wcslen() mentioned in https://bugs.python.org/issue9194#msg339465.

PC/winreg.c
```diff
@@ -749,21 +764,22 @@ Reg2Py(BYTE *retDataBuf, DWORD retDataSize, DWORD typ)
                 }
                 for (index = 0; index < s; index++)
                 {
-                    size_t len = wcslen(str[index]);
-                    if (len > INT_MAX) {
+                    size_t slen = wcsnlen(str[index], len);
```
eryksun, line 767: This might read past the end of `data`. For example, if the buffer is `L"spam\x00eggs"`, then `len` is 9 and `wcsnlen(str[1], 9)` will either include 5 extraneous characters or crash on an access violation.  Adjust for the remaining length using pointer arithmetic.

PC/winreg.c
```diff
@@ -749,21 +764,22 @@ Reg2Py(BYTE *retDataBuf, DWORD retDataSize, DWORD typ)
...
+                    if (slen > INT_MAX) {
```
eryksun: This can't happen.

PC/winreg.c
ZackerySpytz: I might be missing something, but I don't think that would happen: `len -= slen;` is performed on each iteration.

eryksun: Sorry, I overlooked the decrement at the end. But it should be `len -= slen + 1` because `slen` doesn't include the null.

PR #13241, opened 2019-05-11
bpo-36728: Remove PyEval_ReInitThreads() from C API

vstinner:
Remove the PyEval_ReInitThreads() function from the Python C API.
This function was exposed by mistake and should not be called
explicitly anyway.

Rename PyEval_ReInitThreads() to _PyEval_ReInitThreads() and add a
'runtime' parameter.

https://bugs.python.org/issue36728

### DIFF
--- a/Doc/whatsnew/3.8.rst
+++ b/Doc/whatsnew/3.8.rst
@@ -990,6 +990,11 @@ Changes in the Python API
 Changes in the C API
 --------------------
 
+* The :c:func:`PyEval_ReInitThreads` function has been removed from the C API.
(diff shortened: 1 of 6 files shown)


ZackerySpytz:
`PyEval_ReInitThreads()` should be removed from `Doc/c-api/init.rst`, no?

vstinner:
> PyEval_ReInitThreads() should be removed from Doc/c-api/init.rst, no?

Oh, it's documented? It's even documented in "High-level API"... Oh... I wrote PR #13282 to remove the doc.

PR #13247, opened 2019-05-11
bpo-36888: Add multiprocessing.parent_process()

tomMoral:
In the std lib, the `semaphore_tracker` and the `Manager` rely on daemonized processes that are launched with server like loops. The cleaning of such processes is made complicated by the fact that there is no cannonical way to check if the parent process is alive.

I propose to add in context a `parent_process` function that would give access to a `Process` object representing the parent process. This way, we could benefit from sentinel to improve the clean up of these processes that can be left dangling in case of hard stop from the main interpreter.

https://bugs.python.org/issue36888


Lib/test/_test_multiprocessing.py
```diff
@@ -269,6 +269,62 @@ def _test(cls, q, *args, **kwds):
...
+        p.start()
+        p.join()
+        parent_process = rconn.recv()
+        assert parent_process.pid == current_process().pid
+        assert parent_process.name == current_process().name
+
+    @classmethod
+    def _test_send_parent_process(cls, wconn):
+        from multiprocessing.process import parent_process
+        wconn.send(parent_process())
```
tomMoral: It seems a little weird to return this as it might try to return the `_sentinel` object. Why not return just the necessary info to be sure to avoid pickling issue?

Lib/test/_test_multiprocessing.py
```diff
@@ -269,6 +269,62 @@ def _test(cls, q, *args, **kwds):
...
+            self.skipTest('test not appropriate for {}'.format(self.TYPE))
+
+        # Launch a child process. Make it launch a grandchild process. Kill the
+        # child process and make sure that the grandchild notices the death of
+        # its parent (a.k.a the child process).
+        rconn, wconn = self.Pipe(duplex=False)
+        p = self.Process(
+            target=self._test_create_grandchild_process, args=(wconn, ))
+        p.start()
+        time.sleep(1)
```
tomMoral: This is unsafe as the starting time of the grandchild process might exceed 1s.
I would rather have something where you wait for the grandchild to notify that it is alive, then kill the parent. You could read on `rconn` a first time to get the first `"alive"`, then kill the `child` and read the status again which should be `"not alive"`.

Lib/test/_test_multiprocessing.py
```diff
@@ -269,6 +269,62 @@ def _test(cls, q, *args, **kwds):
...
+        p.terminate()
+        p.join()
+        parent_process_status_log = rconn.recv()
+        assert parent_process_status_log == ["alive", "not alive"]
+
+    @classmethod
+    def _test_create_grandchild_process(cls, wconn):
+        p = cls.Process(target=cls._test_report_parent_status, args=(wconn, ))
```
tomMoral, line 319: Don't you need to deamonize it to avoid that the grandchild get killed?

Lib/test/_test_multiprocessing.py
```diff
@@ -269,6 +269,62 @@ def _test(cls, q, *args, **kwds):
...
+    def test_parent_process_attributes(self):
+        if self.TYPE == "threads":
+            self.skipTest('test not appropriate for {}'.format(self.TYPE))
+        from multiprocessing.process import current_process
+        rconn, wconn = self.Pipe(duplex=False)
+        p = self.Process(target=self._test_send_parent_process, args=(wconn,))
+        p.start()
+        p.join()
+        parent_process = rconn.recv()
+        assert parent_process.pid == current_process().pid
```
tomMoral: Could you also check that it is equal to `os.getpid()`? This way, it check that there is no border effects on the `current_process`.

Lib/multiprocessing/spawn.py
```diff
@@ -100,25 +100,22 @@ def spawn_main(pipe_handle, parent_pid=None, tracker_fd=None):
 
         if parent_pid is not None:
             source_process = _winapi.OpenProcess(
-                _winapi.PROCESS_DUP_HANDLE, False, parent_pid)
+                _winapi.PROCESS_ALL_ACCESS, False, parent_pid)
```
tomMoral: Nice!

tomMoral (commented):
Some comments

pierreglaser:
@pitrou want to take a look? ;)

pierreglaser:
Same as in https://github.com/python/cpython/pull/13276, this code will not always be functional if several python processes manipulate the same `SharedMemoryManager`. I am not sure that this represents the canonical use-case of `SharedMemoryManager` though.

tomMoral:
Why is it not functional? The parent process handle should uniquely be open in the parent process no?

pierreglaser:
@tomMoral Pardon me, I realized my last message was not very clear.

`Manager` objects can establish connections to several python processes. It does not make so much sense to differentiate its parent process to other potential processes that are connected to it: especially, we do not want the `Manager` process to start cleaning up ressources if its parent dies, while other processes may still be connected to the `Manager`.

Thus technically, a `Manager` would need one sentinel per process it is connected to (and not only one for its parent process, as it is the case right now). This is particularly hard to implement for `SharedMemoryManager` objects, because when asked to create a new resource, a `SharedMemoryManager` does not return a proxy type, but an actual `SharedMemory` object. Thus, processes accessing the `SharedMemory` object created by the `Manager` do not need to communicate with the `Manager` whenever they want to modify it. 
Therefore, it seems hard, maybe impossible, for a `SharedMemoryManager` to track all the processes manipulating the objects it delivers.

pitrou:
Is this still WIP?

pierreglaser:
Not anymore.

Lib/multiprocessing/spawn.py
pitrou: Out of curiosity, is it possible to narrow the permissions? (for example `PROCESS_ALL_ACCESS + SYNCHRONIZE`?)

Lib/multiprocessing/popen_fork.py
```diff
@@ -70,7 +70,7 @@ def _launch(self, process_obj):
         if self.pid == 0:
             try:
                 os.close(parent_r)
-                code = process_obj._bootstrap()
+                code = process_obj._bootstrap(parent_sentinel=child_w)
```
pitrou: I don't understand. `child_w` will always be "ready" (writable) except if the pipe buffer is full, so how can you use it as a sentinel?

Lib/multiprocessing/spawn.py
```diff
@@ -100,34 +100,31 @@ def spawn_main(pipe_handle, parent_pid=None, tracker_fd=None):
...
-    exitcode = _main(fd)
+        fd = parent_sentinel = pipe_handle
+    exitcode = _main(fd, parent_sentinel)
     sys.exit(exitcode)
 
 
-def _main(fd):
-    with os.fdopen(fd, 'rb', closefd=True) as from_parent:
+def _main(fd, parent_sentinel):
+    with os.fdopen(fd, 'rb', closefd=False) as from_parent:
```
pitrou: Not very nice if `fd` and `parent_sentinel` are different fds... I think the caller should instead ensure those are different fds, possibly using `os.dup`.

Lib/multiprocessing/popen_forkserver.py
```diff
@@ -50,7 +50,7 @@ def _launch(self, process_obj):
 
         self.sentinel, w = forkserver.connect_to_new_process(self._fds)
         self.finalizer = util.Finalize(self, os.close, (self.sentinel,))
-        with open(w, 'wb', closefd=True) as f:
+        with open(w, 'wb', closefd=False) as f:
```
pitrou: Doesn't seem right. See comment about using `os.dup` instead where necessary.

Lib/test/_test_multiprocessing.py
```diff
@@ -269,6 +269,67 @@ def _test(cls, q, *args, **kwds):
...
+            self.skipTest('test not appropriate for {}'.format(self.TYPE))
+        from multiprocessing.process import current_process
+        rconn, wconn = self.Pipe(duplex=False)
+        p = self.Process(target=self._test_send_parent_process, args=(wconn,))
+        p.start()
+        p.join()
+        parent_pid, parent_name = rconn.recv()
+        self.assertEqual(parent_pid, current_process().pid)
+        self.assertEqual(parent_pid, os.getpid())
+        self.assertEqual(parent_name, current_process().name)
```
pitrou: Please also test `parent_process()` from the parent process.

Lib/test/_test_multiprocessing.py
```diff
@@ -269,6 +269,67 @@ def _test(cls, q, *args, **kwds):
             q.put(bytes(current.authkey))
             q.put(current.pid)
 
+    def test_parent_process_attributes(self):
+        if self.TYPE == "threads":
+            self.skipTest('test not appropriate for {}'.format(self.TYPE))
+        from multiprocessing.process import current_process
```
pitrou: Shouldn't this be at the top-level?

pitrou (changes requested):
I have reservations about the implementation. Also, a documentation addition will be required.

Lib/multiprocessing/popen_fork.py
pierreglaser: The idea is to use `_ParentProcess.is_alive`, that itself calls `mutliprocessing.connection.wait` on the sentinel. Does that answer your question?

Lib/multiprocessing/spawn.py
pierreglaser: Woops, agreed.

Lib/multiprocessing/popen_fork.py
pitrou: Not really, because it doesn't answer how it works concretely on the write end of the pipe.
```
>>> import os, select
>>> r, w = os.pipe()
>>> select.select([r], [r], [r])
^CTraceback (most recent call last):
KeyboardInterrupt
>>> select.select([w], [w], [w])
([], [4], [])
```

Lib/test/_test_multiprocessing.py
pierreglaser: OK (Should return `None`)

Lib/multiprocessing/popen_fork.py
pitrou: Though, at least on Linux:
```
>>> select.select([w], [], [w])
^CTraceback (most recent call last):
KeyboardInterrupt
>>> os.close(r)
>>> select.select([w], [], [w])
([4], [], [])
```

Lib/test/_test_multiprocessing.py
pierreglaser: I often see `import` statements inside tests, and I have to admit I do not know what rule is used to determine if an import should be top-level or not.

Lib/multiprocessing/spawn.py
pierreglaser: You mean `PROCESS_DUP_HANDLE` and `SYNCHRONIZE`? I tried a few days ago and I got an `access denied`.

pitrou: Ok. Perhaps @zooba has more insight about this.

Lib/multiprocessing/popen_fork.py
pierreglaser: Actually, I think we should use `child_r` and not `child_w`, as we do for `spawn`. But actually @tomMoral wrote this line, maybe I am missing something?

pierreglaser: I am not a frequent user of `select`, but it does seem odd to have a write end of a pipe into the `rlist` of `select` (which is the behavior we rely on using `multiprocessing.connection.wait`). But actually, I did not write this line, @tomMoral did. Maybe he has other insights on this. 
@pitrou would you be more confortable if we use the read end of a new dedicated pipe as a sentinel instead to mimicate the `sentinel` implementation of the parent process for `fork`?

pitrou: Yes, that would seem like a more strictly POSIX-compliant solution.

Lib/test/_test_multiprocessing.py
pitrou: Generally, it's better to put imports at the top level, except if the module might not be available, or if importing it may have undesirable side-effects.

Lib/multiprocessing/popen_spawn_posix.py
```diff
@@ -63,6 +63,6 @@ def _launch(self, process_obj):
         finally:
             if parent_r is not None:
                 self.finalizer = util.Finalize(self, os.close, (parent_r,))
-            for fd in (child_r, child_w, parent_w):
+            for fd in (child_r, child_w):
```
tomMoral, line 70: I think `parent_w` needs to be registered to be close with a `Finalize` too

Lib/test/_test_multiprocessing.py
pitrou: By the way `multiprocessing.current_process()` works, and perhaps `multiprocessing.parent_process()` should work, too.

Lib/multiprocessing/popen_fork.py
tomMoral: Yes indeed, it seems weird that way. I did not realize this was different compared to `spawn_posix`, where you have access to `child_r`. (But there seems to be a leak with `parent_w` with this implem, cf other comment).

That being said, the goal of the sentinel is simply to know if this pipe is still open or not. That is why it feels a waste to create a new pipe. But there might not be better answer.

Lib/multiprocessing/spawn.py
pierreglaser: Update: after a good night sleep, `PROCESS_DUP_HANDLE` | `SYNCHRONIZE` does work.

Lib/multiprocessing/popen_spawn_posix.py
pierreglaser: As is, I am not sure. `parent_r` needs a Finalizer because it is an attribute of the `Popen` object, which may be deleted later on at some point. OTOH `parent_w` will be gc'ed at the end of this function, but we want the underlying `fd` to remain open until this process ends.

pierreglaser:

Lib/multiprocessing/popen_forkserver.py
```diff
@@ -50,6 +50,7 @@ def _launch(self, process_obj):
 
         self.sentinel, w = forkserver.connect_to_new_process(self._fds)
         self.finalizer = util.Finalize(self, os.close, (self.sentinel,))
+        parent_sentinel = os.dup(w)
```
pitrou: This variable isn't used anywhere. Is it a bug?

Lib/multiprocessing/popen_spawn_win32.py
```diff
@@ -68,7 +68,7 @@ def __init__(self, process_obj):
         else:
             env = None
 
-        with open(wfd, 'wb', closefd=True) as to_child:
+        with open(wfd, 'wb', closefd=False) as to_child:
```
pitrou: Why is that? Is it a leftover from previous attempts?

Lib/multiprocessing/popen_spawn_posix.py
pitrou: I think @tomMoral is right. All running child processes are kept alive in the `_children` set. But when a process object is destroyed (which implies the child process has exited), we need to close the corresponding fds.

Lib/multiprocessing/process.py
```diff
@@ -337,6 +348,36 @@ def __reduce__(self):
...
+        self._parent_pid = None
+        self._popen = None
+        self._closed = False
+        self._sentinel = sentinel
+        self._config = {}
+
+    def close(self):
+        pass
+
+    def is_alive(self):
```
pitrou, line 368: Should `join()` be overriden as well? Or does it work already? I don't see it in the tests.

Lib/multiprocessing/popen_fork.py
```diff
@@ -66,15 +66,18 @@ def kill(self):
...
                 os.close(parent_r)
-                code = process_obj._bootstrap()
+                os.close(parent_w)
+                code = process_obj._bootstrap(parent_sentinel=child_r)
             finally:
                 os._exit(code)
         else:
             os.close(child_w)
+            os.close(child_r)
             self.finalizer = util.Finalize(self, os.close, (parent_r,))
```
pitrou: Same remark as in `popen_spawn_posix.py`: we need to add a finalizer that closes `parent_w`.

Lib/multiprocessing/process.py
```diff
@@ -337,6 +348,36 @@ def __reduce__(self):
...
+        self._identity = ()
+        self._name = name
+        self._pid = pid
+        self._parent_pid = None
+        self._popen = None
+        self._closed = False
+        self._sentinel = sentinel
+        self._config = {}
+
+    def close(self):
```
pitrou: It doesn't look necessary to override this method.

Lib/test/_test_multiprocessing.py
```diff
@@ -269,6 +269,69 @@ def _test(cls, q, *args, **kwds):
...
+        p.start()
+        time.sleep(100)
+
+    @classmethod
+    def _test_report_parent_status(cls, wconn):
+        from multiprocessing.process import parent_process
+        wconn.send("alive" if parent_process().is_alive() else "not alive")
+
+        start_time = time.monotonic()
+        while (parent_process().is_alive() and
```
pitrou: How about calling `join(timeout=5)`? With the sentinel, it seems like it should work.

pitrou (changes requested):
Thanks for the update. There seems to be a couple of issues still.

Lib/multiprocessing/popen_spawn_win32.py
pierreglaser: Good catch, sorry about that.

Lib/multiprocessing/popen_spawn_posix.py
pierreglaser: OK. The implication process object destroyed => child process has exited was what I was missing here. Thanks for the insight.

Lib/multiprocessing/popen_forkserver.py
pierreglaser: It leaves the write end opened as  `w`is closed at the end of the following context manager  (upon your request). But now that I fully understood https://github.com/python/cpython/pull/13247#discussion_r285666739, I guess it is better to have it as a private attribute of the process object and register a Finalizer.

Lib/multiprocessing/popen_forkserver.py
```diff
@@ -49,7 +49,11 @@ def _launch(self, process_obj):
             set_spawning_popen(None)
 
         self.sentinel, w = forkserver.connect_to_new_process(self._fds)
+        # Keep a duplicate of the data pipe's write end as a sentinel of the
+        # parent process used by the child process.
+        self._parent_w = os.dup(w)
         self.finalizer = util.Finalize(self, os.close, (self.sentinel,))
+        self.finalizer = util.Finalize(self, os.close, (self._parent_w,))
```
pierreglaser: `_parent_w` is not the best name, but I did not find convincing, short alternatives. I thought `_child_sentinel` was somewhat unclear. Any ideas anyone?

Lib/multiprocessing/popen_spawn_posix.py
pierreglaser: Not sure why the `parent_r is None` exists, but I did the same for `parent_w`.

Lib/multiprocessing/popen_fork.py
```diff
@@ -66,17 +66,22 @@ def kill(self):
...
+            self.finalizer = util.Finalize(self, os.close, (parent_w,))
             self.sentinel = parent_r
+            self._parent_w = parent_w
```
pitrou: This is not required (but it doesn't hurt either).

pierreglaser:
@pitrou thanks for the review. 

Lib/multiprocessing/popen_forkserver.py
pitrou: I think `_parent_w` is ok. Also you don't need to make it an attribute (see other comment).

Lib/multiprocessing/popen_spawn_posix.py
pitrou: In case `os.pipe` fails probably (it's unlikely, but can happen if the open fd limit is reached).

Lib/multiprocessing/process.py
```diff
@@ -337,6 +348,40 @@ def __reduce__(self):
...
+    @property
+    def ident(self):
+        return self._pid
+
+    def join(self, timeout=None):
+        '''
+        Wait until parent process terminates
+        '''
+        from multiprocessing.connection import wait
+        wait([self._sentinel], timeout=5)
```
pitrou: `timeout=5` looks a bit authoritarian :-)

Lib/test/_test_multiprocessing.py
```diff
@@ -269,6 +269,67 @@ def _test(cls, q, *args, **kwds):
...
+        p = cls.Process(target=cls._test_report_parent_status, args=(wconn, ))
+        p.start()
+        time.sleep(100)
+
+    @classmethod
+    def _test_report_parent_status(cls, wconn):
+        from multiprocessing.process import parent_process
+        wconn.send("alive" if parent_process().is_alive() else "not alive")
+
+        start_time = time.monotonic()
```
pitrou: This isn't used anymore.

Lib/multiprocessing/popen_fork.py
```diff
@@ -66,17 +66,22 @@ def kill(self):
...
-                code = process_obj._bootstrap()
+                os.close(parent_w)
+                code = process_obj._bootstrap(parent_sentinel=child_r)
             finally:
                 os._exit(code)
         else:
             os.close(child_w)
+            os.close(child_r)
             self.finalizer = util.Finalize(self, os.close, (parent_r,))
+            self.finalizer = util.Finalize(self, os.close, (parent_w,))
```
pitrou: Hmm, this is clobbering the previous finalizer. You should use a single callback that closes both fds.

Lib/multiprocessing/popen_forkserver.py
pitrou: However, you're also clobbering the previous finalizer here.

pitrou (commented):
Just a couple more comments.

pierreglaser:
(Hopefully this will not make you cringe :))

pitrou (approved):

pitrou:
Thank you :-)

PR #13250, opened 2019-05-11
gh-80961: Add daemon parameter to ThreadPoolExecutor

hniksic:
See https://bugs.python.org/issue36780 ([GH-80961](https://github.com/python/cpython/issues/80961)).

https://bugs.python.org/issue36780

### DIFF
--- a/Lib/concurrent/futures/thread.py
+++ b/Lib/concurrent/futures/thread.py
@@ -159,7 +159,8 @@ class ThreadPoolExecutor(_base.Executor):
         return WorkerContext.prepare(initializer, initargs)
 
     def __init__(self, max_workers=None, thread_name_prefix='',
-                 initializer=None, initargs=(), **ctxkwargs):
+                 initializer=None, initargs=(), *, daemon=False,
+                 **ctxkwargs):
         """Initializes a new ThreadPoolExecutor instance.
 
         Args:
@@ -168,8 +169,15 @@ class ThreadPoolExecutor(_base.Executor):
             thread_name_prefix: An optional name prefix to give our threads.
             initializer: A callable used to initialize worker threads.
             initargs: A tuple of arguments to pass to the initializer.
+            daemon: If True, worker threads are created as daemon threads.
             ctxkwargs: Additional arguments to cls.prepare_context().
         """
+        if daemon:
+            if not threading._daemon_threads_allowed():
+                raise RuntimeError(
+                    'daemon threads are disabled in this (sub)interpreter')
+        self._daemon = daemon
+
         if max_workers is None:
             # ThreadPoolExecutor is often used to:
             # * CPU bound task which releases GIL
@@ -233,10 +241,12 @@ class ThreadPoolExecutor(_base.Executor):
             t = threading.Thread(name=thread_name, target=_worker,
                                  args=(weakref.ref(self, weakref_cb),
                                        self._create_worker_context(),
-                                       self._work_queue))
+                                       self._work_queue),
+                                 daemon=True if self._daemon else None)
             t.start()
             self._threads.add(t)
-            _threads_queues[t] = self._work_queue
+            if not self._daemon:
+                _threads_queues[t] = self._work_queue
 
     def _initializer_failed(self):
         with self._shutdown_lock:
--- a/Doc/library/concurrent.futures.rst
+++ b/Doc/library/concurrent.futures.rst
@@ -95,7 +95,8 @@ Executor Objects
       return immediately and the resources associated with the executor will be
       freed when all pending futures are done executing.  Regardless of the
       value of *wait*, the entire Python program will not exit until all
-      pending futures are done executing.
+      pending futures are done executing (see the *daemon* parameter of
+      :class:`ThreadPoolExecutor` for an exception to this rule).
 
       If *cancel_futures* is ``True``, this method will cancel all pending
       futures that the executor has not started running. Any futures that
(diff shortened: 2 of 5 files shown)


Lib/concurrent/futures/thread.py
```diff
@@ -211,11 +211,14 @@ def _initializer_failed(self):
                 if work_item is not None:
                     work_item.future.set_exception(BrokenThreadPool(self._broken))
 
-    def shutdown(self, wait=True):
+    def shutdown(self, wait=True, wait_at_exit=True):
```
brianquinlan: `shutdown` is defined in `Executor` because `Executor` is the abstract superclass for both `ThreadPoolExecutor` and `ProcessPoolExecutor`. Unless there is a very strong reason not to, this method should work the same in both executors.

hniksic: @brianquinlan This was intentional - I tested `shutdown(wait=False)` with `ProcessPoolExecutor`, and found that it raised exceptions and hanged the process at exit. (Not just hanged in the sense of waiting for the pending futures, but completely hanged, even when the futures exited.) So the new functionality is only available in and documented for `ThreadPoolExecutor`.

For example, when I run this script on Python 3.7:

```py
import time, concurrent.futures

pool = concurrent.futures.ProcessPoolExecutor()

pool.submit(time.sleep, 5)
print(1)

pool.shutdown(wait=False)
print(2)
```

The expected behavior is for the program to print 1 and 2 and then to wait for 5 seconds before exiting. Instead, it prints 1 and 2, but hangs at exit with the following output:

```
$ python3.7 ~/Desktop/x
1
2
Error in atexit._run_exitfuncs:
  File "/usr/lib/python3.7/concurrent/futures/process.py", line 101, in _python_exit
    thread_wakeup.wakeup()
  File "/usr/lib/python3.7/concurrent/futures/process.py", line 89, in wakeup
    self._writer.send_bytes(b"")
  File "/usr/lib/python3.7/multiprocessing/connection.py", line 183, in send_bytes
    self._check_closed()
  File "/usr/lib/python3.7/multiprocessing/connection.py", line 136, in _check_closed
    raise OSError("handle is closed")
Exception in thread QueueManagerThread:
  File "/usr/lib/python3.7/threading.py", line 917, in _bootstrap_inner
    self.run()
  File "/usr/lib/python3.7/threading.py", line 865, in run
    self._target(*self._args, **self._kwargs)
  File "/usr/lib/python3.7/concurrent/futures/process.py", line 368, in _queue_management_worker
    thread_wakeup.clear()
  File "/usr/lib/python3.7/concurrent/futures/process.py", line 92, in clear
    while self._reader.poll():
  File "/usr/lib/python3.7/multiprocessing/connection.py", line 255, in poll
    self._check_closed()
  File "/usr/lib/python3.7/multiprocessing/connection.py", line 136, in _check_closed
    raise OSError("handle is closed")
OSError: handle is closed
OSError: handle is closed
```

brianquinlan: I really wanted ThreadPoolExecutor and ProcessPoolExecutor to have the same API when I designed them.

Do you have any bandwidth to debug this? If not, I could take a look.

hniksic: Please do take a look if you can. I am not acquainted with the implementation of `ProcessPoolExecutor`, so it would take quite some time for me to trace what's going on.

It would of course be ideal if both classes supported the new flag.

brianquinlan: I was playing with `ProcessPoolExecutor` and it seems like there are a bunch of problems that are triggered when `pool.shutdown(wait=False)` is used. I filed a bug for one issue: https://bugs.python.org/issue39205

Do you think that your PR could hold off until I have a chance to sort some of this out?

hniksic: > Do you think that your PR could hold off until I have a chance to sort some of this out?

Sure, thanks for asking. We have a workaround, so it's no problem to wait for the proper solution. It's just that the workaround is so extremely ugly, involving monkey patch of a private method, that we'd definitely prefer the proper fix to land eventually.

hniksic:
I've now updated this PR to address the feedback from the issue discussion. The `wait_at_exit` parameter has moved from `shutdown()` to the constructor as a `daemon=False` keyword-only parameter, as agreed upon in the issue. The parameter is on the constructor only, keeping the `shutdown()` API consistent across executor types (per @brianquinlan's [review comment](https://github.com/python/cpython/pull/13250#discussion_r346882824)). As before, `shutdown(wait=True)` still joins threads explicitly.

Since Python 3.9 made executor threads non-daemon, the implementation now also creates actual daemon threads, not just removes them from `_threads_queues`.

Tests are updated accordingly.

cc @brianquinlan

hniksic:
Hi @aeros, I updated this PR back in March based on the issue discussion. The revised design specifically addresses the concerns you raised:

- `daemon` is now a keyword-only parameter on the `ThreadPoolExecutor` constructor
- When `daemon=True`, the executor creates real daemon threads
- `shutdown(wait=True)` still joins explicitly; the `shutdown()` API is unchanged
- The docs flag the usual daemon-thread caveats
- Tests in `test_shutdown.py` cover most of the verification scenarios you listed: that daemon threads are actually created, that they're absent from `_threads_queues`, that the interpreter exits cleanly with a still-running worker, and that `shutdown(wait=True)` and context-manager exit still wait for completion. Happy to add more if there are specific finalization paths you want exercised explicitly.

If you have time for another look, I'd appreciate it.

PR #13251, opened 2019-05-11
bpo-36889: Merge asyncio streams

asvetlov:


asvetlov:
The main part is basically done. The functionality is test covered for all significant code paths because the PR doesn't change too much but "just" merges StreamReader and StreamWriter.

There are two options for future steps:
1. Add several tests for `StreamMode` related API, write NEWS, merge and make other improvements in following pull requests. I prefer this way for the sake of simplifying the review process. I have a strong feeling that the proposed change is viable. We did miss it starting from Python 3.5.
2. Another approach is to keep working on this branch to collect all wish changes at once. The review becomes complicated though. 

The wish list is: 
1. Add `asyncio.connect()` and `asyncio.StreamServer()` to avoid working with `stream, stream` where was `reader, writer`. The same for UNIX sockets. 
`StreamServer` should be mimic to `asyncio.AbstractServer` but without boilerplate. The usage is:
 ```
async with StreamServer(host, port) as server:
    await server.serve_forever()
```
`StreamServer` should create low-level asyncio server (`loop.create_server()`) with `start_serving=False`. We discussed the server design with @1st1 offline already.

2. Support both Windows and UNIX pipes (a new API whose implementation is really trivial with unified streams). See `loop.connect_read_pipe` and `loop.connect_write_pipe` for the context.

3. Rewrite `SubprocessStreamProtocol` to add subprotocols from stdin, stdout, and stderr. It can simplify things a lot.

File: Lib/asyncio/streams.py [hunk 1]
```diff
@@ -22,6 +23,26 @@
 _DEFAULT_LIMIT = 2 ** 16  # 64 KiB
 
 
+class StreamMode(enum.Enum):
+    READ = "read"
+    WRITE = "write"
+    READWRITE = "readwrite"
+
+    def is_read(self):
```
1st1: I'd make these prefixed with an `_` underscore.  Exposing them as public api doesn't make a lot of sense imo.  Or maybe just make these methods top level private functions.

Lib/asyncio/streams.py
```diff
@@ -22,6 +23,26 @@
 _DEFAULT_LIMIT = 2 ** 16  # 64 KiB
 
 
+class StreamMode(enum.Enum):
+    READ = "read"
+    WRITE = "write"
+    READWRITE = "readwrite"
```
1st1: I'd make this a Flag enum. Then you can use & and |.

1st1:
Overall looks fine.  Are there any more todo items in this PR?  If no I can review more thoroughly.

Lib/asyncio/streams.py
asvetlov: Nice idea!

Lib/asyncio/streams.py
asvetlov: Converted into private methods.

asvetlov:
The last improvement is spritting stream protocol into client and server parts.
It allows getting rid of complex if-statements in implementation.
Basically, client protocol keeps a weak reference to the stream. It allows detecting stream object deletion by a user before stream closing.
Server protocol should keep a strong reference, a user doesn't provide a stream instance but receives it in a callback.

asvetlov:
I think the PR is ready for careful review.
I'm working on additional tests.

Lib/test/test_asyncio/test_streams.py
```diff
@@ -1095,23 +1161,48 @@ def test_eof_feed_when_closing_writer(self):
...
-                                             _asyncio_internal=True)
         with self.assertWarns(DeprecationWarning):
-            asyncio.StreamWriter('transport', proto, reader, self.loop)
-
-
+            asyncio.StreamWriter
+
+    def test_stream_reader_forbidden_ops(self):
+        async def inner():
+            stream = asyncio.Stream(mode=asyncio.StreamMode.READ)
```
tirkarthi: `DeprecationWarning` over absence of `_asyncio_internal=True`

Lib/test/test_asyncio/test_streams.py
```diff
@@ -1095,23 +1161,48 @@ def test_eof_feed_when_closing_writer(self):
...
+            with self.assertRaisesRegex(RuntimeError, "The stream is read-only"):
+                stream.write_eof()
+            with self.assertRaisesRegex(RuntimeError, "The stream is read-only"):
+                await stream.drain()
+
+        self.loop.run_until_complete(inner())
+
+    def test_stream_writer_forbidden_ops(self):
+        async def inner():
+            stream = asyncio.Stream(mode=asyncio.StreamMode.WRITE)
```
tirkarthi: `DeprecationWarning` over absence of `_asyncio_internal=True`

Lib/test/test_asyncio/test_streams.py
asvetlov: Fixed, thanks!

Lib/test/test_asyncio/test_streams.py
asvetlov: Fixed

icgood:
I agree that StreamReader and StreamWriter were too coupled and merging is a good idea from an implementation perspective, but I'm wondering about the typing perspective. For example, it would be ideal if mypy (or similar) could tell me that I can't `write()` a subprocess stdout stream. Could be solved in typeshed but I wonder if it's an argument for re-adding `StreamReader`/`StreamWriter` as interfaces in `collections.abc` or something?

asvetlov:
Going on this way we need AbstractStreamReader and AbstractStreamWriter.
It makes sense if there are at least more than one alternative implementation for stream class.
I'm open to reconsidering it after getting the real use case.

Side note: we have no AbstractFuture and AbstractTask yet while there is a way to create own future and task implementations:)

Lib/asyncio/streams.py
```diff
@@ -81,16 +122,125 @@
...
+    async def __aexit__(self, exc_type, exc_value, exc_tb):
+        await self.close()
+
+    def _attach(self, stream, task):
+        self._streams[stream] = task
+
+    def _detach(self, stream, task):
+        del self._streams[stream]
+
+    async def _warn_active_tasks(tasks):
```
1st1: _shutdown_active_tasks()?

asvetlov: Done

Lib/asyncio/__init__.py
```diff
@@ -43,3 +44,21 @@
...
+
+__all__ += ('StreamReader', 'StreamWriter')  # deprecated
+
+
+def __getattr__(name):
+    if name == 'StreamReader':
+        warnings.warn("StreamReader is deprecated, use asyncio.Stream instead",
+                      DeprecationWarning,
+                      stacklevel=2)
+        return Stream
```
1st1: Do you think it's a good idea to return `Stream` when the user has *explicitly* asked for a `StreamReader`?  Is there a way we can return `StreamReader` and `StreamWriter` (at least in 3.8)?  I feel that would be a smoother migration path.

Lib/asyncio/streams.py
```diff
@@ -1,27 +1,70 @@
...
+                'UnixStreamServer')
 
 from . import coroutines
 from . import events
 from . import exceptions
 from . import format_helpers
 from . import protocols
 from .log import logger
-from .tasks import sleep
+from .tasks import sleep, wait
```
1st1: Can we import `from . import tasks` and then `tasks.sleep`, `tasks.wait`?

Lib/asyncio/streams.py
```diff
@@ -81,16 +124,185 @@
...
+                                               client_connected_cb,
+                                               loop=loop,
+                                               _asyncio_internal=True)
         return protocol
 
     return await loop.create_server(factory, host, port, **kwds)
 
 
+class _BaseStreamServer:
+    # TODO: add __repr__
```
1st1: I think it's time to add it :)

Lib/asyncio/streams.py
```diff
@@ -81,16 +124,185 @@
...
+    async def __aenter__(self):
+        await self.bind()
+        return self
+
+    async def __aexit__(self, exc_type, exc_value, exc_tb):
+        await self.close()
+
+    def __init_subclass__(cls):
+        if not cls.__module__.startswith('asyncio.'):
+            raise TypeError("Stream server classes are final, don't inherit from them")
```
1st1:
```suggestion
            raise TypeError(f"asyncio.{cls.__name__} class cannot be inherited from")
```

Lib/asyncio/streams.py
```diff
@@ -81,16 +124,185 @@
...
+        self._streams[stream] = task
+
+    def _detach(self, stream, task):
+        del self._streams[stream]
+
+    async def _shutdown_active_tasks(self, tasks):
+        if not tasks:
+            return
+        # NOTE: tasks finished with exception are reported
+        # by Tast/Future __del__ method
```
1st1:
```suggestion
        # by the Task.__del__() method.
```

Lib/asyncio/streams.py
```diff
@@ -81,16 +124,185 @@
...
+        done, pending = await wait(tasks, timeout=self._shutdown_timeout)
+        if not pending:
+            return
+        for task in pending:
+            task.cancel()
+        done, pending = await wait(pending, timeout=self._shutdown_timeout)
+        for task in pending:
+            self._loop.call_exception_handler({
+                "message": f'{task} was not finished on stream server closing'
```
1st1:
```suggestion
                "message": f'{task!r} ignored cancellation request from a closing StreamServer {self!r}',
                "stream_server": self
```

Lib/asyncio/streams.py
```diff
@@ -81,16 +124,185 @@
...
+    def __init__(self, client_connected_cb,
+                 limit=_DEFAULT_LIMIT,
+                 shutdown_timeout=60,
+                 _asyncio_internal=False):
+        if not _asyncio_internal:
+            raise RuntimeError("_ServerStream is a private asyncio class")
+        self._client_connected_cb = client_connected_cb
+        self._limit = limit
+        self._loop = events.get_running_loop()
+        self._low_server = None
```
1st1: "low_server" reads ambiguous. Maybe spell it "_io_server"?

Lib/asyncio/streams.py
```diff
@@ -81,16 +124,185 @@
...
+
+    def addresses(self):
+        # I don't want to expose plain socket.socket objects as low-level
+        # asyncio.Server does but exposing served IP addresses or unix paths
+        # is useful
+        #
+        # multiple value for socket bound to both IPv4 and IPv6 families
+        if self._low_server is None:
+            return []
+        return [sock.getsockname() for sock in self._low_server.sockets]
```
1st1: I don't like the idea of returning lists, they cannot be cached (as users can mutate them). Let's return a tuple or a frozenset.

Also, in uvloop I return a "socket-like" object.  I think it makes more sense than adding multiple ad-hoc methods to introspect the underlying sockets.  Maybe we copy https://github.com/MagicStack/uvloop/blob/master/uvloop/pseudosock.pyx into asyncio 3.8 and have a `StreamServer.sockets -> PseudoSocket` API?

Lib/asyncio/streams.py
```diff
@@ -81,16 +124,185 @@
...
+    async def bind(self):
+        if self._low_server is not None:
+            return
+        self._low_server = await self._bind()
+
+    def is_bound(self):
+        return self._low_server is not None
+
+    def addresses(self):
+        # I don't want to expose plain socket.socket objects as low-level
```
1st1:
```suggestion
        # We don't want to expose plain socket.socket objects as low-level
```

Please don't use "I" :)

Lib/asyncio/streams.py
```diff
@@ -81,16 +124,185 @@
...
+    def __init__(self, client_connected_cb, host=None, port=None, *,
+                 limit=_DEFAULT_LIMIT,
+                 family=socket.AF_UNSPEC,
+                 flags=socket.AI_PASSIVE, sock=None, backlog=100,
+                 ssl=None, reuse_address=None, reuse_port=None,
+                 ssl_handshake_timeout=None,
+                 shutdown_timeout=60):
+        # client_connected_cb name is consistent with legacy API
+        # but it is long and ugly
+        # any suggestion?
```
1st1: Yes -- make it a positional-only argument.

Lib/asyncio/__init__.py
asvetlov: The real question is that: why user need to import `Stream` class at all?
I guess for two reasons: hacking by deriving a custom stream class from `StreamReader` for example. Sorry, this way is broken, `Stream` has no compatible constructor with `StreamReader`.
Note, both `Stream.__init__` and `StreamReader.__init__` are private constructors, a user should never call them directly but use factories like `open_connection()` / `connect()`.
The second reason is `isinstance(obj, StreamReader)` check. 
Now both `isinstance(obj, asyncio.Stream)` and `isinstance(obj, asyncio.StreamReader)` returns the same result, which is good I think.

1st1: > I guess for two reasons: hacking by deriving a custom stream class from StreamReader for example. Sorry, this way is broken, Stream has no compatible constructor with StreamReader.

Right, but it means you are making a backwards incompatible change in 3.8 by returning `Stream`.  We can't do that.

Lib/asyncio/streams.py
asvetlov: Done.

Lib/asyncio/streams.py
asvetlov: Done

Lib/asyncio/streams.py
asvetlov: I mean `low_server` for *low-level server*. I little clumsily, agree.

Honestly, `_io_server` doesn't sound better.
`_impl`, whatever it means?

If you prefer `_io_server` I can live with it, just left a comment

Lib/asyncio/streams.py
asvetlov: I completely understand why do you use PseudoSocket for uvloop implementation.
But for high-level API the only safe server socket operation is `sock.getsockname()` (and `sock.getsockopt()`. IPv4 vs IPv6 can be distinguished by the length of addresses item. Not very cool but works at least.

`setsockopt()` can have dangerous side-effects.
Even `fileno()` can be used for dirty hacks.
For high-level AIP, if we need the functionality provided by `setsockopt` better to support it by keyword-only parameters for stream constructor, as we already do for `backlog`, `reuse_address` and `reuse_port`.

I agree with `tuple` though. More restrictive type is better.
I don't like `frozenset` mostly because I would write `server.addresses()[0]`.

1st1: > setsockopt() can have dangerous side-effects.
> Even fileno() can be used for dirty hacks.

Sometimes such hacks are necessary (e.g. people used it to set NODELAY before we had it).  I don't want to close the door here.  That's why I'm suggesting to return a socket-like object that allows *some* limited functionality, instead of adding `StreamServer.addresses`, `StreamServer.filenos`, `StreamServer.sockopts` etc -- because asyncio users will start demanding us to add them.

asvetlov:
Trying re-run tests several times to make sure that there is no flaky tests

Lib/asyncio/streams.py
```diff
@@ -1,27 +1,121 @@
...
+
+    def __init__(self, sock, _asyncio_internal=False):
+        if not _asyncio_internal:
+            raise TypeError(f"{self.__class__} should be instaniated "
+                            "by asyncio internals only")
+        self._family = sock.family
+        self._type = sock.type
+        self._proto = sock.proto
+        self._addr = sock.getsockname()
+        self._fileno = sock.fileno()
```
1st1: Can you also add `__init_subclass__` to deny inheritance?

Lib/asyncio/streams.py
```diff
@@ -1,27 +1,121 @@
...
+    def fileno(self):
+        return self._fileno
+
+    def __repr__(self):
+        ret = [f"<{self.__class__.__name__}"]
+        ret.append(f"family={self.family}")
+        ret.append(f"type={self.type}")
+        ret.append(f"proto={self.proto}")
+        ret.append(f"addr={self.addr}")
+        ret.append(f"fileno={self.fileno()}")
```
1st1:
```suggestion
        ret.append(f"fileno={self.fileno()}>")
```

Lib/asyncio/streams.py
1st1: `_server_impl` sounds good.

Lib/asyncio/streams.py
```diff
@@ -91,40 +191,276 @@ def factory():
...
+        await self._shutdown_active_tasks(active_tasks)
+
+    async def __aenter__(self):
+        await self.bind()
+        return self
+
+    async def __aexit__(self, exc_type, exc_value, exc_tb):
+        await self.close()
+
+    def __init_subclass__(cls):
```
1st1: Please move this method to the very top of the class.

Lib/asyncio/streams.py
```diff
@@ -802,3 +1028,672 @@ def __aiter__(self):
...
+        self._server._detach(self._stream, self._task)
+        self._stream = None
+
+
+def _swallow_unhandled_exception(task):
+    # Do a trick to suppress unhandled exception
+    # if stream.write() was used without await and
+    # stream.drain() was paused and resumed with an exception
+
+    # TODO: add if not task.cancelled() check!!!!
```
1st1: Need to clean this up before committing.

Lib/asyncio/streams.py
```diff
@@ -802,3 +1028,672 @@ def __aiter__(self):
...
+    """Wraps a Transport.
+
+    This exposes write(), writelines(), [can_]write_eof(),
+    get_extra_info() and close().  It adds drain() which returns an
+    optional Future on which you can wait for flow control.  It also
+    adds a transport property which references the Transport
+    directly.
+    """
+
+    # TODO: add __aenter__ / __aexit__ to close stream
```
1st1: Let's do this.

Lib/asyncio/streams.py
```diff
@@ -802,3 +1028,672 @@ def __aiter__(self):
...
+        val = await self.readline()
+        if val == b'':
+            raise StopAsyncIteration
+        return val
+
+    # async def __aenter__(self):
+    #     return self
+
+    # async def __aexit__(self, exc_type, exc_val, exc_tb):
+    #     await self.close()
```
1st1: Uncomment?

Lib/asyncio/streams.py
```diff
@@ -1,27 +1,121 @@
...
+    @property
+    def type(self):
+        return self._type
+
+    @property
+    def proto(self):
+        return self._proto
+
+    @property
+    def addr(self):
```
1st1: I do want to return instances of this class from `transport.get_extra_info('socket')`.  So I'm -1 on the current design and propose to have `.getsockname()` method instead.

Lib/asyncio/streams.py
asvetlov: The class is gone

Lib/asyncio/streams.py
asvetlov: Done

Lib/asyncio/streams.py
asvetlov: The question is: what API want we provide?
Now we have
```
async with await asyncio.connect(...) as stream:
    ...
```
`async with await` looks ugly.
On the other hand, I like to support 
```
stream = await asyncio.connect(...)
...
await stream.close()
```

To support 
```
async with connect(...) as stream:
     ...
```
the `connect` function should be a plain function that returns a helper class with `__aenter__`, `__aexit__` and `__await__` methods like I do in `aiohttp` and other libraries.
Pretty straightforward, though.

Lib/asyncio/streams.py
asvetlov: The class is removed for the sake of using whatever that `asyncio.Sever.sockets` return.

Lib/asyncio/streams.py
asvetlov: Done

Lib/asyncio/streams.py
asvetlov: Done

Lib/asyncio/streams.py
asvetlov: `asyncio.Server.sockets` are returned.
The type is `list`, I didn't change the type of already existing `sockets` property

Lib/asyncio/streams.py
asvetlov: Done and tested :)

Lib/asyncio/streams.py
asvetlov: Done

asvetlov:
The PR is done and ready for the final review and merging.

Lib/asyncio/streams.py
```diff
@@ -802,3 +1042,675 @@ def __aiter__(self):
...
+        if not self._transport.is_closing():
+            if self._protocol._connection_lost:
+                fut = self._loop.create_future()
+                fut.set_exception(ConnectionResetError('Connection lost'))
+                return fut
+            if not self._protocol._paused:
+                # fast path, the stream is not paused
+                # no need to wait for resume signal
+                return self._complete_fut
+        return _OptionalAwait(self.drain)
```
asvetlov, line 1370: I replaced returning `loop.create_task(drain())` with `_OptionalAwait(self.drain)`.
The reason is: I don't like that `drain()` code is executed even if user didn't ask (plain `stream.write(data)` call).
If `drain()` failed with exception but `write()` was not awaited the exception was discarded.

It works pretty well I believe but confuses code readers. Also, future modifications of `drain()` method can produce a weird side-effect.
Better to avoid execution of dead code at all, the mental model becomes much clearer at least.

auvipy (approved):

Lib/asyncio/streams.py
```diff
@@ -1,27 +1,152 @@
...
+    await loop.connect_read_pipe(
+        lambda: _StreamProtocol(stream, loop=loop,
+                                _asyncio_internal=True),
+        pipe)
+    return stream
+
+
+def connect_write_pipe(pipe, *, limit=_DEFAULT_LIMIT):
+    # Design note:
+    # Don't use decorator approach but exilicit non-async
```
1st1:
```suggestion
    # Don't use decorator approach but explicit non-async
```

Lib/asyncio/streams.py
```diff
@@ -1,27 +1,152 @@
...
+        flags=flags, sock=sock, local_addr=local_addr,
+        server_hostname=server_hostname,
+        ssl_handshake_timeout=ssl_handshake_timeout,
+        happy_eyeballs_delay=happy_eyeballs_delay, interleave=interleave)
+    return stream
+
+
+def connect_read_pipe(pipe, *, limit=_DEFAULT_LIMIT):
+    # Design note:
+    # Don't use decorator approach but exilicit non-async
```
1st1:
```suggestion
    # Don't use decorator approach but explicit non-async
```

Lib/asyncio/streams.py
```diff
@@ -91,40 +222,301 @@ def factory():
...
+        self._server_impl = await self._bind()
+
+    def is_bound(self):
+        return self._server_impl is not None
+
+    @property
+    def sockets(self):
+        # multiple value for socket bound to both IPv4 and IPv6 families
+        if self._server_impl is None:
+            return []
```
1st1:
```suggestion
            return ()
```

Lib/asyncio/streams.py
```diff
@@ -91,40 +222,301 @@ def factory():
...
+        return self._server_impl.sockets
```
1st1, line 270: Is this a list or a tuple? Maybe 
```suggestion
        return tuple(self._server_impl.sockets)
```
?

Lib/asyncio/streams.py
```diff
@@ -802,3 +1084,673 @@ def __aiter__(self):
...
+        self._server._detach(self._stream, self._task)
+        self._stream = None
+
+
+class _OptionalAwait:
+    # The class doesn't create a coroutine
+    # if not awaited
+    # It prevents "coroutine is never awaited" message
+
+    __slot___ = ('_method', '_args', '_kwargs')
```
1st1:
```suggestion
    __slots__ = ('_method', '_args', '_kwargs')
```

Lib/asyncio/streams.py
```diff
@@ -802,3 +1084,673 @@ def __aiter__(self):
...
+
+    def __init__(self, method, *args, **kwargs):
```
1st1: Please drop `*args` and `**kwargs`.  They aren't used by your code but still impose overhead (albeit insignificant)

asvetlov: I thought about generic implementation.
Agree, let's drop.
We can add params later if needed

Lib/asyncio/streams.py
asvetlov: This is a list because `asyncio.Server.sockets` is the list property.
I wanted to be consistent

Lib/asyncio/streams.py
asvetlov: Done

Lib/asyncio/streams.py
asvetlov: Done

Misc/NEWS.d/next/Library/2019-05-14-12-25-44.bpo-36889.MChPqP.rst
```diff
@@ -0,0 +1,2 @@
+Merge asyncio.StreamReader and asyncio.StreamWriter into asyncio.Stream
```
1st1: Please enumerate all new APIs in this NEWS entry.

Lib/asyncio/streams.py
```diff
@@ -91,40 +222,301 @@ def factory():
...
+    async def start_serving(self):
+        await self.bind()
+        await self._server_impl.start_serving()
+
+    async def serve_forever(self):
+        await self.start_serving()
+        await self._server_impl.serve_forever()
+
+    async def close(self):
+        if self._server_impl is None:
```
1st1, line 286: This isn't correct.  Because we use `Stream` in subprocess APIs, this is not a backwards compatible change.  Please make it synchronous and you can return an `_OptionalAwait` from it.

Lib/asyncio/streams.py
```diff
@@ -1,27 +1,152 @@
...
 
 _DEFAULT_LIMIT = 2 ** 16  # 64 KiB
 
 
+class StreamMode(enum.Flag):
+    READ = enum.auto()
+    WRITE = enum.auto()
+    READWRITE = READ | WRITE
+
+    def _check_read(self):
```
1st1: Can you make this a top-level function?

Lib/asyncio/streams.py
1st1: My bad, this is a different class :)

1st1 (approved):
Please try to address my comments prior to merging. Great work, Andrew!

Lib/asyncio/streams.py
asvetlov: Done

Lib/asyncio/streams.py
```diff
@@ -1,27 +1,154 @@
...
 _DEFAULT_LIMIT = 2 ** 16  # 64 KiB
 
 
+class StreamMode(enum.Flag):
+    READ = enum.auto()
+    WRITE = enum.auto()
+    READWRITE = READ | WRITE
+
+
+def _check_read(mode):
```
1st1: nit: I think `_ensure_can_read` would be a better name.

Lib/asyncio/streams.py
1st1: https://github.com/python/cpython/pull/13530 fixes that for server.  Please return an empty tuple here as well.

asvetlov: Done

Lib/asyncio/streams.py
asvetlov: Done

Lib/asyncio/streams.py
asvetlov: Done

PR #13252, opened 2019-05-11
[3.6] bpo-35925: Skip SSL tests that fail due to weak external certs or old TLS (GH-13124)

gpshead:
Modern Linux distros such as Debian Buster have default OpenSSL system
configurations that reject connections to servers with weak certificates
and the use of TLS versions less than TLSv1.2 by default. This causes our
test suite run with external networking resources enabled to skip these tests
when they encounter such a failure or configuration.

(cherry picked from commit 2cc0223)

Additional change to test_ssl.py required as 3.6 and earlier have legacy protocol tests.
The test_httplib.py change is omitted from this backport as self-signed.pythontest.net's
certificate was updated.

Authored-by: Gregory P. Smith <greg@krypto.org>

https://bugs.python.org/issue35925

### DIFF
--- a/Lib/test/test_httplib.py
+++ b/Lib/test/test_httplib.py
@@ -4,6 +4,7 @@ import io
 import itertools
 import os
 import array
+import re
 import socket
 
 import unittest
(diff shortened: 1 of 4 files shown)


gpshead:
This will make the debian buster buildbot green again for 3.6.

@larryhastings you might want to cherry pick this for 3.5 as well if people building 3.5.x on systems with a modern openssl configuration and running -u all tests is a concern.

PR #13253, opened 2019-05-11
[2.7] bpo-35925: Skip SSL tests that fail due to weak external certs or old TLS (GH-13124)

gpshead:
Modern Linux distros such as Debian Buster have default OpenSSL system
configurations that reject connections to servers with weak certificates
and the use of TLS versions less than TLSv1.2 by default. This causes our
test suite run with external networking resources enabled to skip these tests
when they encounter such a failure or configuration.

(cherry picked from commit 2cc0223)

Changes to test_ssl.py required as 2.7 has legacy protocol tests.

The test_httplib.py change is omitted from this backport as
self-signed.pythontest.net's certificate was updated and the
test_nntplib.py change is not applicable on 2.7.

Authored-by: Gregory P. Smith greg@krypto.org

https://bugs.python.org/issue35925

### DIFF
--- a/Lib/test/test_ssl.py
+++ b/Lib/test/test_ssl.py
@@ -19,6 +19,7 @@ import urllib2
 import traceback
 import weakref
 import platform
+import re
 import functools
 from contextlib import closing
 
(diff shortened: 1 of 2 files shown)


gpshead:
This one is derived from the 3.6 change in https://github.com/python/cpython/pull/13252 as it and 2.7 are both old enough to have some test_ssl things to update.

PR #13254, opened 2019-05-11
bpo-27805: Allow opening /dev/std{out,err} in append mode

ZackerySpytz:
For example, `open('/dev/stdout', 'a')` will no longer raise an `OSError`.

https://bugs.python.org/issue27805

### DIFF
--- a/Modules/_io/fileio.c
+++ b/Modules/_io/fileio.c
@@ -477,9 +477,31 @@ _io_FileIO___init___impl(fileio *self, PyObject *nameobj, const char *mode,
            end of file (otherwise, it might be done only on the
            first write()). */
         PyObject *pos = portable_lseek(self, NULL, 2);
-        if (pos == NULL)
+        if (pos != NULL) {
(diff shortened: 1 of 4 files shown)


auvipy (approved):

Modules/_io/fileio.c
```diff
@@ -477,9 +477,31 @@ _io_FileIO___init___impl(fileio *self, PyObject *nameobj, const char *mode,
            end of file (otherwise, it might be done only on the
            first write()). */
         PyObject *pos = portable_lseek(self, NULL, 2);
```
benjaminp, line 479: Wouldn't it be simpler to check for `ESPIPE` in `portable_lseek`. (Perhaps you need to add a parameter?)

benjaminp (changes requested):
(dropping from my queue because of my question)

benjaminp:
Closing because GH-17112 was merged instead. Thank you for the PR, though.

PR #13257, opened 2019-05-12
Docs: Use python -m pip install everywhere

bsolomon1124:
Follows a similar recent Django ticket:

> https://code.djangoproject.com/ticket/30367

Quoting from that ticket/PR:

> Using ``python -m pip install foo`` makes sure the copy of pip looked up and executed is the one installed for the interpreter copy currently in use (be it a system-wide or a virtualenv's).

**This is not technically needed in a virtualenv, but the idea here is consistency: pick a single option that works everywhere and be consistent about it.**

### DIFF
--- a/Doc/howto/pyporting.rst
+++ b/Doc/howto/pyporting.rst
@@ -31,19 +31,19 @@ are:
 
 #. Only worry about supporting Python 2.7
 #. Make sure you have good test coverage (coverage.py_ can help;
-   ``pip install coverage``)
+   ``python -m pip install coverage``)
 #. Learn the differences between Python 2 & 3
-#. Use Futurize_ (or Modernize_) to update your code (e.g. ``pip install future``)
+#. Use Futurize_ (or Modernize_) to update your code (e.g. ``python -m pip install future``)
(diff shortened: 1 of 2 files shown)


BoboTiG (approved):
Thanks @bsolomon1124, I am +1 for those changes.

Doc/howto/pyporting.rst
```diff
@@ -31,19 +31,19 @@ are:
 
 #. Only worry about supporting Python 2.7
 #. Make sure you have good test coverage (coverage.py_ can help;
-   ``pip install coverage``)
+   ``python -m pip install coverage``)
```
brettcannon, line 34: Maybe it would be good to add a note to this section that all installation commands assume the user is in a virtual environment (with appropriate link to the glossary)? We unfortunately can't assume the readers of this HOWTO are experienced developers who know not to install globally.

brettcannon (changes requested):
Could you add a comment explaining that the

bsolomon1124:

brettcannon (approved):

brettcannon:
Thanks!

PR #13258, opened 2019-05-12
Docs: Add bz2 usage examples

bsolomon1124:
- Adds an "Examples of usage" section inspired by the one
  found in the gzip docs
- Corrects the descriptions for ``compresslevel`` and ``data``:
    - ``compresslevel`` must be an `int`, not any number.  For
      instance, passing a float will raise ``TypeError``
    - Note that `data` must be bytes-like

### DIFF
--- a/Doc/library/bz2.rst
+++ b/Doc/library/bz2.rst
@@ -83,7 +83,7 @@ All of the classes in this module may safely be accessed from multiple threads.
 
    The *buffering* argument is ignored. Its use is deprecated since Python 3.0.
 
-   If *mode* is ``'w'`` or ``'a'``, *compresslevel* can be a number between
+   If *mode* is ``'w'`` or ``'a'``, *compresslevel* can be an integer between
    ``1`` and ``9`` specifying the level of compression: ``1`` produces the
    least compression, and ``9`` (default) produces the most compression.
 
@@ -148,7 +148,7 @@ Incremental (de)compression
    incrementally. For one-shot compression, use the :func:`compress` function
    instead.
 
-   *compresslevel*, if given, must be a number between ``1`` and ``9``. The
+   *compresslevel*, if given, must be an integer between ``1`` and ``9``. The
    default is ``9``.
 
    .. method:: compress(data)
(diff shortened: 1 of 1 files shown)


Doc/library/bz2.rst
```diff
@@ -254,3 +254,75 @@ One-shot (de)compression
    .. versionchanged:: 3.3
       Support for multi-stream inputs was added.
 
+.. _bz2-usage-examples:
+
+Examples of usage
+-----------------
+
+Below are some examples of typical usage of the :mod:`bz2` module:.
```
pitrou:
```suggestion
Below are some examples of typical usage of the :mod:`bz2` module.
```

pitrou (commented):
LGTM generally. I'll let someone else review the grammar.

gpshead:
i made an edit that I believe will fix the failing doctest.  otherwise this looks good to me.  waiting on CI now.  as this is just a documentation improvement, no issue or news entry necessary.

bsolomon1124:
Thanks for having a look @gpshead - two thoughts/questions:

1. Why might the `502` (bytes written) output make doctests fail?  Would this number vary by platform?
2. One example that I can be self-critical of is the one with `gen_data()`: because the data is very random, the compression ratio is actually < 1.  (I.e. compression ratio will be much higher for non-random-looking data.)  Do you think that this deserves some type of rewrite to a more realistic scenario, or does it still get the point across?

bsolomon1124:
Looks like the build is failing because a "`" is used inside a function docstring.  News to me that wasn't allowed.

Screen Shot 2019-05-12 at 10 01 00 PM

gpshead:
> Why might the 502 (bytes written) output make doctests fail?

It was in the wrong place as `... 502` as if it were be entered as a line of code in the interactive session rather than `502` as the result. I could've left it there, but it was an irrelevant detail and would need updating if the data string above were ever changed.

btw, i removed the extra ``s.  i just obey whatever sphinx says for this sort of thing.

gpshead:
> One example that I can be self-critical of is the one with gen_data(): because the data is very random, the compression ratio is actually < 1. (I.e. compression ratio will be much higher for non-random-looking data.) Do you think that this deserves some type of rewrite to a more realistic scenario, or does it still get the point across?

heh, true.  compressing random data is rather pointless.  how about just changing that to yield something else.  Even if it just yields chunks of `b'z' * chunksize` that'd be fine.

gpshead:
Thanks for the documentation improvement, always appreciated!

bsolomon1124:
@gpshead appreciate your guidance.  It was a positive experience as my first CPython PR.

PR #13261, opened 2019-05-12
bpo-36871: Ensure method signature is used when asserting mock calls to a method

tirkarthi:
https://bugs.python.org/issue36871

### DIFF
--- a/Lib/unittest/mock.py
+++ b/Lib/unittest/mock.py
@@ -739,9 +739,10 @@ class NonCallableMock(Base):
 
     def _get_call_signature_from_name(self, name):
         """
-        If there is no name like the case for call check against functions
-        with tuples then return the self's spec signature.
-        If the name is not empty then remove () and split by '.' to get
+        * If call objects are asserted against a method/function like obj.meth1
+        then there could be no name for the call object to lookup. Hence just
+        return the spec_signature of the method/function being asserted against.
+        * If the name is not empty then remove () and split by '.' to get
         list of names to iterate through the children until a potential
         match is found. A child mock is created only during attribute access
         so if we get a _SpecState then no attributes of the spec were accessed
--- a/Lib/unittest/mock.py
+++ b/Lib/unittest/mock.py
@@ -737,6 +737,34 @@ class NonCallableMock(Base):
         return message % (expected_string, actual_string)
 
 
+    def _get_call_signature_from_name(self, name):
+        """
+        If there is no name like the case for call check against functions
+        with tuples then return the self's spec signature.
+        If the name is not empty then remove () and split by '.' to get
+        list of names to iterate through the children until a potential
+        match is found.
+        """
+        if not name:
+            return self._spec_signature
+
+        sig = None
+        names = name.replace('()', '').split('.')
+        children = self._mock_children
+
+        for name in names:
+            child = children.get(name)
+            if not child:
+                break
+            else:
+                children = child._mock_children.get(name)
+                sig = child._spec_signature
+                if not children:
+                    break
+
+        return sig
+
+
     def _call_matcher(self, _call):
         """
         Given a call (or simply an (args, kwargs) tuple), return a
@@ -744,7 +772,12 @@ class NonCallableMock(Base):
         This is a best effort method which relies on the spec's signature,
         if available, or falls back on the arguments themselves.
         """
-        sig = self._spec_signature
+
+        if isinstance(_call, tuple) and len(_call) > 2:
+            sig = self._get_call_signature_from_name(_call[0])
+        else:
+            sig = self._spec_signature
+
         if sig is not None:
             if len(_call) == 2:
                 name = ''
--- a/Lib/unittest/mock.py
+++ b/Lib/unittest/mock.py
@@ -743,7 +743,9 @@ class NonCallableMock(Base):
         with tuples then return the self's spec signature.
         If the name is not empty then remove () and split by '.' to get
         list of names to iterate through the children until a potential
-        match is found.
+        match is found. A child mock is created only during attribute access
+        so if we get a _SpecState then no attributes of the spec were accessed
+        and can be safely exited.
         """
         if not name:
             return self._spec_signature
@@ -754,13 +756,11 @@ class NonCallableMock(Base):
 
         for name in names:
             child = children.get(name)
-            if child is None:
+            if child is None or isinstance(child, _SpecState):
                 break
             else:
-                children = child._mock_children.get(name)
+                children = child._mock_children
                 sig = child._spec_signature
-                if children is None:
-                    break
 
         return sig
 
--- a/Lib/unittest/mock.py
+++ b/Lib/unittest/mock.py
@@ -754,12 +754,12 @@ class NonCallableMock(Base):
 
         for name in names:
             child = children.get(name)
-            if not child:
+            if child is None:
                 break
             else:
                 children = child._mock_children.get(name)
                 sig = child._spec_signature
-                if not children:
+                if children is None:
                     break
 
         return sig
--- a/Lib/unittest/test/testmock/testmock.py
+++ b/Lib/unittest/test/testmock/testmock.py
@@ -1344,7 +1344,7 @@ class MockTest(unittest.TestCase):
 
             class Foo:
 
-                def __init__(self): pass
+                def __init__(self, a): pass
                 def meth1(self, a, b): pass
 
         mock_class = create_autospec(Something)
@@ -1354,17 +1354,20 @@ class MockTest(unittest.TestCase):
             m.assert_has_calls([call.meth(1, 2, 3, d=1)])
             m.assert_has_calls([call.meth(1, 2, 3, 1)])
 
-        mock_class = create_autospec(Something)
+        mock_class.reset_mock()
 
         for m in [mock_class, mock_class()]:
             self.assertRaises(AssertionError, m.assert_has_calls, [call.Foo()])
-            m.Foo().meth1(1, 2)
-            m.assert_has_calls([call.Foo(), call.Foo().meth1(1, 2)])
+            m.Foo(1).meth1(1, 2)
+            m.assert_has_calls([call.Foo(1), call.Foo(1).meth1(1, 2)])
+            m.Foo.assert_has_calls([call(1), call().meth1(1, 2)])
 
-        mock_class = create_autospec(Something)
+        mock_class.reset_mock()
+
         invalid_calls = [call.meth(1),
                          call.non_existent(1),
-                         call.Foo().non_existent(1)]
+                         call.Foo().non_existent(1),
+                         call.Foo().meth(1, 2, 3, 4)]
 
         for kall in invalid_calls:
             self.assertRaises(AssertionError,
--- a/Lib/unittest/test/testmock/testmock.py
+++ b/Lib/unittest/test/testmock/testmock.py
@@ -1336,6 +1336,38 @@ class MockTest(unittest.TestCase):
                         )
 
 
+    def test_assert_has_calls_nested_spec(self):
+        class Something:
+
+            def __init__(self, a): pass
+            def meth(self, a, b, c, d=None): pass
+
+            class Foo:
+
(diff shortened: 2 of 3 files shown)


Lib/unittest/mock.py
```diff
@@ -737,14 +737,47 @@ def _format_mock_failure_message(self, args, kwargs):
...
     def _call_matcher(self, _call):
         """
         Given a call (or simply an (args, kwargs) tuple), return a
         comparison key suitable for matching with other calls.
         This is a best effort method which relies on the spec's signature,
         if available, or falls back on the arguments themselves.
         """
-        sig = self._spec_signature
+
+        if isinstance(_call, tuple) and len(_call) > 2:
```
gpshead, line 777: Q: Is being of `tuple` type actually important, or is any sequence okay?
My assumption: `tuple` is easy to express and being an internal API we may be able to guarantee the type so this could be fine.

Lib/unittest/test/testmock/testmock.py
```diff
@@ -1336,6 +1336,46 @@ def test_assert_has_calls(self):
...
+                def meth1(self, a, b): pass
+
+        mock = create_autospec(Something)
+        mock.meth(1, 2, 3, d=1)
+
+        mock.assert_has_calls([call.meth(1, 2, 3, d=1)])
+        mock.assert_has_calls([call.meth(1, 2, 3, 1)])
+
+        mock = create_autospec(Something)
+        mock.Foo().meth1(1, 2)
```
gpshead: You cover two cases here `mocked.meth(...)` and `mocked.Foo().meth1(...)` that this PR fixes the behavior of.  Good!  Lets also cover the cases that I believe were already working for completeness sake:

`mocked().meth(...)` and `mocked().Foo().meth1(...)`.

Lib/unittest/test/testmock/testmock.py
```diff
@@ -1336,6 +1336,46 @@ def test_assert_has_calls(self):
...
+        class Something:
+
+            def __init__(self, a): pass
+            def meth(self, a, b, c, d=None): pass
+
+            class Foo:
+
+                def meth1(self, a, b): pass
+
+        mock = create_autospec(Something)
```
gpshead: given we are testing the mock module, lets not assign to a local name `mock`.  perhaps call this `mocked` or `mock_something`.

Lib/unittest/test/testmock/testmock.py
tirkarthi: Thanks, I have added it. I also made a mistake where `Something.Foo` was not having a constructor and `Something.Foo.meth1` was using `Something.Foo` with signature `(*args, **kwargs)`. So I have added a single argument constructor for this to make sure I am using correct signature of meth1. 

I also noted there was a logical mistake in my iteration where I was using `self._mock_children.get('Foo')` and was again calling `Foo._mock_children.get('Foo')` inside the loop. Due to above constructor mistake accepting `(*args, **kwargs)` it was not caught. There is also an internal implementation where unless an attribute is accessed spec stores `_SpecState` object for the key and that also now has a test.

Lib/unittest/mock.py
tirkarthi: Yes, any sequence can be used. It should be a two element sequence args and kwargs or three element sequence in case of name being present. But most examples and tests construct `call` which inherits from tuple or tuple as a recommended one.

```
>>> from unittest.mock import Mock
>>> m = Mock()
>>> m(1)
<Mock name='mock()' id='4554182024'>
>>> m.assert_has_calls([[(1,), {}]])
```

cjw296: This feels like it would be sensible to just state and ensure that these are only ever `_Call` objects.

tirkarthi:
@mariocj89 Would be helpful to have your review of this.

The main thing I feel is that this lacks tests. I have added some and fixed some bugs in my own test. Also if there is a good way to get to the mock's spec from it's that has parenthesis wrapping around it for nested calls like Foo().Bar().Spam then the logic to get the correct spec in the helper function can be simplified.

Lib/unittest/test/testmock/testmock.py
cjw296: Do all the things you've discovered now have test coverage? If so, can you point me at the tests?

Lib/unittest/test/testmock/testmock.py
cjw296: ...or `mock_`.

cjw296:
What tests do you feel this lacks? What's stopping them being added?

tirkarthi:
I added some more tests. My general concern was that getting to correct spec for nested calls since I was doing text manipulation to generate the list of names to lookup. I have used different signatures for all the methods and checked for non existent attributes in nested calls. I think it's good now. 

Please review the changes. Thanks.

Lib/unittest/mock.py
cjw296: @tirkarthi - thoughts on my comment here ^^

tirkarthi: Though this is an internal method there is a note on the docstring of `_call_matcher` that it could be a tuple. It's not specified in docs that this could be a tuple. I haven't seen a usecase of it in other projects too. There is a test for this usecase but it's not using autospec so it didn't fail with my PR I guess. Relevant test using tuples https://github.com/python/cpython/blob/9fe42b49c79c453d905d0395150ba0607fbab18b/Lib/unittest/test/testmock/testmock.py#L1265

```
Given a call (or simply an (args, kwargs) tuple), return a
comparison key suitable for matching with other calls.
```

There could be a below test that passes with checking `_call` to be a tuple.

```python
    def test_assert_has_calls_tuple(self):
        class Something:

            def __init__(self): pass
            def meth(self, a, b, c, d=None): pass

        m = create_autospec(Something)
        m.meth(1, 2, 3, 1)
        m.assert_has_calls([('meth', (1, 2, 3, 1), {})])
```

On having a strict check of only using `_Call` it will fail since it will go on to use the constructor signature.

```
./python.exe -m unittest Lib.unittest.test.testmock.testmock.MockTest.test_assert_has_calls_tuple
F
FAIL: test_assert_has_calls_tuple (Lib.unittest.test.testmock.testmock.MockTest)
  File "/Users/karthikeyansingaravelan/stuff/python/cpython/Lib/unittest/test/testmock/testmock.py", line 1387, in test_assert_has_calls_tuple
    m.assert_has_calls([('meth', (1, 2, 3, 1), {})])
  File "/Users/karthikeyansingaravelan/stuff/python/cpython/Lib/unittest/mock.py", line 876, in assert_has_calls
    raise AssertionError(
Expected: [('meth', (1, 2, 3, 1), {})]
Actual: [call.meth(1, 2, 3, 1)].

Ran 1 test in 0.063s

FAILED (failures=1)
```

I am leaning towards removing tuple to support only `_Call` since the error message is also not useful. We can perhaps change it in 3.9 to accept only `_Call` and then revert back if someone complains that it breaks their workflow given there is plenty of time. I am not sure about backporting it to 3.8 though since beta is out. Would also like to know thoughts from @voidspace since this was added in the original commit where mock was added.

cjw296: Ah, if it being `_Call`-only causes problems then that's probably a good reason to leave it as it is..

Lib/unittest/test/testmock/testmock.py
```diff
@@ -1336,6 +1336,54 @@ def test_assert_has_calls(self):
...
+
+            class Foo:
+
+                def __init__(self, a): pass
+                def meth1(self, a, b): pass
+
+        mock_class = create_autospec(Something)
+
+        for m in [mock_class, mock_class()]:
+            m.meth(1, 2, 3, d=1)
```
cjw296, line 1353: I'm confused: for the `mock_class` case, why is this expected to work? `meth` isn't a classmethod...

Lib/unittest/mock.py
```diff
@@ -737,14 +737,47 @@ def _format_mock_failure_message(self, args, kwargs):
         return message % (expected_string, actual_string)
 
 
+    def _get_call_signature_from_name(self, name):
+        """
+        If there is no name like the case for call check against functions
+        with tuples then return the self's spec signature.
```
cjw296: I can't really follow what this sentence is saying, could it be reworded?

Lib/unittest/mock.py
cjw296: Splitting hairs: should the ` and len(_call) > 2` move inside `self._get_call_signature_from_name` and maybe renamed that method `_get_signature_from_call`?

cjw296 (approved):
Only nitpicks left from me :-)

Lib/unittest/test/testmock/testmock.py
tirkarthi: I too had the same confusion :) It seems for both class and instance the signature skips self as in the below example. There is some note in docs at https://docs.python.org/3/library/unittest.mock.html#create-autospec but I am also little surprised at this.

> If a class is used as a spec then the return value of the mock (the instance of the class) will have the same spec. You can use a class as the spec for an instance object by passing instance=True. The returned mock will only be callable if instances of the mock are callable.

```python
from unittest.mock import *
import inspect

class A:

    def meth1(self, a): pass

a = create_autospec(A)
a.meth1(1)
print(inspect.signature(a.meth1))
print(a.mock_calls)

# Use as an object
obj = a()
obj.meth1(1)
print(inspect.signature(obj.meth1))
print(obj.mock_calls)
```

```
./python.exe /tmp/bar.py
(a)
[call.meth1(1)]
(a)
[call.meth1(1)]
```

Similar test : 

https://github.com/python/cpython/blob/0f6f73ff3c2e46144884c4d418e67939a042aa23/Lib/unittest/test/testmock/testhelpers.py#L533-L543

Lib/unittest/mock.py
tirkarthi: I have reworded the comment. Feel free to suggest if it can be rephrased better. An explanation using examples would be as below but I am not sure of using examples in the code comment.

* If an assertion is made against a method like `obj.meth1.assert_has_calls([call(1)]` here `call(1)` object has no name to lookup and hence just return `obj.meth1._spec_signature`. 
* If an assertion is made against the obj like `obj.assert_has_calls([call.meth1(1)])` here `call.meth1(1)` will have name 'meth1' that should be used to return signature of `obj.meth1` unlike the bug where `obj` signature was returned.

Thanks

Lib/unittest/test/testmock/testmock.py
cjw296: It's unrelated to this PR, but this just feels like a bug to me. If you've autospecced from a class, the mock shouldn't have things that behave like classmethods if those things aren't classmethods on the object the spec is from.

The "similar test" you've included is just as confusing: why should the mock have `one`, `two` or `three` attributes when the original does not? @voidspace?

Lib/unittest/mock.py
```diff
@@ -739,9 +739,10 @@ def _format_mock_failure_message(self, args, kwargs):
...
     def _get_call_signature_from_name(self, name):
         """
-        If there is no name like the case for call check against functions
-        with tuples then return the self's spec signature.
-        If the name is not empty then remove () and split by '.' to get
+        * If call objects are asserted against a method/function like obj.meth1
+        then there could be no name for the call object to lookup. Hence just
+        return the spec_signature of the method/function being asserted against.
+        * If the name is not empty then remove () and split by '.' to get
         list of names to iterate through the children until a potential
```
cjw296, line 746: much clearer, thanks!

Lib/unittest/test/testmock/testmock.py
tirkarthi: one, two and three are methods of `SomeClass` that is inherited by `Sub`

cjw296: Ah! That makes more sense :-)

ned-deily:
This PR seems to have stalled.  Any core developer want to follow up on merging it?

gpshead:
PR 15578 was the backport to 3.8.
PR 15577 was the backport to 3.7.

PR #13263, opened 2019-05-12
Remove workaround for defaults in namedtuple now that we have the defaults parameter

andresdelfino:
The intention is to prevent the previous more verbose style now that a more fitting solution has been implemented.

### DIFF
--- a/Doc/library/collections.rst
+++ b/Doc/library/collections.rst
@@ -1017,14 +1017,6 @@ fields:
 .. versionchanged:: 3.5
    Property docstrings became writeable.
 
-Default values can be implemented by using :meth:`~somenamedtuple._replace` to
-customize a prototype instance:
-
-    >>> Account = namedtuple('Account', 'owner balance transaction_count')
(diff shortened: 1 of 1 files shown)


remilapeyre (approved):
Thanks !

csabella (changes requested):
Can you also remove one of the blank lines so that only one is left?  Thanks.

csabella:
Although I can see some value in leaving this example (that is, the creation of a Factory namedtuple), the intent of the example was to show the use of defaults, so I agree it should be removed.

andresdelfino:
> Can you also remove one of the blank lines so that only one is left? Thanks.

Sure, sorry about that.

andresdelfino:

csabella (approved):

csabella:
@andresdelfino Thank you for the contribution.  I'm going to just apply this to 3.8 and leave the example in for 3.7 when `defaults` was added.

PR #13264, opened 2019-05-12
Remove outdated time.monotonic reference

bsolomon1124:
Per ae58649, ``time.monotonic`` is always available, making the old note outdated.

### DIFF
--- a/Doc/library/sched.rst
+++ b/Doc/library/sched.rst
@@ -20,8 +20,7 @@ scheduler:
    The :class:`scheduler` class defines a generic interface to scheduling events.
    It needs two functions to actually deal with the "outside world" --- *timefunc*
    should be callable without arguments, and return  a number (the "time", in any
-   units whatsoever). If time.monotonic is not available, the *timefunc* default
-   is time.time instead. The *delayfunc* function should be callable with one
+   units whatsoever).  The *delayfunc* function should be callable with one
    argument, compatible with the output of *timefunc*, and should delay that many
    time units. *delayfunc* will also be called with the argument ``0`` after each
    event is run to allow other threads an opportunity to run in multi-threaded


Doc/library/sched.rst
```diff
@@ -20,8 +20,8 @@ scheduler:
    The :class:`scheduler` class defines a generic interface to scheduling events.
    It needs two functions to actually deal with the "outside world" --- *timefunc*
    should be callable without arguments, and return  a number (the "time", in any
-   units whatsoever). If time.monotonic is not available, the *timefunc* default
```
pganssle, line 23: Is it necessary to retain this sentence at all? It's already conveyed in the function prototype, and no special note is made for `delayfunc`.

I think maybe just remove the sentence entirely now that `timefunc`'s default value is no longer conditional.

Not sure if it's worth adding a `.. versionchanged:: 3.5` that says that `time.monotonic` is the default on all platforms now, since that's essentially covered in `time.monotonic` already, but I don't think it would *hurt*.

bsolomon1124: @pganssle Yes, I think that's all agreeable.  Squashed the recent commit.

pganssle (approved):

pganssle:
Not sure what the rules are for semi-trivial things like this. Seems like at least skipping an issue is fine for this, but it wouldn't hurt to add a news blurb.

csabella:
Asking for review from @vstinner as he made the original change for time.monotonic to be always available.

Doc/library/sched.rst
```diff
@@ -20,8 +20,7 @@ scheduler:
    The :class:`scheduler` class defines a generic interface to scheduling events.
    It needs two functions to actually deal with the "outside world" --- *timefunc*
    should be callable without arguments, and return  a number (the "time", in any
```
vstinner, line 22: I dislike "should" in technical documentations. I suggest to replace "should be" with "is". Like "timefunc is a callback without parameter which returns a number (...)".

bsolomon1124: I didn't write that snippet in the first place, but am glad to change it now 😉 .

I added some additional touchups as well and expanded the scope of the PR slightly.

Doc/library/sched.rst
```diff
@@ -19,21 +19,24 @@ scheduler:
...
+      The *timefunc* and *delayfunc* parameters are optional.
 
    .. versionchanged:: 3.3
-      :class:`scheduler` class can be safely used in multi-threaded
+      :class:`scheduler` can be safely used in multi-threaded
       environments.
 
+   .. versionchanged:: 3.5
+      *timefunc* will always default to :func:`time.monotonic`,
+      given that the function is available on all platforms.
```
vstinner: That's wrong. time.monotonic is used since Python 3.3:
https://docs.python.org/release/3.3.7/library/sched.html#sched.scheduler

vstinner: Complete the 3.3 note, or just remove this 3.5 note.

bsolomon1124: Yes, but I'm using _always_ to denote that only as of 3.5 is `time.monotonic()` guaranteed to be what is actually used as the callable.

https://github.com/python/cpython/blob/bf9cccb2b54ad2c641ea78435a8618a6d251491e/Lib/sched.py#L38

I think it would be most straightforward to just remove the 3.5 note to avoid confusion.

Doc/library/sched.rst
```diff
@@ -114,9 +113,9 @@ Scheduler Objects
    function passed to the constructor) for the next event, then execute it and so
    on until there are no more scheduled events.
 
-   If *blocking* is false executes the scheduled events due to expire soonest
-   (if any) and then return the deadline of the next scheduled call in the
-   scheduler (if any).
+   If *blocking* evaluates to ``False``, execute the scheduled events due to
```
vstinner: "is false" is commonly used in the documentation to mean "evaluates to False". "is false" sounds correct to me.

Doc/library/sched.rst
```diff
@@ -131,8 +130,10 @@ Scheduler Objects
    .. versionchanged:: 3.3
       *blocking* parameter was added.
 
+Instance attributes (read-only):
```
vstinner: I prefer to keep the "read-only" mention on the attribute doc.

bsolomon1124: Both changes made/reverted.

vstinner:
@csabella: "Asking for review from @vstinner as he made the original change for time.monotonic to be always available."

I'm not a native English speaker, so I prefer to skip my turn for doc reviews ;-)

pganssle (approved):
This PR seems to have grown in scope, but these changes look fine to me (though also the old way mostly looked fine to me as well).

Doc/library/sched.rst
```diff
@@ -19,19 +19,18 @@ scheduler:
 
    The :class:`scheduler` class defines a generic interface to scheduling events.
    It needs two functions to actually deal with the "outside world" --- *timefunc*
```
csabella, line 21: So, this sentence isn't exactly proper English.  (I'm using the original version)

``It needs two functions to actually deal with the “outside world” — timefunc should be callable without arguments, and return a number (the “time”, in any units whatsoever). ``

Starting a sentence with `It needs two functions - ` would imply to me that I would then see those two functions after the dash and within the same sentence, which in this case would require more of a rewrite.  However, I believe this is why Victor asked for additional review because it does take a few readings of the paragraph to grasp the meaning.

Based on that, my suggestion is:

> The :class:`scheduler` class defines a generic interface to scheduling events.
  It needs two functions to actually deal with the "outside world".  *timefunc* is a 
 callable without arguments that returns a number representing the "time" (in any
  units whatsoever).  *delayfunc* is a callable with one argument, which is a 
  number compatible with the output of *timefunc*, and that delays 
  for that many  time units. *delayfunc* will also be called with the argument ``0`` after each
   event is run to allow other threads an opportunity to run in multi-threaded
   applications.

Doc/library/sched.rst
```diff
@@ -66,17 +65,17 @@ Scheduler Objects
...
-   with the return value of the *timefunc* function passed  to the constructor.
+   Schedule a new event. The *time* argument is a numeric type compatible
+   with the return value of the *timefunc* function passed to the constructor.
    Events scheduled for the same *time* will be executed in the order of their
    *priority*. A lower number represents a higher priority.
 
    Executing the event means executing ``action(*argument, **kwargs)``.
    *argument* is a sequence holding the positional arguments for *action*.
-   *kwargs* is a dictionary holding the keyword arguments for *action*.
+   *kwargs* is a :class:`dict` holding the keyword arguments for *action*.
```
csabella: I don't think it's necessary to add this markup.  If someone is reading about the `sched` module, then the word dictionary should be enough for them to understand the meaning.

Doc/library/sched.rst
```diff
@@ -133,6 +132,6 @@ Scheduler Objects
 
 .. attribute:: scheduler.queue
 
-   Read-only attribute returning a list of upcoming events in the order they
+   Read-only attribute; a :class:`list` of upcoming events in the order they
```
csabella: Again, I don't think this markup is necessary.  However, I can see your reasoning with wanting to remove the word 'returning'.  Maybe use 'providing' or 'containing' instead?

Doc/library/sched.rst
```diff
@@ -66,17 +65,17 @@ Scheduler Objects
...
 
    Executing the event means executing ``action(*argument, **kwargs)``.
    *argument* is a sequence holding the positional arguments for *action*.
-   *kwargs* is a dictionary holding the keyword arguments for *action*.
+   *kwargs* is a :class:`dict` holding the keyword arguments for *action*.
 
-   Return value is an event which may be used for later cancellation of the event
-   (see :meth:`cancel`).
+   The return value is an event which may be used for later cancellation
+   of the event (see :meth:`cancel`).
```
csabella: Maybe just use `Return an event which`

Doc/library/sched.rst
bsolomon1124: Again, I didn't write any of the `sched` docs in the first place but am glad to help touch it up now 😆

bsolomon1124: In 99da934, I made them more compartmentalized and also clarified the wording based off your suggestion.  I think we are all on the same page here - trying to emphasize that `delayfunc()` effectively takes the numeric output produced by `timefunc()` as its argument in doing the actual scheduling.

bsolomon1124:
Hi @csabella, do the grammatical changes here look okay to you?

csabella:
@bsolomon1124, thanks for working on this.  I've gone back through the commit history and, since no one has changed the wording of this paragraph in the past even when adding additional phrases, I'm going to trust their judgment and backtrack on the suggestions that I made earlier.  I think your original idea of just removing the line `If time.monotonic is not available, the *timefunc* default is time.time instead.` to match the logic change in the program should be the only line modified for this.  Thank you!

bsolomon1124:
Okay @csabella , I've reverted to that one change. Perhaps if anyone still thinks the rest of the doc could use some grammatical updates, that could be addressed in a separate PR.

csabella (approved):

csabella:
@bsolomon1124, thank you for your contribution and your patience with the process.

PR #13265, opened 2019-05-12
[2.7] bpo-14353: Fix detection of bind_textdomain_codeset in libintl

abadger:
In Python-2.7, we were only searching for bind_textdomain_codeset in
libc.  We should have also checked for it in libintl.  This change from
Mel Flynn https://bugs.python.org/file24918/python27-configure.in.patch
fixes that.

This problem does not exist in the Python-3.x tree so this is not a backport but a fresh pull request against the 2.7 branch.

Fixes: https://bugs.python.org/issue14353

https://bugs.python.org/issue14353


abadger:
\cc @benjaminp PR for the configure script detection of bind_textdomain_codeset.

benjaminp:
Okay, but `configure` needs to be regenerated before this is merged.

abadger:
Thanks.  New commit with regenerated configure and pyconfig.h.in

PR #13268, opened 2019-05-12
Name individual Travis CI jobs

GPHemsley:
This adds names to each Travis CI job so that their purpose is more obvious.


GPHemsley:
@methane Thoughts?

asvetlov (approved):
I like it

Carreau (approved):

PR #13269, opened 2019-05-12
gh-66333: Allow to break into pdb with Ctrl-C for all the commands t…

devjoe:
This PR comes from https://bugs.python.org/issue22135.
I mainly use registered `signal_handler` to handle Ctrl-C behavior and add a unit test.

https://bugs.python.org/issue22135

* Issue: gh-66333

### DIFF
--- a/Lib/pdb.py
+++ b/Lib/pdb.py
@@ -158,6 +158,7 @@ class Pdb(bdb.Bdb, cmd.Cmd):
             pass
         self.allow_kbdint = False
         self.nosigint = nosigint
+        self.sigint_triggered_cmd = None
 
         # Read $HOME/.pdbrc and ./.pdbrc
         self.rcLines = []
@@ -188,9 +189,29 @@ class Pdb(bdb.Bdb, cmd.Cmd):
     def sigint_handler(self, signum, frame):
         if self.allow_kbdint:
             raise KeyboardInterrupt
-        self.message("\nProgram interrupted. (Use 'cont' to resume).")
-        self.set_step()
-        self.set_trace(frame)
+        try:
+            self.message("\nReceived SIGINT signal\nResuming Pdb ... (please wait)")
+        except RuntimeError:
+            # 'print' is not safe in signal handlers, check issue24283
+            # if we encounter this issue, we just silently resume the execution
+            pass
+        if self.sigint_triggered_cmd == 'continue':
+            self.set_trace(frame)
+        else:
+            self.set_step()
+
+    def register_sigint_handler(self, triggered_cmd):
+        if self.nosigint:
+            return
+        self.sigint_triggered_cmd = triggered_cmd
+        try:
+            Pdb._previous_sigint_handler = \
+                signal.signal(signal.SIGINT, self.sigint_handler)
+        except ValueError:
+            # ValueError happens when any do_xxx() is invoked from
+            # a non-main thread in which case we just continue without
+            # SIGINT set.
+            pass
 
     def reset(self):
         bdb.Bdb.reset(self)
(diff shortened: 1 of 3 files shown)


Lib/pdb.py
```diff
@@ -976,6 +995,7 @@ def do_until(self, arg):
         or equal to that is reached.  In both cases, also stop when
         the current frame returns.
         """
+        self.register_sigint_handler(triggered_cmd='until')
```
blueyed: This should be probably just before the `set_until` below.

Lib/pdb.py
```diff
@@ -188,9 +189,27 @@ def __init__(self, completekey='tab', stdin=None, stdout=None, skip=None,
     def sigint_handler(self, signum, frame):
         if self.allow_kbdint:
             raise KeyboardInterrupt
-        self.message("\nProgram interrupted. (Use 'cont' to resume).")
+        try:
+            self.message("\nProgram interrupted (please wait).")
```
blueyed: I think the message could/should be clearer (which is also the case for the current one): the program is not interrupted (e.g. a call to `time.sleep` will finish), but only the debugger will stop at the next step.
Therefore "please wait" is already better than the current "Program interrupted. (Use 'cont' to resume)." (where it might take a while until you can use "cont").

Lib/pdb.py
```diff
@@ -188,9 +189,27 @@ def __init__(self, completekey='tab', stdin=None, stdout=None, skip=None,
...
+            # 'print' is not safe in signal handlers, check issue24283
+            # if we encounter this issue, we just silently resume the execution
+            pass
         self.set_step()
-        self.set_trace(frame)
+        if self.sigint_triggered_cmd == 'continue':
+            self.set_trace(frame)
+
+    def register_sigint_handler(self, triggered_cmd):
+        if not self.nosigint:
```
blueyed: Could return for `if self.nosigint` at the top to safe a level of indentation.

Lib/pdb.py
```diff
@@ -188,9 +189,27 @@ def __init__(self, completekey='tab', stdin=None, stdout=None, skip=None,
...
+        try:
+            self.message("\nProgram interrupted (please wait).")
+        except RuntimeError:
+            # 'print' is not safe in signal handlers, check issue24283
+            # if we encounter this issue, we just silently resume the execution
+            pass
         self.set_step()
-        self.set_trace(frame)
+        if self.sigint_triggered_cmd == 'continue':
+            self.set_trace(frame)
```
blueyed, line 199: Why is this only needed / special for "continue"?  Because it might have unset the tracing function (when there are no breakpoints)?

btw: `Pdb/Bdb.set_trace` calls `set_step` itself, so it might not be needed here explicitly before then.

devjoe:
@blueyed thanks for your super fast feedback, will refine this PR in the following days.

Lib/pdb.py
devjoe: You are right.

Lib/pdb.py
devjoe: How about this version?
```
(Pdb) c
^C
Received SIGINT signal
Resuming Pdb ... (please wait)```

Lib/pdb.py
devjoe: Guard Clause. Updated.

Lib/pdb.py
devjoe: Because `set_continue()` does not stop the debugger in an `interaction()` on the next debug event.

Thanks for your suggestion, it works.

devjoe:
PR updated, feel free to me know your thoughts.

devjoe:
@blueyed 🙏

blueyed:
@devjoe 
LGTM (only skimmed it quickly now), but I am not in the position to really judge / acknowledge it unfortunately.

devjoe:
@xdegaye sorry for bothering you, could help to review this PR again? 🙏

PR #13271, opened 2019-05-13
bpo-36896: Clarify that some types constructors are unstable

Carreau:
At least it will represent the current state, which seem to be a
consensus; but is not explicit.

Not sure this requires a News item; it would also be good to add this
info in the constructors docstrings.

https://bugs.python.org/issue36896

### DIFF
--- a/Doc/library/types.rst
+++ b/Doc/library/types.rst
@@ -98,6 +98,9 @@ the types that arise only incidentally during processing such as the
 Typical use of these names is for :func:`isinstance` or
 :func:`issubclass` checks.
 
+
+If you instantiate any of these types, note that signatures may vary between Python versions.
+
 Standard names are defined for the following types:
 
 .. data:: FunctionType
--- a/Misc/NEWS.d/next/Documentation/2019-05-31-10-46-36.bpo-36896.wkXTW9.rst
+++ b/Misc/NEWS.d/next/Documentation/2019-05-31-10-46-36.bpo-36896.wkXTW9.rst
@@ -1,0 +1,2 @@
+Clarify that some types have unstable constructor signature between Python
+versions.


Doc/library/types.rst
```diff
@@ -98,6 +98,10 @@ the types that arise only incidentally during processing such as the
 Typical use of these names is for :func:`isinstance` or
 :func:`issubclass` checks.
 
+.. note::
+
+   These types are not supposed to be instantiated outside of CPython internals and constructor signatures will vary between python versions.
```
csabella: `python` should be capitalized.

Carreau: Oops indeed, good catch thanks, fixed.

encukou:
> current state, which seem to be a consensus; but is not explicit

I'm trying to make the consensus explicit: https://mail.python.org/pipermail/python-dev/2019-May/157795.html
I believe "*not supposed to be instantiated outside of CPython internals*" is too far, with respect to how much code outside CPython is using the constructor.

Carreau:
Thanks @encukou I'm happy to rephrase; 

I 'm just tryin to convey they are supposed to mostly be used for checking isinstance.
I believe like as for AST its unstable; but ast Nodes are undocumented and AFAICT the response is regularly [the same](https://bugs.python.org/issue35894)

Would you prefer something around "stability of constructor signatures is not guaranteed between python versions, not recommend and thus not documented" ?

Doc/library/types.rst
```diff
@@ -98,6 +98,10 @@ the types that arise only incidentally during processing such as the
 Typical use of these names is for :func:`isinstance` or
 :func:`issubclass` checks.
 
+.. note::
```
terryjreedy: Delete this and just add a revised sentence (next comment) to the previous paragraph.
```suggestion

```

Doc/library/types.rst
```diff
@@ -98,6 +98,10 @@ the types that arise only incidentally during processing such as the
...
+
+   These types are not supposed to be instantiated outside of CPython internals and constructor signatures will vary between Python versions.
```
terryjreedy: The 'not supposed to be' part may be true for some of the types not listed, but is definitely not true for at least some of the types.
```suggestion
   If you instantiate any of these types, note that signatures may vary between Python versions.
```

terryjreedy (dismissed):
Add a blurb.  With very few exceptions, there should be one for every bpo issue.  (If multiple PRs for an issue, there would normally still be just one blurb.)

Carreau:

pablogsal (approved):

Misc/NEWS.d/next/Documentation/2019-05-31-10-46-36.bpo-36896.wkXTW9.rst
```diff
@@ -0,0 +1,2 @@
+Clarify that some types have unstable constructor signature between
```
pablogsal: s/between versions/between Python versions/

Carreau: Done.

Carreau:
Thanks !

PR #13272, opened 2019-05-13
bpo-23896: Add a grammar where exec isn't a stmt

isidentical:
Add a grammar where exec isn't a stmt

https://bugs.python.org/issue23896


auvipy (approved):

asottile (approved):
looks good, this is consistent with how [yapf](https://github.com/google/yapf/blob/45cf63a4538a2d5790db5182b84aca8c1b3bc491/yapf/yapflib/pytree_utils.py#L90) and others use this module for parsing py3 syntax

gvanrossum (commented):
Looks fine. I guess this should be used by tools that are actually parsing Python 3... :-)

PR #13283, opened 2019-05-13
bpo-36903: Fix ResourceWarning in test_logging

tirkarthi:
Fix `ResourceWarning` in test_logging by closing the stream used in the logging handler.

https://bugs.python.org/issue36903

### DIFF
--- a/Lib/test/test_logging.py
+++ b/Lib/test/test_logging.py
@@ -684,6 +684,7 @@ class HandlerTest(BaseTest):
 
         self.assertEqual(len(logging._handlers), 0)
         refed_h = _OurHandler()
+        self.addCleanup(refed_h.sub_handler.stream.close)
         refed_h.name = 'because we need at least one for this test'
         self.assertGreater(len(logging._handlers), 0)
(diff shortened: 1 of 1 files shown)


Lib/test/test_logging.py
```diff
@@ -683,7 +682,9 @@ def emit(self, record):
                     self.sub_handler.release()
 
         self.assertEqual(len(logging._handlers), 0)
-        refed_h = _OurHandler()
+        stream = open('/dev/null', 'wt')
+        self.addCleanup(stream.close)
```
vstinner: You can use self.addCleanup(refed_h.stream.close), the .stream attribute is public. No need to modify _OurHandler.

vstinner (approved):
LGTM.

Lib/test/test_logging.py
tirkarthi: Thanks Victor. I think it should have been `self.addCleanup(refed_h.sub_handlerstream.close)` since it's the sub_handler's stream that need to be closed. I have reverted my changes and added your suggestion.

PR #13284, opened 2019-05-13
bpo-36807: Fix typo in NEWS item about IDLE (os.flush() should be os.fsync())

gvanrossum:
Typo reported by Dan Halbert. Should be incorporated into the backports as well.

### DIFF
--- a/Misc/NEWS.d/next/IDLE/2019-05-05-16-27-53.bpo-13102.AGNWYJ.rst
+++ b/Misc/NEWS.d/next/IDLE/2019-05-05-16-27-53.bpo-13102.AGNWYJ.rst
@@ -1,1 +1,1 @@
-When saving a file, call os.flush() so bits are flushed to e.g. USB drive.
+When saving a file, call os.fsync() so bits are flushed to e.g. USB drive.


Misc/NEWS.d/next/IDLE/2019-05-05-16-27-53.bpo-13102.AGNWYJ.rst
```diff
@@ -1 +1 @@
-When saving a file, call os.flush() so bits are flushed to e.g. USB drive.
+When saving a file, call os.fsync() so bits are flushed to e.g. USB drive.
```
terryjreedy, line 1: We missed the type in the file title.   13102 in the filename was the original pr #.  The bpo in the PR, 36807,  is correct and should have been the bpo # in the filename.  I will fix this in the 3.8 and 3.7 changelogs.
#15042 for 3.8.

gvanrossum: Thanks!

PR #13290, opened 2019-05-13
bpo-36894: Fix regression in test_multiprocessing_spawn (no tests run on Windows)

pitrou:
This regression was introduced in [bpo-36867](https://bugs.python.org/issue36867).

https://bugs.python.org/issue36894


pitrou:
@pierreglaser

pierreglaser (approved):

PR #13293, opened 2019-05-13
[2.7] bpo-36807: When saving a file in IDLE, call flush and fsync (GH…

terryjreedy:
…-13102)

(cherry picked from commit 4f098b3)


terryjreedy:
I updated and compiled my local 2.7 branch; ran IDLE; and tested save_as and F5 autosave and run.

PR #13296, opened 2019-05-13
bpo-36909 Simplify the ``LastUpdatedOrderedDict`` example recipe

wimglenn:
Recipe in https://docs.python.org/3/library/collections.html#ordereddict-examples-and-recipes

This example seems to use `super()` unnecessarily. ~I didn't bother to create a bpo issue number because it seems a trivial docs change to me, but let me know if it's required.~

Please add skip news label.

Cheers!

https://bugs.python.org/issue36909

### DIFF
--- a/Doc/library/collections.rst
+++ b/Doc/library/collections.rst
@@ -1141,7 +1141,7 @@ original insertion position is changed and moved to the end::
 
         def __setitem__(self, key, value):
             super().__setitem__(key, value)
-            super().move_to_end(key)
+            self.move_to_end(key)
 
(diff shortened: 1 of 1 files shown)


tonybaloney:
Thanks for your contribution

Yes, it is required. Please add a bpo reference, https://devguide.python.org/pullrequest/

wimglenn:
@tonybaloney   Added a bpo. What about news entry? Too trivial, no?

tonybaloney:
Yes, too trivial. there's a special tag that the merger can add

rhettinger (approved):

PR #13297, opened 2019-05-13
bpo-36908: 'This module is always available' isn't helpful.

nedbat:
https://bugs.python.org/issue36908

### DIFF
--- a/Doc/library/cmath.rst
+++ b/Doc/library/cmath.rst
@@ -6,13 +6,12 @@
 
 --------------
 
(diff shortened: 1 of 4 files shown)


eamanu (approved):
LGTM. :-) IMHO we need to create a section with the phrase 

```
The next modules: math, sys ... are always available. This mean that ...
```
I will mention it on bpo

kushaldas (approved):
This explanation is good. Thank you for making it easier to understand.

PR #13300, opened 2019-05-14
bpo-30535: Explicitly note that sys.meta_path is not empty

Windsooon:
Explicitly note that sys.meta_path is not empty

https://bugs.python.org/issue30535

### DIFF
--- a/Doc/library/sys.rst
+++ b/Doc/library/sys.rst
@@ -926,7 +926,8 @@ always available.
 
     A list of :term:`meta path finder` objects that have their
     :meth:`~importlib.abc.MetaPathFinder.find_spec` methods called to see if one
-    of the objects can find the module to be imported. The
+    of the objects can find the module to be imported. By default, it holds entries
+    to handle the standard kinds of modules (.py files, extension modules...). The
     :meth:`~importlib.abc.MetaPathFinder.find_spec` method is called with at
     least the absolute name of the module being imported. If the module to be
     imported is contained in a package, then the parent package's :attr:`__path__`


Doc/library/sys.rst
```diff
@@ -926,7 +926,8 @@ always available.
 
     A list of :term:`meta path finder` objects that have their
     :meth:`~importlib.abc.MetaPathFinder.find_spec` methods called to see if one
-    of the objects can find the module to be imported. The
+    of the objects can find the module to be imported. by default, holds entries
```
jdemeyer: This is not grammatically correct: a sentence should start with a capital letter and contain a verb.

jdemeyer (changes requested):

Doc/library/sys.rst
Windsooon: Thank you, how about
"By default, it holds entries to handle the standard kinds of modules (.py files, extension modules...)."

jdemeyer (approved):

Doc/library/sys.rst
```diff
@@ -926,7 +926,8 @@ always available.
 
     A list of :term:`meta path finder` objects that have their
     :meth:`~importlib.abc.MetaPathFinder.find_spec` methods called to see if one
-    of the objects can find the module to be imported. The
+    of the objects can find the module to be imported. By default, it holds entries
+    to handle the standard kinds of modules (.py files, extension modules...). The
```
brettcannon, line 930:
```suggestion
   that implement Python's default import semantics. The
```

brettcannon (changes requested):
Technically there aren't different kinds of modules, just different ways of storing code. I have suggested a tweak to the wording.

Doc/library/sys.rst
Windsooon: @brettcannon Do you think we should add examples after that. Like:

> that implement Python's default import semantics. (BuiltinImporter, FrozenImporter, and PathFinder)

brettcannon: No, as that's too easy of a thing to forget to update. Plus there's no guarantee that every interpreter will set up `sys.meta_path` the same.

brettcannon (changes requested):
I don't think the docs should get too specific as interpreters can choose what to put into `sys.meta_path`.

csabella:
@Windsooon, please take a look at the review comments.  Thanks!

AlexWaygood:
Changes were requested by a core dev over two years ago, but have not been made. I am therefore closing this PR.

@Windsooon, if you're still interested in working on this, feel free to open a new PR. Alternatively, ping me, and I'll happily reopen this PR. Thanks! 🙂

PR #13305, opened 2019-05-14
bpo-23378: Add an extend action to argparse

isidentical:
Add an extend action to argparse

https://bugs.python.org/issue23378

### DIFF
--- a/Doc/library/argparse.rst
+++ b/Doc/library/argparse.rst
@@ -797,6 +797,15 @@ how the command-line arguments should be handled. The supplied actions are:
     >>> parser.parse_args(['--version'])
     PROG 2.0
 
+* ``'extend'`` - This stores a list, and extends each argument value to the
+  list.
+  Example usage::
+
+    >>> parser = argparse.ArgumentParser()
+    >>> parser.add_argument("--foo", action="extend", nargs="+", type=str)
+    >>> parser.parse_args("--foo f1 --foo f2 f3 f4".split())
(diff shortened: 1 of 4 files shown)


asottile (commented):
patch looks good, are there other `nargs=...` that this makes sense with? should it produce an error in the nonsensical cases? should it default to `nargs='+'`?

isidentical:
`nargs=N` doesn't make any sense (it doesn't need `extend` but `extend` can be used like `append`).  but i dont think we should produce a `ValueError` etc.

furkanonder (approved):
Nice fix.

Doc/library/argparse.rst
```diff
@@ -797,6 +797,15 @@ how the command-line arguments should be handled. The supplied actions are:
     >>> parser.parse_args(['--version'])
     PROG 2.0
 
+* ``'extend'`` - This stores a list, and extends each argument value to the
+  list.
+  Example usage::
+
+    >>> parser = argparse.ArgumentParser()
+    >>> parser.add_argument("--foo", action="extend", nargs="+", type=str)
+    >>> parser.parse_args("--foo f1 --foo f2 f3 f4".split())
```
gvanrossum: I would remove the `split()` call and just write `parser.parse_args("--foo",  "f1", "--foo", "f2", "f3", " f4")`

isidentical: resolved

gvanrossum (approved):
Yup!

PR #13307, opened 2019-05-14
Doc: Update pip and setuptools when creating the virtual environment

matrixise:
Add a new pip install before `sphinx` etc.. because we should use the last version of `pip` and `setuptools`


JulienPalard (approved):
LGTM

PR #13308, opened 2019-05-14
bpo-36904: new function _PyStack_DictAsVector

jdemeyer:
Provide `_PyStack_DictAsVector` as safe and more efficient replacement for `_PyStack_UnpackDict`.

https://bugs.python.org/issue36904

### DIFF
--- a/Objects/call.c
+++ b/Objects/call.c
@@ -535,19 +535,16 @@ _PyMethodDef_RawFastCallDict(PyMethodDef *method, PyObject *self,
 
     case METH_FASTCALL | METH_KEYWORDS:
     {
-        PyObject *const *stack;
-        PyObject *kwnames;
         _PyCFunctionFastWithKeywords fastmeth = (_PyCFunctionFastWithKeywords)(void(*)(void))meth;
 
-        if (_PyStack_UnpackDict(args, nargs, kwargs, &stack, &kwnames) < 0) {
+        PyObject *kwnames;
+        args = _PyStack_DictAsVector(args, nargs, kwargs, &kwnames);
+        if (args == NULL) {
             goto exit;
         }
 
-        result = (*fastmeth) (self, stack, nargs, kwnames);
-        if (stack != args) {
-            PyMem_Free((PyObject **)stack);
-        }
-        Py_XDECREF(kwnames);
+        result = (*fastmeth) (self, args, nargs, kwnames);
+        _PyStack_DictAsVector_Free(kwnames);
         break;
     }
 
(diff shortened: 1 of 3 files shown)


jdemeyer:
`skip news` because this is an implementation detail

CC @encukou @markshannon @scoder @vstinner

jdemeyer:
I'm wondering whether we should use a special `tuple` subclass for this. It wouldn't really change much in practice, but it would feel more justified to hack with the tuple size.

Objects/call.c
```diff
@@ -1298,59 +1295,80 @@ _PyStack_AsDict(PyObject *const *values, PyObject *kwnames)
...
-        return -1;
+    /* Allocate one big tuple for everything, plus one sentinel NULL */
+    Py_ssize_t n = 2 * nkwargs + nargs + 1;
+    PyObject *kwtuple = PyTuple_New(n);
+    *p_kwnames = kwtuple;
+    if (kwtuple == NULL) {
+        return NULL;
     }
+    /* Truncate the tuple to the desired length */
+    Py_SIZE(kwtuple) = nkwargs;
```
encukou, line 1331: That doesn't look safe at all. Won't such a tuple join the wrong freelist in `tupledealloc`?

jdemeyer:
I didn't think of the freelist but it's true that I already had bad feelings about messing with the tuple size like that. What do you think of my proposal (see above) of making a special `tuple` subclass for this?

encukou:
For a good answer I'd need to write the code :)
Generally, this sounds like an optimization, so consider doing it in two steps:
* Fix the issue simply and robustly (seems to need two allocations and a bunch of refcount poking)
* Afterwards, optimize and measure

jdemeyer:
> For a good answer I'd need to write the code :)

*write* the code or just *see* the code (written by me for example)?

> Generally, this sounds like an optimization, so consider doing it in two steps:
> 
>     * Fix the issue simply and robustly (seems to need two allocations and a bunch of refcount poking)
> 
>     * Afterwards, optimize and measure

This two-step process is going to take more work: it requires to discuss two API designs and implementations instead of one (with the first one going to be thrown away anyway once we do the second).

So, I'd like to do it in one step if possible.

jdemeyer:
>     * bunch of refcount poking

On the other hand, adding the "refcount poking" part is pretty simple. So if you agree with doing just that manually (without changing `_PyStack_UnpackDict` at all), then I'm OK with that.

encukou:
> write the code or just see the code (written by me for example)?

For a relatively complex optimization? *benchmark* the code, before and after.

> This two-step process is going to take more work: it requires to discuss two API designs and implementations instead of one (with the first one going to be thrown away anyway once we do the second).

But without the first, it'd be hard to argue that the tuple subclass actually brings any benefit. Tuple subclasses behave quite differently from exact tuples (they don't use freelists, for one).

> On the other hand, adding the "refcount poking" part is pretty simple. So if you agree with doing just that manually (without changing _PyStack_UnpackDict at all), then I'm OK with that.

Definitely! All I ask for before the optimization is a correct baseline.

It might then very well turn out that the optimization is unnecessary.

jdemeyer:
I decided to drop this idea. It's more controversial than I initially thought and I see why (I can't disagree).

For reference: #14517 optimized `_PyStack_UnpackDict` in other ways, in particular to create a fast path in the case of no keyword arguments.

PR #13310, opened 2019-05-14
Reference zipimport source code from docs

tirkarthi:
zipimport module is written in Python so reference the source code from docs.


tirkarthi:
cc @serhiy-storchaka : zipimport was rewritten in Python with https://github.com/python/cpython/pull/6809/

serhiy-storchaka (approved):

matrixise:
@serhiy-storchaka Do you want to merge this PR?

PR #13313, opened 2019-05-14
bpo-36916: Swallow unhandled exception

asvetlov:
The PR fixes a message about an unhandled exception in a task when `writer.write()` is used without await but `writer.drain()` fails with an exception.

https://bugs.python.org/issue36916

### DIFF
--- a/Lib/asyncio/streams.py
+++ b/Lib/asyncio/streams.py
@@ -335,7 +335,7 @@ def _swallow_unhandled_exception(task):
     # stream.drain() was paused and resumed with an exception
     task.exception()
 
- 
+
 class StreamWriter:
     """Wraps a Transport.
 
--- a/Lib/asyncio/streams.py
+++ b/Lib/asyncio/streams.py
@@ -329,6 +329,13 @@ class StreamReaderProtocol(FlowControlMixin, protocols.Protocol):
             closed.exception()
 
 
+def _swallow_unhandled_exception(task):
+    # Do a trick to suppress unhandled exception
+    # if stream.write() was used without await and
+    # stream.drain() was paused and resumed with an exception
+    task.exception()
+
+ 
 class StreamWriter:
     """Wraps a Transport.
 
@@ -393,7 +400,9 @@ class StreamWriter:
                 # fast path, the stream is not paused
                 # no need to wait for resume signal
                 return self._complete_fut
-        return self._loop.create_task(self.drain())
+        ret = self._loop.create_task(self.drain())
+        ret.add_done_callback(_swallow_unhandled_exception)
+        return ret
 
     def write_eof(self):
         return self._transport.write_eof()
--- a/Lib/test/test_asyncio/test_streams.py
+++ b/Lib/test/test_asyncio/test_streams.py
@@ -851,6 +851,8 @@ os.close(fd)
         # where it never gives up the event loop but the socket is
         # closed on the  server side.
 
+        messages = []
+        self.loop.set_exception_handler(lambda loop, ctx: messages.append(ctx))
         q = queue.Queue()
 
         def server():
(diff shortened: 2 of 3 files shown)


Lib/asyncio/streams.py
```diff
@@ -329,6 +329,13 @@ def __del__(self):
             closed.exception()
 
 
+def _swallow_unhandled_exception(task):
+    # Do a trick to suppress unhandled exception
+    # if stream.write() was used without await and
+    # stream.drain() was paused and resumed with an exception
+    task.exception()
```
vstinner, line 336: Silently ignore exceptions looks like a bad idea. If the task failed, the exception should be at least logged.

vstinner (changes requested):

Lib/asyncio/streams.py
asvetlov: We have a special case here.
Basically, there are two ways to call `write()` now:
1. Legacy `writer.write(b'data')`
2. Recommended `await writer.write(b'data')`

For the second case, an exception raised by nested `writer.drain()` call is propagated to user.
The first case ignores the exception from nested `writer.drain()` call, allowing existing code to work in a backward compatible way.
A subsequent call to `drain()` raises the exception again, like in the following snippet:
```
writer.write(b'data')  # exception is swallowed
await writer.drain()  # previous exception is raised if any
```

Sorry for such a tricky code.
Ideally, we should make `writer.write()` a coroutine from the very beginning.
Proposed workaround is a strategy to keep backward compatibility and provide a smooth transition path.

asvetlov:

Well, actually I did nothing but provide a comment that describes why ignoring the exception in this particular case is ok.

vstinner (approved):

vstinner:
Honestly, I don't understand how it can be a good idea to silently ignore errors. But this PR fix the warning on my Windows VM when running test_drain_raises() on Windows, so I approved it.

My main concern right now is that https://bugs.python.org/issue36870 is breaking the Python CI and the Python workflow.

Lib/asyncio/streams.py
vstinner: > writer.write(b'data')  # exception is swallowed

Again, this exception should be logged somehow. It's a bad practice to *silently* ignore exceptions and don't provide any way to the developer to see these exceptions.

Can you try to write the exception?

I only merged your PR only to attempt to fix https://bugs.python.org/issue36870

vstinner: I'm trying to do exactly the opposite than your PR :-) Provide a way to get ignored exceptions: https://github.com/python/cpython/pull/13187 :-)

asvetlov:
Thanks, @vstinner 
Regarding random fails: looks like the current asyncio code showed the problem that existed for a long time already.
Please let me warm up my Windows VM, I'll cope with it.

Lib/asyncio/streams.py
asvetlov: Well, it is not for suppressing an exception entirely, the exception is raised to the caller on `await writer.write()` if this form of call is used.
It is for disabling the exception logging. We do similar things in asyncio somewhere.

1st1:
@vstinner Please don't merge asyncio PRs without my review.  This one looks fine though.

vstinner:
> @vstinner Please don't merge asyncio PRs without my review. This one looks fine though.

test_asyncio was causing a lot of troubles, please see https://bugs.python.org/issue36870 for the context.

@asvetlov: Can you please also fix Python 3.7? Python 3.7 is also affected by https://bugs.python.org/issue36870

vstinner:
By the way, it's not the first time that the Python CI and the Python workflow is broken by test_asyncio, especially by random failures on Windows. I'm not sure how to prevent these issues.

The policy for the CI is to revert a change which introduces a regression: https://pythondev.readthedocs.io/ci.html#revert-on-fail

That's what I did: PR #13316. But @asvetlov asked me to look at this PR instead.

PR #13315, opened 2019-05-14
[2.7] bpo-30458: Disallow control chars in http URLs. (GH-12755) (GH-13154)

vstinner:
Disallow control chars in http URLs in urllib2.urlopen.  This
addresses a potential security problem for applications that do not
sanity check their URLs where http request headers could be injected.

Disable https related urllib tests on a build without ssl (GH-13032)
These tests require an SSL enabled build. Skip these tests when
python is built without SSL to fix test failures.

Use httplib.InvalidURL instead of ValueError as the new error case's
exception. (GH-13044)

Backport Co-Authored-By: Miro Hrončok <miro@hroncok.cz>

(cherry picked from commit 7e200e0763f5b71c199aaf98bd5588f291585619)

Notes on backport to Python 2.7:

* test_urllib tests urllib.urlopen() which quotes the URL and so is
  not vulerable to HTTP Header Injection.
* Add tests to test_urllib2 on urllib2.urlopen().
* Reject non-ASCII characters: range 0x80-0xff.

### DIFF
--- a/Lib/test/test_urllib.py
+++ b/Lib/test/test_urllib.py
@@ -9,6 +9,10 @@ import os
 import sys
 import mimetools
 import tempfile
+try:
+    import ssl
+except ImportError:
+    ssl = None
 
 from test import test_support
 from base64 import b64encode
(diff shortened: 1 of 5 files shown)


vstinner:
I backported the fix from Python 3.7 to Python 2.7.

Please review it carefully, I had to make multiple changes to adapt the fix to Python 2:

* non-ASCII characters are explicitly rejected
* urllib doesn't reject control characters: they are quoted properly, so I addapted test_urllib
* urllib2 doesn't quote the URL and so reject control characters, I added tests to test_urllib2
* I replaced http.client with httplib
* I replaced urllib.request with urllib or urllib2

vstinner:
@orsenthil @ned-deily @tirkarthi @hroncok: Would you mind to review my backport to Python 2.7?

tirkarthi (approved):
LGTM. Thanks.

ned-deily:
#vstinner, I'm not sure why you asked me to review this.  Perhaps you meant to ask @benjaminp as 2.7. release manager?

vstinner:
> @vstinner I'm not sure why you asked me to review this. Perhaps you meant to ask @benjaminp as 2.7. release manager?

I just copied the list of all people who reviewed the change in other branches. If you have no opinion, that's fine :-)

I would prefer to have more eyes as possible on this tricky backport ;-)

orsenthil (approved):
lgtm. Thanks, Victor.

Lib/test/test_urllib.py
```diff
@@ -257,6 +261,33 @@ def test_url_fragment(self):
         finally:
             self.unfakehttp()
 
+    @unittest.skipUnless(ssl, "ssl module required")
```
gpshead: is this true for these tests?  (not that it matters, all sane platforms have ssl so these tests will be run regardless)

Lib/test/test_urllib2.py
```diff
@@ -1317,6 +1320,52 @@ def test_unsupported_algorithm(self):
...
+            finally:
+                self.unfakehttp()
+
+    @unittest.skipUnless(ssl, "ssl module required")
+    def test_url_with_newline_header_injection_rejected(self):
+        self.fakehttp(b"HTTP/1.1 200 OK\r\n\r\nHello.")
+        host = "localhost:7777?a=1 HTTP/1.1\r\nX-injected: header\r\nTEST: 123"
+        schemeless_url = "//" + host + ":8080/test/?test=a"
+        try:
+            # We explicitly test urllib.request.urlopen() instead of the top
```
gpshead: urllib2.urlopen (here and in the comment in the test above)

Lib/test/test_urllib2.py
```diff
@@ -1,5 +1,5 @@
 import unittest
-from test import test_support
+from test import support
```
gpshead: fwiw moving from test_support to support isn't directly related to this PR.  i don't mind myself.

gpshead (commented):
overall, this looks good to me,  my only comments are minor.

Lib/test/test_urllib2.py
vstinner: reverted

Lib/test/test_urllib2.py
vstinner: fixed

Lib/test/test_urllib.py
vstinner: oops, no, the decorator is wrong: i removed it.

vstinner:
Ok, thanks everybody for reviews. I merged my PR.

PR #12850, opened 2019-04-16
bpo-27987: pymalloc: align by 16bytes on 64bit platform

methane:
https://bugs.python.org/issue27987

PR #13319, opened 2019-05-14
[2.7] bpo-27987: pymalloc: align by 16bytes on 64bit platform (GH-12850)

vstinner:
(cherry picked from commit f0be4bbb9b3cee876249c23f2ae6f38f43fa7495)


vstinner:
@pitrou @methane: Python 2.7 should be fixed as well. Python must respect the x86-64 ABI, no?

methane (approved):

vstinner:
Thanks @methane, I wasn't comfortable with merging this change without your review :-)

PR #13322, opened 2019-05-14
bpo-36760: Clarify subprocess capture_output docs.

gpshead:
Clarify how to capture stdout and stderr combined into one stream.

https://bugs.python.org/issue36760


tirkarthi (approved):
LGTM. Thanks.

PR #13326, opened 2019-05-14
bpo-25652: Fix __rmod__ of UserString

isidentical:
The ``__rmod__`` method of ``collections.UserString`` class had a bug that made it unusable. 

https://bugs.python.org/issue25652

### DIFF
--- a/Lib/test/test_collections.py
+++ b/Lib/test/test_collections.py
@@ -71,10 +71,17 @@ class TestUserObjects(unittest.TestCase):
         self._copy_test(obj)
 
     def test_str_rmod(self):
-        arg = UserString("python")
-        template = "I love %s"
-        self.assertEqual(arg.__rmod__(template), "I love python")
-        self.assertIs(arg.__rmod__(type('Dummy', (), {})), NotImplemented)
+        class ustr2(UserString):
+            pass
+
+
+        class ustr3(ustr2):
+            def __rmod__(self, other):
+                return super().__rmod__(other)
+
+        fmt2 = ustr2('value is %s')
+        str3 = ustr3('TEST')
+        self.assertEqual(fmt2 % str3, 'value is TEST')
 
 ################################################################################
 ### ChainMap (helper class for configparser and the string module)
--- a/Lib/test/test_collections.py
+++ b/Lib/test/test_collections.py
@@ -70,17 +70,6 @@ class TestUserObjects(unittest.TestCase):
         obj[123] = "abc"
         self._copy_test(obj)
 
-    def test_str_rmod(self):
-        class ustr2(UserString):
-            pass
-
-        class ustr3(ustr2):
-            def __rmod__(self, other):
-                return super().__rmod__(other)
-
-        fmt2 = ustr2('value is %s')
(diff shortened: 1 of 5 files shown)


jcgoble3:
As the person who originally filed the bpo issue, I stand by my statement that it's better to fix the bug rather than just remove the method. As demonstrated by my `userstringerror.py` attachment on the issue, a subclass of a subclass of `UserString` can be bitten by this due to Python allowing a subclass to override a dunder of the parent class, even when the subclass object is the right-hand operand. (For example, the sub-subclass may be for a specialized case where pre-processing or normalization is required before formatting.)

Therefore `__rmod__` is useful, and removing it would force subclass authors to write the implementation themselves rather than extending the UserString method..

isidentical:
@rhettinger whats your thought?

gvanrossum:
Yeah, let's fix the bug and add a test.

Lib/test/test_collections.py
```diff
@@ -70,6 +70,11 @@ def test_dict_copy(self):
         obj[123] = "abc"
         self._copy_test(obj)
 
+    def test_str_rmod(self):
+        arg = UserString("python")
+        template = "I love %s"
+        self.assertEqual(arg.__rmod__(template), "I love python")
+        self.assertIs(arg.__rmod__(type('Dummy', (), {})), NotImplemented)
```
gvanrossum: It would be more convincing if the test reproduced the end-to-end example from the bpo issue (https://bugs.python.org/file46854/userstringerror.py).

isidentical: Okey, done.

Lib/test/test_collections.py
```diff
@@ -70,6 +70,18 @@ def test_dict_copy(self):
         obj[123] = "abc"
         self._copy_test(obj)
 
+    def test_str_rmod(self):
+        class ustr2(UserString):
+            pass
+
```
gvanrossum: Drop the extra blank line please.

gvanrossum (commented):
OK, there's one whitespace nit in the new test left.

Lib/test/test_collections.py
isidentical: Ah, sorry for that. Fixed

jcgoble3:
Do the new tests belong in `test_collections.py` or `test_userstring.py`? I'm also confused as to why the three `test_user*.py` files exist when `test_collections.py` also has tests for those classes.

isidentical:
I wasn't aware of these tests. Should i move the test?

jcgoble3:
I'm still new to the contributing process, so I will let the core devs figure out where they belong. I honestly don't know, hence why I asked it as a neutral question.

gvanrossum:
Looks like the tests in `test_collections.py` only make very little mention of `UserString` etc., and the important tests for those classes are in their own test file. So I recommend moving the new tests to `test_userstring.py`.

isidentical:
I moved the test.

gvanrossum (approved):
OK, if the test pass it will be merged. Thanks!

isidentical:
Thank you for reviewing it.

PR #13332, opened 2019-05-15
bpo-1875: Raise SyntaxError in invalid blocks that will be optimized away

pablogsal:
https://bugs.python.org/issue1875

### DIFF
--- a/Lib/test/test_syntax.py
+++ b/Lib/test/test_syntax.py
@@ -696,6 +696,20 @@ class SyntaxTestCase(unittest.TestCase):
     def test_break_outside_loop(self):
         self._check_error("break", "outside loop")
 
+    def test_yield_outside_function(self):
+        self._check_error("if 0: yield",               "outside function")
+        self._check_error("class C:\n    if 0: yield", "outside function")
+
+    def test_return_outside_function(self):
+        self._check_error("if 0: return",               "outside function")
+        self._check_error("class C:\n    if 0: return", "outside function")
+
+    def test_break_outside_loop(self):
+        self._check_error("if 0: break",                "outside loop")
+
+    def test_continue_outside_loop(self):
+        self._check_error("if 0: continue",     "not properly in loop")
+
     def test_unexpected_indent(self):
         self._check_error("foo()\n bar()\n", "unexpected indent",
                           subclass=IndentationError)
--- a/Misc/NEWS.d/next/Core and Builtins/2019-05-15-01-29-29.bpo-1875.9oxXFX.rst
+++ b/Misc/NEWS.d/next/Core and Builtins/2019-05-15-01-29-29.bpo-1875.9oxXFX.rst
@@ -1,0 +1,3 @@
+A :exc:`SyntaxError` is now raised if a code blocks that will be optimized
+away (e.g. if conditions that are always false) contains syntax errors.
+Patch by Pablo Galindo.
--- a/Python/compile.c
+++ b/Python/compile.c
@@ -2546,13 +2546,12 @@ compiler_if(struct compiler *c, stmt_ty s)
         return 0;
 
     constant = expr_constant(s->v.If.test);
-    /* constant = 0: "if 0"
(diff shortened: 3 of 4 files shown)


vstinner:
I wrote similar code in my old fatoptimizer project (AST optimizer):
https://github.com/vstinner/fatoptimizer/blob/master/test_fatoptimizer.py

Extracts:

```
        # comparison between bytes and str can raise BytesWarning depending
        # on runtime option
        self.check_dont_optimize_func('"x" == b"x"')
        self.check_dont_optimize_func('b"x" == "x"')
        self.check_dont_optimize_func('"x" != b"x"')
        self.check_dont_optimize_func('b"x" != "x"')

        # bytes < str raises TypeError
        self.check_dont_optimize_func('b"bytes" < "str"')

        self.check_dont_optimize("""
            def func():
                if 0:
                    yield
        """)

        self.check_dont_optimize("""
            def func(obj):
                return
                if 0:
                    yield from obj
        """)

        self.check_dont_optimize("""
            try:
                pass
            except Exception:
                yield 3
        """)

        # must raise SyntaxError
        self.check_dont_optimize("""
            for x in (1, 2):
                try:
                    pass
                except Exception:
                    func2()
                finally:
                    # continue is final is illegal
                    continue
        """)

        self.check_dont_optimize("""
            for x in (1, 2):
                try:
                    pass
                except Exception:
                    func2()
                else:
                    func3()
                finally:
                    # continue is final is illegal
                    continue
        """)

        self.check_dont_optimize("""
            for x in (1, 2):
                try:
                    pass
                except Exception:
                    try:
                        func2()
                    finally:
                        # continue is final is illegal
                        continue
        """)

        self.check_dont_optimize("""
            if test:
                pass
        """)
```

It would be nice to ensure that our optimizer doesn't optimize these examples.

pablogsal:
@vstinner Ok, I like to add these tests but let's open a different issue for those.

vstinner:
When I wrote an AST optimizer, I implemented correctness checks to ensure that I don't remove code that would raise a SyntaxError:
https://github.com/vstinner/fatoptimizer/blob/master/fatoptimizer/dead_code.py#L7

```
# AST types of nodes that cannot be removed
_CANNOT_REMOVE_TYPES = (ast.Global, ast.Nonlocal, ast.Yield, ast.YieldFrom,
                        # don't remove 'except' block if it contains continue
                        # or break: see can_move_final() for the rationale
                        ast.Continue)

_CANNOT_MOVE_FINAL = (ast.Continue,)
```

My implementation checks if a tree contains yield (for example) before trying to remove it. If it contains yield, global, etc.: the node is not removed.

Maybe we should implement that in the new builtin AST optimizer, rather than in peephole.c which works at the bytecode level?

vstinner:
Hum, my previous comment is unclear. I don't understand how your PR still removes "if 0: print(name)" but keeps "if 0: yield". Or it doesn't?

pablogsal:
> Hum, my previous comment is unclear. I don't understand how your PR still removes "if 0: print(name)" but keeps "if 0: yield". Or it doesn't?

It compiles the block (raising errors if any are found) and moves the removal optimization to the peephole optimizer. The only downside (if you consider any) is that the variables inside the dead code will still be considered when building the code stack size and some jumps may not be optimized. But this is a no-problem for three reasons:

1) We would be prioritizing speed over correctness. 
2) (More important) This only happens when you have dead code that can be detected statically. Slightly unoptimized function compilation and a slightly bigger stack when you have dead code are by far not the biggest problem. The biggest problem is the dead code so I think we are ok if the slightly worst code is generated if dead code blocks are present. 
3) This situation (a dead code block with a falsy constant in the condition) happens extremely rare, so it should not even be a concern. This PR is about correctness more than anything.

vstinner (approved):
LGTM.

First I misunderstood the change. Once I understood it: yeah, it's correct to move an optimization from the compiler to the (peephole) optimizer. First, I didn't understood why a compiler directly implements an optimization.

PR #13336, opened 2019-05-15
bpo-27987: align PyGC_Head to alignof(long double) explicitly

methane:
PyGC_Head consists two uintptr_t variables so it is aligned to 16 byte
on amd64 platform already.

This commit makes this alignment more explicit, although make code
bit ugly.

### DIFF
--- a/Include/cpython/objimpl.h
+++ b/Include/cpython/objimpl.h
@@ -43,20 +43,23 @@ PyAPI_FUNC(Py_ssize_t) _PyGC_CollectIfEnabled(void);
      && (Py_TYPE(o)->tp_is_gc == NULL || Py_TYPE(o)->tp_is_gc(o)))
 
 /* GC information is stored BEFORE the object structure. */
-typedef struct {
-    // Pointer to next object in the list.
-    // 0 means the object is not tracked
-    uintptr_t _gc_next;
+typedef union {
+    struct {
+        // Pointer to next object in the list.
+        // 0 means the object is not tracked
+        uintptr_t next;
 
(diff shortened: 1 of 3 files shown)


Include/cpython/objimpl.h
```diff
@@ -43,20 +43,23 @@ PyAPI_FUNC(Py_ssize_t) _PyGC_CollectIfEnabled(void);
...
+    struct {
+        // Pointer to next object in the list.
+        // 0 means the object is not tracked
+        uintptr_t next;
+
+        // Pointer to previous object in the list.
+        // Lowest two bits are used for flags documented later.
+        uintptr_t prev;
+    } _gc;
+    long double _dummy;  /* for worst alignment */
```
vstinner, line 56: Can you please elaborate the comment? Add "bpo-27987" and explain that x64-64 ABI requires to align on 16 bytes.

Do you have to align to 16 bytes on x86 (32-bit) as well?

vstinner:
> This commit makes this alignment more explicit, although make code bit ugly.

Said differently, it doesn't change PyGC_Head size on 64-bit system, right? Does it change the size on 32-bit system?

Correctness matters more than code beauty here ;-)

Include/cpython/objimpl.h
methane: Oh! I forgot about 32bit platform.
I thought this PR does nothing, but it makes PyGC_Head to 16 bytes from 8 bytes on x86 platform.

Maybe, we can use `__attribute__ ((aligned))` if compiler is GCC or clang.

PR #13338, opened 2019-05-15
bpo-36922: implement Py_TPFLAGS_METHOD_DESCRIPTOR

jdemeyer:
CC @encukou @markshannon 

https://bugs.python.org/issue36922

### DIFF
--- a/Doc/c-api/typeobj.rst
+++ b/Doc/c-api/typeobj.rst
@@ -1045,6 +1045,32 @@ and :c:type:`PyType_Type` effectively act as defaults.)
 
       ???
 
+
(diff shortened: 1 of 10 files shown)


jdemeyer:
Petr: do you think this should be `_Py_TPFLAGS_METHOD_DESCRIPTOR` (with underscore) or is this useful enough to make public right away?

This PR is otherwise independent from the rest of PEP 590, so it's possible to make it public.

encukou:
I don't see this one changing, so I'm OK with making this public.

encukou (commented):
This looks good, thank you! It should go in after PEP is accepted.

jdemeyer:
Rebased to latest master (no other changes).

PR #13339, opened 2019-05-15
bpo-36567: use debian's manpage service for references

isidentical:
use `MANPAGES_URL` env variable for manpage references. If that variable not found, use debians manpage service.

https://bugs.python.org/issue36567

### DIFF
--- a/Doc/conf.py
+++ b/Doc/conf.py
@@ -24,8 +24,7 @@ except ImportError:
     _tkinter = None
 '''
 
-# Manpage config
-manpages_url = os.getenv('MANPAGES_URL', 'https://manpages.debian.org/{path}')
+manpages_url = 'https://manpages.debian.org/{path}'
 
 # General substitutions.
 project = 'Python'
--- a/Doc/conf.py
+++ b/Doc/conf.py
@@ -23,6 +23,10 @@ try:
 except ImportError:
     _tkinter = None
 '''
+
+# Manpage config
+manpages_url = os.getenv('MANPAGES_URL', 'https://manpages.debian.org/{path}')
+
 # General substitutions.
 project = 'Python'
 copyright = '2001-%s, Python Software Foundation' % time.strftime('%Y')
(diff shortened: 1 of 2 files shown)


Misc/NEWS.d/next/Documentation/2019-05-15-14-31-40.bpo-36567.MNbyoI.rst
```diff
@@ -0,0 +1 @@
+Use debian's manpage service website for references.
```
berkerpeksag: NEWS entry isn't needed for trivial PRs. I've set the "skip news" label.

Doc/conf.py
```diff
@@ -23,6 +23,10 @@
 except ImportError:
     _tkinter = None
 '''
+
+# Manpage config
+manpages_url = os.getenv('MANPAGES_URL', 'https://manpages.debian.org/{path}')
```
berkerpeksag: I don't think this is something that needs to be customized via an environment variable.

Setting its value directly should be enough IMO:

```py
manpages_url = 'https://manpages.debian.org/{path}'
```

Doc/conf.py
```diff
@@ -23,6 +23,10 @@
 except ImportError:
     _tkinter = None
 '''
+
+# Manpage config
```
berkerpeksag: Nit: This comment can be dropped. The name of the setting is already self-descriptive.

berkerpeksag (changes requested):

Misc/NEWS.d/next/Documentation/2019-05-15-14-31-40.bpo-36567.MNbyoI.rst
isidentical: Done

Doc/conf.py
isidentical: Done

Doc/conf.py
isidentical: Done

isidentical:

asottile (approved):

berkerpeksag (approved):
LGTM! I just tested this branch locally and it worked as expected.

berkerpeksag:
Ok, let's not bother backporting to 3.7 then.

PR #13340, opened 2019-05-15
bpo-37151: simplify classmethoddescr_call

jdemeyer:
`skip news` since this is an implementation detail.

CC @encukou @markshannon 

### DIFF
--- a/Lib/test/test_descr.py
+++ b/Lib/test/test_descr.py
@@ -1613,8 +1613,8 @@ order (MRO) for bases """
             spam_cm(spam.spamlist())
         self.assertEqual(
             str(cm.exception),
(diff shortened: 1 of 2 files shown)


encukou:
I don't see a problem merging this with PEP 590. By itself, de-optimizing doesn't make sense to me.

auvipy (approved):

jdemeyer:
Can this be merged now?

encukou:
Yes.

FWIW, I'm sorry for my current unavailability. I'm busy with the update to 3.8beta1 in Fedora (re-building and re-testing 2000+ Pytohn packages, we do find failures caused – some caused by removing `tp_print`).

PR #13344, opened 2019-05-15
Fix typos in documentation

tirkarthi:


tirkarthi:
I am not a native English speaker I think it should be "Apple's guidelines" instead of "Apple guidelines" in idlelib.

matrixise:
Do we need a back port to 3.7?

tirkarthi:
Except for "guidlines" in idlelib the other two are 3.8 only.

csabella:
@terryjreedy usually likes to make the IDLE doc fixes into their own separate PR, so you may want to wait for him to comment.

terryjreedy:
My typo, thanks for catching.  Nouns can be used as adjectives, so 'Apple guidelines' is OK, with a slightly different sense from 'Apple's guidelines'.  But I will consider opinions on which seems better, especially for beginners.

Changes to idle.rst should not only be backported, but need to be propagated to idlelib/help.html, which change also should be backported.  It is easier to do it in a separate PR, which I will do.

tirkarthi:
> Changes to idle.rst should not only be backported, but need to be propagated to idlelib/help.html, which change also should be backported. It is easier to do it in a separate PR, which I will do.

Do you want me to revert idlelib changes from this PR? The other two typos are 3.8 only. Hence backport might require manual work.

terryjreedy:
Follow-up is #13354.

tirkarthi:
Thanks @terryjreedy

PR #13345, opened 2019-05-15
bpo-36928: link threading.settrace to sys.settrace

stefanhoelzl:
# link threading.settrace to sys.settrace

linking the documentation of threading.settrace to sys.settrace.

sys.settrace mentions multi-threading but there is not link to threading.settrace which also can be used to set a trace function for threads.

https://bugs.python.org/issue36928


matrixise:
@stefanhoelzl you don't need to push force if you add new commits, there is no problem with that.

matrixise (approved):

PR #13346, opened 2019-05-15
bpo-36921: Deprecate @coroutine for sake of async def

asvetlov:
The second attempt. Now deprecate `@coroutine` only, keep `yield from fut` as is.

https://bugs.python.org/issue36921


Lib/test/test_typing.py
```diff
@@ -1708,9 +1708,8 @@ def __init__(self, value: typing.Iterable[T_a]):
     def __aiter__(self) -> typing.AsyncIterator[T_a]:
         return self
 
-    @asyncio.coroutine
-    def __anext__(self) -> T_a:
-        data = yield from self.value
+    async def __anext__(self) -> T_a:
```
asvetlov, line 1711: The same functionality but without `@coroutine`

ilevkivskyi (approved):
The `typing` part LGTM.

Lib/asyncio/coroutines.py
```diff
@@ -107,6 +108,9 @@ def coroutine(func):
     If the coroutine is not yielded from before it is destroyed,
     an error message is logged.
     """
+    warnings.warn('"@coroutine" decorator is deprecated, use "async def" instead',
```
Carreau:
```suggestion
    warnings.warn('"@coroutine" decorator is deprecated since Python 3.8 , use "async def" instead',
```

asvetlov: Thanks!
The correction is very valuable.

Carreau: Thanks for cleaning up things, and working on that. It's not that much of a correction than a supplement and a small pet-peeve of mine. I find it sometime difficult to know when things were changed and how pressing the updating needs be, so a small nudge could help downstream developers to update their code.

It also make it a tiny bit more probable that this will be removed in a couple of years.

Carreau:
want to update `asyncio-task.rst` to use the `.. deprecated-removed:: 3.8 3.10` directive ?

Carreau:
Also side-question (and maybe this is better suited for bpo , should `asyncio.iscoroutine` and `asyncio.iscoroutinefunction` be deprecated at the same time ? 

Because presumably once `@coroutine` is removed, the two other have quasy no chance of encountering a coroutine that would not be properly detected by `inspect.iscoroutine`.

I'm thinking they are still useful from inter-library compatibility and across python version; so maybe deprecate as well but remove even later  ?

Lib/asyncio/coroutines.py
```diff
@@ -107,6 +108,9 @@ def coroutine(func):
     If the coroutine is not yielded from before it is destroyed,
     an error message is logged.
     """
+    warnings.warn('"@coroutine" decorator is deprecated since Python 3.8 , use "async def" instead',
```
1st1:
```suggestion
    warnings.warn('"@coroutine" decorator is deprecated since Python 3.8, use "async def" instead',
```

Lib/asyncio/locks.py
```diff
@@ -3,12 +3,13 @@
...
 
 import collections
+import types
 import warnings
 
 from . import events
 from . import futures
 from . import exceptions
-from .coroutines import coroutine
+from .coroutines import _is_coroutine
```
1st1:
```suggestion
from .import coroutines
```

Lib/asyncio/locks.py
```diff
@@ -78,6 +79,8 @@ def __iter__(self):
         yield from self.acquire()
         return _ContextManager(self)
 
+    __iter__._is_coroutine = _is_coroutine
```
1st1: Add a comment why we need this: "For legacy asyncio.iscoroutine()"

1st1 (approved):
Overall LGTM.  Please fix the few nits I commented on.

1st1:
> Also side-question (and maybe this is better suited for bpo , should asyncio.iscoroutine and asyncio.iscoroutinefunction be deprecated at the same time ?

We can deprecate them when we remove `@asyncio.coroutine` (in 3.9?)

asvetlov:
@caavery the funny thing that `@coroutine` is already deprecated in docs. Now we have the support of this statement explicitly in code.
Removal is scheduled for 3.10, that's fine.
I agree with @1st1 that `asyncio.iscoroutine` and `asyncio.iscoroutinefunction` should be deprecated later.
As the maintainer of several asyncio based libraries, I want a smooth transition path.
My libraries don't use `@coroutine` already but sometimes accept async callbacks and calls asyncio checkers for them. I want to support user-written old-style coros in my libraries as long as possible (but discourage them).
Basically the same as asyncio itself does.
Deprecation checkers after removal the root functionality absolutely makes sense to me. We can consider deprecation in 3.9 if you want but I pretty sure we shouldn't do it in 3.8.

Doc/library/asyncio-task.rst
```diff
@@ -916,12 +916,13 @@ enforced.
         async def main():
             await old_style_coroutine()
 
-    This decorator is **deprecated** and is scheduled for removal in
-    Python 3.10.
-
     This decorator should not be used for :keyword:`async def`
     coroutines.
 
+    .. deprecated-removed:: 3.8 3.10
```
asvetlov, line 922: Replace free-form deprecation text with more standardized one.

asvetlov:
The last call for a review.
I have a plan to merge the PR is a day.

1st1:
> The last call for a review.

I think it's ready.

PR #13347, opened 2019-05-15
bpo-36929: Modify io/re tests to allow for missing mod name

tekknolagi:
For a vanishingly small number of internal types, CPython sets the
tp_name slot to mod_name.type_name, either in the PyTypeObject or the
PyType_Spec. There are a few minor places where this surfaces:

* Custom repr functions for those types (some of which ignore the
  tp_name in favor of using a string literal, such as _io.TextIOWrapper)
* Pickling error messages

The test suite only tests the former. This commit modifies the test
suite to allow Python implementations to omit the module prefix.

https://bugs.python.org/issue36929

### DIFF
--- a/Lib/test/test_io.py
+++ b/Lib/test/test_io.py
@@ -1085,12 +1085,12 @@ class CommonBufferedTests:
     def test_repr(self):
         raw = self.MockRawIO()
         b = self.tp(raw)
-        clsname = "%s.%s" % (self.tp.__module__, self.tp.__qualname__)
-        self.assertEqual(repr(b), "<%s>" % clsname)
+        clsname = "(%s\.)?%s" % (self.tp.__module__, self.tp.__qualname__)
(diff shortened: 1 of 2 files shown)


tekknolagi:
Hm, looks like I broke something. Poking around.

Mariatta:
Thanks for the PR. I noticed the PR is for 3.6 branch, which is no longer accepting bug fixes (only security fixes).

If this is also a problem in the later versions, then your PR should target the `master` branch, and we will backport once the `master` PR has been merged.

tekknolagi:
Ah, I see. Yes, this appears to also be a problem in later versions. So I should target master, and it will be backported to, 3.6 and 3.7? Or just 3.7?

Mariatta:
It will be backported to 3.7 only at this point. 3.6 is only open for security fixes, and I don't think this one counts as security fix. Thanks.

PR #13348, opened 2019-05-15
Fix #19404: Change context managers to have appropriate docstrings

plokmijnuhby:
Changed functionality so that contextmanager and asyccontextmanager return classes rather than functions. The classes have appropriate docstrings and names. (This made one of the internal classes redundant, so I removed it.)

### DIFF
--- a/Lib/contextlib.py
+++ b/Lib/contextlib.py
@@ -76,26 +76,22 @@ class ContextDecorator(object):
         return inner
 
 
-class _GeneratorContextManagerBase:
-    """Shared functionality for @contextmanager and @asynccontextmanager."""
+def _class_wrapper(cls, func):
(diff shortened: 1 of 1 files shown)


the-knights-who-say-ni:

1st1 (commented):
I'm really not sure about this one.  While I understand the idea and "why", I don't think that the increased complexity of the new code is worth it.

ncoghlan:
Hi, thanks for this attempt, but as discussed on the tracker issue, I don't think changing the way contextlib works to cope with a limitation in pydoc is the right answer here - instead, we need to make pydoc smarter so that contextlib can tell it to return instance-specific documentation, and ignore the documentation for the type.

PR #13349, opened 2019-05-15
bpo-36932: use proper deprecation-removed directive

Carreau:
.. And update some deprecation warnings with version numbers.

https://bugs.python.org/issue36932

### DIFF
--- a/Doc/library/asyncio-task.rst
+++ b/Doc/library/asyncio-task.rst
@@ -280,9 +280,7 @@ Sleeping
    to run.
 
    .. deprecated-removed:: 3.8 3.10
-
-      The *loop* argument is deprecated and scheduled for removal
-      in Python 3.10.
+      The *loop* parameter.
 
    .. _asyncio_example_sleep:
 
@@ -440,9 +438,7 @@ Timeouts
    If the wait is cancelled, the future *aw* is also cancelled.
 
    .. deprecated-removed:: 3.8 3.10
-
-      The *loop* argument is deprecated and scheduled for removal
-      in Python 3.10.
+      The *loop* parameter.
 
    .. _asyncio_example_waitfor:
 
@@ -496,9 +492,7 @@ Waiting Primitives
         done, pending = await asyncio.wait(aws)
 
    .. deprecated-removed:: 3.8 3.10
-
-      The *loop* argument is deprecated and scheduled for removal
-      in Python 3.10.
+      The *loop* parameter.
 
    *timeout* (a float or int), if specified, can be used to control
    the maximum number of seconds to wait before returning.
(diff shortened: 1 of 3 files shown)


pganssle (approved):
Nice, always good to see some more structure and detail added to these things.

Carreau:
I'd love to be able to build an index of deprecated/removal of this is used
consistently. Though  not want to write a Sphinx extension...

isidentical (approved):

Doc/library/sys.rst
```diff
@@ -1353,7 +1353,7 @@ always available.
       This function has been added on a provisional basis (see :pep:`411`
       for details.)  Use it only for debugging purposes.
 
-   .. deprecated:: 3.7
+   .. deprecated-removed:: 3.7 3.8
```
asvetlov, line 1356: Side question:
Do we really want to remove it in 3.8?
If yes -- the time window is very close.
@1st1 we need your judgment for it.

asvetlov (approved):

Doc/library/asyncio-task.rst
```diff
@@ -279,8 +279,10 @@ Sleeping
    ``sleep()`` always suspends the current task, allowing other tasks
    to run.
 
-   The *loop* argument is deprecated and scheduled for removal
-   in Python 3.10.
+   .. deprecated-removed:: 3.8 3.10
+
+      The *loop* argument is deprecated and scheduled for removal
```
asvetlov: The text should be updated.
Now it looks like:

*Deprecated since version 3.8, will be removed in version 3.10:*  The *loop* argument is deprecated and scheduled for removal in Python 3.10

Looks like battology.

asvetlov (changes requested):
Please fix deprecation texts to remove generated wordiness

Doc/library/sys.rst
1st1: I think we should drop it. asyncio no longer needs it, AFAIK Trio considered using it but decided it's too slow.

1st1: I can do it myself tomorrow.

asvetlov: Nice!
Thank you very much!

Carreau: @1st1  I've opened https://bugs.python.org/issue36933 yesterday to track the removal.

Doc/library/asyncio-task.rst
Carreau: I spotted that and didn't wanted to change too much at once, but sure I can take care of it.

asvetlov (approved):
Thanks!

asvetlov:
Well, if doc change cannot be backported to 3.7 automatically I prefer to keep it in 3.8 only.

Thanks, @Carreau !

Carreau:
> Well, if doc change cannot be backported to 3.7 automatically I prefer to keep it in 3.8 only.

Ok, let me know I can work on backporting it otherwise.

asvetlov:
Backporting would be good but I don't insist at all.
We live with existing markup for half a year, nobody complains :)

Carreau:
> Backporting would be good but I don't insist at all.

Well, I prefer to focus on features then, (like top-level-await #13148) before beta1, once we are in the beta period I can work on applying that to 3.7.

PR #11779, opened 2019-02-07
bpo-35926: Add support for OpenSSL 1.1.1b on Windows

paulmon:
After a little more testing I discovered that these changes don't quite pass all of the tests yet.  
I will be looking into these failures tomorrow

amd64 retail - 1 failure
amd64 debug - 0 failures
win32 retail - 1 failures
win32 debug - access violation halts test

test_parse_cert_CVE_2019_5010 only fails win32 debug (access violation)
works for amd64 debug/release and win32 release

test_load_default_certs_env_windows fails on win32 and amd64 retail.  skipped on debug

https://bugs.python.org/issue35926

PR #13350, opened 2019-05-15
[3.7] bpo-35926: Add support for OpenSSL 1.1.1b on Windows (GH-11779)

zooba:
https://bugs.python.org/issue35926


zooba:
This includes some of the ARM[32/64] properties from 3.8, but I'm not worried about them. They're all tied to our OpenSSL sources anyway, which supports it, and it will only make backporting easier in the future.

PR #13352, opened 2019-05-16
bpo-18911: using xmlcharrefreplace when open a file

Windsooon:
See https://bugs.python.org/issue18911

https://bugs.python.org/issue18911


auvipy (approved):

scoder (changes requested):
We should not recommend using the replacement method. See my comment in the ticket.

Windsooon:
@scoder I updated the PR base on your review and suggestions.

scoder (approved):

scoder:
Thanks @Windsooon!

PR #13353, opened 2019-05-16
bpo-31171: add `-lpthread' to build multiprocessing for linux platform

jiahongxujia:
While cross compiling on Linux, AC_RUN_IFELSE to check `-pthread' is always
false, add `-lpthread' to linker, it fixed multiprocessing.BoundedSemaphore
of 32-bit python did not work.

Whether corss compiling or not, whether supports `-pthread' or not, the fix
does not have side effect on Linux platform.

Signed-off-by: Hongxu Jia <hongxu.jia@windriver.com>

https://bugs.python.org/issue31171

### DIFF
--- a/setup.py
+++ b/setup.py
@@ -1597,6 +1597,8 @@ class PyBuildExt(build_ext):
                                ['cjkcodecs/_codecs_%s.c' % loc]))
 
     def detect_multiprocessing(self):
+        multiprocessing_lib = []
+
         # Richard Oudkerk's multiprocessing module
         if MS_WINDOWS:
             multiprocessing_srcs = ['_multiprocessing/multiprocessing.c',
@@ -1619,7 +1621,11 @@ class PyBuildExt(build_ext):
                                    libraries=libs,
                                    include_dirs=["Modules/_multiprocessing"]))
 
+            if HOST_PLATFORM.startswith('linux'):
+                multiprocessing_lib = ['pthread']
+
         self.add(Extension('_multiprocessing', multiprocessing_srcs,
+                           libraries=multiprocessing_lib,
                            include_dirs=["Modules/_multiprocessing"]))
 
     def detect_uuid(self):


auvipy (approved):

setup.py
```diff
@@ -1619,7 +1621,11 @@ def detect_multiprocessing(self):
                                    libraries=libs,
                                    include_dirs=["Modules/_multiprocessing"]))
 
+            if HOST_PLATFORM.startswith('linux'):
+                multiprocessing_lib = ['pthread']
```
pitrou, line 1625: @xdegaye Does it look right to you?

pitrou: Also cc @rossburton.

xdegaye: On Android, the Python platform name is ``linux`` and pthread is not a standalone library but it is included in the libc (named ``bionic``), same as on MacOS X.  Therefore ``-lpthread`` is not used at the link stage.

IMO it is confusing to have both configure and setup.py handle the setting of ``-lpthread`` especially since the configure implementation of pthread variants is already quite involved. It would be better to just fix ``configure.ac`` if it is needed.

It is not clear from the comments on [issue 31171](https://bugs.python.org/issue31171) if the compiler ``-pthread`` option is available. If that option is available then running configure with ``ac_cv_pthread=yes`` sets ``CC="$CC -pthread"`` and that should fix the problem raised in the issue. Quoted from gcc documentation:
```
'-pthread'
     Define additional macros required for using the POSIX threads
     library.  You should use this option consistently for both
     compilation and linking.  This option is supported on GNU/Linux
     targets, most other Unix derivatives, and also on x86 Cygwin and
     MinGW targets.
```

xdegaye: And if the cross-compiler does not support the ``-pthread`` option then run make with LDFLAGS:
```
    LDFLAGS=-lpthread make
```

kumaraditya303:
`setup.py` does not exists anymore, if this is still an issue please create a new PR for autoconf.

PR #13116, opened 2019-05-06
bpo-36766: Typos in docs and code comments

GiantRobato:
Resolves [bpo-36766](https://bugs.python.org/issue36766)

https://bugs.python.org/issue36766

PR #13354, opened 2019-05-16
Fix typos in documentation.  Patch by tirkarthi.

terryjreedy:

### DIFF
--- a/Doc/library/idle.rst
+++ b/Doc/library/idle.rst
@@ -50,7 +50,7 @@ default title and context menu.
 
 On macOS, there is one application menu.  It dynamically changes according
 to the window currently selected.  It has an IDLE menu, and some entries
-described below are moved around to conform to Apple guidlines.
+described below are moved around to conform to Apple guidelines.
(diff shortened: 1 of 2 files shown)


tirkarthi (approved):
LGTM. Thanks.

terryjreedy:
Somehow my working file did not have the 'millesecond' => 'millisecond' correction someone did about a month ago, even though everything was supposed to be up to date.

tirkarthi:
I think this was missed in https://github.com/python/cpython/pull/13116 though my report had it as an item for idle.rst : https://bugs.python.org/issue36766#msg341194

terryjreedy:
You message specified idle.html rather than idle.rst, and there is what got edited.  I should have noticed.  help.html should not be directly edited since it is a generated file -- a trailing whitespace trimmed copy of the file generated from idle.rst by Sphinx.  So when I generated it the first time today, I noticed the reversion and added the change to idle.rst.  This clears up why.

PR #13355, opened 2019-05-16
bpo-36935: Remove usage of the deprecated PyErr_SetFromWindowsErrWithUnicodeFilename()

ZackerySpytz:
In e895de3e7f3cc2f7213b87621cfe9812ea4343f0, the
deprecated function PyErr_SetFromWindowsErrWithUnicodeFilename() was
added in two functions in Modules/_winapi.c. This function was
deprecated in 3.3.

https://bugs.python.org/issue36935

### DIFF
--- a/Modules/_winapi.c
+++ b/Modules/_winapi.c
@@ -502,7 +502,7 @@ _winapi_CreateFileMapping_impl(PyObject *module, HANDLE file_handle,
     Py_END_ALLOW_THREADS
 
     if (handle == NULL) {
-        PyErr_SetFromWindowsErrWithUnicodeFilename(0, name);
+        PyErr_SetFromWindowsErr(0);
         handle = INVALID_HANDLE_VALUE;
     }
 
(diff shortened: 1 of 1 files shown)


Modules/_winapi.c
```diff
@@ -502,7 +502,7 @@ _winapi_CreateFileMapping_impl(PyObject *module, HANDLE file_handle,
     Py_END_ALLOW_THREADS
 
     if (handle == NULL) {
-        PyErr_SetFromWindowsErrWithUnicodeFilename(0, name);
+        PyErr_SetFromWindowsErr(0);
```
vstinner: This change removes the name information which is useful, no?

Can PyErr_SetExcFromWindowsErrWithFilenameObjects() be used with PyUnicode_FromWideChar() here? Basically, copy-paste PyErr_SetFromWindowsErrWithUnicodeFilename() code:

```
    PyObject *name = filename ? PyUnicode_FromWideChar(filename, -1) : NULL;
    PyObject *result = PyErr_SetExcFromWindowsErrWithFilenameObjects(
                                                  PyExc_OSError,
                                                  ierr, name, NULL);
    Py_XDECREF(name);
    return result;
```

These function names are super long. I dislike these APIs :-(

ZackerySpytz: > Can PyErr_SetExcFromWindowsErrWithFilenameObjects() be used with PyUnicode_FromWideChar() here?

Yes. Eryk Sun had suggested the same thing on the bug tracker.

ZackerySpytz:
@vstinner, I have updated the PR. Please take a look.

PR #13357, opened 2019-05-16
bpo-36936: CALL_FUNCTION_KW keyword names must be non-empty

jdemeyer:
`skip news` because this is just documenting existing behavior

https://bugs.python.org/issue36936


encukou:
This is a behavior change, not just documenting existing behavior. It needs a news entry.
Bytecode semantics are not compatible between CPython versions, but they're still documented and usable. There are libraries out there that generate bytecode, and we need to let their authors know about this change.

jdemeyer:
The motivation for this is https://github.com/python/peps/pull/1049 and if that's rejected, it's probably better to close this too.

PR #13362, opened 2019-05-16
bpo-36763: Add _Py_InitializeMain()

vstinner:
* Add again _PyCoreConfig._init_main
* _PyCoreConfig: rename _frozen to pathconfig_warnings, its value is
  not the opposite of Py_FrozenFlag.
* Add an unit test for _init_main=0 and _Py_InitializeMain()

### DIFF
--- a/Include/cpython/coreconfig.h
+++ b/Include/cpython/coreconfig.h
@@ -398,10 +398,14 @@ typedef struct {
        See PEP 552 "Deterministic pycs" for more details. */
     wchar_t *check_hash_pycs_mode;
 
-    /* If greater than 0, suppress _PyPathConfig_Calculate() warnings.
+    /* If greater than 0, suppress _PyPathConfig_Calculate() warnings on Unix.
+       The parameter has no effect on Windows.
 
-       If set to -1 (default), inherit Py_FrozenFlag value. */
-    int _frozen;
+       If set to -1 (default), inherit !Py_FrozenFlag value. */
+    int pathconfig_warnings;
+
+    /* If equal to 0, stop Python initialization before the "main" phase */
+    int _init_main;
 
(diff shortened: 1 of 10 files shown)


vstinner:
@zooba @Yhg1s @ncoghlan: Ok, here you have, a simple change to allow to execute Python code between _Py_InitializeFromConfig() and _Py_InitializeMain(). Using a small change in run_eval_code_obj() (set __builtins__ in globals), it becomes even possible to use import before _Py_InitializeMain() ;-)

See included unit test:

```
static int test_init_main(void)
{
    _PyCoreConfig config = _PyCoreConfig_INIT;
    configure_init_main(&config);
    config._init_main = 0;

    _PyInitError err = _Py_InitializeFromConfig(&config);
    if (_Py_INIT_FAILED(err)) {
        _Py_ExitInitError(err);
    }

    /* sys.stdout don't exist yet: it is created by _Py_InitializeMain() */
    int res = PyRun_SimpleString(
        "import sys; "
        "print('Run Python code before _Py_InitializeMain', "
               "file=sys.stderr)");
    if (res < 0) {
        exit(1);
    }

    err = _Py_InitializeMain();
    if (_Py_INIT_FAILED(err)) {
        _Py_ExitInitError(err);
    }

    return _Py_RunMain();
}
```

Maybe _Py_InitializeMain() can be added to PEP 587 as an experimental feature until we decide how to implement PEP 432.

vstinner:
I merge this PR right now, because it fix a _Py_InitializeFromConfig() bug. _Py_InitializeMainInterpreter() must be called even if pathconfig_warnings is 0. pathconfig_warnings and _init_main must not be related. I was wrong when I removed _PyCoreConfig._init_main.

PR #13364, opened 2019-05-16
GH-81061: Fix refcount issue when returning `None` from a `ctypes.py_object` callback

dgelessus:
~~Right now this only adds a failing test for the issue, to check that the bug is properly detected in CI. Once that runs through I'll push the commit with the actual fix.~~ Done

https://bugs.python.org/issue36880

* Issue: gh-81061


isidentical (approved):

pitrou:
I don't feel competent to review ctypes issues, sorry. Perhaps @serhiy-storchaka

kumaraditya303 (changes requested):
There are merge conflicts.

dgelessus:
Fixed the conflict, thanks. I have made the requested changes; please review again.

Lib/test/test_ctypes/test_refcounts.py
```diff
@@ -97,5 +97,31 @@ def func(a, b):
         f(1, 2)
         self.assertEqual(sys.getrefcount(ctypes.c_int), a)
 
+    @support.refcount_test
+    def test_callback_py_object_none_return(self):
+        """Test that returning ``None`` from a ``py_object`` callback
+        does not affect ``None``'s refcount (bpo-36880)."""
```
eryksun:
```suggestion
        # bpo-36880: test that returning None from a py_object callback
        does not decrement the refcount of None.
```

Lib/test/test_ctypes/test_refcounts.py
```diff
@@ -97,5 +97,31 @@ def func(a, b):
...
+
+        import sys
```
eryksun:
```suggestion
        import sys

        # Increase the initial refcount to avoid crashing due to a regression.  
        refcount_margin = [None] * 20000  
```

Lib/test/test_ctypes/test_refcounts.py
```diff
@@ -97,5 +97,31 @@ def func(a, b):
...
+
+        for FUNCTYPE in (ctypes.CFUNCTYPE, ctypes.PYFUNCTYPE):
+            with self.subTest(FUNCTYPE=FUNCTYPE):
+                proto = FUNCTYPE(ctypes.py_object)
+                @proto
```
eryksun:
```suggestion
                @FUNCTYPE(ctypes.py_object)
```

Lib/test/test_ctypes/test_refcounts.py
```diff
@@ -97,5 +97,31 @@ def func(a, b):
         f(1, 2)
         self.assertEqual(sys.getrefcount(ctypes.c_int), a)
 
+    @support.refcount_test
+    def test_callback_py_object_none_return(self):
```
eryksun, line 101: AFAICT, this test is rendered pointless in 3.12 due to the adoption of [PEP 683](https://peps.python.org/pep-0683/) (Immortal Objects, Using a Fixed Refcount). Because `None` is immortal, executing `Py_INCREF(Py_None)` and `Py_DECREF(Py_None)` will no longer modify its refcount.

dgelessus: Ah, good to know. I assume this isn't implemented yet though, because as of the current state of `main` (447d061bc7b978afedd3b0148715d2153ac726c5), this test still hard-crashes without the fix.

Lib/test/test_ctypes/test_refcounts.py
dgelessus: This doesn't prevent the crash, it just delays it until after the test case finishes (when `refcount_margin` is deallocated). I think the behavior is better without this addition, because then the faulthandler stack trace points at the correct test case that caused the crash.

Lib/test/test_ctypes/test_refcounts.py
```diff
@@ -97,5 +97,31 @@ def func(a, b):
...
+                none_refcount = sys.getrefcount(None)
+                # Because None's refcount can also change for other reasons,
+                # we call func in a loop to ensure that any effects on None's
+                # refcount are clearly visible.
+                for _ in range(10000):
+                    func()
+                # Allow for small variations in None's refcount from other
+                # sources.
+                self.assertAlmostEqual(
+                    sys.getrefcount(None), none_refcount, delta=50)
```
pablogsal: Yeah, in general we don't do these kind of tests. We have an automatic refleak detection mechanism but it will only work if the path that leaks references is exercised in the test suite. This means that you can add a test that ensures that the code you are adding to `callbacks.c` is actually executed but your test doesn't need to try to detect the refleak, that will be done automatically. I suppose the reason this was never detected is that we don't have a test that actually runs that.

dgelessus: Ah okay. I was going by what the other tests in `test_refcounts` are doing. Should I adjust those as well to not check `sys.getrefcount` manually?

pablogsal: > Should I adjust those as well to not check `sys.getrefcount` manually?

Not as part of this issue (one issue per problem - only changing what's required for the problem at hand). I haven't checked those and maybe these are checked more specific things.

pablogsal: You can confirm that your test work running `./python -m test test_ctypes -R'.

Ensure that it detects the problem without your fix first :)

dgelessus: Thanks, I didn't know that just `-m test` works too (I used `-m test.regrtest`). Yes, I checked - with `-R` it now indeed detects the original problem.

Lib/test/test_ctypes/test_refcounts.py
```diff
@@ -97,5 +97,20 @@ def func(a, b):
...
+        # does not decrement the refcount of None.
+
+        for FUNCTYPE in (ctypes.CFUNCTYPE, ctypes.PYFUNCTYPE):
+            with self.subTest(FUNCTYPE=FUNCTYPE):
+                @FUNCTYPE(ctypes.py_object)
+                def func():
+                    return None
+
+                # Check that calling func does not affect None's refcount.
+                for _ in range(10000):
```
kumaraditya303, line 112: Single call should be enough to trigger the refleak checker.

dgelessus: Unfortunately not... I tested again - apparently the refleak checker (with default settings: `-m test --huntrleaks=: test.test_ctypes.test_refcounts`) only detects if `None`'s refcount is too *high*, not if it's too low. The latter case is only detected once `None`'s refcount gets to zero/negative (which triggers an assertion), and that requires a thousand or so iterations of the broken code.

pablogsal: If `None` refcount goes too small it will crash at some point assuming the rest of the code decrements normally

kumaraditya303: > Unfortunately not... I tested again - apparently the refleak checker (with default settings: -m test --huntrleaks=: test.test_ctypes.test_refcounts) only detects if None's refcount is too high, not if it's too low. The latter case is only detected once None's refcount gets to zero/negative (which triggers an assertion), and that requires a thousand or so iterations of the broken code.

I see, it is because of https://github.com/python/cpython/issues/74959.

kumaraditya303 (approved):

kumaraditya303:
Buildbots are happy, merging. Thanks for the PR.

PR #13367, opened 2019-05-16
bpo-36943 Windows test changes for Windows ARM64

paulmon:
Skip tests which are failing on ARM64.
In some cases these failures should be fixed, either before or after the initial checkin.
These test failures illustrate the current state of the port to Windows ARM64

There are 3 other Windows ARM64 PRs:

- libffi changes
- windows build changes
- one header file change

https://bugs.python.org/issue36943

@zooba @zware


zooba:
This should wait until we have ARM32 stabilized and have an ARM64 buildbot ready to go.

paulmon:
This should never be checked in.  I created this only to show where more work is needed for Windows arm64.

PR #13368, opened 2019-05-16
bpo-36945: Add _PyPreConfig.configure_locale

vstinner:
Setting _PyPreConfig.configure_locale to 0 prevents Python to modify
the LC_CTYPE locale. In that case, coerce_c_locale an
coerce_c_locale_warn are set to 0.

https://bugs.python.org/issue36945

### DIFF
--- a/Include/cpython/coreconfig.h
+++ b/Include/cpython/coreconfig.h
@@ -79,6 +79,10 @@ typedef struct {
        set to !Py_IgnoreEnvironmentFlag. */
     int use_environment;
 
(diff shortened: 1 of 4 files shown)


vstinner:
I'm not sure that we actually need this feature, see the discussion at: https://bugs.python.org/issue36945#msg342677

vstinner:
Aha, at least 2 users would like to get this feature: https://bugs.python.org/issue36945#msg342744

So I removed the [WIP] status of my PR. I rebased it and I plan to merge it soon.

PR #8601, opened 2018-08-01
bpo-24564: shutil.copystat(): ignore EINVAL on os.setxattr()

giampaolo:
https://bugs.python.org/issue24564

https://bugs.python.org/issue24564

PR #13369, opened 2019-05-16
bpo-24564: shutil.copystat(): ignore EINVAL on os.setxattr()

yingw787:
[First-time contribution]

Carbon copy of PR #8601, except using tuples instead of sets. New pull request needed because original repo fork no longer exists.

This pull request ignores errno.EINVAL (invalid argument) when copying over extended attributes, which may cause copy to fail when copying files from a filesystem that supports extended attributes to filesystem that does not support extended attributes.

This pull request may also address [bpo-36850](https://bugs.python.org/issue36850).

CC: @giampaolo @serhiy-storchaka @abadger 

https://bugs.python.org/issue24564


yingw787:
@abadger I am not sure how to add labels, but I believe #8601 needed a backport to 3.7 and 2.7 as they are currently accepting bugfixes (or maybe just 3.7 if 2.7 bugfix support will be dropped in 2020).

Misc/NEWS.d/next/Library/2019-05-16-23-40-36.bpo-24564.lIwV_7.rst
```diff
@@ -0,0 +1,3 @@
+shutil.copystat(): ignore EINVAL on os.setxattr() which may occur when copying files on NFS filesystems.
```
BoboTiG: Do you mind adding markup and reword the NEWs entry? I think something like that could be good (think to use `blurb` to format correctly the lines):
```
:func:`shutil.copystat` now ignores :const:`errno.EINVAL` on :func:`os.setxattr` which may occur when copying files on filesystems without extended attributes support. Original patch by Giampaolo Rodola, updated by Ying Wang.
```

BoboTiG (approved):
Thank you @yingw787 for your first contribution :) :champagne: 
LTGM, even better if you would take into account my remark on the NEWs entry ;)

Misc/NEWS.d/next/Library/2019-05-16-23-40-36.bpo-24564.lIwV_7.rst
yingw787: Sounds great, should be pushed 👍

Lib/shutil.py
```diff
@@ -309,15 +309,16 @@ def _copyxattr(src, dst, *, follow_symlinks=True):
...
                 raise
             return
         for name in names:
             try:
                 value = os.getxattr(src, name, follow_symlinks=follow_symlinks)
                 os.setxattr(dst, name, value, follow_symlinks=follow_symlinks)
             except OSError as e:
-                if e.errno not in (errno.EPERM, errno.ENOTSUP, errno.ENODATA):
+                if e.errno not in (errno.EPERM,
+                    errno.ENOTSUP, errno.ENODATA, errno.EINVAL):
```
giampaolo: just a cosmetic/style nitpick - it's better to do this:
```
                if e.errno not in (errno.EPERM, errno.ENOTSUP, errno.ENODATA,
                                   errno.EINVAL):
```

giampaolo (changes requested):

Lib/shutil.py
yingw787: Should be updated!

PR #13370, opened 2019-05-16
[2.7] closes bpo-36755: Suppress noisy error output in test HTTPS server.

benjaminp:
TLS 1.3 has a more efficient handshake protocol. The client can reject the server's credentials and close the connection before the server has even finished writing out all of its initial data. Depending on whether the server finishes writing the rest of its handshake before the it sees the connection is reset, the server will read an empty line or see a ECONNRESET OSError. Nothing is really wrong here with the server or client, so just suppress the error output in the OSError case to fix the test.

This fix isn't required in Python 3 because clients that reject the server's certificate will shut down the TLS layer before closing the TCP connection.

https://bugs.python.org/issue36755

### DIFF
--- a/Lib/test/ssl_servers.py
+++ b/Lib/test/ssl_servers.py
@@ -42,6 +42,11 @@ class HTTPSServer(_HTTPServer):
             raise
         return sslconn, addr
 
+    def handle_error(self, request, client_address):
+        "Suppose noisy error output by default."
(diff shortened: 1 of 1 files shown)


vstinner (approved):
LGTM.

Even if I'm not 100% confident that the change is correct, since it only impacts tests, I'm fine with it :-)

PR #13372, opened 2019-05-17
bpo-36876:  Avoid static locals.

ericsnowcurrently:
This patch touches a lot of files but does only two very specific things:

* move static locals to global scope (if they aren't truly global)
* move all locally scoped `_Py_IDENTIFIER()` to global scope

Moving these to global scope makes it easier to identify problematic usage via `Tools/c-globals/check-c-globals.py`.  It also has the side effect of dropping a number of duplicate `_Py_IDENTIFIER()`, thus saving a little bit of space. :)

The overall objective is to eliminate any global state that isn't truly process-global (and to ensure the remaining global state is thread-safe).  This is only one of the early steps.

### DIFF
--- a/Modules/_datetimemodule.c
+++ b/Modules/_datetimemodule.c
@@ -17,6 +17,17 @@
 #  include <winsock2.h>         /* struct timeval */
 #endif
 
+
+_Py_IDENTIFIER(replace);
+_Py_IDENTIFIER(tzname);
+_Py_IDENTIFIER(time);
+_Py_IDENTIFIER(struct_time);
+_Py_IDENTIFIER(fromtimestamp);
+_Py_IDENTIFIER(timetuple);
+_Py_IDENTIFIER(__getinitargs__);
+_Py_IDENTIFIER(__getstate__);
+_Py_IDENTIFIER(_strptime_datetime);
+
 #define PyDate_Check(op) PyObject_TypeCheck(op, &PyDateTime_DateType)
 #define PyDate_CheckExact(op) (Py_TYPE(op) == &PyDateTime_DateType)
 
@@ -1228,7 +1239,6 @@ static PyObject *
 call_tzname(PyObject *tzinfo, PyObject *tzinfoarg)
 {
     PyObject *result;
-    _Py_IDENTIFIER(tzname);
 
     assert(tzinfo != NULL);
     assert(check_tzinfo_subclass(tzinfo) >= 0);
@@ -1418,7 +1428,6 @@ make_Zreplacement(PyObject *object, PyObject *tzinfoarg)
     PyObject *temp;
     PyObject *tzinfo = get_tzinfo_member(object);
     PyObject *Zreplacement = PyUnicode_FromStringAndSize(NULL, 0);
-    _Py_IDENTIFIER(replace);
 
     if (Zreplacement == NULL)
         return NULL;
@@ -1657,7 +1666,6 @@ time_time(void)
     PyObject *time = PyImport_ImportModuleNoBlock("time");
 
     if (time != NULL) {
-        _Py_IDENTIFIER(time);
 
         result = _PyObject_CallMethodId(time, &PyId_time, NULL);
         Py_DECREF(time);
@@ -1673,7 +1681,6 @@ build_struct_time(int y, int m, int d, int hh, int mm, int ss, int dstflag)
 {
     PyObject *time;
     PyObject *result;
-    _Py_IDENTIFIER(struct_time);
     PyObject *args;
 
 
@@ -2883,7 +2890,6 @@ date_today(PyObject *cls, PyObject *dummy)
 {
     PyObject *time;
     PyObject *result;
-    _Py_IDENTIFIER(fromtimestamp);
 
     time = time_time();
     if (time == NULL)
@@ -3182,7 +3188,6 @@ date_strftime(PyDateTime_Date *self, PyObject *args, PyObject *kw)
     PyObject *result;
     PyObject *tuple;
     PyObject *format;
-    _Py_IDENTIFIER(timetuple);
     static char *kwlist[] = {"format", NULL};
 
     if (! PyArg_ParseTupleAndKeywords(args, kw, "U:strftime", kwlist,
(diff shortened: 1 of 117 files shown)


asvetlov (approved):
asyncio part looks good

Modules/_decimal/libmpdec/context.c
```diff
@@ -48,7 +48,7 @@ void (* mpd_traphandler)(mpd_context_t *) = mpd_dflt_traphandler;
 void
 mpd_setminalloc(mpd_ssize_t n)
 {
-    static int minalloc_is_set = 0;
+    static int minalloc_is_set = 0;  // Static is okay here (process-global).
```
skrah, line 51: Please don't change `libmpdec`, even if it's just a comment. :)

vstinner (changes requested):
> move all locally scoped _Py_IDENTIFIER() to global scope

I dislike this change. I prefer to keep "local" static variables in functions. I don't see how moving them as globals is solving any issue, apart your practical issue of detecting them with your tool.

I would prefer to see an API to support static variable per interpreter.

I would suggest to develop an hash table similar to thread TLS: each static variable would have a global unique identifier, and its value would stored in an hash table in the PyInterpreterState. The problem is how to initialize the global unique identifier in a safe way (thread safety / atomicity). But that could be implemented using "local static" variables.

Modules/_datetimemodule.c
```diff
@@ -1336,10 +1346,10 @@ tzinfo_from_isoformat_results(int rv, int tzoffset, int tz_useconds)
 static PyObject *
 format_ctime(PyDateTime_Date *date, int hours, int minutes, int seconds)
 {
-    static const char * const DayNames[] = {
+    static const char * const DayNames[] = {  // Static is okay here (immutable data).
```
vstinner, line 1349: Are these comments really useful? To me, it seems obvious that static constants are safe. I'm not sure of the value of such comment.

vstinner:
> (...) The problem is  how to initialize the global unique identifier in a safe way (thread  safety / atomicity). But that could be implemented using "local static"  variables.

I forgot to mention that "local static variables" would be easier to initialize properly, since they cannot be accessed before the function is called: before Python is initialized.

The case of global static variables is worse: we cannot statically initialize them, we need something to dynamically initialize them. Maybe at the first access?

About the global unique identifier, one solution would be to use the memory address of the variable: it should be unique. I don't see why a compiler would merge two variables.

Draft pseudo API:

```
void get_interpreter_specific(uintptr_t key, size_t size, void *value);
int set_interpreter_specific(uintptr_t key, size_t size, void *value);

#define PyStaticVariable(TYPE) static TYPE
// FIXME: add assertion to ensure that VAR is a PyStaticVariable
#define PyStaticVariable_get(VAR, VALUE) \
   get_interpreter_specific((uintptr_t)&(VAR), sizeof(VAR), &(VALUE))
#define PyStaticVariable_set(VAR, VALUE) \
   set_interpreter_specific((uintptr_t)&(VAR), sizeof(VAR), &(VALUE))

PyObject* func(void)
{
PyStaticVariable(int) var;

// if var doesn't exist in the interpreter local storage,
// its value is always 0 by default, for any type,
// as global static variables in C
int value;
PyStaticVariable_get(var, value);
value += 1;
if (PyStaticVariable_set(var, value) < 0) { return NULL; }

Py_RETURN_VALUE;
}
```

Problem: we cannot pre-allocate the storage of "var" value into the interpreter "local storage" hash table, so set_interpreter_specific() can fail with a memory allocation failure :-(

tiran (approved):

tiran:
@ericsnowcurrently Please re-generate the ast module:

```
Generated files not up to date
 M Python/Python-ast.c
```

skrah (changes requested):
As Victor has also requested, I'd remove all comments that mention that variables are process local.

Modules/_datetimemodule.c
tiran: A comment sounds useful to flag all local statics as "good" and for a linter that checks for unverified local static vars. I'm ok with the comments, but maybe a shorter one like ``// static ok``.?

berkerpeksag: I agree with @vstinner. If you really want to keep these comments, please move them to separate lines to avoid noise when using `git blame`.

Modules/_datetimemodule.c
```diff
@@ -3007,13 +3013,13 @@ date_fromisoformat(PyObject *cls, PyObject *dtstr)
 static PyObject *
 date_fromisocalendar(PyObject *cls, PyObject *args, PyObject *kw)
 {
-    static char *keywords[] = {
+    static char *kwlist[] = {
```
gpshead, line 3016: is there a reason all of the kwlist's everywhere cannot become static const?  (lemme guess.. our C API complains about the type?)

Objects/classobject.c
```diff
@@ -116,13 +119,13 @@ static PyMemberDef method_memberlist[] = {
...
 method_get_doc(PyMethodObject *im, void *context)
 {
-    static PyObject *docstr;
-    if (docstr == NULL) {
-        docstr= PyUnicode_InternFromString("__doc__");
-        if (docstr == NULL)
-            return NULL;
-    }
-    return PyObject_GetAttr(im->im_func, docstr);
+//    if (cached_str___doc__ == NULL) {
```
gpshead, line 122: checking in a block of commented out code looks wrong... what's up here?

Python/pythonrun.c
```diff
@@ -39,17 +39,27 @@
...
+_Py_IDENTIFIER(code);
+_Py_IDENTIFIER(encoding);
 _Py_IDENTIFIER(excepthook);
+_Py_IDENTIFIER(filename);
 _Py_IDENTIFIER(flush);
 _Py_IDENTIFIER(last_traceback);
 _Py_IDENTIFIER(last_type);
 _Py_IDENTIFIER(last_value);
+_Py_IDENTIFIER(lineno);
+_Py_IDENTIFIER(msg);
```
gpshead, line 54: could the _Py_IDENTIFIER macro be modified such that it _only_ works from the global scope so that these don't creep back into function bodies?

gpshead (changes requested):
overall i'm a little concerned about how fragile this is.  it won't be obvious in all code why something static lives outside of the only place it is used.  it feels like we need a way to prevent future C code authors from making this mistake.

ericsnowcurrently:
Thanks for all the feedback, everyone!  I'm going to re-think this as you've all made good points. :)  I may re-open the PR if it later makes sense to re-purpose it, but I'll probably instead tackle the statics in groups by `Modules`/`Objects`/`Python`.

PR #13373, opened 2019-05-17
bpo-36763: Remove _PyCoreConfig.program

vstinner:


vstinner:
@Yhg1s: Oops, while inspecting _PyCoreConfig to review their documentation, I understood that program is kind of redundant wih program_name. Python 3.6 used argv[0] directly to display command line usage / help. I introduced "program". But after a lot of refactoring (finished by this PR), I managed to use "program_name" which is the **correct** program that should be used to display the command line usage / help.

PR #13377, opened 2019-05-17
Doc: Replace the deprecated highlightlang directive by highlight.

matrixise:
highlightlang is deprecated since April 2018 in Sphinx and will be removed in Sphinx 4.0
See https://github.com/sphinx-doc/sphinx/pull/4845

@JulienPalard for the `skip news` I am not sure because we will stop to use `.. highlightlang ::` and it's not a feature/bugfix of Python.

### DIFF
--- a/Doc/c-api/abstract.rst
+++ b/Doc/c-api/abstract.rst
@@ -1,4 +1,4 @@
-.. highlightlang:: c
+.. highlight:: c
 
 .. _abstract:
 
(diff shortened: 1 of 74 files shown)


kushaldas (approved):
The change looks good, the docs built from this branch looks exactly the same like in the old version.
There are no more instance of `hightlightlang` left in the `Doc/` directory.

PR #9084, opened 2018-09-06
bpo-25750: add testcase on bad descriptor __get__()

jdemeyer:
This adds a testcase for [bpo-25750](https://www.bugs.python.org/issue25750) (#6118). This is a separate pull request since the testcase is possibly controversial.

https://bugs.python.org/issue25750

PR #13381, opened 2019-05-17
bpo-36907: fix refcount bug in _PyStack_UnpackDict()

jdemeyer:
A minimal patch fixing [bpo-36907](https://bugs.python.org/issue36907)

CC @encukou 

### DIFF
--- a/Lib/test/test_call.py
+++ b/Lib/test/test_call.py
@@ -8,6 +8,7 @@ except ImportError:
 import struct
 import collections
 import itertools
+import gc
 
 
 class FunctionCalls(unittest.TestCase):
@@ -466,6 +467,7 @@ class FastCallTests(unittest.TestCase):
                 self.kwargs = kwargs
             def __index__(self):
                 self.kwargs.clear()
+                gc.collect()
                 return 0
         x = IntWithDict(dont_inherit=IntWithDict())
         # We test the argument handling of "compile" here, the compilation
--- a/Lib/test/test_call.py
+++ b/Lib/test/test_call.py
@@ -457,6 +457,21 @@ class FastCallTests(unittest.TestCase):
                 result = _testcapi.pyobject_fastcallkeywords(func, args, kwnames)
                 self.check_result(result, expected)
 
+    def test_fastcall_clearing_dict(self):
+        # Test bpo-36907: the point of the test is just checking that this
+        # does not crash.
+        class IntWithDict:
+            __slots__ = ["kwargs"]
+            def __init__(self, **kwargs):
+                self.kwargs = kwargs
+            def __index__(self):
+                self.kwargs.clear()
+                return 0
+        x = IntWithDict(dont_inherit=IntWithDict())
+        # We test the argument handling of "compile" here, the compilation
+        # itself is not relevant. When we pass flags=x below, x.__index__() is
+        # called, which changes the keywords dict.
+        compile("pass", "", "exec", x, **x.kwargs)
 
 if __name__ == "__main__":
     unittest.main()
--- a/Objects/call.c
+++ b/Objects/call.c
@@ -544,10 +544,14 @@ _PyMethodDef_RawFastCallDict(PyMethodDef *method, PyObject *self,
         }
 
         result = (*fastmeth) (self, stack, nargs, kwnames);
-        if (stack != args) {
+        if (kwnames != NULL) {
+            Py_ssize_t i, n = nargs + PyTuple_GET_SIZE(kwnames);
+            for (i = 0; i < n; i++) {
+                Py_DECREF(stack[i]);
+            }
             PyMem_Free((PyObject **)stack);
+            Py_DECREF(kwnames);
         }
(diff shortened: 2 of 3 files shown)


jdemeyer:
I decided to increase the refcount in `_PyStack_UnpackDict` after all because that seemed the most natural place (and because I didn't want to contradict the "borrowed references" comments). If you disagree, it would also be possible to leave `_PyStack_UnpackDict` alone and increase the refcounts in the code calling `_PyStack_UnpackDict`.

This should probably be backported to 3.7 too.

Objects/call.c
```diff
@@ -1334,8 +1338,11 @@ _PyStack_UnpackDict(PyObject *const *args, Py_ssize_t nargs, PyObject *kwargs,
         return -1;
     }
 
-    /* Copy position arguments (borrowed references) */
-    memcpy(stack, args, nargs * sizeof(stack[0]));
+    /* Copy positional arguments */
+    for (i = 0; i < nargs; i++) {
+        Py_INCREF(args[i]);
+        stack[i] = args[i];
+    }
```
encukou, line 1345: Is there a reason to not keep `memcpy` for the bulk copy, and only add an INCREF loop?

jdemeyer: > Is there a reason to not keep `memcpy` for the bulk copy, and only add an INCREF loop?

Not really, it's mostly a matter of style. I like doing the `INCREF` at the same time as the assignment. I also find the `for` loop variant simpler to understand.

Lib/test/test_call.py
```diff
@@ -457,6 +457,19 @@ def test_fastcall_keywords(self):
...
+        class IntWithDict:
+            __slots__ = ["kwargs"]
+            def __init__(self, **kwargs):
+                self.kwargs = kwargs
+            def __index__(self):
+                self.kwargs.clear()
+                L = [2**i for i in range(10000)]
+                return 0
+        x = IntWithDict(dont_inherit=IntWithDict())
+        compile("pass", "", "exec", x, **x.kwargs)
```
encukou, line 476: compile is quite a special function, so I recommend to add a note:

```suggestion
        # Note: we test compile's argument handling, not actual compilation
        compile("pass", "", "exec", x, **x.kwargs)
```

Lib/test/test_call.py
```diff
@@ -457,6 +457,19 @@ def test_fastcall_keywords(self):
...
+    def test_fastcall_clearing_dict(self):
+        # Test bpo-36907: the point of the test is just checking that this
+        # does not crash.
+        class IntWithDict:
+            __slots__ = ["kwargs"]
+            def __init__(self, **kwargs):
+                self.kwargs = kwargs
+            def __index__(self):
+                self.kwargs.clear()
+                L = [2**i for i in range(10000)]
```
encukou: This is supposed to trigger garbage collection, right? Why not make it explicit?

```suggestion
                gc.collect()
```

(with `import gc` at the top)

jdemeyer: > This is supposed to trigger garbage collection, right? Why not make it explicit?

Not really, it is meant to clobber memory. The bug is an use-after-free but crashing is only guaranteed if the used-after-freed memory is corrupted. But I'll add a comment explaining that.

encukou: IMO we just leave that to the debug allocator, or valgrind runs.

jdemeyer: > IMO we just leave that to the debug allocator, or valgrind runs.

Fine for me, but I have to mention that a similar thing was accepted by you(!) in #9084.

encukou: I hope the things I learned since then are useful.

@vstinner, what's your view on this? Is it useful to clobber memory with `L = [2**i for i in range(10000)]`?

vstinner: > @vstinner, what's your view on this? Is it useful to clobber memory with L = [2**i for i in range(10000)]?

I'm sorry but I don't have the bandwidth right now to follow the discussion on PEP 590 and it's implementation :-( Retry to ping me in 1 or 2 weeks if you want ;-)

encukou: @vstinner Ah, sorry. This isn't really about PEP 590, though. It's a refcounting bug, an object is deallocated while it's still in use.

In normal builds, the bug will manifest only if you clobber the memory after it is deallocated. The debug allocator (or valgrind) should catch it even if the `L = [2**i for i in range(10000)]` is not done.

Is it OK to rely on the debug allocator (or valgrind) to prevent regressions?

(If you don't have bandwidth for that either, let me know)

jdemeyer: In the mean time, I removed that line.

jdemeyer:
Can this be merged please? That would be good for the PEP 590 implementation, which uses this function.

encukou (approved):
I took the liberty to add a `gc.collect()` call.

encukou:
https://github.com/python/cpython/pull/13493 is a backport to 3.7.

PR #13386, opened 2019-05-17
bpo-32972: Async test case

asvetlov:
Add explicit `asyncSetUp` and `asyncTearDown` methods.
The rest is the same as for #13228

`AsyncTestCase` create a loop instance for every test for the sake of test isolation.
Sometimes a loop shared between all tests can speed up tests execution time a lot but it requires control of closed resources after every test finish. Basically, it requires nested supervisors support that was discussed with @1st1 many times. Sorry, asyncio supervisors have no chance to land on Python 3.8.

The PR intentionally does not provide API for changing the used event loop or getting the test loop: use `asyncio.set_event_loop_policy()` and `asyncio.get_event_loop()` instead.

The PR adds four overridable methods to base `unittest.TestCase` class:
```
    def _callSetUp(self):
        self.setUp()

    def _callTestMethod(self, method):
        method()

    def _callTearDown(self):
        self.tearDown()

    def _callCleanup(self, function, /, *args, **kwargs):
        function(*args, **kwargs)
```
It allows using asyncio facilities with minimal influence on the unittest code.

The last but not least: the PR respects contextvars. The context variable installed by `asyncSetUp` is available on test, `tearDown` and a coroutine scheduled by `addCleanup`.

https://bugs.python.org/issue32972


Lib/unittest/async_case.py
```diff
@@ -0,0 +1,143 @@
...
+
+    def _callTestMethod(self, method):
+        self._callMaybeAsync(method)
+
+    def _callTearDown(self):
+        self._callAsync(self.asyncTearDown)
+        self.tearDown()
+
+    def _callCleanup(self, function, *args, **kwargs):
+        self._callMaybeAsync(function, *args, **kwargs)
```
1st1, line 72: Why don't we have an explicit `addAsyncCleanup`?

1st1: For example, we have an explicit async method in AsyncExitStack: https://docs.python.org/3/library/contextlib.html#contextlib.AsyncExitStack.push_async_callback

Even though we toyed with the idea of just reusing the push_callback method and checking if the returned value is an awaitable.

I'm more in favour of `addAsyncCleanup` in this case.

Lib/unittest/async_case.py
```diff
@@ -0,0 +1,143 @@
+import asyncio
+import inspect
+
+from .case import TestCase
+
+
+
+class AsyncioTestCase(TestCase):
```
1st1: Can we call it `IsolatedAsyncioTestCase` to signify that the loop is not shared between the tests?

Lib/unittest/async_case.py
asvetlov: I can add `addAsyncCleanup()` method, it makes sense.
The implementation should push a callback into the same cleanup queue as used for `addCleanup()` though, to keep cleanups order.
I think running all sync cleanups before asyncs and vice versa is a bad idea.
Cleanups should be executed exactly in the reverse adding order.
Also, `self.doCleanups()` should execute all sync and async cleanups altogether.
Do you agree?

Lib/unittest/async_case.py
asvetlov: We can.
I have a hope that non-isolated tests can reuse the same test case but require a little different test runner maybe (to keep a loop during the whole test run session).

Anyway, I'm ok with renaming the class right now and see what's happen when we add a shared loop environment.

asvetlov: Renamed.
Not sure if I like it too much (the long name becomes even longer) but can live with it, sure.

auvipy:
this would be great to see in 3.8

Lib/unittest/async_case.py
1st1: Yes.

1st1 (approved):
Please add `addAsyncCleanup()` before merging.

Lib/unittest/async_case.py
1st1: I don't like it too much either,  but the predictability of the behaviour of this new base class comes first.

asvetlov:
`addAsyncCleanup()` is added

auvipy (approved):

abetkin:
Hi all, I tested unittest in 3.8 and what I see is tests are run not concurrently: I have 5 tests that sleep for 3 seconds and the total time of execution is 15 seconds. What is the reason behind this?

PR #13389, opened 2019-05-17
bpo-36948: Fix NameError in urllib.request.URLopener.retrieve

tirkarthi:
Fix NameError in urllib.request.URLopener.retrieve by using correct imports.

https://bugs.python.org/issue36948

### DIFF
--- a/Lib/test/test_urllib.py
+++ b/Lib/test/test_urllib.py
@@ -1464,11 +1464,15 @@ class URLopener_Tests(unittest.TestCase, FakeHTTPMixin):
                 "//c:|windows%/:=&?~#+!$,;'@()*[]|/path/")
 
     def test_urlopener_retrieve_file(self):
-        with tempfile.NamedTemporaryFile() as file_obj:
-            with self.assertWarns(DeprecationWarning):
-                url = f'file://localhost{file_obj.name}'
-                filename, _ = urllib.request.URLopener().retrieve(url)
-                self.assertEqual(filename, file_obj.name)
+        with self.assertWarns(DeprecationWarning):
+            try:
+                fd, tmp_file = tempfile.mkstemp()
+                fileurl = 'file:' + urllib.request.pathname2url(tmp_file)
+                filename, _ = urllib.request.URLopener().retrieve(fileurl)
+                self.assertEqual(filename, tmp_file)
+            finally:
+                os.close(fd)
+                os.unlink(tmp_file)
 
     def test_urlopener_retrieve_remote(self):
         with self.assertWarns(DeprecationWarning):
--- a/Lib/test/test_urllib.py
+++ b/Lib/test/test_urllib.py
@@ -1465,10 +1465,12 @@ class URLopener_Tests(unittest.TestCase, FakeHTTPMixin):
 
     @support.ignore_warnings(category=DeprecationWarning)
     def test_urlopener_retrieve_file(self):
-        with tempfile.NamedTemporaryFile() as fileobj:
-            fileurl = "file:" + urllib.request.pathname2url(fileobj.name)
-            filename, _ = urllib.request.URLopener().retrieve(fileurl)
-            self.assertEqual(filename, fileobj.name)
+        fd, tmpfile = tempfile.mkstemp()
+        self.addCleanup(os.close, fd)
+        self.addCleanup(os.unlink, tmpfile)
+        fileurl = "file:" + urllib.request.pathname2url(tmpfile)
+        filename, _ = urllib.request.URLopener().retrieve(fileurl)
+        self.assertEqual(filename, tmpfile)
 
     @support.ignore_warnings(category=DeprecationWarning)
     def test_urlopener_retrieve_remote(self):
--- a/Lib/test/test_urllib.py
+++ b/Lib/test/test_urllib.py
@@ -1467,7 +1467,7 @@ class URLopener_Tests(unittest.TestCase, FakeHTTPMixin):
     def test_urlopener_retrieve_file(self):
         with support.temp_dir() as tmpdir:
             fd, tmpfile = tempfile.mkstemp(dir=tmpdir)
-            self.addCleanup(os.close, fd)
+            os.close(fd)
             fileurl = "file:" + urllib.request.pathname2url(tmpfile)
             filename, _ = urllib.request.URLopener().retrieve(fileurl)
             self.assertEqual(filename, tmpfile)
--- a/Lib/test/test_urllib.py
+++ b/Lib/test/test_urllib.py
@@ -1466,7 +1466,7 @@ class URLopener_Tests(unittest.TestCase, FakeHTTPMixin):
     def test_urlopener_retrieve_file(self):
         with tempfile.NamedTemporaryFile() as file_obj:
             with self.assertWarns(DeprecationWarning):
-                url = f'file://{file_obj.name}'
+                url = f'file://localhost{file_obj.name}'
                 filename, _ = urllib.request.URLopener().retrieve(url)
                 self.assertEqual(filename, file_obj.name)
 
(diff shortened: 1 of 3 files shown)


isidentical (changes requested):
Do this PR needs a news entry?

tirkarthi:
> Do this PR needs a news entry?

@isidentical I have added one already at https://github.com/python/cpython/pull/13389/files#diff-1f45e92fbc7f2721e27165810e333552

isidentical:
I'm just asking, i dont think this is a change that requires a news entry.

tirkarthi:
> I'm just asking, i dont think this is a change that requires a news entry.

This is a user impacting bug where even though the function was deprecated it's now causing `NameError` that used to work fine. IMO, it's a bug fix and requires a NEWS entry.

isidentical (approved):

Lib/test/test_urllib.py
```diff
@@ -1463,6 +1463,28 @@ def open_spam(self, url):
                 "spam://c:|windows%/:=&?~#+!$,;'@()*[]|/path/"),
                 "//c:|windows%/:=&?~#+!$,;'@()*[]|/path/")
 
+    def test_urlopener_retrieve_file(self):
+        with self.assertWarns(DeprecationWarning):
```
berkerpeksag: `DeprecationWarning` is already tested elsewhere. It's out of scope of this test. Use the `support.ignore_warnings` decorator to suppress it.

Lib/test/test_urllib.py
```diff
@@ -1463,6 +1463,28 @@ def open_spam(self, url):
...
+            try:
+                fd, tmp_file = tempfile.mkstemp()
```
berkerpeksag: Can't we use `tempfile.NamedTemporaryFile` and context management protocol to avoid handling resources manually?

Lib/test/test_urllib.py
```diff
@@ -1463,6 +1463,28 @@ def open_spam(self, url):
...
+                fileurl = 'file:' + urllib.request.pathname2url(tmp_file)
+                filename, _ = urllib.request.URLopener().retrieve(fileurl)
+                self.assertEqual(filename, tmp_file)
+            finally:
+                os.close(fd)
+                os.unlink(tmp_file)
+
+    def test_urlopener_retrieve_remote(self):
+        with self.assertWarns(DeprecationWarning):
```
berkerpeksag: Same comment as above.

Lib/test/test_urllib.py
```diff
@@ -1463,6 +1463,28 @@ def open_spam(self, url):
...
+            try:
+                self.fakehttp(b"HTTP/1.1 200 OK\r\n\r\nHello!")
+                url = "http://www.python.org/file.txt"
+                filename, _ = urllib.request.URLopener().retrieve(url)
+                self.assertEqual(os.path.splitext(filename)[1], '.txt')
+            finally:
+                self.unfakehttp()
```
berkerpeksag: Use of `self.addCleanup(self.unfakehttp)` should help to get rid of the whole try...finally block.

Lib/test/test_urllib.py
```diff
@@ -1463,6 +1463,28 @@ def open_spam(self, url):
...
+                os.close(fd)
+                os.unlink(tmp_file)
+
+    def test_urlopener_retrieve_remote(self):
+        with self.assertWarns(DeprecationWarning):
+            try:
+                self.fakehttp(b"HTTP/1.1 200 OK\r\n\r\nHello!")
+                url = "http://www.python.org/file.txt"
+                filename, _ = urllib.request.URLopener().retrieve(url)
+                self.assertEqual(os.path.splitext(filename)[1], '.txt')
```
berkerpeksag: Nit: Please use your choice of quoting style consistently in new code.

berkerpeksag (changes requested):

Lib/test/test_urllib.py
tirkarthi: I guess I used it due to Windows permission error in previous CI runs where the created file was not readable but with `mkstemp` it's possible as seen from the docstring. It seems to be a windows only case but it seems to have other problem where the directory could be used by another process as seen in current CI run. Is there a reliable way to generate a temp file for this usecase?

```
    The file is readable and writable only by the creating user ID.
    If the operating system uses permission bits to indicate whether a
    file is executable, the file is executable by no one. The file
    descriptor is not inherited by children of this process.
```

tirkarthi:
I have resorted to using `support.temp_dir` to create a temporary directory for `mkstemp` . I guess the file being used error was to adding `os.close(fd)` to Cleanup where it's closed at the end of tests though the context manager `support.temp_dir` tries to clean it up on `__exit__` causing error on cleaning up tmp_dir with an open file handle. So I have added `os.close(fd)` on the next line itself. I must admit that I resolved the issue with trial and error since I don't have access to Windows. Suggestions if any windows compatible test would be helpful too.

Lib/test/test_urllib.py
berkerpeksag: I think `support.unlink()` can be used to handle "it's being used by another process" errors under Windows.

Also, the order of `self.addCleanup()` calls is important. You might be bitten by a similar issue like https://github.com/python/cpython/commit/d0f5bab21b9828f4764bd45e4afe24f23c1a55b0.

Lib/test/test_urllib.py
```diff
@@ -1445,7 +1445,7 @@ def test_thishost(self):
         self.assertIsInstance(urllib.request.thishost(), tuple)
 
 
-class URLopener_Tests(unittest.TestCase):
+class URLopener_Tests(unittest.TestCase, FakeHTTPMixin):
```
berkerpeksag: The mixin class should be applied first.

Lib/test/test_urllib.py
tirkarthi: NamedTemporaryFile gives me that the file is not readable. Hence I used temp_dir and mkstemp to generate a more safer version of temp file for this test but I am sure I am not the only one having this problem in the testsuite.

Lib/test/test_urllib.py
tirkarthi: All the tests in `test_urllib` use `Mixin` after `TestCase`. Grepping tells me both variants are used across test suite. Is there any difference I am missing since I am using the mixin only for `fakehttp`?

Lib/test/test_urllib.py
berkerpeksag: I don't think calling `os.close()` just after `tempfile.mkstemp()` makes sense. The temp directory will be cleaned up after the `support.temp_dir()` context manager and `fd` will be effectively invalid.

Lib/test/test_urllib.py
berkerpeksag: Yes, there are a dozen of incorrect usages in the test suite (mainly in Lib/test/test_ipaddress.py, Lib/test/test_marshal.py, Lib/test/test_posix.py, Lib/test/test_runpy.py, Lib/test/test_statistics.py, Lib/test/test_with.py) but that doesn't justify adding non-idiomatic code to the test suite just because we are only using `some_unique_helper` from the mixin class.

Contributors often use existing tests as examples when writing new tests for their PRs, so it would be nice to write idiomatic tests as much as possible.

Lib/test/test_urllib.py
tirkarthi: Without calling `os.close(fd)` there will be an open file handle to the tempfile inside the directory and there will be a permission error that file handle is used by another process cannot be cleaned up as `support.temp_dir` context manager exits and due to unclosed fd. I guess Windows doesn't allow deleting directory when there is an open file handle to one of the files inside.

Lib/test/test_urllib.py
tirkarthi: Changed as suggested, thanks.

Lib/test/test_urllib.py
berkerpeksag: Ah, now I understood what you are talking about, and, yes, your analysis is correct. I think the current version of the test is good and reliable enough. Thanks!

berkerpeksag (approved):

tirkarthi:
Thanks @berkerpeksag for the review. This is a 3.8 only issue so I guess the backport PR to 3.7 can be closed.

PR #13390, opened 2019-05-17
bpo-36763: _Py_RunMain() doesn't call Py_Exit() anymore

vstinner:
Py_Main() and _Py_RunMain() now return the exitcode rather than
calling Py_Exit(exitcode) when calling PyErr_Print() if the current
exception type is SystemExit and -i command line is used
(core_config.inspect=1).

* Add _Py_HandleSystemExit()
* Add pymain_exit_err_print()
* Add pymain_exit_print()


vstinner:
@Yhg1s: This PR fix the bugs you told me about, but it doesn't cover the PyRun_SimpleXXX() APIs.

PR #6966, opened 2018-05-18
bpo-33524: Fix the behavior when max_line_length is 0 or None in email.policy

Licht-T:
This fixes [bpo-33524](https://bugs.python.org/issue33524).

https://bugs.python.org/issue33524

PR #13391, opened 2019-05-17
bpo-33524: Fix the folding of email header when max_line_length is 0 or None

maxking:
Updated version of #6966 with requested NEWS entry since that PR looks abandoned.

Original Patch by @Licht-T 

https://bugs.python.org/issue33524


Misc/NEWS.d/next/Library/2019-05-17-11-44-21.bpo-33524.8y_xUU.rst
```diff
@@ -0,0 +1,2 @@
+Fix the folding of email header when the max_line_length is 0 or None.
```
bitdancer: To be clear about what we are fixing this probably ought to say "...is 0 or None and the header contains non-ascii characters".

bitdancer (changes requested):

maxking:

bitdancer (approved):

PR #13392, opened 2019-05-17
bpo-36929: Modify io/re tests to allow for missing mod name

tekknolagi:
For a vanishingly small number of internal types, CPython sets the
tp_name slot to mod_name.type_name, either in the PyTypeObject or the
PyType_Spec. There are a few minor places where this surfaces:

* Custom repr functions for those types (some of which ignore the
  tp_name in favor of using a string literal, such as _io.TextIOWrapper)
* Pickling error messages

The test suite only tests the former. This commit modifies the test
suite to allow Python implementations to omit the module prefix.

https://bugs.python.org/issue36929

### DIFF
--- a/Lib/test/test_io.py
+++ b/Lib/test/test_io.py
@@ -1119,12 +1119,12 @@ class CommonBufferedTests:
     def test_repr(self):
         raw = self.MockRawIO()
         b = self.tp(raw)
-        clsname = "%s.%s" % (self.tp.__module__, self.tp.__qualname__)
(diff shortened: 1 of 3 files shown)


DinoV:
Tested on Mac OS/X, test changes look good, some lines >80 got a little longer, but they're consistent with some other long lines in the same file, and there's no great way to split them.  This should make the test suite more flexible for other Python implementations while still testing the things that matter.

DinoV (approved):

PR #13396, opened 2019-05-17
bpo-36763: _Py_InitializeFromArgs() argc becomes Py_ssize_t

vstinner:
* The type of initlization function 'argc' parameters becomes
  Py_ssize_t, instead of int.
* Change _PyPreConfig_Copy() return type to void, instead of int.
  The function cannot fail anymore.
* Fix compilation warnings on Windows.


vstinner:
I checked AppVeyor: this change fixed all compiler warnings.

PR #13399, opened 2019-05-18
bpo-29847: Path subclasses raise TypeError given kwargs

jsaporta:
TypeError raised in the subclasses as per the discussion on bugs.python.org.

https://bugs.python.org/issue29847


the-knights-who-say-ni:

Lib/pathlib.py
```diff
@@ -1512,13 +1512,23 @@ class PosixPath(Path, PurePosixPath):
     """
     __slots__ = ()
 
+    def __new__(cls, *args, **kwargs):
+        if kwargs:
+            raise TypeError("keyword arguments provided but ignored")
```
eamanu, line 1524: What about start the raise with uppercase?

Lib/pathlib.py
```diff
@@ -1512,13 +1512,23 @@ class PosixPath(Path, PurePosixPath):
...
+
 class WindowsPath(Path, PureWindowsPath):
     """Path subclass for Windows systems.
 
     On a Windows system, instantiating a Path should return this object.
     """
     __slots__ = ()
 
+    def __new__(cls, *args, **kwargs):
+        if kwargs:
```
eamanu: This weird for me because if you don't have kwargs on ```__new__(cls, *args, **kwargs)```
because this will raise a exception, what are you sending on ```super().__new__(cls, *args, **kwargs)```

Lib/pathlib.py
jsaporta: Other errors I've seen start with lowercase letters.
```python
>>> 3 + 'string'
TypeError: unsupported operand type(s) for +: 'int' and 'str'
```

Lib/pathlib.py
jsaporta: Not sure I understand you clearly, but if `kwargs` is an empty dictionary, then it will act as a falsy value in the if statement, and the exception will not be raised.

tirkarthi:
I would suggest adding a NEWS entry and perhaps a versionchanged section to doc since someone could have been passing kwargs though it was silently ignored and undocumented in the past that now raises a TypeError.

EpicWink:
What about a test to see if a new subclass of `pathlib.Path` (or `pathlib.PosixPath` or `pathlib.WindowsPath`) can still use keyword-only parameters in the initialiser?
```python
import pathlib
class MyPath(type(pathlib.Path())):
    def __init__(self, *args, foo):
        self.foo = foo
path = MyPath('bar', foo=42)
```

gvanrossum:
It's not cool to spam the entire team with code review requests.

jsaporta:
I synced my fork with master then rebased the issue branch on top of it before updating it. Looks like that may have been a mistake?

methane:
Please use merge instead of rebase next time.
ref: https://github.com/python/devguide/pull/496

brettcannon:
I'm closing this as it has 893 changed files so there's no chance to easily view what the PR is actually targeting.

PR #13401, opened 2019-05-18
bpo-2180: Treat line continuation at EOF as a `SyntaxError`

asottile:
This makes the parser consistent with the tokenize module (already the case
in `pypy`).

sample
------

```python
x = 5\
```

before
------

```console
$ python3 t.py
$ python3 -mtokenize t.py
t.py:2:0: error: EOF in multi-line statement
```

after
-----

```console
$ ./python t.py
  File "t.py", line 3
    x = 5\

         ^
SyntaxError: unexpected EOF while parsing
$ ./python -m tokenize t.py
t.py:2:0: error: EOF in multi-line statement
```

https://bugs.python.org/issue2180

### DIFF
--- a/Lib/test/test_eof.py
+++ b/Lib/test/test_eof.py
@@ -24,5 +24,11 @@ class EOFTestCase(unittest.TestCase):
         else:
             raise support.TestFailed
 
+    def test_line_continuation_EOF(self):
+        expect = 'unexpected EOF while parsing (<string>, line 1)'
+        with self.assertRaises(SyntaxError) as excinfo:
+            exec('x = 5\\')
+        self.assertEqual(str(excinfo.exception), expect)
+
 if __name__ == "__main__":
     unittest.main()
--- a/Lib/test/test_eof.py
+++ b/Lib/test/test_eof.py
@@ -1,7 +1,10 @@
 """test script for a few new invalid token catches"""
 
-import unittest
+import subprocess
+import sys
+import tempfile
 from test import support
+import unittest
 
 class EOFTestCase(unittest.TestCase):
     def test_EOFC(self):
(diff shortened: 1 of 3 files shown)


Parser/tokenizer.c
```diff
@@ -983,7 +983,8 @@ tok_nextc(struct tok_state *tok)
                         return EOF;
                     /* Last line does not end in \n,
                        fake one */
-                    strcpy(tok->inp, "\n");
+                    if (tok->inp[-1] != '\n')
```
gpshead, line 986: I added an additional test to exercise this logic (it requires parsing a file).

gpshead (approved):

Lib/test/test_eof.py
```diff
@@ -24,5 +27,33 @@ def test_EOFS(self):
...
+            exec('x = 5\\')
+        self.assertEqual(str(excinfo.exception), expect)
+        with self.assertRaises(SyntaxError) as excinfo:
+            exec('\\')
+        self.assertEqual(str(excinfo.exception), expect)
+
+    @unittest.skipIf(not sys.executable, "sys.executable required")
+    def test_line_continuation_EOF_from_file_bpo2180(self):
+        """Ensure tok_nextc() does not add too many ending newlines."""
+        with tempfile.NamedTemporaryFile() as temp_f:
```
tirkarthi: `test.support.script_helper` can be used to simplify the tests and that will also help in Windows tests. Attached is a patch that passes on Windows.

```patch
diff --git a/Lib/test/test_eof.py b/Lib/test/test_eof.py
index b9328a4db9..d579d88ab6 100644
--- a/Lib/test/test_eof.py
+++ b/Lib/test/test_eof.py
@@ -4,6 +4,7 @@ import subprocess
 import sys
 import tempfile
 from test import support
+from test.support import script_helper
 import unittest

 class EOFTestCase(unittest.TestCase):
@@ -40,20 +41,14 @@ class EOFTestCase(unittest.TestCase):
     @unittest.skipIf(not sys.executable, "sys.executable required")
     def test_line_continuation_EOF_from_file_bpo2180(self):
         """Ensure tok_nextc() does not add too many ending newlines."""
-        with tempfile.NamedTemporaryFile() as temp_f:
-            temp_f.write(b'\\')
-            temp_f.flush()
-            proc = subprocess.run([sys.executable, temp_f.name],
-                                  capture_output=True)
-            self.assertIn(b'unexpected EOF while parsing', proc.stderr)
-            self.assertGreater(proc.returncode, 0)
-            temp_f.seek(0)
-            temp_f.write(b'y = 6\\')
-            temp_f.flush()
-            proc = subprocess.run([sys.executable, temp_f.name],
-                                  capture_output=True)
-            self.assertIn(b'unexpected EOF while parsing', proc.stderr)
-            self.assertGreater(proc.returncode, 0)
+        with support.temp_dir() as temp_dir:
+            file_name = support.script_helper.make_script(temp_dir, 'foo', '\\')
+            rc, out, err = support.script_helper.assert_python_failure(file_name)
+            self.assertIn(b'unexpected EOF while parsing', err)
+
+            file_name = support.script_helper.make_script(temp_dir, 'foo', 'y = 6\\')
+            rc, out, err = support.script_helper.assert_python_failure(file_name)
+            self.assertIn(b'unexpected EOF while parsing', err)

 if __name__ == "__main__":
     unittest.main()
```

tirkarthi: `import subprocess` can be removed if `script_helper` is used

gpshead: Thanks, this is nicer. I wasn't aware of script_helper. :)

PR #13402, opened 2019-05-18
bpo-36763: Remove _PyCoreConfig.dll_path

vstinner:


vstinner:
According to @zooba, there is no benefit to configure dll_path:

> > The dll_path field of PyConfig says "Windows only". Does that meant the 
> > struct doesn't have that field except in a Windows build? Or is it 
> > ignored, instead? If it doesn't have that field at all, what #define can 
> > be used to determine if the PyConfig struct will have it or not?
> 
> This field doesn't need to be here. It exists because it was used in 
> getpathp.c, and Victor's internal refactoring has kept it around through 
> all the field movement.

https://mail.python.org/pipermail/python-dev/2019-May/157451.html

I understood that it would allow to decide where Python looks for DLL. It seems like I misunderstood the purpose of this field. So let's remove it :-)

PR #13403, opened 2019-05-18
bpo-36951: Correct some types in the type_members struct in typeobject.c

ZackerySpytz:
https://bugs.python.org/issue36951


jdemeyer (approved):
This is obviously the right thing to do.

PR #13407, opened 2019-05-18
bpo-36950: Add run_until in test.support.script_helper

pierreglaser:
@vstinner

https://bugs.python.org/issue36950

### DIFF
--- a/Lib/test/support/script_helper.py
+++ b/Lib/test/support/script_helper.py
@@ -9,6 +9,7 @@ import os.path
 import subprocess
 import py_compile
 import zipfile
+import time
 
 from importlib.util import source_from_cache
 from test.support import make_legacy_pyc, strip_python_stderr
@@ -82,6 +83,21 @@ class _PythonRunResult(collections.namedtuple("_PythonRunResult",
                              % (self.rc, cmd_line,
                                 out,
                                 err))
+
+
+def run_until(condition, timeout=60, error_msg=None):
+    """Evaluates a condition until it becomes True, or a timeout is reached."""
+    deadline = time.monotonic() + timeout
+    t = 0.1
+    while time.monotonic() < deadline:
+        time.sleep(t)
+        t = min(t*2, 5)
+        if condition():
+            break
+    else:
+        msg = (error_msg or
+               f"The condition was not met after {timeout} seconds")
+        raise TimeoutError(msg)
 
 
 # Executing the interpreter in a subprocess
--- a/Misc/NEWS.d/next/Tests/2019-05-18-18-33-02.bpo-36950.f-K73w.rst
+++ b/Misc/NEWS.d/next/Tests/2019-05-18-18-33-02.bpo-36950.f-K73w.rst
@@ -1,0 +1,1 @@
+Add a new test helper (named run_until) to evaluate a condition repeatedly until it becomes true or a timeout is reached.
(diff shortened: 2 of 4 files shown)


Lib/test/support/script_helper.py
```diff
@@ -84,6 +85,21 @@ def fail(self, cmd_line):
                                 err))
 
 
+def run_until(condition, timeout=60, error_msg=None):
+    """Evaluates a condition until it becomes True, or a timeout is reached."""
+    deadline = time.monotonic() + timeout
+    t = 0.1
+    while time.monotonic() < deadline:
```
pierreglaser, line 92: For infinitesimal `timeout` values, the `while` loop will never be entered, and `condition` will not be evaluated even one single time. Not such a big deal, but maybe worth noting.

Lib/test/support/script_helper.py
```diff
@@ -84,6 +85,21 @@ def fail(self, cmd_line):
                                 err))
 
 
+def run_until(condition, timeout=60, error_msg=None):
```
tirkarthi, line 88: Could this be documented at https://docs.python.org/3/library/test.html#module-test.support.script_helper ?

Misc/NEWS.d/next/Tests/2019-05-18-18-33-02.bpo-36950.f-K73w.rst
```diff
@@ -0,0 +1 @@
+Add a new test helper (named run_until) to evaluate a condition repeatedly until it becomes true or a timeout is reached.
```
tirkarthi:
```suggestion
Add a new test helper :func:`test.support.script_helper.run_until` to evaluate a condition repeatedly until it becomes true or a timeout is reached.
```

tirkarthi (approved):
LGTM. Added a minor suggestion so that the function doc is hyperlinked in changelog.

Thanks for this utility.

pierreglaser:
Thanks for the review @tirkarthi!

Lib/test/support/script_helper.py
```diff
@@ -84,6 +85,21 @@ def fail(self, cmd_line):
...
+    """Evaluates a condition until it becomes True, or a timeout is reached."""
```
vstinner, line 89: I don't understand how I am supposed to use to ... run something.

pierreglaser: I'm sorry, I did not understand your comment. Can you re-phrase it?

vstinner: Example: https://bugs.python.org/issue36950#msg342751

Where do I put this code?

```
                try:
                    smm = shared_memory.SharedMemory(name, create=False)
                except FileNotFoundError:
                    break
```

pierreglaser: This is what is explained [here](https://bugs.python.org/issue36950#msg342804).
From what I saw in the `stdlib` tests, there are not so many cases that are that complex. 
Many tests use much simpler conditions (usually, they check the result of a function)

Example: in `_test_multiprocessing.py`:
```python
    def wait_proc_exit(self):
        # Only the manager process should be returned by active_children()
        # but this can take a bit on slow machines, so wait a few seconds
        # if there are other children too (see #17395).
        join_process(self.proc)
        start_time = time.monotonic()
        t = 0.01
        while len(multiprocessing.active_children()) > 1:
            time.sleep(t)
            t *= 2
            dt = time.monotonic() - start_time
            if dt >= 5.0:
                test.support.environment_altered = True
                print("Warning -- multiprocessing.Manager still has %s active "
                      "children after %s seconds"
                      % (multiprocessing.active_children(), dt),
                      file=sys.stderr)
                break
```

We could change this test into 
```python
def wait_proc_exit(self):
    join_process(self.proc)
    run_until(lambda: len(multiprocessing.active_children()) == 1, timeout=5)
```

You can see a exhaustive list of the tests that can be re-written using the current implementation of `run_until`  here: https://bugs.python.org/issue36950#msg342751. As I said, I did not find so many more complex test cases such as the one you are pointing out.

Knowing this, do we still want a more powerful `run_until`?

csabella:
@vstinner, should this one move forward?  Thanks!

vstinner:
I close the PR because of the lack of activity. The need for this function is not very clear at this point.

PR #13409, opened 2019-05-18
bpo-36953: Delay removal of ABCs from collections.

Carreau:
Bump the removal to 3.9, indicate collections.abc available since 3.3,
replace version-changed directive to deprecated-removed.

https://bugs.python.org/issue36953


Doc/library/collections.rst
```diff
@@ -33,10 +33,10 @@ Python's general purpose built-in containers, :class:`dict`, :class:`list`,
 :class:`UserString`     wrapper around string objects for easier string subclassing
 =====================   ====================================================================
 
-.. versionchanged:: 3.3
+.. deprecated-removed:: 3.3 3.9
     Moved :ref:`collections-abstract-base-classes` to the :mod:`collections.abc` module.
     For backwards compatibility, they continue to be visible in this module through
-    Python 3.7.  Subsequently, they will be removed entirely.
+    Python 3.7.
```
tirkarthi: Should this be changed to 3.8?

```suggestion
    Python 3.8.
```

Carreau: likely yes. I was unsure what this was trying to convey.

tirkarthi: My interpretation was that the import was still visible through `__getattr__` on module for 3.7 indicating 3.7 as last release to do so. But since 3.8 will continue to have that I suggested updating it to 3.8. I am not a native speaker though so this could be reverted if incorrect.

jdemeyer:
Why do we even have to say in advance when the functionality will be removed? For this deprecation, we already said we would remove it in 3.8 but we don't. So why are we now so certain that it will be removed in 3.9?

Carreau:
> Why do we even have to say in advance when the functionality will be removed? For this deprecation, we already said we would remove it in 3.8 but we don't. So why are we now so certain that it will be removed in 3.9?

The policy is AFAICT removal n+2 after deprecation. We do have to remove at some point no? And adding the version number for removal make it easy to search for all that was marked for removal, as well as give expectation to readers.

In any case here we are just discussing bumping the version number, which should be less controversial than discussing if their should be one.

asvetlov (approved):

ilevkivskyi:
I think this deserves a `Misc/NEWS` item.

Carreau:
> I think this deserves a `Misc/NEWS` item.

Why not... done.

Carreau:
Title and commits were mentioning wrong issues; rebased/edited.

ilevkivskyi (approved):
Thanks for updates, LGTM!

PR #13415, opened 2019-05-19
bpo-36949: Implement __repr__ on WeakSet

isidentical:
Implement __repr__ on WeakSet

https://bugs.python.org/issue36949


remilapeyre (approved):
Thanks!

zooba (approved):
Thanks!

PR #13421, opened 2019-05-19
bpo-36965: include windows.h to be able to use STATUS_CONTROL_C_EXIT

erikjanss:
According to the Microsoft documentation at

https://docs.microsoft.com/en-us/windows-hardware/drivers/kernel/using-ntstatus-values

system-supplied status codes are defined in ntstatus.h. and the NTSTATUS type is defined in ntdef.h

this PR includes both ntstatus.h and ntdef.h to be able to use STATUS_CONTROL_C_EXIT when compiling for windows.

https://bugs.python.org/issue36965

### DIFF
--- a/Modules/main.c
+++ b/Modules/main.c
@@ -18,8 +18,9 @@
 #if defined(HAVE_GETPID) && defined(HAVE_UNISTD_H)
 #  include <unistd.h>   /* getpid() */
 #endif
-#ifdef _MSC_VER
-#  include <crtdbg.h>   /* STATUS_CONTROL_C_EXIT */
+#ifdef MS_WINDOWS
+#  include <ntdef.h>      /* NTSTATUS */
+#  include <ntstatus.h>   /* STATUS_CONTROL_C_EXIT */
 #endif
 /* End of includes for exit_sigint() */
 
--- a/Modules/main.c
+++ b/Modules/main.c
@@ -19,8 +19,7 @@
 #  include <unistd.h>   /* getpid() */
 #endif
 #ifdef MS_WINDOWS
-#  include <winternl.h>      /* NTSTATUS */
-#  include <ntstatus.h>   /* STATUS_CONTROL_C_EXIT */
+#  include <windows.h>  /* STATUS_CONTROL_C_EXIT */
 #endif
 /* End of includes for exit_sigint() */
 
--- a/Modules/main.c
+++ b/Modules/main.c
@@ -19,7 +19,7 @@
 #  include <unistd.h>   /* getpid() */
 #endif
 #ifdef MS_WINDOWS
-#  include <ntdef.h>      /* NTSTATUS */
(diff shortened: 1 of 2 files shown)


Modules/main.c
```diff
@@ -18,8 +18,9 @@
 #if defined(HAVE_GETPID) && defined(HAVE_UNISTD_H)
 #  include <unistd.h>   /* getpid() */
 #endif
-#ifdef _MSC_VER
-#  include <crtdbg.h>   /* STATUS_CONTROL_C_EXIT */
+#ifdef MS_WINDOWS
+#  include <winternl.h>      /* NTSTATUS */
```
vstinner: I don't understand what this include is needed, the file doesn't use NTSTATUS.

erikjanss: ntstatus.h contains the value of STATUS_CONTROL_C_EXIT, and that value is of type NTSTATUS.

however ntstatus.h does not contain the definition of NTSTATUS and neither does it include that definition.  So, only including ntstatus.h will not compile.  I found the documentation linked above that specifies that one should include ntdef.h to have the definition of NTSTATUS.

but then again, that seems to have been deprecated, and I found comments in the headers that winternl.h should be used instead of ntdef.h

(I'm not much of a windows developer, so I only tried to parse the available documentation)

erikjanss:
include statements were changed as discussed in bpo-36965

zooba (approved):
I approve this. Since @gpshead was called in too I'll let him weigh in or merge, in case he has some special knowledge about this area.

vstinner:
Azure Pipelines PR: Windows PR Tests win64 failed with timeout:

```
534 0:11:08 load avg: 3.52 [420/421] test_multiprocessing_spawn passed (2 min 24 sec) -- running: test_import (1 min 31 sec)
535running: test_import (2 min 1 sec)
536running: test_import (2 min 31 sec)
(...)
620running: test_import (44 min 32 sec)
621running: test_import (45 min 2 sec)
622Terminate batch job (Y/N)? 
```

I close/reopen the PR to schedule a new job.

vstinner:
Hum, our bot creating backport is broken: https://github.com/python/miss-islington/issues/228

@erikjanss: Can you please try to create manually a PR to backport your fix?

vstinner:
> Sorry, @erikjanss and @vstinner, I could not cleanly backport this to 3.7 due to a conflict.
Please backport using cherry_picker on command line.
> cherry_picker 925af1d99b69bf3e229411022ad840c5a0cfdcf8 3.7

The bot first failed with a timeout, but then run correctly, except that there is a merge conflict.

@erikjanss: Agai, can you please backport manually the fix?

erikjanss:
@vstinner, I'll give the manual backport a try ...

erikjanss:
@vstinner, backport ready in GH-13471, what I did not evaluate though was if that include is needed at all, since STATUS_CONTROL_C_EXIT is not used in 3.7

PR #13428, opened 2019-05-19
gh-81148: Eliminate unnecessary check in _strptime when determining AM/PM

GPHemsley:
This also adds additional tests to improve the coverage in this area.

https://bugs.python.org/issue36967

* Issue: gh-81148


Lib/test/test_strptime.py
```diff
@@ -292,6 +292,10 @@ def test_hour(self):
         self.assertTrue(strp_output[3] == self.time_tuple[3],
                         "testing of '%%I %%p' directive failed; '%s' -> %s != %s" %
                          (strf_output, strp_output[3], self.time_tuple[3]))
+        am_date = (1999, 3, 17, 10, 44, 55, 2, 76, 0)
+        pm_date = (1999, 3, 17, 22, 44, 55, 2, 76, 0)
+        self.assertEqual(_strptime._strptime_time(time.strftime("%I %p", am_date), "%I %p")[3], 10)
```
pganssle: I'm not crazy about testing implementation details like `_strptime`, though I realize that that's what this `test_strptime` file does.

Personally, I think it might be worth migrating more of this stuff into `datetimetester.py` and just testing with `datetime.strftime`/`datetime.strptime` instead.

Lib/test/test_strptime.py
```diff
@@ -292,6 +292,10 @@ def test_hour(self):
...
+        self.assertEqual(_strptime._strptime_time(time.strftime("%I %p", pm_date), "%I %p")[3], 22)
```
pganssle: I'm not crazy about testing implementation details like `_strptime`, though I realize that that's what this `test_strptime` file does.

Personally, I think it might be worth migrating more of this stuff into `datetimetester.py` and just testing with `datetime.strftime`/`datetime.strptime` instead.

pganssle: Also, the special casing is happening at 0:XX and 12:XX, so you may want to add tests for that.

pganssle (changes requested):
Thank you for this contribution @GPHemsley, I think it's a definite improvement.

I've left two comments on the tests. I think adding the special case tests is important. Moving your ampm tests into `datetimetester` is more of a "nice to have" (I have not looked into how difficult it would be, but I'm pretty sure there are `strptime` tests in `datetimetester` already, so it's probably pretty easy).

Thanks!

Lib/test/test_strptime.py
GPHemsley: Because of the interaction of `_strptime.py` and `_datetimemodule.c` in the creation of the `datetime` modules, I think it's worthwhile to have separate tests for each of them. We may very well want to test this functionality in `datetimetester.py` as well, but I think we would want to deliberately duplicate the tests in that case (unless we go so far as to remove `test_strptime.py` altogether).

And I think my motivation for picking 10 was from turning 22 into AM, and 22 was likely chosen because it was used in `__calc_am_pm()`. But you raise a valid point about the noon and midnight special cases; I guess I may as well test all 24 hours.

GPHemsley:
@pganssle I have updated the tests. Let me know if that is sufficient.

pganssle:
> Because of the interaction of `_strptime.py` and `_datetimemodule.c` in the creation of the `datetime` modules, I think it's worthwhile to have separate tests for each of them. We may very well want to test this functionality in `datetimetester.py` as well, but I think we would want to deliberately duplicate the tests in that case (unless we go so far as to remove `test_strptime.py` altogether).

I disagree here. `_strptime.py` and `_datetimemodule.c` are both *implementation details* of the `datetime` module. I'm not sure why `_strptime.py` is being tested directly like this, but I don't think it's a good idea to keep adding direct tests for `_strptime.py`.

The behavior we want to test is the user-facing behavior, which is the one invoked by calling `datetime.strptime`. It doesn't really matter if that logic is in `_strptime` or in `datetime.strptime`, but it *does* matter that it works with [PEP 300](https://www.python.org/dev/peps/pep-0399/) and that it works with subclasses of `datetime`, both of which come "for free" when you add the test in `datetimetester.py`, but that I believe are not handled in `test_strptime.py`.

GPHemsley:
@pganssle I'm not sure if this is what you were asking me to do, but I've now added a similar test to `datetimetester.py` (which is actually missing a lot of tests covering the code in `_strptime.py`).

pganssle:
@GPHemsley Thank you for adding the tests in `datetimetester.py`, they look good to me. However, I think you should drop the tests from `test_strptime.py`, since this case is already getting proper (and better) coverage in `datetimetester.py`

I am aware that a lot of `strptime` is tested in `test_strptime.py` instead of `datetimetester.py`, but I think that is a historical accident and not something we should continue doing just for consistency's sake. Maybe @abalkin or @vstinner has a different opinion of this, but I think all new tests for `strptime` should be hitting the public interface of the module `datetime.strptime` *only*. Frankly, I think it would be a good idea to migrate a lot of the tests in `test_strptime.py` over to `datetimetester.py`, or at least to switch them over to using the same testing methodology (i.e. testing `datetime.strptime` rather than `_strptime` directly, testing with datetime subclasses, and running the tests with and without C extensions), but that's a different issue that can be discussed on its own ticket.

GPHemsley:
@pganssle Removed test from `test_strptime.py`. Also confirmed that `test_strptime.py` dates back to 2002, compared to `datetimetester.py`'s 2010.

Lib/test/datetimetester.py
```diff
@@ -2535,6 +2535,12 @@ def test_strptime(self):
         with self.assertRaises(ValueError): strptime("-2400", "%z")
         with self.assertRaises(ValueError): strptime("-000", "%z")
 
+    def test_strptime_ampm(self):
+        for hour in range(0, 24):
+            with self.subTest(hour=hour):
+                hour_date = (1999, 3, 17, hour, 44, 55, 2, 76, 0)
+                self.assertEqual(self.theclass.strptime(_time.strftime("%I %p", hour_date), "%I %p").hour, hour)
```
pganssle: This line is I think too long for PEP8. If I may suggest a refactoring for this function that I think makes it a bit cleaner:

```python
    def test_strptime_ampm(self):
        dt = datetime(1999, 3, 17, 0, 44, 55, 2, 76, 0)
        for hour in range(0, 24):
            with self.subTest(hour=hour):
                new_dt = dt.replace(hour=hour)
                dt_str = new_dt.strftime("%I %p")

                self.assertEqual(self.theclass.strptime(dt_str, "%I %p").hour,
                                 hour)
```

pganssle (changes requested):
@GPHemsley Yeah, I agree, I am at a conference this weekend, but this came up in another context today, so when I get back home (and have a bit of time to recover from my jet lag), I'm going to open an issue on bpo to propose a strategy going forward with respect to the datetimetester / test_strptime split. Things are mildly complicated by the fact that I didn't fully appreciate the fact that `_strptime` is in its own module because it's got some shared code between `time` and `datetime` (I grepped incorrectly for `import _strptime`, when in fact `timemodule.c` uses another mechanism to import it).

Anyway, developing a coherent strategy for how to organize the `strptime` tests is way outside the scope of this PR, so I don't want that to block your contribution, for which I am grateful.

I have left one final comment suggesting a refactoring of the test, but otherwise this looks good to me.

pganssle:
@vstinner @pablogsal Do you think this rises to the level of needing a NEWS entry? It's a pretty minor refactoring so I added "skip news", but if you think it would be better to add a changelog I will remove the label and add that to the requested changes.

Other than the one little requested change and possibly the news entry, I think this is ready to merge.

vstinner:
> @vstinner @pablogsal  Do you think this rises to the level of needing a NEWS entry? It's a  pretty minor refactoring so I added "skip news", (...)

If the change doesn't modify datetime behavior, there is no need to annoy users reading the Changelog. "skip news" is correct here.

Lib/test/datetimetester.py
GPHemsley: I likely opted to ignore line length, since much of the rest of the file has PEP8 issues, but I can change that.

Out of curiosity, why do you feel that using `dt.replace` is cleaner than generating a new datetime value directly each iteration?

pganssle: > Out of curiosity, why do you feel that using `dt.replace` is cleaner than generating a new datetime value directly each iteration?

Technically `replace` does generate a new value every iteration (`datetime` is immutable), but I get what you mean. I mainly thought it was more readable because:

1. I find the `time.strftime` interface to be a bit more "low level" and harder to read than an `strftime` call on a `datetime` object.
2. Using `replace(hour=hour)` makes it more obvious that we didn't have any sort of off-by-one error when constructing the time tuple - it would be hard to notice in a tuple if you were actually cycling through different values for the minute instead of hour.

AlexWaygood:
@pganssle, this old PR is marked as "awaiting changes", but I'm not sure it actually is.

Is this something you'd still be interested in reviewing, and potentially merging?

StanFromIreland:
@pganssle Is the only change left to do to the test? I don't the person who originally opened this pr has been active recently, I can finish this off :-)

Lib/test/datetimetester.py
```diff
@@ -2536,6 +2536,12 @@ def test_strptime(self):
...
```
StanFromIreland:
```suggestion
    def test_strptime_ampm(self):
        dt = datetime(1999, 3, 17, 0, 44, 55, 2, 76, 0)
        for hour in range(0, 24):
            with self.subTest(hour=hour):
                new_dt = dt.replace(hour=hour)
                dt_str = new_dt.strftime("%I %p")

                self.assertEqual(self.theclass.strptime(dt_str, "%I %p").hour,
                                 hour)
```
@pganssle 's suggestion

StanFromIreland:

Lib/test/datetimetester.py
```diff
@@ -2924,6 +2924,16 @@ def test_strptime(self):
         with self.assertRaises(ValueError): strptime("-000", "%z")
         with self.assertRaises(ValueError): strptime("z", "%z")
 
+    def test_strptime_ampm(self):
+        dt = datetime(1999, 3, 17, 0, 44, 55, 2, 76, 0)
```
StanFromIreland:
```suggestion
        dt = datetime(1999, 3, 17, 0, 44, 55, 2)
```

StanFromIreland (approved):
I updated and merged @pganssle's *Requested Changes* (with minor fixup).

LGTM

pganssle (approved):

PR #1067, opened 2017-04-09
bpo-29883: Add UDP support to Windows Proactor Event Loop

ameily:
This PR adds UDP support to the Windows Proactor Event Loop. Several changes were performed:

 * Added wrappers around Winsock UDP functions:
   * `WSAConnect` - ConnectEx doesn't support binding a UDP endpoint to a socket so I had to add this. WSAConnect completes immediately for UDP sockets.
   * `WSARecvFrom`
   * `WSASendTo`
 * Implement `_make_datagram_transport()` and the corresponding UDP transport class `_ProactorDatagramTransport`.
   * Bound UDP endpoints can call `WSARecv` and `WSASend`. Unbound UDP endpoints require calls to `WSARecvFrom` and `WSASendTo`.
 * Several new unit tests exercising UDP proactor support.

I've verified that the change can be cleanly applied to master, 3.5, and 3.6. So, back porting this feature will be straightforward, which I would like personally since I require this feature on a Python 3.5 project.

One of the SSL unit tests is failing and I've verified that the same test fails with the same error in `master`. 

```
cpython\lib\asyncio\sslproto.py:335: ResourceWarning: unclosed transport
<asyncio.sslproto._SSLProtocolTransport object at 0x00000171EABA8A90>
  source=self)
```

https://bugs.python.org/issue29883

PR #13440, opened 2019-05-20
bpo-29883: Asyncio proactor udp

asvetlov:
Follow-up for #1067

https://bugs.python.org/issue29883


asvetlov:
@vstinner would you review this PR? If you have no time please let me know anyway.

It is a continuation of #1067 with the following differences:
1. Support `scope_id` and `flowinfo` for IPv6 addresses 
2. Fix tp_traverse to process data structures for `TYPE_WRITE_TO` and `TYPE_READ_FROM` modes
3. Increase buffer size for `readfrom` from 4KiB to 256Kib (256 is used by selector-based event loops)
4. Support flow-control for data sending, now proactor implementation has the same functionality as selector-based one.
5. Convert mutable `bytearray` and `memoryview` on writing to immutable `bytes`. The selector-based implementation does the same already.
6. Add more tests.

I believe that the Pull Request is written well and ready for merging.

Lib/asyncio/proactor_events.py
```diff
@@ -425,6 +444,134 @@ def _pipe_closed(self, fut):
...
+
+    def get_write_buffer_size(self):
+        return sum(len(data) for data, _ in self._buffer)
+
+    def abort(self):
+        self._force_close(None)
+
+    def sendto(self, data, addr=None):
+        if not isinstance(data, (bytes, bytearray, memoryview)):
+            raise TypeError('data argument must be byte-ish (%r)',
```
tirkarthi:
```suggestion
            raise TypeError('data argument must be bytes-like object (%r)',
```

`byte-ish` could be rephrased as `bytes-like object` which is a more common term used in other places where same check is made.

```
$ rg -A 2 'not isinstance\(data, \(bytes, bytearray, memoryview\)'

Lib/asynchat.py
193:        if not isinstance(data, (bytes, bytearray, memoryview)):
194-            raise TypeError('data argument must be byte-ish (%r)',
195-                            type(data))

Lib/asyncio/proactor_events.py
307:        if not isinstance(data, (bytes, bytearray, memoryview)):
308-            raise TypeError(
309-                f"data argument must be a bytes-like object, "

Lib/asyncio/sslproto.py
382:        if not isinstance(data, (bytes, bytearray, memoryview)):
383-            raise TypeError(f"data: expecting a bytes-like instance, "
384-                            f"got {type(data).__name__}")

Lib/asyncio/selector_events.py
852:        if not isinstance(data, (bytes, bytearray, memoryview)):
853-            raise TypeError(f'data argument must be a bytes-like object, '
854-                            f'not {type(data).__name__!r}')
--
979:        if not isinstance(data, (bytes, bytearray, memoryview)):
980-            raise TypeError(f'data argument must be a bytes-like object, '
981-                            f'not {type(data).__name__!r}')
```

asvetlov: Good catch!
`byte-ish` was living in asyncio from the very beginning and for 5 year.

auvipy (approved):

Lib/asyncio/proactor_events.py
```diff
@@ -22,6 +23,24 @@
 from .log import logger
 
 
+def _set_socket_extra(transport, sock):
+    transport._extra['socket'] = sock
```
1st1: Please wrap into the new TransportSocket

Modules/overlapped.c
```diff
@@ -1362,15 +1499,213 @@ Overlapped_traverse(OverlappedObject *self, visitproc visit, void *arg)
...
+        return NULL;
+    }
+
+    Length = sizeof(AddressBuf);
+    Length = parse_address(AddressObj, Address, Length);
+    if (Length < 0) {
+        return NULL;
+    }
+
+    Py_BEGIN_ALLOW_THREADS
```
1st1, line 1550: If WSAConnect returns immediately, do we need to release GIL?

1st1 (approved):
Looks good. Merge with caution as I'm not a Windows expert.

Lib/asyncio/proactor_events.py
asvetlov: Done

Modules/overlapped.c
asvetlov: Technically it is a syscall.
We have the practice to release GIL on `socket.bind()` for example

auvipy (approved):

auvipy (approved):

asvetlov:
Thank you very much @ameily 
You did the most part of work!

PR #13441, opened 2019-05-20
bpo-26185: Fix repr() on empty ZipInfo object

BoboTiG:
It was failing on `AttributeError` due to inexistant but required attributes `file_size` and `compress_size`.
They are now initialized to `0` in `ZipInfo.__init__()`.

Note that this is a second patch: Matthew Zipay wrote a first one back in 2016 (see it on BPO).
I wanted propose something else less "magic". I do not mind closing this one if it is not the good patch to apply :)

https://bugs.python.org/issue26185

### DIFF
--- a/Lib/test/test_zipfile.py
+++ b/Lib/test/test_zipfile.py
@@ -1657,6 +1657,33 @@ class OtherTests(unittest.TestCase):
         self.assertRaises(ValueError,
                           zipfile.ZipInfo, 'seventies', (1979, 1, 1, 0, 0, 0))
 
+    def test_create_empty_zipinfo_repr(self):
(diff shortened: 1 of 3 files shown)


takluyver:
This isn't up to me, but I prefer the more explicit, less magical approach you've used here than the `__getattr__` trick in the original patch.

There's a `hasattr(zinfo, 'file_size')` elsewhere in the file which should be unnecessary with this change.

BoboTiG:
You are right, thank you :)
I simply removed the check to let all atributes being initialized to 0 and keep consistency.

Carreau (approved):

BoboTiG:
[EuroPython 2019] cc @zooba.

DinoV (approved):

PR #13443, opened 2019-05-20
bpo-36763: Fix encoding/locale tests in test_embed

vstinner:
* Fix encoding and locale tests in test_embed.InitConfigTests.
* InitConfigTests now only computes EXPECTED_CONFIG once.
* Add tests for PYTHONWARNINGS and PYTHONPATH env vars


vstinner:
This change should fix tests on FreeBSD.

PR #13444, opened 2019-05-20
bpo-36763: Fix C locale coercion

vstinner:
* Fix _Py_LegacyLocaleDetected(): don't attempt to coerce the
  C locale if LC_ALL environment variable is set. Add 'warn'
  parameter: emit_stderr_warning_for_legacy_locale() must check for
  the LC_ALL env var.
* _PyPreConfig_Write() sets coerce_c_locale to 0 if
  _Py_CoerceLegacyLocale() fails.

Altogether, _PyRuntime.preconfig.coerce_c_locale can now be used to
check if the C locale has been coerced.


ncoghlan:
@vstinner Makes sense, thanks.

PR #13445, opened 2019-05-20
gh-81087: Add str.dedent and cache it at compile time

remilapeyre:
* Issue: gh-81087


Lib/textwrap.py
```diff
@@ -426,40 +427,10 @@ def dedent(text):
...
-        for line in text.split("\n"):
-            assert not line or line.startswith(margin), \
-                   "line = %r, margin = %r" % (line, margin)
 
-    if margin:
-        text = re.sub(r'(?m)^' + margin, '', text)
-    return text
+    warnings.warn("textwrap.dedent is deprecated, use the str.dedent method instead",
+            PendingDeprecationWarning)
+    return text.dedent()
```
Carreau: Can we:
1) "is deprecated" -> is pending deprecation.
2) if/when we deprecate add in the deprecation message "Since Python X.y", it make it easier to manage expectations/ find to remove later.

Python/ast_opt.c
```diff
@@ -471,6 +471,39 @@ astfold_mod(mod_ty node_, PyArena *ctx_, int optimize_)
...
+    _Py_IDENTIFIER(dedent);
+    PyObject *dedent = PyUnicode_FromString("dedent");
+    if (!dedent) {
+        return 0;
+    }
+    if (asdl_seq_LEN(node_->v.Call.args) != 0 ||
+        asdl_seq_LEN(node_->v.Call.keywords) != 0 ||
+        node_->v.Call.func->kind != Attribute_kind) {
+        return 1;
+    }
```
Carreau, line 474: Likely dumb question, but isn't this block likely to be the one the triggers an early return most of the time?
Should it be the first in the function, allowing to not always have to create the `dedent` object ?

Lib/test/test_configparser.py
```diff
@@ -867,10 +866,10 @@ def test_invalid_multiline_value(self):
         if self.allow_no_value:
             self.skipTest('if no_value is allowed, ParsingError is not raised')
 
-        invalid = textwrap.dedent("""\
+        invalid = """\
             [DEFAULT]
             test {0} test
-            invalid""".format(self.delimiters[0])
+            invalid""".dedent().format(self.delimiters[0]
         )
```
gpshead: move this ) up

Lib/test/test_unicode.py
```diff
@@ -2436,6 +2436,124 @@ def test_free_after_iterating(self):
         support.check_free_after_iterating(self, iter, str)
         support.check_free_after_iterating(self, reversed, str)
 
+    def assertUnchanged(self, text):
```
gpshead: This would read better as `dedent_and_assert_unchanged`

Objects/unicodeobject.c
```diff
@@ -13885,6 +13886,153 @@ unicode_sizeof_impl(PyObject *self)
     return PyLong_FromSsize_t(size);
 }
 
+/*[clinic input]
+str.dedent as unicode_dedent
```
gpshead, line 14126: I think it would be easier to maintain if we kept the implementation of this method in Python.  I suggest just keeping it in the textwrap module.  Just rename `textwrap.dedent` to `_dedent` and turn `dedent` into the public wrapper that adds the PendingDeprecationWarning.  That benefits other Python implementations (PyPy) who can continue using it and is nicer than this literal translation of the Python code into CPython API calls.

This C API version becomes a simple import of textwrap, lookup of dedent, and returning the result of the call.

gpshead (changes requested):
missing: A documentation update documenting the new `str.dedent` method.

Lib/textwrap.py
```diff
@@ -426,40 +427,10 @@ def dedent(text):
...
-    if 0 and margin:
-        for line in text.split("\n"):
-            assert not line or line.startswith(margin), \
-                   "line = %r, margin = %r" % (line, margin)
 
-    if margin:
-        text = re.sub(r'(?m)^' + margin, '', text)
-    return text
+    warnings.warn("textwrap.dedent is deprecated, use the str.dedent method instead",
+            PendingDeprecationWarning)
```
tirkarthi: Please use `stacklevel=2` to make sure the relevant call site from user program is printed instead of this line in the traceback. Please align `PendingDeprecationWarning` with the brace.

Python/ast_opt.c
remilapeyre: This makes sense, thanks.

csabella:
@remilapeyre, if your changes are ready to be reviewed again, please add a PR-level comment as requested by @bedevere-bot.  Thanks!

remilapeyre:
@csabella, sorry, I have made the requested changes; please review again

csabella:
@remilapeyre, please resolve the merge conflicts.  Thank you!

remilapeyre:
Hi @csabella, it should be good now. There is a lot of PendingDeprecationWarnings in the tests. I'm not sure what I should do, should I fix all calls or let them as is for now?

csabella:

remilapeyre:
Hi @csabella and @gpshead, I fix the conflicts and updated the Pull Request. Can you review again?

gpshead:
One thing I think would help this PR is not updating so many `Lib/test/test_*.py` files to replace their existing textwrap.dedent usage.  It'd avoid the need for nearly so many conflict resolutions.

Changing those to use the new `.dedent()` method can be done in followups.

ammaraskar:
@remilapeyre Would you like to untangle the code changes here from updating existing usages of `textwrap.dedent` like Gregory suggested or do you mind if someone else bases off this work with a `Co-authored-by` tag with you?

hauntsaninja (changes requested):
Requesting changes as per above; we should separate the implementation of the feature from its use elsewhere in the codebase

Lib/idlelib/idle_test/htest.py
```diff
@@ -209,7 +208,7 @@ def _wrapper(parent):  # htest #
 _linenumbers_drag_scrolling_spec = {
     'file': 'sidebar',
     'kwds': {},
-    'msg': textwrap.dedent("""\
+    'msg': """\
```
terryjreedy, line 211: I don't like having '.dedent moved from here to the end 16 lines down.

In any case, I want both idlelib changes removed from this PR to avoid gratuitous differences between versions.

terryjreedy (changes requested):

StanFromIreland:
Closing this, it requires a PEP and is now quite a bit out of date.

PR #13450, opened 2019-05-20
bpo-36974: document PEP 590

jdemeyer:
**Work in progress**

CC @encukou @markshannon 


encukou:
I'll work on updating the docs today.
@jdemeyer I'll push changes your branch directly, if you don't mind.

encukou:
Rebased & updated

jdemeyer:
In `Doc/c-api/structures.rst`, I planned to give a high-level explanation of the vectorcall protocol (basically a short version of PEP 590). Right now this is not really explained. That's why I defined `vectorcallfunc` there.

Do you still think that this is a good idea?

encukou:
I don't think `c-api/structures` is a good place, since vectorcall is not a structure.
But perhaps a new chapter on calling under "Abstract Objects Layer" would make sense – calling takes up half of `c-api/object` already.

PR #10700, opened 2018-11-25
bpo-35310: Retry select() calls on EINTR even if deadline has passed

oranav:
select() calls are retried on EINTR (per PEP 475).  However, if a
timeout was provided and the deadline has passed after running signal
handlers, we should still retry select() with a non-blocking call -- to
make sure rlist, wlist and xlist are initialized appropriately.

https://bugs.python.org/issue35310

PR #13451, opened 2019-05-20
gh-81158: Make SharedMemoryManager release its resources if its parent process dies

pierreglaser:
https://bugs.python.org/issue36977

* Issue: gh-81158

### DIFF
--- a/Lib/multiprocessing/managers.py
+++ b/Lib/multiprocessing/managers.py
@@ -159,6 +159,10 @@ class Server(object):
         self.id_to_local_proxy_obj = {}
         self.mutex = threading.Lock()
 
+    def _track_parent(self):
+        process.parent_process().join()
+        self.stop_event.set()
+
     def serve_forever(self):
         '''
         Run the server forever
@@ -167,8 +171,10 @@ class Server(object):
         process.current_process()._manager_server = self
         try:
             accepter = threading.Thread(target=self.accepter)
-            accepter.daemon = True
+            watcher = threading.Thread(target=self._track_parent)
+            accepter.daemon = watcher.daemon = True
             accepter.start()
+            watcher.start()
             try:
                 while not self.stop_event.is_set():
                     self.stop_event.wait(1)
--- a/Lib/multiprocessing/managers.py
+++ b/Lib/multiprocessing/managers.py
@@ -160,8 +160,9 @@ class Server(object):
         self.mutex = threading.Lock()
 
     def _track_parent(self):
-        process.parent_process().join()
-        self.stop_event.set()
+        if parent_process := process.parent_process():
+            parent_process.join()
+            self.stop_event.set()
 
     def serve_forever(self):
         '''
--- a/Lib/multiprocessing/managers.py
+++ b/Lib/multiprocessing/managers.py
@@ -160,8 +160,8 @@ class Server(object):
         self.mutex = threading.Lock()
 
     def _track_parent(self):
-        if parent_process := process.parent_process():
-            parent_process.join()
+        if process.parent_process() is not None:
+            process.parent_process().join()
             self.stop_event.set()
 
     def serve_forever(self):
@@ -172,10 +172,14 @@ class Server(object):
         process.current_process()._manager_server = self
         try:
             accepter = threading.Thread(target=self.accepter)
-            watcher = threading.Thread(target=self._track_parent)
-            accepter.daemon = watcher.daemon = True
+            accepter.daemon = True
             accepter.start()
-            watcher.start()
+
+            if process.parent_process() is not None:
+                watcher = threading.Thread(target=self._track_parent)
+                watcher.daemon = True
+                watcher.start()
+
             try:
                 while not self.stop_event.is_set():
                     self.stop_event.wait(1)
--- a/Lib/test/_test_multiprocessing.py
+++ b/Lib/test/_test_multiprocessing.py
@@ -3852,7 +3852,8 @@ class _TestSharedMemory(BaseTestCase):
         t = 0.1
         while time.monotonic() < deadline:
             time.sleep(t)
-            t = min(t*2, 5)
+            # The shared_memory segment should be released by the
+            # SharedMemoryManager server process before it shuts down.
             try:
                 sl = shared_memory.SharedMemory(shm_name, create=False)
             except FileNotFoundError:
--- a/Lib/test/_test_multiprocessing.py
+++ b/Lib/test/_test_multiprocessing.py
@@ -3824,6 +3824,44 @@ class _TestSharedMemory(BaseTestCase):
 
         smm.shutdown()
 
+    def test_shared_memory_SharedMemoryManager_cleanup_if_parent_dies(self):
+        # If a process that spawned a SharedMemoryManager was terminated, its
+        # manager process should notice it, unlink the resources the manager
+        # delivered, and shut down.
+        cmd = '''if 1:
+            import sys, time
+
+            from multiprocessing.managers import SharedMemoryManager
+
+
+            smm = SharedMemoryManager()
+            smm.start()
+            sl = smm.ShareableList([1, 2, 3])
+
+            sys.stdout.write(f'{sl.shm.name}\\n')
+            sys.stdout.flush()
+            time.sleep(100)
+        '''
+
+        p = subprocess.Popen([sys.executable, '-E', '-c', cmd],
+                             stdout=subprocess.PIPE, stderr=subprocess.PIPE)
+        shm_name = p.stdout.readline().decode().strip()
+        p.terminate()
+        start_time = time.monotonic()
+        deadline = start_time + 60
+        t = 0.1
+        while time.monotonic() < deadline:
+            time.sleep(t)
+            t = min(t*2, 5)
+            try:
+                sl = shared_memory.SharedMemory(shm_name, create=False)
+            except FileNotFoundError:
+                break
+        else:
+            raise AssertionError(
+                "A SharedMemoryManager prevented some resources to be cleaned "
+                "up after its parent process was terminated")
+
     @unittest.skipIf(os.name != "posix", "resource_tracker is posix only")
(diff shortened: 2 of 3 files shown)


pierreglaser:
@ogrisel

Lib/test/_test_multiprocessing.py
```diff
@@ -3824,6 +3824,46 @@ def test_shared_memory_SharedMemoryServer_ignores_sigint(self):
...
 
+    def test_shared_memory_SharedMemoryManager_cleanup_if_parent_dies(self):
+        # If a process that spawned a SharedMemoryManager was terminated, its
+        # manager process should notice it, unlink the resources the manager
+        # delivered, and shut down.
+        cmd = '''if 1:
+            import os, sys, time
+            import multiprocessing as mp
+            from multiprocessing import Process
+
```
eamanu, line 3833: Why don't remove this line?

eamanu (commented):
Hello, IMO you should test the if your changes. I think that you are not test it, isn't?

Are you making this changes for the test? I think if you want to say anything about (in this
example) process child are alive yet, you need to say it on the code (no in the test). 
Because, anyone will know that the process are alive.

Lib/test/_test_multiprocessing.py
```diff
@@ -3824,6 +3824,44 @@ def test_shared_memory_SharedMemoryServer_ignores_sigint(self):
...
+        p = subprocess.Popen([sys.executable, '-E', '-c', cmd],
+                             stdout=subprocess.PIPE, stderr=subprocess.PIPE)
+        shm_name = p.stdout.readline().decode().strip()
+        p.terminate()
+        start_time = time.monotonic()
+        deadline = start_time + 60
+        t = 0.1
+        while time.monotonic() < deadline:
+            time.sleep(t)
```
pablogsal, line 3854: I don't feel comfortable having tests that rely on sleep, they are almost assured to fail in some of the buildbots
 (check https://github.com/python/cpython/pull/10700 for example). Could you restructure this
tests to use synchronization primitives?

pablogsal (dismissed):

Lib/test/_test_multiprocessing.py
pierreglaser: I understand your worries. I am having trouble thinking of a test that does not require a timeout though (else, we risk hanging forever waiting for the synchronization event to occur). Inherently, this tests a feature that relies on `connection.wait`, which can block for a non-deterministic  amount of time.

Note that using this pattern:
```python
while time.monotonic() < deadline:
    time.sleep(t)
    if condition():
        break
else:
    raise AssertionError
```

is already more robust than a simple 
```python
time.sleep(dt)
self.assertTrue(condition())
```

It was discussed here: https://bugs.python.org/issue36867, and @vstinner seemed OK with its usage. I may be missing something though.

Lib/test/_test_multiprocessing.py
```diff
@@ -3824,6 +3824,44 @@ def test_shared_memory_SharedMemoryServer_ignores_sigint(self):
...
+            t = min(t*2, 5)
+            try:
+                sl = shared_memory.SharedMemory(shm_name, create=False)
+            except FileNotFoundError:
+                break
```
ogrisel, line 3860: Maybe add a comment to make it explicit that `shm_name` has been successfully collected by the `SharedMemoryManager` server before shutting down.

Lib/test/_test_multiprocessing.py
pablogsal: `multiprocessing` remains the module with the higher amount of random failures due to sleeps/race conditions. I will dismiss my review in case there is no other way, but I still uneasy about adding more `sleep()` :(

pierreglaser:
Maybe @pitrou wants to have a look at this.

Lib/multiprocessing/managers.py
```diff
@@ -159,6 +159,10 @@ def __init__(self, registry, address, authkey, serializer):
         self.id_to_local_proxy_obj = {}
         self.mutex = threading.Lock()
 
+    def _track_parent(self):
+        process.parent_process().join()
```
pitrou: What if `process.parent_process()` returns None? For example if the Server was launched in an independent Python instance?

Lib/test/_test_multiprocessing.py
```diff
@@ -3824,6 +3824,44 @@ def test_shared_memory_SharedMemoryServer_ignores_sigint(self):
...
+            sys.stdout.flush()
+            time.sleep(100)
+        '''
+
+        p = subprocess.Popen([sys.executable, '-E', '-c', cmd],
+                             stdout=subprocess.PIPE, stderr=subprocess.PIPE)
+        shm_name = p.stdout.readline().decode().strip()
+        p.terminate()
+        start_time = time.monotonic()
+        deadline = start_time + 60
```
pitrou, line 3851: 10 seconds sounds enough, no?

Lib/test/_test_multiprocessing.py
```diff
@@ -3824,6 +3824,44 @@ def test_shared_memory_SharedMemoryServer_ignores_sigint(self):
...
+        t = 0.1
+        while time.monotonic() < deadline:
+            time.sleep(t)
+            t = min(t*2, 5)
```
pitrou: There is no point doing this.

pitrou (commented):
Some comments below. It would also be nice if @applio could take a look and say if it looks fine on the principle.

Lib/test/_test_multiprocessing.py
ogrisel: In case of overloaded CI workers one can see weird things on rare occasions.

99.9% of the time it will probably take less than 10ms (although I have not checked), this limit is just a timeout, so I think it's fine to keep a large value to make it extremely unlikely to get false positive failures on CI servers.

pitrou: These are seconds here not milliseconds.

pitrou: But, no strong feelings either.

Lib/multiprocessing/managers.py
pierreglaser: Did not know this was a supported use-case. But we can add a guard.

pierreglaser: However, adding a guard will simply discard the case where `parent_process() is None`, which is not the best solution. Yet, I feel that tracking the usage of manager-allocated resources by other processes than the manager's direct parent is tricky, especially for `SharedMemoryManager` (see more detailed explanation here https://github.com/python/cpython/pull/13247#issuecomment-493694766).

Do we want to make this feature more robust by trying to make the manager wait on any process that tries to connect to it?

Lib/multiprocessing/managers.py
```diff
@@ -167,8 +172,10 @@ def serve_forever(self):
         process.current_process()._manager_server = self
         try:
             accepter = threading.Thread(target=self.accepter)
-            accepter.daemon = True
+            watcher = threading.Thread(target=self._track_parent)
```
pitrou: Can simply launch the watcher only if `parent_process() is not None`.

Lib/multiprocessing/managers.py
```diff
@@ -159,6 +159,11 @@ def __init__(self, registry, address, authkey, serializer):
         self.id_to_local_proxy_obj = {}
         self.mutex = threading.Lock()
 
+    def _track_parent(self):
+        if parent_process := process.parent_process():
```
pitrou: Please use decent syntax...

Lib/multiprocessing/managers.py
pitrou: I'm afraid I'm not following you here. We're simply trying to fix the case where the manager was launched by its parent process, no? (that's what the issue title says)

The case where the manager is autonomous looks pretty much unfixable on its own.

Lib/multiprocessing/managers.py
pierreglaser: I did not know PEP 572 assignment expressions were out of line for new code?

pitrou: Sorry. I just don't like them :-)

Lib/multiprocessing/managers.py
pierreglaser: > We're simply trying to fix the case where the manager was launched by its parent process

Ok, lets stick to this in this PR.

> The case where the manager is autonomous looks pretty much unfixable on its own.

I agree, but there are in-betweens: if a manager is created by a process (its parent), and then connected to by other processes, then should the manager just watch its parent? Technically, if other processes are also connected to it, it probably should not shut down unless ALL processes connected to it (and not only its parent) terminated.

Lib/multiprocessing/managers.py
pierreglaser: I added a new guard, but left the one in `_track_parent` (does not hurt)

Lib/multiprocessing/managers.py
pierreglaser: Ok. No strong opinion on my end, so I removed it.

Lib/multiprocessing/managers.py
pitrou: Hmm... So, conversely, if the manager loses all its connections _and_ was created by a parent process, then it should die?

Perhaps that should be the right heuristic?

In other words: `SharedMemoryServer.handle_request` can easily keep track of the number of connected clients. At the end, it decrements that number and, if it falls to 0 _and_ there is a parent process, then die?

pierreglaser: Yes, I think this is already a more robust solution. However, as opposed to with `SyncManager`, manipulating/modifying `SharedMemoryManager` delivered memory segments does not necessarily trigger communication with the manager server process. This is because `SharedMemoryManager` delivers `SharedMemory` objects, and not proxies. Thus, keeping track of the number of connected clients does not equate keep track of the processes manipulating those objects. But maybe this approach is still good enough for now.

Also, the recent `os.memfd_create` mentioned in [this python issue](https://bugs.python.org/issue26836) looks interesting, but I need to look more into it.

PR #13452, opened 2019-05-20
bpo-36969: Make PDB args command display keyword only arguments

remilapeyre:
https://bugs.python.org/issue36969

### DIFF
--- a/Lib/pdb.py
+++ b/Lib/pdb.py
@@ -1132,9 +1132,9 @@ class Pdb(bdb.Bdb, cmd.Cmd):
         """
         co = self.curframe.f_code
         dict = self.curframe_locals
-        n = co.co_argcount
-        if co.co_flags & 4: n = n+1
-        if co.co_flags & 8: n = n+1
+        n = co.co_argcount + co.co_kwonlyargcount
+        if co.co_flags & inspect.CO_VARARGS: n = n+1
+        if co.co_flags & inspect.CO_VARKEYWORDS: n = n+1
         for i in range(n):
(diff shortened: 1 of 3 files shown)


Lib/pdb.py
```diff
@@ -1132,9 +1132,9 @@ def do_args(self, arg):
         """
         co = self.curframe.f_code
         dict = self.curframe_locals
-        n = co.co_argcount
-        if co.co_flags & 4: n = n+1
-        if co.co_flags & 8: n = n+1
+        n = co.co_argcount + co.co_kwonlyargcount
```
pablogsal, line 1135: You also need to add co.co_posonlyargcount to handle PEP570 (check the issue). Although this will difficult backporting, so maybe you want this to be done in a separate PR.

remilapeyre: Thank @pablogsal, it may be simple to do so in a separate PR I think.

Lib/pdb.py
```diff
@@ -1132,9 +1132,9 @@ def do_args(self, arg):
         """
         co = self.curframe.f_code
         dict = self.curframe_locals
-        n = co.co_argcount
-        if co.co_flags & 4: n = n+1
```
pablogsal, line 1136: Thanks for removing these evil magic numbers!

pablogsal (approved):

Lib/pdb.py
remilapeyre: Definitely the improvement Python needed ^^

PR #13454, opened 2019-05-20
bpo-36511: Fix Windows arm32 buildbot pythoninfo, scp, and ssh

paulmon:
When running the buildbot for Windows ARM32 test.pythoninfo must be run remotely.
Also seperating deploy step from both pythoninfo and test.bat

Requires buildmaster config change to work: 
https://github.com/python/buildmaster-config/pull/91

@zooba @zware


paulmon:
The debug part of the test pass should pass locally with these fixes, except for test_datetime.
See this PR for test_datetime: https://github.com/python/cpython/pull/13073
I am re-running test.bat now with these fixes, which takes more than an hour.  
The last pass took 4 hours and 23 minutes on my secondary ARM test device, but on the buildbot ARM test device it's been taking about an hour and a half.
I still need to try the WindowsArm32ReleaseBuild steps, but expect the same results.

Lib/test/test_ssl.py
```diff
@@ -1345,6 +1346,7 @@ def test_load_verify_cadata(self):
             ctx.load_verify_locations(cadata=b"broken")
 
 
+    @unittest.skipIf(Py_DEBUG, "Crashes on debug python builds")
```
zooba: This is certainly not a general issue - can we scope it down to the platform(s) where it crashes?

zooba (commented):
Would prefer a tighter skip on the tests, and deferring to @zware on the buildbot scripts (as with the build config change).

Lib/test/test_ssl.py
paulmon: I will scope it down.  

I think it's limited to UCRT.  The problem is that SSL has  FILE* from the ucrtbase.dll table and python has a FILE* from the ucrtbase**d**.dll table.  

I think we could limit it to win32+arm or platform.win32_is_iot( ).  Unless there's a way to check if python is using UCRT? I couldn't find one.

zooba: Oh, it's because of mixing FILE* across debug/release builds? In that case we scope it down to Windows and enhance the skip message to say "Avoid mixing debug/release CRT on Windows"

paulmon: Done.

paulmon:
Enabling is_python_build( ) to return true on the test target uncovered more failing tests.
There is still one failure in test_pyexpat that I will be investigating next.

```
FAIL: test_exception (test.test_pyexpat.HandlerExceptionTest)
  File "C:\py\lib\test\test_pyexpat.py", line 453, in test_exception
    parser.Parse(b"<a><b><c/></b></a>", 1)
RuntimeError: a

  File "C:\py\lib\test\test_pyexpat.py", line 469, in test_exception
    self.assertIn('call_with_frame("StartElement"', entries[1][3])
AssertionError: 'call_with_frame("StartElement"' not found in ''

```

paulmon:
I simplified the deployment script to copy entire directories instead of picking files, except for PCBuild and Modules which have many files that aren't needed for the test run.  This should make the script more resilient to test changes.  

I think the Windows ARM32 test pass is green now when I run the scripts manually, if https://github.com/python/cpython/pull/13073 is included.

Lib/test/test_pyexpat.py
```diff
@@ -465,7 +466,7 @@ def test_exception(self):
                                        "pyexpat.c", "StartElement")
             self.check_traceback_entry(entries[2],
                                        "test_pyexpat.py", "StartElementHandler")
-            if sysconfig.is_python_build():
+            if sysconfig.is_python_build() and not (sys.platform == 'win32' and platform.machine() == 'ARM'):
```
paulmon, line 469: I'd rather fix this than skip it, but I haven't figured out where to start looking yet.

Lib/test/test_ssl.py
```diff
@@ -28,6 +28,9 @@
 
 from ssl import TLSVersion, _TLSContentType, _TLSMessageType, _TLSAlertType
 
+Py_DEBUG = hasattr(sys, 'gettotalrefcount')
+Py_DEBUG_Win32 = Py_DEBUG and sys.platform=='win32'
```
zooba:
```suggestion
Py_DEBUG_WIN32 = Py_DEBUG and sys.platform == 'win32'
```

I'll let you update the references to this variable.

Lib/test/test_math.py
```diff
@@ -468,6 +468,8 @@ def testCos(self):
             self.assertRaises(ValueError, math.cos, NINF)
         self.assertTrue(math.isnan(math.cos(NAN)))
 
+    @unittest.skipIf(sys.platform=='win32' and (platform.machine()=='ARM' or platform.machine()=='ARM64'),
```
zooba:
```suggestion
    @unittest.skipIf(sys.platform == 'win32' and platform.machine() in ('ARM', 'ARM64')),
```

paulmon:
@tiran can you please review?
Some of these changes are required to enable the Windows arm32 buildbot to start working.
Thanks!

zooba:
@paulmon I'm happy to merge without Christian's review once the updates I suggested in my last one are done.

paulmon:
@zooba I merged your suggested changes.  I think it's ready

zooba:
@paulmon The `Py_DEBUG_WIN32` change needs the references to that variable updated as well. I didn't go through and suggest a change on each one.

paulmon:
> @paulmon The `Py_DEBUG_WIN32` change needs the references to that variable updated as well. I didn't go through and suggest a change on each one.

I'm not sure I'm following you.  Are suggesting adding a check for arm/arm64 to Py_DEBUG_WIN32?  That would be incorrect for Windows IoT Core x86/x64 which also use UCRT for everything.

zooba:
> @paulmon The Py_DEBUG_WIN32 change needs the references to that variable updated as well. I didn't go through and suggest a change on each one.
> 
> I'm not sure I'm following you.  Are suggesting adding a check for arm/arm64 to Py_DEBUG_WIN32?  That would be incorrect for Windows IoT Core x86/x64 which also use UCRT for everything.

No, I'm saying elsewhere you are referring to `Py_DEBUG_Win32` which is now undefined. They need to be changed to `Py_DEBUG_WIN32` as well

paulmon:
> No, I'm saying elsewhere you are referring to `Py_DEBUG_Win32` which is now undefined. They need to be changed to `Py_DEBUG_WIN32` as well

Thanks.  I noticed the whitespace change but somehow didn't see the uppercase change.

paulmon:
@zooba Thanks for the help
I fixed test_ssl and test_math so they pass on my dev machine.  
All of the PR checks pass now.

zooba:
@paulmon Seems like it still fails: https://buildbot.python.org/all/#/builders/197/builds/338

paulmon:
> @paulmon Seems like it still fails: https://buildbot.python.org/all/#/builders/197/builds/338

The raspberry pi was in a bad state.  I have restarted it

paulmon:
Also, ssh.exe and scp.exe somehow don't exist when running under the buildbot, which is breaking the script.

zooba:
@paulmon Different PATH setting? Or perhaps it's running with different file redirection settings (as a 32-bit app, maybe)?

paulmon:
> @paulmon Different PATH setting? Or perhaps it's running with different file redirection settings (as a 32-bit app, maybe)?

The PATH is echoed at the top of the log and includes the PATH to ssh.  
I hadn't thought WoW as a possible cause.  I installed from the link on python.org so I got the 32-bit version of python.  I'll update to the 64-bit python and see if that helps.

PR #13457, opened 2019-05-21
bpo-36984: Improve version added references in `typing` module docs

asottile:
https://bugs.python.org/issue36984

### DIFF
--- a/Doc/library/typing.rst
+++ b/Doc/library/typing.rst
@@ -555,7 +555,7 @@ The module defines the following classes, functions and decorators:
 
    A generic version of :class:`collections.abc.Collection`
 
-   .. versionadded:: 3.6
+   .. versionadded:: 3.6.0
 
 .. class:: AbstractSet(Sized, Collection[T_co])
 
@@ -599,6 +599,7 @@ The module defines the following classes, functions and decorators:
 
    A generic version of :class:`collections.deque`.
 
+   .. versionadded:: 3.5.4
    .. versionadded:: 3.6.1
 
 .. class:: List(list, MutableSequence[T])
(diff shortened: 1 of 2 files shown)


Doc/library/typing.rst
```diff
@@ -599,6 +599,7 @@ The module defines the following classes, functions and decorators:
 
    A generic version of :class:`collections.deque`.
 
+   .. versionadded:: 3.5.4
    .. versionadded:: 3.6.1
```
Carreau, line 603: Why is there 2 "versionadded" here ? was this added both ein 3.5.4 and 3.6.1 ? Meaning it's not in 3.6.0 ?
If that's the case that might look like a typo in the docs from a casual reader, and might need a  clarification paragraph.

Doc/library/typing.rst
```diff
@@ -708,12 +719,14 @@ The module defines the following classes, functions and decorators:
 
    A generic version of :class:`collections.Counter`.
 
+   .. versionadded:: 3.5.4
    .. versionadded:: 3.6.1
```
Carreau, line 723: Same on all of these; that feel super weird;

Carreau (approved):
Version added are always super useful.

Doc/library/typing.rst
asottile: that's correct! it _**is**_ pretty weird!

ilevkivskyi: Yeah, the story with versions is related to the fact that `typing` module used to be provisional, so features were added at random time, even between minor releases.

ilevkivskyi (approved):
Thanks, LGTM!

PR #11993, opened 2019-02-22
bpo-36084: add native thread ID (TID) to threading.Thread objects

jaketesler:
This functionality adds a native Thread ID to threading.Thread objects. This ID (`TID`), similar to the `PID` of a process, is assigned by the OS (kernel) and is generally used for externally monitoring resources consumed by the running thread (or process).
This does not replace the `ident` attribute within Thread objects, which is assigned by the Python interpreter and is guaranteed as unique for the lifetime of the Python instance.

This new functionality fully supports Linux, Apple, and Windows platforms.

https://bugs.python.org/issue36084

PR #13458, opened 2019-05-21
Revert "bpo-36084: Add native thread ID to threading.Thread objects (GH-11993)"

vstinner:
This reverts commit 4959c33d2555b89b494c678d99be81a65ee864b0.

### DIFF
--- a/Doc/library/_thread.rst
+++ b/Doc/library/_thread.rst
@@ -85,18 +85,6 @@ This module defines the following constants and functions:
    may be recycled when a thread exits and another thread is created.
 
 
-.. function:: get_native_id()
-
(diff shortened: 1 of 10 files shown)


vstinner:
The commit 4959c33d2555b89b494c678d99be81a65ee864b0 broke the AIX buildbot for longer than week:
https://bugs.python.org/issue36084#msg342263

The CI for the policy is to revert a change which breaks the CI and no one is able to come up with a fix in 1 day:
https://pythondev.readthedocs.io/ci.html#revert-on-fail

So I revert the change.

It doesn't mean that the change is wrong and will never land in Python. It is just a way to repair the CI and give more time to find the right design which fits all platforms running on the CI.

cc @jaketesler @pitrou

PR #13460, opened 2019-05-21
bpo-36974: remove _PyObject_HasFastCall

jdemeyer:
**After implementing PEP 590** (#13185)


encukou (approved):
Looks good!
(Not merging now as I won't be able to watch buildbots)

PR #13461, opened 2019-05-21
bpo-36994: add test for profiling method_descriptor with **kwargs

jdemeyer:
https://bugs.python.org/issue36994


rbtcollins (approved):
LGTM

PR #13462, opened 2019-05-21
bpo-36797: Reduce levels of indirection in outdated distutils docs

ncoghlan:
https://bugs.python.org/issue36797


jaraco (approved):
Looks good to me.

jaraco:
Is there any value in leaving an orphaned 'packageindex.rst', given that we can always reference it at a permalink prior to acceptance of this change?

PR #13463, opened 2019-05-21
bpo-36084: add native thread ID (TID) to threading.Thread objects (V2)

jaketesler:
This is the second version of PR #11993 (original reverted due to test failures on unsupported operating systems).

cc @vstinner @pitrou 

From the original PR:
>This functionality adds a native Thread ID to threading.Thread objects. This ID (TID), similar to the PID of a process, is assigned by the OS (kernel) and is generally used for externally monitoring resources consumed by the running thread (or process).
This does not replace the ident attribute within Thread objects, which is assigned by the Python interpreter and is guaranteed as unique for the lifetime of the Python instance.
>
>This new functionality fully supports Linux, Apple, and Windows platforms.


Lib/_dummy_thread.py
```diff
@@ -71,6 +71,10 @@ def get_ident():
     """
     return 1
 
+def get_native_id():
+    """Dummy implementation of _thread.get_native_id()."""
+    return None
```
vstinner: None is not a non-negative integer. It contradicts the doc, no?

Lib/test/test_threading.py
```diff
@@ -104,6 +104,11 @@ def test_various_ops(self):
             self.assertRegex(repr(t), r'^<TestThread\(.*, initial\)>$')
             t.start()
 
+        if hasattr(_thread, 'get_native_id'):
```
vstinner: Can you test threading.get_native_id() instead to avoid relying on private module?

Lib/threading.py
```diff
@@ -34,6 +34,10 @@
 _allocate_lock = _thread.allocate_lock
 _set_sentinel = _thread._set_sentinel
 get_ident = _thread.get_ident
+try:
+    get_native_id = _thread.get_native_id
+except AttributeError:
+    get_native_id = lambda: None
```
vstinner: Hum no, I would prefer to not define the function if it is not supported. Returning None contradicts the doc.

Lib/threading.py
```diff
@@ -891,6 +896,9 @@ def _bootstrap(self):
     def _set_ident(self):
         self._ident = get_ident()
 
+    def _set_native_id(self):
```
vstinner: I would prefer to make the attribute optionale as well

Misc/NEWS.d/next/Core and Builtins/2019-02-22-23-03-20.bpo-36084.86Eh4X.rst
```diff
@@ -0,0 +1 @@
+Add native thread ID (TID) to threading.Thread objects
```
vstinner: Maybe mention on which platform it is supported.

Python/thread_pthread.h
```diff
@@ -302,6 +308,29 @@ PyThread_get_thread_ident(void)
...
+#ifdef __APPLE__
+    uint64_t native_id;
+    pthread_threadid_np(NULL, &native_id);
+#elif defined(__linux__)
+    pid_t native_id;
+    native_id = syscall(__NR_gettid);
+#elif defined(__FreeBSD__)
+    pid_t native_id;
+    native_id = pthread_getthreadid_np();
+#else
```
vstinner: The #else block should not exist.

Python/thread_pthread.h
```diff
@@ -302,6 +308,29 @@ PyThread_get_thread_ident(void)
...
+    unsigned long native_id;
+    native_id = 0;
+#endif
+    return (unsigned long) native_id;
```
vstinner, line 327: Hum,  this downcast leads to integer overflow on macOS which uses uint64_t if unsigned long is 32-bit, no?

What about using uint64_t type for the return type?

vstinner (changes requested):

Lib/_dummy_thread.py
jaketesler: The doc will be updated – this change reflects your suggestion from the previous PR (to have the value be `None` if not started, matching the behavior of`ident`)

Lib/test/test_threading.py
jaketesler: @vstinner The goal here is to determine whether the function is defined in C. `threading.get_native_id()` is always available in the Python module but will return `None` if the capability is not supported by the OS. 

If you think `threading.get_native_id()` shouldn't even exist on the Python side if the functionality isn't available, can you suggest a mechanism to remove the Python function only on unsupported platforms, or alternatively, a mechanism to avoid defining it altogether (on those unsupported platforms only)?

Lib/threading.py
jaketesler: @vstinner Re: my comment above, when you say "not define the function" do you mean simply declare `get_native_id = None`, or remove it altogether from the module (via `del` or some similar mechanism)?

Python/thread_pthread.h
jaketesler: From the last PR: https://github.com/python/cpython/pull/11993#discussion_r283107275

>On Apple platforms, the `pthread_threadid_np()` function wasn’t introduced until Mac OS 10.5, after Mac OS primarily became 64-bit. I figured it was acceptable to have the 64-bit type here since Python 3 for Mac only builds 64-bit versions and therefore the only Mac OS versions capable of supporting this feature would have to be those that also support Python 3 and therefore be 64-bit.

Lib/_dummy_thread.py
jaketesler: @vstinner should `dummy_thread` be updated to remove the function if the OS doesn't support the capability, or does it not matter since it always returns `None`?

jaketesler:
Per @bedevere-bot – 
`I have made the requested changes; please review again`.

Python/thread_nt.h
```diff
@@ -227,6 +231,22 @@ PyThread_get_thread_ident(void)
...
+ * The native ID of a thread is valid and guaranteed to be unique system-wide
+ * from the time the thread is created until the thread has been terminated.
+ */
+unsigned long
+PyThread_get_thread_native_id(void)
+{
+    if (!initialized)
+        PyThread_init_thread();
+
+    return GetCurrentThreadId();
```
vstinner: Would you mind to explicitly cast from DWORD to unsigned long?

```
DWORD tid = GetCurrentThreadId(); 
return (unsigned long)tid;
```

Doc/library/_thread.rst
```diff
@@ -85,6 +85,18 @@ This module defines the following constants and functions:
    may be recycled when a thread exits and another thread is created.
 
 
+.. function:: get_native_id()
+
+   Return the native integral Thread ID of the current thread assigned by the kernel.
+   This is a non-negative integer, or ``None`` if the thread has not been started.
```
vstinner: I don't understand "if the thread has not been started". The _thread modules has no Threading class and a .start() method. It only has a start_new_thread() function. It seems like this doc is more for the threading.Thread.native_id attribute, no? In the _thread module, the function must never return None, no?

Doc/library/threading.rst
```diff
@@ -49,6 +49,18 @@ This module defines the following functions:
    .. versionadded:: 3.3
 
 
+.. function:: get_native_id()
+
+   Return the native integral Thread ID of the current thread assigned by the kernel.
+   This is a non-negative integer, or ``None`` if the thread has not been started.
```
vstinner: Same question about None.

Doc/library/threading.rst
```diff
@@ -297,6 +309,25 @@ since it is impossible to detect the termination of alien threads.
       another thread is created.  The identifier is available even after the
       thread has exited.
 
+   .. attribute:: native_id
+
+      The native integral thread ID of this thread or ``None`` if the thread has not
+      been started.  This is a non-negative integer.  See the
```
vstinner: I suggest to move "or None" in the second sentence.

```
The native integral thread ID of this thread.
This is a non-negative integer,  or ``None`` if the thread has not
been started. 
```

Doc/library/threading.rst
```diff
@@ -297,6 +309,25 @@ since it is impossible to detect the termination of alien threads.
...
+      :func:`get_native_id` function.
+      This represents the Thread ID (``TID``) as assigned to the
+      thread by the OS (kernel).  Its value may be used to uniquely identify
+      this particular thread system-wide.
```
vstinner: You may add "(until the thread terminates, after which the value may be recycled by the OS)".

I don't think it's redundant with the following note.

Lib/_dummy_thread.py
vstinner: The threading module is now always available, so I don't understand the purpose of the dummy_threading module!?

_dummy_thread.start_new_thread(func) runs in the same thread. I understand that dummy threading design is that it behaves as if all threads are always running. So the dummy get_native_id() should return an integer.

Honestly, it seems way simpler to not provide a fake get_native_id(), as the function is not always available: just remove it.

Lib/threading.py
```diff
@@ -891,6 +899,10 @@ def _bootstrap(self):
...
```
vstinner: I would prefer to not defined the function if _HAVE_THREAD_NATIVE_ID is false:

```
if _HAVE_THREAD_NATIVE_ID:
    def _set_native_id(self): ...
```

It prevents the need of an assertion ;-)

Lib/threading.py
```diff
@@ -1077,6 +1091,18 @@ def ident(self):
         assert self._initialized, "Thread.__init__() not called"
         return self._ident
 
+    @property
```
vstinner: The property must not be defined if _HAVE_THREAD_NATIVE_ID is false:

```
if _HAVE_THREAD_NATIVE_ID:
    @property
    def ...
```

Python/thread_nt.h
```diff
@@ -227,6 +231,22 @@ PyThread_get_thread_ident(void)
...
+#ifdef PY_HAVE_THREAD_NATIVE_ID
+/*
+ * Return the native Thread ID (TID) of the calling thread.
+ * The native ID of a thread is valid and guaranteed to be unique system-wide
+ * from the time the thread is created until the thread has been terminated.
+ */
+unsigned long
+PyThread_get_thread_native_id(void)
+{
+    if (!initialized)
```
vstinner: PEP 7: please add { ... }. Old code is left unchanged, but new code should respect PEP 7 ;-)

Python/thread_pthread.h
```diff
@@ -302,6 +308,26 @@ PyThread_get_thread_ident(void)
...
+        PyThread_init_thread();
+#ifdef __APPLE__
+    uint64_t native_id;
+    pthread_threadid_np(NULL, &native_id);
+#elif defined(__linux__)
+    pid_t native_id;
+    native_id = syscall(__NR_gettid);
+#elif defined(__FreeBSD__)
+    pid_t native_id;
+    native_id = pthread_getthreadid_np();
```
vstinner, line 325: Are you sure about the "pid_t" type? The following FreeBSD manual page documents the return type as "int":
https://www.freebsd.org/cgi/man.cgi?query=pthread_getthreadid_np&sektion=3&manpath=freebsd-release-ports

Python/thread_pthread.h
```diff
@@ -12,6 +12,12 @@
 #endif
 #include <signal.h>
 
+#if defined(__linux__)
+#include <sys/syscall.h>
+#elif defined(__FreeBSD__)
+#include <pthread_np.h>
+#endif
```
vstinner, line 19: nitpick: I know that one of the few who like indenting macros for readbility, but I suggest:

```
#if defined(__linux__)
#  include <sys/syscall.h>   /* syscall(__NR_gettid) */
#elif defined(__FreeBSD__)
#  include <pthread_np.h>   /* pthread_getthreadid_np() */
#endif
```

Python/thread_pthread.h
```diff
@@ -302,6 +308,26 @@ PyThread_get_thread_ident(void)
...
+PyThread_get_thread_native_id(void)
+{
+    if (!initialized)
+        PyThread_init_thread();
+#ifdef __APPLE__
+    uint64_t native_id;
+    pthread_threadid_np(NULL, &native_id);
+#elif defined(__linux__)
+    pid_t native_id;
+    native_id = syscall(__NR_gettid);
```
vstinner: SYS_gettid exists as well. In posixmodule.c, we use "syscall(SYS_getrandom, ...)". In _posixsubprocess.c, we use "syscall(SYS_getdents64, ..."). I suggest to use SYS_gettid as well here.

To be clear, SYS_gettid and __NR_gettid are the same :-) https://stackoverflow.com/questions/22351033/difference-between-nr-gettid-and-sys-gettid I just prefer consistency in our code base ;-)

Python/thread_pthread.h
```diff
@@ -302,6 +308,26 @@ PyThread_get_thread_ident(void)
...
 
+#ifdef PY_HAVE_THREAD_NATIVE_ID
+unsigned long
+PyThread_get_thread_native_id(void)
+{
+    if (!initialized)
+        PyThread_init_thread();
+#ifdef __APPLE__
+    uint64_t native_id;
+    pthread_threadid_np(NULL, &native_id);
```
vstinner: In general, the function *can* fail. But in our case, it cannot fail. So I suggest to add (void) to express that we ignore the return value:

```
/* this call cannot fail: ignore return value */ 
(void)pthread_threadid_np(NULL, &native_id);
```

Implementation of pthread_threadid_np in darwin libpthread, simplified code:

```
int
pthread_threadid_np(pthread_t thread, uint64_t *thread_id)
{
	int res = 0;
	...
	if (thread == NULL) {
		*thread_id = self->thread_id;
	} else {
		...
	}
	return res;
}
```

Full code:
https://github.com/apple/darwin-libpthread/blob/master/src/pthread.c#L1111

Python/thread_pthread.h
vstinner: I checked my notes, Python 3.8 only supports macOS 10.6 and newer, so it's ok :-) https://pythondev.readthedocs.io/platforms.html#well-supported-platforms

Python/thread_pthread.h
jaketesler: Quite right – I think this was left over from the Linux implementation. Fixed :+1:

Doc/library/_thread.rst
jaketesler: In this case, yes, you are correct! :) I copy+pasted the doclines, shame on me. :(

The `_thread.get_native_id()` function will never return `None`. Fixed.

Doc/library/threading.rst
jaketesler: Same answer as above...fixed.

jaketesler:
@vstinner please advise if there are any additional suggestions you have or changes that should be implemented before the merge.

aixtools:
> 
> 
> This is the second version of PR #11993 (original reverted due to test failures on unsupported operating systems).
> 
> cc @vstinner @pitrou
> 
> From the original PR:
> 
> > This functionality adds a native Thread ID to threading.Thread objects. This ID (TID), similar to the PID of a process, is assigned by the OS (kernel) and is generally used for externally monitoring resources consumed by the running thread (or process).
> > This does not replace the ident attribute within Thread objects, which is assigned by the Python interpreter and is guaranteed as unique for the lifetime of the Python instance.
> > This new functionality fully supports Linux, Apple, and Windows platforms.

I expect the reason it broke AIX - is because AIX supports - for nearly 30 years is my guess (not known if it was already in AIX 3.X, but always in AIX 4.X).
From the AIX 6.1 documentation:

```
pthread_self Subroutine
Purpose
Returns the calling thread's ID.
Library
Threads Library (libpthreads.a)
Syntax
#include <pthread.h>
pthread_t pthread_self (void);
Description
The pthread_self subroutine returns the calling thread's ID.
Note: The pthread.h header file must be the first included file of each source file using the threads
library. Otherwise, the -D_THREAD_SAFE compilation flag should be used, or the cc_r compiler
used. In this case, the flag is automatically set.
Return Values
The calling thread's ID is returned.
Errors
No errors are defined.
The pthread_self function will not return an error code of EINTR.
```
The 'key' issue being that <pthread.h> needs to be first (or -D_THREAD_SAFE, which may be 'better' to ensure that all C-code is compiled "thread-safe" when using gcc. I normally use xlc_r, which sets _THREAD_SAFE - but I cannot assume that gcc (which is what the AIX bots currently use) are compiling everything "thread-safe".
Iirc, among other things, "thread-safe" means the linker links with the "thread_safe" libraries (provided by IBM. No guarantee that third-party libraries are thread-safe. etc..

Anyway, moving forward - how can I best contribute some code specific to AIX. Never tried working "together" before.

jaketesler:
To be clear, in all the POSIX-based platforms that I’ve worked with to build this feature (notably _excluding_ AIX, which I have zero experience with), `*_self()`-type functions are NOT the same value/ID as the one used by this feature. 

Can you confirm whether or not AIX has the capability to return the integral ID, or whether the `pthread_self()` function simply returns the same struct reference as other POSIX systems? Given the return type `pthread_t`, at first glance I’m inclined to believe it’s not the same functionality.

jaketesler:
Looking at docs, I think this might be what we’re looking for: https://www.ibm.com/support/knowledgecenter/en/ssw_aix_72/com.ibm.aix.basetrf2/thread_self.htm

I will look into this and see if it’d be easy to add for this PR.

pitrou:
I'd recommend tackling AIX as a separate PR.

jaketesler:
Either way :)

In that case, I think this is mostly good to go (pending review of course).

vstinner (approved):

vstinner:
I could continue to nitpick to death, but well, you addressed my main concern: not add the function/attribute on unsupported platforms, so I approve your change. Thanks for fixing it!

PR #13467, opened 2019-05-21
bpo-36998: Fix distutils' sdist command for undecodable filenames

abadger:
It is occassionally desirable to have undecodable filenames inside of an sdist.
For instance, testcases may want to use a filename that is undecodable to
simulate what the application will see in real-world usage.  Prior to this
change, distutils was correctly reading the filename from disk (using the
surrogatescape error handler to handle the undecodable bytes) but failing to
write the filename back out to the MANIFEST file because the characters could
not be encoded.  This change fixes that by using the surrogateescape error
handler at the point that we write the filenames out, matching with the reading
of the data.

https://bugs.python.org/issue36998


auvipy (approved):

pganssle:
@abadger This looks like a very nice pull request, but these days `distutils` is more or less frozen and not accepting new changes. The recommended guidance is to send bugs and fixes to [`setuptools`](https://github.com/pypa/setuptools) instead, because of 1. backwards compatibility concerns and 2. the fact that `setuptools` supports old versions of Python immediately (and as such this comes with a backport "for free"). Eventually `setuptools` will become the reference implementation for `distutils`.

As a maintainer of `setuptools`, I think this does look like a reasonable change to make, so I don't think you'd be wasting your time by adapting the patch to our codebase.

iritkatriel:
Closing this PR following the closure of its b.p.o issue (https://bugs.python.org/issue36998)

PR #13475, opened 2019-05-21
bpo-37003: handle f-debug in ast-unparse,

Carreau:
Ast-unparse was mistaking f'{x=}' for f"{x!r}"

https://bugs.python.org/issue37003


Carreau:
I'm not quite all that is possible with this new format so this patch is likely wrong (or at least insufficient).

Carreau:
Not quite sure that requires a news item...

PR #13477, opened 2019-05-21
bpo-36941: clean up pcbuild changes for Windows ARM64

paulmon:
I found some typos in pcbuild.sln and python.vcxproj while trying to narrow down the failure in test_SEH

https://bugs.python.org/issue36941


eamanu (approved):
LGTM

PR #13479, opened 2019-05-21
bpo-36878: Track extra text added to 'type: ignore' in the AST

msullivan:
#13238 made extra text after a `# type: ignore` accepted by the parser.
This finishes the job and actually plumbs the extra text through the
parser and makes it available in the AST.


gvanrossum (approved):
Perfect!

ilevkivskyi (approved):

PR #13480, opened 2019-05-22
Fix so that test.test_distutils can be executed by unittest and not j…

abadger:
…ust regrtest

Our developer instructions list unittest with test discovery as the
method to execute individual tests so we should make that work.  This is
similar to add7cbfb0527aaa15879f5ea9aaf2c8b322f0868 but for testing
distutils rather than testing unittest.  For instance, this change will
enable the following to work:

./python -m unittest -v distutils.tests.test_file_util

### DIFF
--- a/Lib/test/test_distutils.py
+++ b/Lib/test/test_distutils.py
@@ -10,8 +10,14 @@ import test.support
 
 
 def test_main():
+    # used by regrtest
     test.support.run_unittest(distutils.tests.test_suite())
     test.support.reap_children()
+
+
+def load_tests(*_):
+    # used by unittest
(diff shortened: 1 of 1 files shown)


abadger:
@merwok hey, here's a fix to enable the distutils test suite to work with unittest as well as regrtest.  The strategy was copied from Michael Foord's change to the test_unittest.py file.

Let me know if you'd like me to open a bug (and if you'd like me to add a NEWS entry).  I wasn't sure if this was so trivial that it didn't need either of those.

auvipy (approved):

merwok (approved):

PR #13484, opened 2019-05-22
bpo-34616: Document top level async in whatsnew/3.8.

Carreau:
https://bugs.python.org/issue34616

### DIFF
--- a/Doc/whatsnew/3.8.rst
+++ b/Doc/whatsnew/3.8.rst
@@ -244,28 +244,7 @@ The :func:`compile` built-in has been improved to accept the
 ``ast.PyCF_ALLOW_TOP_LEVEL_AWAIT`` flag. With this new flag passed,
 :func:`compile` will allow top-level ``await``, ``async for`` and ``async with``
 constructs that are usually considered invalid syntax. Asynchronous code object
-marked with the ``CO_COROUTINE`` flag may then be returned. They
-can be used in conjunction with ``await eval(...)`` (instead of ``exec``) to – for
-example – run async snippets in doctest, or build an async-repl.
-
-Example of a bare-bones async REPL ::
-
-    import asyncio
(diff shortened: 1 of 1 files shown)


Doc/whatsnew/3.8.rst
```diff
@@ -237,6 +237,37 @@ asyncio
...
 On Windows, the default event loop is now :class:`~asyncio.ProactorEventLoop`.
 
+builtins
+--------
+
+The :func:`compile` built-in has been improved to accept the
+``ast.PyCF_ALLOW_TOP_LEVEL_AWAIT`` flag. With this new flag passed,
+:func:`compile` will allow top-level ``await``, ``async for`` and ``async with``
+constructs that are usually considered invalid syntax. Asynchronous code object
+marked with the ``CO_COROUTINE`` flag may then be returned. Those code-objects
```
1st1: I'd drop it after "Those code-objects". It's a bit too much detail for a what's new entry.

Doc/whatsnew/3.8.rst
```diff
@@ -237,6 +237,37 @@ asyncio
...
+can be used in conjunction with ``await eval(...)`` (instead of ``exec``) to for
```
tirkarthi: Should "to" be "too" as in "Code objects can be used in conjunction with ``await eval(...)`` (instead of ``exec``) too. For example run async snippets in doctest, or build an async-repl."

Doc/whatsnew/3.8.rst
Carreau: Done.

Doc/whatsnew/3.8.rst
Carreau: I don't believe so; it's "They can be use to run", not "they can be used also". But it does not matter as per Yuri comment, I removed this and the example.

1st1 (approved):

tirkarthi (approved):
LGTM. Thanks.

asvetlov (approved):

PR #13486, opened 2019-05-22
bpo-31904: Only UTF-8 encoding is supported on VxWorks

hliu0:
VxWorks uses UTF-8 as the system encoding. So the test case test_c_locale_coercion.py need to be updated to use UTF-8 on VxWorks.

https://bugs.python.org/issue31904

### DIFF
--- a/Doc/library/locale.rst
+++ b/Doc/library/locale.rst
@@ -331,6 +331,8 @@ The :mod:`locale` module defines the following exception and functions:
       The function now always returns ``UTF-8`` on Android or if the UTF-8 mode
       is enabled.
 
+   On VxWorks, always return ``'UTF-8'``, the locale and the *do_setlocale*
+   argument are ignored.
 
 .. function:: normalize(localename)
 
--- a/Lib/test/test_c_locale_coercion.py
+++ b/Lib/test/test_c_locale_coercion.py
@@ -52,6 +52,10 @@ elif sys.platform == "cygwin":
     # TODO: Work out a robust dynamic test for this that doesn't rely on
     #       CPython's own locale handling machinery
     EXPECT_COERCION_IN_DEFAULT_LOCALE = False
+elif sys.platform == "vxworks":
+    # VxWorks defaults to using UTF-8 for all system interfaces
+    EXPECTED_C_LOCALE_STREAM_ENCODING = "utf-8"
+    EXPECTED_C_LOCALE_FS_ENCODING = "utf-8"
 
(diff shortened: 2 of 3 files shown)


Doc/library/locale.rst
```diff
@@ -331,6 +331,8 @@ The :mod:`locale` module defines the following exception and functions:
       The function now always returns ``UTF-8`` on Android or if the UTF-8 mode
       is enabled.
 
+   On VxWorks, always return ``'UTF-8'``, the locale and the *do_setlocale*
+   argument are ignored.
```
vstinner, line 335:
```suggestion
   argument is ignored.
```

vstinner (changes requested):
Your doc is wrong: getpreferredencoding() isn't hardcoded to UTF-8 on VxWorks.

You have to hack Lib/_bootlocale.py and Lib/locale.py using sys.platform == "vxworks" to make this effective.

I guess that you can just replace ``if hasattr(sys, 'getandroidapilevel'):`` with ``if hasattr(sys, 'getandroidapilevel') and sys.platform == "vxworks":``, and update the related comments.

hliu0:
> Your doc is wrong: getpreferredencoding() isn't hardcoded to UTF-8 on VxWorks.
> 
> You have to hack Lib/_bootlocale.py and Lib/locale.py using sys.platform == "vxworks" to make this effective.
> 
> I guess that you can just replace `if hasattr(sys, 'getandroidapilevel'):` with `if hasattr(sys, 'getandroidapilevel') and sys.platform == "vxworks":`, and update the related comments.

On VxWorks, always return "UTF-8" with sys.flags.utf8_mode is True in both locale.py&_bootlocale.py:
if sys.flags.utf8_mode:
    return 'UTF-8'
So it does't need to add judgement with sys.platform == "vxworks".

Doc/library/locale.rst
hliu0: > 

"are" is used for "the locale" and "the *do_setlocale* argument".

This line is mostly the same as comment for Android at line 328:

On Android or in the UTF-8 mode (:option:`-X` ``utf8`` option), always
return ``'UTF-8'``, the locale and the *do_setlocale* argument are ignored.

hliu0:

pxinwr:
The author of this PR has quit the job. A new PR #19448 was created to replace this one. Please anyone helps to close this PR.

corona10:
@pxinwr Thank you for letting me know

PR #13489, opened 2019-05-22
bpo-25068: urllib.request.ProxyHandler now lowercases the dict keys

ZackerySpytz:
https://bugs.python.org/issue25068

### DIFF
--- a/Lib/urllib/request.py
+++ b/Lib/urllib/request.py
@@ -799,6 +799,7 @@ class ProxyHandler(BaseHandler):
         assert hasattr(proxies, 'keys'), "proxies must be a mapping"
         self.proxies = proxies
         for type, url in proxies.items():
+            type = type.lower()
             setattr(self, '%s_open' % type,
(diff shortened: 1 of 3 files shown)


BoboTiG (approved):
Thanks @ZackerySpytz, LGTM :)

Lib/urllib/request.py
```diff
@@ -799,6 +799,7 @@ def __init__(self, proxies=None):
         assert hasattr(proxies, 'keys'), "proxies must be a mapping"
         self.proxies = proxies
         for type, url in proxies.items():
+            type = type.lower()
```
tirkarthi, line 802: Is it worth adding this behavior change in https://docs.python.org/3/library/urllib.request.html#urllib.request.ProxyHandler with a `versionchanged` directive note?

ZackerySpytz: For what it's worth, I don't think that this change needs to be documented, but it would be nice to see what a core dev thinks.

matrixise (approved):
Thank you for your PR

PR #13490, opened 2019-05-22
bpo-36829: Add test.support.catch_unraisable_exception()

vstinner:
* Copy test_exceptions.test_unraisable() to
  test_sys.UnraisableHookTest().
* test_exceptions.test_unraisable() uses catch_unraisable_exception();
  simplify the test. test_sys now checks the exact output.
* Use catch_unraisable_exception() in test_coroutines,
  test_exceptions, test_generators.

### DIFF
--- a/Lib/test/support/__init__.py
+++ b/Lib/test/support/__init__.py
@@ -3042,15 +3042,13 @@ class catch_unraisable_exception:
 
     Usage:
 
-        try:
-            with support.catch_unraisable_exception() as cm:
-                ...
-
-            # check the expected unraisable exception
+        with support.catch_unraisable_exception() as cm:
             ...
-        finally:
-            # Explicitly break reference cycle
-            cm = None
+
+            # check the expected unraisable exception: use cm.unraisable
+            ...
+
+        # cm.unraisable is None here (to break a reference cycle)
 
     The finally block is required: cm.unraisable contains a traceback object
     which indirectly references the 'cm' variable, it's a reference cycle.
@@ -3069,4 +3067,6 @@ class catch_unraisable_exception:
         return self
 
     def __exit__(self, *exc_info):
+        # Clear the unraisable exception to explicitly break a reference cycle
+        self.unraisable = None
         sys.unraisablehook = self._old_hook
--- a/Lib/test/support/__init__.py
+++ b/Lib/test/support/__init__.py
@@ -3034,3 +3034,39 @@ def collision_stats(nbins, nballs):
         collisions = k - occupied
         var = dn*(dn-1)*((dn-2)/dn)**k + meanempty * (1 - meanempty)
         return float(collisions), float(var.sqrt())
+
+
+class catch_unraisable_exception:
+    """
+    Context manager catching unraisable exception using sys.unraisablehook.
+
+    Usage:
+
+        try:
+            with support.catch_unraisable_exception() as cm:
+                ...
+
+            # check the expected unraisable exception
+            ...
+        finally:
+            # Explicitly break reference cycle
+            cm = None
+
+    The finally block is required: cm.unraisable contains a traceback object
+    which indirectly references the 'cm' variable, it's a reference cycle.
+    """
+
+    def __init__(self):
+        self.unraisable = None
+        self._old_hook = None
+
+    def _hook(self, unraisable):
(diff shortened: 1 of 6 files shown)


vstinner:
I chose to leave test_io unchanged on purpose: I have a local branch which fix also https://bugs.python.org/issue36918 in Lib/_pyio.py **and** use this new catch_unraisable_exception() function. Once this PR will be merged, I will write a second PR based on it.

Lib/test/test_coroutines.py
```diff
@@ -2342,12 +2342,24 @@ def test_unawaited_warning_when_module_broken(self):
...
+                    coro = corofn()
+                    coro_repr = repr(coro)
+
+                    # clear reference to the coroutine without awaiting for it
+                    del coro
+                    support.gc_collect()
+
+                self.assertEqual(repr(cm.unraisable.object), coro_repr)
+                self.assertEqual(cm.unraisable.exc_type, ZeroDivisionError)
+                self.assertIn("was never awaited", stream.getvalue())
```
vstinner, line 2357: It's a little bit strange to have to catch half of the error message from stderr, and the other half from sys.unraisablehook. My PR #13488 will allow to inject the error message in the unraisable exception to catch both at the same time.

Lib/test/support/__init__.py
```diff
@@ -3034,3 +3034,39 @@ def collision_stats(nbins, nballs):
...
+
+    Usage:
+
+        try:
+            with support.catch_unraisable_exception() as cm:
+                ...
+
+            # check the expected unraisable exception
+            ...
+        finally:
```
graingert: this try/with/finally is somewhat tricky, how about:
```python
try:
    with support.throw_unraisable_exceptions():
        ...
except Exception as e:
    ... # the exception is now here
```

eg:
```python
@contextlib.contextmanager
def throw_unraisable_exceptions():
    unraisable = None
    old_hook = sys.unraisablehook

    def hook(exc):
        nonlocal unraisable
        unraisable = exc

    sys.unraisablehook = hook
    try:
        yield
        if unraisable is not None:
            raise unraisable
    finally:
        unraisable = None
        sys.unraisablehook = old_hook
```

graingert: or recommend usage like this:
```python
    with support.catch_unraisable_exceptions() as cm:
        ...
        cm.unraisable  # only available here
    assert cm.unraisable is None  # now it's gone
```
by updating `__exit__`: 
```python
   def __exit__(self, *exc_info):
        self.unraisable = None  # clear unraisable here
        sys.unraisablehook = self._old_hook
```

vstinner: Good idea :-) I implemented your ``__exit__`` idea to avoid the need for try/finally.

throw_unraisable_exceptions() might be useful, but modified tests needs to access the 'obj' attribute of the unraisable hook. Later, they might also want to get access to the 'err_msg' attribute: https://github.com/python/cpython/pull/13488

graingert: There's still a sharp edge here:

how about making `cm.unraisable` throw if accessed outside the context instead of being None, eg:

```
def __exit__(self, *exc_info):
    del self.unraisable
    sys.unraisablehook = self._old_hook
```

graingert: and if you want to access additional unraisablehook info using throw_unraisable_exceptions:

```python
class UnraisableException(Exception):
    def __init__(self, unraisable):
        self.unraisable = unraisable
        super().__init__(self, unraisable)


@contextlib.contextmanager
def throw_unraisable_exceptions():
    unraisable = None
    old_hook = sys.unraisablehook

    def hook(unraisable_):
        nonlocal unraisable
        unraisable = unraisable_

    sys.unraisablehook = hook
    try:
        yield
        if unraisable is not None:
            raise UnraisableException(unraisable) from unraisable.exc_value
    finally:
        sys.unraisablehook = old_hook
```

then you can use it with:
```python
try:
    with throw_unraisable_exceptions():
        ...
except UnraisableException as e:
    print(repr(e.__cause__))
    err_msg = e.unraisable.err_msg
```

vstinner: I wrote PR #13554 to implement your "del self.unraisable" idea.

PR #13492, opened 2019-05-22
bpo-37008: make mock_open handle able to honor next()

Anvil:
I've reported the issue on https://bugs.python.org/issue37008 and now I'm trying to bring a solution to this minor issue.

I think it could be trivially backported to 3.7 branch.

https://bugs.python.org/issue37008

### DIFF
--- a/Lib/unittest/test/testmock/testmock.py
+++ b/Lib/unittest/test/testmock/testmock.py
@@ -1710,6 +1710,8 @@ class MockTest(unittest.TestCase):
         lines = [line for line in f1]
         self.assertEqual(lines[0], '3rd line')
         self.assertEqual(list(f1), [])
+        with self.assertRaises(StopIteration):
+            next(h)
 
     def test_mock_open_write(self):
         # Test exception in file writing write()
--- a/Lib/unittest/test/testmock/testmock.py
+++ b/Lib/unittest/test/testmock/testmock.py
@@ -1705,13 +1705,15 @@ class MockTest(unittest.TestCase):
     def test_mock_open_using_next(self):
         mocked_open = mock.mock_open(read_data='1st line\n2nd line\n3rd line')
         f1 = mocked_open('a-name')
-        self.assertEqual(next(f1), '1st line\n')
-        self.assertEqual(f1.__next__(), '2nd line\n')
+        line1 = next(f1)
+        line2 = f1.__next__()
         lines = [line for line in f1]
+        self.assertEqual(line1, '1st line\n')
+        self.assertEqual(line2, '2nd line\n')
         self.assertEqual(lines[0], '3rd line')
         self.assertEqual(list(f1), [])
         with self.assertRaises(StopIteration):
-            next(h)
+            next(f1)
 
     def test_mock_open_write(self):
         # Test exception in file writing write()
--- a/Lib/unittest/test/testmock/testmock.py
+++ b/Lib/unittest/test/testmock/testmock.py
@@ -1702,6 +1702,15 @@ class MockTest(unittest.TestCase):
         self.assertEqual(lines[1], 'Norwegian Blue')
         self.assertEqual(list(f1), [])
 
+    def test_mock_open_using_next(self):
+        mocked_open = mock.mock_open(read_data='1st line\n2nd line\n3rd line')
+        f1 = mocked_open('a-name')
+        self.assertEqual(next(f1), '1st line\n')
+        self.assertEqual(f1.__next__(), '2nd line\n')
+        lines = [line for line in f1]
+        self.assertEqual(lines[0], '3rd line')
+        self.assertEqual(list(f1), [])
+
     def test_mock_open_write(self):
         # Test exception in file writing write()
         mock_namedtemp = mock.mock_open(mock.MagicMock(name='JLV'))
--- a/Misc/NEWS.d/next/Library/2019-05-22-15-26-08.bpo-37008.WPbv31.rst
+++ b/Misc/NEWS.d/next/Library/2019-05-22-15-26-08.bpo-37008.WPbv31.rst
@@ -1,0 +1,2 @@
+Add support for calling `next` with the mock resulting from
+:func:`unittest.mock.mock_open`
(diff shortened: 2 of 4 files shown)


the-knights-who-say-ni:

mariocj89:
Can you add a test that reproduces the issue today and shows it is fixed (to make sure this does not break in the future). The example you have in the issue is good, you can just add it to [the `testmock.py` file](https://github.com/python/cpython/blob/master/Lib/unittest/test/testmock/testmock.py#L1705).

You also need to add a news entry for this change, something in the lines of:

> Add support for calling `next` with the mock resulting from :func:`unittest.mock.mock_open`.

Great fix by the way! :)

tirkarthi (commented):
Please also test `__next__` along with `next(mock)` and also your `next` calls after exhausting `state[0]` raises `StopIteration`  so add a test for that too.

I would prefer having these tests grouped under the class for reference https://github.com/python/cpython/blob/33e71e01e95506cf8d93fd68251fc56352bc7b39/Lib/unittest/test/testmock/testwith.py#L142

This is a code change and would require CLA. Please sign it. Change requires a NEWS entry : https://devguide.python.org/committing/?highlight=news#what-s-new-and-news-entries

Thanks

Anvil:
@mariocj89 @tirkarthi thank you for your support and your advice. It's been appreciated. I've updated the NEWS and the unittests.

Sorry about the push force, btw, I realized in the mean time i was still using an old user.email to commit and had to amend the commits.

Anvil:
CLA has been signed, btw.

Lib/unittest/test/testmock/testmock.py
```diff
@@ -1702,6 +1702,15 @@ def test_mock_open_dunder_iter_issue(self):
         self.assertEqual(lines[1], 'Norwegian Blue')
         self.assertEqual(list(f1), [])
 
+    def test_mock_open_using_next(self):
+        mocked_open = mock.mock_open(read_data='1st line\n2nd line\n3rd line')
+        f1 = mocked_open('a-name')
+        self.assertEqual(next(f1), '1st line\n')
```
tirkarthi: Personally, grouping the expected data with variables from assert calls reads nice.  

```python
line1 = next(f1)
line2 = f1.__next__()
# line 3

self.assertEqual(line1, '1st line\n')
# assert for line 2
# assert for line 3
```

Also add a test for `StopIteration` after exhausting f1. You can use

```python
with self.assertRaises(StopIteration):
    next(f1)
```

tirkarthi: Would prefer moving it to `TestMockOpen` for these tests but would wait for other's opinion on this

Anvil:
I've added tests to both `testmock.py` and `testwith.py`.

Lib/unittest/test/testmock/testmock.py
Anvil: I've applied the changed you've suggested and also added the missing StopIteration test.

Anvil: Also, would it be more consistent to test `__next__()` prior to `next()` ?

tirkarthi: Sorry I implied moving the testmock test to testwith. It seems like two copies.

Misc/NEWS.d/next/Library/2019-05-22-15-26-08.bpo-37008.WPbv31.rst
```diff
@@ -0,0 +1,2 @@
+Add support for calling `next` with the mock resulting from
```
mariocj89: Sorry, I should have been more specific. This should probably be ```:func:`next` ``` or double backtick.
I'd suggest ``` :func:`next` ```

Lib/unittest/test/testmock/testmock.py
Anvil: Yeah, I've understood that :) but since I see some other already-redundant tests in both files and and since @mariocj89 said testmock and you said testwith.py, I've added both. No offense but I dont know either of you two so I honestly dont who is right. :)

Misc/NEWS.d/next/Library/2019-05-22-15-26-08.bpo-37008.WPbv31.rst
Anvil: My bad. Fixed.

mariocj89 (approved):
Looking great IMHO! Thanks, great fix for `mock_open`.

Lib/unittest/test/testmock/testmock.py
tirkarthi: Well, it was just an opinion of mine that to have tests grouped under the class. It's up to the core dev who will be merging this.

asvetlov (approved):
LGTM

asvetlov:
Thanks!

tirkarthi:
@cjw296 `__next__` [doesn't seem to exist on python 2](https://github.com/testing-cabal/mock/blob/e0180b98d0e07e895a3f699b7e9afcac4716fc03/mock/mock.py#L97
) and this might need to be python 3 only code during backport. Perhaps the test could be marked as python 3 only such in the backport test suite.

```python
import mock

mocked_open = mock.mock_open(read_data='one\n')
f1 = mocked_open('a-name')
line1 = f1.__next__()
```

mock with this patch applied to backport

```
python /tmp/foo.py
  File "/tmp/foo.py", line 3, in <module>
    mocked_open = mock.mock_open(read_data='one\n')
  File "/Users/karthikeyansingaravelan/stuff/python/27-venv/lib/python2.7/site-packages/mock/mock.py", line 2567, in mock_open
    handle.__next__.side_effect = _next_side_effect
  File "/Users/karthikeyansingaravelan/stuff/python/27-venv/lib/python2.7/site-packages/mock/mock.py", line 695, in __getattr__
    raise AttributeError("Mock object has no attribute %r" % name)
AttributeError: Mock object has no attribute '__next__'
```

Anvil:
Thank you everyone.

asvetlov:
@tirkarthi good point.
I think this is not a reason to reject the PR though :)

tirkarthi:
@asvetlov Chris is maintaining the mock backport in PyPI that supports Python 2 and hence I was just wanted to make a note of this for Python 2. I am happy with this improving Python 3 :)

PR #13493, opened 2019-05-22
[3.7] bpo-36907: fix refcount bug in _PyStack_UnpackDict() (GH-13381)

jdemeyer:
Backport of #13381

CC @encukou 


encukou:
Thank you!

PR #13498, opened 2019-05-22
bpo-36974: handle inheritance of vectorcall protocol

jdemeyer:


asvetlov (approved):
looks good for asyncio part

encukou (approved):
Looks  good overall!
(Not merging right now as I won't be able to watch buildbots)

PR #13499, opened 2019-05-22
Doc: Add missing forward reference in the tutorial.

JulienPalard:
Reported on https://mail.python.org/pipermail/docs/2019-March/039788.html

### DIFF
--- a/Doc/tutorial/controlflow.rst
+++ b/Doc/tutorial/controlflow.rst
@@ -482,9 +482,9 @@ When a final formal parameter of the form ``**name`` is present, it receives a
 dictionary (see :ref:`typesmapping`) containing all keyword arguments except for
 those corresponding to a formal parameter.  This may be combined with a formal
 parameter of the form ``*name`` (described in the next subsection) which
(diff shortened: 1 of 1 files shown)


Doc/tutorial/controlflow.rst
```diff
@@ -482,9 +482,9 @@ When a final formal parameter of the form ``**name`` is present, it receives a
 dictionary (see :ref:`typesmapping`) containing all keyword arguments except for
 those corresponding to a formal parameter.  This may be combined with a formal
 parameter of the form ``*name`` (described in the next subsection) which
-receives a tuple containing the positional arguments beyond the formal parameter
-list.  (``*name`` must occur before ``**name``.) For example, if we define a
-function like this::
+receives a tuple (see :ref:`tut-tuples`) containing the positional
```
csabella: Do you think just making `tuple` a link would be OK instead of adding the `see` part?

```suggestion
receives a :ref:`tuple <tut-tuples>` containing the positional
```

JulienPalard:
@csabella I like the idea :)

PR #13504, opened 2019-05-22
bpo-36878: Only allow text after `# type: ignore` if first character ASCII

msullivan:
This disallows things like `# type: ignoreé`, which seems wrong.

Also switch to using Py_ISALNUM for the alnum check, for consistency
with other code (and maybe correctness re: locale issues?).


msullivan:
@ilevkivskyi @gvanrossum last one in this sequence, I think!

gvanrossum (approved):

gvanrossum:
Yeah, `Py_ISALNUM` uses a table that's hard-coded for ASCII, and that's what we want here.

msullivan:
Buildbot failure is an infrastructure failure, it looks like: `OSError: [Errno 28] No space left on device: 'areallylongpackageandmodulenametotestreprtruncation'`

PR #13512, opened 2019-05-22
bpo-18748: _pyio.IOBase emits unraisable exception

vstinner:
In development (-X dev) mode and in a debug build, IOBase finalizer
of the _pyio module now logs the exception if the close() method
fails. The exception is ignored silently by default in release build.

test_io: test_error_through_destructor() now uses
support.catch_unraisable_exception() rather than capturing stderr.

https://bugs.python.org/issue18748

### DIFF
--- a/Doc/whatsnew/3.8.rst
+++ b/Doc/whatsnew/3.8.rst
@@ -303,7 +303,7 @@ for :func:`property`, :func:`classmethod`, and :func:`staticmethod`::
 io
 --
 
-In development mode (:opt:`-X` ``env``) and in debug build, the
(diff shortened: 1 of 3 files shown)


vstinner:
Ah, test_io failed on Travis CI. Example:

```
FAIL: test_error_through_destructor (test.test_io.CBufferedReaderTest)
  File "/home/travis/build/python/cpython/Lib/test/test_io.py", line 1107, in test_error_through_destructor
    self.assertIsNotNone(cm.unraisable)
AssertionError: unexpectedly None
```

PR #8610, opened 2018-08-01
bpo-1230540: Invoke custom sys.excepthook for threads in threading

vlasovskikh:
Threads created via threading.Thread didn't invoke sys.excepthook if there was an uncaught exception. Since the docs for sys.excepthook are not specific about using it from threads and, in fact, this hook works for threads created via the _thread module, it makes sense to enable it the threading module as well.

https://bugs.python.org/issue1230540

PR #13515, opened 2019-05-23
bpo-1230540: Add threading.excepthook()

vstinner:
Add a new threading.excepthook() function which handles uncaught
Thread.run() exception. It can be overridden to control how uncaught
exceptions are handled.

threading.ExceptHookArgs is not documented on purpose: it should not
be used directly.

* threading.excepthook() and threading.ExceptHookArgs.
* Add _PyErr_Display(): similar to PyErr_Display(), but accept a
  'file' parameter.
* Add _thread._excepthook(): C implementation of the exception hook
  calling _PyErr_Display().
* Add _thread._ExceptHookArgs: structseq type.
* Add threading._invoke_excepthook_wrapper() which handles the gory
  details to ensure that everything remains alive during Python
  shutdown.
* Add unit tests.

https://bugs.python.org/issue1230540

### DIFF
--- a/Doc/library/sys.rst
+++ b/Doc/library/sys.rst
@@ -298,7 +298,11 @@ always available.
    before the program exits.  The handling of such top-level exceptions can be
    customized by assigning another three-argument function to ``sys.excepthook``.
 
-   See also :func:`unraisablehook` which handles unraisable exceptions.
+    .. seealso::
+
+       The :func:`sys.unraisablehook` function handles unraisable exceptions
+       and the :func:`threading.excepthook` function handles uncaught
+       :func:`threading.Thread.run` exceptions.
 
 
 .. data:: __breakpointhook__
--- a/Doc/library/sys.rst
+++ b/Doc/library/sys.rst
@@ -298,11 +298,11 @@ always available.
    before the program exits.  The handling of such top-level exceptions can be
    customized by assigning another three-argument function to ``sys.excepthook``.
 
-    .. seealso::
+   .. seealso::
 
-       The :func:`sys.unraisablehook` function handles unraisable exceptions
-       and the :func:`threading.excepthook` function handles uncaught
-       :func:`threading.Thread.run` exceptions.
+      The :func:`sys.unraisablehook` function handles unraisable exceptions
+      and the :func:`threading.excepthook` function handles exception raised
+      by :func:`threading.Thread.run`.
 
 
 .. data:: __breakpointhook__
--- a/Doc/library/threading.rst
+++ b/Doc/library/threading.rst
@@ -38,6 +38,30 @@ This module defines the following functions:
    returned.
 
 
+.. function:: excepthook(args, /)
+
+   Handle uncaught :func:`Thread.run` exception.
+
+   The *args* argument has the following attributes:
+
+   * *exc_type*: Exception type.
+   * *exc_value*: Exception value, can be ``None``.
+   * *exc_traceback*: Exception traceback, can be ``None``.
+   * *thread*: Thread which raised the exception, can be ``None``.
+
+   If *exc_type* is :exc:`SystemExit`, the exception is silently ignored.
+
+   If it raises an exception, :func:`sys.excepthook` is called to handle it.
+
+   :func:`threading.excepthook` can be overridden to control how uncaught
+   :func:`Thread.run` exceptions are handled.
+
+   .. seealso::
+      :func:`sys.excepthook` handles uncaught exceptions.
+
+   .. versionadded:: 3.8
+
+
 .. function:: get_ident()
 
    Return the 'thread identifier' of the current thread.  This is a nonzero
@@ -190,6 +214,10 @@ called is terminated.
 
 A thread has a name.  The name can be passed to the constructor, and read or
 changed through the :attr:`~Thread.name` attribute.
+
+If the :meth:`~Thread.run` method raises an exception,
+:func:`threading.excepthook` is called to handle it. By default,
+:func:`threading.excepthook` ignores silently :exc:`SystemExit`.
 
 A thread can be flagged as a "daemon thread".  The significance of this flag is
 that the entire Python program exits when only daemon threads are left.  The
--- a/Doc/library/threading.rst
+++ b/Doc/library/threading.rst
@@ -40,7 +40,7 @@ This module defines the following functions:
 
 .. function:: excepthook(args, /)
 
-   Handle uncaught :func:`Thread.run` exception.
+   Handle uncaught exception raised by :func:`Thread.run`.
 
    The *args* argument has the following attributes:
 
@@ -50,11 +50,13 @@ This module defines the following functions:
    * *thread*: Thread which raised the exception, can be ``None``.
 
    If *exc_type* is :exc:`SystemExit`, the exception is silently ignored.
+   Otherwise, the exception is printed out on :data:`sys.stderr`.
 
-   If it raises an exception, :func:`sys.excepthook` is called to handle it.
+   If  this function raises an exception, :func:`sys.excepthook` is called to
+   handle it.
 
    :func:`threading.excepthook` can be overridden to control how uncaught
-   :func:`Thread.run` exceptions are handled.
+   exceptions raised by :func:`Thread.run` are handled.
 
    .. seealso::
       :func:`sys.excepthook` handles uncaught exceptions.
--- a/Lib/threading.py
+++ b/Lib/threading.py
@@ -5,7 +5,6 @@ import sys as _sys
 import _thread
 
 from time import monotonic as _time
-from traceback import format_exc as _format_exc
 from _weakrefset import WeakSet
 from itertools import islice as _islice, count as _count
 try:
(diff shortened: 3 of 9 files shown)


vstinner:
cc @vlasovskikh

vstinner:
See PR #8610 for a different approach: reuse sys.excepthook.

vstinner:
In my current implementation, I chose to continue to rely on t_bootstrap() of Modules/_threadmodule.c to log exception in threading.excepthook().

Maybe another option would be to attempt to reuse sys.excepthook() to log threading.excepthook() new failure?

Maybe the best option would be to later reimplement threading.excepthook() in C which reduces a lot the risk of error during Python shutdown. In C, it's more reliable to call functions during Python shutdown. Especially functions which don't rely on Python modules, but only access the interpreter state, thread state and the current exception.

I prefer to start with a simple PR to introduce the new API, and later experiment the different options to make the code even more reliable.

Doc/library/sys.rst
```diff
@@ -248,7 +248,11 @@ always available.
    before the program exits.  The handling of such top-level exceptions can be
    customized by assigning another three-argument function to ``sys.excepthook``.
 
-   See also :func:`unraisablehook` which handles unraisable exceptions.
+    .. seealso::
+
+       The :func:`sys.unraisablehook` function handles unraisable exceptions
+       and the :func:`threading.excepthook` function handles uncaught
+       :func:`threading.Thread.run` exception.
```
graingert: Grammar

vstinner: English is not my first language, so if you spot something, please suggest a fix :-)

vstinner:
I rewrote my PR, see the details in the issue: https://bugs.python.org/issue1230540#msg343443

vstinner:
Oops, I didn't revert properly my change which added "stderr" into arguments passed to threading.excepthook(). It's now fixed.

Doc/library/sys.rst
graingert:
```suggestion
       :func:`threading.Thread.run` exceptions.
```

graingert: Or maybe:

```
the uncaught
       :func:`threading.Thread.run` exception.
```

Depends what you're going for though

auvipy (approved):

vstinner:
@serhiy-storchaka @pablogsal @pitrou: Would you mind to review the updated PR? See https://bugs.python.org/issue1230540 for the discussion.

pitrou:
From the doc: `Handle uncaught :func:``Thread.run`` exception.`

What if some C extension wants to report uncaught exceptions in C-created threads? Does it have to create a dummy `Thread` object?

For example, how about threads created by `_thread.start_new_thread`?

vstinner:
I modified my PR to use threading identitifer (threading.get_ident()) as the thread name if args.thread is None.

vstinner:
As you can see in test_excepthook_thread_None(), it's not possible to call directly threading.excepthook() directly only using the public API: threading._make_excepthook_args() is needed.

```
args = threading._make_excepthook_args([*sys.exc_info(), None])
```

Should it be made public? Maybe as threading.ExceptHookArgs?

Maybe it can be made public but not documented?

vstinner:
I squashed my commits, rebased my PR on master and I renamed _make_excepthook_args as threading.ExceptHookArgs.

I chose to not document ExceptHookArgs. In the Python implementation, threading.ExceptHookArgs is a wrapper to threading._ExceptHookArgs: threading._ExceptHookArgs constructor takes N arguments, whereas ExceptHookArgs takes 1 argument.

Using threading.ExceptHookArgs, it's now possible to call directly threading.excepthook(threading.ExceptHook(...)) to log an error from a thread not created using threading.Thread, but using directly _thread.start_new_thread.

@pitrou: Does it look better to you?

vstinner:
I plan to merge this PR wednesday (in 2 days).

I don't see a strong opposition, and nobody came up with a working implementation for the idea of reusing sys.excepthook.

There is PR #8610 but I dislike it: threading.Thread doesn't call sys.excepthook to handle run() exception by default, it only calls sys.excepthook if it's overridden. Moreover, when sys.excepthook is called, the hook doesn't get access to the thread object :-(
https://bugs.python.org/issue1230540#msg343260

Lib/threading.py
```diff
@@ -1163,6 +1116,104 @@ def getName(self):
...
+
+try:
+    from _thread import (_excepthook as excepthook,
+                         _ExceptHookArgs as ExceptHookArgs)
+except ImportError:
+    # Simple Python implementation if _thread._excepthook() is not available
+    from traceback import print_exception as _print_exception
+    from collections import namedtuple
+
+    _ExceptHookArgs = namedtuple(
```
asvetlov, line 1128: Is namedtuple for possible future extending a list of provided arguments?

Python/pythonrun.c
```diff
@@ -953,10 +953,11 @@ print_exception_recursive(PyObject *f, PyObject *value, PyObject *seen)
 }
 
 void
-PyErr_Display(PyObject *exception, PyObject *value, PyObject *tb)
+_PyErr_Display(PyObject *file, PyObject *exception, PyObject *value, PyObject *tb)
```
asvetlov, line 956: Make it public maybe? `PyErr_DisplayToFile` or something like this?

Doc/library/threading.rst
```diff
@@ -38,6 +38,30 @@ This module defines the following functions:
...
+   The *args* argument has the following attributes:
+
+   * *exc_type*: Exception type.
+   * *exc_value*: Exception value, can be ``None``.
+   * *exc_traceback*: Exception traceback, can be ``None``.
+   * *thread*: Thread which raised the exception, can be ``None``.
+
+   If *exc_type* is :exc:`SystemExit`, the exception is silently ignored.
+
+   If it raises an exception, :func:`sys.excepthook` is called to handle it.
```
pitrou: What is "it"? "If this function raises an exception", perhaps?

Doc/library/threading.rst
```diff
@@ -38,6 +38,30 @@ This module defines the following functions:
    returned.
 
 
+.. function:: excepthook(args, /)
+
+   Handle uncaught :func:`Thread.run` exception.
```
pitrou: I would say: `Handle an uncaught exception raised by :func:`Thread.run``.

Doc/library/threading.rst
```diff
@@ -38,6 +38,30 @@ This module defines the following functions:
...
+
+   The *args* argument has the following attributes:
+
+   * *exc_type*: Exception type.
+   * *exc_value*: Exception value, can be ``None``.
+   * *exc_traceback*: Exception traceback, can be ``None``.
+   * *thread*: Thread which raised the exception, can be ``None``.
+
+   If *exc_type* is :exc:`SystemExit`, the exception is silently ignored.
```
pitrou, line 52: Perhaps add: "Otherwise, the exception is printed out on ``sys.stderr``".

Doc/library/sys.rst
```diff
@@ -298,7 +298,11 @@ always available.
    before the program exits.  The handling of such top-level exceptions can be
    customized by assigning another three-argument function to ``sys.excepthook``.
 
-   See also :func:`unraisablehook` which handles unraisable exceptions.
+    .. seealso::
```
pitrou: Is there a misindent here?

Modules/_threadmodule.c
```diff
@@ -1309,6 +1311,144 @@ requiring allocation in multiples of the system memory page size\n\
...
+        return -1;
+    }
+
+    PyObject *name = NULL;
+    if (thread != Py_None) {
+        name = PyObject_GetAttrString(thread, "name");
+    }
+    if (name != NULL) {
+        if (PyFile_WriteObject(name, file, Py_PRINT_RAW) < 0) {
+            return -1;
```
pitrou, line 1330: `Py_DECREF(name)` here?

Modules/_threadmodule.c
```diff
@@ -1309,6 +1311,144 @@ requiring allocation in multiples of the system memory page size\n\
...
+        PyObject *str = PyUnicode_FromFormat("%lu", ident);
+        if (str != NULL) {
+            if (PyFile_WriteObject(str, file, Py_PRINT_RAW) < 0) {
+                Py_DECREF(str);
+                return -1;
+            }
+            Py_DECREF(str);
+        }
+        else {
+            if (PyFile_WriteString("<failed to get thread name>", file) < 0) {
```
pitrou, line 1349: `PyErr_Clear()` before this?

pitrou (changes requested):
Looks mostly good to me. Some comments below.

Lib/threading.py
vstinner: Exactly. See sys.unraisablehook: I used the same approach and today I added a new "err_msg" attribute ;-)

Python/pythonrun.c
vstinner: I would prefer to not add a new function to the public C API unless there is a clear need from users. In the past, the C API evolved organically and now it's a mess because nobody knows what is really used or not :-(

Modules/_threadmodule.c
vstinner: fixed

Modules/_threadmodule.c
vstinner: fixed

Doc/library/sys.rst
vstinner: oops, fixed

Doc/library/threading.rst
vstinner: done

Doc/library/threading.rst
vstinner: done

Doc/library/threading.rst
vstinner: done

vstinner:

pitrou (approved):

vstinner:
Thanks for the review @pitrou!

vstinner:
> For example, how about threads created by _thread.start_new_thread?

I created https://bugs.python.org/issue37076 for this specific function ;-)

PR #13516, opened 2019-05-23
bpo-37017: Make PyObject_CallMethodObjArgs use LOAD_METHOD optimization

msullivan:
Update PyObject_CallMethodObjArgs and _PyObject_CallMethodIdObjArgs
to use _PyObject_GetMethod to avoid creating a bound method object
in many cases.

On a microbenchmark of PyObject_CallMethodObjArgs calling a method on
an interpreted Python class, this optimization resulted in a 1.7x
speedup.

https://bugs.python.org/issue37017

### DIFF
--- a/Misc/NEWS.d/next/C API/2019-05-22-17-33-52.bpo-37107.8BVPR-.rst
+++ b/Misc/NEWS.d/next/C API/2019-05-22-17-33-52.bpo-37107.8BVPR-.rst
@@ -1,2 +1,4 @@
-Update PyObject_CallMethodObjArgs and _PyObject_CallMethodIdObjArgs to use
-_PyObject_GetMethod to avoid creating a bound method object in many cases.
+Update c:func:PyObject_CallMethodObjArgs and c:func:_PyObject_CallMethodIdObjArgs
+to use c:func:_PyObject_GetMethod to avoid creating a bound method object in many
+cases.
+Patch by Michael J. Sullivan.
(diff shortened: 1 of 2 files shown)


auvipy (approved):

Objects/call.c
```diff
@@ -1210,23 +1215,29 @@ object_vacall(PyObject *callable, va_list vargs)
...
+    PyObject *callable, *result;
+    int is_method;
 
-    if (callable == NULL || name == NULL) {
+    if (obj == NULL || name == NULL) {
         return null_error();
     }
 
-    callable = PyObject_GetAttr(callable, name);
+    callable = NULL;
```
methane:
```suggestion
    PyObject *callable = NULL;
```

Objects/call.c
```diff
@@ -1210,23 +1215,29 @@ object_vacall(PyObject *callable, va_list vargs)
...
+    is_method = _PyObject_GetMethod(obj, name, &callable);
```
methane:
```suggestion
    int is_method = _PyObject_GetMethod(obj, name, &callable);
```

auvipy (approved):

Misc/NEWS.d/next/C API/2019-05-22-17-33-52.bpo-37107.8BVPR-.rst
```diff
@@ -0,0 +1,2 @@
+Update PyObject_CallMethodObjArgs and _PyObject_CallMethodIdObjArgs to use
```
pablogsal: Please, use markdown syntax here (for example :c:func:`PyObject_CallMethodObjArgs`) and add "Patch by PyObject_CallMethodObjArgs" at the end.

PR #13520, opened 2019-05-23
Doc: Correct the creation year and the credits of the Logo Programming language

matrixise:
See https://mail.python.org/pipermail/docs/2019-May/040679.html


csabella (approved):
LGTM.  Looking at the wiki history, there were 4 names listed for awhile, so the information had probably been modified sometime after this doc was originally created.

Carreau (approved):

PR #13522, opened 2019-05-23
bpo-37007: Implement socket.if_nametoindex() and socket.if_indextoname() on Windows

ZackerySpytz:
https://bugs.python.org/issue37007


zooba (approved):
Approved, but I asked on bpo whether it's worth doing if_nameindex as well.

ZackerySpytz:
socket.if_nameindex() is  Implemented on Windows

testInterfaceNameIndex() in test_socket.py is now run on Windows (as hasattr(socket, 'if_nameindex') is now True). Two other tests that aren't directly related to if_nameindex() have been disabled on Windows. These tests only use if_nameindex() for the purpose of picking a random network interface.

ZackerySpytz:
socket.if_nameindex() is now implemented on Windows.

testInterfaceNameIndex() in test_socket.py is now run on Windows (as hasattr(socket, 'if_nameindex') is now True). Two other tests that aren't directly related to if_nameindex() have been disabled on Windows. These tests only use if_nameindex() for the purpose of picking a random network interface.

Modules/socketmodule.c
```diff
@@ -6588,28 +6590,54 @@ Set the default timeout in seconds (float) for new socket objects.\n\
...
-    PyObject *list;
+    PyObject *list = PyList_New(0);
+    if (list == NULL) {
+        return NULL;
+    }
+#ifdef MS_WINDOWS
+    PMIB_IF_TABLE2 tbl;
+    if (GetIfTable2Ex(MibIfTableRaw, &tbl) != NO_ERROR) {
+        Py_DECREF(list);
+        return PyErr_SetFromErrno(PyExc_OSError);
```
zooba: I think we want `SetFromWindowsError` here (or something like that). `SetFromErrno` will use the POSIX `errno` variable, while we want to use `GetLastError()`.

Modules/socketmodule.c
```diff
@@ -6588,28 +6590,54 @@ Set the default timeout in seconds (float) for new socket objects.\n\
...
+    }
+    for (ULONG i = 0; i < tbl->NumEntries; ++i) {
+        MIB_IF_ROW2 r = tbl->Table[i];
+        WCHAR buf[NDIS_IF_MAX_STRING_SIZE + 1];
+        if (ConvertInterfaceLuidToNameW(&r.InterfaceLuid, buf,
+                                        NDIS_IF_MAX_STRING_SIZE + 1)) {
+            Py_DECREF(list);
+            FreeMibTable(tbl);
+            return PyErr_SetFromErrno(PyExc_OSError);
```
zooba: Same here, but we'll need to capture `GetLastError()` in a variable before calling `FreeMibTable`.

Modules/socketmodule.c
```diff
@@ -6588,28 +6590,54 @@ Set the default timeout in seconds (float) for new socket objects.\n\
...
+        }
+        PyObject *tuple = Py_BuildValue("IN", r.InterfaceIndex,
+                PyUnicode_FromWideChar(buf, -1));
```
zooba: `Py_BuildValue` can take `u` and a `wchar_t*` and will do the string conversion for you.

Modules/socketmodule.c
ZackerySpytz: Apparently, `GetLastError()` can't be used with [GetIfTable2Ex()](https://docs.microsoft.com/en-us/windows/desktop/api/netioapi/nf-netioapi-getiftable2ex#return-value) or [ConvertInterfaceLuidToNameW()](https://docs.microsoft.com/en-us/windows/desktop/api/netioapi/nf-netioapi-convertinterfaceluidtonamew#return-value). The return value of those two functions must be used (I've verified this locally).

Modules/socketmodule.c
ZackerySpytz: Done.

Modules/socketmodule.c
ZackerySpytz: See above,

ZackerySpytz:
Hi, @zooba. Thanks for the review. I've updated the PR.

Modules/socketmodule.c
ZackerySpytz: See above.

Modules/socketmodule.c
```diff
@@ -6668,28 +6670,54 @@ Set the default timeout in seconds (float) for new socket objects.\n\
...
+    PyObject *list = PyList_New(0);
+    if (list == NULL) {
+        return NULL;
+    }
+#ifdef MS_WINDOWS
+    PMIB_IF_TABLE2 tbl;
+    int ret;
+    if ((ret = GetIfTable2Ex(MibIfTableRaw, &tbl)) != NO_ERROR) {
+        Py_DECREF(list);
+        return PyErr_SetFromWindowsErr(ret);
```
zooba, line 6689:
```suggestion
        // ret is used instead of GetLastError()
        return PyErr_SetFromWindowsErr(ret);
```

I had to double check this, so probably worth calling it out. It's unusual for Windows APIs to return the error code.

Modules/socketmodule.c
```diff
@@ -6668,28 +6670,54 @@ Set the default timeout in seconds (float) for new socket objects.\n\
...
+    }
+    for (ULONG i = 0; i < tbl->NumEntries; ++i) {
+        MIB_IF_ROW2 r = tbl->Table[i];
+        WCHAR buf[NDIS_IF_MAX_STRING_SIZE + 1];
+        if ((ret = ConvertInterfaceLuidToNameW(&r.InterfaceLuid, buf,
+                                               Py_ARRAY_LENGTH(buf)))) {
+            Py_DECREF(list);
+            FreeMibTable(tbl);
+            return PyErr_SetFromWindowsErr(ret);
```
zooba, line 6699:
```suggestion
            // ret is used instead of GetLastError()
            return PyErr_SetFromWindowsErr(ret);
```

zooba (commented):
Just a suggested comment to avoid me having to look up the APIs again next time I read this :)

Modules/socketmodule.c
ZackerySpytz: Done.

Modules/socketmodule.c
ZackerySpytz: Done.

zooba:
Thanks @ZackerySpytz!

PR #13474, opened 2019-05-21
bpo-35907, CVE-2019-9948: urllib rejects local_file:// scheme

vstinner:
CVE-2019-9948: Avoid file reading as disallowing the unnecessary URL
scheme in URLopener().open() of urllib.request.

Co-Authored-By: SH <push0ebp@gmail.com>

https://bugs.python.org/issue35907

PR #13523, opened 2019-05-23
bpo-35907: Clarify the NEWS entry

vstinner:
https://bugs.python.org/issue35907


vstinner:
Gregory commented my master PR:

"This text doesn't make sense as written. "as disallowing the unnecessary URL scheme" says nothing. WHAT url scheme did we just break our users applications by refusing to support? Those need to be listed in the NEWS entry."
https://github.com/python/cpython/pull/13474/files#r286726212

I wrote this PR to clarify which scheme is rejected.

@gpshead @tirkarthi @push0ebp: Would it sound better said like that?

Once this PR will be merged, I will also update the NEWS entry in 2.7 and 3.7, and then update my 3.5 and 3.6 backports.

vstinner:
> [whatsnew/changelog:12] "`" found in "local_file://`"

Oh, I made a typo. It's now fixed.

gpshead (approved):

PR #13527, opened 2019-05-23
bpo-20443: make code objects filename an absolute path

isidentical:
make code objects filename an absolute path


benjaminp:
The compiler isn't an appropriate place to do this. `compiler("some", "exec", "<some file>")` should return a code object with `co_filename == "<some file>"`.

isidentical:
Then, where to it'll return absolute path? For not breaking backwards compability, we can set up a future flag if it is OK?

terryjreedy:
> Then, where to it'll return absolute path?

I think you meant, where to patch to that co_filename == <absolute path>?  I agree with what Benjamin implied, that the call site should pass in the absolute patch when there is one.   The passed-in 'filename' can be anything and does not have to correspond to a real file on external storage.

> For not breaking backwards compability, we can set up a future flag if it is OK?

This would definitely require a PEP.

isidentical:
> This would definitely require a PEP.
I know, i posted to python-ideas.

vstinner:
I concur with Benjamin: code object constructor is the wrong place to fix the issue. The filename must be made absolute somewhere in the caller. I'm not against https://bugs.python.org/issue20443 I'm only against the proposed implementation.

PR #13528, opened 2019-05-23
bpo-32528: Make asyncio.CancelledError a BaseException.

1st1:
This will address the common mistake many asyncio users make:
an "except Exception" clause breaking Tasks cancellation.

In addition to this change, we stop inheriting asyncio.TimeoutError
and asyncio.InvalidStateError from their concurrent.futures.*
counterparts.  There's no point for these exceptions to share the
inheritance chain.

https://bugs.python.org/issue32528

### DIFF
--- a/Lib/asyncio/base_events.py
+++ b/Lib/asyncio/base_events.py
@@ -186,7 +186,7 @@ def _interleave_addrinfos(addrinfos, first_address_family_count=1):
 def _run_until_complete_cb(fut):
     if not fut.cancelled():
         exc = fut.exception()
-        if isinstance(exc, BaseException) and not isinstance(exc, Exception):
+        if isinstance(exc, (SystemExit, KeyboardInterrupt)):
             # Issue #22429: run_forever() already finished, no need to
             # stop it.
             return
@@ -1196,7 +1196,7 @@ class BaseEventLoop(events.AbstractEventLoop):
 
         try:
             await waiter
-        except Exception:
+        except BaseException:
             transport.close()
             conmade_cb.cancel()
             resume_cb.cancel()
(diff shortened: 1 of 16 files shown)


asvetlov:
I see a small problem with the PR in `test_base_events.py`:
```
    def test_run_until_complete_loop_orphan_future_close_loop(self):
        class ShowStopper(BaseException):
            pass

        async def foo(delay):
            await asyncio.sleep(delay)

        def throw():
            raise ShowStopper

        self.loop._process_events = mock.Mock()
        self.loop.call_soon(throw)
        try:
            self.loop.run_until_complete(foo(0.1))
        except ShowStopper:
            pass

        # This call fails if run_until_complete does not clean up
        # done-callback for the previous future.
        self.loop.run_until_complete(foo(0.2))
```
The test passes but a message is printed:
```
test_run_until_complete_loop_orphan_future_close_loop (test.test_asyncio.test_base_events.BaseEventLoopTests) ... Exception in callback BaseEventLoopTests.test_run_until_complete_loop_orphan_future_close_loop.<locals>.throw() at /home/andrew/projects/cpython/Lib/test/test_asyncio/test_base_events.py:485
handle: <Handle BaseEventLoopTests.test_run_until_complete_loop_orphan_future_close_loop.<locals>.throw() at /home/andrew/projects/cpython/Lib/test/test_asyncio/test_base_events.py:485>
  File "/home/andrew/projects/cpython/Lib/asyncio/events.py", line 81, in _run
    self._context.run(self._callback, *self._args)
  File "/home/andrew/projects/cpython/Lib/test/test_asyncio/test_base_events.py", line 486, in throw
    raise ShowStopper
test.test_asyncio.test_base_events.BaseEventLoopTests.test_run_until_complete_loop_orphan_future_close_loop.<locals>.ShowStopper
ok
```

I suggest a little different implementation.
On catching exceptions let's don't change `except Exception` to `except BaseException` but replace it with `except (Exception, asyncio.CancelledError)`.

By this, we keep exception handling logic in asyncio unchanged but put CacelledError at the proper place in exceptions hierarchy.
I don't know if users use custom `BaseException` derived classes in their code. This PR has a potential regression risk, while my proposal eliminates it (I hope).

asvetlov (commented):
After offline talk @1st1 convinced me that we should process all exceptions as proposed by the PR.
Eventually in Python 3.9 we will get rid of custom handling of SystemExit and KeyboardInterrupt.

asvetlov (approved):

PR #13530, opened 2019-05-23
bpo-37027: Return a proxy socket object from transp.get_extra_info('socket')

1st1:
Return a safe to use proxy socket object from `transport.get_extra_info('socket')`

https://bugs.python.org/issue37027

### DIFF
--- a/Lib/asyncio/trsock.py
+++ b/Lib/asyncio/trsock.py
@@ -1,4 +1,5 @@
 import socket
+import warnings
 
 
 class TransportSocket:
@@ -16,7 +17,11 @@ class TransportSocket:
         self._sock = sock
 
     def _na(self, what):
-        raise TypeError(f'asyncio.TransportSocket does not support {what}')
+        warnings.warn(
+            f"Using {what} on sockets returned from get_extra_info('socket') "
+            f"will be prohibited in asyncio 3.9. Please report your use case "
+            f"to bugs.python.org.",
+            DeprecationWarning, source=self)
 
     @property
     def family(self):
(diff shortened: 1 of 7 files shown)


asvetlov (approved):
asyncio saves at least "process" and "pipe" in extra info.
Should we wrap these things too?
Should we wrap sockets from `asyncio.Server()`?

1st1:
Andrew, please re-review this PR (and please hit the merge button if it's ready).  I've relaxed this PR - now all socket methods can be called as they are in 3.7, but we issue a warning that we won't support that in 3.9.

Lib/asyncio/trsock.py
```diff
@@ -0,0 +1,206 @@
...
+
+    def dup(self):
+        return self._sock.dup()
+
+    def get_inheritable(self):
+        return self._sock.get_inheritable()
+
+    def shutdown(self, how):
+        # asyncio doesn't currently provide a high-level transport API
+        # to shutdown the connection.
```
asvetlov, line 75: Technically there is `transport.write_eof()` for closing writing.
There is no API for closing reading though.

asvetlov (approved):

Lib/asyncio/trsock.py
1st1: Yes, and I know some people use `sock.shutdown` for that.  Should we add an API to transports?

asvetlov: Maybe, to make the transition to new Python smoother.
Not urgent, though. Minor thing.

PR #13531, opened 2019-05-23
bpo-36876: Use a consistent variable name for kwlist.

ericsnowcurrently:
Having a consistent variable name helps simplify efforts to identify C globals that need to be moved.  (A consistent name is easier to ignore.)


vstinner:
I'm not super excited by PR only changing the coding style. Can't you use a tool which understand that the static is a constant and it's value is constant?

ericsnowcurrently:
You're right. :)

vstinner:
Sorry :-(

PR #13536, opened 2019-05-24
gh-81211: hide undocumented commands in cmd module

aldwinaldwin:
A flag, defaulting to false. If true, :meth:`do_help` and :meth:`completenames`
won't include undocumented commands (that is, there are do_*() methods without
corresponding help_*() methods).

https://bugs.python.org/issue37030

* Issue: gh-81211

### DIFF
--- a/Doc/library/cmd.rst
+++ b/Doc/library/cmd.rst
@@ -217,6 +217,11 @@ Instances of :class:`Cmd` subclasses have some public instance variables:
    :mod:`readline`, on systems that support it, the interpreter will automatically
    support :program:`Emacs`\ -like line editing  and command-history keystrokes.)
 
+.. attribute:: Cmd.hide_undoc
+
+    A flag, defaulting to false. If true, :meth:`do_help` and :meth:`completenames`
+    won't include undocumented commands (that is, there are do_*() methods without
+    corresponding help_*() methods).
 
 .. _cmd-example:
 
--- a/Doc/library/cmd.rst
+++ b/Doc/library/cmd.rst
@@ -223,6 +223,8 @@ Instances of :class:`Cmd` subclasses have some public instance variables:
     won't include undocumented commands (that is, there are do_*() methods without
     corresponding help_*() methods).
 
+    .. versionadded:: 3.8
+
 .. _cmd-example:
 
 Cmd Example
--- a/Lib/cmd.py
+++ b/Lib/cmd.py
@@ -72,6 +72,7 @@ class Cmd:
     undoc_header = "Undocumented commands:"
     nohelp = "*** No help on %s"
     use_rawinput = 1
+    hide_undoc = 0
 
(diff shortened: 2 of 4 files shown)


aldwinaldwin:
Test: https://github.com/aldwinaldwin/cmd_hide_undoc

Lib/cmd.py
```diff
@@ -245,8 +246,16 @@ def completedefault(self, *ignored):
...
     def completenames(self, text, *ignored):
+        """ Method returns all commands.
+
+        If hide_undoc is True, only the commands with "help_" are returned.
+        """
         dotext = 'do_'+text
-        return [a[3:] for a in self.get_names() if a.startswith(dotext)]
+        names = self.get_names()
+        commands = [a[3:] for a in names if a.startswith(dotext)]
+        if self.hide_undoc:
```
tirkarthi, line 256: Please add a test similar to https://github.com/python/cpython/blob/f1e17e9f97d9a4e97a5d99574775ee343a3a74fb/Lib/test/test_cmd.py#L55 . In the doctest you can set `mycmd.hide_undoc = True` and make sure "shell" is not returned by the `completenames` function for "s". Set it `mycmd.hide_undoc = False` for rest of the tests to make sure to make sure other tests are not affected.

Doc/library/cmd.rst
```diff
@@ -217,6 +217,11 @@ Instances of :class:`Cmd` subclasses have some public instance variables:
    :mod:`readline`, on systems that support it, the interpreter will automatically
    support :program:`Emacs`\ -like line editing  and command-history keystrokes.)
 
+.. attribute:: Cmd.hide_undoc
```
tirkarthi, line 220: This should have a `.. versionadded:: 3.8` directive since it's a new attribute.

tirkarthi (commented):
Left comment for doc and test. But in general would like to hear from a core dev on this.

Doc/library/cmd.rst
aldwinaldwin: ok, done.

Lib/cmd.py
aldwinaldwin: understood, but this is new to me, so i'll need a few days to figure this out and do this correctly

aldwinaldwin: ok, figured it out and made commit.

aldwinaldwin:
@tirkarthi Could you review the changes made following your suggestions? Thank you.

aldwinaldwin:

rhettinger:
Why would we want to encourage undocumented commands?  This seems like an anti-pattern to me.

aldwinaldwin:
> 

The reason was to have some secret functions for professional users or to have some try out functions that normal users would not start calling support about immediately. On the other had it bothered me that they are in the list to ask help for, while there is no help available for these anyway.

PR #13540, opened 2019-05-24
bpo-36710: add tstate parameter in errors.c

vstinner:
Add 'PyThreadState *tstate' parameter to errors.c functions to avoid
rely on global variables (indirectly on _PyRuntime).


vstinner:
cc @ericsnowcurrently

PR #13546, opened 2019-05-24
bpo-36379: __ipow__ must be a ternaryfunc, not a binaryfunc

ZackerySpytz:
If a type's __ipow__ method was implemented in C, attempting to use
the *modulo* parameter would cause crashes.

https://bugs.python.org/issue36379


auvipy (approved):

tiran (approved):
Thanks!

PR #13549, opened 2019-05-24
bpo-21536: On Cygwin, C extensions must be linked with libpython

embray:
It is also possible to link against a library or executable with a
statically linked libpython, but not both with the same DLL.  In fact
building a statically linked python is currently broken on Cygwin
for other (related) reasons.

The same problem applies to other POSIX-like layers over Windows
(MinGW, MSYS) but Python's build system does not seem to attempt
to support those platforms at the moment.

Follow-up to #12946

@vstinner

### DIFF
--- a/Lib/distutils/command/build_ext.py
+++ b/Lib/distutils/command/build_ext.py
@@ -714,20 +714,32 @@ class build_ext(Command):
                 # don't extend ext.libraries, it may be shared with other
                 # extensions, it is a reference to the original list
                 return ext.libraries + [pythonlib]
-        # On Android only the main executable and LD_PRELOADs are considered
-        # to be RTLD_GLOBAL, all the dependencies of the main executable
-        # remain RTLD_LOCAL and so the shared libraries must be linked with
-        # libpython when python is built with a shared python library (issue
-        # bpo-21536).
         else:
+            # On Android only the main executable and LD_PRELOADs are considered
+            # to be RTLD_GLOBAL, all the dependencies of the main executable
+            # remain RTLD_LOCAL and so the shared libraries must be linked with
+            # libpython when python is built with a shared python library (issue
+            # bpo-21536).
+            # On Cygwin (and if required, other POSIX-like platforms based on
+            # Windows like MinGW) it is simply necessary that all symbols in
+            # shared libraries are resolved at link time.
             from distutils.sysconfig import get_config_var
+            link_libpython = False
             if get_config_var('Py_ENABLE_SHARED'):
-                # Either a native build on an Android device or the
-                # cross-compilation of Python.
-                if (hasattr(sys, 'getandroidapilevel') or
-                        ('_PYTHON_HOST_PLATFORM' in os.environ and
-                         get_config_var('ANDROID_API_LEVEL') != 0)):
-                    ldversion = get_config_var('LDVERSION')
-                    return ext.libraries + ['python' + ldversion]
+                # A native build on an Android device or on Cygwin
+                if hasattr(sys, 'getandroidapilevel'):
+                    link_libpython = True
+                elif sys.platform == 'cygwin':
+                    link_libpython = True
+                elif '_PYTHON_HOST_PLATFORM' in os.environ:
+                    # We are cross-compiling for one of the relevant platforms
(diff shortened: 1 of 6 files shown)


Lib/distutils/command/build_ext.py
```diff
@@ -714,20 +714,32 @@ def get_libraries(self, ext):
...
+                if hasattr(sys, 'getandroidapilevel'):
+                    link_libpython = True
+                elif sys.platform == 'cygwin':
+                    link_libpython = True
+                elif '_PYTHON_HOST_PLATFORM' in os.environ:
+                    # We are cross-compiling for one of the relevant platforms
+                    if get_config_var('ANDROID_API_LEVEL') != 0:
+                        link_libpython = True
+                    elif get_config_var('MACHDEP') == 'cygwin':
+                        link_libpython = True
```
embray, line 739: Here I just mostly "unwrapped" the single `if` condition above into multiple explicit branches, which I think makes the logic clearer, especially with more cases added on...

vstinner:
LGTM.I merged your PR, thanks!

embray:
Thank you, though perhaps it was too quick!  There is a bug in my logic in configure.ac.

vstinner:
> Thank you, though perhaps it was too quick! There is a bug in my logic in configure.ac.

Ah? I didn't notice. Write a second PR in that case ;-)

configure
```diff
@@ -15129,9 +15129,9 @@ LDVERSION='$(VERSION)$(ABIFLAGS)'
 { $as_echo "$as_me:${as_lineno-$LINENO}: result: $LDVERSION" >&5
 $as_echo "$LDVERSION" >&6; }
 
-# On Android the shared libraries must be linked with libpython.
+# On Android and Cygwin the shared libraries must be linked with libpython.
 
-if test -z "$ANDROID_API_LEVEL"; then
+if test -z "$ANDROID_API_LEVEL" -o "$MACHDEP" != "cygwin"; then
```
embray, line 15134: This should be `-a` not `-o` technically, but it would probably be even clearer if the test weren't negated, and were instead

```
if test -n "$ANDROID_API_LEVEL" -o "$MACHDEP" = "cygwin";
```

vstinner:
Hum, is ``if test -z "$ANDROID_API_LEVEL" -o "$MACHDEP" != "cygwin"; then`` the issue?

I'm not comfortable with double negation :-) Maybe write:

```
if test -n "$ANDROID_API_LEVEL" -o "$MACHDEP" = "cygwin"; then
   LIBPYTHON=...
else
  LIBPYTHON=
fi
```

configure
vstinner: "even clearer if the test weren't negated" yeah, please invert the test conditions to avoid double negation ;-)

embray: Glad we're on the same page.  It's the double-negation that tripped me up in the first place, and probably you too.

vstinner: > probably you too.

yes

Lib/distutils/command/build_ext.py
```diff
@@ -714,20 +714,32 @@ def get_libraries(self, ext):
...
-        # bpo-21536).
         else:
+            # On Android only the main executable and LD_PRELOADs are considered
+            # to be RTLD_GLOBAL, all the dependencies of the main executable
+            # remain RTLD_LOCAL and so the shared libraries must be linked with
+            # libpython when python is built with a shared python library (issue
+            # bpo-21536).
+            # On Cygwin (and if required, other POSIX-like platforms based on
+            # Windows like MinGW) it is simply necessary that all symbols in
+            # shared libraries are resolved at link time.
```
eli-schwartz, line 725: @embray trying to better understand this change...

On Android, before this PR, it is technically possible to refrain from linking to libpython if and only if python itself is configured as a static build.

This Cygwin change applies here also within `if get_config_var('Py_ENABLE_SHARED'):` -- does Cygwin only need libpython in shared builds, like Cygwin, or all the time, like MSVC?

The commit message glosses over it by pointing out that static Cygwin builds are broken anyway. I am simply curious what the most correct approach is (when implementing similar logic outside of distutils).

PR #13552, opened 2019-05-24
bpo-21536: Follow-up to #13549

embray:
This test in #13549 was not correct.  I rewrote it a bit more clearly.  This is necessary for `python-config` to work properly on Cygwin.

I also added a small fix to a (trivial) bug I noticed in `configure`.  @vstinner 


vstinner:
Merged, thanks again.

PR #12463, opened 2019-03-20
bpo-32573: Handle the case when sys.argv is not set

uellue:
I encountered a situation where `sys.argv` wasn't set when using Python as an embedded interpreter. The proposed update will avoid an AttributeError in that case. If `sys.argv` is set, it should not change behavior.

There's discussions related to correct behavior in this case: 
https://bugs.python.org/issue32573

Is this proposed update to spawn.py the correct fix, or should the embedded case be handled differently, somehow?

https://bugs.python.org/issue32573

PR #13553, opened 2019-05-24
[3.7] bpo-32573: sys.argv always exists and is non-empty

vstinner:
Ensure that sys.argv always exists and is non-empty.

When Python is embedded in an application and ``argv`` is not set,
use [''].

https://bugs.python.org/issue32573

### DIFF
--- a/Lib/test/test_embed.py
+++ b/Lib/test/test_embed.py
@@ -287,7 +287,7 @@ class InitConfigTests(EmbeddingTestsMixin, unittest.TestCase):
         'coerce_c_locale_warn': 0,
 
         'program_name': './_testembed',
-        'argv': [],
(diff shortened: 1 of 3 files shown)


vstinner:
@ncoghlan: Do you consider that https://bugs.python.org/issue32573 is a bug that should be fixed in Python 3.7? Is it ok to change the behavior in a minor Python 3.7 release?

See also PR #12463.

ned-deily:
I don't think we should change this behavior in a maintenance release unless it were a regression in 3.7.  It's not, though, right?

vstinner:
> I don't think we should change this behavior in a maintenance release unless it were a regression in 3.7. It's not, though, right?

Ok. I close my PR.

PR #13554, opened 2019-05-24
bpo-36829: Document test.support.catch_unraisable_exception()

vstinner:
catch_unraisable_exception() now also removes its 'unraisable'
attribute at the context manager exit.

### DIFF
--- a/Doc/library/test.rst
+++ b/Doc/library/test.rst
@@ -1081,6 +1081,26 @@ The :mod:`test.support` module defines the following functions:
    :exc:`PermissionError` is raised.
 
 
+.. function:: catch_unraisable_exception()
+
+   Context manager catching unraisable exception using
+   :func:`sys.unraisablehook`.
+
+   Usage::
+
+       with support.catch_unraisable_exception() as cm:
+           # code creating an "unraisable exception"
+           ...
+
+           # check the unraisable exception: use cm.unraisable
+           ...
+
+       # cm.unraisable attribute no longer exists at this point
+       # (to break a reference cycle)
+
+   .. versionadded:: 3.8
+
+
 .. function:: find_unused_port(family=socket.AF_INET, socktype=socket.SOCK_STREAM)
 
    Returns an unused port that should be suitable for binding.  This is
--- a/Lib/test/test_io.py
+++ b/Lib/test/test_io.py
@@ -1098,18 +1098,14 @@ class CommonBufferedTests:
         # Test that the exception state is not modified by a destructor,
         # even if close() fails.
         rawio = self.CloseFailureIO()
-        try:
-            with support.catch_unraisable_exception() as cm:
-                with self.assertRaises(AttributeError):
-                    self.tp(rawio).xyzzy
+        with support.catch_unraisable_exception() as cm:
+            with self.assertRaises(AttributeError):
+                self.tp(rawio).xyzzy
 
             if not IOBASE_EMITS_UNRAISABLE:
                 self.assertIsNone(cm.unraisable)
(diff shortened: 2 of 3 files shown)


vstinner:
cc @tirkarthi @graingert

Doc/library/test.rst
```diff
@@ -1419,3 +1419,21 @@ script execution tests.
...
+
+.. function:: catch_unraisable_exception()
+
+   Context manager catching unraisable exception using
+   :func:`sys.unraisablehook`.
+
+   Usage::
+
+       with support.catch_unraisable_exception() as cm:
+           # code creating an "unraisable exception"
```
graingert: Might be worth suggesting gc.collect()

vstinner: gc.collect() is not needed. Why do you want to force a GC collection?

vstinner:
Oh. "del self.unraisable" was a good idea. I only used the context manager in a few places, but I already get it wrong in test_io :-) It's now fixed by this PR!

Doc/library/test.rst
```diff
@@ -1419,3 +1419,21 @@ script execution tests.
    containing the *source*.  If *compiled* is ``True``, both source files will
    be compiled and added to the zip package.  Return a tuple of the full zip
    path and the archive name for the zip file.
+
+.. function:: catch_unraisable_exception()
```
tirkarthi: This is placed under `script_helper` module doc. I Would prefer moving it `test.support` section. `.. versionadded:: 3.8` directive could be added to this.

Doc/library/test.rst
graingert: It's not needed but it's often a good trigger of these sort of exceptions

Lib/test/test_io.py
```diff
@@ -1103,10 +1103,10 @@ def test_error_through_destructor(self):
...
-                self.assertIsNone(cm.unraisable)
-            elif cm.unraisable is not None:
-                self.assertEqual(cm.unraisable.exc_type, OSError)
+                if not IOBASE_EMITS_UNRAISABLE:
+                    self.assertIsNone(cm.unraisable)
+                elif cm.unraisable is not None:
+                    self.assertEqual(cm.unraisable.exc_type, OSError)
         finally:
             # Explicitly break reference cycle
             cm = None
```
graingert: Redundant

vstinner: Nice catch, removed.

Doc/library/test.rst
vstinner: Oops, I moved the doc. I also added the markup.

Doc/library/test.rst
vstinner: Well, tests using the context manager *expect* an unraisable exception and fail if it's not the case.

vstinner: Oh. By the way. I also had the idea of installing a custom sys.unraisablehook in regrtest and repeat all errors at exit. Maybe even make the test fail if at least one unraisable exception is logged.

graingert: Can you URL me to that commit?

vstinner: It's not a commit but an idea :-) I didn't implement it.

graingert: Ok, :$ ping me when implemented!

vstinner:
@tirkarthi @graingert: Would you mind to review the updated PR?

graingert (approved):

vstinner:
@graingert: Thanks for your very good feedback, it's very valuable!

PR #13556, opened 2019-05-24
bpo-36842: Fix reference leak in tests by running out-of-proc

zooba:
https://bugs.python.org/issue36842

### DIFF
--- a/Lib/test/test_audit.py
+++ b/Lib/test/test_audit.py
@@ -10,111 +10,47 @@ from test import support
 if not hasattr(sys, "addaudithook") or not hasattr(sys, "audit"):
     raise unittest.SkipTest("test only relevant when sys.audit is available")
 
-
-class TestHook:
-    """Used in standard hook tests to collect any logged events.
-
-    Should be used in a with block to ensure that it has no impact
-    after the test completes. Audit hooks cannot be removed, so the
-    best we can do for the test run is disable it by calling close().
-    """
-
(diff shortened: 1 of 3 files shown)


Lib/test/test_audit.py
```diff
@@ -10,111 +10,47 @@
...
-        if event == "cpython._PySys_ClearAuditHooks":
-            raise RuntimeError("Should be ignored")
-        elif event == "cpython.PyInterpreterState_Clear":
-            raise RuntimeError("Should be ignored")
-
-
-def run_finalize_test():
-    """Called by test_finalize_hooks in a subprocess."""
-    sys.addaudithook(TestFinalizeHook())
+AUDIT_TESTS_PY = os.path.join(os.path.dirname(__file__), "audit-tests.py")
```
vstinner: support.findfile("audit-tests.py") can used here.

vstinner (commented):
It sounds fine to run tests in a subprocess to prevent false when hunting reference leaks (hooks cannot be unregistered). But I'm not sure about running individual tests one by one. I suggest to run the whole script, so rely on unittest.main() test discovery. Otherwise, there is a risk of forgetting the related test method in test_audit.py. test_eintr uses a similar approach.

zooba:
If the tests are run separately, it reduces the risk of them interfering with each other, and also helps with tracking test results over time.

PR #13563, opened 2019-05-25
bpo-34632: fix installation of importlib.metadata

yan12125:
https://bugs.python.org/issue34632


jaraco:
Thank you @yan12125  for putting this together. I was unaware of this subtle footgun.

PR #13565, opened 2019-05-25
bpo-34632: fix test_importlib with installed CPython

yan12125:
"Installed" builders on buildbot.python.org [1] run tests on an
installed CPython copy [2]. Running tests in this way requires all
testing data also installed.

This commit fixes the following error during running tests on an
installed copy:

> ModuleNotFoundError: No module named 'test.test_importlib.data'

[1] https://buildbot.python.org/all/#/builders?tags=%2Binstalled&tags=%2B3.x
[2] https://github.com/python/buildmaster-config/blob/master/master/custom/factories.py

https://bugs.python.org/issue34632


jaraco:
I missed this earlier. It's since been superseded by #13566.

PR #13567, opened 2019-05-25
bpo-26836: Add os.memfd_create()

ZackerySpytz:
https://bugs.python.org/issue26836


Modules/posixmodule.c
```diff
@@ -11897,6 +11901,29 @@ os_urandom_impl(PyObject *module, Py_ssize_t size)
     return bytes;
 }
 
+#ifdef HAVE_MEMFD_CREATE_SYSCALL
+/*[clinic input]
+os.memfd_create
+
+    name: FSConverter
+    flags: unsigned_int(bitwise=True)
```
tiran: According to PEP 446, all file descriptors should be non-inheritable by default. The default flags should be MFD_CLOEXEC.

Modules/posixmodule.c
```diff
@@ -14003,6 +14031,13 @@ all_ins(PyObject *m)
     if (PyModule_AddIntMacro(m, GRND_RANDOM)) return -1;
     if (PyModule_AddIntMacro(m, GRND_NONBLOCK)) return -1;
 #endif
+#ifdef HAVE_MEMFD_CREATE_SYSCALL
+    if (PyModule_AddIntMacro(m, MFD_CLOEXEC)) return -1;
+    if (PyModule_AddIntMacro(m, MFD_ALLOW_SEALING)) return -1;
+#ifdef MFD_HUGETLB
+    if (PyModule_AddIntMacro(m, MFD_HUGETLB)) return -1;
```
tiran, line 14049: Does it make sense to expose the flag without the supporting HUGETLB flags from ``HUGETLB_FLAG_ENCODE_64KB`` to ``HUGETLB_FLAG_ENCODE_16GB``?

Modules/posixmodule.c
```diff
@@ -11897,6 +11901,29 @@ os_urandom_impl(PyObject *module, Py_ssize_t size)
...
+    /
+
+[clinic start generated code]*/
+
+static PyObject *
+os_memfd_create_impl(PyObject *module, PyObject *name, unsigned int flags)
+/*[clinic end generated code: output=6681ede983bdb9a6 input=5659b8bdae914e4f]*/
+{
+    int fd;
+    if ((fd = syscall(__NR_memfd_create, PyBytes_AS_STRING(name),
```
tiran: Please use ``memfd_create()`` instead of syscall API.

tiran (changes requested):
Good work! :+1: 

Please modify the patch to use the glibc wrapper instead of the syscall interface and to use ``MFD_CLOEXEC`` as default value for the new FD. Feel free to use bits and pieces from my pending PR https://github.com/python/cpython/compare/master...tiran:bpo-26836-memfd-seal (Co-authored-by appreciated).

ZackerySpytz:
Thank you, @tiran, for the comments. I will address them soon.

Modules/posixmodule.c
ZackerySpytz: With the current implementation, both *name* and *flags* must be passed, so adding a default here will not have any effect. I'm not sure if *flags* should have a default value, but I will make that change if it is desired.

ZackerySpytz:
`MFD_CLOEXEC` is now the default flag.

There were some issues with `configure`. It seems that someone had modified `configure.ac` without regenerating `configure` in a recent commit.

Doc/library/os.rst
```diff
@@ -2982,6 +2982,44 @@ features:
...
+          MFD_HUGE_32MB
+          MFD_HUGE_256MB
+          MFD_HUGE_512MB
+          MFD_HUGE_1GB
+          MFD_HUGE_2GB
+          MFD_HUGE_16GB
+
+   These flags can be passed to :func:`memfd_create`.
+
+   .. availability:: Linux 3.17 and newer.  The ``MFD_HUGE*`` flags are only
```
tiran: Please mention that glibc >= 2.27 is required, too. http://man7.org/linux/man-pages/man2/memfd_create.2.html

Lib/test/test_os.py
```diff
@@ -3122,6 +3122,19 @@ def test_stty_match(self):
         self.assertEqual(expected, actual)
 
 
+@unittest.skipUnless(hasattr(os, 'memfd_create'), 'requires os.memfd_create')
+class MemfdCreateTests(unittest.TestCase):
+    def test_memfd_create(self):
+        fd = os.memfd_create("Hi", os.MFD_CLOEXEC)
```
tiran, line 3128: This may leak an fd if one of the checks is failing. Please make sure that the fd is ``os.close()``-ed on error.

Lib/test/test_os.py
```diff
@@ -3122,6 +3122,19 @@ def test_stty_match(self):
...
+        self.assertNotEqual(fd, -1)
+        self.assertFalse(os.get_inheritable(fd))
+        with open(fd, "wb") as f:
+            f.write(b'memfd_create')
+            self.assertEqual(f.tell(), 12)
+        fd = os.memfd_create("Hi")
```
tiran: This definitively leaks an fd.

tiran (changes requested):

Lib/test/test_os.py
ZackerySpytz: I'm sorry. You are right.

Doc/library/os.rst
ZackerySpytz: Done.

ZackerySpytz:

Lib/test/test_os.py
ZackerySpytz: Right. I added `self.addCleanup(os.close, fd)` after the `self.assertNotEqual(fd, -1)` in order to avoid calling `os.close()` on `-1` in the unlikely event of a failure there (as the `self.assertNotEqual(fd, -1)` should never fail at that point).

ZackerySpytz: (I was just posting my own thought process -- I don't have a lot of experience with the unit testing of CPython and CPython internals).

tiran (approved):
:+1:

tiran:
Argh, I wanted to edit the commit message, too.

Thanks for your hard work. Is there a PR for sealing, too?

tiran:
@ZackerySpytz Please see buildbot logs. It looks like some HUGE TABLE constants are not defined on all platforms. You may have to ifdef them all :(

ZackerySpytz:
@tiran Okay, I'll take a look.

PR #13571, opened 2019-05-25
bpo-37045: PEP 591: Add final qualifiers to typing module

ilevkivskyi:
cc @msullivan 

The implementation is straightforward, it just mimics `ClassVar` (since the latter is also a name/access qualifier, not really a type). Also it is essentially copied from `typing_extensions`.

https://bugs.python.org/issue37045

### DIFF
--- a/Doc/library/typing.rst
+++ b/Doc/library/typing.rst
@@ -940,6 +940,30 @@ The module defines the following classes, functions and decorators:
 
    See :pep:`484` for details and comparison with other typing semantics.
 
+.. decorator:: final
+
+   A decorator to indicate to type checkers that the decorated method
+   cannot be overridden, and the decorated class cannot be subclassed.
+   For example::
+
+      class Base:
+          @final
+          def done(self) -> None:
+              ...
+      class Sub(Base):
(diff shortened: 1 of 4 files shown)


Doc/library/typing.rst
```diff
@@ -1104,6 +1128,24 @@ The module defines the following classes, functions and decorators:
 
    .. versionadded:: 3.5.3
 
+.. data:: Final
+
+   A special typing construct to indicate that a name cannot be
```
gvanrossum: Add "to type checkers" (after "indicate").

Lib/typing.py
```diff
@@ -398,6 +400,24 @@ class Starship:
     be used with isinstance() or issubclass().
     """)
 
+Final = _SpecialForm('Final', doc=
+    """Special typing construct to indicate final names.
```
gvanrossum: add "to type checkers"

gvanrossum (approved):
Just some doc nits. (The PR is almost all docs anyway. ;-) Feel free to fix those and merge.

gvanrossum (approved):
Ship it!

msullivan (approved):

PR #13572, opened 2019-05-25
bpo-37046: PEP 586: Add Literal to typing module

ilevkivskyi:
cc @Michael0x2a @JukkaL

The implementation is straightforward and essentially is just copied from `typing_extensions`.

https://bugs.python.org/issue37046

### DIFF
--- a/Doc/library/typing.rst
+++ b/Doc/library/typing.rst
@@ -1078,6 +1078,22 @@ The module defines the following classes, functions and decorators:
    ``Callable[..., Any]``, and in turn to
    :class:`collections.abc.Callable`.
 
+.. data:: Literal
+
+   A type that can be used to indicate to type checkers that the
+   corresponding value has a value literally equivalent to the
+   provided parameter.  For example::
+
+      var: Literal[4] = 4
+
+   The type checker understands that var is literally equal to the
+   value 4 and no other value.
+
+   ``Literal[...]`` cannot be subclassed. There is no runtime checking
+   verifying that the parameter is actually a value instead of a type.
+
+   .. versionadded:: 3.8
+
 .. data:: ClassVar
 
    Special type construct to mark class variables.
--- a/Doc/library/typing.rst
+++ b/Doc/library/typing.rst
@@ -1081,16 +1081,22 @@ The module defines the following classes, functions and decorators:
 .. data:: Literal
 
    A type that can be used to indicate to type checkers that the
-   corresponding value has a value literally equivalent to the
-   provided parameter.  For example::
+   corresponding variable or function parameter has a value equivalent to
+   the provided literal (or one of several literals). For example::
 
-      var: Literal[4] = 4
+      def validate_simple(data: Any) -> Literal[True]:  # always returns True
+          ...
 
-   The type checker understands that var is literally equal to the
-   value 4 and no other value.
+      MODE = Literal['r', 'rb', 'w', 'wb']
+      def open_helper(file: str, mode: MODE) -> str:
+          ...
 
-   ``Literal[...]`` cannot be subclassed. There is no runtime checking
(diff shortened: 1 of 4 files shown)


Doc/library/typing.rst
```diff
@@ -1078,6 +1078,22 @@ The module defines the following classes, functions and decorators:
    ``Callable[..., Any]``, and in turn to
    :class:`collections.abc.Callable`.
 
+.. data:: Literal
+
+   A type that can be used to indicate to type checkers that the
+   corresponding value has a value literally equivalent to the
+   provided parameter.  For example::
```
gvanrossum: Writing "literally" here sounds strangely like the common (ab)use of "literally" in colloquial speech, e.g. "I was literally blown away by her performance".

More importantly, somehow this example doesn't convey why this would be an interesting thing to declare. I understand that you don't want to duplicate half the PEP in the docs, but maybe the example could still be a little more useful? Perhaps use a function and use a literal union in the example, and explain that the call must statically use one of the indicated literals. And then a link to the PEP.

Doc/library/typing.rst
```diff
@@ -1078,6 +1078,22 @@ The module defines the following classes, functions and decorators:
...
+
+      var: Literal[4] = 4
+
+   The type checker understands that var is literally equal to the
+   value 4 and no other value.
+
+   ``Literal[...]`` cannot be subclassed. There is no runtime checking
+   verifying that the parameter is actually a value instead of a type.
```
gvanrossum: There is no parameter in the example. Also, probably "verifying" -> "checking".

Lib/test/test_typing.py
```diff
@@ -489,6 +489,66 @@ def test_ellipsis_in_generic(self):
         typing.List[Callable[..., str]]
 
 
+class LiteralTests(BaseTestCase):
+    def test_basics(self):
+        # All of these are allowed.
+        Literal[1]
+        Literal[1, 2, 3]
+        Literal["x", "y", "z"]
+        Literal[None]
```
gvanrossum, line 498: Maybe add True or False and a collection using more than one type?

Lib/test/test_typing.py
```diff
@@ -489,6 +489,66 @@ def test_ellipsis_in_generic(self):
...
+
+    def test_illegal_parameters_do_not_raise_runtime_errors(self):
+        # Type checkers should reject these types, but we do not
+        # raise errors at runtime to maintain maximium flexibility.
+        Literal[int]
+        Literal[Literal[1, 2], Literal[4, 5]]
+        Literal[3j + 2, ..., ()]
+        Literal[b"foo", u"bar"]
```
gvanrossum: What's illegal about this one? The PEP shows examples of both `b"..."` and `u"..."` (L167-168).

Lib/test/test_typing.py
```diff
@@ -489,6 +489,66 @@ def test_ellipsis_in_generic(self):
...
+        Literal[1]
+        Literal[1, 2, 3]
+        Literal["x", "y", "z"]
+        Literal[None]
+
+    def test_illegal_parameters_do_not_raise_runtime_errors(self):
+        # Type checkers should reject these types, but we do not
+        # raise errors at runtime to maintain maximium flexibility.
+        Literal[int]
+        Literal[Literal[1, 2], Literal[4, 5]]
```
gvanrossum: The PEP has a similar example at L194 and claims it should be valid.

Lib/test/test_typing.py
ilevkivskyi: Likely this and the one below were prohibited in an older version of the PEP or during discussion.

gvanrossum (approved):
Great, LGTM now!

PR #13573, opened 2019-05-25
bpo-37049: PEP 589: Add TypedDict to typing module

ilevkivskyi:
cc @JukkaL

The implementation is straightforward and essentially is just copied from `typing_extensions`.

https://bugs.python.org/issue37049

### DIFF
--- a/Doc/library/typing.rst
+++ b/Doc/library/typing.rst
@@ -880,11 +880,11 @@ The module defines the following classes, functions and decorators:
 
 .. class:: TypedDict(dict)
 
-   A simple typed name space. At runtime it is equivalent to
+   A simple typed namespace. At runtime it is equivalent to
    a plain :class:`dict`.
 
    ``TypedDict`` creates a dictionary type that expects all of its
-   instances to have a certain set of keys, with each key
+   instances to have a certain set of keys, where each key is
    associated with a value of a consistent type. This expectation
    is not checked at runtime but is only enforced by type checkers.
    Usage::
@@ -899,11 +899,16 @@ The module defines the following classes, functions and decorators:
 
       assert Point2D(x=1, y=2, label='first') == dict(x=1, y=2, label='first')
 
-   The type info for introspection could be accessed via ``Point2D.__annotations__``
-   and ``Point2D.__total__``.  Backward-compatible usage::
+   The type info for introspection can be accessed via ``Point2D.__annotations__``
+   and ``Point2D.__total__``.  To allow using this feature with older versions
+   of Python that do not support :pep:`526`, ``TypedDict`` supports two additional
+   equivalent syntactic forms::
 
       Point2D = TypedDict('Point2D', x=int, y=int, label=str)
       Point2D = TypedDict('Point2D', {'x': int, 'y': int, 'label': str})
+
+   See :pep:`589` for more examples and detailed rules of using ``TypedDict``
+   with type checkers.
 
    .. versionadded:: 3.8
 
--- a/Doc/library/typing.rst
+++ b/Doc/library/typing.rst
@@ -878,6 +878,34 @@ The module defines the following classes, functions and decorators:
       The ``_field_types`` and ``__annotations__`` attributes are
       now regular dictionaries instead of instances of ``OrderedDict``.
 
+.. class:: TypedDict(dict)
+
+   A simple typed name space. At runtime it is equivalent to
+   a plain :class:`dict`.
+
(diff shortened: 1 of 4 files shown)


Doc/library/typing.rst
```diff
@@ -878,6 +878,34 @@ The module defines the following classes, functions and decorators:
       The ``_field_types`` and ``__annotations__`` attributes are
       now regular dictionaries instead of instances of ``OrderedDict``.
 
+.. class:: TypedDict(dict)
+
+   A simple typed name space. At runtime it is equivalent to
```
gvanrossum: "Namespace" is usually written as one word (Google says "Did you mean: namespace?").

Doc/library/typing.rst
```diff
@@ -878,6 +878,34 @@ The module defines the following classes, functions and decorators:
...
+   a plain :class:`dict`.
+
+   ``TypedDict`` creates a dictionary type that expects all of its
+   instances to have a certain set of keys, with each key
```
gvanrossum: "with each key" -> "where each key is"

Doc/library/typing.rst
```diff
@@ -878,6 +878,34 @@ The module defines the following classes, functions and decorators:
...
+          x: int
+          y: int
+          label: str
+
+      a: Point2D = {'x': 1, 'y': 2, 'label': 'good'}  # OK
+      b: Point2D = {'z': 3, 'label': 'bad'}           # Fails type check
+
+      assert Point2D(x=1, y=2, label='first') == dict(x=1, y=2, label='first')
+
+   The type info for introspection could be accessed via ``Point2D.__annotations__``
```
gvanrossum: could -> can

Doc/library/typing.rst
```diff
@@ -878,6 +878,34 @@ The module defines the following classes, functions and decorators:
...
+   and ``Point2D.__total__``.  Backward-compatible usage::
```
gvanrossum: I'd start a new sentence before "Backward-compatible usage" and elaborate a bit, e.g. that this is meant for compatibility with Python 3.5 and earlier, i.e. pre-PEP-526 -- this is a bit different than NamedTuple, where the functional notation is not considered to be needed for backward compatibility only (many users seem to prefer its compactness).

Also, the PEP consistently writes "backwards compatible".

Lib/test/test_typing.py
```diff
@@ -2549,6 +2561,101 @@ def test_pickle(self):
...
+        self.assertEqual(Emp.__module__, __name__)
+        self.assertEqual(Emp.__bases__, (dict,))
+        self.assertEqual(Emp.__annotations__, {'name': str, 'id': int})
+        self.assertEqual(Emp.__total__, True)
+
+    def test_basics_keywords_syntax(self):
+        Emp = TypedDict('Emp', name=str, id=int)
+        self.assertIsSubclass(Emp, dict)
+        self.assertIsSubclass(Emp, typing.MutableMapping)
+        if sys.version_info[0] >= 3:
```
gvanrossum: Always true...

Lib/test/test_typing.py
```diff
@@ -2549,6 +2561,101 @@ def test_pickle(self):
             self.assertEqual(jane2, jane)
 
 
+class TypedDictTests(BaseTestCase):
+    def test_basics_iterable_syntax(self):
```
gvanrossum: I think it's most often referred to as "functional syntax".

Lib/typing.py
```diff
@@ -87,6 +87,7 @@
     'Set',
     'FrozenSet',
     'NamedTuple',  # Not really a type.
+    'TypedDict',
```
gvanrossum: Maybe also add the same comment? You can't use it in `isinstance()` for example.

Lib/typing.py
```diff
@@ -1417,6 +1418,89 @@ def __new__(self, typename, fields=None, **kwargs):
...
+        tp_dict.__annotations__ = anns
+        if not hasattr(tp_dict, '__total__'):
+            tp_dict.__total__ = total
+        return tp_dict
+
+    __instancecheck__ = __subclasscheck__ = _check_fails
+
+
+class TypedDict(dict, metaclass=_TypedDictMeta):
+    """A simple typed name space. At runtime it is equivalent to a plain dict.
```
gvanrossum: namespace

gvanrossum (approved):
LGTM!

PR #13574, opened 2019-05-25
bpo-37047: Refactor AsyncMock setup logic for autospeccing

tirkarthi:
* Initialize awaited attribute for AsyncMock during autospec.
* Fix late binding issue for mock helpers API to ensure correct function is called during attribute access.
* Document newly implemented async methods.
* Fix assert_not_awaited error message.

https://bugs.python.org/issue37047

### DIFF
--- a/Lib/unittest/mock.py
+++ b/Lib/unittest/mock.py
@@ -2525,7 +2525,6 @@ def create_autospec(spec, spec_set=False, instance=False, _parent=None,
                  name=_name, **_kwargs)
 
     if isinstance(spec, FunctionTypes):
-        wrapped_mock = mock
         # should only happen at the top level because we don't
         # recurse for functions
         mock = _set_signature(mock, spec)
--- a/Lib/unittest/mock.py
+++ b/Lib/unittest/mock.py
@@ -2181,7 +2181,7 @@ class AsyncMockMixin(Base):
         """
         self = _mock_self
         if self.await_count != 0:
-            msg = (f"Expected {self._mock_name or 'mock'} to have been not awaited."
+            msg = (f"Expected {self._mock_name or 'mock'} to not have been awaited."
                    f" Awaited {self.await_count} times.")
             raise AssertionError(msg)
 
--- a/Lib/unittest/mock.py
+++ b/Lib/unittest/mock.py
@@ -2487,7 +2487,7 @@ def create_autospec(spec, spec_set=False, instance=False, _parent=None,
         spec = type(spec)
 
     is_type = isinstance(spec, type)
-    is_async_func = _is_async_obj(spec)
+    is_async_func = _is_async_func(spec)
     _kwargs = {'spec': spec}
     if spec_set:
         _kwargs = {'spec_set': spec}
--- a/Lib/unittest/mock.py
+++ b/Lib/unittest/mock.py
@@ -51,6 +51,13 @@ def _is_async_obj(obj):
         return False
 
 
+def _is_async_func(func):
+    if getattr(func, '__code__', None):
+        return asyncio.iscoroutinefunction(func)
+    else:
+        return False
+
+
 def _is_instance_mock(obj):
     # can't use isinstance on Mock objects because they override __class__
     # The base class for all mocks is NonCallableMock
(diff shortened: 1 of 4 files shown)


tirkarthi:
I am not sure if a NEWS entry is needed since the changes were not released but added one just in case it doesn't make it for 3.8 beta 1. Note to self and reviewer : 

* `_is_async_obj` and `_is_async_func` look very much same. Except there is `isawaitable` which also seems to accept coroutine. Since only coroutine function has to be accepted as spec I have kept them separate.
* I have added test for much of the attribute access. So test might seem little verbose and can be removed since we can assuming one method call working properly would make sure late binding is handled.
* `autospec=True` for patch uses `create_autospec` internally so a test has been added for the same.
* I haven't added a test for error message change.

pablogsal:
CC: @mariocj89

Lib/unittest/mock.py
```diff
@@ -51,6 +51,13 @@ def _is_async_obj(obj):
         return False
 
 
+def _is_async_func(func):
+    if getattr(func, '__code__', None):
```
mariocj89, line 55: Is it possible to have an async callable class? If so this won’t work.

Would a check for ‘callable’ work here?

Lib/unittest/mock.py
```diff
@@ -225,6 +232,29 @@ def reset_mock():
...
+    mock.awaited = _AwaitEvent(mock)
+
+    for attribute in ('assert_awaited',
+                      'assert_awaited_once',
+                      'assert_awaited_with',
+                      'assert_awaited_once_with',
+                      'assert_any_await',
+                      'assert_has_awaits',
+                      'assert_not_awaited'):
+        def f(attribute, *args, **kwargs):
```
mariocj89: I would use a different name than arte for this function. It makes it harder to read.
Or even better, take this function out of the loop.

mariocj89: Why do we do this over directly assigning the result of getattr here? Is it because the mock might still not be configured? Just confirming.

tirkarthi: > I would use a different name than arte for this function. It makes it harder to read. Or even better, take this function out of the loop.

I moved it out of the loop and used a different variable name. I have added a comment. Feel free to suggest if it can be rephrased better.

> Why do we do this over directly assigning the result of getattr here? Is it because the mock might still not be configured? Just confirming.

Yes, this is similar to a synchronous function being specced where a new function object is returned that needs to be configured like in [_setup_func](https://github.com/python/cpython/pull/13574/files#diff-ff75b1b83c21770847ade91fa5bb2525L215) where synchronous mock helpers are attached to function object and when called delegate to mock's helper. The original async setup code was trying to do the same except defining a named function for every helper it uses a wrapper function and uses getattr to get the corresponding method as it is being called.

```python
def assert_called_with(*args, **kwargs):
    return mock.assert_called_with(*args, **kwargs)
funcopy.assert_called_with = assert_called_with

# Could be rewritten as in this PR
def wrapper(attribute, *args, **kwargs):
    return getattr(mock, attr)(*args, **kwargs)
setattr(funcopy, 'assert_called_with', partial(wrapper, 'assert_called_with'))
```

Lib/unittest/mock.py
tirkarthi: I don't understand "Asunción callable class" Do you mean something like `async class Foo: pass` which I think is not supported. Please add an example of the same.

asvetlov: There is no async version of `callable()` builtin like `acallable()`.
`asyncio.iscoroutinefunction()` works for functions only

asvetlov (approved):

Lib/unittest/mock.py
mariocj89: 👍 note that the following will probably produce a standard mock rather than an async mock:

```python
class A:
   async def __call__(self):
       await asyncio.sleep(1)

a = A()
# Creating a spec for a will not quite likely not create the right kind of mock, as a does not have __code__.
```

We can cover this in a different PR if possible I guess.

tirkarthi: Yes, it produces a MagicMock but there could be a class with a mix of sync and async dunder methods too so not sure of the behavior and how to detect classes like this. I would like to cover it as part of different PR since this contain fixes for `create_autospec`. Thanks for catching this.

```python
from unittest.mock import create_autospec

class A:

    async def __call__(self):
        await asyncio.sleep(1)

a = create_autospec(A)
b = a()
print(a)
print(b())
```

```
<MagicMock spec='A' id='4410501232'>
<MagicMock name='mock()()' id='4416843600'>
```

mariocj89: Totally fair indeed.

Misc/NEWS.d/next/Library/2019-05-26-01-20-06.bpo-37047.K9epi8.rst
```diff
@@ -0,0 +1,2 @@
+Handle late binding and attribute access in AsyncMock setup for
+autospeccing. Document newly implemented async methods in MagicMock.
```
mariocj89: Note: If you want to make this a link you can convert the AsyncMock and MagicMock to
:class:`unittest.mock.AsyncMock` and :class:`unittest.mock.MagicMock`

mariocj89 (approved):

Misc/NEWS.d/next/Library/2019-05-26-01-20-06.bpo-37047.K9epi8.rst
tirkarthi: Done. Thanks.

mariocj89 (approved):

1st1 (approved):

1st1:
@tirkarthi Do you want me or Andrew to merge this?

tirkarthi:
@1st1 Yes, this can be merged and all the items listed in the issue are complete. Thanks for the review everyone.

1st1:
Thank you, @tirkarthi!

PR #13576, opened 2019-05-25
bpo-37039: IDLE - zoomheight fixes

terryjreedy:
Move doc entry to match menu; refactor def zoom_height and adjust bottom margin for
Windows 10.

https://bugs.python.org/issue37039


terryjreedy:
Will add blurb when think ready to merge.

terryjreedy:
News will be part of followup patch.

PR #13578, opened 2019-05-26
bpo-22385: Support output separators in hex methods.

gpshead:
Also in binascii.hexlify aka b2a_hex.

The underlying implementation behind all hex generation in CPython uses the
same pystrhex.c implementation.  This adds support to bytes, bytearray,
and memoryview objects.

The binascii module functions exist rather than being slated for deprecation
because they return bytes rather than requiring an intermediate step through a
str object.

This change was inspired by MicroPython which supports sep in its binascii
implementation (and does not yet support the .hex methods).

https://bugs.python.org/issue22385


gpshead:
and now my branch is messed up.  WTF git.

Doc/library/binascii.rst
```diff
@@ -155,6 +155,24 @@ The :mod:`binascii` module defines the following functions:
    Similar functionality (but returning a text string) is also conveniently
    accessible using the :meth:`bytes.hex` method.
 
+   If *sep* is specified, it must be a single character str or bytes object.
+   It will be inserted in the output after every *bytes_per_sep* input bytes.
+   Separator placement is counted from the right end of the output by default,
+   if you wish to count from the left, supply a negative *bytes_per_sep* value.
+
+      >>> import binascii
```
tirkarthi, line 163:
```suggestion
     >>> import binascii
```

Seems to be misaligned with below set of >>>.

Doc/library/stdtypes.rst
```diff
@@ -2404,8 +2404,26 @@ data and are closely related to string objects in a variety of other ways.
       >>> b'\xf0\xf1\xf2'.hex()
       'f0f1f2'
 
+      If you want to make the hex string easier to read, you can specify a
+      single character separator *sep* parameter to include in the output.
+      By default between each byte.  A second optional *bytes_per_sep*
+      parameter controls the spacing.  Positive values calculate the
+      separator position from the right, negative values from the left.
+
+      >>> value = b'\xf0\xf1\xf2'
```
tirkarthi:
```suggestion
     >>> value = b'\xf0\xf1\xf2'
```

tirkarthi: Sorry, looking at the above examples it seems the other statements have to be shifted right by one space so that they are grouped under hex method like the rest. Please ignore my suggestion.

Doc/library/binascii.rst
gpshead: I don't see that in my files.

tirkarthi: Can confirm locally that there is no extra space here and below docs. Could be something wrong with GitHub rendering. Sorry for the noise.

PR #13579, opened 2019-05-26
bpo-36853: Fix suspicious.py to actually print the unused rules

asottile:
There were two issues here:
- `self.warn` isn't a thing in modern sphinx
- `logger.warning` terminates immediately and then doesn't print the info lines below it

https://bugs.python.org/issue36853

### DIFF
--- a/Doc/tools/extensions/suspicious.py
+++ b/Doc/tools/extensions/suspicious.py
@@ -153,14 +153,15 @@ class CheckSuspiciousMarkupBuilder(Builder):
         self.any_issue = True
         self.write_log_entry(lineno, issue, text)
         if py3:
-            self.logger.warn('[%s:%d] "%s" found in "%-.120s"' %
-                             (self.docname, lineno, issue, text))
+            self.logger.warning('[%s:%d] "%s" found in "%-.120s"' %
+                                (self.docname, lineno, issue, text))
         else:
-            self.logger.warn('[%s:%d] "%s" found in "%-.120s"' % (
-                self.docname.encode(sys.getdefaultencoding(),'replace'),
-                lineno,
-                issue.encode(sys.getdefaultencoding(),'replace'),
-                text.strip().encode(sys.getdefaultencoding(),'replace')))
+            self.logger.warning(
+                '[%s:%d] "%s" found in "%-.120s"' % (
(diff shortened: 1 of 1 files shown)


Doc/tools/extensions/suspicious.py
```diff
@@ -115,10 +115,12 @@ def write_doc(self, docname, doctree):
     def finish(self):
         unused_rules = [rule for rule in self.rules if not rule.used]
         if unused_rules:
-            self.warn('Found %s/%s unused rules:' %
```
tirkarthi: Does similar change of using `self.logger.warning` also need to be done in `report_issue` method?

* https://github.com/python/cpython/blob/aaf47caf35984e614d93bd8bea5227df55e0e3e6/Doc/tools/extensions/suspicious.py#L154
* https://github.com/python/cpython/blob/aaf47caf35984e614d93bd8bea5227df55e0e3e6/Doc/tools/extensions/suspicious.py#L157

asottile: probably!  let me fix those too -- thanks for pointing them out 👍

Carreau (approved):

tirkarthi (approved):
LGTM. Thanks.

tirkarthi:
@JulienPalard Would be helpful to have a review of this. This fixes the error reported in https://bugs.python.org/issue36853. Thanks.

jaraco (approved):

drj11:
Ah, cool. I came here (independently) to report this bug, but found this PR that fixes it instead. :)

asottile:
seems @zooba also applied a similar patch, but used the deprecated `logger.warn` method

I've rebased this to resolve the conflicts

asottile:
In 60419a7e96577cf783b3b45bf3984f9fb0d7ddff

asottile:
@jaraco anything left to do here?

ned-deily:
This PR seems to have stalled. @JulienPalard, can you take a look please?

jaraco:
Right. No need to leave this to languish longer.

jaraco:
@asottile Would you be willing to look at the cherry-pick? Thanks.

asottile:
yep! will do

PR #13580, opened 2019-05-26
bpo-29699: shutil.rmtree should not fail with FileNotFoundError (race condition)

websurfer5:
Ignore FileNotFoundError exceptions in shutil.rmtree() when removing a directory tree because some other process may have already deleted the directory entry for us after we started scanning directory entries (we don't care because we were going to delete it anyways).

https://bugs.python.org/issue29699


auvipy (approved):

Lib/shutil.py
```diff
@@ -701,6 +701,8 @@ def onerror(*args):
             os.close(fd)
     else:
         try:
+            if not os.path.exists(path):
+                raise FileNotFoundError("cannot call rmtree on a missing path")
```
mangrisano: I think it would be better replace 'cannot' with 'Cannot' and add a full stop after 'path'. Thus:

`Cannot call rmtree on a missing path.`

websurfer5: I updated the capitalization but not the punctuation to remain consistent with the other error messages in the function.

jaraco (changes requested):
In the last couple of commits, I've attempted to resolve the conflicts in the full spirit of the original contribution.

A major concern I have about this change is that there are no unit tests capturing this behavior, meaning that there are untested code paths and that this logic could be removed without breaking the tests. Our best hope is to reason about the implementation and hope our reasoning is sound.

Is there a way we could simulate the failures that inspired this contribution in a test or series of tests to avoid these shortcomings?

csabella:
@websurfer5, please address the questions raised in the code review.  Thank you!

the-knights-who-say-ni:

## CLA Missing

@mrkn

CoolCat467:
@websurfer5 "CLA not signed" tag is still here even though you did because "mrkn" committed but didn't also sign CLA

mrkn:
@CoolCat467 @the-knights-who-say-ni I'm sorry for the too-late response.  I couldn't notice the comment one year ago. 
 I've finished signing CLA.

jaraco:
Given this PR is over two years old and there's been no response to my request to add tests, I'm closing. Feel free to revisit and resubmit or request to re-open this PR if that's helpful.

PR #13335, opened 2019-05-15
[3.7] bpo-27987: align PyGC_Head to alignof(long double)

methane:
https://bugs.python.org/issue27987

PR #13581, opened 2019-05-26
[3.7] bpo-27987: align PyGC_Head to alignof(long double) (GH-13335)

methane:
Reverts python/cpython#13569

#13318 fixed the issue happened by merging #13335.

### DIFF
--- a/Include/objimpl.h
+++ b/Include/objimpl.h
@@ -255,7 +255,11 @@ typedef union _gc_head {
         union _gc_head *gc_prev;
         Py_ssize_t gc_refs;
     } gc;
-    double dummy;  /* force worst-case alignment */
+    long double dummy;  /* force worst-case alignment */
+    // malloc returns memory block aligned for any built-in types and
+    // long double is the largest standard C type.
+    // On amd64 linux, long double requires 16 byte alignment.
+    // See bpo-27987 for more discussion.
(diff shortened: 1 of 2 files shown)


gpshead:
I'm going to test this out with PGO and ubsan builds first.  I agree that the obmalloc alignment leading to the ubsan issues is probably fixed by #13318 at this point.

What I'm missing is _why_ this is even necessary at this point given the obmalloc alignment change?  What does this change actually cause to happen?

methane:
> What I'm missing is _why_ this is even necessary at this point given the obmalloc alignment change? What does this change actually cause to happen?

Alignment is all or nothing.  If this is not aligned, no need to align pointer returned by obmalloc.
See this figure.

alignment

#13318 make pointer PyObject_Malloc returned (a) aligned to 16 byte.  Now `PyGC_Head *gc` is aligned to 16 byte.

This pull request make sizeof(PyGC_Head) (b) aligned to 16 byte. Now `PyObject *o = (PyObject*)(gc+1)` is aligned to 16 byte.

C compiler align (c) part automatically.  Now member of PyObject is aligned.

gpshead:
thanks, that was helpful :)  also, testing this on an ubsan build it shows a notable improvement.  the two changes together work.

PR #13582, opened 2019-05-26
bpo-31163: Added return values to pathlib.Path instance's rename and replace methods.

shangdahao:
After a rename/replace the Path instance keeps "pointing" to the previous path, and does not return the new path, which is a little misunderstand. This patch make Path instance's rename and replace methods return the new Path instance.

https://bugs.python.org/issue31163

### DIFF
--- a/Doc/library/pathlib.rst
+++ b/Doc/library/pathlib.rst
@@ -933,23 +933,26 @@ call fails (for example because the path doesn't exist).
 
 .. method:: Path.rename(target)
 
-   Rename this file or directory to the given *target*.  On Unix, if
-   *target* exists and is a file, it will be replaced silently if the user
-   has permission.  *target* can be either a string or another path object::
+   Rename this file or directory to the given *target*, and return a new Path
+   instance pointing to *target*.  On Unix, if *target* exists and is a file,
+   it will be replaced silently if the user has permission.  *target* can be
+   either a string or another path object::
 
       >>> p = Path('foo')
       >>> p.open('w').write('some text')
       9
       >>> target = Path('bar')
       >>> p.rename(target)
+      PosixPath('bar')
       >>> target.open().read()
       'some text'
 
 
 .. method:: Path.replace(target)
 
-   Rename this file or directory to the given *target*.  If *target* points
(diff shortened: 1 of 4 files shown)


Doc/library/pathlib.rst
```diff
@@ -933,23 +933,26 @@ call fails (for example because the path doesn't exist).
 
 .. method:: Path.rename(target)
 
-   Rename this file or directory to the given *target*.  On Unix, if
-   *target* exists and is a file, it will be replaced silently if the user
-   has permission.  *target* can be either a string or another path object::
+   Rename this file or directory to the given *target*, and return a new Path
```
tirkarthi, line 936: Please add a `versionchanged` directive to note the change in return value behavior.

shangdahao: Done, thanks tirkarthi.

asvetlov:
I'd like to get @pitrou review for `pathlib` changes.

jaraco (approved):
~Antoine is no longer reviewing pathlib~, but I agree the approach here seems reasonable as does @brettcannon.

Doc/library/pathlib.rst
```diff
@@ -933,23 +933,32 @@ call fails (for example because the path doesn't exist).
...
+   instance pointing to *target*.  On Unix, if *target* exists and is a file,
+   it will be replaced silently if the user has permission.  *target* can be
+   either a string or another path object::
 
       >>> p = Path('foo')
       >>> p.open('w').write('some text')
       9
       >>> target = Path('bar')
       >>> p.rename(target)
+      PosixPath('bar')
```
jaraco, line 946: Hmm. I wonder if this docstring will fail doctests on Windows (or violate Windows users' expectations). @zooba can you comment?

pitrou:
Sounds fine on the principle, in any case.

Doc/library/pathlib.rst
jaraco: On deeper inspection, I see that `PosixPath` is what the docs above use, so this string is the most consistent with the current behavior. If these docs are tested, they're not tested on Windows.

PR #13583, opened 2019-05-26
bpo-37053: handle strings like u"bar" correctly in Tools/parser/unparse.py

yan12125:
Constant.kind is added in https://bugs.python.org/issue36280.
Current possible values for Constant.kind are "u" or None.

For r'bar' and b'bar', Constant.kind value is None, so there's no need
for special handling.

https://bugs.python.org/issue37053


pablogsal (approved):
LGTM

Confirmed that this fixes the broken buildbots:

./python -m test test_tools -vvvv -uall -m test_files
== CPython 3.8.0a4+ (heads/master:91f4380ced, May 26 2019, 17:30:21) [GCC 8.3.0]
== Linux-5.0.15-1-MANJARO-x86_64-with-glibc2.29 little-endian
== cwd: /home/pablogsal/github/cpython/build/test_python_25649
== CPU count: 12
== encodings: locale=UTF-8, FS=utf-8
Run tests sequentially
0:00:00 load avg: 0.50 [1/1] test_tools
test_files (test.test_tools.test_unparse.DirectoryTestCase) ... Testing /home/pablogsal/github/cpython/Lib/typing.py
Testing /home/pablogsal/github/cpython/Lib/test/test_typing.py
ok

Ran 1 test in 0.586s

OK

== Tests result: SUCCESS ==

1 test OK.

Total duration: 671 ms
Tests result: SUCCESS


The manual check is needed because the CI in GitHub does not run `test_tools` over all files so it does not take forever.

pablogsal (changes requested):
[Comment deleted as it was tested on a buildbot with a corrupted checkout]

pablogsal (approved):

ilevkivskyi (approved):

yan12125:
Thanks for reviewing!

pablogsal:
@yan12125 Thanks for the PR and the report! :)

PR #13585, opened 2019-05-26
bpo-37058: PEP 544: Add Protocol to typing module

ilevkivskyi:
cc @ambv @JukkaL

I tried to get rid of the `_ProtocolMeta`, but unfortunately it didn'y work. My idea to return a generic alias from `@runtime_checkable` made runtime protocols unpickleable. I am not sure what is worse (a custom metaclass or having some classes unpickleable), so I decided to stick with the status quo (since there were no complains so far). So essentially this is a copy of the implementation in `typing_extensions` with two modifications:
* Rename `@runtime` to `@runtime_checkable` (plus corresponding updates).
* Allow protocols that extend `collections.abc.Iterable` etc.

https://bugs.python.org/issue37058

### DIFF
--- a/Doc/library/typing.rst
+++ b/Doc/library/typing.rst
@@ -17,7 +17,8 @@
 
 --------------
 
-This module supports type hints as specified by :pep:`484` and :pep:`526`.
+This module supports type hints as specified by :pep:`484`, :pep:`526`,
+:pep:`544`, :pep:`586`, :pep:`589`, and :pep:`591`.
 The most fundamental support consists of the types :data:`Any`, :data:`Union`,
 :data:`Tuple`, :data:`Callable`, :class:`TypeVar`, and
 :class:`Generic`.  For full specification please see :pep:`484`.  For
@@ -392,6 +393,48 @@ it as a return value) of a more specialized type is a type error. For example::
 Use :class:`object` to indicate that a value could be any type in a typesafe
 manner. Use :data:`Any` to indicate that a value is dynamically typed.
 
+
+Nominal vs structural subtyping
+-------------------------------
+
+Initially :pep:`484` defined Python static type system as using
+*nominal subtyping*. This means that a class ``A`` is allowed where
+a class ``B`` is expected if and only if ``A`` is a subclass of ``B``.
+
+This requirement previously also applied to abstract base classes, such as
+:class:`Iterable`. The problem with such approach is that a class had
+to be explicitly marked to support them, which is unpythonic and unlike
+what one would normally do in idiomatic dynamically typed Python code.
+For example, this conforms to the :pep:`484`::
+
+   from typing import Sized, Iterable, Iterator
+
+   class Bucket(Sized, Iterable[int]):
+       ...
+       def __len__(self) -> int: ...
+       def __iter__(self) -> Iterator[int]: ...
+
+The :pep:`544` allows to solve this problem by allowing users to write
+the above code without explicit base classes in the class definition,
+allowing ``Bucket`` to be implicitly considered a subtype of both ``Sized``
+and ``Iterable[int]`` by static type checkers. This is known as
+*structural subtyping* (or static duck-typing)::
+
+   from typing import Iterator, Iterable
+
+   class Bucket:  # Note: no base classes
+       ...
+       def __len__(self) -> int: ...
+       def __iter__(self) -> Iterator[int]: ...
+
+   def collect(items: Iterable[int]) -> int: ...
+   result: int = collect(Bucket())  # Passes type check
+
+Moreover, by subclassing a special class :class:`Protocol`, a user
+can define new custom protocols to fully enjoy structural subtyping
+(see examples below).
+
+
 Classes, functions, and decorators
 ----------------------------------
 
@@ -480,7 +523,7 @@ The module defines the following classes, functions and decorators:
       func(C())  # Passes static type check
 
    See :pep:`544` for details. Protocol classes decorated with
-   :decorator:`runtime_checkable` act as simple-minded runtime protocols that
+   :func:`runtime_checkable` act as simple-minded runtime protocols that
    check only the presence of given attributes, ignoring their type signatures.
    Protocol classes can be generic, they are defined as::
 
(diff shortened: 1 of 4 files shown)


Doc/library/typing.rst
```diff
@@ -17,7 +17,8 @@
 
 --------------
 
-This module supports type hints as specified by :pep:`484` and :pep:`526`.
+This module supports type hints as specified by :pep:`484`, :pep:`526`,
```
gvanrossum: "This module provides runtime support for type hints ..."

Doc/library/typing.rst
```diff
@@ -392,6 +393,48 @@ it as a return value) of a more specialized type is a type error. For example::
...
+
+Nominal vs structural subtyping
+-------------------------------
+
+Initially :pep:`484` defined Python static type system as using
+*nominal subtyping*. This means that a class ``A`` is allowed where
+a class ``B`` is expected if and only if ``A`` is a subclass of ``B``.
+
+This requirement previously also applied to abstract base classes, such as
+:class:`Iterable`. The problem with such approach is that a class had
```
gvanrossum: such -> this

Doc/library/typing.rst
```diff
@@ -392,6 +393,48 @@ it as a return value) of a more specialized type is a type error. For example::
...
+For example, this conforms to the :pep:`484`::
+
+   from typing import Sized, Iterable, Iterator
+
+   class Bucket(Sized, Iterable[int]):
+       ...
+       def __len__(self) -> int: ...
+       def __iter__(self) -> Iterator[int]: ...
+
+The :pep:`544` allows to solve this problem by allowing users to write
```
gvanrossum: There's no "The" before "PEP NNN". (Essentially, "PEP NNN" is treated as a name, while just "PEP" is treated as a noun.)

Doc/library/typing.rst
```diff
@@ -459,6 +502,37 @@ The module defines the following classes, functions and decorators:
           except KeyError:
               return default
 
+.. class:: Protocol(Generic)
+
+   Base class for protocol classes. Protocol classes are defined as::
```
gvanrossum: as -> like this

Doc/library/typing.rst
```diff
@@ -459,6 +502,37 @@ The module defines the following classes, functions and decorators:
...
+      class C:
+          def meth(self) -> int:
+              return 0
+
+      def func(x: Proto) -> int:
+          return x.meth()
+
+      func(C())  # Passes static type check
+
+   See :pep:`544` for details. Protocol classes decorated with
```
gvanrossum, line 525: This is okay for now, but at some point we may want to make it so that the docs are self-contained without any references to PEPs. (This is so that in the future the implementation can evolve, and the docs should describe the current implementation, while the PEP describes the implementation as it was at the time the PEP was finalized.)

Doc/library/typing.rst
```diff
@@ -459,6 +502,37 @@ The module defines the following classes, functions and decorators:
...
+   :func:`runtime_checkable` act as simple-minded runtime protocols that
+   check only the presence of given attributes, ignoring their type signatures.
+   Protocol classes can be generic, they are defined as::
```
gvanrossum: Insert blank line before, and change "they are defined as" to "for example"

Doc/library/typing.rst
```diff
@@ -459,6 +502,37 @@ The module defines the following classes, functions and decorators:
...
+          def meth(self) -> int:
+              return 0
+
+      def func(x: Proto) -> int:
+          return x.meth()
+
+      func(C())  # Passes static type check
+
+   See :pep:`544` for details. Protocol classes decorated with
+   :func:`runtime_checkable` act as simple-minded runtime protocols that
```
gvanrossum: Is there no `:decorator:` thing you can use? Also maybe insert "(described later)" since this isn't the full description of runtime_checkable, just a summary of features?

Doc/library/typing.rst
```diff
@@ -1033,6 +1107,26 @@ The module defines the following classes, functions and decorators:
    Note that returning instances of private classes is not recommended.
    It is usually preferable to make such classes public.
 
+.. decorator:: runtime_checkable
+
+   Mark a protocol class as a runtime protocol.
+
+   Such protocol can be used with :func:`isinstance` and :func:`issubclass`.
```
gvanrossum: Such -> Such a

Doc/library/typing.rst
```diff
@@ -1033,6 +1107,26 @@ The module defines the following classes, functions and decorators:
...
+   Raise :exc:`TypeError` if applied to a non-protocol class.  This allows
```
gvanrossum: Raise -> This raises; if -> when

Doc/library/typing.rst
```diff
@@ -1033,6 +1107,26 @@ The module defines the following classes, functions and decorators:
...
+   a simple-minded structural check very similar to one trick ponies in
```
gvanrossum: comma after "check", and put quotes around "one trick ponies" (it's a rather colloquial expression).

gvanrossum (commented):
I didn't really review the code, I trust it's just a copy of what's in the external typing.py or typing_extensions.py.

Doc/library/typing.rst
ilevkivskyi: Yes, we can expand the docs later.

Doc/library/typing.rst
ilevkivskyi: > Is there no `:decorator:` thing you can use?

I think unfortunately no. At least I didn't find `:deco:` or anything similar (this would be something good to have).

ilevkivskyi:
> I trust it's just a copy of what's in the external typing.py or typing_extensions.py.

To be precise it is not literally a copy. I added some dcostrings, few more comments, and factored-out couple pieces of code into private helpers. The tests however are 1:1 copied, so there are no semantic changes.

ilevkivskyi:
@gvanrossum Thanks for review! (Also thank you for reviewing the other PEP PRs!)

gvanrossum (approved):

PR #13586, opened 2019-05-26
bpo-36686: Improve the documentation of the std* params in loop.subprocess_exec

sbstp:
https://bugs.python.org/issue36686

### DIFF
--- a/Doc/library/asyncio-eventloop.rst
+++ b/Doc/library/asyncio-eventloop.rst
@@ -1217,32 +1217,52 @@ async/await code consider using the high-level
 
    Other parameters:
 
-   * *stdin*: either a file-like object representing a pipe to be
-     connected to the subprocess's standard input stream using
-     :meth:`~loop.connect_write_pipe`, or the
-     :const:`subprocess.PIPE`  constant (default). By default a new
-     pipe will be created and connected.
+   * *stdin* can be any of these:
+   
+     * a file-like object representing a pipe to be connected to the
+       subprocess's standard input stream using 
+       :meth:`~loop.connect_write_pipe`
+     * the :const:`subprocess.PIPE` constant (default) which will create a new
+       pipe and connect it,
+     * the value ``None`` which will make the subprocess inherit the file
+       descriptor from this process
+     * the :const:`subprocess.DEVNULL` constant which indicates that the
+       special :data:`os.devnull` file will be used
 
-   * *stdout*: either a file-like object representing the pipe to be
-     connected to the subprocess's standard output stream using
-     :meth:`~loop.connect_read_pipe`, or the
-     :const:`subprocess.PIPE` constant (default). By default a new pipe
-     will be created and connected.
+   * *stdout* can be any of these:
 
-   * *stderr*: either a file-like object representing the pipe to be
-     connected to the subprocess's standard error stream using
-     :meth:`~loop.connect_read_pipe`, or one of
-     :const:`subprocess.PIPE` (default) or :const:`subprocess.STDOUT`
-     constants.
+     * a file-like object representing a pipe to be connected to the
+       subprocess's standard output stream using 
+       :meth:`~loop.connect_write_pipe`
+     * the :const:`subprocess.PIPE` constant (default) which will create a new
+       pipe and connect it,
+     * the value ``None`` which will make the subprocess inherit the file
+       descriptor from this process
+     * the :const:`subprocess.DEVNULL` constant which indicates that the
+       special :data:`os.devnull` file will be used
 
-     By default a new pipe will be created and connected. When
-     :const:`subprocess.STDOUT` is specified, the subprocess' standard
-     error stream will be connected to the same pipe as the standard
-     output stream.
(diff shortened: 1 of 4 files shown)


Doc/library/asyncio-eventloop.rst
```diff
@@ -1217,32 +1217,52 @@ async/await code consider using the high-level
...
+       error stream to the process' standard output stream
 
    * All other keyword arguments are passed to :class:`subprocess.Popen`
-     without interpretation, except for *bufsize*, *universal_newlines*
-     and *shell*, which should not be specified at all.
+     without interpretation, except for *bufsize*, *universal_newlines*,
+     *shell*, *text*, *encoding* and *errors*, which should not be specified
+     at all.
+
+     The ``asyncio`` subprocess API does not yet support decoding the streams
```
asvetlov: Let's not make assumptions the future in docs.
I mean `subprocess API does not yet support` should be replaced with just `subprocess API does not support`.
I'm open for discussion about text mode support though, please create a separate ussue for that.

asvetlov (changes requested):
The PR is for documentation update, that's nice.
`base_events.py` in `subprocess_exec` and `subprocess_shell` explicitly forbids non-default `universal_newlines`, `shell` and `bufsize`.
It would be nice if `text`, `encoding` and `errors` arguments are explicitly forbidden too by code

sbstp:
> It would be nice if `text`, `encoding` and `errors` arguments are explicitly forbidden too by code

I don't mind writing the code for that, but it seems like there's a chance it will cause breakage if some programs used those params.

1st1 (approved):

1st1:
Please use `blurb` to generate a NEWS entry.

Lib/asyncio/base_events.py
```diff
@@ -1570,6 +1571,10 @@ def _log_subprocess(self, msg, stdin, stdout, stderr):
             raise ValueError("shell must be True")
         if bufsize != 0:
             raise ValueError("bufsize must be 0")
+        if encoding or errors or text:
```
asvetlov: Please split this `if` into three separate checks. It makes error messages more specific.

Lib/asyncio/base_events.py
```diff
@@ -1586,13 +1591,19 @@ def _log_subprocess(self, msg, stdin, stdout, stderr):
...
+                              shell=False, bufsize=0,
+                              encoding=None, errors=None, text=None,
+                              **kwargs):
         if universal_newlines:
             raise ValueError("universal_newlines must be False")
         if shell:
             raise ValueError("shell must be False")
         if bufsize != 0:
             raise ValueError("bufsize must be 0")
+        if encoding or errors or text:
```
asvetlov: The same

asvetlov (changes requested):
Almost perfect! Please fix last nits

sbstp:
I have made the requested changes; please review again. :)

Lib/asyncio/base_events.py
```diff
@@ -1570,6 +1571,13 @@ def _log_subprocess(self, msg, stdin, stdout, stderr):
             raise ValueError("shell must be True")
         if bufsize != 0:
             raise ValueError("bufsize must be 0")
+        if text:
+            raise ValueError("text must be False")
+        if encoding:
```
asvetlov:
```suggestion
        if encoding is not None:
```

Lib/asyncio/base_events.py
```diff
@@ -1570,6 +1571,13 @@ def _log_subprocess(self, msg, stdin, stdout, stderr):
...
+            raise ValueError("encoding must be None")
+        if errors:
```
asvetlov:
```suggestion
        if errors is not None:
```

Lib/asyncio/base_events.py
```diff
@@ -1586,13 +1594,22 @@ def _log_subprocess(self, msg, stdin, stdout, stderr):
...
+                              **kwargs):
         if universal_newlines:
             raise ValueError("universal_newlines must be False")
         if shell:
             raise ValueError("shell must be False")
         if bufsize != 0:
             raise ValueError("bufsize must be 0")
+        if text:
+            raise ValueError("text must be False")
+        if encoding:
```
asvetlov:
```suggestion
        if encoding is not None:
```

Lib/asyncio/base_events.py
```diff
@@ -1586,13 +1594,22 @@ def _log_subprocess(self, msg, stdin, stdout, stderr):
...
+            raise ValueError("encoding must be None")
+        if errors:
```
asvetlov:
```suggestion
        if errors is not None:
```

asvetlov (changes requested):
Please use suggested strict checks to distinguish, e.g. empty string from `None`.
I know that empty string for `errors` is invalid currently but still...

asvetlov (approved):
Thanks!

sbstp:
Thanks for being patient with me!

PR #13599, opened 2019-05-27
Add @maxking to CODEOWNERS file

maxking:


eamanu (approved):

Mariatta:
Hmm we've never tried adding non core developers into `CODEOWNERS` yet, so I'm not sure if this will work. The effect of `codeowners` file is that you will be automatically requested review for the matched PR, and I don't know if we can request review from people outside of Python org.

But anyway, let's try this, and we can always revert the change.

Mariatta (approved):

maxking:
Thanks @Mariatta ! Now all I need to do is open a PR and see if it works ;-)

tirkarthi:
https://help.github.com/en/articles/about-code-owners

> People with admin or owner permissions can set up a CODEOWNERS file in a repository. The people you choose as code owners must have write permissions for the repository.

brettcannon:
I opened #13660 to revert this since @tirkarthi found that reference saying @maxking would need to have write access to get notifications.

Maybe once the new "triage" access is launched that will work with CODEOWNERS and then we can give @maxking that level of access.

PR #13601, opened 2019-05-27
bpo-37054, _pyio: Fix BytesIO and TextIOWrapper __del__()

vstinner:
Fix destructor _pyio.BytesIO and _pyio.TextIOWrapper: initialize
their _buffer attribute as soon as possible (in the class body),
because it's used by __del__() which calls close().

https://bugs.python.org/issue37054


pitrou (approved):
+1

PR #13603, opened 2019-05-27
bpo-33725: multiprocessing uses spawn by default on macOS

vstinner:
The multiprocessing module now uses the "spawn" start method by
default on macOS.

https://bugs.python.org/issue33725

### DIFF
--- a/Doc/library/multiprocessing.rst
+++ b/Doc/library/multiprocessing.rst
@@ -102,7 +102,7 @@ to start a process.  These *start methods* are
     will not be inherited.  Starting a process using this method is
     rather slow compared to using *fork* or *forkserver*.
 
-    Available on Unix and Windows.  The default on Windows.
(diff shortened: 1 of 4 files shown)


vstinner:
cc @warsaw @applio @pitrou

vstinner:
I didn't test my PR. I'm not a multiprocessing expert at all, so I'm not sure that it even does what it is supposed to do :-D

vstinner:
I tested my PR on macOS: it seems like it works as expected :-)

```
macbook:master haypo$ ./python.exe
Python 3.8.0a4+ (heads/pr/13603:d117d0f, May 28 2019, 14:37:04) 
[Clang 10.0.0 (clang-1000.10.44.4)] on darwin
>>> import multiprocessing; multiprocessing.get_context()
<multiprocessing.context.SpawnContext object at 0x1009c9550>
```

PR #13606, opened 2019-05-28
bpo-26423: Fix possible overflow in wrap_lenfunc()

ZackerySpytz:
https://bugs.python.org/issue26423


Misc/NEWS.d/next/Core and Builtins/2019-05-27-18-00-19.bpo-26423.RgUOE8.rst
```diff
@@ -0,0 +1 @@
+Fix possible overflow in ``wrap_lenfunc()``.
```
vstinner: Would you mind that the issue only affects systems where sizeof(long) < sizeof(size_t): only 64-bit Windows if I understand correctly.

ZackerySpytz:
@vstinner Thank you for your review. I have added more detail to the news entry.

PR #13610, opened 2019-05-28
[3.6] bpo-36520: Email header folded incorrectly (GH-13608)

websurfer5:
Folding long email headers that contain multiple UTF-8 words in the first line results in corruption when UTF-8 words are found in subsequent lines. The fix is to reset the variable tracking the encoded word location in the line when a new line is started.

https://bugs.python.org/issue36520


Mariatta:
Thanks for this PR. I'm closing this because the 3.6 branch is only accepting security fixes only. It doesn't accept bug fixes/new features anymore. Additionally, we will do the backport PR only after the PR on master has been reviewed and merged.

PR #13612, opened 2019-05-28
bpo-37029:  keep usable_arenas in sorted order without searching

tim-one:
This adds a vector of "search fingers" so that usable_arenas can be kept in sorted order (by number of free pools) via constant-time operations instead of linear search.

This should reduce worst-case time for reclaiming a great many objects from O(A**2) to O(A), where A is the number of arenas.

https://bugs.python.org/issue37029


Objects/obmalloc.c
```diff
@@ -1777,6 +1821,9 @@ pymalloc_free(void *ctx, void *p)
             usable_arenas->prevarena = ao;
         usable_arenas = ao;
         assert(usable_arenas->address != 0);
+        if (nfp2lasta[1] == NULL) {
+            nfp2lasta[1] = ao;
+        }
```
methane, line 1826: Is this needed?  Arn't only nfp2lasta[2 ... MAX_POOLS_IN_ARENA-1] used as fingers?

obmalloc does not have any hysteresis / buffer.  So heavy malloc / free loop can be
allocate and free a pool again and again:

```
for ... {
    buf = PyObject_Malloc(...);  // nfreepool become 0
    // ...
    PyObject_Free(buf);  // nfreepool become 1
}
```

I'm not sure about worst case speed yet.  But I want to minimize code and memory access
when nfreepool becomes 0 or 1.

tim-one: Yes, `nfp2lasta[1]` csn be used (also, on the other end, `nfp2lasta[MAX_POOLS_IN_ARENA]` is set for a brief time when a new arena is obtained from the system but before the first pool in it is carved out).

Suppose `usable_arenas` is empty, and many thousands of arenas have been fully allocated.  Call them arena #1, arena #2, ...  Then a pool is freed in arena #1, then in #2, then in #3, ..., and then in #2000.  Now we have 2000 arenas in `usable_arenas`, all with a single free pool, starting with arena #2000 at the head of the list and ending with arena #1.  The code you're asking about set `nfp2last1[1] = arena #1` on the first of those pool frees, and it stays that way.

Now when a _second_ pool is freed from arena #2000, its number of free pools increases to 2, and `nfp2last[1]` immediately tells us arena #2000 has to be moved from the head of `usable_arenas` to follow arena #1, nearly 2000 positions down the list.  Doing that without searching is, of course, the point of this patch.

The code you gave as an example doesn't show "a problem" anyway, at least not as is:  remember that obmalloc does _not_ return pools, it returns blocks, and every pool contains multiple blocks.  To get the kind of thrashing you have in mind, the loop would have to contain multiple `PyObject_Malloc()` calls followed by the same number of `PyObject_Free()` calls, and exactly the right number of calls to exactly exhaust a pool.

I agree "exactly the right number of oalls" may be one, if the arena just happens to have one block remaining in a single free pool at the loop start, but that has to be rare.  Especially because arenas have 64 pools to begin with, so you're compounding improbabilities to contrive a case :wink:.  Even then, without the patch it's going to incur the overheads of repeatedly unlinking and linking the arena from/to `usable_arenas` (etc), and the patch just adds a test-&-branch to those.  It adds some cycles, but doesn't change the asymptotics.

I'd be more concerned about, say, thrashing between 3 and 4 free pools, because that's cheaper today than thrashing between 0 and 1, and keeping `nfp2lasta` up to date adds some cycles to _all_ cases when the number of free pools changes.  The new overheads are a bigger percentage hit to the cases that are cheaper today.

methane (approved):

Objects/obmalloc.c
```diff
@@ -1172,6 +1189,9 @@ static struct arena_object* unused_arena_objects = NULL;
  */
```
pitrou, line 1189: Add comment documenting that they are sorted in number of free pools?

pitrou (commented):
Looks neat, just a comment.

Objects/obmalloc.c
tim-one: I'd rather not, unless you feel strongly about it.  This has been documented all along, but up where the _main_ explanation of arenas lives (look at the comment block about `usable_arenas` at line 1150).

There are a bunch of moving parts here, and the style of the file is to give long-ish English explanations in uninterrupted comment blocks, with only the briefest of memory-jogging comments on declaration lines.

PR #13616, opened 2019-05-28
bpo-37075: Fix string concatenation in assert_has_awaits error message

tirkarthi:
* Fix the implicit string concatenation in `assert_has_awaits` error message.
* Use "await" instead of "call" in `assert_awaited_with` error message.

https://bugs.python.org/issue37075


tirkarthi:
I feel this doesn't require NEWS entry since this is change in error message for unreleased code. If needed I can add one.

asvetlov (approved):

tirkarthi:
Thanks @asvetlov :)

PR #13617, opened 2019-05-28
bpo-37076: _thread.start_new_thread() calls _PyErr_WriteUnraisableMsg()

vstinner:
_thread.start_new_thread() now logs uncaught exception raised by the
function using sys.unraisablehook(), rather than sys.excepthook(), so
the hook gets access to the function which raised the exception.

https://bugs.python.org/issue37076

### DIFF
--- a/Doc/library/_thread.rst
+++ b/Doc/library/_thread.rst
@@ -43,12 +43,22 @@ This module defines the following constants and functions:
 
 .. function:: start_new_thread(function, args[, kwargs])
 
-   Start a new thread and return its identifier.  The thread executes the function
-   *function* with the argument list *args* (which must be a tuple).  The optional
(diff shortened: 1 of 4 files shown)


vstinner:
See https://bugs.python.org/issue37076 for my choice of sys.unraisablehook() rather than threading.excepthook().

vstinner:
My PR changes the error message:

```
Unhandled exception in thread started by {func!r}"
Traceback ...
```

becomes:

```
Exception ignored in thread started by: {func!r}"
Traceback ...
```

Technically, _PyErr_WriteUnraisableMsg() internals would allow to override the whole error message. We chose to hardcode "Exception ignored" prefix. IMHO it's a good thing that error messages are more similar. So changing the error message is a good thing.

serhiy-storchaka (approved):

PR #6375, opened 2018-04-04
bpo-24882: Let ThreadPoolExecutor reuse idle threads before creating new thread

iUnknwn:
The change fixes an issue where ThreadPoolExecutor doesn't check if existing threads are idle before spinning up new ones. To do this, we use a simple counter within the Executor to track how many threads are idle, and atomically increment and decrement the count as needed. 

One question - the previous code to spin up a new thread in `_adjust_thread_count` does not appear thread safe - if two threads are both submitting items to the executor at the same time, the call to check the number of threads could be invalid. Is this something that should also be fixed, or is this not an issue because of the global interpreter lock?

https://bugs.python.org/issue24882

PR #13618, opened 2019-05-28
bpo-35279: reduce default max_workers of ThreadPoolExecutor

methane:
https://bugs.python.org/issue35279

### DIFF
--- a/Doc/library/concurrent.futures.rst
+++ b/Doc/library/concurrent.futures.rst
@@ -161,12 +161,12 @@ And::
 
    .. versionchanged:: 3.8
       Default value of *max_workers* is changed to ``min(32, os.cpu_count() + 4)``.
-      This default value preserves at least 5 worker for I/O bound task.  It utilize
-      at most 32 CPU cores for CPU bound task which release GIL.  And it avoid using
-      very large resource implicitly on many core machines.
+      This default value preserves at least 5 workers for I/O bound tasks.
+      It utilizes at most 32 CPU cores for CPU bound tasks which release the GIL.
+      And it avoids using very large resources implicitly on many-core machines.
 
-      ThreadPoolExecutor now reuses idle worker threads before starting *max_workers*
-      worker threads too.
+      ThreadPoolExecutor now reuses idle worker threads before starting
+      *max_workers* worker threads too.
 
 
 .. _threadpoolexecutor-example:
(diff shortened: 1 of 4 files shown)


asvetlov (approved):
LGTM!

Doc/library/concurrent.futures.rst
```diff
@@ -159,6 +159,15 @@ And::
    .. versionchanged:: 3.7
       Added the *initializer* and *initargs* arguments.
 
+   .. versionchanged:: 3.8
+      Default value of *max_workers* is changed to ``min(32, os.cpu_count() + 4)``.
+      This default value preserves at least 5 worker for I/O bound task.  It utilize
```
tirkarthi: Not a native English speaker but I think it should be "It utilizes" and "And it avoids" in the next sentence.

csabella: Also, `at least 5 workers`.  And I'm not sure if it should be `for I/O bound tasks` or 'for the I/O bound task`.

Doc/library/concurrent.futures.rst
```diff
@@ -159,6 +159,15 @@ And::
...
+      at most 32 CPU cores for CPU bound task which release GIL.  And it avoid using
```
csabella: `for CPU bound tasks` and `which release the GIL`.

Doc/library/concurrent.futures.rst
```diff
@@ -159,6 +159,15 @@ And::
...
+      very large resource implicitly on many core machines.
```
csabella: s/resource/resources

Doc/library/concurrent.futures.rst
```diff
@@ -159,6 +159,15 @@ And::
    .. versionchanged:: 3.7
       Added the *initializer* and *initargs* arguments.
 
+   .. versionchanged:: 3.8
+      Default value of *max_workers* is changed to ``min(32, os.cpu_count() + 4)``.
+      This default value preserves at least 5 workers for I/O bound tasks.
+      It utilizes at most 32 CPU cores for CPU bound tasks which release the GIL.
+      And it avoids using very large resources implicitly on many-core machines.
+
+      ThreadPoolExecutor now reuses idle worker threads before starting
```
wyaadarsh, line 168: Where is the corresponding change in the code? For reusing the idle worker threads before starting *max_workers* worker threads?

methane: https://github.com/python/cpython/pull/6375

PR #13624, opened 2019-05-28
bpo-37077: Add native thread ID (TID) for AIX

aixtools:
This is the followup  for issue36084

https://bugs.python.org/issue37077

### DIFF
--- a/Doc/library/threading.rst
+++ b/Doc/library/threading.rst
@@ -355,7 +355,7 @@ since it is impossible to detect the termination of alien threads.
          system-wide) from the time the thread is created until the thread
          has been terminated.
 
-      .. availability:: Windows, FreeBSD, Linux, macOS, OpenBSD, NetBSD, AIX.
+      .. availability:: Require :func:`get_native_id` function.
 
       .. versionadded:: 3.8
 
--- a/Doc/library/threading.rst
+++ b/Doc/library/threading.rst
@@ -82,7 +82,7 @@ This module defines the following functions:
    Its value may be used to uniquely identify this particular thread system-wide
    (until the thread terminates, after which the value may be recycled by the OS).
 
-   .. availability:: Windows, FreeBSD, Linux, macOS, OpenBSD.
+   .. availability:: Windows, FreeBSD, Linux, macOS, OpenBSD, AIX.
 
    .. versionadded:: 3.8
 
@@ -355,7 +355,7 @@ since it is impossible to detect the termination of alien threads.
          system-wide) from the time the thread is created until the thread
          has been terminated.
 
-      .. availability:: Require :func:`get_native_id` function.
+      .. availability:: Windows, FreeBSD, Linux, macOS, OpenBSD, AIX.
 
       .. versionadded:: 3.8
 
--- a/Misc/NEWS.d/next/Core and Builtins/2019-05-28-11-47-44.bpo-37077.S1h0Fc.rst
+++ b/Misc/NEWS.d/next/Core and Builtins/2019-05-28-11-47-44.bpo-37077.S1h0Fc.rst
@@ -1,2 +1,2 @@
-Add native thread ID (TID) for AIX
+Add :func:`threading.get_native_id` support for AIX.
 Patch by M. Felt
--- a/Misc/NEWS.d/next/Core and Builtins/2019-05-28-11-47-44.bpo-37077.S1h0Fc.rst
+++ b/Misc/NEWS.d/next/Core and Builtins/2019-05-28-11-47-44.bpo-37077.S1h0Fc.rst
@@ -1,0 +1,2 @@
+Add native thread ID (TID) for AIX
+Patch by M. Felt
--- a/Doc/library/_thread.rst
+++ b/Doc/library/_thread.rst
@@ -106,7 +106,7 @@ This module defines the following constants and functions:
    Its value may be used to uniquely identify this particular thread system-wide
    (until the thread terminates, after which the value may be recycled by the OS).
 
-   .. availability:: Windows, FreeBSD, Linux, macOS, OpenBSD.
+   .. availability:: Windows, FreeBSD, Linux, macOS, OpenBSD, AIX.
 
    .. versionadded:: 3.8
 
--- a/Include/pythread.h
+++ b/Include/pythread.h
@@ -26,7 +26,7 @@ PyAPI_FUNC(unsigned long) PyThread_start_new_thread(void (*)(void *), void *);
 PyAPI_FUNC(void) _Py_NO_RETURN PyThread_exit_thread(void);
 PyAPI_FUNC(unsigned long) PyThread_get_thread_ident(void);
 
-#if defined(__APPLE__) || defined(__linux__) || defined(__FreeBSD__) || defined(__OpenBSD__) || defined(_WIN32)
+#if defined(__APPLE__) || defined(__linux__) || defined(__FreeBSD__) || defined(__OpenBSD__) || defined(_WIN32) || defined(_AIX)
 #define PY_HAVE_THREAD_NATIVE_ID
 PyAPI_FUNC(unsigned long) PyThread_get_thread_native_id(void);
 #endif
--- a/Python/thread_pthread.h
+++ b/Python/thread_pthread.h
@@ -18,6 +18,8 @@
 #   include <pthread_np.h>      /* pthread_getthreadid_np() */
 #elif defined(__OpenBSD__)
 #   include <unistd.h>          /* getthrid() */
+#elif defined(_AIX)
+#   include <sys/thread.h>      /* thread_self() */
 #endif
 
 /* The POSIX spec requires that use of pthread_attr_setstacksize
@@ -328,6 +330,9 @@ PyThread_get_thread_native_id(void)
 #elif defined(__OpenBSD__)
     pid_t native_id;
     native_id = getthrid();
+#elif defined(_AIX)
+    tid_t native_id;
+    native_id = thread_self();
 #endif
     return (unsigned long) native_id;
 }


aixtools:
@jaketesler - Hi, and thanks in advance. Many thanks for offering to be "tagged"!

This seems straight forward - and I hope complete (docs updated). (also thanks to @ncoghlan iirc who pointed at thread_self() - the "kernel" routine versus the posix thread *_self routine!

tiran (dismissed):
-1

As you already said, pthread_self does not return the TID, https://linux.die.net/man/3/pthread_self

> The thread ID returned by pthread_self() is not the same thing as the kernel thread ID returned by a call to gettid(2).

I'd rather not provide the feature at all than to provide an implementation that does not match with other implementations.

aixtools:
> 
> 
> -1
> 
> As you already said, pthread_self does not return the TID, https://linux.die.net/man/3/pthread_self
> 
> > The thread ID returned by pthread_self() is not the same thing as the kernel thread ID returned by a call to gettid(2).
> 
> I'd rather not provide the feature at all than to provide an implementation that does not match with other implementations.

And I did not use pthread_self() - I used thread_self() - also a routine that has been around for years.

A short example - with also the verification from "ps" that the value returned by thread_self is - indeed - the KERNEL tid and not the POSIX_tid.

```
root@x066:[/data/prj/aixtools/tests/posix]./thread_self
thread_self: pid:7602402 tid:12779547 ptid:1
[1] + Stopped (SIGTSTP)        ./thread_self
root@x066:[/data/prj/aixtools/tests/posix]ps -p 7602402 -mo THREAD
    USER     PID    PPID       TID ST  CP PRI SC    WCHAN        F     TT BND COMMAND
    root 7602402 7209084         - T    0  60  1 f1000a040034adb0   200011  pts/0   - ./thread_self
       -       -       -  12779547 T    0  60  1 f1000a040034adb0   410400      -   - -
root@x066:[/data/prj/aixtools/tests/posix]cat thread_self.c
#include <pthread.h>
#include <sys/thread.h>
#include <stdio.h>

main()
{
        pid_t pid;
        tid_t tid;
        pthread_t ptid;

        pid = getpid();
        tid = thread_self();
        ptid = pthread_self();

        fprintf(stderr,"thread_self: pid:%d tid:%d ptid:%d\n", pid, tid, ptid);
        sleep(300);     /* give time to run ps -mo THREAD */
}

```

So, not understanding the GIT process - you requested a change - but I do not see anything to change. Other than perhaps my comment where I thanked 'someone' for pointing out the issue BEFORE I submitted anything. 

Regards,
Michael

jaketesler:
This may provide some clarity – https://www.ibm.com/support/knowledgecenter/en/ssw_aix_72/com.ibm.aix.basetrf2/thread_self.htm

ncoghlan:
@aixtools @encukou has cleared out the review that was based on misreading the update. It's likely worth adding a comment to the new AIX code, referencing the link that @jaketesler's provided.

Do you have any thoughts on how we might be able to properly test this change? The existing native thread ID code doesn't appear to have any obvious tests either, so I wouldn't consider that a merge blocker, but it would still be nice if we could add that missing test coverage while folks have eyes on the code anyway.

aixtools:
On 09/06/2019 07:38, Nick Coghlan wrote:
> @aixtools @encukou has cleared out the review that was based on misreading the update. It's likely worth adding a comment to the new AIX code, referencing the link that @jaketesler's provided.
>
> Do you have any thoughts on how we might be able to properly test this change? The existing native thread ID code doesn't appear to have any obvious tests either, so I wouldn't consider that a merge blocker, but it would still be nice if we could add that missing test coverage while folks have eyes on the code anyway.
>
With all mails coming from "Nicole Trudeau", sometime hard to know who
is saying what.

a) I have no issue with adding a reference to
https://www.ibm.com/support/knowledgecenter/en/ssw_aix_72/com.ibm.aix.basetrf2/thread_self.htm,
but I wonder why to make an exception for the AIX bit.

I had replied directly to "whoever" had made the review - that I had
found the same comment in a book (white books I called them back then)
on AIX 4.1 documentation from June 1995).

I'll bow to your wisdom - but it feels funny giving AIX "special
attention". iirc, the other platforms do not have any link re: their
documentation.


b) as to how to verify the the difference between "pthread"_tid() -
whatever it name and "kernel"_tid() - for AIX I can only come up with
something (that would work on AIX, do not know how to implement it on
other platforms) is to verify the current PID is the "owner" of the TID
would be to call, and parse a command such as "ps -p $PID -mo tid".

ps -p $PID # fairly common"

-o tid # I expect to be 'common' as well

-m (on AIX stands for multi-line),

An example - not as root, so it works for all:

ps -p $$ -mo tid
      TID
        -
 22937633

So, while it could be used "as proof" - I think it would need to be a
separate PR/issue - as it will require "platform" input for all platforms.

As a footnote - I would not mind learning how to expand a test (even if
only for AIX) to make this comparison - as long as you know, in advance,
the basic approach I would take.

Regards,

Michael

aixtools:
Someone will need to send me better instructions on how to rebase. Locally it only showed the difference from master - this does not look 'rebased' at all.

The real changes are the same 4 files - and the same sort of changes - as were made for OpenBSD - bpo-37087, PR 13654

encukou:
> Someone will need to send me better instructions on how to rebase.

Which instructions were you using?
What you did here is a rebase *followed by a merge*.

skrah:
Why do you need to rebase?  I'm usually fine with `git pull https://github.com/python/cpython/ master` **after committing my local changes**, which then merges `master` into the local branch.

encukou:
> Why do you need to rebase?

The surrounding code changed.

asvetlov:
@aixtools your PR has 234 commits from which about 4 are yours, others are the result of false merge/rebase.
I personally usually use `git merge upstream/master` when working on PR because we squash all PR's commits before merging into master anyway -- but other people can have different preferences.

aixtools:
@encukou - besides reading the git manual I also reviewed https://www.atlassian.com/git/tutorials/rewriting-history/git-rebase. 

It looked - fine - from my forks perspective, or is that local copy - but when I pushed it here it seems to sees it only as a merge. Perhaps I need to add a --force flag - which I have never used to date - to make it all look 'cleaner'.

This is not my day job - and although I generally like git - it still gets the better of me - frequently.
I hope the real content is clear - the 4 files I had merge conflicts are the same files that we both updated to add this function - 

michael@x071:[/data/prj/python/git/python3-3.8]vi Doc/library/_thread.rst Doc/library/threading.rst Include/pythread.h Python/thread_pthread.h

I do not want to clutter this with my steps. I am saving the screen output, and if someone will be kind enough to explain where I went wrong I, hopefully, will not get caught like this again.

skrah:
> The surrounding code changed.

I guess technically you're correct but the point is that one does not have to call `git rebase` explicitly, fixing conflicts is part of the merge.

asvetlov:
Try `git merge master` or `git  merge upstream/master` depending on where `https://github.com/python/cpython` repo is present in your fork.
It can eliminate all foreign commits for your PR.

aixtools:
> I guess technically you're correct but the point is that one does not have to call `git rebase` explicitly, fixing conflicts is part of the merge.

As I said - not my day job - and when someone writes - elsewhere - that a "PR needs rebase" I made, the (wrong?) assumption - that that is the expected flow. In this case it would have been much easier to do it all manually. 

Big Sigh. There will be a next time and I'll probably feel forced to close the "old" PR and start a new one.  :( I do not like "messy".

> Try `git merge master` or `git merge upstream/master` depending on where `https://github.com/python/cpython` repo is present in your fork.
> It can eliminate all foreign commits for your PR.

So, I need to work from a combined setup - as I cannot commit/push to cpython.

My normal workflow is:
cd cypthon-master
git checkout master
git fetch cpython
git pull cpython master
git push --all aixtools # update my fork

And then switch to a clone of my fork
git fetch aixtools
git pull aixtools master
... # create branch bpo-XXXXX-tag, etc..
git commit -m"message"
git push --set-upstream aixtools bpo-XXXXX-tag

If I would add cpython as an additional upstream - can I "rebase" from it, and then push to my fork (as I cannot push to cpython) - and the foreign commits are eliminated?

asvetlov:
I use the so-called triangular workflow: https://github.blog/2015-07-29-git-2-5-including-multiple-worktrees-and-triangular-workflows/

I have a fork of cpython project here: https://github.com/asvetlov/cpython
My `cpython/.git/config` looks like:
```
[core]
	repositoryformatversion = 0
	filemode = true
	bare = false
	logallrefupdates = true
[remote "upstream"]
	url = git@github.com:python/cpython.git
	fetch = +refs/heads/*:refs/remotes/upstream/*
[branch "master"]
	remote = upstream
	merge = refs/heads/master
[remote "origin"]
	url = git@github.com:asvetlov/cpython.git
	fetch = +refs/heads/*:refs/remotes/origin/*
[branch "3.7"]
	remote = upstream
	merge = refs/heads/3.7
[branch "dangling-threads-in-watcher"]
	remote = origin
	merge = refs/heads/dangling-threads-in-watcher
```

So my local master is `upstream/master` but my working branches are `origin/feature-branch`.
To sync my feature branch with upstream master I do:
```
cd cpython
git checkout master
git pull
git checkout feature-branch
git merge master
git push -u origin feature-branch
```
Thus my clone has all my work but my master is `https://github.com/python/cpython.git`

I hope it helps you

aixtools:
Thanks. I'll do something similar - it has to be better than what I am doing now. :)

aixtools:
So, do I need to (try to) clean this up, or close the PR, and create a new one that is based on current master? Or will the process be able to proceed? I do not wish to make extra work for someone else (e.g., clean up actions).

methane:
I think Github will fix this automatically.
I saw similar trouble sometime.  They are fixed automatically after several hours.

asvetlov:
Whatever works for you.
Usually, we don't close existing PR to keep conversation history.
You can create a new clear branch, apply your 4 pull requests (by making patch file or cherry-picking) and force-push the new local branch to this existing github-side branch without closing the PR.

aixtools:
I'll try the cherry-pick, at least look at it, and force a new PR.

aixtools:
This looks cleaner. Thanks for the suggestions.

asvetlov (changes requested):
Cool!
Please resolve conflicts.
This is a job for PR author, not a merger.

A patch for NetBSD was landed before AIX :)

aixtools:
So, I get to learn - again - if I have mastered the "rebase" process :)

methane:
No need to learn "rebase".  Just use "merge".
In our workflow, "rebase" does not have any merit.  It only cause troubles like this.

ericvsmith:
But https://devguide.python.org/gitbootcamp/#syncing-with-upstream recommends rebase if you have local changes. Or maybe I misunderstand the use case here.

methane:
> But https://devguide.python.org/gitbootcamp/#syncing-with-upstream recommends rebase if you have local changes. Or maybe I misunderstand the use case here.

I think it is just an preference of the document author, not official recommendation.

skrah:
> But https://devguide.python.org/gitbootcamp/#syncing-with-upstream recommends rebase if you have local changes. Or maybe I misunderstand the use case here.

Yes, it looks a bit like a recommendation.  FWIW, I recently had this conversation with a highly advanced git user, and his opinion was to merge master into the PR for all projects that squash merges, because the alternative is too complex for PRs that are open for a long time.

aixtools:
OK - third time is the charm they say.

Question: line 358 of  Doc/library/threading.rst: the OpenBSD PR change this line. I am following the line from the original PR. So which is correct?

OpenBSD PR with:  .. availability:: Require :func:`get_native_id` function.

Or "original model": .. availability:: Windows, FreeBSD, Linux, macOS, OpenBSD, NetBSD, AIX.

If OpenBSD is correct, I'll update the PR again.

@methane,  @ericvsmith and @skrah - thanks for the extra info.

FYI - elsewhere I was told to use "rebase" - and I finally (think I) understand why the operation is named 'rebase' - and it looks nice on my repository - but horrible here (i.e., where it cam from). @asvetlov gave a very helpful "list" that is working nicely for me now.

methane:
> Yes, it looks a bit like a recommendation. FWIW, I recently had this conversation with a highly advanced git user, and his opinion was to merge master into the PR for all projects that squash merges, because the alternative is too complex for PRs that are open for a long time.

I agree with him.  While I use rebase often (especially for branches before sending pull request), it requires high git knowledge.  Additionally, it requires `-f` option when `git push`.  It is dangerous for new git users too.
I think "git rebase" shouldn't be recommended for new users.

aixtools:
I am thinking the "rebase" - read "merge" should be written up in the devguide. I tried reading up on it yesterday and this morning - and my head hurts - how to make it simple.

What I think has been vital for me was seeing how to merge from "upstream". I thought I had to maintain my fork at 100% - so I had a separate directory for "maintaining" the fork and then did all my cloneing, merge, pull, etc. from my fork rather than from "upstream".

I suppose when in a team using git - rather than "solo" (aka wannabe 'member') getting advice and direction in how to use git effectively is more efficient than googling through multiple opinions written up as tutorials.

:) so happy this "merge" went smoothly.

methane:
I sent pull request to stop recommend "rebase": python/devguide#496

Misc/NEWS.d/next/Core and Builtins/2019-05-28-11-47-44.bpo-37077.S1h0Fc.rst
```diff
@@ -0,0 +1,2 @@
+Add native thread ID (TID) for AIX
```
asvetlov:
```suggestion
:func:`threading.get_native_id` now also supports AIX.
```

Doc/library/threading.rst
```diff
@@ -355,7 +355,7 @@ since it is impossible to detect the termination of alien threads.
          system-wide) from the time the thread is created until the thread
          has been terminated.
 
-      .. availability:: Require :func:`get_native_id` function.
+      .. availability:: Windows, FreeBSD, Linux, macOS, OpenBSD, NetBSD, AIX.
```
asvetlov:
```suggestion
      .. availability:: Requires :func:`get_native_id` function.
```
Please keep the text from master.
@vstinner approved the change for NetBSD, I think we don't have to change it again (and fix doc in two places once next implementation will be added).

asvetlov (changes requested):

Doc/library/threading.rst
jaketesler: “Require” or “Requires”?

asvetlov: grepping python doc provides the only similarity from `os.rst`: 
```
.. availability:: Unix, Windows.  :func:`spawnlp`, :func:`spawnlpe`, :func:`spawnvp`
```
Also doesn't look perfect.

I don't know, I'm not a native speaker. Ok with any decision.

jaketesler: I’d suggest “Requires.” Maybe this will set a good precedent. ¯\_(ツ)_/¯

asvetlov: Updated suggestion

aixtools:

vstinner (approved):
LGTM.

asvetlov (approved):

PR #13630, opened 2019-05-28
bpo-35621: Support running subprocesses in asyncio when loop is executed in non-main thread

asvetlov:
Implementation adds `MultiLoopChildWatcher` and uses it by default.


Lib/asyncio/unix_events.py
```diff
@@ -1101,6 +1108,118 @@ def _do_waitpid_all(self):
                 callback(pid, returncode, *args)
 
 
+class MultiLoopChildWatcher(AbstractChildWatcher):
+    # The class keeps compatibility with AbstractChildWatcher ABC
+    # To achieve this it has empty attach_loop() method
+    # and don't accept explicit loop argument
```
1st1: don't -> doesn't

Lib/asyncio/unix_events.py
```diff
@@ -1101,6 +1108,118 @@ def _do_waitpid_all(self):
...
+            del self._callbacks[pid]
+            return True
+        except KeyError:
+            return False
+
+    def attach_loop(self, loop):
+        # Don't attach the loop but initialize itself if called first time
+        # The reason to do it here is that attach_loop() is called from
+        # unix policy only for the main thread.
+        # Main thread is required for subscription on SIGCHLD signal
```
1st1, line 1165: Can you add an assertion that the current thread is indeed the main thread?

1st1 (commented):
Is it possible to make MultiChildWatcher to work like FastChildWatcher? Can we just fix them without introducing a new class? I think we discussed these options but I don't remember the resolution.

Lib/asyncio/unix_events.py
asvetlov: The assertion is not needed: `signal.signal` raises an exception if the thread is not main.

Lib/asyncio/unix_events.py
asvetlov: Done

Lib/asyncio/unix_events.py
1st1: Right, never mind.

asvetlov:
> Is it possible to make MultiChildWatcher to work like FastChildWatcher? Can we just fix them without introducing a new class? I think we discussed these options but I don't remember the resolution.

Sorry, no.
1. FastChildWatcher uses `loop.add_signal_handler()` for subscription on SIGCHLD, so this watcher is tightly coupled with loop. Changing this behavior is not backward compatible, I prefer to keep existing watchers as is.
2. `FastChildWatcher` is dangerous: it reaps all terminated processes, even not started by asyncio code.
Basically, if you have a thread that uses `subprocess.run()` that call can never wait for the process finish because `FastChildWatcher` swallows this.

Currently we have O(N) `SafeChildWatcher` default. `MultiLoopChildWatcher` has the same O(N) complexity but works with many threads.
The only requirement is that this watcher should be initialized in the main thread.
It can be accomplished in two ways:
1. Start a loop in the main thread (`asyncio.get_event_loop()` installs the watcher is executed from the main thread). `asyncio.new_event_loop()` doesn't install child watcher for a new loop unfortunatelly.
2. Explicitly call `asyncio.get_event_loop_policy().get_child_watcher()` from the main thread before starting a loop in sub-thread. I know, looks tricky. I'll document it carefully. 

That's how asyncio works now and from the very beginning. I'm not very comfortable with changing policy on-demand initialization too much now.

Calling `self._init_watcher()` from `_UnixDefaultEventLoopPolicy.__init__` if the calling thread is main might make sense, though.
The change is minimal, safe and backward compatible.
We don't create a policy on import time anyway.
The change allows to get `MultiLoopChildWatcher` installed and ready for `asyncio.new_event_loop()` if called from the main thread. Also instead of `asyncio.get_event_loop_policy().get_child_watcher()` a user can write just `asyncio.get_event_loop_policy()`. I believe it simplifies things.

1st1:
> 1. Start a loop in the main thread (`asyncio.get_event_loop()` installs the watcher is executed from the main thread). `asyncio.new_event_loop()` doesn't install child watcher for a new loop unfortunatelly.
> 2. Explicitly call `asyncio.get_event_loop_policy().get_child_watcher()` from the main thread before starting a loop in sub-thread. I know, looks tricky. I'll document it carefully.

3. We use `Py_AddPendingCall` from `asynciomodule.c` to schedule a callback to initialize `MultiLoopChildWatcher`.  What do you think?  Ideally we need to expose `Py_AddPendingCall` via the `sys.module`.

1st1:
I've quickly prototyped this: https://github.com/python/cpython/pull/13656

1st1:
> Explicitly call asyncio.get_event_loop_policy().get_child_watcher() from the main thread before starting a loop in sub-thread. I know, looks tricky. I'll document it carefully.

How about we don't allow users to use subprocess APIs from non-main threads if we don't have a working child watcher ready?

asvetlov:
I suspect that `sys.addpendingcall()` will raise a long discussion.
Right now I prefer adding `MultiLoopChildWatcher` and initializing it on policy creation time if the loop is main.

> How about we don't allow users to use subprocess APIs from non-main threads if we don't have a working child watcher ready?
It is tricky: watcher has no API to tell is it "ready". Existing watchers are "ready" between `watcher.attach_loop(loop)` and `watcher.attach_loop(None)` calls.
The watcher provided by this PR is "ready" after the first `attach_loop()` call (doesn't matter if `loop` was passed or `None`). All subsequent `attach_loop()` calls do nothing. The watcher becomes "not ready"` after `watcher.close()` call.

I can implement `AbstractChileWatcher.is_ready()` method. The watcher is ready if `is_ready` returns True. `AbstractChileWatcher.is_ready()` should return `True` for backward compatibility with existing third-party watchers (not sure if somebody provides custom watchers but still). asyncio watchers will override the method to return proper state.

`UnixEventLoop._make_subprocess_transport` will raise `RuntimeError` if the watcher is not ready. At the moment of the call the watcher always exists but can be "not ready" as described above.

Also, I don't like to create a waiting thread per subprocess. I suspect that this approach doesn't scale well for 10,000 subprocesses. I don't want to apply this change in the last day before the feature freeze.

We can return to a discussion about subprocess support redesign in Python 3.9

1st1 (approved):

1st1:
How about we rename ThreadedChildWatcher to DefaultChildWatcher? My reasoning: if it's called the latter, it could implement different watching strategy based on OS or Python version. On macOS we could use kwueue, on linux threads etc.

Alternatively, we can keep this as is, implement KQueue watcher (in 3.9) and let asyncio decide which one should be the default at import time.

Thoughts?

1st1:
Let's keep it named ThreadedChildWatcher :)

PR #13631, opened 2019-05-28
bpo-37081: Test with OpenSSL 1.1.1c

tiran:

https://bugs.python.org/issue37081


tiran:
@ned-deily Should I update 3.6, too?

ned-deily:
@tiran, Based on your proposal to not fully support 1.1.1 in 3.6 (https://bugs.python.org/issue32947), perhaps not?

tiran:
@ned-deily Sounds good to me! I'm removing the backport to 3.6 label again.

PR #6286, opened 2018-03-28
bpo-33164: Blake2 dependency update

ghost:
Updating the module following-up recent changes

https://github.com/BLAKE2/libb2/commit/620681a3b15c4d7239b9323b9da5ea208a959d3d

https://bugs.python.org/issue33164

PR #13633, opened 2019-05-28
bpo-33164: blake2 fix for HP-UX

ghost:
Backporting fix on blake2 side too.

Fixes from @michael-o mainly.

https://bugs.python.org/issue33164

### DIFF
--- a/Modules/_blake2/impl/blake2-impl.h
+++ b/Modules/_blake2/impl/blake2-impl.h
@@ -140,6 +140,9 @@ static inline void secure_zero_memory(void *v, size_t n)
 {
 #if defined(_WIN32) || defined(WIN32)
   SecureZeroMemory(v, n);
+#elif defined(__hpux)
+  static void *(*const volatile memset_v)(void *, int, size_t) = &memset;
+  memset_v(v, 0, n);
 #else
 // prioritize first the general C11 call
 #if defined(HAVE_MEMSET_S)
--- a/configure.ac
+++ b/configure.ac
@@ -3527,6 +3527,8 @@ AC_CHECK_FUNCS(alarm accept4 setitimer getitimer bind_textdomain_codeset chown \
  getgrouplist getgroups getlogin getloadavg getpeername getpgid getpid \
  getpriority getresuid getresgid getpwent getpwnam_r getpwuid_r getspnam getspent getsid getwd \
  if_nameindex \
+ initgroups kill killpg lchmod lchown lockf linkat lstat lutimes mmap \
+ memrchr mbrtowc mkdirat mkfifo \
  madvise mkfifoat mknod mknodat mktime mremap nice openat pathconf pause pipe2 plock poll \
  posix_fallocate posix_fadvise posix_spawn posix_spawnp pread preadv preadv2 \
  pthread_condattr_setclock pthread_init pthread_kill putenv pwrite pwritev pwritev2 \
--- a/configure
+++ b/configure
@@ -785,6 +785,7 @@ infodir
 docdir
 oldincludedir
 includedir
+runstatedir
 localstatedir
 sharedstatedir
 sysconfdir
@@ -897,6 +898,7 @@ datadir='${datarootdir}'
 sysconfdir='${prefix}/etc'
 sharedstatedir='${prefix}/com'
 localstatedir='${prefix}/var'
+runstatedir='${localstatedir}/run'
 includedir='${prefix}/include'
 oldincludedir='/usr/include'
 docdir='${datarootdir}/doc/${PACKAGE_TARNAME}'
@@ -1149,6 +1151,15 @@ do
   | -silent | --silent | --silen | --sile | --sil)
     silent=yes ;;
 
+  -runstatedir | --runstatedir | --runstatedi | --runstated \
+  | --runstate | --runstat | --runsta | --runst | --runs \
+  | --run | --ru | --r)
+    ac_prev=runstatedir ;;
+  -runstatedir=* | --runstatedir=* | --runstatedi=* | --runstated=* \
+  | --runstate=* | --runstat=* | --runsta=* | --runst=* | --runs=* \
+  | --run=* | --ru=* | --r=*)
+    runstatedir=$ac_optarg ;;
+
   -sbindir | --sbindir | --sbindi | --sbind | --sbin | --sbi | --sb)
     ac_prev=sbindir ;;
   -sbindir=* | --sbindir=* | --sbindi=* | --sbind=* | --sbin=* \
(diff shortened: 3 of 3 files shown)


Modules/_blake2/impl/blake2-impl.h
```diff
@@ -140,6 +140,9 @@ static inline void secure_zero_memory(void *v, size_t n)
 {
 #if defined(_WIN32) || defined(WIN32)
   SecureZeroMemory(v, n);
+#elif deifned(__hpux)
```
tiran: typo, should be "defined"

configure.ac
```diff
@@ -3527,7 +3527,9 @@ AC_CHECK_FUNCS(alarm accept4 setitimer getitimer bind_textdomain_codeset chown \
  getgrouplist getgroups getlogin getloadavg getpeername getpgid getpid \
  getpriority getresuid getresgid getpwent getpwnam_r getpwuid_r getspnam getspent getsid getwd \
  if_nameindex \
- madvise mkfifoat mknod mknodat mktime mremap nice openat pathconf pause pipe2 plock poll \
```
tiran: What and why did you modify the configure script?

tiran (changes requested):

Modules/_blake2/impl/blake2-impl.h
ghost: good catch ...

configure.ac
ghost: In fact I got the fixes from a HP-UX specialist, if it works without the better.

ghost: will remove the changes ...

michael-o: The change comes actually from the original PR which broke it, https://github.com/python/cpython/commit/51aa35e9e17eef60d04add9619fe2a7eb938358c#diff-67e997bcfdac55191033d57a16d1408a too much has been removed as opposed to the `configure` script.

michael-o: The `configure.ac` needs to retain:

```
initgroups kill killpg lchmod lchown lockf linkat lstat lutimes mmap \
memrchr mbrtowc mkdirat mkfifo \
```

Modules/_blake2/impl/blake2-impl.h
```diff
@@ -140,6 +140,9 @@ static inline void secure_zero_memory(void *v, size_t n)
 {
 #if defined(_WIN32) || defined(WIN32)
   SecureZeroMemory(v, n);
+#elif defined(__hpux)
+  static void *(*const volatile memset_v)(void *, int, size_t) = &memset;
+  memset_v(v, 0,n);
```
michael-o: This one misses a space after the second comma

Modules/_blake2/impl/blake2-impl.h
michael-o: Here is my upstream fix: https://github.com/BLAKE2/libb2/commit/37062e55cc52895f5c79aeebe40f45409c008c90

configure.ac
michael-o: Here is the patch I have sent via email:

```
# diff -U 1 configure.ac.orig configure.ac
--- configure.ac.orig   2019-05-28 14:34:58 +0000
+++ configure.ac        2019-05-28 14:38:23 +0000
@@ -3526,6 +3526,8 @@
 getgrouplist getgroups getlogin getloadavg getpeername getpgid getpid \
 getpriority getresuid getresgid getpwent getpwnam_r getpwuid_r getspnam getspent getsid getwd \
 if_nameindex \
+ initgroups kill killpg lchmod lchown lockf linkat lstat lutimes mmap \
+ memrchr mbrtowc mkdirat mkfifo \
 mkfifoat mknod mknodat mktime mremap nice openat pathconf pause pipe2 plock poll \
 posix_fallocate posix_fadvise posix_spawn posix_spawnp pread preadv preadv2 \
 pthread_condattr_setclock pthread_init pthread_kill putenv pwrite pwritev pwritev2 \
```

Please reinstantiate otherwise all of these won't be available at compile time.

@dcarlier-afilias please readd otherwise people are back to the configure error.

michael-o:
@dcarlier-afilias looks very good now. Let me double check on the server tomorrow and then you can merge into master! Thank you very much for the quick reaction.

Modules/_blake2/impl/blake2-impl.h
```diff
@@ -140,6 +140,9 @@ static inline void secure_zero_memory(void *v, size_t n)
 {
 #if defined(_WIN32) || defined(WIN32)
   SecureZeroMemory(v, n);
+#elif defined(__hpux)
+  static void *(*const volatile memset_v)(void *, int, size_t) = &memset;
+  memset_v(v, 0, n);
```
methane, line 145: How is it different from just `memset(v, 0, n)`?

This code looks different from https://github.com/BLAKE2/libb2/commit/b4b241a34824b51956a7866606329a065d397525

ghost: there was a recent change to fix HP-UX build which is backported from [this version](https://github.com/BLAKE2/libb2/blob/master/src/blake2-impl.h#L144).
There is a subtle difference, this prevents the compiler to optimise away the buffer clearing, the volatile function pointer plays its role here.

methane: I got [reply in upstream](https://github.com/BLAKE2/libb2/pull/26#discussion_r288364105) too.

While I like comment here, it's up to upstream project.  We should copy code as-is to reduce maintenance cost.

configure.ac
methane: I'm sorry about I merged the previous PR.  I thought travis checked configure script is really generated from configure.ac.

methane:
I will merge this after @michael-o 's check.

michael-o (approved):
Compiles for me on top of 0ae022c6a47abffce22ec185552e319b7b93dbf4 with my outstanding PR: https://github.com/python/cpython/pulls/michael-o. Would be nice if someone could take a look at them. They are trivial.

tiran:
@methane please revert the merge. The modifications to autoconf script are incorrect.

michael-o:
@tiran, why? It restores the `configure.ac` before it was broken by the blake2 upgrade + added functions.

methane:
@tiran "minimize patch manually after autoreconf" makes this type of trouble.
I created #13651 which commit all generated files by autoreconf as-is.

tiran:
@michael-o The commit d8b755167235e0621814eb5ac39163b3db6879bb only changes ``configure.ac``. CPython also has the ``configure`` script in VCS. Any change to ``configure.ac`` must also update ``configure``.

methane:
@tiran Mismatch between `configure` and `configure.ac` was introduced by #6286.
https://github.com/python/cpython/pull/6286/files?short_path=2328065#diff-e2d5a00791bce9a01f99bc6fd613a39d

This PR resolves the mismatch.  I'm sorry about merging #6286.

methane:
I re-run `autoreconf` and commit generated files without manual modification, except `aclocal.m4` (GH-13651).

PR #13634, opened 2019-05-28
bpo-37072: Fix crash in PyAST_FromNodeObject() when flags is NULL

gvanrossum:
I'm confident that this fixes the reported crash. flags=NULL is treated as using the latest minor version.

https://bugs.python.org/issue37072


gvanrossum:
This is a one-line fix, I'm just going to merge it.

PR #13638, opened 2019-05-28
Doc: Add an optional obsolete header.

JulienPalard:
This is to be backported to old versions of documentation, and when activated can look like this:

### DIFF
--- a/Doc/README.rst
+++ b/Doc/README.rst
@@ -116,14 +116,8 @@ see the make targets above).
 Deprecation header
 ==================
 
-You can define the ``outdated_message`` and ``outdated_link_text``
-variables in ``html_context`` to show a red banner on each page
-redirecting to the "latest" version.
-
-The ``outdated_message`` comes first, the ``outdated_link_text`` comes
-afterwards as a link, like this::
-
-    {{ outdated_message }}<a href="to new version">{{ outdated_link_text }}</a>.
+You can define the ``outdated`` variable in ``html_context`` to show a
+red banner on each page redirecting to the "latest" version.
 
 The link points to the same page on ``/3/``, sadly for the moment the
 language is lost during the process.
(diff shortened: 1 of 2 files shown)


JulienPalard:
@theacodes moved here :)

Doc/tools/templates/layout.html
```diff
@@ -1,5 +1,14 @@
 {% extends "!layout.html" %}
 
+{% block header %}
+{%- if outdated_message is defined and outdated_link_text is defined %}
+<div id="outdated-warning" style="padding: .5em; text-align: center; background-color: #FFBABA; color: #6A0E0E;">
+    {{ outdated_message }}
+    <a href="/3/{{ pagename }}{{ file_suffix }}">{{ outdated_link_text }}</a>.
```
theacodes: I would just make a flag to *show* the outdated message, and then just have the message itself and the link text be static text wrapped in `{% trans %}` to ease translation.

JulienPalard: translations? I like translations!

theacodes: Good for the heart, good for the soul. :)

JulienPalard: It would be nice to have two translations strings:

> This document is for an old version of Python that is no longer supported. You should upgrade, and read the {link}.

and

> Python documentation for the last stable release

Using the current tags I'll get:

> This document is for an old version of Python that is no longer supported. You should upgrade, and read the

and

> Python documentation for the last stable release

Almost the same, but in the current version translations can't move the link inside the sentence, and they can't touch the trailing dot.

eamanu (approved):

Carreau (approved):

asvetlov (approved):

Doc/README.rst
```diff
@@ -113,6 +113,15 @@ Then, from the ``Doc`` directory, run ::
...
 see the make targets above).
 
+Deprecation header
+==================
+
+You can define the ``outdated`` variable in ``html_context`` to show a
+red banner on each page redirecting to the "latest" version.
+
+The link points to the same page on ``/3/``, sadly for the moment the
+language is lost during the process.
```
merwok, line 123: For the record, the 2.7 PR addressed this.

PR #13639, opened 2019-05-28
bpo-36919: make test_source_encoding.test_issue2301 implementation-independent

BCSharp:
https://bugs.python.org/issue36919


auvipy (approved):

jaraco (approved):
Looks good. If this pattern is extended to more than one other case, it will be advisable to extract the behavior as a helper function that describes the justification for the behavior.

PR #13640, opened 2019-05-28
Fix markup and minor grammar improvements in Code_of_conduct.md

hydrogen-mvm:
The old link had a > in the url which prevented the browser from jumping down to the correct section on that page.

That PSF page itself has an error: There's a duplicate "the" in that paragraph that needs to be removed: "...and conform to **the the** Python Community Code of Conduct."

While I was editing this file, I also fixed some grammar and bolded the 3 important keywords so that they catch the viewer's eyes. I can revert these changes if they are unwanted.

Thanks.

### DIFF
--- a/CODE_OF_CONDUCT.md
+++ b/CODE_OF_CONDUCT.md
@@ -2,11 +2,11 @@
 
 Please note that all interactions on
 [Python Software Foundation](https://www.python.org/psf-landing/)-supported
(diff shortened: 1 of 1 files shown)


Mariatta (approved):

Mariatta:
Thanks @hydrogen-mvm for your first contribution to CPython 🎉

PR #13642, opened 2019-05-29
Regenerate pydoc topics file

pablogsal:


pablogsal:
For some reason, the newlines looks wrong....investigating...

Lib/pydoc_data/topics.py
```diff
@@ -2558,22 +2558,25 @@
              '(see\n'
              'section The standard type hierarchy):\n'
              '\n'
-             '   funcdef                 ::= [decorators] "def" funcname "(" '
+             '   funcdef                   ::= [decorators] "def" funcname "(" '
              '[parameter_list] ")"\n'
              '               ["->" expression] ":" suite\n'
-             '   decorators              ::= decorator+\n'
-             '   decorator               ::= "@" dotted_name ["(" '
+             '   decorators                ::= decorator+\n'
```
eamanu, line 2564: Why white-spaces is added?

pablogsal: This file is automatically generated

Carreau:
Out of curiosity how are these regenerated, and where can I find this information ?

pablogsal:
@Carreau `make pydoc-topics` in the Docs folder, but usually is done by the release manager before releases (I just learned that fact looking at the git history for that file).

PR #13645, opened 2019-05-29
bpo-37094: Add example for TestCase.skipTest in unittest doc

MakDon:
Also includes other minor test skipping doc improvements.

https://bugs.python.org/issue37094

### DIFF
--- a/Doc/library/unittest.rst
+++ b/Doc/library/unittest.rst
@@ -539,7 +539,7 @@ This is the output of running the example above in verbose mode::
 
    test_format (__main__.MyTestCase) ... skipped 'not supported in this library version'
    test_nothing (__main__.MyTestCase) ... skipped 'demonstrating skipping'
-   test_skip_me (tmp.MyTestCase) ... skipped 'skip'
+   test_skip_me (__main__.MyTestCase) ... skipped 'skip'
    test_windows_support (__main__.MyTestCase) ... skipped 'requires Windows'
 
    ----------------------------------------------------------------------
--- a/Doc/library/unittest.rst
+++ b/Doc/library/unittest.rst
@@ -510,7 +510,8 @@ that is broken and will fail, but shouldn't be counted as a failure on a
 :class:`TestResult`.
 
 Skipping a test is simply a matter of using the :func:`skip` :term:`decorator`
-or one of its conditional variants.
+or one of its conditional variants. And :meth:`TestCase.skipTest` could be used to
+skip a test.
 
 Basic skipping looks like this::
 
@@ -530,6 +531,9 @@ Basic skipping looks like this::
        def test_windows_support(self):
            # windows specific testing code
            pass
+
+       def test_skip_me(self):
+           self.skipTest("skip")
 
 This is the output of running the example above in verbose mode::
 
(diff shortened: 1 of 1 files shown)


auvipy (approved):

csabella:
Hi @MakDon, please create a [bug tracker issue](https://bugs.python.org/) for this and link the bpo number in the title.  Thanks!

MakDon:
Hi @csabella , according to [Python Developer’s Guide](https://devguide.python.org/), Trivial issues (e.g. typo fixes) do not require any issue to be created.

It's just a document updating, so maybe it is not important enough to create an issue on issues tracker.

Thank you for your reply! :D

csabella:
Yes, it's not obvious what trivial means, but it's really just a pull request with only a few typo fixes.  Adding an additional example is more than trivial.  Also, some core devs primary way of discovering issues is through the bug tracker, so adding a PR without a bug tracker link increases that likelihood of a change not being reviewed.  It's usually better to have a bpo number than to not have one.  🙂

MakDon:
@csabella Maybe you are right. I created an [issue](https://bugs.python.org/issue37094) on issue tracker. Thanks.

Doc/library/unittest.rst
```diff
@@ -531,16 +532,20 @@ Basic skipping looks like this::
            # windows specific testing code
            pass
 
+       def test_skip_me(self):
```
zware: If you take my second suggestion for the body of this example method, it could use a better name ("test_maybe_skipped" or similar, perhaps).

Doc/library/unittest.rst
```diff
@@ -531,16 +532,20 @@ Basic skipping looks like this::
...
+           self.skipTest("skip")
```
zware: Nit, but I'd prefer not to encourage meaningless skip messages :).  How about

```suggestion
           self.skipTest("skipped from within")
```

or similar?  Even better might be to show that this can be called conditionally, something like:

```suggestion
           if not external_resource_available():
               self.skipTest("external resource not available")
           # test code that depends on the external resource
           pass
```

Doc/library/unittest.rst
```diff
@@ -510,7 +510,8 @@ that is broken and will fail, but shouldn't be counted as a failure on a
 :class:`TestResult`.
 
 Skipping a test is simply a matter of using the :func:`skip` :term:`decorator`
-or one of its conditional variants.
+or one of its conditional variants. And :meth:`TestCase.skipTest` could be used to
```
zware: The added sentence is not complete.  How about:

```suggestion
or one of its conditional variants, calling :meth:`TestCase.skipTest` within a
:meth:`~TestCase.setUp` or test method, or raising :exc:`SkipTest` directly.
```

zware (changes requested):
Needs a little bit of cleanup, but I think this will end up as a nice improvement!

MakDon:
@zware Thanks for your suggestions.
The document is still WIP.Please wait for a minute.
Thank your again!

MakDon:

zware (approved):
LGTM

MakDon:
Cause [Python 3.8.0b1 branch cut-off date moved to Monday](https://discuss.python.org/t/python-3-8-0b1-branch-cut-off-date-moved-to-monday/1775), could i see this document update in 3.8? :D

zware:
There is no restriction on documentation improvements; this will apply to 3.7 and beyond :)

PR #12843, opened 2019-04-15
bpo-36624: Add platform constants in Lib/test/support (support.MS_WINDOWS, support.AIX, etc)

aixtools:
This is just the first phase - an example of what could be - although I have added other constants as discussed during #7800.

During the previous discussion, backport was discussed, and if manual work is needed for the backport(s) I'll do that as well.

https://bugs.python.org/issue36624

PR #13648, opened 2019-05-29
bpo-36624: Add platform constants in test.support

taleinat:
Add platform constants in `Lib/test/support.py`: `support.MS_WINDOWS`, `support.AIX`, etc.

This is a clean rebase of the original PR by @aixtools, GH-12843.

I've also fixed two minor errors which broke `test_stat` and `test_subprocess`.

https://bugs.python.org/issue36624

### DIFF
--- a/Lib/test/test_decimal.py
+++ b/Lib/test/test_decimal.py
@@ -35,7 +35,7 @@ import locale
 from test.support import (run_unittest, run_doctest, is_resource_enabled,
                           requires_IEEE_754, requires_docstrings)
 from test.support import (import_fresh_module, TestFailed,
-                          run_with_locale, cpython_only)
+                          run_with_locale, cpython_only, MS_WINDOWS)
 import random
 import inspect
 import threading
@@ -4862,7 +4862,7 @@ class CWhitebox(unittest.TestCase):
         for attr in ('prec', 'Emin', 'Emax', 'capitals', 'clamp'):
             self.assertRaises(OverflowError, setattr, c, attr, int_max+1)
             self.assertRaises(OverflowError, setattr, c, attr, -int_max-2)
-            if sys.platform != 'win32':
+            if not MS_WINDOWS:
                 self.assertRaises(ValueError, setattr, c, attr, int_max)
                 self.assertRaises(ValueError, setattr, c, attr, -int_max-1)
 
@@ -4937,7 +4937,7 @@ class CWhitebox(unittest.TestCase):
         # OverflowError, general ValueError
         self.assertRaises(OverflowError, setattr, c, '_allcr', int_max+1)
         self.assertRaises(OverflowError, setattr, c, '_allcr', -int_max-2)
-        if sys.platform != 'win32':
+        if not MS_WINDOWS:
             self.assertRaises(ValueError, setattr, c, '_allcr', int_max)
             self.assertRaises(ValueError, setattr, c, '_allcr', -int_max-1)
 
@@ -4945,7 +4945,7 @@ class CWhitebox(unittest.TestCase):
         for attr in ('_flags', '_traps'):
             self.assertRaises(OverflowError, setattr, c, attr, int_max+1)
             self.assertRaises(OverflowError, setattr, c, attr, -int_max-2)
-            if sys.platform != 'win32':
+            if not MS_WINDOWS:
                 self.assertRaises(TypeError, setattr, c, attr, int_max)
                 self.assertRaises(TypeError, setattr, c, attr, -int_max-1)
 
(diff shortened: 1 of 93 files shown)


Lib/test/test_decimal.py
```diff
@@ -4862,7 +4862,7 @@ def test_c_context_errors(self):
         for attr in ('prec', 'Emin', 'Emax', 'capitals', 'clamp'):
             self.assertRaises(OverflowError, setattr, c, attr, int_max+1)
             self.assertRaises(OverflowError, setattr, c, attr, -int_max-2)
-            if sys.platform != 'win32':
+            if not MS_WINDOWS:
```
skrah, line 4865: I'm not sure that this adds a lot of clarity.  Could you remove that?

aixtools:
> 
> 
> Add platform constants in `Lib/test/support.py`: `support.MS_WINDOWS`, `support.AIX`, etc.
> 
> This is a clean rebase of the original PR by @aixtools, [GH-12843](https://github.com/python/cpython/pull/12843).
> 
> I've also fixed two minor errors which broke `test_stat` and `test_subprocess`.
> 
> https://bugs.python.org/issue36624

Many thanks for the rebase - and corrections!

skrah:
I'm a bit uncertain about the benefit of these in `test.support`.  In other locations like `setup.py` one still needs `sys.platform == name`, which IMO is actually quite nice.  So I don't feel a need to replace it.

asvetlov (approved):
LGTM for asyncio part

skrah (changes requested):
Could you take out test_decimal? I don't quite like this change.

Lib/test/test_decimal.py
aixtools: It is not about just this one file. The request was, as much as possible, to use constants in the tests rather than different approaches such as 'sys.platform.startswith("win")' and sys.platform == 'win32'. 

Compare with (if pasting images works)
image
 Otherwise:

Look at this file (Lib/test/_test_multiprocessing.py, https://github.com/python/cpython/pull/13648/files#diff-b046ab474480855fb4a01de88cfc82bb) - where, for windows both a local constant (WIN32) and sys.platform {==|!=} 'win32' is used.

imho - this adds clarity to the whole.

skrah:
On Wed, May 29, 2019 at 02:43:16AM -0700, Michael Felt wrote:
> imho - this adds clarity to the whole.

I don't think so: Now I have to wonder which of the different idioms hides behind MS_WINDOWS.

And in other project's setup.py files I still need sys.platform, so now there's even one additional way to accomplish the same thing.

aixtools:
> 
> 
> Could you take out test_decimal? I don't quite like this change.

As I tried to show above - it is not just one file. IMHO - this needs to be unanimous. Without that there is no future where a code review could say - use constants.

And, yes, some platforms such as bsd* and solaris do not have constants - yet. 

"My" intent is that they are also done - asap. Where there is a will, there is way.

aixtools:
> 
> On Wed, May 29, 2019 at 02:43:16AM -0700, Michael Felt wrote: imho - this adds clarity to the whole.
> I don't think so: Now I have to wonder which of the different idioms hides behind MS_WINDOWS. And in other project's setup.py files I still need sys.platform, so now there's even one additional way to accomplish the same thing.

So, Python does nothing?

Doing this in "test" gives "the example" on how it should be done.

As to the areas outside of test, and even within test (e.g., for bsd*) I am willing to make this "my mission" to bring uniformity wherever possible. This was not even my idea :) - this was a suggestion on how I could contribute something to the whole.

Again, if not clear before - I do not see it stopping here - but would clearly welcome direction on where to set priorities.

It is not my call - I am not in python-dev mailing list - so I will not see a discussion there. If the group feels this is wrong - I have no direct interest in any decision made - other than my attempt - read desire - to contribute to Python.

Regards,
Michael

> 
> On Wed, May 29, 2019 at 02:43:16AM -0700, Michael Felt wrote: imho - this adds clarity to the whole.
> I don't think so: Now I have to wonder which of the different idioms hides behind MS_WINDOWS. And in other project's setup.py files I still need sys.platform, so now there's even one additional way to accomplish the same thing.

taleinat:
Considering this is a rather minor change, discussion about it should be done on the bug tracker, bpo-36624 (except when discussion specifics of the implementation in this PR).

@skrah, your input would be helpful there.

@aixtools, you've made your point rather clearly, let's see what some other devs think about this now that there's a PR to look at.

skrah:
@taleinat I'm aware of the bug tracker.

taleinat:
@skrah, I have no doubt about that :)

I wanted to stop the discussion from continuing here where it could be missed by those following the issue on the bug tracker, so I wrote to clarify for everyone.

ned-deily:
@taleinat, FWIW, I disagree wih your assertion that this is a "rather minor change".  See my comments on the bug tracker issue.

taleinat:
The discussion here and on the issue is clearly going towards not making this change, so I'm closing this PR.

PR #13649, opened 2019-05-29
 bpo-21131: don't use SIGSTKSZ for stack size in sigaltstack

peadar:
Following up discussion on http://bugs.python.org/issue21131

vstinner suggested using the logic in `thread_pthread.h` to work out the "good" stack size for sigaltstack. I've integrated the logic in there, but it doesn't actually provide an accessible way of finding the default thread stack size - it just nominally requests pthread_create to create a stack of a "default" size by not poking at the stack size attribute.

So, instead, I've rejigged the macros in that file to provide PLATFORM_STACK_SIZE and PLATFORM_STACK_PREFERRED that are available regardless of the ability of pthreads to set the thread stack size attribute (i.e., regardless of the presence of _POSIX_THREAD_ATTR_STACK_SIZE) (and simplified some of the conditionals in the code using THREAD_STACK_SIZE in the process). The eventual sizes past to `pthread_attr_setstacksize` are unaffected.

This is then used as a fallback in the new function, `size_t _Pthread_preferred_stacksize()` , which will first attempt to use a default-constructed pthread_attr_t to work out at runtime what size stack the system prefers for new threads. This value is far more likely to be usable than the value defined in SIGSTKSZ.

If that call fails, or the thread size attribute is not available, we default to the maximum of PTHREAD_STACK_MIN, whatever we've defined for PLATFORM_STACK_SIZE for those platforms with an inadequate default stacksize, or finally the arbitrary 32k setting previously used as a floor.

All feedback gratefully appreciated.

https://bugs.python.org/issue21131

### DIFF
--- a/Include/internal/pycore_thread.h
+++ b/Include/internal/pycore_thread.h
@@ -14,7 +14,7 @@ extern "C" {
 /*
  * Find the size of the stack the system uses by default.
  */
-PyAPI_FUNC(size_t) _PyThread_preferred_stacksize();
+PyAPI_FUNC(size_t) _PyThread_preferred_stacksize(void);
 
 #ifdef __cplusplus
 }
--- a/Include/internal/pycore_thread.h
+++ b/Include/internal/pycore_thread.h
@@ -1,0 +1,22 @@
+#ifndef Py_INTERNAL_THREAD_H
+#define Py_INTERNAL_THREAD_H
+
+#ifdef __cplusplus
+extern "C" {
+#endif
+
+#include <stdint.h>
+
+#ifndef Py_BUILD_CORE
+#  error "this header requires Py_BUILD_CORE define"
+#endif
+
+/*
+ * Find the size of the stack the system uses by default.
+ */
+PyAPI_FUNC(size_t) _PyThread_preferred_stacksize();
+
+#ifdef __cplusplus
+}
+#endif
+#endif /* !Py_INTERNAL_THREAD_H */
--- a/Python/thread_nt.h
+++ b/Python/thread_nt.h
@@ -496,3 +496,13 @@ PyThread_tss_get(Py_tss_t *key)
     SetLastError(error);
     return result;
 }
+
+size_t
+_PyThread_preferred_stacksize()
+{
+    /*
+     * Windows default stack size is 1MB, overridable by the linker in the
+     * executable
+     */
+    return 1024 * 1024;
+}
--- a/Python/thread_pthread.h
+++ b/Python/thread_pthread.h
@@ -24,13 +24,6 @@
 #   include <lwp.h>             /* _lwp_self() */
 #endif
 
-/* The POSIX spec requires that use of pthread_attr_setstacksize
-   be conditional on _POSIX_THREAD_ATTR_STACKSIZE being defined. */
-#ifdef _POSIX_THREAD_ATTR_STACKSIZE
-#ifndef THREAD_STACK_SIZE
-#define THREAD_STACK_SIZE       0       /* use default stack size */
-#endif
-
 /* The default stack size for new threads on OSX and BSD is small enough that
  * we'll get hard crashes instead of 'maximum recursion depth exceeded'
  * exceptions.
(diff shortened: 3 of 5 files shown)


auvipy (changes requested):
news entry needed if it is not private.

justbennet:
Is there any chance that the priority of this could get promoted prior to the next release?  This causes the FaultHandler test to segfault when compiling on the Skylake architecture.  This seems somewhat serious, yet it's not been reviewed, and I think perhaps is should not be a type-enhancement but an actual bug?

Adding @ambv here in lieu of sending mail per the release candidate notes at
https://www.python.org/downloads/release/python-380b3/

justbennet:
I just updated the Bug report at https://bugs.python.org/issue21131 to report that this only seems to apply when Python is configured to use shared libraries.  Using
```
$ ./configure --prefix=/path/to/install
```
leads to `make test` passing all tests.  But, using
```
$ ./configure --prefix=/path/to/install --enable-shared
```
leads to the `faulthandler` test failing.  The patch file at the Python bugs issue will fix the problem, and the `faulthandler` test then succeeds.

I am not sure why that is, but it may explain some people's being unable to reproduce the problem.

Python/thread_nt.h
```diff
@@ -496,3 +496,13 @@ PyThread_tss_get(Py_tss_t *key)
...
 }
+
+size_t
+_PyThread_preferred_stacksize()
+{
+    /*
+     * Windows default stack size is 1MB, overridable by the linker in the
+     * executable
+     */
+    return 1024 * 1024;
```
vstinner, line 507: I'm not comfortable with hardcoded values, especially the comment "overridable by the linker in the executable". Maybe we should get the size written by the linker in this case? We don't really need this function for faulthandler, maybe remove the Windows implementation?

Python/thread_pthread.h
```diff
@@ -24,41 +24,58 @@
...
  * exceptions.
  *
  * The default stack sizes below are the empirically determined minimal stack
  * sizes where a simple recursive function doesn't cause a hard crash.
  */
-#if defined(__APPLE__) && defined(THREAD_STACK_SIZE) && THREAD_STACK_SIZE == 0
-#undef  THREAD_STACK_SIZE
+#if defined(__APPLE__)
 /* Note: This matches the value of -Wl,-stack_size in configure.ac */
-#define THREAD_STACK_SIZE       0x1000000
```
vstinner, line 44: Previously, THREAD_STACK_SIZE was redefined if THREAD_STACK_SIZE=0. With your patch, it seems like it's always overriden: THREAD_STACK_SIZE value is no longer tested. I would prefer to minimize changes in this PR.

Include/internal/pycore_thread.h
```diff
@@ -0,0 +1,22 @@
...
+#include <stdint.h>
+
+#ifndef Py_BUILD_CORE
+#  error "this header requires Py_BUILD_CORE define"
+#endif
+
+/*
+ * Find the size of the stack the system uses by default.
+ */
+PyAPI_FUNC(size_t) _PyThread_preferred_stacksize();
```
vstinner:
```suggestion
PyAPI_FUNC(size_t) _PyThread_preferred_stacksize(void);
```

Python/thread_nt.h
```diff
@@ -496,3 +496,13 @@ PyThread_tss_get(Py_tss_t *key)
     SetLastError(error);
     return result;
 }
+
+size_t
+_PyThread_preferred_stacksize()
```
vstinner, line 501:
```suggestion
_PyThread_preferred_stacksize(void)
```

Python/thread_pthread.h
```diff
@@ -872,3 +881,31 @@ PyThread_tss_get(Py_tss_t *key)
     assert(key != NULL);
     return pthread_getspecific(key->_key);
 }
+
+/*
+ * Provide the preferred stack size to use if we need to create a stack and
+ * need to know the actual amount of space to allocate.
+ */
+size_t
+_PyThread_preferred_stacksize()
```
vstinner:
```suggestion
_PyThread_preferred_stacksize(void)
```

vstinner (changes requested):
I'm not comfortable with this change since it modify the core thread module only to fix a faulthandler unit test. faulthandler uses SIGSTKSZ. It's not obvious to me how it relates to the "preferred thread stack size".

Python/thread_nt.h
peadar: Sure - I added it for completeness/documentation, but it's not called. I can remove it.

Python/thread_pthread.h
peadar: The end result is the same, but with less code - Originally THREAD_STACK_SIZE was always defined to 0 if not previously defined, and then was undefined/redefined if it was 0 in response to the platform detection macros. The code here was really just trying to extract the platform's preferred stack size (as reported by pthreads), and to not replicate this between the threads and faulthandler modules. Let me minimize the diff then.

csabella:
The bug tracker issue was resolved with GH-15276, so I am going to close this pull request.

vstinner:
> The bug tracker issue was resolved with GH-15276, so I am going to close this pull request.

Oops, thanks @csabella: I forgot to close this one!

PR #13651, opened 2019-05-29
autoreconf

methane:
Run autoreconf and commit all updated files.

Environment:  Ubuntu 19.04 (autoconf 2.69)


aclocal.m4
```diff
@@ -166,11 +166,11 @@ and $1[]_LIBS to avoid the need to call pkg-config.
 See the pkg-config man page for more details.])
 
 if test $pkg_failed = yes; then
-        AC_MSG_RESULT([no])
+   	AC_MSG_RESULT([no])
         _PKG_SHORT_ERRORS_SUPPORTED
         if test $_pkg_short_errors_supported = yes; then
 	        $1[]_PKG_ERRORS=`$PKG_CONFIG --short-errors --print-errors --cflags --libs "$2" 2>&1`
-        else
+        else
```
methane: trailing space is ugly, but it is generated by `autoreconf`.

methane: I reverted this file because I don't know about how this file is managed well.

PR #13656, opened 2019-05-29
bpo-37088: Add the sys.addpendingcall() function.

1st1:
https://bugs.python.org/issue37088

### DIFF
--- a/Doc/library/sys.rst
+++ b/Doc/library/sys.rst
@@ -43,6 +43,20 @@ always available.
       functions will not see the hook.
 
 
+.. function:: addpendingcall(callback)
+
+   Schedule a *callback* to eventually execute in the main thread.
+
+   A :exc:`RuntimeError` is raised if the *callback* could not be
+   scheduled.
+
+   If *callback* raises an exception it is converted into an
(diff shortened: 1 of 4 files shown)


asvetlov (commented):
I like it but we need more reviewers I guess

asvetlov:
@vstinner what do you think?

asvetlov (approved):
The more I think the more I like it.
`addpendingcall()` is obviously the very low-level API but it can help asyncio a lot for initialization event loop policies and SIGCHLD handling on Linux.
Python allows subscribing on a signal from the main thread only.
asyncio has on-demand initialization strategy for loops and policies.
So, if I want to use asyncio from sub-thread only and want to execute subprocesses a way to execute some code in main thread if I'm now in sub-thread context the feature is pretty cool.

csabella:
This looked like it was ready to be merged, maybe pending some comments from the bug tracker.  @1st1, what do you think?

asvetlov:
@vstinner had objections, perhaps this PR should be rejected.

vstinner:
There is not clear rationale to justify the additon of the function, so I reject the feature. If you are still interesting by the function, please elaborate the rationale in https://bugs.python.org/issue37088

PR #13657, opened 2019-05-29
[3.6] Doc: Add an optional obsolete header. (GH-13638).

JulienPalard:
(cherry picked from commit 46ed90dd014010703c7a3b2a61c4927644fa8210)

Co-authored-by: Julien Palard <julien@palard.fr>

### DIFF
--- a/Doc/tools/templates/layout.html
+++ b/Doc/tools/templates/layout.html
@@ -5,7 +5,7 @@
 <div id="outdated-warning" style="padding: .5em; text-align: center; background-color: #FFBABA; color: #6A0E0E;">
     {% trans %}This document is for an old version of Python that is no longer supported.
     You should upgrade, and read the {% endtrans %}
-    <a href="/3/{{ pagename }}{{ file_suffix }}">{% trans %} Python documentation for the last stable release {% endtrans %}</a>.
+    <a href="/3/{{ pagename }}{{ file_suffix }}">{% trans %} Python documentation for the current stable release {% endtrans %}</a>.
 </div>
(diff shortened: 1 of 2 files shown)


JulienPalard:
@ned-deily Hi Ned, would you please merge this, if you like it? See screenshot here: https://github.com/python/cpython/pull/13638

Doc/tools/templates/layout.html
```diff
@@ -1,4 +1,15 @@
 {% extends "!layout.html" %}
+
+{% block header %}
+{%- if outdated %}
+<div id="outdated-warning" style="padding: .5em; text-align: center; background-color: #FFBABA; color: #6A0E0E;">
+    {% trans %}This document is for an old version of Python that is no longer supported.
+    You should upgrade, and read the {% endtrans %}
+    <a href="/3/{{ pagename }}{{ file_suffix }}">{% trans %} Python documentation for the last stable release {% endtrans %}</a>.
```
ned-deily: Suggest using "most recent" or "current" rather than "last" to avoid ambiguity ("most recent" vs "final").

ned-deily (changes requested):
Are you also planning to add `outdated` to `conf.py`?

JulienPalard:
> Are you also planning to add `outdated` to `conf.py`?

I think I'll add it thrue docsbuild-scripts, not via the file, so for future releases I won't have to commit the boolean to conf.py.

JulienPalard:

ned-deily (approved):
LGTM, thanks!

PR #13661, opened 2019-05-29
bpo-37015: Ensure tasks created by _accept_connection2 due to AsyncMock are completed

tirkarthi:
From 3.8 async functions used with mock.patch return an `AsyncMock`. `_accept_connection2` is an async function where create_task is also mocked. Don't mock `create_task` so that tasks are created out of coroutine returned by `AsyncMock` and the tasks are completed.

https://bugs.python.org/issue37015

### DIFF
--- a/Lib/test/test_asyncio/test_selector_events.py
+++ b/Lib/test/test_asyncio/test_selector_events.py
@@ -364,16 +364,15 @@ class BaseSelectorEventLoopTests(test_utils.TestCase):
         backlog = 100
         # Mock the coroutine generation for a connection to prevent
         # warnings related to un-awaited coroutines. _accept_connection2
-        # is an async function that is patched with AsyncMock. Use MagicMock
-        # since the coroutine is not awaited.
+        # is an async function that is patched with AsyncMock. create_task
+        # creates a task out of coroutine returned by AsyncMock, so use
+        # asyncio.sleep(0) to ensure created tasks are complete to avoid
+        # task pending warnings.
         mock_obj = mock.patch.object
-        with mock_obj(self.loop, '_accept_connection2',
-                      new=mock.MagicMock()) as accept2_mock:
-            accept2_mock.return_value = None
-            with mock_obj(self.loop, 'create_task') as task_mock:
-                task_mock.return_value = None
+        with mock_obj(self.loop, '_accept_connection2') as accept2_mock:
                 self.loop._accept_connection(
                     mock.Mock(), sock, backlog=backlog)
+        self.loop.run_until_complete(asyncio.sleep(0))
         self.assertEqual(sock.accept.call_count, backlog)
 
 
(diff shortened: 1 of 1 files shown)


tirkarthi:
The other way to solve this would be to remove `mock.patch` for `create_task` so that `_accept_connection2` returns an `AsyncMock` and `self.create_task` creates a task from the coroutine. This would require a call like `loop.run_until_complete(asyncio.sleep(0))` to ensure the task is completed I guess. I prefer this since this just restores 3.7 behavior without additional code change but I can change the approach if needed.

Command to reproduce the warnings : 

```
./python.exe -Wall -X tracemalloc -m unittest -vv test.test_asyncio.test_selector_events.BaseSelectorEventLoopTests.test_accept_connection_multiple
```

Lib/test/test_asyncio/test_selector_events.py
```diff
@@ -363,9 +363,12 @@ def test_accept_connection_multiple(self):
...
         backlog = 100
         # Mock the coroutine generation for a connection to prevent
-        # warnings related to un-awaited coroutines.
+        # warnings related to un-awaited coroutines. _accept_connection2
+        # is an async function that is patched with AsyncMock. Use MagicMock
+        # since the coroutine is not awaited.
         mock_obj = mock.patch.object
-        with mock_obj(self.loop, '_accept_connection2') as accept2_mock:
+        with mock_obj(self.loop, '_accept_connection2',
+                      new=mock.MagicMock()) as accept2_mock:
```
mariocj89: Interesting! I usually use new for created objects and `new_callable` when I want a different mock type.

mariocj89 (approved):

asvetlov:
> The other way to solve this would be to remove mock.patch for `create_task`

Please do it. Double mocking for both `_accept_connection2` and `create_task` looks redundant.

P.S.
Thank you very much for taking care of tests.
I'm pretty busy now and very appreciate your help with solving non-urgent but extremely important tasks,

Lib/test/test_asyncio/test_selector_events.py
tirkarthi: This change in behavior has to be documented as noted in the tracker. Currently there is below text where MagicMock is documented as the target type but that changed in 3.8 with `AsyncMock` for async functions and `MagicMock` for normal functions when new is omitted in mock.patch.

https://docs.python.org/3/library/unittest.mock.html#patch

> If new is omitted, then the target is replaced with a MagicMock

tirkarthi:
You're welcome @asvetlov . I have made the requested changes to make sure only `_accept_connection2` is mocked and the tasks are completed. My assumption is that `create_task` creates the task but doesn't execute them and using `asyncio.sleep(0)` there is a control switch to the created tasks without sleeping and they are completed. I have added a comment. Feel free to add suggestions if I have misunderstood something.

Thanks

asvetlov (approved):
LGTM!

PR #13456, opened 2019-05-21
bpo-36983: Fix typing.__all__ and add test for exported names

asottile:
https://bugs.python.org/issue36983

PR #13662, opened 2019-05-29
[3.7] bpo-36983: Fix typing.__all__ and add test for exported names (GH-13456)

asottile:
https://bugs.python.org/issue36983

Fixes issue 36983


asottile:
CC @gvanrossum

gvanrossum (approved):

PR #13664, opened 2019-05-29
Update socketserver.py

geryogam:
Change `BaseServer.__init__(self, *args)` to `super().__init__(*args)`.


ZackerySpytz:
Such changes are usually not accepted by CPython.

tiran:
Thanks for your PR.

However as Zackery already pointed out, we don't land code cleanups like that. It's creating too much noise for no compelling reasons. We clean up when we touch the code for other reasons.

PR #13665, opened 2019-05-29
bpo-36974: Fix GDB integration

encukou:
**buildbot fix for #13185**

As it changes the way functions are called, the PEP 590 implementation
skipped the functions that the GDB integration is looking for
(by name) to find function calls.

Looking for the new helper `cfunction_call_varargs` should fix buildbots.

The changed frame nuber in test_gdb is due to there being fewer
C calls when calling a built-in funciton/method.

### DIFF
--- a/Lib/test/test_gdb.py
+++ b/Lib/test/test_gdb.py
@@ -887,7 +887,7 @@ id(42)
                                           breakpoint='time_gmtime',
                                           cmds_after_breakpoint=['py-bt-full'],
                                           )
-        self.assertIn('#2 <built-in method gmtime', gdb_output)
+        self.assertIn('#1 <built-in method gmtime', gdb_output)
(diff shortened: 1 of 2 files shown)


jdemeyer:
Could you change this comment:
```
        # Tested function must not be defined with METH_NOARGS or METH_O,
        # otherwise call_function() doesn't call PyCFunction_Call()
```

encukou:
I'm planning a follow-up with a better fix; I'll change the comment there. This is an quick patch to fix the buildbots.

jdemeyer:
Thanks a lot (assuming that this actually fixes the buildbots)!

encukou:
It dit! :tada: 

And https://github.com/python/cpython/pull/13668 is that follow-up.

PR #13670, opened 2019-05-30
bpo-36373: Deprecate explicit loop parameter in all public asyncio APIs [task] [WIP]

eamanu:
This PR deprecate explicit loop parameters in all public asyncio APIs

This issues is split to be easier to review.

First step: tasks.py

### DIFF
--- a/Lib/asyncio/tasks.py
+++ b/Lib/asyncio/tasks.py
@@ -116,10 +116,9 @@ class Task(futures._PyFuture):  # Inherit Python Task implementation
         return _all_tasks_compat(loop)
 
     def __init__(self, coro, *, loop=None, name=None):
-        if loop:
-            warnings.warn("The loop argument is deprecated since Python 3.8, "
-                          "and scheduled for removal in Python 3.10.",
-                           DeprecationWarning, stacklevel=2)
+        warnings.warn("The loop argument is deprecated since Python 3.8, "
+                      "and scheduled for removal in Python 3.10.",
+                       DeprecationWarning, stacklevel=2)
         super().__init__(loop=loop)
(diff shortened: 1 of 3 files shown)


Carreau (approved):

tirkarthi (commented):
Please handle `DeprecationWarning` in the tests. Running with `-Werror` would give test failures due to warnings.

asvetlov (changes requested):
Agree with @tirkarthi 
`./python -Werror -m test` should pass
You can find the following command useful in development:
`./python -Werror -m test -v test_asyncio"

Lib/asyncio/tasks.py
```diff
@@ -659,6 +672,10 @@ class _GatheringFuture(futures.Future):
     """
 
     def __init__(self, children, *, loop=None):
+        if loop:
```
asvetlov: This check is redundant: `asyncio.gather()` always passes `loop` to internal `_GatheringFuture` constructor.

asvetlov (changes requested):
Please run the PR with `./python -Werror -m test --check-env-changed -v test_asyncio` and make sure that the test run is succeded.

Lib/asyncio/tasks.py
```diff
@@ -116,6 +116,10 @@ def all_tasks(cls, loop=None):
         return _all_tasks_compat(loop)
 
     def __init__(self, coro, *, loop=None, name=None):
+        if loop:
+            warnings.warn("The loop argument is deprecated since Python 3.8, "
```
asvetlov: `Task` is called with explicit `loop` by `loop.create_task()` method, please drop the check

PR #13671, opened 2019-05-30
bpo-36373:  Deprecate explicit loop parameter in all public asyncio APIs [streams]

eamanu:
This PR deprecate explicit loop parameters in all public asyncio APIs

This issues is split to be easier to review.

Second step: streams.py

### DIFF
--- a/Lib/asyncio/streams.py
+++ b/Lib/asyncio/streams.py
@@ -175,6 +175,10 @@ async def open_connection(host=None, port=None, *,
                   stacklevel=2)
     if loop is None:
         loop = events.get_event_loop()
+    else:
+        warnings.warn("The loop argument is deprecated since Python 3.8, "
+                      "and scheduled for removal in Python 3.10.",
+                      DeprecationWarning, stacklevel=2)
     reader = StreamReader(limit=limit, loop=loop)
     protocol = StreamReaderProtocol(reader, loop=loop, _asyncio_internal=True)
     transport, _ = await loop.create_connection(
@@ -213,6 +217,10 @@ async def start_server(client_connected_cb, host=None, port=None, *,
                   stacklevel=2)
     if loop is None:
         loop = events.get_event_loop()
+    else:
+        warnings.warn("The loop argument is deprecated since Python 3.8, "
+                      "and scheduled for removal in Python 3.10.",
+                      DeprecationWarning, stacklevel=2)
 
     def factory():
         reader = StreamReader(limit=limit, loop=loop)
@@ -414,6 +422,10 @@ if hasattr(socket, 'AF_UNIX'):
                       stacklevel=2)
         if loop is None:
             loop = events.get_event_loop()
+        else:
+            warnings.warn("The loop argument is deprecated since Python 3.8, "
+                          "and scheduled for removal in Python 3.10.",
+                          DeprecationWarning, stacklevel=2)
         reader = StreamReader(limit=limit, loop=loop)
         protocol = StreamReaderProtocol(reader, loop=loop,
                                         _asyncio_internal=True)
@@ -473,6 +485,10 @@ if hasattr(socket, 'AF_UNIX'):
                       stacklevel=2)
         if loop is None:
             loop = events.get_event_loop()
+        else:
+            warnings.warn("The loop argument is deprecated since Python 3.8, "
+                          "and scheduled for removal in Python 3.10.",
+                          DeprecationWarning, stacklevel=2)
 
         def factory():
             reader = StreamReader(limit=limit, loop=loop)
@@ -535,10 +551,6 @@ class FlowControlMixin(protocols.Protocol):
             self._loop = events.get_event_loop()
         else:
             self._loop = loop
-            warnings.warn("The loop argument is deprecated since Python 3.8, "
-                          "and scheduled for removal in Python 3.10.",
-                          DeprecationWarning, stacklevel=2)
-
         if not _asyncio_internal:
             # NOTE:
             # Avoid inheritance from FlowControlMixin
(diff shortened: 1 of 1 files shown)


Lib/asyncio/streams.py
```diff
@@ -535,6 +535,10 @@ def __init__(self, loop=None, *, _asyncio_internal=False):
             self._loop = events.get_event_loop()
         else:
             self._loop = loop
+            warnings.warn("The loop argument is deprecated since Python 3.8, "
```
asvetlov: Please drop this warning.
`FlowControlMixin` is a private API.
Actually, we always pass the loop explicitly in asyncio.

Lib/asyncio/streams.py
```diff
@@ -610,6 +614,10 @@ class StreamReaderProtocol(FlowControlMixin, protocols.Protocol):
 
     def __init__(self, stream_reader, client_connected_cb=None, loop=None,
                  *, _asyncio_internal=False):
+        if loop:
```
asvetlov: Private class, please dro the check

Lib/asyncio/streams.py
```diff
@@ -753,6 +761,9 @@ def __init__(self, limit=_DEFAULT_LIMIT, loop=None):
             self._loop = events.get_event_loop()
         else:
             self._loop = loop
+            warnings.warn("The loop argument is deprecated since Python 3.8, "
```
asvetlov: Private API, people should never create StreamReader class explicitly but use `await open_connection()` and family.

Lib/asyncio/streams.py
```diff
@@ -1297,6 +1308,9 @@ def __init__(self, mode, *,
...
```
asvetlov: Private API

asvetlov (changes requested):
Please remove all deprecations in public API but add these warnings to `open_connection()`, `start_server()`, `open_unix_connection()`, `start_unix_server()` instead.

Lib/asyncio/streams.py
eamanu: Ready, I dropped the check

Lib/asyncio/streams.py
eamanu: removed it

Lib/asyncio/streams.py
eamanu: removed it

Lib/asyncio/streams.py
eamanu: ready

eamanu: ready

eamanu:

eamanu:
@asvetlov the News Could be skipped?

Lib/asyncio/streams.py
```diff
@@ -610,6 +626,9 @@ class StreamReaderProtocol(FlowControlMixin, protocols.Protocol):
 
     def __init__(self, stream_reader, client_connected_cb=None, loop=None,
                  *, _asyncio_internal=False):
+        warnings.warn("The loop argument is deprecated since Python 3.8, "
```
asvetlov: This class is actually *private API*. Please remove the warning, I pretty sure that asyncio factories like `open_connection` always send the `loop`. Double deprecation is redundant.

asvetlov (changes requested):
Please run your PR by `./python -Weeror -m test -v test_asyncio`.
I suspect that some tests will fail (but didn't check it myself).
These tests should be wrapped by `with self.assertWarns(DeprecationWarning): ...` construction

asvetlov (changes requested):
please drop the check here

Lib/asyncio/streams.py
Carreau: If so it should  be documented as private, it does not appear to be : https://docs.python.org/3.6/library/asyncio-stream.html#streamreaderprotocol

And a quick search returns case of people passing the loop argument to it. 
https://github.com/ssfdust/pyclewn/blob/555f2e5485f9938078adfe19388a0e1bb8af67f5/lib/clewn/tty.py#L119

asvetlov: Agree, it should be documented properly.
I've added `_asyncio_internal` flag.
Also, `asincio.__init__` has `__getattr__()` now to raise `DeprecationWarning` for importing `StreamReader`, `StreamWriter` and `StreamReaderProtocol`.
We have an agreement with @1st1 to land asyncio streams code changes first and document them later before Release Candidate stage.

I supposed to make a PR for documentation. If somebody would help me -- I very appreciate this.
Especially taking into account that my English is rough :(
Usually Yuri replaces almost every my sentence on review :)

eamanu: Hi @asvetlov I will help with the docs. :-)

eamanu:

eamanu:
@asvetlov I run the ```./python -Weeror -m test -v test_asyncio``` and the test are ok

```
Ran 2145 tests in 80.981s

OK (skipped=35)
test_asyncio passed in 1 min 21 sec

== Tests result: SUCCESS ==

1 test OK.

Total duration: 1 min 21 sec
Tests result: SUCCESS
```

asvetlov (approved):

asvetlov:
Thank you @eamanu 
Last days I was busy by firefighting with bugs to make 3.8 beta released.

The next big target is asyncio documentation cleanup (plus I need to land multithreaded support back again).

asvetlov:
@eamanu would you take a look on `asyncio/locks.py` and `asyncio/queues.py`?

eamanu:
> @eamanu would you take a look on `asyncio/locks.py` and `asyncio/queues.py`?

@asvetlov  Yes yes, I will start work on that today. I was busy the last weekend. And I have pending #13670

asvetlov:
Thank you. No rush, we have enough time during betas

PR #13672, opened 2019-05-30
bpo-37105: Add deprecated-remove information on stream doc

eamanu:
According to the code on streams.py the functions:
``open_connection()``, ``start_server()``, ``open_unix_connection()``,
``start_unix_server()`` are deprecated. I inform that on
documentation.

https://bugs.python.org/issue37105

### DIFF
--- a/Doc/library/asyncio-stream.rst
+++ b/Doc/library/asyncio-stream.rst
@@ -67,7 +67,7 @@ and work with streams:
 
       The *ssl_handshake_timeout* parameter.
 
-   .. deprecated-removed :: 3.8 3.10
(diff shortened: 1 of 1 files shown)


auvipy (approved):

eamanu:
I think the news could be skipped

I will open a bpo

Carreau (approved):

asvetlov (approved):
The problem is: the new stream API is not documented yet :)
I'm merging this PR but we need to write docs for new streams before release candidate.
Any help is welcome as usual

eamanu:
> The problem is: the new stream API is not documented yet :)
@asvetlov I will take this :-)

PR #13675, opened 2019-05-30
bpo-36610: shutil.copyfile(): use sendfile() on Linux only

giampaolo:
...and avoid using it on Solaris as it can raise EINVAL if offset is equal or bigger than the size of the file, see:
https://bugs.python.org/issue36610

https://bugs.python.org/issue36610


stellarpower:
I have just installed Python 3.10a6 because of this but am still getting this error. I'm inside a Ubuntu LX zone so it's actually coming all the way through into the guest system, I am guessing because the patch is using the OS name and correctly determining the guest to be linux. I'll ask with the Solaris people to see if this is possibly an issue with our LX translation of syscalls.

PR #13676, opened 2019-05-30
bpo-37096: Add large-file tests for modules using sendfile(2)

giampaolo:
https://bugs.python.org/issue37096

https://bugs.python.org/issue37096


asvetlov (approved):

asvetlov:
Looks good when green

PR #13677, opened 2019-05-30
bpo-37098: Skip memfd_create test before Linux 3.17 (GH-13677)

tiran:
https://bugs.python.org/issue37098


tiran (commented):
Let's try the Kernel check first. The test might still fail on containerized platforms that have the memfd syscall restricted.

PR #13678, opened 2019-05-30
bpo-37039: Make IDLE's Zoom Height adjust to users' screens

taleinat:
When "zoom height" is first activated, this implementation measures the window manager's size for maximized windows by maximizing IDLE, and from that point uses that size. Subsequent activations will use the cached value for the dimensions of maximized windows.

This avoids hard-coded values which aren't always correct, e.g. because docks/taskbars can be resized, moved or hidden and window-manager themes can be customized.

This does cause a one-time visual artifact on macOS, where on the first activation, the window is maximized, un-maximized, and then set to the maximum height. Subsequent activation uses cached information, so this doesn't happen again.

https://bugs.python.org/issue37039


aroberge:
On my computer (Windows 10, resolution: 3000 x 2000, scaling of UI elements set to 200%), clicking on Zoom Height makes Idle go full screen. Clicking on the same menu (now called Restore Height) does absolutely nothing: Idle stays in full screen mode.

If I run the program on a secondary monitor (4k), the height only changes, but nowhere enough to be equal to the height of the screen.

csabella:
On Windows 10, I'm seeing the same thing as @aroberge, but there's some more.  I'll enumerate the steps.

1.  Start IDLE.
2.  Zoom Height -- window now takes up full screen.
3.  Restore Height -- nothing happens.
4.  Use Maximize/Minimize from title bar -- window goes back to original size and location.
5.  Zoom Height -- window expands past bottom of screen, but top of the screen doesn't move.  The status bar (and bottom of the editor frame) cannot be seen.
6.  Restore Height -- restores to original size and location.
7.  Moving the window around (that is, where the top of the screen is above or below the original), the window always 'snaps' to the same top, which is where is was originally opened.

Another path:
1.  Start IDLE.
2.  Maximize window using the title bar icon.  
3.  Menu says 'Zoom Height', but selecting it does nothing.
4.  Continue from step 4 above.

This seems like a different bug/enhancement (to not allow zooming when the window is maximized), but I wanted to mention it since I was checking the behavior.

Thanks!

taleinat:
I've fixed three different issues on Windows 10, including those brought up here. I've also verified that it still works as intended on macOS (10.14.4).

@aroberge, @csabella, would you mind giving this another go?

aroberge:
It works correctly on my system. 

It is a bit surprising to see the window taking the full screen for a second and then having its width reduced again, but it only does that when changing the zoom status for the first time.   Moving the window to a second monitor with a different resolution after having clicked on zoom height at least once results in the zoom height feature not working properly as it remembers the settings from the first window.  However, if one first moves the window to a second monitor and then clicks on the button, it works.

taleinat:
> Moving the window to a second monitor with a different resolution after having clicked on zoom height at least once results in the zoom height feature not working properly as it remembers the settings from the first window.

This is true and very unfortunate. Many people use multiple monitors with different resolutions these days, so this should be considered a common scenario.

To avoid this, we could check the screen dimensions and only cache values for screens with identical dimensions. We'll also need to properly handle screens with identical resolutions but different scaling settings.

terryjreedy:
The test failures were all for test_enum.  False positives with respect to this patch.  Since test_idle passed on my machine, no need to retest with close/open -- wait until another commit.

terryjreedy:
There seem to be three issues.

1. Initial full-screen flash:  really annoying.  If we cannot find something we have missed so far (I don't expect this), it might be better to pause it a bit.  Idea: maximize the root window with text such as "Calibrating zoom".  Or do this on startup as a splash screen with text such as "IDLE is measuring\nyour screen."  The latter might require pulling startup code from pyshell into a separate startup file that does not immediately import all of idlelib. I don't currently know where the startup time goes, and if the time between creating root and displaying a shell or editor screen could be made long enough with pyshell as startup.

After this, Zoom and Restore work as they should.

2.  User maximizes the window and then 'zooms'.  Nothing happens as the window is already zoomed and zoom/restore do not change width.  'Zoom' is not changes to 'Restore'.  Perhaps a comparison

taleinat:
The newest change calculates and caches the needed info once per screen resolution.

This will still misbehave slightly in truly rare edge-cases, such as changing the size or position of the taskbar/dock.

taleinat:
> 1. Initial full-screen flash:  really annoying.  If we cannot find something we have missed so far (I don't expect this), it might be better to pause it a bit.  Idea: maximize the root window with text such as "Calibrating zoom".  Or do this on startup as a splash screen with text such as "IDLE is measuring\nyour screen."  The latter might require pulling startup code from pyshell into a separate startup file that does not immediately import all of idlelib. I don't currently know where the startup time goes, and if the time between creating root and displaying a shell or editor screen could be made long enough with pyshell as startup.

IMO the number of users who actually use "zoom height" is small, so I don't think this justifies annoying 100% of the users with a maximized (i.e. almost full-screen) splash screen.

I've been working around and through the limitations of Tk as far as possible, and have reached the limit of what is possible, to the best of my understanding and research. To do better would require using OS APIs directly, e.g. as done by the [screeninfo library](https://pypi.org/project/screeninfo/), but IMO this is not warranted in this case.

> 2. User maximizes the window and then 'zooms'.  Nothing happens as the window is already zoomed and zoom/restore do not change width.  'Zoom' is not changed to 'Restore'.  Perhaps a comparison needs to remove '='?  Perhaps when no calibration is stored, max can be stored and 'zoom' flipped?

I've currently made "zoom height" be a no-op when the window isn't in the "normal" state, e.g. when it is maximized. I can handle the case where the screen is already maximized if you find the current "do nothing" behavior surprising. Or I can just add a `.bell()` call to make it clear that the command failed.

> 3. Moving between a window between multiple screens.  Given multiple screens, how common is this?

These days, with widespread use of laptops, it is very common: People connect and disconnect to/from external screens, using them instead of/in addition to the laptop screen. Also, multi-monitor desktop setups are increasingly common for developers, and often they do not have the same resolution, e.g. having one monitor situated in a "portrait" orientation.

> ... We should be able to identify screens sufficiently by size and scaling.  It does suggest that a startup check is not enough, and that both zoom and scale calibration need to be per window.

The latest commit in the PR handles this properly.

Lib/idlelib/zoomheight.py
```diff
@@ -2,42 +2,121 @@
...
 
 
 class ZoomHeight:
 
     def __init__(self, editwin):
         self.editwin = editwin
 
+        # Cached values for maximized window dimensions, one for each set
+        # of screen dimensions.
+        self._max_height_and_y_coords = {}
```
terryjreedy: Make this a class attribute so we only get one flash per screen, for the first window zoomed.  I think this reduction makes the flashing tolerable.  On my system, the unneeded flashes for additional windows are often more obnoxious.

Lib/idlelib/zoomheight.py
```diff
@@ -2,42 +2,121 @@
...
+
+    @property
```
terryjreedy: I think this should just be  `self.top = self.editwin.top`, which is done once and makes each access faster.  EDIT: done.

As long as there is one ZoomHeight instance per window, setting self.editwin and self.top in __init__ is right.  If we switch to an app menu, there would be only one instance and zoom_height_event would grab editwin from the event.  I believe this can be done now, by changing editor.py, but it should be a separate issue.

Lib/idlelib/zoomheight.py
```diff
@@ -2,42 +2,121 @@
...
+
+        if zoomed is not None:
+            menu_status = 'Restore' if zoomed else 'Zoom'
+            self.editwin.update_menu_label(menu='options', index='* Height',
+                                           label=f'{menu_status} Height')
+
         return "break"
 
+    def zoom_height(self):
+        width, height, x, y = get_window_geometry(self.top)
```
terryjreedy: I strongly prefer having multiple uses of `self.top` reduced to `top`, as it wap before, after adding `top = self.top`. I will do it if you prefer.  Ditto for the next function.
EDIT: both done. Thank you.

Lib/idlelib/zoomheight.py
```diff
@@ -2,42 +2,121 @@
...
+
+        if self.top.wm_state() != 'normal':
+            # Can't zoom/restore window height for windows not in the 'normal'
+            # state, e.g. maximized and full-screen windows.
+            return None
```
terryjreedy, line 41: I now agree that this is correct.  Previously, touching the window when maximized messed up the behavior of the OS WM Restore button (at least on Windows, which is enough).  Besides which, restoring only the height and not the width would be wrong and restoring both would be redundant.  So add `top.beep()`.

This should be added to the doc and blurb.  I plan to edit both when code is settled.

Lib/idlelib/zoomheight.py
```diff
@@ -2,42 +2,121 @@
...
-    # The constants below for Windows and Mac Aqua are visually determined
-    # to avoid taskbar or menubar and app icons.
-    newy, bot_y = ((0, 72) if sys.platform == 'win32' else
-                   (22, 88) if macosx.isAquaTk() else
-                   (0, 88) ) # Guess for anything else.
-    newheight = newheight - newy - bot_y
-    newgeom = '' if height >= newheight else f"{width}x{newheight}+{x}+{newy}"
-    top.wm_geometry(newgeom)
-    return newgeom != ""
+    return width, height, x, y
```
terryjreedy: Just return `map...` and unpack where used.

Lib/idlelib/zoomheight.py
```diff
@@ -2,42 +2,121 @@
...
+
+        return self._max_height_and_y_coords[screen_dimensions]
+
+
+def get_window_geometry(top):
     geom = top.wm_geometry()
     m = re.match(r"(\d+)x(\d+)\+(-?\d+)\+(-?\d+)", geom)
     if not m:
         top.bell()
         return
```
terryjreedy: By PEP 8, this should be `return None` because of explicit return below.  But unpacking None at the call site will crash IDLE anyway.  So we might as well remove the check here.

https://www.tcl.tk/man/tcl8.6/TkCmd/tkvars.htm says
"The following variables are only guaranteed to exist in wish executables; the Tk library does not define them itself but many Tk environments do.
geometry
    If set, contains the user-supplied geometry specification to use for the main Tk window. "

idlelib, including tests, is loaded with unguarded geometry calls, so IDLE and the test suite would fail on any tk without geometry.  So there must be none on systems Python and IDLE currently run on.

terryjreedy (changes requested):
I addressed my points 1 and 2 in the comments.  Without a multiscreen setup, I cannot comment much on 3.  I am willing to have zoomheight be imperfect if someone changes their taskbar or screen settings.  I will add this also to the doc, something like "After the first zoom on a screen, zoomheight assumes that taskbar and screen settings remain the same."

I have not yet looked at the rest of the code of `zoom_height` and that of `get_max...`, but I expect this should get into 3.7.4b1 on 6/17.

Lib/idlelib/zoomheight.py
taleinat: Note that this part of the code is unchanged...

If the existing norm is not guarding against `geometry()` returning bad values, I'll remove the check.

taleinat:

terryjreedy:
@zooba Azure failure on Mac due to running more that 60 minutes
https://dev.azure.com/Python/cpython/_build/results?buildId=44703&view=logs&j=18d1a34d-6940-5fc1-f55b-405e2fba32b1

@taleinat I assume that you consider this essentially ready to merge.  If, after review, I agree, I will add idlelib NEWS entry, which will trigger retest.  We can ignore the Azure failure.

Lib/idlelib/zoomheight.py
terryjreedy: Fixed, thank you.  The one flash is okay.

Lib/idlelib/zoomheight.py
terryjreedy: Fixed, thank you.  The one flash is okay.

Misc/NEWS.d/next/IDLE/2019-06-04-23-27-33.bpo-37039.FN_fBf.rst
```diff
@@ -1 +1,4 @@
-Fix incorrect behavior of IDLE's "Zoom Height" command across platforms.
+Adjust "Zoom Height" zooms to individual screens by momemtarily maximizing
+the window the first time one is zoomed on a particular screen.  Zooms will
+not be correct if the resolution and scaling for a screen is subsequently
+changed.  While a window is maximized, "Zoom Height" has no effect.
```
taleinat: As far as I can tell, changing the resolution and/or scaling of a screen will simply cause Tk to observe a different screen resolution, after which the current implementation will again maximize a screen once to get the proper values for the new screen configuration.  So this is incorrect: "Zooms will not be correct if the resolution and scaling for a screen is subsequently changed."

The only cases where we could get things slightly wrong is when the dock/taskbar is moved or changed, or perhaps moving a window to a screen with identical effective resolution but with other differences, e.g. with/without a dock/taskbar.

taleinat:
@terryjreedy, indeed this appears ready to merge IMO, apart for my comments regarding the NEWS entry.

Misc/NEWS.d/next/IDLE/2019-06-04-23-27-33.bpo-37039.FN_fBf.rst
terryjreedy: Of course.  The dict key will change and not be found.  I will simplify to "Zooms may not be correct if the screen is subsequently changed."  and then merge.

terryjreedy (approved):

PR #13680, opened 2019-05-30
bpo-36999: Add asyncio.Task.get_coro()

agronholm:
https://bugs.python.org/issue36999

### DIFF
--- a/Doc/library/asyncio-task.rst
+++ b/Doc/library/asyncio-task.rst
@@ -844,7 +844,7 @@ Task Object
 
    .. method:: get_coro()
 
-      Return the coroutine object that was passed to the constructor.
+      Return the coroutine object wrapped by the :class:`Task`.
 
       .. versionadded:: 3.8
 
--- a/Doc/library/asyncio-task.rst
+++ b/Doc/library/asyncio-task.rst
@@ -842,6 +842,12 @@ Task Object
       The *file* argument is an I/O stream to which the output
       is written; by default output is written to :data:`sys.stderr`.
 
+   .. method:: get_coro()
+
+      Return the coroutine object that was passed to the constructor.
+
+      .. versionadded:: 3.8
+
    .. method:: get_name()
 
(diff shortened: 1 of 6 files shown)


Doc/library/asyncio-task.rst
```diff
@@ -842,6 +842,12 @@ Task Object
       The *file* argument is an I/O stream to which the output
       is written; by default output is written to :data:`sys.stderr`.
 
+   .. method:: get_coro()
+
+      Return the coroutine object that was passed to the constructor.
```
asvetlov: I dislike mentioning `constructor` here.

Above we have the text:
```
   Use the high-level :func:`asyncio.create_task` function to create
   Tasks, or the low-level :meth:`loop.create_task` or
   :func:`ensure_future` functions.  Manual instantiation of Tasks
   is discouraged.
```
Maybe rephrase this line to `Return the coroutine object that was used for the task creation.` or something like this?
Sorry, I'm not a native speaker. Don't insist on the exact wording but you've got my idea.

1st1: I'd just say this: "Return the coroutine object wrapped by the Task".

1st1 (approved):

Doc/library/asyncio-task.rst
asvetlov:
```suggestion
      Return the coroutine object wrapped by the :class:`Task`.
```
@1st1 like this?

1st1: Yeah, I think it's great.  If you have time just merge Alex's PR and we can fix this ourselves. Up to you.

agronholm: I changed the wording now as requested.

asvetlov: Great!

asvetlov (approved):

asvetlov:
Thanks!

PR #9296, opened 2018-09-14
bpo-26467: Adds AsyncMock for asyncio Mock library support

lisroach:
This is my initial pass at supporting coroutine mocking via a new Mock subclass, CoroutineMock.

It can be used to mock out coroutines and have them validate as coroutines:

```python
>>> mock = CoroutineMock()
>>> asyncio.iscoroutinefunction(mock)
True
```

Test that a coroutine was awaited:

```python
class TestTests(unittest.TestCase):
    def test_assert_awaited(self):
        async def main():
            mock = CoroutineMock()
            with self.assertRaises(AssertionError):
                mock.assert_awaited()
            await mock()
            mock.assert_awaited()
        asyncio.run(main())
```
Also awaited_with, awaited_once_with, etc.

Things that I could use advice on:

1. Some of the code has been borrowed by asynctest (https://pypi.org/project/asynctest/) which is licensed under Apache 2. I have reached out to the author and am waiting to hear back, but am not sure how we feel about having Apache 2 licensed stuff in CPython. I could rewrite those spots if needed.

2. `inspect.iscoroutine` will return False for CoroutineMock objects. This is because the check is `isinstance(object, types.CoroutineType)`, and CoroutineTypes are concrete. I've looked into using something like, register, but it doesn't work here, so I need someway to make a Mock look like a types.CoroutineType but I am unsure how.

3. In CoroutineMockAssert tests, I have `asyncio.events._event_loop_policy` unset because it causes the env to change when running tests if it is not unset. There is probably a better way to do this where the environment doesn't get polluted?

4. I plan on working on getting more example test cases for the CoroutineArguments test section, would be happy for advice though.

And I will be writing up the documentation once the code has settled.

https://bugs.python.org/issue26467

PR #13681, opened 2019-05-30
docs: Add references to AsyncMock in unittest.mock.patch

mariocj89:
Update the docs as patch can now return an AsyncMock if the patched object is an async function.

Just noticed this, happy to create a bug report & news entry but not sure if it is worth it.

CC: @lisroach @tirkarthi @cjw296 

Relates-to: https://github.com/python/cpython/pull/9296

### DIFF
--- a/Doc/library/unittest.mock.rst
+++ b/Doc/library/unittest.mock.rst
@@ -1307,8 +1307,10 @@ patch
     is patched with a *new* object. When the function/with statement exits
     the patch is undone.
 
-    If *new* is omitted, then the target is replaced with a
-    :class:`MagicMock`. If :func:`patch` is used as a decorator and *new* is
+    If *new* is omitted, then the target is replaced with an
+    :class:`AsyncMock` if the patched object is an async function or
+    a :class:`MagicMock` otherwise.
+    If :func:`patch` is used as a decorator and *new* is
     omitted, the created mock is passed in as an extra argument to the
     decorated function. If :func:`patch` is used as a context manager the created
     mock is returned by the context manager.
@@ -1326,8 +1328,8 @@ patch
     patch to pass in the object being mocked as the spec/spec_set object.
 
     *new_callable* allows you to specify a different class, or callable object,
-    that will be called to create the *new* object. By default :class:`MagicMock` is
-    used.
+    that will be called to create the *new* object. By default :class:`AsyncMock`
+    is used for async functions and :class:`MagicMock` for the rest.
 
     A more powerful form of *spec* is *autospec*. If you set ``autospec=True``
     then the mock will be created with a spec from the object being replaced.
@@ -1490,6 +1492,10 @@ work as expected::
     ...     assert sys.non_existing_attribute == 42
     ...
     >>> test()
+
+.. versionchanged:: 3.8
+
+    :func:`patch` now returns an :class:`AsyncMock` if the target is an async function.
 
 
 patch.object
@@ -2273,6 +2279,12 @@ create_autospec
 
 See :ref:`auto-speccing` for examples of how to use auto-speccing with
 :func:`create_autospec` and the *autospec* argument to :func:`patch`.
+
+
+.. versionchanged:: 3.8
+
+    :func:`create_autospec` now returns an :class:`AsyncMock` if the target is
+    an async function.
 
 
 ANY
--- a/Lib/unittest/mock.py
+++ b/Lib/unittest/mock.py
@@ -1598,8 +1598,9 @@ def patch(
     is patched with a `new` object. When the function/with statement exits
     the patch is undone.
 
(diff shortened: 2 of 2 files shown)


Doc/library/unittest.mock.rst
```diff
@@ -1307,8 +1307,10 @@ patch
     is patched with a *new* object. When the function/with statement exits
     the patch is undone.
 
-    If *new* is omitted, then the target is replaced with a
-    :class:`MagicMock`. If :func:`patch` is used as a decorator and *new* is
+    If *new* is omitted, then the target is replaced with an
+    :class:`AsyncMock` if the patched object is an async function or
```
cjw296, line 1311: Is it just async functions? What about async methods? Is there such a thing as an async `__init__` or `__new__`?

mariocj89: This is what it is controlling it: https://github.com/python/cpython/blob/master/Lib/unittest/mock.py#L49 It is checking for `__code__` so it seems to be only for functions at the moment.

[I was wondering about that](https://github.com/python/cpython/pull/13574#discussion_r287576935) for callable class. My main concern was what if I change a function by a callable instance but it seems there is no good way to check that at the moment. [`asyncio.iscoroutinefunction` works only for functions](https://github.com/python/cpython/pull/13574#discussion_r287586606).

cjw296: Hmm, doesn't seem sensible to make changes to the docs until the rest of this is bottomed out.

mariocj89: Note this is landed already and docs are updated in some other places. The documentation for "patch" does not do what it says it does for async functions right now. Also, I am not sure at all this is going to change, why do you think so?

tirkarthi:
Thank you @mariocj89 for this. I had the same concern while fixing warnings caused due to this change in behavior in the testsuite : https://bugs.python.org/issue37015#msg343352 . I am okay with this being an AsyncMock because it means that there was a coroutine created and it was not awaited where raising a warning makes sense and the user can resort to using MagicMock if they are want it to be not awaited. I would prefer thoughts from @asvetlov @1st1 to discuss this behavior change from 3.7 to ensure the documentation is also updated and if needed an issue for this do discuss. I believe create_autospec has similar behavior : https://github.com/python/cpython/blob/98ef92002ec289bf8086b0ef3d4f96c2589f4e68/Lib/unittest/mock.py#L2612

cjw296:
I guess my reticence on this comes party from the fact that the AsyncMock stuff is something I'm still quite unsure about. If I'm honest, I was surprised to see the PR landed when it did and I'm a little concerned about unintended consequences of the stuff that has been added, along with the chance that this now means the backport is pretty much stuck until it can ditch support for all python versions that don't have the required async syntax.

mariocj89:
> ensure the documentation is also updated 

Good point, I'll add a "versionadded" tag for the patch function.

> I believe create_autospec has similar behavior :

Good point, I'll add a "versionadded" to mark the change.

> I guess my reticence on this comes party from the fact that the AsyncMock stuff is something I'm still quite unsure about. 

If you are saying that you are looking at reverting this behaviour soon, then sure, but otherwise we have out of date documentation, as beta 1 of 3.8 is getting released tomorrow. Not saying that we have to rush things, but people (from tomorrow) won't be seeing this change in the docs.

Even if anything was reverted, we can revert the docs and implementation together at the same time.

auvipy (approved):

cjw296:
@mariocj89 - reverting would be quite an extreme move, and one I'm not comfortable taking, particularly with my very limited knowledge of async stuff.  @asvetlo merged the original PR, IIRC, and it sounded like he's part of a team that had been using this already. 

I'm not sure what the best way to bottom out the remaining issues might be, including the questions I asked above.

asvetlov:
I'll review the PR just after the feature freeze.
Sorry, I'm pretty busy before this date (tomorrow, actually).

mariocj89:
Ping @asvetlov

This is a change in 3.8 that will go undocumented otherwise.

Doc/library/unittest.mock.rst
lisroach: This looks like correct wording to me, perhaps we will find a way to make the callable check a little more granular but there is no reason to block this PR on what might be a future change in implementation.

lisroach (approved):
I think this documentation looks good, @asvetlov I'll wait to merge so you have a chance to take a look as well.

asvetlov (approved):
LGTM, please merge.
```
:class:`AsyncMock`
``` 
doesn't exist in the module documentation though.

lisroach:
I thought I did @bedevere-bot :(

cjw296:
This catches me just about every time too :'(

PR #13682, opened 2019-05-30
bpo-36974: add some assertions for PEP 590

jdemeyer:


jdemeyer:
CC @encukou

auvipy (approved):

encukou:
Is it better to have the checks here than in `_PyType_CheckConsistency`?

encukou:
The answer is in the comment – "we require this before inheriting". I now agree that's a good thing to do.

PR #13687, opened 2019-05-30
bpo-37104: docs: Add docs about the `Logger.disabled` attribute

mariocj89:
Add documentation about the attribute that allows disabling logging for the specific logger.

I can add a news-entry but I really think it is not worth it for this change.

Should this be backported to previous versions? The disabled attribute has been in Logger since the beginning.

https://bugs.python.org/issue37104

### DIFF
--- a/Doc/library/logging.rst
+++ b/Doc/library/logging.rst
@@ -352,6 +352,15 @@ is the module's name in the Python package namespace.
    .. versionchanged:: 3.7
       Loggers can now be pickled and unpickled.
 
+   .. attribute:: Logger.disabled
+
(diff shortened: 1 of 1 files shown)


auvipy (approved):

vsajip:
I'd like to avoid documenting this attribute, as it's really meant for internal use by the configuration machinery. I realise that it's named `disabled` rather than `_disabled`, which might indicate that it's public, but it isn't meant to be, and the name can't be changed now for compatibility reasons, in case some code somewhere is relying on it! I don't want someone to ask for "disabled" to be explicitly supported in configurations, for instance - I can currently argue that it's not meant to be public, as it isn't documented.

PR #13688, opened 2019-05-30
bpo-12639: msilib.Directory.start_component() fails if *keyfile* is not None

ZackerySpytz:
msilib.Directory.start_component() was passing an extra argument to
CAB.gen_id().

https://bugs.python.org/issue12639

### DIFF
--- a/Lib/msilib/__init__.py
+++ b/Lib/msilib/__init__.py
@@ -273,7 +273,7 @@ class Directory:
         if AMD64:
             flags |= 256
         if keyfile:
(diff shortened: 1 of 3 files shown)


auvipy (approved):

pablogsal:
This is possibly the cause of the following leak:

https://buildbot.python.org/all/#/builders/132/builds/505/steps/4/logs/stdio
https://buildbot.python.org/all/#/builders/33/builds/604

PR #13696, opened 2019-05-31
Update data model docs to include missing attributes for code objects

pablogsal:


mariocj89 (commented):
Should we also add keyword only?

pablogsal:
Done!

mariocj89 (approved):
I'd change the commit title on squash to incolde co_kwonlyargcount  as well.

PR #13698, opened 2019-05-31
bpo-37115: Support annotations in positional-only arguments

pablogsal:
https://bugs.python.org/issue37115


mariocj89 (approved):
[🙄](https://www.python.org/dev/peps/pep-0007/#code-lay-out) (Click me :P)

pablogsal:
> [🙄](https://www.python.org/dev/peps/pep-0007/#code-lay-out) (Click me :P)

In general is also important to maintain internal consistency (also explained in PEP7), so or we change the style in all surroundings or we just blend with whatever it is/

PR #13699, opened 2019-05-31
bpo-36974: Make tp_call=PyVectorcall_Call work for inherited types

encukou:
When inheriting a heap subclass from a vectorcall class that sets `.tp_call=PyVectorcall_Call` (as recommended), the subclass does not inherit `_Py_TPFLAGS_HAVE_VECTORCALL`, and thus `PyVectorcall_Call` does not work for it.

This attempts to solve the issue by:
* always inheriting `tp_vectorcall_offset` unless `tp_call` is overridden in the subclass
* inheriting _Py_TPFLAGS_HAVE_VECTORCALL for static types, unless `tp_call` is overridden
* making `PyVectorcall_Call` ignore `_Py_TPFLAGS_HAVE_VECTORCALL`

This means it'll be ever more important to only call `PyVectorcall_Call` on classes that support vectorcall. In `PyVectorcall_Call`'s intended role as `tp_call` filler, that's not a problem.

cc @jdemeyer (not Mark, as he seemed to want to stay out of subclassing – but he's nosy on the issue)

### DIFF
--- a/Lib/test/test_capi.py
+++ b/Lib/test/test_capi.py
@@ -515,9 +515,10 @@ class TestPEP590(unittest.TestCase):
 
     def test_vectorcall(self):
         # Test a bunch of different ways to call objects:
-        # 1. normal call
-        # 2. vectorcall using _PyObject_Vectorcall()
-        # 3. vectorcall using PyVectorcall_Call()
+        # 1. vectorcall using PyVectorcall_Call()
+        #   (only for objects that support vectorcall directly)
+        # 2. normal call
+        # 3. vectorcall using _PyObject_Vectorcall()
         # 4. call as bound method
         # 5. call using functools.partial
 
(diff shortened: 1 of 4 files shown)


jdemeyer:
I don't have much time to look into this, but your idea sounds reasonable.

jdemeyer:
Reviewing this now after the fact... looks mostly good to me.

One potential issue: it is safer to inherit `tp_vectorcall_offset` unconditionally, even if `tp_call` is overridden. This allows a custom `tp_call` implementation falling back to the `tp_call` of the base class. Essentially the C equivalent of
```
def __call__(self, *args, **kwds):
    # do something interesting here
    return super().__call__(*args, **kwds)
```
Do you agree?

And one minor detail: in `_PyVectorcall_Function()`, the check `Py_TYPE(callable)->tp_call` is an assertion `assert(PyCallable_Check(callable));` so I wonder why you made it a real check here.

encukou:
> it is safer to inherit `tp_vectorcall_offset` unconditionally

Yes, that's a good point. Do you want to write the test for that?

> in `_PyVectorcall_Function()`, the check `Py_TYPE(callable)->tp_call` is an assertion `assert(PyCallable_Check(callable));`

Ah, that's a remnant of my investigations that I didn't adjust back to match `_PyVectorcall_Function` again. I don't feel strongly about this.

PR #13707, opened 2019-05-31
bpo-36842: Pass positional only parameters to code_new audit hook

pablogsal:
https://bugs.python.org/issue36842


zooba (approved):
Approved, though only because we haven't shipped this event before. In future we'd have to create a new event name and raise both here.

pablogsal:
Thanks @zooba for the review!

PR #12360, opened 2019-03-15
bpo-33608: Factor out a private, per-interpreter _Py_AddPendingCall().

ericsnowcurrently:
This is effectively an un-revert of #11617 and #12024 (reverted in #12159).  Portions of those were merged in other PRs (with lower risk) and this represents the remainder.  Note that I found 3 different bugs in the original PRs and have fixed them here.

@vstinner, would you mind running this against that VM you set up, to see if it causes the original problem?  I expect it won't.

https://bugs.python.org/issue33608

PR #12159, opened 2019-03-04
Revert: bpo-33608: Factor out a private, per-interpreter _Py_AddPendingCall(). (GH-11617)

vstinner:
* Revert "[bpo-33608](https://bugs.python.org/issue33608): Simplify ceval's DISPATCH by hoisting eval_breaker ahead of time. (gh-12062)"
* Revert "[bpo-33608](https://bugs.python.org/issue33608): Use _Py_AddPendingCall() in _PyCrossInterpreterData_Release(). (gh-12024)"
* Revert "[bpo-33608](https://bugs.python.org/issue33608): Factor out a private, per-interpreter _Py_AddPendingCall(). (GH-11617)"    

https://bugs.python.org/issue33608

PR #13714, opened 2019-06-01
bpo-33608: Factor out a private, per-interpreter _Py_AddPendingCall().

ericsnowcurrently:
This is effectively an un-revert of #11617 and #12024 (reverted in #12159). Portions of those were merged in other PRs (with lower risk) and this represents the remainder. Note that I found 3 different bugs in the original PRs and have fixed them here.

I also tried again (with some tweaks) in #12360, which I ended up reverting in #12806.  For this attempt I am dropping the code that calls any remaining pending calls during runtime finalization.  We'll see if the FreeBSD buildbot still hates me. :)

https://bugs.python.org/issue33608


csabella:
Looks like the Mac CI's hate you now.  😉

PR #13717, opened 2019-06-01
bpo-36411: Python 3 f.tell() gets out of sync with file pointer in binary append+read mode

websurfer5:
Implement buffered tell() as a wrapper for buffered seek(0, io.SEEK_CUR) to fix a problem with the internally maintained file position getting out of sync with the kernel when using read/write append mode. The fix applies to both binary and text files.

https://bugs.python.org/issue36411


auvipy (approved):

Modules/_io/bufferedio.c
```diff
@@ -1301,6 +1287,22 @@ _io__Buffered_seek_impl(buffered *self, PyObject *targetobj, int whence)
...
+       around _io__Buffered_seek_impl() because it
+       keeps the location information properly
+       updated. The optimizations that had previously
+       been implemented are prone to get out of sync
+       with the kernel when the file has been opened
+       in append mode. See issue #36411.
+    */
+
+    CHECK_INITIALIZED(self)
+    return _io__Buffered_seek_impl(self, PyLong_FromLong(0), SEEK_CUR);
```
ZackerySpytz: For what it's worth, I believe `PyLong_FromLong(0)` should be checked for failure (even though this call should never fail due to the current implementation of `PyLong_FromLong()`).

websurfer5: > For what it's worth, I believe `PyLong_FromLong(0)` should be checked for failure (even though this call should never fail due to the current implementation of `PyLong_FromLong()`).

I agree, and made the change.

Modules/_io/bufferedio.c
```diff
@@ -1301,6 +1287,25 @@ _io__Buffered_seek_impl(buffered *self, PyObject *targetobj, int whence)
...
+       around _io__Buffered_seek_impl() because it
+       keeps the location information properly
+       updated. The optimizations that had previously
+       been implemented are prone to get out of sync
+       with the kernel when the file has been opened
+       in append mode. See bpo-36411.
+    */
+
+    CHECK_INITIALIZED(self)
+    PyObject *z_obj = PyLong_FromLong(0);
```
benjaminp, line 1303: This looks like it leaks a reference to `z_obj`.

benjaminp (changes requested):
Won't this break calling `tell()` on buffered streams wrapping raw io that supports `tell()` but not `seek()`?

pewscorner:
Wouldn't it be possible to use the old buffered_tell() code if the stream doesn't support seek()? As far as I know, the bug only happens after seeking, so it ought to be fine to fall back to the old approach when seeking isn't possible, I would imagine.

StanFromIreland:
Closing as stale, the requested changes have not been applied in seven years.

PR #13719, opened 2019-06-01
bpo-37120: Add SSLContext.num_tickets (GH-13719)

tiran:

https://bugs.python.org/issue37120

### DIFF
--- a/Doc/library/ssl.rst
+++ b/Doc/library/ssl.rst
@@ -1959,6 +1959,19 @@ to speed up repeated connections from the same clients.
 
    .. versionadded:: 3.7
 
+.. attribute:: SSLContext.num_tickets
+
+   Control the number of TLS 1.3 session tickets of a
+   :attr:`TLS_PROTOCOL_SERVER` context. The setting has no impact on TLS
+   1.0 to 1.2 connections.
+
+   .. note::
+
+     This attribute is not available unless the ssl module is compiled
+     with OpenSSL 1.1.1 or newer.
+
+   .. versionadded:: 3.8
+
 .. attribute:: SSLContext.options
 
    An integer representing the set of SSL options enabled on this context.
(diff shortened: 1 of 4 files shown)


Doc/library/ssl.rst
```diff
@@ -1959,6 +1959,18 @@ to speed up repeated connections from the same clients.
...
+
+   Control the number of TLSv1.3 session tickets of a
+   :attr:`TLS_PROTOCOL_SERVER` context.
+
+   .. note::
+
+     This attribute is not available unless the ssl module is compiled
+     with OpenSSL 1.1.1 or newer.
+
+   .. versionadded:: 3.8
```
njsmith, line 1973: Probably worth noting explicitly that it only affects TLS 1.3 connections, and possibly that it doesn't affect connections that are resumed using a ticket.

Lib/test/test_ssl.py
```diff
@@ -1634,6 +1634,24 @@ class MySSLObject(ssl.SSLObject):
...
+        self.assertEqual(ctx.num_tickets, 1)
+        ctx.num_tickets = 0
+        self.assertEqual(ctx.num_tickets, 0)
+        with self.assertRaises(ValueError):
+            ctx.num_tickets = -1
+        with self.assertRaises(TypeError):
+            ctx.num_tickets = None
+
+        ctx = ssl.SSLContext(ssl.PROTOCOL_TLS_CLIENT)
+        self.assertEqual(ctx.num_tickets, 2)
```
njsmith, line 1651: It's maybe slightly nicer to make this error out too, I guess?

njsmith (commented):
Two small comments.

Doc/library/ssl.rst
tiran: SGTM, the first sentence mentions that it's about TLS 1.3. I added another sentence to explicitly state that the setting does not affect TLS 1.0 to 1.2.

Lib/test/test_ssl.py
tiran: You mean getattr access? I'm not sure how much trouble it's going to create.

PR #13721, opened 2019-06-01
Fix documentation

bentyeh:
Based on the source code https://github.com/python/cpython/blob/4a686504eb2bbf69adf78077458508a7ba131667/Lib/multiprocessing/pool.py#L755 AsyncResult.successful() raises a ValueError, not an AssertionError.

### DIFF
--- a/Doc/library/multiprocessing.rst
+++ b/Doc/library/multiprocessing.rst
@@ -2277,7 +2277,11 @@ with the :class:`Pool` class.
    .. method:: successful()
 
       Return whether the call completed without raising an exception.  Will
-      raise :exc:`ValueError` if the result is not ready.
+      raise :exc:`AssertionError` if the result is not ready.
+      
+      .. versionchanged:: 3.7
+         If the result is not ready, :exc:`ValueError` is raised instead of
+         :exc:`AssertionError`.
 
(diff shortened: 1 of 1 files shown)


the-knights-who-say-ni:

mangrisano:
I'm not a core-dev but it sounds good to me. I don't know if it is required to open an issue on the tracker. @csabella what's the behavior in this case? Thank you and sorry for the mention!

mangrisano:
I'm not a core-dev but it looks good to me. I don't know if it is required to open an issue on the tracker. @csabella what's the behavior in this case? Thank you and sorry for the mention!

tirkarthi (approved):
LGTM. It seems similar changes were done in bd73e72b4a9f019be514954b1d40e64dc3a5e81c to change assert statements and use proper exceptions and messages though docs were not updated. Thanks.

tirkarthi (commented):
Please add a `versionchanged` directive along with this change since it was changed in 3.7 and above with versions below 3.7 raising `AssertionError` and it's clarified better that it is 3.7+ only change.

asvetlov (approved):
LGTM.

I'd like to postpone the merging until "Backport to 3.8" github label is added.

PR #13727, opened 2019-06-01
bpo-36993: Improve error detection of extra field in ZipFile

berkerpeksag:
https://bugs.python.org/issue36993


serhiy-storchaka (commented):
An IndexError can still be raised if file_size, compress_size and header_offset are all set to 0xffffffff and ln < 24.

berkerpeksag:
Superseded by https://github.com/python/cpython/pull/14656.

PR #13729, opened 2019-06-01
bpo-37126: Allow structseq objects to be tracked by the GC

pablogsal:
https://bugs.python.org/issue37126

`structseq` objects can be involved in reference cycles but they are not tracked by the gc. One example of such case is:
https://bugs.python.org/issue37126


pablogsal:
The `threading` leak is solved by this PR:

```
❯ ./python -m test test_threading -R :
Run tests sequentially
0:00:00 load avg: 1.45 [1/1] test_threading
beginning 9 repetitions
123456789
.........
test_threading passed in 1 min 32 sec

== Tests result: SUCCESS ==

1 test OK.

Total duration: 1 min 32 sec
Tests result: SUCCESS
```

nascheme:
These changes LGTM.

PR #1927, opened 2017-06-03
bpo-30555: Fix WindowsConsoleIO errors in the presence of fd redirection

segevfiner:
This works by not caching the handle and instead getting the handle from the file descriptor each time, so that if the actual handle changes by fd redirection closing/opening the console handle beneath our feet, we will keep working correctly.

I think I also fixed some resource handling issues along the way since ucrt/msvcrt fds *own* the HANDLEs they use and this module tried to close them itself.

See https://github.com/pytest-dev/py/issues/103 for the issue it causes in Pytest. Also see: https://github.com/pytest-dev/pytest/pull/2462. I was able to run the Pytest test suite, and my own little test described in the PR, to completion after this change.

Having your console handle closed by fd redirection is likely a change in the ucrt/msvcrt or a change in newer Windows versions (Console handles were rewritten in Windows 8).

This likely also affects any other module caching `GetStdHandle` or `_get_osfhandle`. For example colorama (https://github.com/pytest-dev/pytest/issues/2465 - https://github.com/tartley/colorama/pull/131), and I can think of "pyreadline", "win-unicode-console" and "click" probably suffering from the same issue but I haven't tested them.

I'm also not sure of the performance implications of this change.

This affects Python 3.6-3.7

P.S. `python -m test -R 3:2 test.test_winconsoleio` reports ref leaks on master for me 😛

CC @zooba @eryksun

- bpo-30555

PR #13739, opened 2019-06-02
gh-77046: Pass the _O_NOINHERIT flag to _open_osfhandle() calls

ZackerySpytz:
On Windows, do not use inheritable file descriptors to wrap handles
in os.pipe().

https://github.com/python/cpython/issues/77046

* Issue: gh-77046

### DIFF
--- a/Misc/NEWS.d/next/Core and Builtins/2019-06-01-23-55-50.bpo-32865.Ha11I1.rst
+++ b/Misc/NEWS.d/next/Core and Builtins/2019-06-01-23-55-50.bpo-32865.Ha11I1.rst
@@ -1,0 +1,2 @@
+On Windows, do not use inheritable file descriptors to wrap handles in
+:func:`os.pipe`.
--- a/Modules/_io/winconsoleio.c
+++ b/Modules/_io/winconsoleio.c
@@ -478,9 +478,11 @@ _io__WindowsConsoleIO_fileno_impl(winconsoleio *self)
     if (self->fd < 0 && self->handle != INVALID_HANDLE_VALUE) {
         _Py_BEGIN_SUPPRESS_IPH
         if (self->writable)
-            self->fd = _open_osfhandle((intptr_t)self->handle, _O_WRONLY | _O_BINARY);
+            self->fd = _open_osfhandle((intptr_t)self->handle,
+                                       _O_WRONLY | _O_BINARY | _O_NOINHERIT);
         else
-            self->fd = _open_osfhandle((intptr_t)self->handle, _O_RDONLY | _O_BINARY);
+            self->fd = _open_osfhandle((intptr_t)self->handle,
+                                       _O_RDONLY | _O_BINARY | _O_NOINHERIT);
         _Py_END_SUPPRESS_IPH
     }
     if (self->fd < 0)
--- a/Modules/posixmodule.c
+++ b/Modules/posixmodule.c
@@ -9200,8 +9200,8 @@ os_pipe_impl(PyObject *module)
     _Py_BEGIN_SUPPRESS_IPH
     ok = CreatePipe(&read, &write, &attr, 0);
     if (ok) {
-        fds[0] = _open_osfhandle((intptr_t)read, _O_RDONLY);
-        fds[1] = _open_osfhandle((intptr_t)write, _O_WRONLY);
+        fds[0] = _open_osfhandle((intptr_t)read, _O_RDONLY | _O_NOINHERIT);
+        fds[1] = _open_osfhandle((intptr_t)write, _O_WRONLY | _O_NOINHERIT);
         if (fds[0] == -1 || fds[1] == -1) {
             CloseHandle(read);
             CloseHandle(write);


ZackerySpytz:
This change should have a unit test, though it is difficult to create one that doesn't cause debug assertion failures in the MSVC runtime library when run.

auvipy (approved):

eryksun:
> This change should have a unit test, though it is difficult to create one that doesn't cause debug assertion failures in the MSVC runtime library when run.

Do you mean the dialog box used by the debug CRT for warnings, errors, and failed assertions, such as a failed assertion if an invalid fd is passed to `msvcrt.get_osfhandle`? That can be suppressed with the context manager [`support.SuppressCrashReport`](https://github.com/python/cpython/blob/v3.8.0a4/Lib/test/support/__init__.py#L2648). 

(It also calls `SetErrorMode` to suppress the WER dialog from an unhandled OS exception -- if the system is configured to show it. An issue should be opened to update this test helper to use `msvcrt.SetErrorMode` instead of ctypes. We could also add support to msvcrt for `GetErrorMode`, `GetThreadErrorMode` and `SetThreadErrorMode`.)

ZackerySpytz:
Thank you for the comment, @eryksun.

> Do you mean the dialog box used by the debug CRT for warnings, errors, and failed assertions, such as a failed assertion if an invalid fd is passed to msvcrt.get_osfhandle?

Yes.

> An issue should be opened to update this test helper to use msvcrt.SetErrorMode instead of ctypes.

Why exactly? Such a change is not likely to be accepted by CPython unless there is an actual bug to be fixed.

eryksun:
> > Do you mean the dialog box used by the debug CRT for warnings, errors, and failed assertions, 
> 
> Yes.

Okay, so a unit test can be added to ensure the fd isn't inherited as long as we suppress these warnings. 

Note that fixing this means that `os.set_inheritable` can no longer enable inheriting these pipe fds via `os.spawn*`.  Maybe a note should be added to explain that there's no simple way in Windows to make the fds inheritable. It's possible, but convoluted. We have to `dup` the fd; `dup2` the result back to the original fd; and then close the duped fd.  Once it's made inheritable, there's no way to make the file descriptor properly non-inheritable again. We can only make the underlying OS handle non-inheritable and hope no other file takes that handle value in the child process before the CRT startup code gets a chance to validate the fd. Applications should be encouraged to switch to the subprocess module, which is based on OS handles, instead of file descriptors.

---

> > An issue should be opened to update this test helper to use msvcrt.SetErrorMode instead of 
>> ctypes.
> 
> Why exactly? Such a change is not likely to be accepted by CPython unless there is an actual 
>bug to be fixed.

There's no reason to use ctypes if we already have a builtin module that provides the function. Dependence on ctypes in the unit tests should be minimized if possible. One day we might want to make it optional in Windows as well. For example, say if Windows supports an architecture that libffi doesn't support or doesn't support for the required calling convention. 

---

Another issue to be addressed is the Windows implementation of `os.dup` (`_Py_dup` in Python/fileutils.c). Someone has a peculiar idea that handles for character (`FILE_TYPE_CHAR`) files, including console files, cannot be made non-inheritable. Of course they can. Due to this misunderstanding, if we open "NUL", "COM1", "CON", "CONIN$", "CONOUT$", or any other character device, and subsequently `dup` the fd, the duplicated fd and/or handle is inheritable via `os.system`, `os.spawn*`, or `subprocess.Popen` with `close_fds=False`.

Maybe the behavior of console applications has confused the author of this code. When a console process is created, if its console is inherited and if its startup info doesn't have the flag `STARTF_USESTDHANDLES`, then the parent's standard handles (`StandardInput`, `StandardOutput`, and `StandardError`) are *duplicated* to it, not inherited. The handles don't have to be inheritable, and the resulting handle values may be different in the child. It's okay since standard handles are referenced via the process parameters in the PEB. Subsequently the CRT initializes fds 0-2 from the standard handles. This should agree with fds 0-2 in the parent process (in most cases) because the CRT keeps them in sync with the standard handles as long as a process doesn't directly call WINAPI `SetStdHandle`.

vstinner:
> This change should have a unit test, though it is difficult to create one that doesn't cause debug assertion failures in the MSVC runtime library when run.

IMO the existing test_os.test_pipe() is enough:

```
    @unittest.skipUnless(hasattr(os, 'pipe'), "need os.pipe()")
    def test_pipe(self):
        rfd, wfd = os.pipe()
        self.addCleanup(os.close, rfd)
        self.addCleanup(os.close, wfd)
        self.assertEqual(os.get_inheritable(rfd), False)
        self.assertEqual(os.get_inheritable(wfd), False)
```

Misc/NEWS.d/next/Core and Builtins/2019-06-01-23-55-50.bpo-32865.Ha11I1.rst
```diff
@@ -0,0 +1,2 @@
+On Windows, do not use inheritable file descriptors to wrap handles in
+:func:`os.pipe`.
```
vstinner, line 2: Would you mind to also document _io._WindowsConsoleIO.fileno() fix?

vstinner (approved):
LGTM, but I added a minor request. Can you try to rebase your PR?

IMO we should always pass _O_NOINHERIT flag in msvcrt.open_osfhandle(). For example, we do the same in select.epoll(). It can be addressed in a separated issue, or the same PR. It's up to you.

Misc/NEWS.d/next/Core and Builtins/2019-06-01-23-55-50.bpo-32865.Ha11I1.rst
ZackerySpytz: Thank you, Victor, for looking at this PR -- I will update it soon.

iritkatriel (changes requested):
This has merge conflicts now.

vstinner:
Maybe a better approach would replace _open_osfhandle() with _Py_open_osfhandle_noraise(), and modify _Py_open_osfhandle_noraise() so it always sets the _O_NOINHERIT flag.

erlend-aasland:
ISTM this was fixed by #1927.

vstinner:
> ISTM this was fixed by https://github.com/python/cpython/pull/1927.

@erlend-aasland: PR #1927 added `_Py_open_osfhandle()` helper function to wrap `_open_osfhandle()` call between `_Py_BEGIN_SUPPRESS_IPH` and `_Py_END_SUPPRESS_IPH`, but it doesn't always set the `_O_NOINHERIT` which is the purpose of this PR.

PR #13743, opened 2019-06-02
bpo-37134: Add PEP570 notation to the documentation

pablogsal:
https://bugs.python.org/issue37134


vstinner (approved):
LGTM.

vstinner:
Only zlib uses positional only for the first paramater? Maybe other compression libs should do the same?

willingc (approved):
Looks good to me. Thanks @pablogsal

PR #13745, opened 2019-06-02
Fix typos in docs and docstrings

tirkarthi:
I have limited typos to user facing output and avoided typos in source code comments to reduce diff.

perfoming -> performing
specifiy -> specify
accross -> across
filname -> filename
prepand -> prepend
reminaing -> remaining
atttributes -> attributes
inherints -> inherits
brower -> browser
Informaton -> Information
hostame -> hostname
imput -> input
toogles betweem -> toggles between
Coersion -> Coercion

PR #13749, opened 2019-06-02
Fix typos in docs and comments in idlelib

tirkarthi:
extenstons -> extensions
Attrbutes -> Attributes
delagator -> delegator
diplayable -> displayable
corresonding -> corresponding
examime -> examine
contination -> continuation
seach -> search
create_widgers -> create_widgets
Double-cliking -> Double-clicking
tkintter -> tkinter

### DIFF
--- a/Lib/idlelib/config.py
+++ b/Lib/idlelib/config.py
@@ -12,7 +12,7 @@ database keys are tuples (config-type, section, item).  As implemented,
 there are  separate dicts for default and user values.  Each has
 config-type keys 'main', 'extensions', 'highlight', and 'keys'.  The
 value for each key is a ConfigParser instance that maps section and item
-to values.  For 'main' and 'extenstons', user values override
+to values.  For 'main' and 'extensions', user values override
 default values.  For 'highlight' and 'keys', user sections augment the
 default sections (and must, therefore, have distinct names).
 
--- a/Lib/idlelib/configdialog.py
+++ b/Lib/idlelib/configdialog.py
@@ -199,7 +199,7 @@ class ConfigDialog(Toplevel):
     def help(self):
         """Create textview for config dialog help.
 
-        Attrbutes accessed:
+        Attributes accessed:
             note
 
         Methods:
--- a/Lib/idlelib/delegator.py
+++ b/Lib/idlelib/delegator.py
@@ -14,7 +14,7 @@ class Delegator:
 
     def resetcache(self):
         "Removes added attributes while leaving original attributes."
-        # Function is really about resetting delagator dict
+        # Function is really about resetting delegator dict
         # to original state.  Cache is just a means
         for key in self.__cache:
             try:
--- a/Lib/idlelib/editor.py
+++ b/Lib/idlelib/editor.py
@@ -315,7 +315,7 @@ class EditorWindow(object):
                   self.CodeContext(self).toggle_code_context_event)
 
     def _filename_to_unicode(self, filename):
-        """Return filename as BMP unicode so diplayable in Tk."""
+        """Return filename as BMP unicode so displayable in Tk."""
         # Decode bytes to unicode.
         if isinstance(filename, bytes):
             try:
--- a/Lib/idlelib/idle_test/htest.py
+++ b/Lib/idlelib/idle_test/htest.py
@@ -12,7 +12,7 @@ or function) whose .__name__ attribute is also X (the usual situation).
 The first parameter of X must be 'parent'.  When called, the parent
 argument will be the root window.  X must create a child Toplevel
 window (or subclass thereof).  The Toplevel may be a test widget or
-dialog, in which case the callable is the corresonding class.  Or the
+dialog, in which case the callable is the corresponding class.  Or the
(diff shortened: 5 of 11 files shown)


tirkarthi:
This is similar to #13745 but the raised the changes to idlelib folder as a separate PR to assist backporting workflow for idlelib.

asvetlov (dismissed):
Please fix red builds

tirkarthi:
I am not sure why code comments changes cause test failure and especially it happens in test_asyncio and not idlelib related test failure. I have rebased and retriggered the run.

terryjreedy (approved):
How did I (and previous spell checks) miss some of these?  Thanks for finding these and doing them separately.

tirkarthi:
Since new docs are added everyday I just try to run the spell check every 2-3 months. I use aspell and it reports lot of false positives like variable names, function names, tests etc. For Docs folder it reports 10k words and Lib folder 30k words as typos. Hence manual skimming is required and sometimes I overlook the typo to find it next time.

terryjreedy:
Are you reducing .py files to strings and comments first?

tirkarthi:
aspell takes the input and splits it into words based on word boundary I think to get all words from the source. Below is the command I use

`aspell list < Lib/idlelib/**/*py | tr '[:upper:]' '[:lower:]' | sort -u | less`

terryjreedy:
The following, using the regexes in idlelib.colorizer, strips a file down to comments and strings.  If you could feed the output for all (idlelib or other stdlib) py files to aspell, there would be less to wade through, as the result have only a small fraction of the variable names.  (Since I am on Windows and don't know the aspell API, I cannot be more specific.)
```
import re

comment = r"#[^\n]*"
stringprefix = r"(?i:r|u|f|fr|rf|b|br|rb)?"
sqstring = stringprefix + r"'[^'\\\n]*(\\.[^'\\\n]*)*'?"
dqstring = stringprefix + r'"[^"\\\n]*(\\.[^"\\\n]*)*"?'
sq3string = stringprefix + r"'''[^'\\]*((\\.|'(?!''))[^'\\]*)*(''')?"
dq3string = stringprefix + r'"""[^"\\]*((\\.|"(?!""))[^"\\]*)*(""")?'
comstring = "|".join([comment, sq3string, dq3string, sqstring, dqstring])

pattern = re.compile(comstring, re.S)
with open("c:/Programs/Python38/Lib/idlelib/autocomplete.py") as f:
    for m in pattern.finditer(f.read()):
        print(m.group(0))
```

18z:
Hi @tirkarthi 

Just came up with the idea.
Is it possible that we can create a typo checker (bot) ?

Like the-knights-who-say-ni or  bedevere-bot, 
when receive a PR, the bot automatically checks the typo.

18z:
More details.

Every time when the typo checker finds a typo, it warns the PR submitter.
PR submitter can fix the typo based on the report.

It is okay, if the typos were found and submitter doesn't fix it.
Since it is probably intended to spell the word that way.

terryjreedy:
I had that in the back of my mind when I posted the code above.  However, the problem is that even after reducing the code checked, there will still be mostly false positives, and for many modules, there will always be false positives.  @tirkarthi could get some numbers after interfacing the filter with aspell.  And of course, the bot would have to have aspell available.  The report could be made optional with a spellcheck label.

krnick:
> More details.
> 
> Every time when the typo checker finds a typo, it warns the PR submitter.
> PR submitter can fix the typo based on the report.
> 
> It is okay, if the typos were found and submitter doesn't fix it.
> Since it is probably intended to spell the word that way.

I also think it would be great to add this feature to the CI test. For example,  trailing whitespace testing.

tirkarthi:
I share the same concern about high number of false positives. I mostly took the code from the rust-lang book where there is a separate dictionary.txt file with list of false positives and this [shell script](https://github.com/rust-lang/book/blob/master/ci/spellcheck.sh) to check spelling. It can work at the scale of book but not sure about Python docs, code level docstrings and comments that can have lot of variables, function names and so on that there is a higher rate of changes to keep updating dictionary.txt.

18z:
Hi @terryjreedy  and @tirkarthi 

I summarized the discussion above.

**Consensus:**
- Reducing typo in a constructive way.

**To be Solved:**
- Even after reducing the code checked, there will still be mostly false positives.
- High frequency of updating dictionary.txt (false positive typo).

**Next Step:**
- Interfacing the filter with aspell to get some numbers.

**MISC:**
- The report could be made optional with a spellcheck label.
- Code for reducing .py files to strings and comments.
- Command line for spellcheck. 

Is @tirkarthi  currently working on interfacing the filter? 
I can help otherwise :D

tirkarthi:
No, I am not working on the filters data or improving aspell side and manual check for every 2-3 months seems good. Feel free to continue with the approach and I may not have time to continue with this.

Thanks

terryjreedy:
Further general discussion should be on the core-workflow list.  A specific proposal for git should be an issue for the core-workflow repository, which I believe is where bots originate.

18z:
Thanks @terryjreedy , I'll move the discussion to core-workflow repository. :D

PR #13754, opened 2019-06-02
bpo-35621: Fix tests when SafeChildWatcher is expected instead of ThreadedChildWatcher

asvetlov:

### DIFF
--- a/Lib/test/test_asyncio/test_unix_events.py
+++ b/Lib/test/test_asyncio/test_unix_events.py
@@ -1825,6 +1825,7 @@ class PolicyTests(unittest.TestCase):
 
         def f():
             policy.set_event_loop(policy.new_event_loop())
+
             self.assertIsInstance(policy.get_event_loop(),
                                   asyncio.AbstractEventLoop)
             watcher = policy.get_child_watcher()
--- a/Lib/test/test_asyncio/test_unix_events.py
+++ b/Lib/test/test_asyncio/test_unix_events.py
@@ -1824,6 +1824,7 @@ class PolicyTests(unittest.TestCase):
     def test_get_child_watcher_thread(self):
 
         def f():
+            policy.set_event_loop(policy.new_event_loop())
             self.assertIsInstance(policy.get_event_loop(),
                                   asyncio.AbstractEventLoop)
             watcher = policy.get_child_watcher()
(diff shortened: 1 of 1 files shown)


pablogsal:
This is failing for me because there is no event loop in the thread that executes `f`:

```exception
❯ ./python -m test test_asyncio -m test_unix_events
Run tests sequentially
0:00:00 load avg: 0.65 [1/1] test_asyncio
Exception in thread Thread-1:
  File "/home/pablogsal/github/cpython/Lib/threading.py", line 923, in _bootstrap_inner
    self.run()
  File "/home/pablogsal/github/cpython/Lib/threading.py", line 865, in run
    self._target(*self._args, **self._kwargs)
  File "/home/pablogsal/github/cpython/Lib/test/test_asyncio/test_unix_events.py", line 1828, in f
    self.assertIsInstance(policy.get_event_loop(),
  File "/home/pablogsal/github/cpython/Lib/asyncio/events.py", line 639, in get_event_loop
    raise RuntimeError('There is no current event loop in thread %r.'
RuntimeError: There is no current event loop in thread 'Thread-1'.
/home/pablogsal/github/cpython/Lib/asyncio/base_events.py:646: ResourceWarning: unclosed event loop <_UnixSelectorEventLoop running=False closed=False debug=False>
  _warn(f"unclosed event loop {self!r}", ResourceWarning, source=self)
```

Lib/test/test_asyncio/test_unix_events.py
```diff
@@ -1824,8 +1824,6 @@ def test_get_child_watcher_after_set(self):
     def test_get_child_watcher_thread(self):
 
         def f():
-            policy.set_event_loop(policy.new_event_loop())
```
pablogsal: What about:

```
def f():
    policy.set_event_loop(policy.new_event_loop())
    policy.set_child_watcher(asyncio.SafeChildWatcher())
```

asvetlov: Changed to keep `policy.set_event_loop()` in a thread but setup the watcher in the main thread (it is crucial for the test).

pablogsal (approved):
LGTM

I tested it locally and get no failures anymore :)

asvetlov:
Thanks @pablogsal

PR #13757, opened 2019-06-02
bpo-35814: Update the annotated assignment docs

ilevkivskyi:
cc @gvanrossum @rhettinger 

Btw, when writing this I am now thinking is there a reason why we don't allow also star expressions in the r.h.s. of annotated assignments (so that it mirrors normal assignment, not augmented assignment)? For example `t: Tuple[int, ...] = x, y, *z` is a `SyntaxError`. Should we try allowing this? Or is it too late in the release cycle?

https://bugs.python.org/issue35814

### DIFF
--- a/Doc/reference/simple_stmts.rst
+++ b/Doc/reference/simple_stmts.rst
@@ -329,7 +329,8 @@ Annotated assignment statements
 statement, of a variable or attribute annotation and an optional assignment statement:
 
 .. productionlist::
-   annotated_assignment_stmt: `augtarget` ":" `expression` ["=" `expression`]
+   annotated_assignment_stmt: `augtarget` ":" `expression`
+                            : ["=" (`expression_list` | `yield_expression`)]
 
 The difference from normal :ref:`assignment` is that only single target and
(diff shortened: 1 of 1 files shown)


ilevkivskyi:
See https://github.com/python/peps/pull/1096 for the corresponding PEP update.

gvanrossum (approved):
Re: whether to allow `x, y, *z` on the RHS, I guess PEP 526 again doesn't specify this either way. I vaguely recall that during the discussion we weren't too sure about various extensions, but perhaps the most important thing we didn't want to allow was multiple variables (e.g. `x, y: int` would be ambiguous).

I'm okay with allowing that, and it's probably better to do it sooner than later. The fix presumably is very simple (just update the grammar) but I'd request a test to make sure we didn't miss anything obvious in the code generation. (But `x: Tuple[int, ...] = (1, 2, *xs)` does work, so probably nothing.)

ilevkivskyi:
OK, I will make a separate PR tomorrow (it is almost 1AM here).

PR #13758, opened 2019-06-02
bpo-36027 bpo-36974: Fix "incompatible pointer type" compiler warnings

encukou:

### DIFF
--- a/Objects/longobject.c
+++ b/Objects/longobject.c
@@ -4243,7 +4243,7 @@ long_invmod(PyLongObject *a, PyLongObject *n)
 
     Py_DECREF(c);
     Py_DECREF(n);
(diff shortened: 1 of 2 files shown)


pablogsal (approved):

Objects/longobject.c
```diff
@@ -4243,7 +4243,7 @@ long_invmod(PyLongObject *a, PyLongObject *n)
 
     Py_DECREF(c);
     Py_DECREF(n);
-    if (long_compare(a, _PyLong_One)) {
+    if (long_compare(a, (PyObject *)_PyLong_One)) {
```
encukou, line 4246: This is wrong, it should be `PyLongObject`.
https://github.com/python/cpython/pull/13761 is the proper fix.

PR #13761, opened 2019-06-03
bpo-36027: Really fix "incompatible pointer type" compiler warning

encukou:
Apologies for the earlier hasty attempt.


encukou:
test_asyncio is flaky, it sometimes "alters the execution environment". It looks unrelated to any recent changes.
I don't know what I can do to fix it :(

PR #13762, opened 2019-06-03
bpo-37069: tests use catch_unraisable_exception()

vstinner:
Modify test_coroutines, test_cprofile, test_generators, test_raise,
test_ssl and test_yield_from to use
support.catch_unraisable_exception() rather than
support.captured_stderr().

test_thread: remove test_save_exception_state_on_error() which is now
updated. test_unraisable_exception() checks that sys.unraisablehook()
is called to handle _thread.start_new_thread() exception.

test_cprofile now rely on unittest for test discovery: replace
support.run_unittest() with unittest.main().

https://bugs.python.org/issue37069

### DIFF
--- a/Lib/test/test_coroutines.py
+++ b/Lib/test/test_coroutines.py
@@ -2032,11 +2032,17 @@ class CoroutineTest(unittest.TestCase):
     def test_fatal_coro_warning(self):
         # Issue 27811
         async def func(): pass
-        with warnings.catch_warnings(), support.captured_stderr() as stderr:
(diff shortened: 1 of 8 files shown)


vstinner:
Travis CI: "Documentation tests" failed with:

```
Warning, treated as error:
duplicate object description of email.message, other instance in library/email.message, use :noindex: for one of them
```

vstinner:
> Travis CI: "Documentation tests" failed with: (...)

https://bugs.python.org/issue37136

PR #13763, opened 2019-06-03
bpo-5680: IDLE: Customize running a module.

csabella:
The initialize options are 1) add command line options, which are appended to sys.argv as if passed on a real command line, and 2) skip the shell restart.  The customization dialog is accessed by a new item on the Run menu.

https://bugs.python.org/issue5680


Lib/idlelib/query.py
```diff
@@ -83,7 +84,7 @@ def __init__(self, parent, title, message, *, text0='', used_names={},
             self.deiconify()  # Unhide now that geometry set.
             self.wait_window()
 
-    def create_widgets(self):  # Call from override, if any.
+    def create_widgets(self, ok_text='OK'):  # Call from override, if any.
```
csabella, line 87: I made this configurable so that the 'Custom Run' dialog could change the `OK` to `Run`.

Lib/idlelib/query.py
```diff
@@ -302,10 +303,61 @@ def entry_ok(self):
...
+        frame = self.frame
+        self.restartvar = BooleanVar(self, value=True)
+        restart = Checkbutton(frame, variable=self.restartvar, onvalue=True,
+                              offvalue=False, text='Restart shell')
+        self.args_error = Label(frame, text=' ', foreground='red',
+                                font=self.error_font)
+
+        restart.grid(column=0, row=4, columnspan=3, padx=5, sticky='w')
+        self.args_error.grid(column=0, row=12, columnspan=3, padx=5,
+                             sticky='we')
```
csabella, line 333: I'm not using `self.args_error` yet, but it added nice spacing and I figured it would come in handy for an edit such as `must restart with arguments` as mentioned on the bug tracker.

Lib/idlelib/query.py
```diff
@@ -302,10 +303,61 @@ def entry_ok(self):
...
+
+    def cli_args_ok(self):
+        "Validity check and parsing for command line arguments."
+        cli_args = self.entry.get().strip()
+        if not cli_args:
+            self.showerror('no arguments specified.')
+            return None
```
csabella: Not sure about this edit, but it makes everything a little nicer right now.  The reason I'm not sure about it is due to the behavior of the current `Run Module`.  Should `Run Module` use the last settings from `Custom Run` or should it work exactly as it works now (while in a subprocess)?  `Run Module` without a subprocess uses the `Custom Run` settings from the last run.

Lib/idlelib/query.py
```diff
@@ -302,10 +303,61 @@ def entry_ok(self):
...
+        try:
+            lex = shlex.split(cli_args, posix=True)
+        except ValueError as err:
+            self.showerror(str(err))
+            return None
+        return lex
+
+    def restart_ok(self):
+        return self.restartvar.get()
```
csabella: Could add edits here for when `restart` cannot be False.

Lib/idlelib/runscript.py
```diff
@@ -131,22 +138,27 @@ def _run_module_event(self, event):
...
+        interp.runcommand(f"""if 1:
             __file__ = {filename!r}
             import sys as _sys
             from os.path import basename as _basename
+            argv = {argv!r}
             if (not _sys.argv or
-                _basename(_sys.argv[0]) != _basename(__file__)):
-                _sys.argv = [__file__]
+                _basename(_sys.argv[0]) != _basename(__file__) or
+                len(argv) > 1):
```
csabella, line 160: Checking the length so that, if there are additional args, they will replace what was already there.  However, existing args won't be replaced by no args.  Currently, F5 still runs with no args and restarts the shell, but that's something to keep in mind if F5 changes.

Lib/idlelib/query.py
terryjreedy: Nice alternative, at least for now, to factoring out the query code I want to reuse and starting with a separate class.  There is another issue about standardizing dialog box behavior that we should revisit after this one.

Lib/idlelib/query.py
terryjreedy: I will have to reread more of the query code to understand that.  We will have to comment out or remove the restart button until it works.

terryjreedy (changes requested):
Test of behavior without looking much at code yet.  Will respond to other comments later.  Immediate issue:  I have a custom keyset without new pseudoevent binding and on startup see

 Warning: config.py - IdleConf.GetCoreKeys -
 problem retrieving key binding for event '<<run-custom>>'
 from key set 'test'.
 returning default value: ['<Shift-Key-F5>']

The fix is simple and I know how to do it if you don't beat me. 

Separate issue if not done yet: write back keyset with new bindings.

Otherwise, works great.

taleinat:
1. I find the name "Run Custom" unclear and confusing, and I image some users would too. Is "Run Module With Arguments" too long? Perhaps "Run With Arguments..."?
2. It would be great to remember the last used arguments for each file, and pre-populate the field with them.

Lib/idlelib/query.py
csabella: Sounds good!  🙂

Lib/idlelib/query.py
csabella: Yes, `restart` works now, but it's never disabled.  I wasn't sure which edits would make sense around it.  On the ticket, you had mentioned that using CLI arguments would make sense on a restart, but right now it allows changing the CLI arguments without restarting the shell.  That's where my other comments (about this dialog not allowing blank arguments) and what the current `Run Module`'s behavior should be.

csabella:

@terryjreedy, I added this pseudoevent to the `former_extension_events` set.  I think that's what you wanted to fix -- to suppress the warning for the new binding and then to work on updating any custom keysets in another ticket.

csabella:
Thanks for looking at this, @taleinat!

> I find the name "Run Custom" unclear and confusing, and I image some users would too. Is "Run Module With Arguments" too long? Perhaps "Run With Arguments..."?

Terry wants to have more things on that dialog to be able to change the run environment, so using 'arguments' on the menu isn't inclusive enough.  If 'Run Custom' is awkward, maybe 'Customize Run'?

>    It would be great to remember the last used arguments for each file, and pre-populate the field with them.

Yes, there was a few things that I listed on the bpo issue that I didn't do in this first round, including saving the values. Terry requested creating a minimal version to start with.  Saving the items wouldn't (shouldn't?) complicate the code, but I wanted to try to keep it simple and then iterate over it in the next round.  He had also mentioned printing the run options in the shell window so that multiple runs would be annotated.  Both features would be helpful.

taleinat:
> > I find the name "Run Custom" unclear and confusing, and I image some users would too. Is "Run Module With Arguments" too long? Perhaps "Run With Arguments..."?
> 
> Terry wants to have more things on that dialog to be able to change the run environment, so using 'arguments' on the menu isn't inclusive enough. If 'Run Custom' is awkward, maybe 'Customize Run'?

I see. Perhaps just "Run..."?

Lib/idlelib/idle_test/test_query.py
```diff
@@ -242,6 +242,57 @@ def test_entry_ok_helpsource(self):
...
+        def __init__(self, dummy_entry):
+            self.entry.set(dummy_entry)
+            self.entry_error['text'] = ''
+        def showerror(self, message):
+            self.entry_error['text'] = message
+
+    def test_blank_args(self):
+        dialog = self.Dummy_CustomRun(' ')
+        self.assertEqual(dialog.cli_args_ok(), None)
+        self.assertIn('no arguments', dialog.entry_error['text'])
```
taleinat: This should use `assertRegex` to do a case-insensitive check.
```suggestion
        self.assertRegex(r'(?i)no arguments', dialog.entry_error['text'])
```

Lib/idlelib/idle_test/test_query.py
```diff
@@ -242,6 +242,57 @@ def test_entry_ok_helpsource(self):
...
+
+    def test_invalid_args(self):
+        dialog = self.Dummy_CustomRun("'no-closing-quote")
+        self.assertEqual(dialog.cli_args_ok(), None)
+        self.assertIn('No closing', dialog.entry_error['text'])
```
taleinat, line 256: This should use assertRegex to do a case-insensitive check.
```suggestion
        self.assertRegex(r'(?i)no closing', dialog.entry_error['text'])
```

Lib/idlelib/idle_test/test_query.py
```diff
@@ -242,6 +242,57 @@ def test_entry_ok_helpsource(self):
                 self.assertEqual(dialog.entry_ok(), result)
 
 
+class CustomRunCLIargsokTest(unittest.TestCase):
+    "Test cli_ok method of the CustomRun subclass of Query."
+
+    class Dummy_CustomRun:
+        cli_args_ok = query.CustomRun.cli_args_ok  # Function being tested.
+        entry = Var()
```
taleinat: Why is this a class attribute? Besides being unclear, this also doesn't reflect the actual `CustomRun` class.

Lib/idlelib/idle_test/test_query.py
```diff
@@ -242,6 +242,57 @@ def test_entry_ok_helpsource(self):
...
+        entry_error = {}
```
taleinat: It might be simpler to make `Dummy_CustomRun` a `Mock` instance and then check calls to `showerror()`, rather than the `entry_error['text']` stuff which is more implementation-specific.

Lib/idlelib/idle_test/test_query.py
```diff
@@ -242,6 +242,57 @@ def test_entry_ok_helpsource(self):
...
+            return self.cli_args
+        def restart_ok(self):
+            return self.restart
+
+    def test_entry_ok_customrun(self):
+        dialog = self.Dummy_CustomRun()
+        for dialog.restart in {True, False}:
+            for cli_args, result in ((None, None),
+                                     ('my arg', ('my arg', dialog.restart))):
+                with self.subTest():
```
taleinat:
```suggestion
                with self.subTest(restart=dialog.restart, cli_args=cli_args):
```

Lib/idlelib/query.py
taleinat: IMO "Run Module" should always ignore the latest "Custom Run" settings, regardless of whether or not a subprocess is used.

Lib/idlelib/runscript.py
```diff
@@ -113,15 +114,21 @@ def run_module_event(self, event):
...
 
-    def _run_module_event(self, event):
+    def run_custom_event(self, event):
+        run_args = CustomRun(event.widget, "Customize Run").result
+        # User cancelled.
+        if not run_args:
+            return 'break'
+        return self._run_module_event(event, cli_args=run_args[0], restart=run_args[1])
+
+    def _run_module_event(self, event, *, cli_args=None, restart=True):
```
taleinat: The "Run Module" and "Run Custom" should each have their own (simple) event handler, with the common code extracted into a common function used by both, which isn't an event handler.

Misc/NEWS.d/next/IDLE/2019-06-03-00-39-29.bpo-5680.VCQfOO.rst
```diff
@@ -0,0 +1 @@
+Run modules from the editor using command line arguments.
```
taleinat: This needs to be a bit more explicit, mentioning that a new menu option is added.

Lib/idlelib/idle_test/test_query.py
```diff
@@ -348,5 +399,23 @@ def test_click_help_source(self):
...
+    def setUpClass(cls):
+        requires('gui')
+
+    def test_click_args(self):
+        root = Tk()
+        root.withdraw()
+        dialog =  query.CustomRun(root, 'Title', _utest=True)
+        dialog.entry.insert(0, 'okay')
+        dialog.button_ok.invoke()
+        self.assertEqual(dialog.result, (['okay'], True))
```
taleinat, line 398: The should probably be a call to `root.update()` after the `.invoke()` call, to ensure that Tk has a chance to do all of its processing.

Lib/idlelib/idle_test/test_query.py
```diff
@@ -348,5 +399,23 @@ def test_click_help_source(self):
...
+        del dialog
+        root.destroy()
```
taleinat, line 399: I suggest cleaning up using `self.addCleanup()` immediately after creating the `Tk()` instance, e.g.:

```python
    root = Tk()
    def cleanup():
        root.update()
        root.destroy()
    self.addCleanup(cleanup)

File: Lib/idlelib/idle_test/test_query.py [hunk 15]
```diff
@@ -348,5 +399,23 @@ def test_click_help_source(self):
...
+        requires('gui')
+
+    def test_click_args(self):
+        root = Tk()
+        root.withdraw()
+        dialog =  query.CustomRun(root, 'Title', _utest=True)
+        dialog.entry.insert(0, 'okay')
+        dialog.button_ok.invoke()
+        self.assertEqual(dialog.result, (['okay'], True))
+        del dialog
```
taleinat: Are this `del` call, and the one below, actually needed?

Lib/idlelib/query.py
```diff
@@ -302,10 +303,61 @@ def entry_ok(self):
...
+                             sticky='we')
+
+    def cli_args_ok(self):
+        "Validity check and parsing for command line arguments."
+        cli_args = self.entry.get().strip()
+        if not cli_args:
+            self.showerror('no arguments specified.')
+            return None
+        try:
+            lex = shlex.split(cli_args, posix=True)
```
taleinat: What does `lex` stand for? For me, this is confusing both here and in `entry_ok()`. Perhaps `args_list`?

taleinat (commented):
As a general note, I'm not sure about the usefulness of the non-GUI tests for the RunCustom dialog (`CustomRunCLIargsokTest`, `CustomRunEntryokTest`) . They rely rather heavily on the internal implementation. Since we have GUI tests run by the CI now, I prefer to have simpler tests actually using Tk with less mocking. Would it be possible to simply move the tests from these classes into `CustomRunGuiTest`?

terryjreedy:
I am reviewing and editing with the intention of getting this into 3.7.4.  While the user UI can be changed after that, it is my primary focus.  Summary responses to comments above.
* Config warning: fixed.
* Menu entry: We could flip 'Run custom', but I prefer to keep to a verb form.  ('Python Shell' might better be 'Open Shell'.) I think 'Run customized' or 'Run ... customized' as an abbreviation for 'Run module with customized settings' is better, and should be good enough for now.  I will see how latter looks.
* Restart (or not) button: now functional.  Add to blurb.
* Arguments, without restart: These extend sys.argv.  'Command line' arguments do not make sense without restart, but extending sys.argv does.  So this is OK.
* No arguments, with restart: I changed my mind and will remove the current error check.  No args will be ok with other settings, and if someone opens box to see available customizations, I now think they should be able to run without changing anything.
(Note: check button will be replaced with radio buttons when another mutually exclusive option is added, such as 'O Run module in system console' (no issue yet).) 

* Code comments, especially tests: I will make the changes I currently agree with.
* Blurb: will expand.

Lib/idlelib/runscript.py
```diff
@@ -113,15 +114,21 @@ def run_module_event(self, event):
         else:
             return self._run_module_event(event)
 
-    def _run_module_event(self, event):
+    def run_custom_event(self, event):
+        run_args = CustomRun(event.widget, "Customize Run").result
+        # User cancelled.
+        if not run_args:
```
terryjreedy: I am removing this clause.

Lib/idlelib/runscript.py
terryjreedy: _run_module_event, with initial '_', is already the common code for run_module_event and run_custom_event.  It is only an 'event handler' to implement the Mac-specific delayed call via an internal pseudo pseudoevent.  I am assuming that someone tried calling it directly after the delay.  As a separate issue, we might test whether the workaround is needed in 8.6.

Misc/NEWS.d/next/IDLE/2019-06-03-00-39-29.bpo-5680.VCQfOO.rst
terryjreedy: Expanded.

Lib/idlelib/runscript.py
terryjreedy: I believe that this can only matter if one used the deprecated -n option.  I have no intention of changing F5

Lib/idlelib/query.py
terryjreedy: I believe I covered this elsewhere.

Lib/idlelib/query.py
terryjreedy: I believe Run Module is currently unchanged and agree that this should continue.

Lib/idlelib/runscript.py
```diff
@@ -113,15 +114,21 @@ def run_module_event(self, event):
...
+            return 'break'
+        return self._run_module_event(event, cli_args=run_args[0], restart=run_args[1])
```
terryjreedy: Changed to pass run_args and unpack in _run_module.

Lib/idlelib/query.py
terryjreedy: Agreed. I switched to cli_string and cli_args.  cli_args_ok should return cli_args.

Lib/idlelib/query.py
terryjreedy: No longer possible.

Lib/idlelib/idle_test/test_query.py
terryjreedy: I wrote the query tests, and Cheryl followed the pattern I set.  For testing with one instance at a time, or ever, class and instance attributes work the same.  For attributes never rebound, I attached them to the class, especially if there was no other need for an __init__ method.  Very clear to me.  However, I moved Var attributes to __init__ when there is one as it saves a line.

The cli_args_ok method being tested accesses self.query.  tk Vars and Entries both have a get method that does the same thing, so a mock Var replaces an Entry quite nicely.

Lib/idlelib/idle_test/test_query.py
terryjreedy: I used this in all the other functional method tests.  The dummy classes have and do exactly what is needed to test the method.  A unittest.mock would be much more complicated and difficult and would run much slower.  Test speed is important.

Lib/idlelib/idle_test/test_query.py
terryjreedy: This assert is gone.

Lib/idlelib/idle_test/test_query.py
terryjreedy: The test is clear and it works.

Lib/idlelib/idle_test/test_query.py
terryjreedy: Done.

Lib/idlelib/idle_test/test_query.py
terryjreedy: I believe invoke blocks until done.  I have run this test at least 20 times without issue.  event_generate is different (and flakey).  There are 55 invokes in idle tests and my spot check of nearly 10 showed none followed by update.  And I know of no resulting failures.

Lib/idlelib/idle_test/test_query.py
terryjreedy: No.  Gone.  Added class attributes need deletion, but function locals already go on return, and there is no evident issue here with destroying first.

Lib/idlelib/idle_test/test_query.py
terryjreedy: I far prefer explicit calls.  Update is not generally needed.  update_idletasks is more often needed.  I have not seen any of the tclerror messages that so indicate.

terryjreedy (approved):

terryjreedy:
buglet to fix: focus is not being shifted to shell correctly.

terryjreedy:
Changes: 1) Check code before bother to customize. 2) Add module title to title bar.  3) Make shell.text the parent, so focus goes to shell.  Good enough to merge.  See issue for 2 remaining issues.

csabella:
@taleinat, thank you for the review and @terryjreedy, thank you for the review and making the additional changes.

PR #11636, opened 2019-01-21
gh-79888: support __index__ and __float__ in time functions

jdemeyer:
https://bugs.python.org/issue35707

Note that there is substantial code duplication in my patch, but that is only because the existing code already had substantial duplication. It would be possible to fix that, at the expense of a slightly more complicated (and therefore probably slightly slower) implementation.

* Issue: gh-79888

PR #13768, opened 2019-06-03
bpo-37086: fixed time.sleep error message

Windsooon:
https://bugs.python.org/issue37086

### DIFF
--- a/Lib/test/test_time.py
+++ b/Lib/test/test_time.py
@@ -165,7 +165,7 @@ class TimeTestCase(unittest.TestCase):
 
     def test_sleep_invalid(self):
         with self.assertRaisesRegex(
-                TypeError, r'an integer or float is required \(got type str\)'):
+                TypeError, 'an integer or float is required'):
             time.sleep('foo')
 
     def test_strftime(self):
--- a/Lib/test/test_time.py
+++ b/Lib/test/test_time.py
@@ -165,9 +165,8 @@ class TimeTestCase(unittest.TestCase):
 
     def test_sleep_invalid(self):
         with self.assertRaisesRegex(
-                TypeError, 'an integer or a float is required \(got type str\)') as te:
+                TypeError, r'an integer or a float is required \(got type str\)'):
             time.sleep('foo')
-
 
     def test_strftime(self):
         tt = time.gmtime(self.t)
--- a/Lib/test/test_time.py
+++ b/Lib/test/test_time.py
@@ -165,7 +165,7 @@ class TimeTestCase(unittest.TestCase):
 
     def test_sleep_invalid(self):
         with self.assertRaisesRegex(
-                TypeError, r'an integer or a float is required \(got type str\)'):
+                TypeError, r'an integer or float is required \(got type str\)'):
             time.sleep('foo')
 
     def test_strftime(self):
--- a/Lib/test/test_time.py
+++ b/Lib/test/test_time.py
@@ -163,6 +163,12 @@ class TimeTestCase(unittest.TestCase):
         self.assertRaises(ValueError, time.sleep, -1)
         time.sleep(1.2)
 
+    def test_sleep_invalid(self):
+        with self.assertRaisesRegex(
+                TypeError, 'an integer or a float is required \(got type str\)') as te:
+            time.sleep('foo')
+
+
     def test_strftime(self):
         tt = time.gmtime(self.t)
         for directive in ('a', 'A', 'b', 'B', 'c', 'd', 'H', 'I',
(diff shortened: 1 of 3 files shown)


Windsooon:
I'm not sure should I add news or not.

Lib/test/test_time.py
```diff
@@ -163,6 +163,12 @@ def test_sleep(self):
         self.assertRaises(ValueError, time.sleep, -1)
         time.sleep(1.2)
 
+    def test_sleep_invalid(self):
+        with self.assertRaisesRegex(
+                TypeError, 'an integer or a float is required \(got type str\)') as te:
+            time.sleep('foo')
+
+
```
ericvsmith: It looks like all of the other tests have just one blank line between them. Please delete one of the lines.

Lib/test/test_time.py
```diff
@@ -163,6 +163,12 @@ def test_sleep(self):
         self.assertRaises(ValueError, time.sleep, -1)
         time.sleep(1.2)
 
+    def test_sleep_invalid(self):
+        with self.assertRaisesRegex(
+                TypeError, 'an integer or a float is required \(got type str\)') as te:
```
ericvsmith: You don't need the `as te` part of this expression.

ericvsmith: You want this to be a raw (`r''`) string, in order to escape the parens for the regex engine.

ericvsmith (changes requested):
A few nits, but otherwise looks good.

I don't think you need a news entry.

Windsooon:
Thank you for review my code. I updated the PR based on your review.

Objects/longobject.c
```diff
@@ -207,7 +207,7 @@ _PyLong_FromNbIndexOrNbInt(PyObject *integral)
     nb = Py_TYPE(integral)->tp_as_number;
     if (nb == NULL || (nb->nb_index == NULL && nb->nb_int == NULL)) {
         PyErr_Format(PyExc_TypeError,
-                     "an integer is required (got type %.200s)",
+                     "an integer or a float is required (got type %.200s)",
```
ericvsmith: I know this is trivial, but it seems more consistent with existing code to say `"an integer or float is required"` (that is, drop the `"a"`). Would you mind making that change, too? I apologize for not catching that earlier.

ericvsmith (changes requested):
Sorry for the one last nit. Once you make this one and the tests pass, I'll commit it.

Lib/test/test_time.py
Windsooon: Yes. I think you are right. It came to me earlier but I forgot to delete it. I updated the PR, thanks.

ericvsmith:
It looks like test_datetime and test_socket are failing because they check the exception text. They'll need to be fixed, too.

Windsooon:
@ericvsmith I have some questions
1. Should we use "an integer or float is required." or "an integer or float is required. (got type %s), ob->ob_name", I found other tests didn't use a dynamic string in the PyErr_SetString function. 
2. Should I change `Invalid value NaN (not a number)` to `Not a float number` in (https://github.com/python/cpython/blob/master/Python/pytime.c#L408).

ericvsmith:
@Windsooon: 
1. If you're talking about the tests, I'd just look for `an integer or float is required`. But for the error message, I'd include the found type. It's very helpful in debugging.
2. I wouldn't change that as part of this PR. If it needs doing, I'd create a separate issue and PR.

jdemeyer (changes requested):
Won't this remove support for objects with `__int__`? See also bpo-35707.

jdemeyer:
Indeed, this breaks
```python
>>> import time, decimal; time.sleep(decimal.Decimal(1))
TypeError: an integer or float is required.
```

Windsooon:
Thank you :D, I will update the pr as well as the tests in a few days.

ericvsmith:
I think the earlier version of the patch (which didn't move the tests around) was a better design.

jdemeyer:
I think it's better to merge #11636 first since that will change the structure of this code anyway and it has all the tests that you need.

Python/pytime.c
```diff
@@ -401,7 +401,11 @@ static int
 _PyTime_FromObject(_PyTime_t *t, PyObject *obj, _PyTime_round_t round,
                    long unit_to_ns)
 {
-    if (PyFloat_Check(obj)) {
+    if (!PyFloat_Check(obj) && !PyLong_Check(obj)) {
```
iritkatriel, line 404: I don't think this is correct. In the else case below it calls PyLong_AsLongLong, which can handle value which are not PyLongs as long as they can be converted to a PyLong via _PyNumber_Index

iritkatriel, line 404: I don't think this is correct. In the else case below it calls PyLong_AsLongLong, which can handle values which are not PyLongs as long as they can be converted to a PyLong via _PyNumber_Index. 

If this change didn't break any unit tests, then a test is probably missing.

iritkatriel: ah, I see you've commented on that. Sorry for the noise.

StanFromIreland:
Completed in https://github.com/python/cpython/commit/155c44b9015089eacc6e2ace449391c12bfb8b8d

PR #13772, opened 2019-06-03
bpo-34788: Add support for scoped IPv6 addresses

opavlyuk:
https://bugs.python.org/issue34788


aeros (commented):
@opavlyuk Welcome and thanks for the contribution!

I would advise breaking up this PR into multiple smaller ones (approximately the size of the commits or smaller, but as self-contained chunks starting with the least dependent ones). This would allow for both easier review and debugging of the integration tests. Otherwise, this process will likely take significantly longer.

Also from the perspective of readability, I would recommend replacing the longer code blocks of repeating `assertBadSplit()` with iteration. 

For example on lines 359-368, instead of using this:

```python
        assertBadSplit("3ffe::1::1%scope")
        assertBadSplit("1::2::3::4:5%scope")
        assertBadSplit("2001::db:::1%scope")
        assertBadSplit("3ffe::1::%scope")
        assertBadSplit("::3ffe::1%scope")
        assertBadSplit(":3ffe::1::1%scope")
        assertBadSplit("3ffe::1::1:%scope")
        assertBadSplit(":3ffe::1::1:%scope")
        assertBadSplit(":::%scope")
        assertBadSplit('2001:db8:::1%scope')
```

I would recommend something like this:

```python
# since this is just an example, I didn't include all of them
bad_split_asserts = ["3ffe::1::1%scope", "1::2::3::4:5%scope", "2001::db:::1%scope",]
for assertion in bad_split_asserts:
    assertBadSplit(assertion)
```

This would significantly improve the overall readability of the code and more quickly conveys the message of what is being done. This doesn't have to be done when using a similar format for 2-3 lines, but it starts to add an unnecessary number of extra lines and characters when it's done 5+ times. 

Edit: Also I forgot to mention that we generally use underscores to represent spaces, not camelCase. So `assertBadSplit()` should be `assert_bad_split()`.

opavlyuk:
@aeros167  Thanks for your review!

> I would advise breaking up this PR into multiple smaller ones (approximately the size of the commits or smaller, but as self-contained chunks starting with the least dependent ones).

As for me, this PR could not be broken into multiple smaller ones, because it refers to and covers a single issue. Also, all changes depend on each other and could not be merged separately (e.g. tests after main code or vice versa). According to [devguide](https://devguide.python.org/pullrequest/#making-good-prs), tests and documentation should be included to PR.  So, seems it would not be right to make separate PR`s.
May be, some tests could be considered redundant, but [devguide](https://devguide.python.org/runtests/#writing) insists on writing tests both for normal behaviour and for error conditions. 


> Also from the perspective of readability, I would recommend replacing the longer code blocks of repeating `assertBadSplit()` with iteration.

Please, take a look at
https://github.com/python/cpython/blob/fb99b6b939c0a971a05228bbd951851660ed5779/Lib/test/test_ipaddress.py#L348-L357 In my PR I only extend this test with cases, consisting scope IPv6 addresses. It was already written in such a style. According to [devguide](https://devguide.python.org/pullrequest/#making-good-commits), it isn't recommended to make cosmetic changes to unrelated code.  [PEP8](https://www.python.org/dev/peps/pep-0008/#naming-conventions) also tells that, if an existing library has a different style, internal consistency is preferred.  So, to keep existing style of test, I haven't put assertions in loop. But, if it is a must, I will fix that lines, of course. 

> Edit: Also I forgot to mention that we generally use underscores to represent spaces, not camelCase. So `assertBadSplit()` should be `assert_bad_split()`.

Again, https://github.com/python/cpython/blob/fb99b6b939c0a971a05228bbd951851660ed5779/Lib/test/test_ipaddress.py#L343 was written before me, so I consider not to fix naming style. Also, as far as unitest library is a port of JUnit, it inherits Java naming style in assertion methods. Just take a look at all other assertion method names, they are camelCased. In my opinion, it would be right to keep existing code style in the module.

aeros167:
@opavlyuk 

> As for me, this PR could not be broken into multiple smaller ones, because it refers to and covers a single issue. Also, all changes depend on each other and could not be merged separately (e.g. tests after main code or vice versa). According to devguide, tests and documentation should be included to PR. So, seems it would not be right to make separate PR`s.
May be, some tests could be considered redundant, but devguide insists on writing tests both for normal behaviour and for error conditions.

Usually, for more involved issues, a single PR represents a self-contained step towards the resolution. Self-contained meaning that the PR also contains the respective documentation changes, test coverage, and is compatible with the existing code-base by itself. This doesn't mean that the entire issue has to be covered at once. 

As for this PR specifically, I'm not entirely certain into what self-contained pieces it could be further fragmented into, it's just a general guideline. If it's not reasonably possible to do this, it's not at all required. The most substantial benefit of fragmenting it into multiple PRs is easier debugging when the integration tests aren't passing.

> In my PR I only extend this test with cases, consisting scope IPv6 addresses. It was already written in such a style. According to devguide, it isn't recommended to make cosmetic changes to unrelated code. PEP8 also tells that, if an existing library has a different style, internal consistency is preferred. So, to keep existing style of test, I haven't put assertions in loop.

Upon reconsideration and further inspection of the existing tests, keeping the current style of the library is understandable, especially given the sheer volume. Perhaps I'll work on refactoring it in a future PR, but it'd likely have to be done all at once for consistency. When I had first submitted my initial review, I had underestimated the size of the existing tests. Improving the readability would likely be a project on it's own. I wouldn't expect you to refactor all of the entire existing tests within the same issue.

The above also applies to the naming conventions. I'm far more used to our more recent unit tests using the underscore naming. My apologies, it looks like I focused a bit too much on your additions rather than the context.

> But, if it is a must, I will fix that lines, of course.

As someone who isn't a core dev, my reviews are just recommendations based on my experiences thus far working within the Python repositories. So if you have reasons for doing anything a particular way and the core devs agree with your justification, you can definitely keep it as it is.

opavlyuk:
@aeros167, I've added entry to Misc/NEWS.d and resolved issues with suspicious lines in documentation, which caused checks to fail. Now, all checks are passing.
Looking forward to make all necessary fixes and have an approved PR :octocat:.

Misc/NEWS.d/next/Library/2019-07-17-08-26-14.bpo-34788.pwV1OK.rst
```diff
@@ -0,0 +1 @@
+Add support for scoped IPv6 addresses to `ipaddress <https://github.com/python/cpython/blob/master/Lib/ipaddress.py>`_. Patch by Oleksandr Pavliuk.
```
aeros:
```suggestion
Add support for scoped IPv6 addresses to :mod:`ipaddress`. Patch by Oleksandr 
Pavliuk.
```

Lib/ipaddress.py
```diff
@@ -1834,8 +1854,23 @@ def __init__(self, address):
         addr_str = str(address)
         if '/' in addr_str:
             raise AddressValueError("Unexpected '/' in %r" % address)
+        addr_str, self._scope_id = self._split_scope_id(addr_str)
+
         self._ip = self._ip_int_from_string(addr_str)
 
+    def __str__(self):
+        ip_str = super().__str__()
+        return ip_str if self.scope_id is None else ip_str + '%' + self.scope_id
```
aeros: This is simply reversing the conditional so that None doesn't have to be explicitly specified: 
``a if x is None else b`` == ``b if x else a``

```suggestion
        return ip_str + '%' + self.scope_id if self.scope_id else ip_str
```

This is another shorter alternative, but the previous one might be a bit easier to read:
```python
return ip_str + ('%' + self.scope_id if self.scope_id else '')
```

aeros (approved):
The documentation and changes to `ipaddress` look good as far as I can tell. Next, we just have give the core devs some time to review the PR. Based on the [experts index](https://devguide.python.org/experts/#stdlib) pmoody is the expert for `ipaddress`, but there is not a github account listed on [their issue tracker profile](https://bugs.python.org/user9293). As a result, I won't be able to mention them in the PR's comment. They're already on the issue's nosy list though.

As a minor fix, the news entry column width is supposed to be 80 characters according to the devguide, so I split it into two lines. Also, I trimmed the path to `ipaddress` down by replacing it with reST inline markup (mod is an abbreviation of module).

To simply a conditional, I made a minor suggestion for line 1863 of `ipaddress.py`.

Lib/ipaddress.py
opavlyuk: https://github.com/python/cpython/blob/033f0313013775ba841a8d965e32037469ea727d/Lib/ipaddress.py#L1799-L1803
Here I tried to exclude cases with blank `scope_id`, but your proposal is really reasonable if someone sets protected `self._scope_id` an empty value.

opavlyuk:
Sorry, I haven't erquested everybodies request, it happens automatically after I pushed revert commit. Trying to dismiss those requests.

Lib/ipaddress.py
```diff
@@ -1785,6 +1785,24 @@ def _reverse_pointer(self):
...
+            ip_str: A string, the IPv6 address.
+
+        Returns:
+            (addr, scope_id) tuple.
+        """
+        addr, sep, scope_id = ip_str.partition('%')
+        if not sep:
+            scope_id = None
+        elif not scope_id:
+            raise AddressValueError('Invalid IPv6 address: "%r"' % ip_str)
```
gnprice, line 1856: You should also give an error if `'%' in scope_id`. That, and a blank value, appear to be the only two forms forbidden in the RFC:

    An implementation MAY support other kinds of non-null strings as
    <zone_id>.  However, the strings must not conflict with the delimiter
    character.  The precise format and semantics of additional strings is
    implementation dependent.

Lib/ipaddress.py
```diff
@@ -1834,8 +1854,23 @@ def __init__(self, address):
...
+
         self._ip = self._ip_int_from_string(addr_str)
 
+    def __str__(self):
+        ip_str = super().__str__()
+        return ip_str + '%' + self._scope_id if self._scope_id else ip_str
+
+    @property
+    def scope_id(self):
+        """Identifier of a particular zone of the address's scope.
```
gnprice, line 1931: So from looking through [the RFC](https://tools.ietf.org/html/rfc4007), I think the standard terminology is: this identifier identifies a "zone", not a scope. (This description in the docstring looks exactly correct!) So a good name would be `zone_id`.

(In fact that's exactly the name the RFC uses when discussing the syntax.)

Lib/ipaddress.py
```diff
@@ -1834,8 +1854,23 @@ def __init__(self, address):
...
+
+        Returns:
+            A string, identifying the zone of the address if specified, else None.
+        """
```
gnprice, line 1938: Let's mention RFC 4007 in the docstring. This concept will be unfamiliar to many people (even people who know a bit about IPv6), so a reference will be helpful.

Lib/ipaddress.py
```diff
@@ -1834,8 +1854,23 @@ def __init__(self, address):
...
+    def __str__(self):
+        ip_str = super().__str__()
+        return ip_str + '%' + self._scope_id if self._scope_id else ip_str
+
+    @property
+    def scope_id(self):
+        """Identifier of a particular zone of the address's scope.
+
+        Returns:
+            A string, identifying the zone of the address if specified, else None.
```
gnprice: Small nit: delete the comma in "A string identifying ...". Otherwise it reads as though it's always a string.

Lib/ipaddress.py
```diff
@@ -1817,6 +1835,8 @@ def __init__(self, address):
             AddressValueError: If address isn't a valid IPv6 address.
 
         """
+        self._scope_id = None
```
gnprice: What's the reason for setting this up here? We just set it again below when we actually get the data, right?

... Ah I see, this sets it for the `int` and `bytes` cases below.

I think it'd be better to be explicit: write `self._scope_id = None` (well, actually `self._zone_id = None`) in each of those two cases. That makes it super clear that in those cases the zone ID is always None. And it only costs one extra line of code! :wink:

Lib/ipaddress.py
```diff
@@ -1989,7 +2024,7 @@ def hostmask(self):
         return self.network.hostmask
 
     def __str__(self):
-        return '%s/%d' % (self._string_from_ip_int(self._ip),
+        return '%s/%d' % (super().__str__(),
```
gnprice, line 2094: Is this a standard way of writing this combination? Do you know a reference for that?

Ah yes, it's section 11.7 of RFC 4007. Half tempted to put a comment mentioning that, but it's fine without.

gnprice (changes requested):
Thanks @opavlyuk for this contribution!

A few comments below. One is particularly important because it goes to the API: the ID here is really a "zone ID", not a "scope ID". (E.g. the scope might be "link-local", and in that case the zone is the particular link it's local to.)

opavlyuk:
@gnprice Thanks for your review!
> A few comments below. One is particularly important because it goes to the API: the ID here is really a "zone ID", not a "scope ID". (E.g. the scope might be "link-local", and in that case the zone is the particular link it's local to.)

I agree with you, [RFC 4007](https://tools.ietf.org/html/rfc4007#section-11.3) calls this part "zone_id".
At the same time, in existing Python code and documentation "scope id" is widely used too, e.g.:
[socket module](https://github.com/python/cpython/blob/master/Modules/socketmodule.c#L1879-L1881), [socket module](https://github.com/python/cpython/blob/master/Modules/socketmodule.c#L6649), [socket module tests](https://github.com/python/cpython/blob/master/Lib/test/test_socket.py#L1693-L1744), [socket module docs](https://docs.python.org/3/library/socket.html).
That examples include using of "scope_id" as part of AF_INET6 address family representation.
So, as for me, naming of attribute is the subject for discussion.

Lib/ipaddress.py
opavlyuk: [Fixed](https://github.com/python/cpython/pull/13772/commits/ac19fadfbcf73e314816c15c05f110cceede5b73)

Lib/ipaddress.py
opavlyuk: [Fixed](https://github.com/python/cpython/pull/13772/commits/b0a4fe901b2300b0be57619bf30d60e904e43fec)

Lib/ipaddress.py
opavlyuk: I agree with you, [RFC 4007](https://tools.ietf.org/html/rfc4007#section-11) calls this part "zone_id".
At the same time, in existing Python code and documentation "scope id" is widely used too, e.g.:
[socket module](https://github.com/python/cpython/blob/master/Modules/socketmodule.c#L1879-L1881), [socket module](https://github.com/python/cpython/blob/master/Modules/socketmodule.c#L6649), [socket module tests](https://github.com/python/cpython/blob/master/Lib/test/test_socket.py#L1693-L1744), [socket module docs](https://docs.python.org/3/library/socket.html).
That examples include using of "scope_id" as part of AF_INET6 address family representation.
So, as for me, naming of attribute is the subject for discussion.

Lib/ipaddress.py
opavlyuk: @gnprice Thanks for your review!

[Fixed](https://github.com/python/cpython/pull/13772/commits/25d3ac7bbabab78cacab6e0c9cd99fce5d2e4641) and [added tests](https://github.com/opavlyuk/cpython/commit/66db707bb02f58c0f1c79e4d891d1e714a3e0d7e).

Lib/ipaddress.py
gnprice: Oof, I see -- and in fact it's that way in the OS-level interfaces on [Linux](http://man7.org/linux/man-pages/man7/ipv6.7.html), [BSD](https://www.freebsd.org/cgi/man.cgi?query=inet6&sektion=4&apropos=0&manpath=FreeBSD+12.0-RELEASE), and Windows ([discussion](https://docs.microsoft.com/en-us/windows/win32/winsock/link-local-and-site-local-addresses-2), [struct](https://docs.microsoft.com/en-us/windows/win32/winsock/sockaddr-2)). Here's Linux:
```
           struct sockaddr_in6 {
               sa_family_t     sin6_family;   /* AF_INET6 */
               in_port_t       sin6_port;     /* port number */
               uint32_t        sin6_flowinfo; /* IPv6 flow information */
               struct in6_addr sin6_addr;     /* IPv6 address */
               uint32_t        sin6_scope_id; /* Scope ID (new in 2.4) */
           };
```
The others are similar, and in particular they all call the field `sin6_scope_id`.

If it were just Python, I'd be inclined to fix it. But it looks like someone misread a draft of the RFC circa 2000 (see reference in the BSD manpage linked above) when implementing it, and the OSes are all stuck with that. I think it's best to follow the OSes' naming, so I withdraw the suggestion.

I appreciate this clear and succinct description in the docstring! I think under the circumstances that's the best we can do. (That, and similar phrasing in documentation.)

Lib/ipaddress.py
```diff
@@ -1834,8 +1855,25 @@ def __init__(self, address):
         addr_str = str(address)
         if '/' in addr_str:
             raise AddressValueError("Unexpected '/' in %r" % address)
+        addr_str, self._scope_id = self._split_scope_id(addr_str)
+
         self._ip = self._ip_int_from_string(addr_str)
 
+    def __str__(self):
+        ip_str = super().__str__()
+        return ip_str + '%' + self._scope_id if self._scope_id else ip_str
```
gnprice, line 1916: So this adds the scoped syntax, with `%`, to the result of `str` on an address.

Are there places where that's likely to wind up in a URL? The RFC in section 11.7 expressed concern about this syntax's interaction with URLs -- in particular, the `%` will probably need to be percent-encoded.

I think if that causes a problem, it's a bug in the code that made the URL; so we can't do everything to prevent it. But:
* To help give notice to third-party code, we should be sure to make the scoped-address case visible in the docs and in What's New.
* We should check the stdlib itself for such bugs. :slightly_smiling_face:

... OK, it looks like nothing else in the stdlib refers to this module, so that takes care of point 2.

aixtools:
FYI - performing a "ci" for AIX (# git pr 13772; <build>; make test.

If I have time later I'll report the results this evening, otherwise tomorrow.

FYI: AIX does not "work like Linux" when it comes to scoped ids and IPv6. This has come up before, and sadly, I lack the IPv6 skills to develop suitable tests to figure out what needs to be done to react similarly.

Doc/library/ipaddress.rst
```diff
@@ -268,6 +273,12 @@ write code that handles both IP versions correctly.  Address objects are
       ``::FFFF/96``), this property will report the embedded IPv4 address.
       For any other address, this property will be ``None``.
 
+   .. attribute:: scope_id
+
+      For addresses that appear to be scoped addresses (containing
+      ``%<scope_id>``) as defined by :RFC:`4007`, this property will consist
+      scope id.  For any other address, this property will be ``None``.
```
gnprice: I like the docstring's wording better. Here, how about:

> For scoped addresses as defined by :RFC:`4007`, this property identifies the particular zone of the address's scope that the address belongs to, as a string. When no scope zone is specified, this property will be ``None``.

Doc/library/ipaddress.rst
```diff
@@ -219,9 +219,14 @@ write code that handles both IP versions correctly.  Address objects are
       ``"::abc:7:def"``.
    2. An integer that fits into 128 bits.
    3. An integer packed into a :class:`bytes` object of length 16, big-endian.
+   4. String IPv6 address may contain scope (or zone) id in format <address>%<scope_id>.
+      Scope id could not be blank or negative integer.
+      See :RFC:`4007` for details.
```
gnprice: Here's an alternate wording. One substantive change: negative integers are fine, and `%` is disallowed. (The RFC says non-negative integers *must* be supported, but everything else that's non-blank and has no `%` is OK to support.)

... Oh also, this should be part of point 1. Let's perhaps make it a second paragraph there.

... Oh hey, GitHub has this "suggest edit" feature. I'll try it.

Doc/library/ipaddress.rst
```diff
@@ -219,9 +219,14 @@ write code that handles both IP versions correctly.  Address objects are
       ``"::abc:7:def"``.
```
gnprice, line 219: (This would replace item 4 just below.)

```suggestion
       ``"::abc:7:def"``.

       Optionally, the string may also have a scope zone ID, expressed with a suffix ``%<scope_id>``. If present, the scope ID must be non-empty, and may not contain ``%``. See :RFC:`4007` for details. For example, ``fe80::1234%1`` might identify address ``fe80::1234`` on the first link of the node.  
```

Doc/library/ipaddress.rst
```diff
@@ -311,15 +322,17 @@ IPv6).
 Comparison operators
 """"""""""""""""""""
 
-Address objects can be compared with the usual set of comparison operators.  Some
-examples::
+Address objects can be compared with the usual set of comparison operators. Scope id
+is not considered while comparing IPv6 address objects. Some examples::
```
gnprice: Ah, is this the behavior we want?

I think probably not. In particular the whole point of this scoped-address system, if I'm understanding it right, is that the address may be local to a particular scope -- so that both link 1 and link 2 might have an address called `fe80::1234`, and those would be completely unrelated addresses that might belong to entirely different machines.

So we really want `IPv6Address('fe80::1234%1') != IPv6Address('fe80::1234%2')`.

Doc/library/ipaddress.rst
```diff
@@ -219,9 +219,14 @@ write code that handles both IP versions correctly.  Address objects are
...
    2. An integer that fits into 128 bits.
    3. An integer packed into a :class:`bytes` object of length 16, big-endian.
+   4. String IPv6 address may contain scope (or zone) id in format <address>%<scope_id>.
+      Scope id could not be blank or negative integer.
+      See :RFC:`4007` for details.
 
    >>> ipaddress.IPv6Address('2001:db8::1000')
    IPv6Address('2001:db8::1000')
+   >>> ipaddress.IPv6Address('2001:db8::1000%1')
+   IPv6Address('2001:db8::1000%1')
```
gnprice: Is this example actually well-formed? If I'm understanding the RFC right, this example and the one above it can't actually both be valid (even if the module doesn't comprehensively validate such things and so the transcript is accurate.)

For the first example to be valid, `2001:db8::1000` must be a global address.

But in that case it's inappropriate to attach a scope zone ID to it. That only makes sense for non-global addresses. So a better example would be one from the RFC, like `ff02::5678%5`.

gnprice (changes requested):
Thanks for the updates!

I withdraw the suggestion about the API -- I see now that the system-level interfaces (on all major OSes, which followed each other) are stuck with the "scope ID" misreading, and I think that makes us stuck with it too. (Details in subthread.)

I appreciate the clear description in the docstring! Describing the real situation accurately in docs is the best we can do to mitigate the confusion.

Some further comments below, particularly on docs. (And one previous comment, on `self._scope_id = None`, is still open.)

aixtools:
Hmm. Thought I left a comment earlier. No matter.

I did a manual CI of the PR for AIX - results are green.

== Tests result: SUCCESS ==

395 tests OK.

24 tests skipped:
    test_dbm_gnu test_devpoll test_epoll test_gdb test_idle
    test_kqueue test_msilib test_ossaudiodev test_readline test_spwd
    test_sqlite test_startfile test_tcl test_tix test_tk
    test_ttk_guionly test_ttk_textonly test_turtle test_unicode_file
    test_unicode_file_functions test_winconsoleio test_winreg
    test_winsound test_zipfile64

Total duration: 16 min 52 sec
Tests result: SUCCESS

Doc/library/ipaddress.rst
opavlyuk: While writing code, I investigated this issue and came to the decision, that same IPv6 addresses with different zone id (or scope id) should be considered equal. Unfortunately, can not provide RFC or other 100% proof, but in [open-jdk](https://github.com/AdoptOpenJDK/openjdk-jdk11/blob/master/src/java.base/share/classes/java/net/Inet6Address.java#L269-L276) zone id (scope id) is not considered while comparing IPv6 addresses. Also, [stackowerflow discussion](https://stackoverflow.com/questions/1787530/how-do-you-determine-equality-between-two-ipv6-addresses).

Doc/library/ipaddress.rst
opavlyuk: I agree with you. As it is not prohibited by [RFC](https://tools.ietf.org/html/rfc4007#section-11.1), it could bring some misunderstanding. Thanks, fixed.

Lib/ipaddress.py
opavlyuk: As far as I know, it is the best practice to avoid attribute initialization in `if `statement. According to this and [DRY](https://en.wikipedia.org/wiki/Don%27t_repeat_yourself), I prefer to leave it as is.
If it creates any issues, I will definitely fix it.

Lib/ipaddress.py
opavlyuk: Could you clarify you suggestion, please?

Doc/library/ipaddress.rst
opavlyuk: Fixed, thanks.

Doc/library/ipaddress.rst
opavlyuk: Sounds reasonable, fixed.

Doc/library/ipaddress.rst
opavlyuk: Fixed.

Doc/library/ipaddress.rst
aeros: > Unfortunately, can not provide RFC or other 100% proof, but in open-jdk zone id (scope id) is not considered while comparing IPv6 addresses.

@opavlyuk Is there anything else besides open-jdk which also considers two of the same IPv6 addresses to be equal even when their scope IDs differ?

Personally, I think it's reasonable for this to be a separate PR and/or issue. At the moment, there is no support at all for scoped IPv6 addresses (as far as I'm aware) within stdlib. In it's current state, this PR provides a significant improvement.

Lib/ipaddress.py
aeros: I think following DRY here is the right thing to do, and it's not at all uncommon to initialize default values for attributes in ``__init__``. Generally, DRY isn't heavily prioritized when it's just two repetitions. But if more cases/conditional blocks are added in the future, writing ``self._scope_id = None`` (or ``self._zone_id``) in all of them could get quite repetitive.

Doc/library/ipaddress.rst
opavlyuk: Found out, that in [dotnet core](https://github.com/dotnet/corefx/blob/master/src/System.Net.Primitives/src/System/Net/IPAddress.cs#L553-L578) scope (zone) id **is considered** while comparing IPv6 addresses.

Need help with the comparison issue, because can not find any reputable sources.

aeros: @opavlyuk If there's no response after a decent period in the PR itself, it might be worthwhile to open a new topic over on Python-Dev to see what the others think the behavior should be.

Lib/ipaddress.py
gnprice: Hmm, I find my feeling is I would actually much rather write this explicitly in the two `if` cases.

Some thoughts that perhaps explain why:

* This isn't a default value. It's the value that applies in two possible cases... specifically the two cases that are sort of off to the side and special. In the case that (a) is conceptually the main one, as evidenced for example by appearing in virtually all of the examples in the module's docs, and (b) contains the most code, the value set here always gets clobbered by a later assignment.

  So when you're reading this function and just want to first understand the main, normal, case, this line is a red herring because it never has an effect.

  (It happens that in the main case, this value also often/usually ends up being `None`. But that just makes having this line more confusing, because it *looks* so much like it's what causes that to happen, when in fact that's entirely controlled by `_split_scope_id`.)

* It's true that there are two cases where this value applies. But that's more like two facts than one fact: the `int` form of the constructor doesn't take a scope zone ID, and the `bytes` form doesn't take one, though the main `str` (or str-able) form of the constructor does.

  One way of seeing that: if we imagine that some fourth form were to be added to this constructor in the future, it's not at all obvious whether that form would accept a scope zone ID or wouldn't.

  Because this is two independent facts about two different forms of the constructor, it's best to express each of those facts separately.

  If the code expressing this fact were more complicated, I'd certainly be in favor of writing down that complexity once... and then I'd again want two separate lines to invoke it (e.g. to call some helper method) for the two different cases.

Doc/library/ipaddress.rst
```diff
@@ -217,16 +217,20 @@ write code that handles both IP versions correctly.  Address objects are
       :RFC:`4291` for details.  For example,
       ``"0000:0000:0000:0000:0000:0abc:0007:0def"`` can be compressed to
       ``"::abc:7:def"``.
+
+      Optionally, the string may also have a scope (zone) id, expressed
+      with a suffix ``%<scope_id>``. If present, the scope id must be non-empty,
+      and may not contain ``%``.
+
+      See :RFC:`4007` for details.
```
gnprice, line 224: Small nit: remove paragraph break. With the paragraph break, it looks like this reference is about the entire item 1, rather than specifically the "scope zone ID" bit.

Doc/library/ipaddress.rst
```diff
@@ -217,16 +217,20 @@ write code that handles both IP versions correctly.  Address objects are
       :RFC:`4291` for details.  For example,
       ``"0000:0000:0000:0000:0000:0abc:0007:0def"`` can be compressed to
       ``"::abc:7:def"``.
+
+      Optionally, the string may also have a scope (zone) id, expressed
+      with a suffix ``%<scope_id>``. If present, the scope id must be non-empty,
+      and may not contain ``%``.
```
gnprice, line 223: Spelling nit: uncapitalized "id" is indeed a word, but it is not the one you mean here. :wink: Here's how one dictionary puts it:

> The part of the mind in which innate instinctive impulses and primary processes are manifest.
https://www.lexico.com/en/definition/id

Like everything else does, the abbreviation for "identifier" gets lowercased in some code contexts -- but in normal text it's always capitalized: an ID.

Doc/library/ipaddress.rst
```diff
@@ -217,16 +217,20 @@ write code that handles both IP versions correctly.  Address objects are
       :RFC:`4291` for details.  For example,
       ``"0000:0000:0000:0000:0000:0abc:0007:0def"`` can be compressed to
       ``"::abc:7:def"``.
+
+      Optionally, the string may also have a scope (zone) id, expressed
```
gnprice: The text I suggested was "scope zone ID", but this version makes it "scope (zone) ID" (modulo the spelling nit mentioned separately.)

There's a big difference there!
* With the parens, this says: "this is an ID for a scope, and oh by the way a scope is also called a zone."
* Without parens, this says: "this is an ID for a scope zone".

I wrote "scope zone ID" because in fact the thing this identifies is exactly the thing the RFC calls a "scope zone". And it's not at all true that "scope" and "zone" mean the same thing -- in the RFC's terms, they are quite different, related, things.

Here's how it describes the relationships between scopes and zones:

> A scope zone, or simply a zone, is a connected region of topology of a given scope.  For example, the set of links connected by routers within a particular (multicast) site, and the interfaces attached to those links, comprise a single zone of multicast site-local scope.
>
> Note that a zone is a particular instance of a topological region (e.g., Alice's site or Bob's site), whereas a scope is the size of a topological region (e.g., a site or a link).

We're constrained to saying "scope ID" when we need a short phrase, rather than saying "zone ID", because of the interfaces that are now baked into OSes.

But that is always going to be confusing for anyone who sees "scope ID", and then goes and reads the RFC and tries to understand what "a scope" actually is, because "scope" is actually the term for a different concept in the same design.

We can at least help mitigate the confusion by once, when first introducing the concept, using the full phrase "scope zone ID". That provides a connection to the RFC's term "scope zone".

gnprice: Oh also, my suggested text included a brief example, parallel with the example in the non-scoped case just above. This isn't as critical as the above remark, but I think an example helps a lot in making it concrete, and in particular in helping readers who don't have the jargon of this particular corner of IPv6 fresh in their heads -- i.e. most of them :slightly_smiling_face: -- either recognize where they've seen these things, or easily determine that it isn't some unfamiliar way of talking about a thing they have in fact seen.

Doc/library/ipaddress.rst
gnprice: Some person's opinion on StackOverflow is not a particularly reliable source for deciding the right way to handle something. :smile: It's often a good source of links, or names of things to go read up on in the documentation/standards/etc... but there are a *lot* of bad and buggy suggestions there.

That leaves us, at the end of the day, responsible for our own reading of the underlying standard and other technical context.

And what I see from that reading is: The entire point of RFC 4007 -- of the whole system of scoped addresses, zones, and writing down as part of an address some identifier for a scope zone -- is that the same 128-bit address can mean different things in different zones. The zone identifier is there to distinguish those different meanings.

The [Wikipedia article](https://en.wikipedia.org/wiki/IPv6_address#Scoped_literal_IPv6_addresses) (following a link from that SO thread! just like I mentioned :wink: ) puts this nicely:

> For addresses with other than global scope [...], and in particular for link-local addresses [...] the same address may be valid in different zones, and be in use by a different host in each of those zones.

(There's nothing new there -- the RFC says all the same things, just with more words.)

Given that the whole point of the system is to give different meanings to the same 128 bits in different zones, the only consistent semantics for `==` on such things is that the same 128-bit address plus two different zone IDs makes two distinct, not-equal values.


> Found out, that in dotnet core scope (zone) id is considered while comparing IPv6 addresses.

Cool. This is helpful for confirming that we won't somehow face interoperability problems where the whole rest of the world turns out to have settled on a shared buggy behavior here.


> I think it's reasonable for this to be a separate PR and/or issue.

I have to disagree with this. Sometimes it's acceptable to introduce broken semantics in a new area, and plan to fix it up later, when implementing the right semantics is a lot of work... but that certainly isn't the case here. We shouldn't introduce the buggy behavior in the first place.

gnprice (changes requested):
Thanks @opavlyuk for the updates!

Some comments below. Some are small nits, and others are easy to fix but I think matter a great deal for what this PR is doing.

I also want to mention one thing about the review/response loop itself. I offered a suggested text for the "constitutes a valid IPv6 address" part of the docs, and you replied "Fixed." but actually had made several edits to that text, one of them quite significant.

I nearly missed that you had made those edits, because you'd made that reply and didn't say anything about choosing different text instead. It's naturally fine to write something different from someone's suggestion, but it's important to communicate clearly about it so the reviewer isn't misled. Please always be transparent -- something like "Fixed, with some small edits; take a look." does the job.

Lib/ipaddress.py
gnprice: Sure -- the request is this bit:

> To help give notice to third-party code, we should be sure to make the scoped-address case visible in the docs and in What's New.

Concretely that means something like:
* This new feature should be mentioned in the "What's New in 3.9" document. That's `Doc/whatsnew/3.9.rst`. (After all, it's one of the nice new things we'll have in 3.9!)
* Each of the main places in the docs that have a series of examples should include an example with a scoped address, so that it's visible that these addresses with `%` in them are a thing that can happen. That means something like:
  * The initial "Convenience factory functions" section in the module doc. For a lot of people this may be the one section they really read. One example in either the `ip_address` or the `ip_network` case should be enough.
  * The docs bit on the `IPv6Address` constructor. Already there in this version -- thanks!
  * The "Conversion to Strings and Integers" section in the doc.
  * The first block of examples in the "IP Host Addresses" section at the top of `Doc/howto/ipaddress.rst` aka [here](https://docs.python.org/3/howto/ipaddress.html). That's prominently linked to from the module doc, so a lot of people may rely more on that.

Doc/library/ipaddress.rst
```diff
@@ -217,11 +217,20 @@ write code that handles both IP versions correctly.  Address objects are
...
    2. An integer that fits into 128 bits.
    3. An integer packed into a :class:`bytes` object of length 16, big-endian.
 
+
    >>> ipaddress.IPv6Address('2001:db8::1000')
    IPv6Address('2001:db8::1000')
+   >>> ipaddress.IPv6Address('ff02::5678%1')
+   IPv6Address('ff02::5678%1')
 
    .. attribute:: compressed
```
gnprice, line 236: I believe this documentation on `compressed` is no longer accurate. Is that right? If so, let's update it.

(Noticed this while skimming the doc just now to answer the question about where to make this feature visible with examples.)

Lib/ipaddress.py
aeros: @gnprice 

> This isn't a default value. It's the value that applies in two possible cases... specifically the two cases that are sort of off to the side and special.

I may have been a bit presumptuous then and assumed that ``self._scope_id`` being ``None`` was a significantly more common scenario. If it is fairly uncommon that the address passed into the ``IPv6Address`` class will be an integer or byte, and it is unlikely that other potential conditional blocks would set also to ``None``, I would agree with your argument to explicitly declare it within those two blocks instead of its current position.

So just to clarify this suggestion, would it look something like this? I omitted the comments to make it more succinct.

```python3
def __init__(self, address):
        if isinstance(address, int):
            self._scope_id = None
            self._check_int_address(address)
            self._ip = address
            return

        if isinstance(address, bytes):
            self._scope_id = None
            self._check_packed_address(address, 16)
            self._ip = int.from_bytes(address, 'big')
            return

```

gnprice: > So just to clarify this suggestion, would it look something like this?

Yep -- just like that except that I'd probably put the `_scope_id` line just after the `_ip` line, rather than before:

        if isinstance(address, int):
            self._check_int_address(address)
            self._ip = address
            self._scope_id = None
            return

(and similarly in the other block.)

I'd lean that way because:
* the `_ip` part is the main thing this code is about, and the `_scope_id` is the extra added feature that's sometimes important; so leading with the main thing makes it quicker to follow the main logic and tune out the rest
* in the `...%...` syntax where they both appear (the one handled in the main codepath, below), the part corresponding to `_scope_id` comes after the part corresponding to `_ip`. So that makes it natural to think of `_ip` as coming first, then `_scope_id`.

Lib/ipaddress.py
aeros: A comment here would be helpful, the reasoning behind the format isn't inherently obvious.

Doc/library/ipaddress.rst
aeros: @gnprice 

> Some person's opinion on StackOverflow is not a particularly reliable source for deciding the right way to handle something.

Definitely agreed.

> Cool. This is helpful for confirming that we won't somehow face interoperability problems where the whole rest of the world turns out to have settled on a shared buggy behavior here.

> I have to disagree with this. Sometimes it's acceptable to introduce broken semantics in a new area, and plan to fix it up later, when implementing the right semantics is a lot of work... but that certainly isn't the case here. We shouldn't introduce the buggy behavior in the first place.

The distinction between "buggy behavior" vs limited comparison accuracy is not always immediately obvious, that's why I was asking what behavior other platforms utilized. If it is only open-jdk that considers two of the same IPv6 addresses with differing scope IDs to be equal, it seems to lean more towards being a bug. 

Also, as you mentioned, interoperability can also be a concern. Based on the information gathered since then, it seems that won't be an issue for the most part.

If the comparison can be fixed with a reasonable amount of effort within the same PR and the majority of other platforms consider the scopes during comparisons, I agree that the behavior should be adjusted accordingly.

Doc/library/ipaddress.rst
aeros: So, in conclusion, the section should be changed to:

```
      Optionally, the string may also have a scope (zone) ID, expressed
      with a suffix ``%<scope_id>``. If present, the scope ID must be non-empty,
      and may not contain ``%``.
```

I agree with the suggestion, but it can sometimes be easier for the author when the recommended changes are explicitly mentioned.

gnprice: > it can sometimes be easier for the author when the recommended changes are explicitly mentioned.

True indeed! :slightly_smiling_face: See here:
https://github.com/python/cpython/pull/13772#discussion_r313085202

I'd already suggested text for this, which is why this comment focused on explaining my reasoning about a difference between my suggested text and the version subsequently added to the PR.

aeros: Ah, I see. It can be quite difficult to keep track of all of the comments in a PR, especially once the total number of them starts to get high. (:

Lib/ipaddress.py
opavlyuk: @gnprice , great thanks, really very helpful. In progress.

Doc/whatsnew/3.9.rst
```diff
@@ -202,6 +202,16 @@ now raises :exc:`ImportError` instead of :exc:`ValueError` for invalid relative
 import attempts.
 (Contributed by Ngalim Siregar in :issue:`37444`.)
 
+ipaddress
+---------
+
+:mod:`ipaddress` now supports IPv6 Scoped Addresses (IPv6 address with suffix ``%<scope_id>``).
+
+Scoped IPv6 could be parsed to :class:`ipaddress.IPv6Address` constructor.
```
aeros: I would recommend making some minor grammar changes: 

1) "could be" should instead be "can be" 
2) "parsed to" should be "parsed using"
3) I don't think it's necessary to specify "constructor", that's typically implied when referring to a specific class. 

```suggestion
Scoped IPv6 addresses can be parsed using :class:`ipaddress.IPv6Address`.
```

Doc/whatsnew/3.9.rst
```diff
@@ -202,6 +202,16 @@ now raises :exc:`ImportError` instead of :exc:`ValueError` for invalid relative
...
+If present, scope zone ID is avalaible through :attr:`scope_id` attribute.
```
aeros: 1) Spelling fix: "avalaible" -> "available"
2) Added a "the"

```suggestion
If present, scope zone ID is available through the :attr:`scope_id` attribute.
```

opavlyuk: @aeros  Thank you! After checks pass, I`m going to ask for core review via python-dev@python.org again.

Doc/whatsnew/3.9.rst
```diff
@@ -202,6 +202,16 @@ now raises :exc:`ImportError` instead of :exc:`ValueError` for invalid relative
 import attempts.
 (Contributed by Ngalim Siregar in :issue:`37444`.)
 
+ipaddress
+---------
+
+:mod:`ipaddress` now supports IPv6 Scoped Addresses (IPv6 address with suffix ``%<scope_id>``).
+
+Scoped IPv6 addresses can be parsed using :class:`ipaddress.IPv6Address`.
+If present, scope zone ID is available through the :attr:`scope_id` attribute.
```
asvetlov:
```suggestion
If present, scope zone ID is available through the :attr:`~ipaddress.IPv6Address.scope_id` attribute.
```
`:attr:` should mention the full name if not used inside a class context itself.

Lib/ipaddress.py
```diff
@@ -1836,6 +1836,25 @@ def _reverse_pointer(self):
         reverse_chars = self.exploded[::-1].replace(':', '')
         return '.'.join(reverse_chars) + '.ip6.arpa'
 
+    @staticmethod
+    def _split_scope_id(ip_str):
+        """Helper function to parse IPv6 string address with scope id.
```
asvetlov, line 1841: We don't use docstrings for private API usually.
If you want to document the method -- that's fine, please use a comment for this.

Also, the project never uses docstrings for the documentation generation, `Arg:` and `Returns:` doesn't make much sense here.

Lib/ipaddress.py
```diff
@@ -1872,21 +1891,49 @@ def __init__(self, address):
...
 
+    def __str__(self):
+        ip_str = super().__str__()
+        return ip_str + '%' + self._scope_id if self._scope_id else ip_str
+
+    def __hash__(self):
+        return hash((self._ip, self._scope_id))
+
+    def __eq__(self, other):
+        address_equal = _BaseAddress.__eq__(self, other)
```
asvetlov:
```suggestion
        address_equal = super().__eq__(other)
```

Lib/ipaddress.py
```diff
@@ -1872,21 +1891,49 @@ def __init__(self, address):
...
+        if not address_equal or address_equal is NotImplemented:
```
asvetlov:
```suggestion
        if address_equal is NotImplemented:
            return NotImplemented
        if not address_equal:
            return False
```

asvetlov (changes requested):
The main question is: why do you use `scope_id` attribute while RFC 4007 uses `zone_id` notation, e.g. `<address>%<zone_id>`?

Additionally, I have minor notes below.

Lib/ipaddress.py
opavlyuk: Some comments regarding existing library style.
As far, as I understand, ipaddress module came from google and was written according to their  style convention. All private methods in ipaddress are documented in such a style. So, I decided to keep existing code style while adding this method.
In such a case, should I bring it to the cpython code style or let it be as is?

Lib/ipaddress.py
```diff
@@ -1872,21 +1892,52 @@ def __init__(self, address):
...
+    def __hash__(self):
+        return hash((self._ip, self._scope_id))
+
+    def __eq__(self, other):
+        address_equal = super().__eq__(other)
+        if address_equal is NotImplemented:
+            return NotImplemented
+        if not address_equal:
+            return False
+        return self._scope_id == other._scope_id
```
aeros: What do we want to do in the scenario that the other object doesn't have a `_scope_id` attribute? An `AttributeError` would not be particularly clear from a user perspective when comparing the equality of two different IPv6 address objects.

Returning False seems to make the most sense to me, but only in the scenario that self actually has a `_scope_id` that isn't None (since we already established that the addresses are equal at this point). 

```suggestion
        try:
            return self._scope_id == other._scope_id
        except AttributeError:
            if self._scope_id is None:
                return True
            else:
                return False    
```

opavlyuk:
> The main question is: why do you use `scope_id` attribute while RFC 4007 uses `zone_id` notation, e.g. `<address>%<zone_id>`?
> 
> Additionally, I have minor notes below.

Thanks for your review!

I agree, RFC 4007 calls this part "zone_id".
At the same time, in existing Python code and documentation "scope id" is widely used too, e.g.:
[socket module](https://github.com/python/cpython/blob/master/Modules/socketmodule.c#L1873-L1874), [socket module](https://github.com/python/cpython/blob/master/Modules/socketmodule.c#L6643), [socket module tests](https://github.com/python/cpython/blob/master/Lib/test/test_socket.py#L1693-L1744), [socket module docs](https://docs.python.org/3/library/socket.html#socket-families).
That examples include using of "scope_id" as part of AF_INET6 address family representation.

Also, according to @gnpince,


> Oof, I see -- and in fact it's that way in the OS-level interfaces on [Linux](http://man7.org/linux/man-pages/man7/ipv6.7.html), [BSD](https://www.freebsd.org/cgi/man.cgi?query=inet6&sektion=4&apropos=0&manpath=FreeBSD+12.0-RELEASE), and Windows ([discussion](https://docs.microsoft.com/en-us/windows/win32/winsock/link-local-and-site-local-addresses-2), [struct](https://docs.microsoft.com/en-us/windows/win32/winsock/sockaddr-2)). Here's Linux:

           struct sockaddr_in6 {
               sa_family_t     sin6_family;   /* AF_INET6 */
               in_port_t       sin6_port;     /* port number */
               uint32_t        sin6_flowinfo; /* IPv6 flow information */
               struct in6_addr sin6_addr;     /* IPv6 address */
               uint32_t        sin6_scope_id; /* Scope ID (new in 2.4) */
           };
>The others are similar, and in particular they all call the field sin6_scope_id.

So, I decided to name it `scope_id`, but I`m open for discussion.

Lib/ipaddress.py
opavlyuk: As I see it, `IPv6Address` object shall always have a `_scope_id` attribute. In case it is created from IPv6 address without scope id, the attribute would be set to `None`.
Case with absent `_scope_id` attribute could appear only if someone removes it explicitly.
May be, it would be ok to use public property `scope_id` in `__eq__` method, and add `try` block to it?
```
    @property
    def scope_id(self):
        """Identifier of a particular zone of the address's scope.

        See RFC 4007 for details.

        Returns:
            A string identifying the zone of the address if specified, else None.

        """
        return getattr(self, '_scope_id', None)    

    def __eq__(self, other):
        address_equal = super().__eq__(other)
        if address_equal is NotImplemented:
            return NotImplemented
        if not address_equal:
            return False
        return self.scope_id == other.scope_id

Comment (aeros) [reply Lib/ipaddress.py]: @opavlyuk
> As I see it, IPv6Address object shall always have a _scope_id attribute.
> Case with absent _scope_id attribute could appear only if someone removes it explicitly.

If the object was instantiated from the version of the `IPv6Address` class in this PR (where `_scope_id` was just added), then yes. But if it's an object instantiated from an older version of the class or some other custom implementation, then `_scope_id` might not be present.

IMO, it seems worthwhile to provide some basic amount of compatibility with objects that don't have a `_scope_id` attribute, considering it is just being added now.

```py
    @property
    def scope_id(self):
        return getattr(self, '_scope_id', None)    

    def __eq__(self, other):
        address_equal = super().__eq__(other)
        if address_equal is NotImplemented:
            return NotImplemented
        if not address_equal:
            return False
        return self.scope_id == other.scope_id
```

The above would _only_ cover the situations in which the `self` object doesn't have a `_scope_id`, no? My concern is with the `other` object, as `self` is much more likely to have a `_scope_id` attribute in the first place. The above situation I was referring to was when `other` is an object that wasn't instantiated from this version of `IPv6Address` or is using some other implementation.

Also, in this case, I find the EAFP (`try-except`) approach in my above suggestion to be more explicit than LBYL (if `hasattr()` or `getattr()` with default). Mostly because the situation in which the other object does not have a `_scope_id` is an *exception* and not just a general conditional to check for.

A `try-except` is generally better performance wise (when in the vast majority of cases, the exception does not occur) in comparison to using `getattr()` with a default every single time the property is accessed or `if hasattr()` every time `__eq__()` is used.   

What do you think @asvetlov?

opavlyuk:

Lib/ipaddress.py
opavlyuk: @aeros , now I understood, and it makes sense.
At the same time, following seems more clean for me:
```
    def __eq__(self, other):
        address_equal = super().__eq__(other)
        if address_equal is NotImplemented:
            return NotImplemented
        if not address_equal:
            return False
        return self.scope_id == getattr(other, _'scope_id', None)
```

So, we consider, if other object has no `_scope_id` attribute, then it is the same as it is `None`.
Having measured the execution time on Python 3.7.4 on my GNU\Linux machine, I saw that option with `getattr` seems to be slightly faster in case when attribute exists and 5x faster in case with non-existent attribute.

Lib/ipaddress.py
asvetlov: Yes, you are right.
Let's keep as is

Lib/ipaddress.py
asvetlov: I like the last version

asvetlov:
Ok, `scope_id` naming after the socket module makes sense.
However, `socket` uses `scopeid` without underscore.
This name is used for documentation only and never present as Python name (e.g. attribute).
I think we should either update existing docs or rename the attribute from this PR.
Any solution is good. The compatibility with `sin6_scope_id` (with an underscore) is not important now, very few Python devs work with C structures at the same time. I expect they are experienced enough to figure out the difference.
The naming problem is minor but I want to prevent silly problems in the future.

Lib/ipaddress.py
aeros: I have no objections to the last version either, ``self.scope_id == getattr(other, _'scope_id', None)`` should work fine. My main issue with the first one was using ``return getattr(self, '_scope_id', None)`` as the getter for ``_scope_id``, but that's not an issue in the most recent version.

Edit: I still think the EAFP approach might be a bit more idiomatic in this case. However, if it's a significant performance difference, ``getattr`` with a default should be okay.

aeros: I'm curious as to how this would perform in comparison to the above two, I might test it out later when I get the chance:

```py

try:
    other_scope_id = other._scope_id
except AttributeError:
    other_scope_id = None
return self._scope_id == other_scope_id

Comment (opavlyuk) [reply Lib/ipaddress.py]: @aeros I've pushed my version, but lets make the final decision after your benchmark.

**Comment by opavlyuk** [comment]
@asvetlov As for me, `scope_id` with underscore looks more pythonic.
 Also, [socket](https://docs.python.org/3/library/socket.html?highlight=socket#module-socket) documentation uses `scopeid` as parameter name and `%scope` as part of  IPv6 address. So I tried to unify naming in `socket` documentation and to make it fit `sin6_scope_id`, which is underscored too.

File: Lib/ipaddress.py (hunk 25 above)
Comment (aeros) [reply Lib/ipaddress.py]: @opavlyuk
> I've pushed my version, but lets make the final decision after your benchmark.

Sounds good to me. I should have time to evaluate it later, I've been a bit occupied recently with studying for finals and working on other projects this last week.

If it gets to the point where everything else is finished and this is the last remaining step, feel free to move ahead with the PR. At the end of the day, I think either version would be acceptable.

**Comment by opavlyuk** [comment]

File: Doc/whatsnew/3.9.rst [hunk 26]
```diff
@@ -202,6 +202,16 @@ now raises :exc:`ImportError` instead of :exc:`ValueError` for invalid relative
...
 (Contributed by Ngalim Siregar in :issue:`37444`.)
 
+ipaddress
+---------
+
+:mod:`ipaddress` now supports IPv6 Scoped Addresses (IPv6 address with suffix ``%<scope_id>``).
+
+Scoped IPv6 addresses can be parsed using :class:`ipaddress.IPv6Address`.
+If present, scope zone ID is available through the :attr:`~ipaddress.IPv6Address.scope_id` attribute.
+
```
aeros: Minor nit: could you remove the empty line of whitespace between the news entry and the "contributed by"?

```suggestion
```

Lib/ipaddress.py
```diff
@@ -1849,7 +1869,7 @@ class IPv6Address(_BaseV6, _BaseAddress):
 
     """Represent and manipulate single IPv6 Addresses."""
 
-    __slots__ = ('_ip', '__weakref__')
+    __slots__ = ('_ip', '_scope_id', '__weakref__')
```
aeros:
```suggestion
    __slots__ = ('_ip', 'scope_id', '__weakref__')
```
I believe the underscore before `scope_id` is unneeded here and in the `__init__()`. IIRC, when an instance attribute has a defined `@property`, `class.attr` will call the defined property method instead of directly accessing the attribute.

Doc/whatsnew/3.9.rst
opavlyuk: Thank you!

opavlyuk:
@asvetlov , please, could you review my changes?

terryjreedy:
About putting repeated assertions in a loop:
1. Before the addition of subTest in 3.4, doing so was only cosmetic;
    with subTest, it is now a slight improvement.
2. Putting 2 calls in a loop is overkill;  doing so with 3 or even 4 might be considered marginal;
    putting 8 calls in a loop is not.

So, if I were maintaining this module, I would want the refactoring.
BUT: simply adding new lines makes review of this large patch easier.
So I agree with leaving it as is for now.

the-knights-who-say-ni:

## Recognized GitHub username

@damani42, @RandyMcMillan, @Baljak

terryjreedy:
The hundreds of commits listed above is an error that might be fixed by a simple update.  But the file conflict has to be fixed first.  I will try to do so.  The conflict shows why What's New entries should not be added until a PR is merge-ready.

opavlyuk:
@terryjreedy Sorry, previous reviewers strongly recommended adding that What'sNew (https://github.com/python/cpython/pull/13772#discussion_r313070495, https://github.com/python/cpython/pull/13772#discussion_r314589999), so I followed their advice.
Should I keep it, or remove until PR is merge-ready?

ethanfurman:
@terryjreedy, I thought the point of the new blurb tool was to avoid What's New troubles -- am I missing something?

terryjreedy:
@lisroach   Lisa, you approved the idea on the tracker, and are the only core dev to give an opinion.  Can you review at least part of the patch?  Do you know ipaddress well enough to merge?

terryjreedy:
@ethanfurman Blurb solved conflict problems with the single news items file, Misc/NEWS.rst, linked to after html-ified, as *changelog* from What's New.   The single whatsnew/3.9.rst has the same issue as having a single NEWS.rst did.  Traditionally, entries were not added with commits, but were selectively added later by a few editors trolling the news file.  Adding as we go, with changes in PRs that sit around for months, recreates the old problem.

terryjreedy:
@opavlyuk There are 3 issues more important at the moment than the whatsnew item.

1. CLA: bot initially said 'signed', then revoked that without explanation, more that a day ago, though I no longer see the initial revocation notice.

2. Commit #: header says you want 'to merge 294 commits', which is wrong.  We usually close PRs opened like this and ask for a resubmission.  But I believe mistake came later, apparently after initial reviews, which we do not want to close, perhaps with an unneeded and bad force push.  (Github is not showing the history of your commits here like it usually does.)  I hoped that update merges would fix this (as it did on another issue), but no.

3. Open change request: click [+-] box after gnprice to find them again.

Please ask on core-mentorship list how to deal with 1 and 2 and anything about 3 that you are not sure of.

Doc/library/ipaddress.rst
opavlyuk: Fixed.

Doc/library/ipaddress.rst
opavlyuk: Fixed.

Lib/ipaddress.py
opavlyuk: Fixed.

opavlyuk:
@terryjreedy , about issues:
1. Fixed while resolving 2-nd issue;
2. I figured out, that extra commits appeared after I merged `master` branch from cpython repo directly to current issue branch, instead of merge it first to `master` branch of my repo and futher merge to issue branch. Fixed it with force push.
3. As for now, I have a lack of time to write expalnations why I haven't listened to every gnprice`s advice, but I`ve already applied all suggestions, that I consider reasonable.

terryjreedy:
@gnprice  Of your many suggestions, opavlyuk has "applied all suggestions, that I consider reasonable".  Of the rest, are there any you consider critical, or would you be comfortable merging as is?

Doc/whatsnew/3.9.rst
```diff
@@ -330,6 +339,13 @@ Optimizations
 Build and C API Changes
 =======================
 
+* Provide :c:func:`Py_EnterRecursiveCall` and :c:func:`Py_LeaveRecursiveCall`
+  as regular functions for the limited API. Previously, there were defined as
+  macros, but these macros didn't work with the limited API which cannot access
+  ``PyThreadState.recursion_depth`` field. Remove ``_Py_CheckRecursionLimit``
+  from the stable ABI.
+  (Contributed by Victor Stinner in :issue:`38644`.)
+
```
asvetlov: Please drop these lines, they are added by accident.
Victor decided to move this text to another place.

asvetlov (approved):
Looks good and ready to merge.
Please drop a few lines from `whatsnew`.

Sorry, for the long review, I was pretty busy.
Thanks for your patience.

Doc/whatsnew/3.9.rst
opavlyuk: Thanks a lot! Fixed and also corrected typo in my own surname (what a shame).

asvetlov:
Thanks!

terryjreedy:
Error message is
```
command timed out: 1200 seconds without output running ['./configure', '--prefix', '$(PWD)/target', '--with-pydebug'], attempting to kill
process killed by signal 9
program finished with exit code -1
elapsedTime=2048.933014
checking for linux/vm_sockets.h... 
```
Since patch did not add a C include, it is not obvious that this is really related.  Since there have not been any subsequent merges and builds, I triggered a rebuild.

opavlyuk:
@terryjreedy thanks! 

```Traceback (most recent call last):
--- <exception caught here> ---
  File "/data/buildbot/venv/lib/python3.8/site-packages/buildbot/process/build.py", line 375, in startBuild
    ping_success_or_failure = yield workerforbuilder.ping()
twisted.spread.pb.PBConnectionLost: [Failure instance: Traceback (failure with no frames): <class 'twisted.internet.error.ConnectionLost'>: Connection to the other side was lost in a non-clean fashion.
]
```
I have no idea how my patch could cause such things. It is likely an environment issue.

terryjreedy:
The next build, 288, passed, so I think we are good for this issue.

asvetlov:
Agree

PR #13779, opened 2019-06-03
bpo-37137: Fix test_asyncio: use TestCase.set_event_loop()

vstinner:
Replace asyncio.set_event_loop() with TestCase.set_event_loop() of
test_asyncio.utils: this method calls TestCase.close_loop() which
waits until the executor completes, to avoid leaking dangling
threads.

### DIFF
--- a/Lib/test/test_asyncio/test_tasks.py
+++ b/Lib/test/test_asyncio/test_tasks.py
@@ -2775,7 +2775,7 @@ class BaseTaskIntrospectionTests:
         self.assertEqual(asyncio.all_tasks(loop), set())
 
 
-class PyIntrospectionTests(unittest.TestCase, BaseTaskIntrospectionTests):
+class PyIntrospectionTests(test_utils.TestCase, BaseTaskIntrospectionTests):
     _register_task = staticmethod(tasks._py_register_task)
     _unregister_task = staticmethod(tasks._py_unregister_task)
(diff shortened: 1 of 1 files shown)


asvetlov (approved):
LGTM and thank you very much.
You are jedi for tests fighting!

Lib/test/test_asyncio/test_tasks.py
```diff
@@ -2799,12 +2799,7 @@ class BaseCurrentLoopTests:
     def setUp(self):
         super().setUp()
         self.loop = asyncio.new_event_loop()
-        asyncio.set_event_loop(self.loop)
-
-    def tearDown(self):
-        self.loop.close()
-        asyncio.set_event_loop(None)
-        super().tearDown()
+        self.set_event_loop(self.loop)
```
vstinner, line 2802: @asvetlov: I'm not sure about this change, but since tests still pass, it seems like the change is safe.

vstinner:
> LGTM and thank you very much.

To be clear: I have no idea if this PR fix the issue, since I'm unable to reproduce https://bugs.python.org/issue37137

> You are jedi for tests fighting!

Well, I modified regrtest to make a test fail if there is "dangling thread", so I know this issue very well. I also reported https://bugs.python.org/issue34037 And I wrote TestCase.close_loop() and TestCase.set_event_loop() :-)

PR #13780, opened 2019-06-03
Revert "bpo-33608: Factor out a private, per-interpreter _Py_AddPendingCall(). (gh-13714)"

vstinner:
This reverts commit 6a150bcaeb190d1731b38ab9c7a5d1a352847ddc.

https://bugs.python.org/issue33608

### DIFF
--- a/Include/internal/pycore_ceval.h
+++ b/Include/internal/pycore_ceval.h
@@ -12,22 +12,19 @@ extern "C" {
 #include "pycore_pystate.h"
 #include "pythread.h"
 
+PyAPI_FUNC(void) _Py_FinishPendingCalls(_PyRuntimeState *runtime);
 PyAPI_FUNC(void) _PyEval_Initialize(struct _ceval_runtime_state *);
 PyAPI_FUNC(void) _PyEval_FiniThreads(
-    struct _ceval_runtime_state *);
+    struct _ceval_runtime_state *ceval);
 PyAPI_FUNC(void) _PyEval_SignalReceived(
-    struct _ceval_runtime_state *);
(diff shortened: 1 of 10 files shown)


ericsnowcurrently:
Please wait. :)  I'm going to take a look.

vstinner:
> Please wait. :) I'm going to take a look.

There is a release scheduled for today. Nobody is able to reproduce nor debug the issue. The change was pushed 2 days ago: the buildbot policy says that you have 24h to fix regressions:
https://pythondev.readthedocs.io/ci.html#revert-on-fail

Sorry but we are busy enough with enough other bugs:
https://mail.python.org/pipermail/python-dev/2019-June/157826.html

I don't want to have to investigate this issue. Moreover, this bug is known for months: see https://bugs.python.org/issue33608 history.

The only motivation for the revert is to repair the CI. It doesn't mean that your change is wrong.

Your change doesn't have to be in the master branch to be able to investigate: use "git checkout 6a150bcaeb190d1731b38ab9c7a5d1a352847ddc" to get it.

vstinner:
> I don't want to have to investigate this issue.

Hum, to be clear, I already spent 2 hours today to attempt to investigate this issue, Pablo also investigated the issue:

* https://bugs.python.org/issue37135
* https://bugs.python.org/issue37143

ericsnowcurrently (approved):
As noted in the tracker, I agree that this should be reverted for the sake of the beta 1 release.

ericsnowcurrently:
Thanks for again spending time trying to figure this out!  The worst thing for me is wasting other people's time.

PR #13787, opened 2019-06-03
bpo-37146: disable opcache when Py_DEBUG is defined

methane:
--with-pydebug is commonly used to find memory leaks.
But opcache makes it harder.
So disable opcache when Py_DEBUG is defined.

https://bugs.python.org/issue37146

### DIFF
--- a/Python/ceval.c
+++ b/Python/ceval.c
@@ -103,7 +103,14 @@ static long dxp[256];
 #endif
 
 /* per opcode cache */
+#ifdef Py_DEBUG
+// --with-pydebug is used to find memory leak.  opcache makes it harder.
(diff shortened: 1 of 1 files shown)


pablogsal:
Check out https://github.com/python/cpython/pull/13789/files for a possible more complete fix

vstinner (approved):
LGTM. That's a reasonable workaround for https://bugs.python.org/issue37146 for beta1. We can design a better solution after beta1.

pablogsal (approved):

vstinner:
@methane: Thanks. I merged your PR to make Refleak buildbots green again (I hope!).

PR #13788, opened 2019-06-03
gh-81274: Allow http.client to parse non-ASCII header names

tipabu:
Previously, when `http.client` tried to parse a response from an out-of-spec server that sent a header with a non-ASCII name, `email.feedparser` would assume that the non-compliant header must be part of a message body and abort parsing. However, `http.client` already determined the boundary between headers and body and only passed the headers to the parser. As a result, any headers after the first non-compliant one would be silently (!) ignored. This could include headers important for message framing like `Content-Length` and `Transfer-Encoding`.

In the long-long ago, this parsing was handled by the `rfc822` module, which didn't care about which bytes were in the header as long as there was a colon in the line.

Now, add an optional argument to the email parsers to decide whether to require strict RFC-compliant header names. Default this to `True` to minimize the possibility of breaking other callers. In `http.client`, which already knows where the headers end and body begins, use `False`.

Note that the non-ASCII names will be decoded as ISO-8859-1 in keeping with how header values are decoded.

https://bugs.python.org/issue37093

* Issue: gh-81274

### DIFF
--- a/Lib/email/parser.py
+++ b/Lib/email/parser.py
@@ -38,38 +38,53 @@ class Parser:
         self._class = _class
         self.policy = policy
 
-    def parse(self, fp, headersonly=False):
+    def parse(self, fp, headersonly=False, strictheaders=True):
         """Create a message structure from the data in a file.
 
         Reads all the data from the file and returns the root of the message
-        structure.  Optional headersonly is a flag specifying whether to stop
-        parsing after reading the headers or not.  The default is False,
-        meaning it parses the entire contents of the file.
+        structure.
+
+        Optional headersonly is a flag specifying whether to stop parsing
+        after reading the headers or not.  The default is False, meaning it
+        parses the entire contents of the file.
+
+        Optional strictheaders is a flag specifying whether to require
+        RFC-compliant header names.  The default is True, causing parsing to
+        abort if non-compliant header names are encountered.
         """
-        feedparser = FeedParser(self._class, policy=self.policy)
+        feedparser = FeedParser(
+            self._class, policy=self.policy, strictheaders=strictheaders)
         if headersonly:
             feedparser._set_headersonly()
         while data := fp.read(8192):
             feedparser.feed(data)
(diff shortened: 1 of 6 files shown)


Lib/email/parser.py
```diff
@@ -38,15 +38,15 @@ def __init__(self, _class=None, *, policy=compat32):
...
+    def parse(self, fp, headersonly=False, strictheaders=True):
         """Create a message structure from the data in a file.
 
         Reads all the data from the file and returns the root of the message
         structure.  Optional headersonly is a flag specifying whether to stop
         parsing after reading the headers or not.  The default is False,
         meaning it parses the entire contents of the file.
         """
-        feedparser = FeedParser(self._class, policy=self.policy)
+        feedparser = FeedParser(self._class, policy=self.policy, strictheaders=strictheaders)
```
ZackerySpytz: Please limit lines to 79 characters (PEP 8).

Misc/NEWS.d/next/Library/2019-06-17-08-42-34.bpo-37093.T2sOF8.rst
```diff
@@ -0,0 +1 @@
+http.client now parses non-ASCII header names.
```
ZackerySpytz:
```rst
:mod:`http.client`
```

ZackerySpytz (commented):
Please update the documentation.

Misc/NEWS.d/next/Library/2019-06-17-08-42-34.bpo-37093.T2sOF8.rst
tipabu: Done.

Lib/email/parser.py
tipabu: There's still a litany of line-length violations in `Lib/http/client.py` and `Lib/test/test_httplib.py` but I think now at least I'm not making things any _worse_.

PR #13789, opened 2019-06-03
bpo-37146: Allow to configure opcode cache and fix huntleaks by presetting it to 1

pablogsal:
https://bugs.python.org/issue37146

### DIFF
--- a/Include/internal/pycore_pystate.h
+++ b/Include/internal/pycore_pystate.h
@@ -43,6 +43,10 @@ struct _pending_calls {
     int last;
 };
 
+struct _opcodecache_config {
+    int minruns;
+};
+
 struct _ceval_runtime_state {
     int recursion_limit;
     /* Records whether tracing is on for any thread.  Counts the number
@@ -60,6 +64,7 @@ struct _ceval_runtime_state {
     /* Request for checking signals. */
     _Py_atomic_int signals_pending;
     struct _gil_runtime_state gil;
+    struct _opcodecache_config opcodeconfig;
 };
 
 /* interpreter state */
@@ -311,6 +316,10 @@ PyAPI_FUNC(PyStatus) _PyInterpreterState_Enable(_PyRuntimeState *runtime);
 PyAPI_FUNC(void) _PyInterpreterState_DeleteExceptMain(_PyRuntimeState *runtime);
 
 PyAPI_FUNC(void) _PyGILState_Reinit(_PyRuntimeState *runtime);
+
+PyAPI_FUNC(void) _PyEval_SetOpcodeCacheMinRuns(int minruns);
+
+PyAPI_FUNC(int) _PyEval_GetOpcodeCacheMinRuns();
 
 #ifdef __cplusplus
 }
--- a/Include/internal/pycore_pystate.h
+++ b/Include/internal/pycore_pystate.h
@@ -43,10 +43,6 @@ struct _pending_calls {
     int last;
 };
 
-struct _opcodecache_config {
-    int minruns;
-};
-
 struct _ceval_runtime_state {
     int recursion_limit;
     /* Records whether tracing is on for any thread.  Counts the number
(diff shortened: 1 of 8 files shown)


Include/internal/pycore_pystate.h
```diff
@@ -43,6 +43,10 @@ struct _pending_calls {
     int last;
 };
 
+struct _opcodecache_config {
+    int minruns;
+};
```
vstinner: the struct is useless, just put directly the int into _ceval_runtime_state.

pablogsal:
Tested locally:

```
❯ ./python -m test -R 3:3 test_pprint
Run tests sequentially
0:00:00 load avg: 1.57 [1/1] test_pprint
beginning 6 repetitions
123456
......

== Tests result: SUCCESS ==

1 test OK.

Total duration: 751 ms
Tests result: SUCCESS
```

Include/internal/pycore_pystate.h
```diff
@@ -312,6 +317,10 @@ PyAPI_FUNC(void) _PyInterpreterState_DeleteExceptMain(_PyRuntimeState *runtime);
 
 PyAPI_FUNC(void) _PyGILState_Reinit(_PyRuntimeState *runtime);
 
+PyAPI_FUNC(void) _PyEval_SetOpcodeCacheMinRuns(int minruns);
+
+PyAPI_FUNC(int) _PyEval_GetOpcodeCacheMinRuns();
```
vstinner: you should use: (void)

Lib/test/libregrtest/runtest.py
```diff
@@ -206,6 +206,8 @@ def _test_module(the_module):
 def _runtest_inner2(ns, test_name):
     # Load the test function, run the test function, handle huntrleaks
     # and findleaks to detect leaks
+    minruns = sys._getopcacheminruns()
```
vstinner: I don't think that it matters to restore the old value. I'm not sure that sys._getopcacheminruns() is needed.

Python/pylifecycle.c
```diff
@@ -737,6 +737,9 @@ _Py_PreInitializeFromPyArgv(const PyPreConfig *src_config, const _PyArgv *args)
         return status;
     }
 
+    // XXX: Add this properly to the commandline config
```
vstinner: pleave remove this XXX. It doesn't *have to* be a cmdline option.

pablogsal:
Tested also with `./python -m test -R 3:3 -j0 -r -x test_asyncio -x test_multiprocessing_spawn`

Lib/test/libregrtest/runtest.py
```diff
@@ -206,6 +206,8 @@ def _test_module(the_module):
...
+    sys._setopcacheminruns(1)
```
vstinner: That's wrong. Only dash_R() should be modified in libregrtest/refleak.py. You must not disable the opcache in runtest.py.

vstinner (changes requested):

methane:
The code object will be optimized only when `++co->co_opcache_flag == opcacheminruns`.
So decreasing min_runs by `_setopcacheminruns()`will cause some hot codes will be not
optimized forever.  I don't want to expose such switch.

Please use #13787 for 3.8b1.

vstinner:
> Please use #13787 for 3.8b1.

I like #13787, it's a reasonable workaround for beta1. We can design a better fix after the beta.

Lib/test/libregrtest/runtest.py
pablogsal: Sorry, this was supposed to reset the original value. Check out my last fixup (bd3a951ff81eab28b77d05290a2466573c41eda2),

Include/internal/pycore_pystate.h
pablogsal: I made a struct for easy extending more stuff in the config in the future.

I will remove it

Lib/test/libregrtest/runtest.py
vstinner: I think you misunderstood my comment. You must not modify this function but dash_R() function.

pablogsal:
Ok, I will close this PR then.

vstinner:
> Ok, I will close this PR then.

Maybe keep it around, it's sounds like a good fix for post-beta1.

PR #13790, opened 2019-06-03
bpo-36889: Make StreamServer.close() tests more robust

asvetlov:
There are reports about hanging stream tests on buildbots.

This PR rewrites tests for `StreamServer.close()`
It uses more future objects for fine-grained synchronization between client and server parts of tests.


ambv:
@vstinner, look at this.

1st1 (approved):

asvetlov:
Thanks, @1st1

vstinner:
This change fixed https://bugs.python.org/issue37142 if I understood correctly.

asvetlov:
Yes, you are right.
Thanks for the clarification

PR #13793, opened 2019-06-03
Revert "bpo-35621: Support running subprocesses in asyncio when loop is executed in non-main thread (#13630)"

asvetlov:


vstinner (approved):
LGTM.

PR #13760, opened 2019-06-02
bpo-35814: Allow unpacking in r.h.s of annotated assignment expressions

pablogsal:
https://bugs.python.org/issue35814

PR #13794, opened 2019-06-04
bpo-35814: More updates to the annotated assignments docs

ilevkivskyi:
A documentation update for https://github.com/python/cpython/pull/13760

cc @pablogsal @gvanrossum 

https://bugs.python.org/issue35814

### DIFF
--- a/Doc/reference/simple_stmts.rst
+++ b/Doc/reference/simple_stmts.rst
@@ -330,10 +330,9 @@ statement, of a variable or attribute annotation and an optional assignment stat
 
 .. productionlist::
    annotated_assignment_stmt: `augtarget` ":" `expression`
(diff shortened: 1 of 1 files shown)


gvanrossum (approved):

ilevkivskyi:
This is purely a doc fix, so I think this can be merged now.

pablogsal:
Thanks, @ilevkivskyi for finishing the docs! One question, should we add these changes also to "What's new in 3.8?

ilevkivskyi:
I don't know TBH, on one hand it is really minor, on the other hand it is a change to the syntax. Maybe somewhere in the "Other Language Changes" section?

gvanrossum:
These together definitely deserve to be mentioned in What's New and also a versionadded comment in the section describing the syntax.

PR #13795, opened 2019-06-04
Revert "bpo-36818: Add PyInterpreterState.runtime field. (gh-13129)"

vstinner:
This reverts commit 396e0a8d9dc65453cb9d53500d0a620602656cfe.

https://bugs.python.org/issue36818


vstinner:
I tested manually that this PR fix https://bugs.python.org/issue37135

pablogsal (approved):
Checked locally:

396e0a8d9d -> ./python stress.py crashes
396e0a8d9d^ -> ./python stress.py does not crash

PR #13797, opened 2019-06-04
bpo-27513: email.utils.getaddresses() now handles Header objects

ZackerySpytz:
getaddresses() should be able to handle a Header object if passed
one.

https://bugs.python.org/issue27513

### DIFF
--- a/Lib/test/test_email/test_email.py
+++ b/Lib/test/test_email/test_email.py
@@ -3265,9 +3265,8 @@ Foo
 
     def test_getaddresses_header_obj(self):
         """Test the handling of a Header object."""
-        eq = self.assertEqual
         addrs = utils.getaddresses([Header('Al Person <aperson@dom.ain>')])
-        eq(addrs[0][1], 'aperson@dom.ain')
+        self.assertEqual(addrs[0][1], 'aperson@dom.ain')
 
     def test_make_msgid_collisions(self):
         # Test make_msgid uniqueness, even with multiple threads
--- a/Lib/test/test_email/test_email.py
+++ b/Lib/test/test_email/test_email.py
@@ -3263,6 +3263,12 @@ Foo
         addrs = utils.getaddresses(['User ((nested comment)) <foo@bar.com>'])
         eq(addrs[0][1], 'foo@bar.com')
 
+    def test_getaddresses_header_obj(self):
+        """Test the handling of a Header object."""
+        eq = self.assertEqual
+        addrs = utils.getaddresses([Header('Al Person <aperson@dom.ain>')])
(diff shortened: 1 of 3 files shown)


csabella:
cc @maxking

Lib/test/test_email/test_email.py
```diff
@@ -3204,6 +3204,12 @@ def test_getaddresses_embedded_comment(self):
         addrs = utils.getaddresses(['User ((nested comment)) <foo@bar.com>'])
         eq(addrs[0][1], 'foo@bar.com')
 
+    def test_getaddresses_header_obj(self):
+        """Test the handling of a Header object."""
+        eq = self.assertEqual
```
maxking: I don't think we should create an alias if we are going to use this only once :)

Misc/NEWS.d/next/Library/2019-06-03-23-53-25.bpo-27513.qITN7d.rst
```diff
@@ -0,0 +1,2 @@
+:func:`email.charset.getaddresses` can now handle
```
maxking: The import path here is wrong, you meant, `email.utils.getaddress`. Also a stray `~` before the class import path.

Also, I suggest a slight reword:

```
:func:`email.utils.getaddress` now accepts :class:`email.header.Header` objects along with string values. Patch by Zackery Spytz.
```
What do you think?

maxking (changes requested):
Thanks for fixing this, overall it looks good to me. 
Can we also change the docs to reflect that [getaddress](https://docs.python.org/3/library/email.utils.html#email.utils.getaddresses) can accept header values or header objects and perhaps an example in the docs as well. There are other examples for header values in there.

taleinat (changes requested):
All of @maxking's comments are spot on, including adding mention in the docs, except that adding an example of this in the docs would be nice to have, but not necessary. Please make the changes suggested in his comments.

Lib/test/test_email/test_email.py
terryjreedy: Agreed.  I use this alias all the time, but it needs at least 2 and usually 3 uses to pay for itself.

csabella:

TZanke:
After upgrading our customers to Python3 we now have problems with some mails. Please take care of this fix, right know we are stuck manually handling the failing mails.

BoboTiG:
That easy one would be great for 3.10. @ambv do you think this is doable?

BoboTiG:
Wow thanks a lot for your reactivity :muscle: :)

PR #13802, opened 2019-06-04
Revert "bpo-34037, asyncio: add BaseEventLoop.wait_executor_on_close (GH-13786)"

ambv:
This reverts commit 0f0a30f4da4b529e0f7df857b9f575b231b32758.

We need to revert this before b1 due to it being a new documented public API.


ambv:
Also, @1st1 requested this change not to be released as comment on the original PR.

1st1 (approved):

ambv:
There will come a day when I remember to replace # with GH-.

PR #13812, opened 2019-06-04
bpo-36742: Corrects fix to handle decomposition in usernames

zooba:


zooba:
@ambv This is a security fix for a bug that hasn't been released yet - can we merge it before b1 and avoid having to create a new CVE?

PR #13817, opened 2019-06-04
bpo-37144: Convert path-like object to regular path

0x29a:
Steps to reproduce the issue:

1. Create gzipped tar.
```sh
echo "test" > test
tar -cvzf test.tar.gz test
``` 
2. Try to pass path-like object in `tarfile.open`:
```python
import tarfile
import pathlib
path = pathlib.Path('test.tar.gz')
tarfile.open(path, 'w|gz')
```
Result will be:
```
  File "/usr/lib/python3.6/tarfile.py", line 1599, in open
    stream = _Stream(name, filemode, comptype, fileobj, bufsize)
  File "/usr/lib/python3.6/tarfile.py", line 382, in __init__
    self._init_write_gz()
  File "/usr/lib/python3.6/tarfile.py", line 430, in _init_write_gz
    if self.name.endswith(".gz"):
AttributeError: 'PosixPath' object has no attribute 'endswith'
```

https://bugs.python.org/issue37144

### DIFF
--- a/Lib/tarfile.py
+++ b/Lib/tarfile.py
@@ -349,7 +349,7 @@ class _Stream:
             fileobj = _StreamProxy(fileobj)
             comptype = fileobj.getcomptype()
 
-        self.name     = name or ""
+        self.name     = os.path.abspath(name) if name else ""
         self.mode     = mode
         self.comptype = comptype
         self.fileobj  = fileobj
--- a/Lib/tarfile.py
+++ b/Lib/tarfile.py
@@ -1454,6 +1454,7 @@ class TarFile(object):
         self.mode = mode
         self._mode = modes[mode]
 
+        name = os.fspath(name) if name else None
         if not fileobj:
             if self.mode == "a" and not os.path.exists(name):
                 # Create nonexistent files in append mode.
@@ -1469,7 +1470,6 @@ class TarFile(object):
                 self._mode = fileobj.mode
             self._extfileobj = True
 
-        name = os.fspath(name) if name else None
         self.name = os.path.abspath(name) if name else None
         self.fileobj = fileobj
 
--- a/Lib/tarfile.py
+++ b/Lib/tarfile.py
@@ -1468,7 +1468,9 @@ class TarFile(object):
             if hasattr(fileobj, "mode"):
                 self._mode = fileobj.mode
             self._extfileobj = True
-        self.name = os.fspath(name) if name else None
+
+        name = os.fspath(name) if name else None
+        self.name = os.path.abspath(name) if name else None
         self.fileobj = fileobj
 
         # Init attributes.
@@ -1938,12 +1940,13 @@ class TarFile(object):
            excluded from the archive.
         """
         self._check("awx")
+        name = os.fspath(name)
 
         if arcname is None:
             arcname = name
 
         # Skip if somebody tries to archive the archive...
-        if self.name is not None and os.fspath(name) == self.name:
+        if self.name is not None and os.path.abspath(name) == self.name:
             self._dbg(2, "tarfile: Skipped %r" % name)
             return
 
--- a/Lib/tarfile.py
+++ b/Lib/tarfile.py
@@ -349,7 +349,7 @@ class _Stream:
             fileobj = _StreamProxy(fileobj)
             comptype = fileobj.getcomptype()
 
-        self.name     = os.path.abspath(name) if name else ""
+        self.name     = os.fspath(name) if name else ""
         self.mode     = mode
         self.comptype = comptype
         self.fileobj  = fileobj
--- a/Lib/tarfile.py
+++ b/Lib/tarfile.py
@@ -1468,7 +1468,7 @@ class TarFile(object):
             if hasattr(fileobj, "mode"):
                 self._mode = fileobj.mode
             self._extfileobj = True
-        self.name = os.path.abspath(name) if name else None
+        self.name = os.fspath(name) if name else None
         self.fileobj = fileobj
 
         # Init attributes.
@@ -1943,7 +1943,7 @@ class TarFile(object):
             arcname = name
 
         # Skip if somebody tries to archive the archive...
-        if self.name is not None and os.path.abspath(name) == self.name:
+        if self.name is not None and os.fspath(name) == self.name:
             self._dbg(2, "tarfile: Skipped %r" % name)
             return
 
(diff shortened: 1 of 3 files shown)


Lib/tarfile.py
```diff
@@ -349,7 +349,7 @@ def __init__(self, name, mode, comptype, fileobj, bufsize):
             fileobj = _StreamProxy(fileobj)
             comptype = fileobj.getcomptype()
 
-        self.name     = name or ""
+        self.name     = os.path.abspath(name) if name else ""
```
tirkarthi: `os.fspath` looks like a good solution to me. Please add a test in `Lib/test/test_tarfile.py` that opens `gzipname` with mode 'w|gz' flag to make sure there is no `AttributeError`

0x29a: @tirkarthi after opening `gzipname` with mode `w|gz` all remaining tests, that need to extract data from `gzipname` will not pass, because opening rewrites file content.

I added test to `StreamWriteTest` class, where I open for write `tmpname`, because it's designed for writing and insensitive to content change. Also, since this test inside of `StreamWriteTest`, opening by path-like filename will be performed for each mode (gz, lz, bz).

Lib/tarfile.py
```diff
@@ -1468,7 +1468,7 @@ def __init__(self, name=None, mode="r", fileobj=None, format=None,
             if hasattr(fileobj, "mode"):
                 self._mode = fileobj.mode
             self._extfileobj = True
-        self.name = os.path.abspath(name) if name else None
+        self.name = os.fspath(name) if name else None
```
asvetlov: these lines are not equal.
The correct code should be something like:
```
name = os.fspath(name)
self.name = os.path.abspath(name)
```
I've skipped the check for empty/None name.

Lib/tarfile.py
```diff
@@ -1943,7 +1943,7 @@ def add(self, name, arcname=None, recursive=True, *, filter=None):
             arcname = name
 
         # Skip if somebody tries to archive the archive...
-        if self.name is not None and os.path.abspath(name) == self.name:
+        if self.name is not None and os.fspath(name) == self.name:
```
asvetlov: the same

asvetlov (changes requested):

Lib/test/test_tarfile.py
```diff
@@ -1421,6 +1421,11 @@ def test_file_mode(self):
         finally:
             os.umask(original_umask)
 
+    def test_open_by_path_object(self):
+        # Test for issue #37144: broken open for stream write by path-like object
+        tar = tarfile.open(pathlib.Path(tmpname), self.mode)
```
tirkarthi: You can use a context manager that automatically calls close on the object.

```python
with tarfile.open(pathlib.Path(tmpname), self.mode) as f:
	pass
```

zygocephalus:

Lib/test/test_tarfile.py
```diff
@@ -1421,6 +1421,11 @@ def test_file_mode(self):
         finally:
             os.umask(original_umask)
 
+    def test_open_by_path_object(self):
+        # Test for issue #37144: broken open for stream write by path-like object
```
asvetlov: Too long line as well.
Please break it into the two-lines comment.
The comment is valuable, thanks

asvetlov (changes requested):

Lib/tarfile.py
```diff
@@ -1943,7 +1945,7 @@ def add(self, name, arcname=None, recursive=True, *, filter=None):
             arcname = name
 
         # Skip if somebody tries to archive the archive...
-        if self.name is not None and os.path.abspath(name) == self.name:
+        if self.name is not None and os.path.abspath(os.fspath(name)) == self.name:
```
asvetlov: Please move `os.fspath(name)` to the very beginning of the function, just after `self._check(...)`.

zygocephalus:

Lib/tarfile.py
```diff
@@ -1468,6 +1468,8 @@ def __init__(self, name=None, mode="r", fileobj=None, format=None,
             if hasattr(fileobj, "mode"):
                 self._mode = fileobj.mode
             self._extfileobj = True
+
+        name = os.fspath(name) if name else None
```
asvetlov: Move this line before `if not fileobj`: the code calls `os.path.exists()` and `bltn_open()` with `name.`

Both functions accept`pathlib.Path` object but let's be consistent in `Path -> str` conversion, please

asvetlov (changes requested):

Lib/tarfile.py
0x29a: @asvetlov 
Oh, finally I got It. Updated.

asvetlov:
Now looks much better!

After briefly looking in `tarfile.py` I see that `gettarinfo()`, `extractall`, `extract`, `_getmember` should convert their `name` and `path` arguments too.

Lib/test/test_tarfile.py
```diff
@@ -1368,7 +1368,7 @@ def write(self, data):
 
             f = BadFile()
             with self.assertRaises(exctype):
-                tar = tarfile.open(tmpname, self.mode, fileobj=f,
+                tarfile.open(tmpname, self.mode, fileobj=f,
                                    format=tarfile.PAX_FORMAT,
                                    pax_headers={'non': 'empty'})
```
mangrisano, line 1373: These two arguments should be indented under `tmpname`.

Lib/tarfile.py
```diff
@@ -1468,6 +1469,7 @@ def __init__(self, name=None, mode="r", fileobj=None, format=None,
             if hasattr(fileobj, "mode"):
                 self._mode = fileobj.mode
             self._extfileobj = True
+
```
matrixise, line 1472: remove this line, we have to get a minimalist diff for the merge. Thank you

Lib/test/test_tarfile.py
matrixise: you need to keep the same indent.

Lib/test/test_tarfile.py
```diff
@@ -1421,6 +1421,12 @@ def test_file_mode(self):
         finally:
             os.umask(original_umask)
 
+    def test_open_by_path_object(self):
+        # Test for issue #37144:
```
matrixise, line 1425: remove these comments.

matrixise (changes requested):

csabella:
@0x29a, please address the code reviews.  Thank you!

PR #13823, opened 2019-06-04
bpo-34282: Remove deprecated _convert method

zware:
https://bugs.python.org/issue34282


ambv:
The test was added in 0fb9fadd3b3e9e3698647e0b92d49b0b7aacd979, we are following up after the 3.9 branch got created and all buildbots and pull requests are now 🔴 .

ambv (approved):

PR #13832, opened 2019-06-05
bpo-37158: Simplify and speed-up statistics.fmean()

rhettinger:
https://bugs.python.org/issue37158


serhiy-storchaka (approved):
Very clever!

PR #13834, opened 2019-06-05
bpo-34767: Do not always create a collections.deque() in asyncio.Lock()

ZackerySpytz:
https://bugs.python.org/issue34767

### DIFF
--- a/Lib/asyncio/locks.py
+++ b/Lib/asyncio/locks.py
@@ -182,8 +182,8 @@ class Lock(_ContextManagerMixin):
         This method blocks until the lock is unlocked, then sets it to
         locked and returns True.
         """
-        if not self._locked and (self._waiters is None or \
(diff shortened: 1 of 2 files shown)


Lib/asyncio/locks.py
```diff
@@ -182,10 +182,13 @@ def locked(self):
         This method blocks until the lock is unlocked, then sets it to
         locked and returns True.
         """
-        if not self._locked and all(w.cancelled() for w in self._waiters):
+        if not self._locked and (self._waiters is None or \
```
asvetlov: Very minor: please close in brackets all condition `if (test1 and test2)`, drop backslash.

asvetlov (changes requested):
Looks good, please fix the minor comment

asvetlov (approved):

PR #13838, opened 2019-06-05
bpo-26836: Document os.memfd_create() name parameter

vstinner:
https://bugs.python.org/issue26836


vstinner:
Oh, I missed the "name" parameter of memfd_create().

The documentation starts with "Create an anonymous file":
https://docs.python.org/dev/library/os.html#os.memfd_create

Wait, an *anonymous* file... which has a name? IMHO it would help to document the purpose of the name parameter :-) This PR just copy/paste the paragraph from the manual page into Python doc.

vstinner:
cc @ZackerySpytz

tiran (approved):

PR #13842, opened 2019-06-05
gh-72894: Dynamically allocate select fd_sets on Windows based on inputs size

andzn:
https://bugs.python.org/issue28708

* Issue: gh-72894


asvetlov:
By this PR you basically increased the memory consumption for every `select` call from 12 KiB to 384 KiB.
I believe the number is too high.

See the issue's comment added recently

jdemeyer (changes requested):
To test the new `seqsize()` code path, I would like to see a test added where `select.select()` is called with an iterable which is not a `tuple` or `list`, for example `iter([fd])` instead of `[fd]`.

jdemeyer (changes requested):
You are also missing error checks where you call `seqsize()`.

Modules/selectmodule.c
```diff
@@ -58,6 +58,13 @@ extern void bzero(void *, int);
 #  define SOCKET int
 #endif
 
+#if defined(MS_WINDOWS)
+#define FD_SET_WIN(fd, set, setsize) do { \
```
asvetlov: Please use a regular function here instead of macros.
The function may be `static inline` if you really care about speed (personally I don't think that `select` with many sockets can be performant at all).
@vstinner wants to get rid of macros in CPython, let's don't add a new one.

Modules/selectmodule.c
```diff
@@ -94,24 +101,37 @@ typedef struct {
...
 }
 
 
+/* returns the size of a given Python sequence
+   returns -1 if the input is not a sequence
+*/
+static Py_ssize_t
+seqsize(PyObject *seq)
+{
+    PyObject* fast_seq = PySequence_Fast(seq, "argument 1 must be sequence");
```
asvetlov: What is the benefit of using `PySequence_Fast()` here instead of `PySequence_Length()` or even `PyObject_Length()`?
(Despite the fact that you missed `Py_DECREF(fast_seq)`)

Modules/selectmodule.c
```diff
@@ -202,10 +226,6 @@ set2list(fd_set *set, pylist fd2obj[FD_SETSIZE + 1])
     return NULL;
 }
 
-#undef SELECT_USES_HEAP
-#if FD_SETSIZE > 1024
-#define SELECT_USES_HEAP
```
asvetlov: I think `SELECT_USES_HEAP` should be preserved for non-Windows platforms.
What is the reason for removing this mode?

asvetlov (changes requested):

Modules/selectmodule.c
andzn: It's just my first PR to python - so I'm not really familiar with the code-base. I just did what I saw in other functions in the same file. But I'll update it - PySequence_Length/PyObject_Length look promising.

Modules/selectmodule.c
andzn: I started from vstinner's prototype: #13845 which removed all the places where it was used. Since it was not used anymore, I removed the definition of it too. But I can add it back if needed.

andzn:

andzn:
@jdmeyer I removed the seqsize function completely and replaced calls with PySequence_Length.

NewUserHa:
hope this pass soon.
select() limit really is pain ass.

WindowsProactorEventLoopPolicy seems still has this limit

csabella:
@andzn, please resolve the merge conflict.  Thank you!

andzn:
@csabella , merge conflicts are now resolved.

andzn:
@asvetlov, @jdemeyer  can you please take a look again at the PR? The requested changes have already been implemented a year ago.

andzn:
> @andzn Could you sign the new CLA by clicking `not signed` button in the cpython-cla-bot's message, please?

@arhadthedev done.

Modules/selectmodule.c
```diff
@@ -63,6 +63,17 @@ extern void bzero(void *, int);
 #  define SOCKET int
 #endif
 
+#if defined(MS_WINDOWS)
+
+void fd_set_win(SOCKET fd, fd_set *set, size_t setsize) {
+    do {
+        if (((fd_set FAR *)(set))->fd_count < setsize)
+            ((fd_set FAR *)(set))->fd_array[((fd_set FAR *)(set))->fd_count++]=(fd);
+    } while(0);
```
vstinner, line 72: What is the purpose of the surrounding code ``do { ... } while (0)``?

Modules/selectmodule.c
```diff
@@ -106,24 +118,23 @@ typedef struct {
 } pylist;
 
 static void
-reap_obj(pylist fd2obj[FD_SETSIZE + 1])
+reap_obj(pylist *fd2obj, size_t setsize)
 {
-    unsigned int i;
-    for (i = 0; i < (unsigned int)FD_SETSIZE + 1 && fd2obj[i].sentinel >= 0; i++) {
+    size_t i;
+    for (i = 0; i < setsize + 1 && fd2obj[i].sentinel >= 0; i++) {
```
vstinner, line 124:
```suggestion
    for (size_t i = 0; i < setsize + 1 && fd2obj[i].sentinel >= 0; i++) {
```

Modules/selectmodule.c
```diff
@@ -196,7 +211,7 @@ set2list(fd_set *set, pylist fd2obj[FD_SETSIZE + 1])
         return NULL;
 
     i = 0;
```
vstinner, line 213: Can you declare i here?

Modules/selectmodule.c
```diff
@@ -257,26 +272,43 @@ select_select_impl(PyObject *module, PyObject *rlist, PyObject *wlist,
...
+#endif /* !defined(SELECT_USES_HEAP) */
+#else
+    Py_ssize_t rsize, wsize, xsize;
+#endif /* !defined(MS_WINDOWS) */
     struct timeval tv, *tvp;
     int imax, omax, emax, max;
     int n;
     _PyTime_t timeout, deadline = 0;
 
+    setsize = FD_SETSIZE;
```
vstinner, line 294: I would prefer to initialize the variable where it's declared. Either declare it at this line, or move the declaration here. So the variable is never uninitialized.

Modules/selectmodule.c
```diff
@@ -257,26 +272,43 @@ select_select_impl(PyObject *module, PyObject *rlist, PyObject *wlist,
...
+    rsize = PySequence_Length(rlist);
+    wsize = PySequence_Length(wlist);
+    xsize = PySequence_Length(xlist);
+
+    if (rsize < 0 || wsize < 0 || xsize < 0) {
+        PyErr_SetString(PyExc_ValueError, "arguments 1-3 must be sequences");
+        return NULL;
+    }
+
+    if ((size_t)rsize > setsize) setsize = (size_t)rsize;
```
vstinner, line 307: You may write: ``setsize = Py_MAX(setsize, (size_t)rsize);`` (same for the 2 lines below). IMO it's more explicit (and how I would write it in Python).

Modules/selectmodule.c
```diff
@@ -298,30 +330,55 @@ select_select_impl(PyObject *module, PyObject *rlist, PyObject *wlist,
...
+    /* Allocate memory for the sets */
+    /* We ought to allocate setsize * sizeof(SOCKET) +
+    sizeof(int) bytes. With this approach on 64-bit we
+    allocate 4 additional unused bytes but the allocation
+    code looks a lot simpler.*/
+    ifdset = (fd_set*) PyMem_NEW(SOCKET, setsize + 1);
+    ofdset = (fd_set*) PyMem_NEW(SOCKET, setsize + 1);
+    efdset = (fd_set*) PyMem_NEW(SOCKET, setsize + 1);
+    if (ifdset->fd_array == NULL || ofdset->fd_array == NULL || efdset->fd_array == NULL) {
+        if (ifdset->fd_array) PyMem_DEL(ifdset->fd_array);
```
vstinner, line 360: PyMem_DEL() is a alias to PyMem_Free(): use it directly.

The NULL test is useless and can be removed. PyMem_Free(NULL) is well defined, it does nothing: https://docs.python.org/dev/c-api/memory.html#c.PyMem_Free

Modules/selectmodule.c
```diff
@@ -257,26 +272,43 @@ select_select_impl(PyObject *module, PyObject *rlist, PyObject *wlist,
...
-    pylist rfd2obj[FD_SETSIZE + 1];
-    pylist wfd2obj[FD_SETSIZE + 1];
-    pylist efd2obj[FD_SETSIZE + 1];
-#endif /* SELECT_USES_HEAP */
     PyObject *ret = NULL;
-    fd_set ifdset, ofdset, efdset;
+    fd_set *ifdset, *ofdset, *efdset;
+#if !defined(MS_WINDOWS)
+    fd_set ifdset_stack, ofdset_stack, efdset_stack;
+#if !defined(SELECT_USES_HEAP)
```
vstinner, line 281: It would make the code simpler if the SELECT_USES_HEAP macro would always be defined on Windows, no?

PR #13844, opened 2019-06-05
bpo-36974: expand call protocol documentation

jdemeyer:
CC @encukou 

I'm also adding Petr Viktorin as contributor for vectorcall in the "what's new" section.


jdemeyer:
Note that the first commit just moves the existing documentation, the second commit actually makes the changes.

jdemeyer:
@vstinner: I moved your doc fixes from #13890 here.

jdemeyer:
Rebased to latest master, squashed most commits. The commit simply moving the documentation and the one coming from @vstinner are still separate.

Can this be reviewed please? It's a documentation-only patch, so it shouldn't be controversial. There are other PRs open for review or planned that will affect the documentation.

Please also add the labels `skip news` and `needs backport to 3.8`.

CC @methane

jdemeyer:
I give up

jdemeyer:
I'm rebooting this PR since I don't expect any further changes to the call API. As before, there are two commits: one to create `Doc/c-api/call.rst` and one making actual modifications.

Doc/c-api/call.rst
```diff
@@ -0,0 +1,344 @@
...
+A call is made using a tuple for the positional arguments
+and a dict for the keyword arguments, just like ``f(*args, **kwargs)``
+in the Python language.
+*args* must be non-NULL
+(use an empty tuple if there are no arguments)
+but *kwargs* may be *NULL* if there are no keyword arguments.
+
+This convention is not only used by *tp_call*:
+also :c:member:`~PyTypeObject.tp_new` and :c:member:`~PyTypeObject.tp_init`
+pass arguments this way
```
epicfaace:
```suggestion
also pass arguments this way
```

Doc/c-api/call.rst
```diff
@@ -0,0 +1,344 @@
...
+
+A call is made using a tuple for the positional arguments
+and a dict for the keyword arguments, just like ``f(*args, **kwargs)``
+in the Python language.
+*args* must be non-NULL
+(use an empty tuple if there are no arguments)
+but *kwargs* may be *NULL* if there are no keyword arguments.
+
+This convention is not only used by *tp_call*:
+also :c:member:`~PyTypeObject.tp_new` and :c:member:`~PyTypeObject.tp_init`
```
epicfaace:
```suggestion
:c:member:`~PyTypeObject.tp_new` and :c:member:`~PyTypeObject.tp_init`
```

Doc/c-api/call.rst
```diff
@@ -0,0 +1,344 @@
...
+
+The *tp_call* Protocol
+----------------------
+
+Classes can implement callables by setting :c:member:`~PyTypeObject.tp_call`.
+
+A call is made using a tuple for the positional arguments
+and a dict for the keyword arguments, just like ``f(*args, **kwargs)``
+in the Python language.
+*args* must be non-NULL
```
epicfaace: Why are these lines so short?

Doc/c-api/call.rst
```diff
@@ -0,0 +1,344 @@
...
+   - *callable* is the object being called.
+   - *args* is a C array consisting of the positional arguments followed by the
+     values of the keyword arguments.
+     This can be *NULL* if there are no arguments.
+   - *nargsf* is the number of positional arguments plus possibly the
+     :const:`PY_VECTORCALL_ARGUMENTS_OFFSET` flag.
+     To get the actual number of positional arguments from *nargsf*,
+     use :c:func:`PyVectorcall_NARGS`.
+   - *kwnames* is a tuple containing the names of the keyword arguments,
+     in other words the keys of the kwargs dict.
```
epicfaace:
```suggestion
     in other words, the keys of the kwargs dict.
```

Doc/c-api/call.rst
```diff
@@ -0,0 +1,344 @@
...
+
+   - *callable* is the object being called.
+   - *args* is a C array consisting of the positional arguments followed by the
+     values of the keyword arguments.
+     This can be *NULL* if there are no arguments.
+   - *nargsf* is the number of positional arguments plus possibly the
+     :const:`PY_VECTORCALL_ARGUMENTS_OFFSET` flag.
+     To get the actual number of positional arguments from *nargsf*,
+     use :c:func:`PyVectorcall_NARGS`.
+   - *kwnames* is a tuple containing the names of the keyword arguments,
```
epicfaace:
```suggestion
   - *kwnames* is a tuple containing the names of the keyword arguments;
```

Doc/c-api/call.rst
```diff
@@ -0,0 +1,344 @@
...
+
+   Whenever they can do so cheaply (without additional allocation), callers
+   are encouraged to use :const:`PY_VECTORCALL_ARGUMENTS_OFFSET`.
+   Doing so will allow callables such as bound methods to make their onward
+   calls (which include a prepended *self* argument) very efficiently.
+
+.. warning::
+
+   As rule of thumb, CPython will internally use the vectorcall
+   protocol if the callable supports it. However, this is not a hard rule,
```
epicfaace:
```suggestion
   protocol if the callable supports it. However, this is not a hard rule;
```

Doc/c-api/call.rst
```diff
@@ -0,0 +1,344 @@
...
+   *tp_call* may be used instead.
+   Therefore, a class supporting vectorcall must also set
```
epicfaace: Maybe "implement" instead of "set"?
```suggestion
   Therefore, a class supporting vectorcall must also implement
```

Doc/c-api/call.rst
```diff
@@ -0,0 +1,344 @@
...
+
+   .. versionadded:: 3.9
+
+
+.. c:function:: PyObject* PyObject_Call(PyObject *callable, PyObject *args, PyObject *kwargs)
+
+   Call a callable Python object *callable*, with arguments given by the
+   tuple *args*, and named arguments given by the dictionary *kwargs*.
+
+   *args* must not be *NULL*, use an empty tuple if no arguments are needed.
```
epicfaace:
```suggestion
   *args* must not be *NULL*; use an empty tuple if no arguments are needed.
```

Doc/c-api/call.rst
```diff
@@ -0,0 +1,344 @@
...
+
+   .. versionchanged:: 3.4
+      The type of *format* was changed from ``char *``.
+
+
+.. c:function:: PyObject* PyObject_CallMethod(PyObject *obj, const char *name, const char *format, ...)
+
+   Call the method named *name* of object *obj* with a variable number of C
+   arguments.  The C arguments are described by a :c:func:`Py_BuildValue` format
+   string that should  produce a tuple.
```
epicfaace:
```suggestion
   string that should produce a tuple.
```

Doc/c-api/call.rst
```diff
@@ -0,0 +1,344 @@
...
+   :c:func:`PyObject_CallMethodObjArgs` is a faster alternative.
+
+   .. versionchanged:: 3.4
+      The types of *name* and *format* were changed from ``char *``.
+
+
+.. c:function:: PyObject* PyObject_CallFunctionObjArgs(PyObject *callable, ..., NULL)
+
+   Call a callable Python object *callable*, with a variable number of
+   :c:type:`PyObject\*` arguments.  The arguments are provided as a variable number
```
epicfaace:
```suggestion
   :c:type:`PyObject \*` arguments.  The arguments are provided as a variable number
```

Doc/c-api/call.rst
```diff
@@ -0,0 +1,344 @@
...
+   Return the result of the call on success, or raise an exception and return
+   *NULL* on failure.
+
+   This is the equivalent of the Python expression:
+   ``callable(arg1, arg2, ...)``.
+
+
+.. c:function:: PyObject* PyObject_CallMethodObjArgs(PyObject *obj, PyObject *name, ..., NULL)
+
+   Calls a method of the Python object *obj*, where the name of the method is given as a
```
epicfaace:
```suggestion
   Call a method of the Python object *obj*, where the name of the method is given as a
```

Doc/c-api/call.rst
```diff
@@ -0,0 +1,344 @@
...
+   Python string object in *name*.  It is called with a variable number of
+   :c:type:`PyObject\*` arguments.  The arguments are provided as a variable number
```
epicfaace:
```suggestion
   :c:type:`PyObject \*` arguments.  The arguments are provided as a variable number
```

Doc/c-api/call.rst
```diff
@@ -0,0 +1,344 @@
...
+
+      This function is provisional and expected to become public in Python 3.9,
+      with a different name and, possibly, changed semantics.
+      If you use the function, plan for updating your code for Python 3.9.
+
+   .. versionadded:: 3.8
+
+.. c:function:: PyObject* _PyObject_FastCallDict(PyObject *callable, PyObject *const *args, size_t nargsf, PyObject *kwdict)
+
+   Call *callable* with positional arguments passed exactly as in the vectorcall_ protocol
```
epicfaace:
```suggestion
   Call *callable* with positional arguments passed exactly as in the vectorcall protocol,
```

Doc/c-api/call.rst
```diff
@@ -0,0 +1,344 @@
...
+   for future extensions.
+
+   .. versionadded:: 3.8
+
+.. c:function:: vectorcallfunc _PyVectorcall_Function(PyObject *op)
+
+   If *op* does not support the vectorcall protocol (either because the type
+   does not or because the specific instance does not), return *NULL*.
+   Otherwise, return the vectorcall function pointer stored in *op*.
+   This function never sets an exception.
```
epicfaace:
```suggestion
   This function never raises an exception.
```

Doc/c-api/call.rst
```diff
@@ -0,0 +1,344 @@
...
+   :c:member:`~PyTypeObject.tp_call` slot or be used in an implementation of ``tp_call``.
+   It does not check the :const:`_Py_TPFLAGS_HAVE_VECTORCALL` flag
+   and it does not fall back to ``tp_call``.
+
+   .. versionadded:: 3.8
+
+.. c:function:: PyObject* _PyObject_VectorcallMethod(PyObject *name, PyObject *const *args, size_t nargsf, PyObject *kwnames)
+
+   Call a method using the vectorcall calling convention. The name of the method
+   is given as Python string *name*. The object whose method is called is
```
epicfaace:
```suggestion
   is given as the Python string *name*. The object whose method is called is
```

Doc/c-api/call.rst
```diff
@@ -0,0 +1,344 @@
...
+   *args[0]* and the *args* array starting at *args[1]* represents the arguments
```
epicfaace:
```suggestion
   *args[0]*, and the *args* array starting at *args[1]* represents the arguments
```

Doc/c-api/exceptions.rst
```diff
@@ -695,13 +695,17 @@ The following functions are used to create and modify Unicode exceptions from C.
...
+.. _recursion:
+
 Recursion Control
 =================
 
 These two functions provide a way to perform safe recursive calls at the C
 level, both in the core and in extension modules.  They are needed if the
 recursive code does not necessarily invoke Python code (which tracks its
 recursion depth automatically).
+It's also not needed for *tp_call* implementations
```
epicfaace:
```suggestion
They're also not needed for *tp_call* implementations
```

encukou:
I hope to look at this at this week's sprint. To make it faster, I plan to push changes directly to your branch, @jdemeyer. Let me know if you want more back-and-forth.

jdemeyer:
> To make it faster, I plan to push changes directly to your branch, @jdemeyer.

Sure. But please give me a final look before merging :-)

Doc/c-api/call.rst
jdemeyer: It doesn't matter in the built documentation. I personally prefer breaking such lines logically, instead of at an arbitrary number of characters. So I tend to use a newline at the end of a sentence. If this is considered bad style, I can fix it.

Doc/c-api/call.rst
jdemeyer: What's wrong with the reference `vectorcall_`?

encukou:
Here are my changes. Sorry for the big diff; I can't do the rewordings and reorganization separately and I don't think it's practical to split afterwards.

encukou:
@jdemeyer, are you OK with this?

jdemeyer:
I don't mind adding
```
.. warning::

   The vectorcall API is provisional and expected to become public in
   Python 3.9, with a different name and, possibly, changed semantics.
   If you use the function, plan for updating your code for Python 3.9.
```
to the general section about vectorcall, but this text refers to a specific function, so it should be reworded.

jdemeyer:
I really like the table :+1: but I have a few comments:
1. `PyObject_CallOneArg` -> `_PyObject_CallOneArg` (missing underscore)
1. `PyObject_CallFunction` and `_PyObject_CallMethodNoArgs` are missing
1. I would write "vectorcall" under "args" in the `_PyObject_FastCallDict()` line because the handling of "args" really is the same as `_PyObject_Vectorcall`.
1. For methods, you always write `char*` but in some cases it's really a Python object with the name instead of a `char*`.

Doc/c-api/call.rst
epicfaace: I don't see `vectorcall_` anywhere else in the code or doc -- is the underscore some kind of rst syntax? I guess this is moot anyway since this part was removed.

jdemeyer:
Other than that, looks good!

encukou (approved):

vstinner:
Good idea to put all call documentation at the same place ;-)

PR #13845, opened 2019-06-05
[WIP] bpo-28708: Add setsize keywoard-only parameter to select.select()

vstinner:
It becomes possible to configure the set size used internally by
select.select() at runtime.

https://bugs.python.org/issue28708

### DIFF
--- a/Modules/selectmodule.c
+++ b/Modules/selectmodule.c
@@ -94,10 +94,10 @@ typedef struct {
 } pylist;
 
 static void
-reap_obj(pylist fd2obj[FD_SETSIZE + 1])
+reap_obj(pylist *fd2obj, size_t setsize)
 {
-    unsigned int i;
-    for (i = 0; i < (unsigned int)FD_SETSIZE + 1 && fd2obj[i].sentinel >= 0; i++) {
+    size_t i;
+    for (i = 0; i < setsize + 1 && fd2obj[i].sentinel >= 0; i++) {
         Py_CLEAR(fd2obj[i].obj);
     }
     fd2obj[0].sentinel = -1;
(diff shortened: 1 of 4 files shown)


vstinner:
I have no idea if this PR works :-) Is it really possible to change FD_SETSIZE at runtime?

Modules/selectmodule.c
```diff
@@ -242,29 +251,26 @@ descriptors can be used.
...
-     * - add one
-     * - that's the size
-     * See: Stevens, APitUE, $12.5.1
-     */
-    pylist rfd2obj[FD_SETSIZE + 1];
-    pylist wfd2obj[FD_SETSIZE + 1];
-    pylist efd2obj[FD_SETSIZE + 1];
-#endif /* SELECT_USES_HEAP */
     PyObject *ret = NULL;
     fd_set ifdset, ofdset, efdset;
```
andzn, line 260: I think these three sets will need to be allocated differently. We'll probably need a different fd_set structure something like: 
```
typedef struct fd_set_2
{
    u_int fd_count
    SOCKET *fd_array;
} fd_set_2;
```
and then dynamically allocate the fd_array to be a SOCKET* array of size set_size.
Knowing that the select implementation only looks at the fd_count for the size of the array (and doesn't do any boundary checking), we can cast those sets to fd_set and then pass them to the select call.

andzn: Actually that struct type isn't going to work, because that adds a level of indirection that the select implementation doesn't expect.
We'll need to allocate a continuous area of memory of size (sizeof(u_int) + setsize*sizeof(SOCKET))

vstinner:
It seems like @andzn went further than me in PR #13842, I close my WIP PR.

PR #13847, opened 2019-06-05
bpo-18299: Improving eintrdata's test suites

shihai1991:
https://bugs.python.org/issue18299

### DIFF
--- a/Lib/test/eintrdata/eintr_tester.py
+++ b/Lib/test/eintrdata/eintr_tester.py
@@ -16,12 +16,12 @@ import platform
 import select
 import signal
 import socket
-import subprocess
 import sys
 import time
 import unittest
 
 from test import support
+from test.support.script_helper import spawn_python
 
 @contextlib.contextmanager
 def kill_on_error(proc):
@@ -71,10 +71,6 @@ class EINTRBaseTest(unittest.TestCase):
         if hasattr(faulthandler, 'cancel_dump_traceback_later'):
             faulthandler.cancel_dump_traceback_later()
 
-    def subprocess(self, *args, **kw):
-        cmd_args = (sys.executable, '-c') + args
-        return subprocess.Popen(cmd_args, **kw)
-
 
 @unittest.skipUnless(hasattr(signal, "setitimer"), "requires setitimer()")
 class OSEINTRTest(EINTRBaseTest):
@@ -82,7 +78,7 @@ class OSEINTRTest(EINTRBaseTest):
 
     def new_sleep_process(self):
         code = 'import time; time.sleep(%r)' % self.sleep_time
-        return self.subprocess(code)
+        return spawn_python('-c', code)
 
     def _test_wait_multiple(self, wait_func):
         num = 3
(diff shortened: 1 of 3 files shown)


Lib/test/support/script_helper.py
```diff
@@ -122,11 +122,14 @@ def run_python_until_end(*args, **env_vars):
...
         env['TERM'] = ''
 
+    universal_newlines = env_vars.pop('__universal_newlines', False)
+
     env.update(env_vars)
     cmd_line.extend(args)
     proc = subprocess.Popen(cmd_line, stdin=subprocess.PIPE,
                          stdout=subprocess.PIPE, stderr=subprocess.PIPE,
-                         env=env, cwd=cwd)
+                         env=env, cwd=cwd,
```
mangrisano: I think this line and the over one should be indented under `cmd_line`.

Lib/test/support/script_helper.py
```diff
@@ -122,11 +122,14 @@ def run_python_until_end(*args, **env_vars):
...
+                         universal_newlines=universal_newlines)
```
mangrisano: Same thing as over.

Lib/test/support/script_helper.py
shihai1991: done:)

Lib/test/support/script_helper.py
shihai1991: done:)

mangrisano:
Did you test this change? If not, please provide it.

mangrisano:
Furthermore, you also need to fix the doc adding the parameter in the signature.

shihai1991:
> Did you test this change? If not, please provide it.
Let us waiting other developer's opinion, if they think this patch is fine, i will replenish it;)

shihai1991:
@vstinner Hi, Victor. Looks i choose the complicated way of script_helper. Maybe i could find the way which you said to supply popen more easily.

gpshead:
For a new feature in `test.support.script_helper` it would help to have some motivating examples refactoring some existing test suite to cleaner due to the new API.

shihai1991:
@gpshead Thanks for review. According your and victor's opinion, I refactor the eintr_tester's test suites.
And pls forget my previous patch;)

Lib/test/eintrdata/eintr_tester.py
```diff
@@ -71,18 +71,14 @@ def tearDown(self):
...
-
 
 @unittest.skipUnless(hasattr(signal, "setitimer"), "requires setitimer()")
 class OSEINTRTest(EINTRBaseTest):
     """ EINTR tests for the os module. """
 
     def new_sleep_process(self):
         code = 'import time; time.sleep(%r)' % self.sleep_time
-        return self.subprocess(code)
+        return spawn_python('-c', code)
```
gpshead, line 81: I don't see how this is beneficial.  The point of that method is to avoid having to type '-c' everywhere.

gpshead:
This doesn't materially improve the eintrdata test suite and the issue it links to is about script_helper, _assert_python, and text mode to the child.  which don't seem to be a problem in eintrdata_tester.py

Lib/test/eintrdata/eintr_tester.py
shihai1991: > This doesn't materially improve the eintrdata test suite and the issue it links to is about script_helper, _assert_python, and text mode to the child. which don't seem to be a problem in eintrdata_tester.py

ok, the spawn_python in script_helper is good enough. Forget this.

PR #13850, opened 2019-06-05
bpo-37166: inspect.findsource() no longer triggers IndexError...

mibpl:
...when co_firstlineno is larger than the size of the linecache entry

https://bugs.python.org/issue37166

### DIFF
--- a/Lib/inspect.py
+++ b/Lib/inspect.py
@@ -861,7 +861,10 @@ def findsource(object):
     if iscode(object):
         if not hasattr(object, 'co_firstlineno'):
             raise OSError('could not find function definition')
-        lnum = object.co_firstlineno - 1
+        lnum = min(
+            object.co_firstlineno - 1,
+            len(lines) - 1
+        )
         pat = re.compile(r'^(\s*def\s)|(\s*async\s+def\s)|(.*(?<!\w)lambda(:|\s))|^(\s*@)')
         while lnum > 0:
             if pat.match(lines[lnum]): break
--- a/Lib/inspect.py
+++ b/Lib/inspect.py
@@ -861,10 +861,7 @@ def findsource(object):
     if iscode(object):
         if not hasattr(object, 'co_firstlineno'):
             raise OSError('could not find function definition')
-        lnum = min(
-            object.co_firstlineno - 1,
-            len(lines) - 1
-        )
+        lnum = min(object.co_firstlineno - 1, len(lines) - 1)
         pat = re.compile(r'^(\s*def\s)|(\s*async\s+def\s)|(.*(?<!\w)lambda(:|\s))|^(\s*@)')
         while lnum > 0:
             if pat.match(lines[lnum]): break
--- a/Lib/test/test_inspect.py
+++ b/Lib/test/test_inspect.py
@@ -689,6 +689,28 @@ class TestBuggyCases(GetSourceBase):
         finally:
             del linecache.cache[co.co_filename]
 
+    def test_findsource_code_too_short(self):
+        lines = [
+            "",
+            "",
+            "",
+            "def foo(): pass",
+        ]
+        co = compile("\n".join(lines), "f.py", "exec")
+        linecache.cache[co.co_filename] = (
+            1, None, lines, co.co_filename
+        )
+        try:
+            _, lnum = inspect.findsource(co)
+            self.assertEqual(lnum, 3)
+            linecache.cache[co.co_filename] = (
+                1, None, [""] * lnum, co.co_filename
+            )
+            _, lnum = inspect.findsource(co)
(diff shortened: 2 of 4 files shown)


Lib/inspect.py
```diff
@@ -834,7 +834,10 @@ def findsource(object):
     if iscode(object):
         if not hasattr(object, 'co_firstlineno'):
             raise OSError('could not find function definition')
-        lnum = object.co_firstlineno - 1
+        lnum = min(
+            object.co_firstlineno - 1,
+            len(lines) - 1
+        )
```
mangrisano: Is there a reason for putting on different rows the args?

mangrisano (approved):

Misc/NEWS.d/next/Library/2019-06-05-14-47-06.bpo-37166.cDr9zg.rst
```diff
@@ -0,0 +1,2 @@
+inspect.findsource() no longer triggers IndexError when co_firstlineno is
+larger than the size of the linecache entry
```
vstinner: Would you mind how that can happen? I understand that the issue can happen if the source file is modified after the code is loaded in Python, right?

Lib/test/test_inspect.py
```diff
@@ -674,6 +674,28 @@ def test_findsource_code_in_linecache(self):
...
+        try:
+            _, lnum = inspect.findsource(co)
+            self.assertEqual(lnum, 3)
+            linecache.cache[co.co_filename] = (
+                1, None, [""] * lnum, co.co_filename
+            )
+            _, lnum = inspect.findsource(co)
+            self.assertEqual(lnum, 0)  # maybe OSError?
+        finally:
+            del linecache.cache[co.co_filename]
```
vstinner, line 712: I dislike this test because it makes test_inspect too dependent of the internal linecache cache format. Honestly, I'm not sure that it's worth it to have a test. I would be fine to only test the change manually.

**or** please rewrite the test without injecting data into linecache.cache, but create a file on disk instead. Basically, write a program which reproduce what you did manually.

vstinner (changes requested):

Lib/test/test_inspect.py
mibpl: Happy to remove the test and test manually instead. Please note that we already have this dependency elsewhere in test_inspect (see: test_findsource_code_in_linecache for instance).

Misc/NEWS.d/next/Library/2019-06-05-14-47-06.bpo-37166.cDr9zg.rst
mibpl: https://bugs.python.org/issue37166 describes the broad issue.

In our case, we compile the application into an executable package (.par). It may not contain the sources. When the executable is ran, inspect may pick up a source file from the current working directory instead.

This is a common occurrence, as developers tend to run the executable from the root of the repository, and the filenames will always align - and the skew in file lengths between the executable and the working directory is likely.

The following would reproduce the issue closer to the real-life scenario:

_a.py:_
```python
import inspect








def foo(): pass

inspect.getsource(foo)
```

```bash
python3 -m compileall -b a.py
black a.py  # to remove the empty lines
python3 a.pyc  # Index error
```

In the real-life scenario the equivalent of a.pyc could be an executable built in an arbitrary point in time, and distributed into the development environment.

vstinner: I suggest:

:func:`inspect.findsource` no longer triggers :exc:`IndexError` when ``co_firstlineno`` is
larger than the size of the :mod:`linecache` entry. It can happen when the source file
is modified after the code object creation.

mibpl:

mibpl:
@vstinner: would you mind taking a look?

vstinner:
> @vstinner: would you mind taking a look?

I have a question: https://bugs.python.org/issue37166#msg345922

csabella:
I believe @mibpl answered the questions on the bug tracker.

vstinner:
Sorry, I'm unavailable to review this PR.

lisroach:
@1st1 would you have time to take a look? I'm afraid I'm not familiar enough with this code.

1st1:
@mibpl Do you think we can add a test for this?

mibpl:
@1st1 Sure, would this one be sufficient? https://github.com/python/cpython/pull/13850/commits/683a2d6decff8e66cf06a4669af44be2224f8096

See this comment on why it was removed, I'm happy to bring it back: https://github.com/python/cpython/pull/13850#discussion_r293994387

iritkatriel:
Closing the PR following the closure of its b.p.o issue (https://bugs.python.org/issue37166)

PR #13851, opened 2019-06-05
bpo-37134: Use PEP570 syntax for sum()

pablogsal:
https://bugs.python.org/issue37134

### DIFF
--- a/Doc/library/functions.rst
+++ b/Doc/library/functions.rst
@@ -1562,7 +1562,7 @@ are always available.  They are listed here in alphabetical order.
    about strings, see :ref:`textseq`.
 
 
-.. function:: sum(iterable[, start])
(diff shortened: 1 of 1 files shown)


Doc/library/functions.rst
```diff
@@ -1562,7 +1562,7 @@ are always available.  They are listed here in alphabetical order.
    about strings, see :ref:`textseq`.
 
 
-.. function:: sum(iterable[, start])
+.. function:: sum(iterable, /[, start])
```
gvanrossum: Maybe `start=0` ?

pablogsal: Done in e035eee

I have eliminated the text "*start* defaults to ``0``" as is now redundant.

gvanrossum (approved):

PR #13861, opened 2019-06-06
bpo-37173: Show passed class in inspect.getfile error

flying-sheep:
Currently, inspect.getfile(str) will report nonsense:

```pytb
>>> inspect.getfile(str)
TypeError: <module 'builtins' (built-in)> is a built-in class
```

This fixes that

https://bugs.python.org/issue37173

### DIFF
--- a/Lib/inspect.py
+++ b/Lib/inspect.py
@@ -659,9 +659,9 @@ def getfile(object):
         raise TypeError('{!r} is a built-in module'.format(object))
     if isclass(object):
         if hasattr(object, '__module__'):
-            object = sys.modules.get(object.__module__)
-            if getattr(object, '__file__', None):
-                return object.__file__
+            module = sys.modules.get(object.__module__)
+            if getattr(module, '__file__', None):
+                return module.__file__
         raise TypeError('{!r} is a built-in class'.format(object))
     if ismethod(object):
         object = object.__func__
--- a/Lib/test/test_inspect.py
+++ b/Lib/test/test_inspect.py
@@ -513,6 +513,24 @@ class TestRetrievingSourceCode(GetSourceBase):
     def test_getfile(self):
         self.assertEqual(inspect.getfile(mod.StupidGit), mod.__file__)
 
+    def test_getfile_builtin_module(self):
+        with self.assertRaises(TypeError) as e:
+            inspect.getfile(sys)
+        self.assertTrue(str(e.exception).startswith('<module'))
+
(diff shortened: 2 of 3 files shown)


mangrisano (approved):

flying-sheep:
I think we can skip the news entry on this tiny fix

asvetlov (changes requested):
I don't follow how the variable renaming can fix anything.
Please provide a test

flying-sheep:
May I convince you to look again instead? It’s 4 lines of very simple code…

1. We go into the branch when `object` is a class
2. `object` gets reassigned to the class’ module
3. An error gets thrown, with the message talking about the class.

 I simply introduce a new variable instead of reassigning `object`, so it stays assigned to the class and the error message is correct.

terryjreedy:
I agree that the change appears to be is correct, as explained.  See the end below.

In general, bpo issues get blurbs.  A bugfix like this needs a blurb so that anyone who has run into it will see that it is fixed.  We are not starved for news space. "The exception message for inspect.getfile(<built-in class>) now correctly reports the class rather than module builtins.

test_inspect needs 3 new methods added after test_getfile:
```
    def test_getfile_builtin_module(self):
        with self.assertRaises(TypeError) as e:
            inspect.getfile(sys)
        self.assertTrue(str(e.exception).startswith('<module'))

    def test_getfile_builtin_class(self):
        with self.assertRaises(TypeError) as e:
            inspect.getfile(int)
        self.assertTrue(str(e.exception).startswith('<class'))

    def test_getfile_builtin_function_or_method(self):
        with self.assertRaises(TypeError) as e:
            inspect.getfile(abs)
        self.assertIn('expected, got', str(e.exception))
        with self.assertRaises(TypeError) as e:
            inspect.getfile(list.append)
        self.assertIn('expected, got', str(e.exception))
```
Note that test_getfile_builtin_class current fails because the message currently starts with '<module'.  The patch should make it pass.

flying-sheep:
OK, I was opposed to just add a test for the thing I fixed (basically as a regression test), but proposing that the whole thing ought to be tested makes sense to me. I’ll add both news and tests

flying-sheep:
@asvetlov OK, done

asvetlov (approved):
lgtm

flying-sheep:
Great! Thank you for the tests @terryjreedy

terryjreedy:
@flying-sheep Our general policy is to add an initially failing test, to prevent a regression, for every bugfix.

PR #13865, opened 2019-06-06
bpo-36922: use Py_TPFLAGS_METHOD_DESCRIPTOR in lookup_maybe_method()

jdemeyer:
`lookup_maybe_method` is an obvious place where the use case of `Py_TPFLAGS_METHOD_DESCRIPTOR` applies.

IMHO, this is safe enough for 3.8 but the reviewer should judge that.

CC @encukou  @vstinner 

https://bugs.python.org/issue36922

### DIFF
--- a/Lib/test/test_descr.py
+++ b/Lib/test/test_descr.py
@@ -4647,9 +4647,11 @@ order (MRO) for bases """
     def test_mixing_slot_wrappers(self):
         class X(dict):
             __setattr__ = dict.__setitem__
+            __neg__ = dict.copy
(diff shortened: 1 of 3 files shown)


mangrisano (approved):

jdemeyer:
CC @methane

methane:
OK, I think this is safe enough, and this is part of PEP 590 implementation.
Polishing a new feature in beta period is allowed when it is safe enough.

jdemeyer:
Thanks!

PR #13867, opened 2019-06-06
bpo-37138: fix undefined behaviour with memcpy() on NULL array

jdemeyer:
https://bugs.python.org/issue37138

### DIFF
--- a/Objects/classobject.c
+++ b/Objects/classobject.c
@@ -71,7 +71,11 @@ method_vectorcall(PyObject *method, PyObject *const *args,
         }
         /* use borrowed references */
         newargs[0] = self;
-        memcpy(newargs + 1, args, totalargs * sizeof(PyObject *));
+        if (totalargs) { /* bpo-37138: if totalargs == 0, then args may be
+                          * NULL and calling memcpy() with a NULL pointer
+                          * is undefined behaviour. */
+            memcpy(newargs + 1, args, totalargs * sizeof(PyObject *));
+        }
         result = _PyObject_Vectorcall(func, newargs, nargs+1, kwnames);
         PyMem_Free(newargs);
     }


mangrisano:
@vstinner Would you be able to review this fix, please? Thanks a lot. :)

mangrisano:
In any case, I think should be added a test.

gpshead:
I'm not sure an explicit test for this is really needed. (nice to have though)

From the bug, this code path was called all the time with the relevant 0 / null based on the undefined behavior sanitizer buildbot output (though I never bothered trying to trace what codepath's triggered it).  So this case is covered by something.  I'm just not sure explicitly how.

If you feel you can add a test that triggers this specific code path, go for it.

also, i've added the skip news label as this isn't really worth mentioning in news.  its an internal only bugfix for a behavior in beta1 that shouldn't be tripping anyone up in reality.

gpshead (approved):
Approving, the code change makes sense.  I'm not clicking merge for now in case @jdemeyer wants to add an explicit test.  I'll leave that up to Jeroen.

jdemeyer:
It occurs for example every time that `_PyObject_CallNoArg()` is used on a method object. There are plenty of code paths where this happens. To name just one: the implementation of `with obj:` calls `obj.__enter__()` (which is typically a method) this way.

jdemeyer:
In fact, there is already an explicit test in `test_fastcall` in `Lib/test/test_call.py`

vstinner:
> In fact, there is already an explicit test in test_fastcall in Lib/test/test_call.py

Yep:

```
                if not args:
                    # args=NULL, nargs=0
                    result = _testcapi.pyobject_fastcall(func, None)
                    self.check_result(result, expected)
```

I wrote it because this code path caused troubles in the past ;-)

Objects/classobject.c
```diff
@@ -71,7 +71,11 @@ method_vectorcall(PyObject *method, PyObject *const *args,
         }
         /* use borrowed references */
         newargs[0] = self;
-        memcpy(newargs + 1, args, totalargs * sizeof(PyObject *));
+        if (totalargs) { /* bpo-37138: if totalargs == 0, then args may be
```
vstinner, line 74: I don't think that the comment is required.

@gpshead: what do you think?

jdemeyer: I wrote the comment because the reason for that `if (totalargs)` is totally not obvious. It's an obscure point in the C standard that I was unaware of. I think that there are plenty of developers that do not know that `memcpy(dst, NULL, 0)` is undefined behaviour.

gpshead: I'm in favor of such comments. :)

gpshead (approved):

PR #13869, opened 2019-06-06
bpo-37177: make IDLE's search dialogs transient

taleinat:
https://bugs.python.org/issue37177

### DIFF
--- a/Lib/idlelib/idle_test/test_searchbase.py
+++ b/Lib/idlelib/idle_test/test_searchbase.py
@@ -4,7 +4,7 @@
 
 import unittest
 from test.support import requires
-from tkinter import Tk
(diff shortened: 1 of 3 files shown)


taleinat:
@IrvKalb, your review would be appreciated!

terryjreedy:
Even if this does not fix all hiding issues, it should be merged, with a blurb added, as a fix for bpo-24139.
Possible text:  "Properly 'attach' search dialogs to their edit windows so that they work like other dialogs and do not get hidden beneath."

taleinat:
> as a fix for bpo-24139

That appears to be the wrong issue.

terryjreedy:
Sorry, bpo-24039.

PR #13873, opened 2019-06-06
bpo-15913 : Add PyBuffer_SizeFromFormat()

nanjekyejoannah:
Added implementation for PyBuffer_SizeFromFormat().

https://bugs.python.org/issue15913

### DIFF
--- a/Doc/c-api/buffer.rst
+++ b/Doc/c-api/buffer.rst
@@ -463,7 +463,7 @@ Buffer-related functions
 .. c:function:: Py_ssize_t PyBuffer_SizeFromFormat(const char *format)
 
    Return the implied :c:data:`~Py_buffer.itemsize` from :c:data:`~Py_buffer.format`. on success, the
-   size of the return *format* and -1 Otherwise.
+   size of the return *format*. Otherwise, raise an exception and return -1 on error.
 
    .. versionadded:: 3.9
 
--- a/Doc/c-api/buffer.rst
+++ b/Doc/c-api/buffer.rst
@@ -462,8 +462,8 @@ Buffer-related functions
 
 .. c:function:: Py_ssize_t PyBuffer_SizeFromFormat(const char *format)
 
-   Return the implied :c:data:`~Py_buffer.itemsize` from :c:data:`~Py_buffer.format`. on success, the
-   size of the return *format*. Otherwise, raise an exception and return -1 on error.
+   Return the implied :c:data:`~Py_buffer.itemsize` from :c:data:`~Py_buffer.format`. 
+   On error, raise an exception and return -1.
 
    .. versionadded:: 3.9
 
--- a/Doc/c-api/buffer.rst
+++ b/Doc/c-api/buffer.rst
@@ -462,7 +462,7 @@ Buffer-related functions
 
 .. c:function:: Py_ssize_t PyBuffer_SizeFromFormat(const char *format)
 
-   Return the implied :c:data:`~Py_buffer.itemsize` from :c:data:`~Py_buffer.format`. 
+   Return the implied :c:data:`~Py_buffer.itemsize` from :c:data:`~Py_buffer.format`.
    On error, raise an exception and return -1.
 
    .. versionadded:: 3.9
--- a/Doc/c-api/buffer.rst
+++ b/Doc/c-api/buffer.rst
@@ -463,6 +463,8 @@ Buffer-related functions
 .. c:function:: Py_ssize_t PyBuffer_SizeFromFormat(const char *format)
 
    Return the implied :c:data:`~Py_buffer.itemsize` from :c:data:`~Py_buffer.format`.
+   
+   .. versionchanged:: 3.9
 
 
 .. c:function:: int PyBuffer_IsContiguous(Py_buffer *view, char order)
--- a/Doc/c-api/buffer.rst
+++ b/Doc/c-api/buffer.rst
@@ -462,9 +462,10 @@ Buffer-related functions
 
 .. c:function:: Py_ssize_t PyBuffer_SizeFromFormat(const char *format)
 
-   Return the implied :c:data:`~Py_buffer.itemsize` from :c:data:`~Py_buffer.format`.
+   Return the implied :c:data:`~Py_buffer.itemsize` from :c:data:`~Py_buffer.format`. on success, the
+   size of the *format* is returned. Otherwise, return -1.
 
-   .. versionchanged:: 3.9
+   .. versionadded:: 3.9
 
 
 .. c:function:: int PyBuffer_IsContiguous(Py_buffer *view, char order)
--- a/Doc/c-api/buffer.rst
+++ b/Doc/c-api/buffer.rst
@@ -463,7 +463,7 @@ Buffer-related functions
 .. c:function:: Py_ssize_t PyBuffer_SizeFromFormat(const char *format)
 
    Return the implied :c:data:`~Py_buffer.itemsize` from :c:data:`~Py_buffer.format`.
-   
+
    .. versionchanged:: 3.9
 
 
--- a/Doc/c-api/buffer.rst
+++ b/Doc/c-api/buffer.rst
@@ -460,10 +460,9 @@ Buffer-related functions
    :c:func:`PyObject_GetBuffer`.
 
 
-.. c:function:: Py_ssize_t PyBuffer_SizeFromFormat(const char *)
+.. c:function:: Py_ssize_t PyBuffer_SizeFromFormat(const char *format)
 
    Return the implied :c:data:`~Py_buffer.itemsize` from :c:data:`~Py_buffer.format`.
-   This function is not yet implemented.
 
 
 .. c:function:: int PyBuffer_IsContiguous(Py_buffer *view, char order)
--- a/Doc/c-api/buffer.rst
+++ b/Doc/c-api/buffer.rst
@@ -463,7 +463,7 @@ Buffer-related functions
 .. c:function:: Py_ssize_t PyBuffer_SizeFromFormat(const char *format)
 
    Return the implied :c:data:`~Py_buffer.itemsize` from :c:data:`~Py_buffer.format`. on success, the
-   size of the *format* is returned. Otherwise, return -1.
+   size of the return *format* and -1 Otherwise.
 
    .. versionadded:: 3.9
 
--- a/Lib/test/test_buffer.py
+++ b/Lib/test/test_buffer.py
@@ -4422,7 +4422,7 @@ class TestBufferProtocol(unittest.TestCase):
         # basic tests
         for format in ("i", "3s", "0i", " "):
             self.assertEqual(_testcapi.pybuffer_size_from_format(format),
-                             struct.calcsize(format))
+                             struct.calcsize(format.encode()))
 
 if __name__ == "__main__":
     unittest.main()
--- a/Lib/test/test_buffer.py
+++ b/Lib/test/test_buffer.py
@@ -4418,7 +4418,6 @@ class TestBufferProtocol(unittest.TestCase):
         self.assertRaises(BufferError, memoryview, x)
 
     @support.cpython_only
-    @unittest.skipUnless(_testcapi, 'requires _testcapi')
     def test_pybuffer_size_from_format(self):
         # basic tests
         for format in ("i", "3s", "0i", " "):
--- a/Lib/test/test_buffer.py
+++ b/Lib/test/test_buffer.py
@@ -4422,7 +4422,7 @@ class TestBufferProtocol(unittest.TestCase):
         # basic tests
         for format in ("i", "3s", "0i", " "):
             self.assertEqual(_testcapi.pybuffer_size_from_format(format),
-                struct.calcsize(format.encode()))
+                            struct.calcsize(format))
 
 if __name__ == "__main__":
     unittest.main()
--- a/Lib/test/test_buffer.py
+++ b/Lib/test/test_buffer.py
@@ -43,6 +43,11 @@ try:
 except ImportError:
     numpy_array = None
 
+try:
+    import _testcapi
+except ImportError:
+    _testcapi = None
+
 
 SHORT_TEST = True
 
@@ -4412,21 +4417,13 @@ class TestBufferProtocol(unittest.TestCase):
         x = ndarray([1,2,3], shape=[3], flags=ND_GETBUF_FAIL)
         self.assertRaises(BufferError, memoryview, x)
 
-    # Test PyBuffer_SizeFromFormat()
     @support.cpython_only
+    @unittest.skipUnless(_testcapi, 'requires _testcapi')
     def test_pybuffer_size_from_format(self):
-        from _testcapi import pybuffer_size_from_format
-
         # basic tests
         for format in ("i", "3s", "0i", " "):
-            self.assertEqual(pybuffer_size_from_format(format.encode()),
+            self.assertEqual(_testcapi.pybuffer_size_from_format(format),
                 struct.calcsize(format))
-
-        # invalid tests
-        for format in ("nn", "xxx", "xx"):
-            self.assertEqual(pybuffer_size_from_format(format.encode()),
-                struct.calcsize(format))
-
 
 if __name__ == "__main__":
     unittest.main()
--- a/Lib/test/test_buffer.py
+++ b/Lib/test/test_buffer.py
@@ -4422,7 +4422,7 @@ class TestBufferProtocol(unittest.TestCase):
         # basic tests
         for format in ("i", "3s", "0i", " "):
             self.assertEqual(_testcapi.pybuffer_size_from_format(format),
-                            struct.calcsize(format))
+                             struct.calcsize(format))
 
 if __name__ == "__main__":
     unittest.main()
--- a/Lib/test/test_buffer.py
+++ b/Lib/test/test_buffer.py
@@ -4423,7 +4423,7 @@ class TestBufferProtocol(unittest.TestCase):
         # basic tests
         for format in ("i", "3s", "0i", " "):
             self.assertEqual(_testcapi.pybuffer_size_from_format(format),
-                struct.calcsize(format))
+                struct.calcsize(format.encode()))
 
 if __name__ == "__main__":
     unittest.main()
--- a/Lib/test/test_buffer.py
+++ b/Lib/test/test_buffer.py
@@ -4412,6 +4412,22 @@ class TestBufferProtocol(unittest.TestCase):
         x = ndarray([1,2,3], shape=[3], flags=ND_GETBUF_FAIL)
         self.assertRaises(BufferError, memoryview, x)
 
+    # Test PyBuffer_SizeFromFormat()
+    @support.cpython_only
+    def test_pybuffer_size_from_format(self):
+        from ctypes import pythonapi
+        PyBuffer_SizeFromFormat = pythonapi.PyBuffer_SizeFromFormat
+
+        # basic tests
+        self.assertIs(PyBuffer_SizeFromFormat("abc") > 0 , True)
+
+        #invalid input
+        self.assertRaises(struct.error, PyBuffer_SizeFromFormat, b'gg')
+
+        #Empty strings
+        self.assertIs(PyBuffer_SizeFromFormat('') > 0 , True)
+        self.assertIs(PyBuffer_SizeFromFormat('%s') > 0 , True)
+
 
 if __name__ == "__main__":
     unittest.main()
--- a/Lib/test/test_buffer.py
+++ b/Lib/test/test_buffer.py
@@ -4415,20 +4415,17 @@ class TestBufferProtocol(unittest.TestCase):
     # Test PyBuffer_SizeFromFormat()
     @support.cpython_only
     def test_pybuffer_size_from_format(self):
-        from ctypes import pythonapi
-        PyBuffer_SizeFromFormat = pythonapi.PyBuffer_SizeFromFormat
+        from _testcapi import pybuffer_size_from_format
 
         # basic tests
-        self.assertIs(PyBuffer_SizeFromFormat("3si") > 0 , True)
-        self.assertIs(PyBuffer_SizeFromFormat("3s") > 0 , True)
-        self.assertIs(PyBuffer_SizeFromFormat("0i") > 0 , True)
+        for format in ("i", "3s", "0i", " "):
+            self.assertEqual(pybuffer_size_from_format(format.encode()),
+                struct.calcsize(format))
 
-        #invalid input
-        self.assertRaises(struct.error, PyBuffer_SizeFromFormat, b'gg')
-
-        #Empty strings
-        self.assertIs(PyBuffer_SizeFromFormat('') > 0 , True)
-        self.assertIs(PyBuffer_SizeFromFormat('%s') > 0 , True)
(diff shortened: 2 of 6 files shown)


nanjekyejoannah:
Requesting review @skrah @pitrou @vstinner

pablogsal:
I think this needs tests. Implementation aside, calling the current function with `3si 3s 0i` segfaults for me.

nanjekyejoannah:
@pablogsal , thought about that  also but am still wondering where to place this test?

pablogsal:
> @pablogsal , thought about that also but am still wondering where to place this test?

I would say exposed in `Modules/_testcapimodule.c` and then tested with `Lib/test/test_capi.py` but Victor or Antoine would know better regarding this specific API.

Doc/c-api/buffer.rst
```diff
@@ -460,10 +460,11 @@ Buffer-related functions
    :c:func:`PyObject_GetBuffer`.
 
 
-.. c:function:: Py_ssize_t PyBuffer_SizeFromFormat(const char *)
+.. c:function:: Py_ssize_t PyBuffer_SizeFromFormat(const char *format)
 
    Return the implied :c:data:`~Py_buffer.itemsize` from :c:data:`~Py_buffer.format`.
-   This function is not yet implemented.
+
+   .. versionchanged:: 3.9
```
vstinner: the function is new in 3.9 no? In that case, it should be "versionadded:: 3.9".

Objects/abstract.c
```diff
@@ -495,6 +500,30 @@ _Py_add_one_to_index_C(int nd, Py_ssize_t *index, const Py_ssize_t *shape)
     }
 }
 
+int
+PyBuffer_SizeFromFormat(const char *format)
+{
+    PyObject *result;
+    int size;
+
+    structmodule = PyImport_ImportModule("struct");
```
vstinner, line 507: it's a strong reference: you must Py_DECREF(structmodule) to prevent a reference leak. I suggest to add a "done:" label which does DECREF and replace return NULL with "goto done".

Objects/abstract.c
```diff
@@ -495,6 +500,30 @@ _Py_add_one_to_index_C(int nd, Py_ssize_t *index, const Py_ssize_t *shape)
...
+    if (structmodule == NULL)
+        return NULL;
```
vstinner: wait, NULL is a pointer. The function return type is int: you should return -1 in case of error, no? Same code for others if below.

Doc/c-api/buffer.rst
```diff
@@ -460,10 +460,11 @@ Buffer-related functions
    :c:func:`PyObject_GetBuffer`.
 
 
-.. c:function:: Py_ssize_t PyBuffer_SizeFromFormat(const char *)
+.. c:function:: Py_ssize_t PyBuffer_SizeFromFormat(const char *format)
 
    Return the implied :c:data:`~Py_buffer.itemsize` from :c:data:`~Py_buffer.format`.
```
vstinner, line 465: Please document how errors are reported. I suggest to return -1.

vstinner (changes requested):

vstinner:
Would it be possible to add an unit test? Expose the function in _testcapi and add a test in test_buffer using _testcapi (such test should be marked with ``@cpython_only``, or skipped if _testcapi is None).

nanjekyejoannah:
I have made some changes locally that will address some of the comments @vstinner. I will update the PR when I finish the tests.

vstinner:
> I have made some changes locally that will address some of the comments @vstinner. I will update the PR when I finish the tests.

Ok. Ping me when I should review again ;-)

Lib/test/test_buffer.py
```diff
@@ -4412,6 +4412,24 @@ def test_issue_7385(self):
...
         self.assertRaises(BufferError, memoryview, x)
 
+    # Test PyBuffer_SizeFromFormat()
+    @support.cpython_only
+    def test_pybuffer_size_from_format(self):
+        from ctypes import pythonapi
+        PyBuffer_SizeFromFormat = pythonapi.PyBuffer_SizeFromFormat
+
+        # basic tests
+        self.assertIs(PyBuffer_SizeFromFormat("3si") > 0 , True)
```
vstinner: I suggest to test a few formats using a loop and compare the result to struct.calcsize(). Something like:

```
for format in ("i", "3s", "0i", ""):
    self.assertEqual(PyBuffer_SizeFromFormat(format.encode()), struct.calcsize(format))
```

You must explicitly encode the Unicode format string to UTF-8 to pass it as a byte string (char*), no?

Objects/abstract.c
```diff
@@ -495,6 +494,47 @@ _Py_add_one_to_index_C(int nd, Py_ssize_t *index, const Py_ssize_t *shape)
...
+
+    fmt = PyBytes_FromString(format);
+    if (fmt == NULL)
+        goto error;
+
+    structobj = PyObject_CallFunctionObjArgs(Struct, fmt, NULL);
+    if (structobj == NULL)
+        goto error;
+
+    itemsize = (intptr_t)structobj;
```
vstinner: I don't think that it works. structobj is a Python object. You should use PyLong_AsSsize_t() to get it as Py_ssize_t.

Objects/abstract.c
```diff
@@ -495,6 +494,47 @@ _Py_add_one_to_index_C(int nd, Py_ssize_t *index, const Py_ssize_t *shape)
     }
 }
 
+int
```
vstinner: The doc uses Py_ssize_t, but you chose int. I would prefer to use Py_ssize_t.

Objects/abstract.c
```diff
@@ -495,6 +494,47 @@ _Py_add_one_to_index_C(int nd, Py_ssize_t *index, const Py_ssize_t *shape)
...
+PyBuffer_SizeFromFormat(const char *format)
+{
+    PyObject *structmodule;
+    PyObject *Struct = NULL;
+    PyObject *structobj = NULL;
+    PyObject *fmt = NULL;
+    int itemsize;
```
vstinner: initialize itemsize to -1, so the error label can be removed, and you can just use "goto done".

Objects/abstract.c
```diff
@@ -495,6 +494,47 @@ _Py_add_one_to_index_C(int nd, Py_ssize_t *index, const Py_ssize_t *shape)
     }
 }
 
+int
+PyBuffer_SizeFromFormat(const char *format)
+{
+    PyObject *structmodule;
+    PyObject *Struct = NULL;
+    PyObject *structobj = NULL;
```
vstinner: I suggest to rename it to "res".

vstinner (changes requested):

Lib/test/test_buffer.py
```diff
@@ -4412,6 +4412,24 @@ def test_issue_7385(self):
         x = ndarray([1,2,3], shape=[3], flags=ND_GETBUF_FAIL)
         self.assertRaises(BufferError, memoryview, x)
 
+    # Test PyBuffer_SizeFromFormat()
+    @support.cpython_only
+    def test_pybuffer_size_from_format(self):
+        from ctypes import pythonapi
+        PyBuffer_SizeFromFormat = pythonapi.PyBuffer_SizeFromFormat
```
pablogsal: This test should be able to run even if `ctypes` is not compiled. Could you expose this function in `Modules/_testcapimodule.c`?

Objects/abstract.c
```diff
@@ -495,6 +494,47 @@ _Py_add_one_to_index_C(int nd, Py_ssize_t *index, const Py_ssize_t *shape)
...
+    Py_DECREF(structmodule);
+    if (Struct == NULL)
+        goto error;
+
+    fmt = PyBytes_FromString(format);
+    if (fmt == NULL)
+        goto error;
+
+    structobj = PyObject_CallFunctionObjArgs(Struct, fmt, NULL);
+    if (structobj == NULL)
```
pablogsal: All of these conditionals should comply with pep7 (use braces).

Objects/abstract.c
```diff
@@ -495,6 +494,47 @@ _Py_add_one_to_index_C(int nd, Py_ssize_t *index, const Py_ssize_t *shape)
...
+PyBuffer_SizeFromFormat(const char *format)
+{
+    PyObject *structmodule = NULL;
+    PyObject *Struct = NULL;
+    PyObject *res = NULL;
+    PyObject *fmt = NULL;
+    Py_ssize_t itemsize = -1;
+
+    structmodule = PyImport_ImportModule("struct");
+    if (structmodule == NULL)
```
vstinner: Nitpick: PEP 7 requires {...} here.

Doc/c-api/buffer.rst
```diff
@@ -460,10 +460,12 @@ Buffer-related functions
    :c:func:`PyObject_GetBuffer`.
 
 
-.. c:function:: Py_ssize_t PyBuffer_SizeFromFormat(const char *)
+.. c:function:: Py_ssize_t PyBuffer_SizeFromFormat(const char *format)
 
-   Return the implied :c:data:`~Py_buffer.itemsize` from :c:data:`~Py_buffer.format`.
-   This function is not yet implemented.
+   Return the implied :c:data:`~Py_buffer.itemsize` from :c:data:`~Py_buffer.format`. on success, the
+   size of the return *format* and -1 Otherwise.
```
vstinner: Sorry, I don't understand the second sentence. I would suggest to replace it with:

Raise an exception and return -1 on error.

Objects/abstract.c
```diff
@@ -495,6 +494,47 @@ _Py_add_one_to_index_C(int nd, Py_ssize_t *index, const Py_ssize_t *shape)
...
+Py_ssize_t
+PyBuffer_SizeFromFormat(const char *format)
+{
+    PyObject *structmodule = NULL;
+    PyObject *Struct = NULL;
+    PyObject *res = NULL;
+    PyObject *fmt = NULL;
+    Py_ssize_t itemsize = -1;
+
+    structmodule = PyImport_ImportModule("struct");
```
vstinner, line 507: You are leaking one reference to structmodule: Py_DECREF(structmodule) somewhere is needed.

You can check for refleak using: ./python -m test -R 3:3 test_buffer -m test_pybuffer_size_from_format

Objects/abstract.c
```diff
@@ -495,6 +494,47 @@ _Py_add_one_to_index_C(int nd, Py_ssize_t *index, const Py_ssize_t *shape)
     }
 }
 
+Py_ssize_t
+PyBuffer_SizeFromFormat(const char *format)
+{
+    PyObject *structmodule = NULL;
+    PyObject *Struct = NULL;
```
vstinner: can you please rename it to "calcsize"? It's not a Struct class.

Modules/_testcapimodule.c
```diff
@@ -3301,6 +3301,19 @@ getbuffer_with_null_view(PyObject* self, PyObject *obj)
     Py_RETURN_NONE;
 }
 
+/* test PyBuffer_SizeFromFormat() */
+static Py_ssize_t
```
vstinner: That's wrong: you must return PyObject*.

Modules/_testcapimodule.c
```diff
@@ -3301,6 +3301,19 @@ getbuffer_with_null_view(PyObject* self, PyObject *obj)
...
+pybuffer_size_from_format(PyObject* self, PyObject *args)
+{
+    const char *format;
+
+    if (!PyArg_ParseTuple(args, "s:pybuffer_size_from_format",
+                          &format))
+        return -1;
+
+    return PyBuffer_SizeFromFormat(format);
```
vstinner: You can use for example:

```
    return PyLong_FromSsize_t(PyBuffer_SizeFromFormat(format));
```

Modules/_testcapimodule.c
```diff
@@ -3301,6 +3301,19 @@ getbuffer_with_null_view(PyObject* self, PyObject *obj)
...
 }
 
+/* test PyBuffer_SizeFromFormat() */
+static Py_ssize_t
+pybuffer_size_from_format(PyObject* self, PyObject *args)
+{
+    const char *format;
+
+    if (!PyArg_ParseTuple(args, "s:pybuffer_size_from_format",
+                          &format))
```
vstinner: nitpick: please add {...}, I know that old code doesn't respect PEP 7, but new code should respect it.

nanjekyejoannah:
have made the requested changes; please review again.

nanjekyejoannah:

Lib/test/test_buffer.py
```diff
@@ -4412,6 +4417,13 @@ def test_issue_7385(self):
         x = ndarray([1,2,3], shape=[3], flags=ND_GETBUF_FAIL)
         self.assertRaises(BufferError, memoryview, x)
 
+    @support.cpython_only
```
vstinner, line 4420: @support.cpython_only should be removed: it's redundant with @unittest.skipUnless(_testcapi, 'requires _testcapi').

Misc/NEWS.d/next/Core and Builtins/2019-06-06-20-52-38.bpo-15913.5Sg5cv.rst
```diff
@@ -0,0 +1 @@
+`PyBuffer_SizeFromFormat()` now added.
```
vstinner: Since it's documented, you can use the :c:func: markup to add a link:

```
:c:func:`PyBuffer_SizeFromFormat` now added.
```

Objects/abstract.c
```diff
@@ -7,7 +7,6 @@
 #include "longintrepr.h"
 
 
-
```
vstinner: This change is unrelated, please revert it.

Doc/c-api/buffer.rst
```diff
@@ -460,10 +460,12 @@ Buffer-related functions
    :c:func:`PyObject_GetBuffer`.
 
 
-.. c:function:: Py_ssize_t PyBuffer_SizeFromFormat(const char *)
+.. c:function:: Py_ssize_t PyBuffer_SizeFromFormat(const char *format)
 
-   Return the implied :c:data:`~Py_buffer.itemsize` from :c:data:`~Py_buffer.format`.
-   This function is not yet implemented.
+   Return the implied :c:data:`~Py_buffer.itemsize` from :c:data:`~Py_buffer.format`. on success, the
+   size of the return *format*. Otherwise, raise an exception and return -1 on error.
```
vstinner: I don't understand "on success, the size of the return *format*." sentence, and it's redundant with the first sentence, no? I suggest to remove it, and just add:

On error, raise an exception and return -1.

vstinner (changes requested):

Lib/test/test_buffer.py
```diff
@@ -4412,6 +4417,12 @@ def test_issue_7385(self):
...
+    def test_pybuffer_size_from_format(self):
+        # basic tests
+        for format in ("i", "3s", "0i", " "):
+            self.assertEqual(_testcapi.pybuffer_size_from_format(format),
+                struct.calcsize(format.encode()))
```
vstinner: .encode() is useless and a little bit inconsistenct since format isn't encoded to call pybuffer_size_from_format().

Would you mind to remove the .encode() call (just pass format) but also indent the code so "struct.calcsize..." is aligned to "_testcapi.pybuffer..."?

vstinner (approved):
LGTM. But please make the last requested changes.

Objects/abstract.c
```diff
@@ -495,6 +495,48 @@ _Py_add_one_to_index_C(int nd, Py_ssize_t *index, const Py_ssize_t *shape)
...
+{
+    PyObject *structmodule = NULL;
+    PyObject *calcsize = NULL;
+    PyObject *res = NULL;
+    PyObject *fmt = NULL;
+    Py_ssize_t itemsize = -1;
+
+    structmodule = PyImport_ImportModule("struct");
+    if (structmodule == NULL) {
+        goto done;
```
ZackerySpytz: If this `goto done;` is executed, a segfault will occur at `Py_DECREF(structmodule);`.

Modules/_testcapimodule.c
```diff
@@ -3301,6 +3301,20 @@ getbuffer_with_null_view(PyObject* self, PyObject *obj)
...
+pybuffer_size_from_format(PyObject *self, PyObject *args)
+{
+    const char *format;
+
+    if (!PyArg_ParseTuple(args, "s:pybuffer_size_from_format",
+                          &format)) {
+        return NULL;
+    }
+
+    return PyLong_FromSsize_t(PyBuffer_SizeFromFormat(format));
```
ZackerySpytz: I don't believe `PyBuffer_SizeFromFormat()` should be inlined like this. Please check it for failure before the `PyLong_FromSsize_t()` call.

Objects/abstract.c
```diff
@@ -495,6 +495,48 @@ _Py_add_one_to_index_C(int nd, Py_ssize_t *index, const Py_ssize_t *shape)
...
+    }
+
+    res = PyObject_CallFunctionObjArgs(calcsize, fmt, NULL);
+    if (res == NULL) {
+        goto done;
+    }
+
+    itemsize = PyLong_AsSsize_t(res);
+    if (itemsize < 0) {
```
ZackerySpytz, line 528: This check is not needed.

Objects/abstract.c
nanjekyejoannah: How did you reproduce the segfault? I suspected a leak but verified none. I just ran one of the tests that use this function. using `./python -m test -R 3:3 test_buffer -m test_pybuffer_size_from_format`

vstinner: You can add "structmodule = NULL;" before the if to reproduce the crash.

Modules/_testcapimodule.c
vstinner: Oh right, NULL should be returned if the result is -1.

Objects/abstract.c
vstinner: calcsize() might return a integer which doesn't fit into a C Py_ssize_t. It's unlikely, but I prefer to keep the test.

Objects/abstract.c
ZackerySpytz: The segfault would only occur if `PyImport_ImportModule()` failed (which is unlikely, but possible). It is enough to just use `return -1;` here instead of the `goto`.

Lib/test/test_buffer.py
nanjekyejoannah: I have also seen that not encoding weirdly works and tests pass locally but will fail on the CI. So I tried to revert the .encode() on struct.calsize().

Objects/abstract.c
ZackerySpytz: I know that. I meant that the explicit check is not needed because the control flow will fall through to the `done` label.

vstinner:
Without .encode, the test failed with:

```
-----------------------------------
  File "/home/travis/build/python/cpython/Lib/test/test_buffer.py", line 4425, in test_pybuffer_size_from_format
    struct.calcsize(format))
BytesWarning: Comparison between bytes and string
```

You should be able to reproduce rhe issue using: ./python -bb -m tets test_buffer

The test should pass without .encode. You should investigate why it fails. struct.calcsize() accepts format as a Unicode string, no?

nanjekyejoannah:
The test should pass without .encode. You should investigate why it fails. struct.calcsize() accepts format as a Unicode string, no?

I needed to change `fmt = PyBytes_FromString(format)` to `fmt = PyUnicode_FromString(format`); in order to do away with the comparison between bytes and string error.

The test passes now :

```
gitpod /workspace/cpython $ ./python -bb -m test test_buffer
Run tests sequentially
0:00:00 load avg: 0.69 [1/1] test_buffer

== Tests result: SUCCESS ==

1 test OK.

```

nanjekyejoannah:

Misc/NEWS.d/next/Core and Builtins/2019-06-06-20-52-38.bpo-15913.5Sg5cv.rst
```diff
@@ -0,0 +1 @@
+:c:func:`PyBuffer_SizeFromFormat()` now added.
```
vstinner: I suggest:

```Implement :c:func:`PyBuffer_SizeFromFormat()` function (previously documented but not implemented): call :func:`struct.calcsize`.```

File: Lib/test/test_buffer.py [hunk 28]
```diff
@@ -4412,6 +4417,12 @@ def test_issue_7385(self):
         x = ndarray([1,2,3], shape=[3], flags=ND_GETBUF_FAIL)
         self.assertRaises(BufferError, memoryview, x)
 
+    @support.cpython_only
+    def test_pybuffer_size_from_format(self):
+        nitems = randrange(1, 30)
```
vstinner: I would prefer to avoid using random in such test. iter_format() returns 91 items: it sounds overkill to me. IMHO tTesting 3 formats should but enough, like:

```
for fmt in ('', 'ii', '3s'):
```

Modules/_testcapimodule.c
```diff
@@ -5087,6 +5107,7 @@ static PyMethodDef TestMethods[] = {
     {"test_pep3118_obsolete_write_locks", (PyCFunction)test_pep3118_obsolete_write_locks, METH_NOARGS},
 #endif
     {"getbuffer_with_null_view", getbuffer_with_null_view, METH_O},
+    {"pybuffer_size_from_format", (PyCFunction)pybuffer_size_from_format, METH_VARARGS},
```
vstinner: Maybe reuse the same name:
```suggestion
    {"PyBuffer_SizeFromFormat", test_PyBuffer_SizeFromFormat, METH_VARARGS},
```

Note: I don't think that (PyCFunction) cast is needed here.

vstinner (changes requested):

Modules/_testcapimodule.c
```diff
@@ -3301,6 +3301,26 @@ getbuffer_with_null_view(PyObject* self, PyObject *obj)
...
 }
 
+/* PyBuffer_SizeFromFormat() */
+static PyObject *
+test_PyBuffer_SizeFromFormat(PyObject *self, PyObject *args)
+{
+    const char *format;
+    Py_ssize_t result;
+
+    if (!PyArg_ParseTuple(args, "s:pybuffer_size_from_format",
```
vstinner: You forget to update the function name here.

Lib/test/test_buffer.py
```diff
@@ -4412,6 +4417,11 @@ def test_issue_7385(self):
         x = ndarray([1,2,3], shape=[3], flags=ND_GETBUF_FAIL)
         self.assertRaises(BufferError, memoryview, x)
 
-
+    @support.cpython_only
+    def test_pybuffer_size_from_format(self):
+        # basic tests
+        for format in ("i", "3s", "0i", " "):
```
vstinner: " " <= my proposal was to test an empty format string

nanjekyejoannah:
@vstinner I don't think this CI fail is now my business. Is it? i.e

`Windows PR Tests win64: "The job running on agent Azure Pipelines 7 has exceeded the maximum execution time of 60. For more information, see https://go.microsoft.com/fwlink/?linkid=2077134 "`

vstinner:
I merged your PR @nanjekyejoannah, thanks!

> @vstinner I don't think this CI fail is now my business.

I don't know why Azure CI failed, but this CI is not mandatory yet.

PR #13876, opened 2019-06-07
IDLE: Standardize naming convention for DummyEditwin in tests

csabella:
Change from Dummy_Editwin to DummyEditwin to match other tests.


terryjreedy:
LGTM.  Now we  just need to remember.

PR #13881, opened 2019-06-07
bpo-37188: Fix a divide-by-zero in arrays of size-0 objects

eric-wieser:
`itemsize` can be zero when `_type_` is an empty struct or length-0 array.

This line used to be `length * itemsize < 0`, which didn't exhibit this problem.

https://bugs.python.org/issue37188

### DIFF
--- a/Lib/ctypes/test/test_arrays.py
+++ b/Lib/ctypes/test/test_arrays.py
@@ -208,6 +208,21 @@ class ArrayTestCase(unittest.TestCase):
             _type_ = c_int
             _length_ = 0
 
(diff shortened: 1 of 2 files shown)


eric-wieser:
@skrah, mind taking a look since this blocks #5576?

skrah:
@eric-wieser  Thanks for catching this!  This appears to have been backported to 2.7 and the auto-backport fails. Do you have time to do a manual backport PR?

skrah:
Actually I missed `assert sizeof(obj) == 0`, which should use `self.assertEqual()`.  I can change that in a follow-up.

PR #13887, opened 2019-06-07
[3.8] bpo-37191: Avoid declaration-after-statement in header included from Python.h

encukou:
When compiled with GCC's -Werror=declaration-after-statement ("intermingled declarations" in PEP7), cpython/abstract.h (included from <Python.h>) errors on vectorcall helper functions added in 3.8.0 Beta 1.

It's well within our rights to ignore this: since 3.6 we require intermingled declarations.
But, when re-compiling Fedora we've seen several projects fail with this warning (so far: pygobject3, python-dbus, xen; more will likely come).

@ambv: Dear Release Manager, should we patch 3.8 to avoid this? The patch is simple, and it would give projects that we(re) dutifully tested with the Alphas one more release to adapt.

I don't think it's worth changing for 3.9 (but if we do we should test it).

https://bugs.python.org/issue37191

### DIFF
--- a/Include/cpython/abstract.h
+++ b/Include/cpython/abstract.h
@@ -81,13 +81,14 @@ static inline vectorcallfunc
 _PyVectorcall_Function(PyObject *callable)
 {
     PyTypeObject *tp = Py_TYPE(callable);
+    Py_ssize_t offset = tp->tp_vectorcall_offset;
+    vectorcallfunc *ptr;
     if (!PyType_HasFeature(tp, _Py_TPFLAGS_HAVE_VECTORCALL)) {
         return NULL;
     }
     assert(PyCallable_Check(callable));
-    Py_ssize_t offset = tp->tp_vectorcall_offset;
     assert(offset > 0);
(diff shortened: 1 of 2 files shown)


jdemeyer:
> I don't think it's worth changing for 3.9 (but if we do we should test it).

To use an argument that you like to use yourself: fixing this both on 3.8 and 3.9 would be better for backportability. It's not unlikely that bugfixes are going to be made to the vectorcall functionality.

vstinner:
I'm working on a different approach: https://bugs.python.org/issue37194

vstinner:
My https://bugs.python.org/issue37194 idea was based on assumptions which were false. This change is really tiny and fix a very practical issue. We can argue forever about the C standard, but I don't think that it's worth it :-)

> To use an argument that you like to use yourself: fixing this both on 3.8 and 3.9 would be better for backportability. It's not unlikely that bugfixes are going to be made to the vectorcall functionality.

I'm in favor of also applying this change to Python 3.9. But it's up to you @encukou ;-)

PR #13888, opened 2019-06-07
bpo-37169: Rewrite _PyObject_IsFreed() unit tests

vstinner:
Replace two Python function calls with a single one to ensure that no
memory allocation is done between the invalid object is created and
when _PyObject_IsFreed() is called.

https://bugs.python.org/issue37169


pablogsal (approved):
LGTM

Why we have never seen this test fail?

vstinner:
> Why we have never seen this test fail?

No idea why https://bugs.python.org/issue37169 only fails on the openSUSE builder. It might depend on compiler optimizations like PGO or LTO?

PR #13889, opened 2019-06-07
[WIP] bpo-36402: add random sleep to test_threads_join_2()

vstinner:
Add random short sleep of around 10 ms to test_threads_join_2() to
reproduce https://bugs.python.org/issue36402 crash.

### DIFF
--- a/Lib/test/test_threading.py
+++ b/Lib/test/test_threading.py
@@ -894,36 +894,34 @@ class SubinterpThreadingTests(BaseTestCase):
         self.assertEqual(os.read(r, 1), b"x")
 
     def test_threads_join_2(self):
-        # Same as above, but a delay gets introduced after the thread's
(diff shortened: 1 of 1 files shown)


pablogsal:
How long the test takes to run? I am asking because if it is low, we can run the test multiple times so it does not magically pass on the main CI and then fail somewhere else. Or is more or less reliable that will fail almost all the time?

vstinner:
@pablogsal: Using this PR, the test is way more likely when run in a loop using:

```
./python -m test  -F -m test_threads_join_2  test_threading
```

The CI only runs the test once and so was just lucky. You might tune my PR to make the crash more reliable, but I didn't bother to go that far ;-)

vstinner:
I used this PR to debug https://bugs.python.org/issue36402

I merged my PR #13948 fix which also includes changes of this PR.

PR #13893, opened 2019-06-07
bpo-37193: remove thread objects which finished process its request

maru-n:
In ThreadingMixIn class, it hold child processes on list (self._threads variable) to join it when server is closed.
This list also contain thread objects which finished its process.
And it will be extended eternally.
So Memory usage keep increasing while running server until server_close() called in spite of all threads finished to process request in real time and computer resource is enough to process all request.
This PR remove thread objects after finished to avoid it.

https://bugs.python.org/issue37193

### DIFF
--- a/Lib/socketserver.py
+++ b/Lib/socketserver.py
@@ -652,11 +652,13 @@ class ThreadingMixIn:
         except Exception:
             self.handle_error(request, client_address)
         finally:
-            self.shutdown_request(request)
-            thread = threading.current_thread()
-            with self._threads_lock:
-                if self._threads and not thread.daemon:
-                    self._threads.remove(thread)
+            try:
+                self.shutdown_request(request)
+            finally:
+                thread = threading.current_thread()
+                with self._threads_lock:
+                    if self._threads and not thread.daemon:
+                        self._threads.remove(thread)
 
     def process_request(self, request, client_address):
         """Start a new thread to process the request."""
--- a/Lib/socketserver.py
+++ b/Lib/socketserver.py
@@ -663,7 +663,10 @@ class ThreadingMixIn:
         if not thread.daemon:
             with self._threads_lock:
                 if self._threads is not None:
-                    self._threads.remove(thread)
+                    try:
+                        self._threads.remove(thread)
+                    except ValueError:
+                        pass
 
     def process_request(self, request, client_address):
         """Start a new thread to process the request."""
--- a/Lib/socketserver.py
+++ b/Lib/socketserver.py
@@ -652,6 +652,11 @@ class ThreadingMixIn:
             self.handle_error(request, client_address)
         finally:
             self.shutdown_request(request)
+            t = threading.current_thread()
+            try:
+                self._threads.remove(t)
+            except AttributeError:
+                pass
 
     def process_request(self, request, client_address):
         """Start a new thread to process the request."""
--- a/Lib/socketserver.py
+++ b/Lib/socketserver.py
@@ -639,7 +639,7 @@ class ThreadingMixIn:
     # For non-daemonic threads, list of threading.Threading objects
     # used by server_close() to wait for all threads completion.
     _threads = None
-    _threads_lock = threading.Lock()
+    _threads_lock = None
 
     def process_request_thread(self, request, client_address):
         """Same as in BaseServer but as a thread.
@@ -661,6 +661,8 @@ class ThreadingMixIn:
         """Remove a current thread from threads list."""
         thread = threading.current_thread()
         if not thread.daemon:
+            if self._threads_lock is None:
+                self._threads_lock = threading.Lock()
             with self._threads_lock:
                 if self._threads is not None:
                     try:
@@ -674,6 +676,8 @@ class ThreadingMixIn:
                              args = (request, client_address))
         t.daemon = self.daemon_threads
         if not t.daemon and self.block_on_close:
+            if self._threads_lock is None:
+                self._threads_lock = threading.Lock()
             with self._threads_lock:
                 if self._threads is None:
                     self._threads = []
@@ -683,6 +687,8 @@ class ThreadingMixIn:
     def server_close(self):
         super().server_close()
         if self.block_on_close:
+            if self._threads_lock is None:
+                self._threads_lock = threading.Lock()
             with self._threads_lock:
                 threads = self._threads
                 self._threads = None
--- a/Lib/socketserver.py
+++ b/Lib/socketserver.py
@@ -654,7 +654,8 @@ class ThreadingMixIn:
             self.shutdown_request(request)
             thread = threading.current_thread()
             try:
-                self._threads.remove(thread)
+                if not thread.daemon:
+                    self._threads.remove(thread)
             except AttributeError:
                 pass
 
--- a/Lib/socketserver.py
+++ b/Lib/socketserver.py
@@ -652,9 +652,9 @@ class ThreadingMixIn:
             self.handle_error(request, client_address)
         finally:
             self.shutdown_request(request)
-            t = threading.current_thread()
+            thread = threading.current_thread()
             try:
-                self._threads.remove(t)
+                self._threads.remove(thread)
             except AttributeError:
                 pass
 
--- a/Lib/socketserver.py
+++ b/Lib/socketserver.py
@@ -655,11 +655,15 @@ class ThreadingMixIn:
             try:
                 self.shutdown_request(request)
             finally:
-                thread = threading.current_thread()
-                if not thread.daemon:
-                    with self._threads_lock:
-                        if self._threads is not None:
-                            self._threads.remove(thread)
+                self._remove_thread()
+
+    def _remove_thread(self):
+        """Remove a current thread from threads list."""
+        thread = threading.current_thread()
+        if not thread.daemon:
+            with self._threads_lock:
+                if self._threads is not None:
+                    self._threads.remove(thread)
 
     def process_request(self, request, client_address):
         """Start a new thread to process the request."""
--- a/Lib/socketserver.py
+++ b/Lib/socketserver.py
@@ -676,8 +676,8 @@ class ThreadingMixIn:
     def server_close(self):
         super().server_close()
         if self.block_on_close:
-            threads = self._threads
             with self._threads_lock:
+                threads = self._threads
                 self._threads = None
             if threads:
                 for thread in threads:
--- a/Lib/socketserver.py
+++ b/Lib/socketserver.py
@@ -639,6 +639,7 @@ class ThreadingMixIn:
     # For non-daemonic threads, list of threading.Threading objects
     # used by server_close() to wait for all threads completion.
     _threads = None
+    _threads_lock = threading.Lock()
 
     def process_request_thread(self, request, client_address):
         """Same as in BaseServer but as a thread.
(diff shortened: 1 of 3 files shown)


mangrisano:
Hi and thank you for the pull request. Just one question about the change: did you test it? 
I'm not a core-dev so feel free to don't follow my hint :)

mangrisano (approved):

maru-n:
Hello, Thank you for your kindly comment.

> Just one question about the change: did you test it?

I only checked that this PR fixed the problem by memory monitoring tool and sample code written on the issue (https://bugs.python.org/issue37193) and didn't write test code because I couldn't decide whether it is necessary. (of course, it passed original test, test_socketserver)
If test code is required, I'm going to add it.

mangrisano:
> If test code is required, I'm going to add it.

Perhaps it's better to wait for a core-dev for reviewing. :)

Lib/socketserver.py
```diff
@@ -652,6 +652,11 @@ def process_request_thread(self, request, client_address):
             self.handle_error(request, client_address)
         finally:
             self.shutdown_request(request)
+            t = threading.current_thread()
```
vstinner: You need a lock to protect self._threads. You should not call .remove() if t.daemon is true.

Please rename t to thread.

We should be extra careful and use a nested finally block in case shutdown_request() raises a new exception:

```
finally:
  try:
    self.shutdown_request(request)
  finally:
    ...
```

maru-n: thank for your reviewing.
I fixed and committed points you suggested.
Could you re-check this?

Lib/socketserver.py
```diff
@@ -651,24 +652,32 @@ def process_request_thread(self, request, client_address):
         except Exception:
             self.handle_error(request, client_address)
         finally:
-            self.shutdown_request(request)
+            try:
+                self.shutdown_request(request)
+            finally:
+                thread = threading.current_thread()
```
vstinner: Can you please put this code into a private sub-method, something like: self._remove_thread()?

Lib/socketserver.py
```diff
@@ -651,24 +652,32 @@ def process_request_thread(self, request, client_address):
...
+                with self._threads_lock:
+                    if self._threads and not thread.daemon:
```
vstinner: I would prefer to test: "self._threads is not None".

Would you mind to test "if not thread.daemon" before "with self._threads_lock:"?

```
if not thread.daemon:
   with self._threads_lock:
     if self._threads  is not None: ...
```

It's just to avoid holding the lock if it's useless.

Lib/socketserver.py
```diff
@@ -651,24 +652,32 @@ def process_request_thread(self, request, client_address):
...
+                        self._threads.remove(thread)
```
vstinner: I would prefer to ignore ValueError in case thread is already removed from _threads. It shouldn't happen, but I prefer to be extra careful.

Lib/socketserver.py
```diff
@@ -651,24 +652,32 @@ def process_request_thread(self, request, client_address):
...
+                    self._threads = []
+                self._threads.append(t)
         t.start()
 
     def server_close(self):
         super().server_close()
         if self.block_on_close:
             threads = self._threads
-            self._threads = None
+            with self._threads_lock:
```
vstinner: You must move "threads = self._threads" inside the with block for correctness.

Lib/socketserver.py
```diff
@@ -639,6 +639,7 @@ class ThreadingMixIn:
     # For non-daemonic threads, list of threading.Threading objects
     # used by server_close() to wait for all threads completion.
     _threads = None
+    _threads_lock = threading.Lock()
```
vstinner: Wait. Written like that, *all* class instances share the same lock. It seems wrong to me. Each instance must have its own lock, no?

vstinner (changes requested):

Lib/socketserver.py
maru-n: Yes, I was careless. You are totally right.
Thank you for your comment. I fixed this.

maru-n:
Hello @vstinner 
Thank you for your good review again.

By the way, now it seems that source code around _threads_lock (and _threads list) is redundant and not clear for me. (especially testing if it is _none_ or not and initialize it)
I would prefer to put it together, for example in initialization,
`
def __init__(self, *args, **kwargs):
    super().__init__(*args, **kwargs)
    self._threads_lock = threading.Lock()
    self._threads = []
`
How you do you feel about this?

Lib/socketserver.py
```diff
@@ -651,24 +652,46 @@ def process_request_thread(self, request, client_address):
...
     def process_request(self, request, client_address):
         """Start a new thread to process the request."""
         t = threading.Thread(target = self.process_request_thread,
                              args = (request, client_address))
         t.daemon = self.daemon_threads
         if not t.daemon and self.block_on_close:
-            if self._threads is None:
-                self._threads = []
-            self._threads.append(t)
+            if self._threads_lock is None:
```
vstinner: Would it be possible to initialize _threads_lock and _threads *before* creating the first thread?

vstinner:
The problem is that ThreadingMixIn is used with multiple inheritance. If you add an __init__ method, the __init__ method of the second inherited class would no longer be called. Examples:

```
class ThreadingUDPServer(ThreadingMixIn, UDPServer): pass
class ThreadingTCPServer(ThreadingMixIn, TCPServer): pass
```

Lib/socketserver.py
```diff
@@ -651,24 +652,46 @@ def process_request_thread(self, request, client_address):
...
+                self.shutdown_request(request)
+            finally:
+                self._remove_thread()
+
+    def _remove_thread(self):
+        """Remove a current thread from threads list."""
+        thread = threading.current_thread()
+        if not thread.daemon:
+            if self._threads_lock is None:
+                self._threads_lock = threading.Lock()
```
vstinner: That's wrong: the lock must exist before the first thread is spawned. Otherwise, two threads might use two different locks... (There would be a race condition on the lock creation.)

Lib/socketserver.py
```diff
@@ -651,24 +652,46 @@ def process_request_thread(self, request, client_address):
...
+            with self._threads_lock:
+                if self._threads is not None:
```
vstinner: I don't think that self._threads can be None there. I suggest to remove the if.

Lib/socketserver.py
```diff
@@ -651,24 +652,46 @@ def process_request_thread(self, request, client_address):
...
+                    try:
+                        self._threads.remove(thread)
+                    except ValueError:
```
vstinner: I suggest to happen a comment:

```
# should not happen, but ignoring the error is safe
```

vstinner (dismissed):
Can you try to add a NEWS entry using the tool "blurb"? python3 -m pip install --user blurb, then run blurb in a command line.

Lib/socketserver.py
maru-n: @vstinner 
If my understanding is correct, self._threads can be "None" here.
It will occur when server_close() called before the process_request_thread reached to this statement.
So my propose is keep this statement or don't set "none" to self._threads on server_close(). for example,
```
def server_close(self):
    super().server_close()
    if self.block_on_close and self._threads_lock is not None:
        with self._threads_lock:
            threads = self._threads
            self._threads = []
        for thread in threads:
            thread.join()
```
(In this case, _remove_thread() will rise ValueError when called server_close() called, but it will be ignored.)
How wold you think this?

csabella:
@maru-n, please address the latest review comments by @vstinner.  It looks like there was a lot of activity on this issue, so maybe it can be finalized soon.  Thanks!

pablogsal:
Will try review this soon, feel free to ping me if I have not said anything in a while :)

Lib/socketserver.py
vadmium: I agree there is a window where __threads_ is None. Maru's proposed change to _server_close_ looks okay to me. It simplifies the logic a bit too.

Lib/socketserver.py
vadmium: The lock won't have been initialized if _block_on_close_ was false.

csabella:
Thanks @pablogsal.  Hope this ping isn't too soon.  🙂

pablogsal:
Maybe I am missing something, but there are still some changes requested by @vstinner that are not addressed. @maru-n, cold you check that everything requested is addressed and also request a new review (check https://github.com/python/cpython/pull/13893#issuecomment-505622726)?

csabella:
@maru-n, please make the requested code review changes.  Thank you!

Lib/socketserver.py
jaraco: In the latest commits, this behavior is moved to `_threads.remove_current()`.

Lib/socketserver.py
jaraco: In the latest commit, `_threads.remove_current()` does not hold the lock for daemon threads.

Lib/socketserver.py
jaraco: `_threads.remove()` now suppresses `ValueError`.

Lib/socketserver.py
jaraco: Creation of `self._threads` is now done with `set_default` to make it as atomic as possible without a lock outside the instance.

Lib/socketserver.py
jaraco: These are now created _before_ creating the first thread.

Lib/socketserver.py
jaraco: The lock is now created with the list of threads as a `_Threads` object. The `block_on_close` is used to determine the type of object to create for `_threads` so that it's referenced exactly once and has no effect after the first request is handled.

Lib/socketserver.py
jaraco: In b107b913c17f8eab25be0c1c24edfd25bfbed6a9, I've added a test that captures this behavior (where server_close is called before process_request) and in a1aff7673d947fe63e2f303b1766686602c9e4dc, I've addressed the issue.

Lib/socketserver.py
jaraco: Done.

ned-deily:
I don't have an opinion on the merits of the PR at this point. I think it first needs to be reviewed. merged to master,  and exposed on builtbots before considering backports to other branches.

jaraco:
In 7d1f367941c411ce01c4bf5ec5b020680e7c5313, I've added a test against the ancestor commit to this PR, capturing the reported failure and in dadeb0c71460931ce8c525b77e059587472b3021 merged that commit into the fix, demonstrating its effectiveness.

dmitry-prokopchenkov:
@maru-n , @jaraco Thanks for working on this! I guess we've faced with the issue which could be fixed by this PR. So I'm curious what is the status of this PR? Why is it not merged yet?

jaraco:
I've reviewed it and believe I've addressed the concerns by Victor. Given that there's been no comment after several pings, I recommend to proceed.

pablogsal:
This commit has introduced reference leaks:

---------------------------------------------------------------------
Ran 202 tests in 21.654s
OK (skipped=1)
......
test_logging leaked [20, 20, 20] references, sum=60
test_logging leaked [20, 20, 20] memory blocks, sum=60
2 tests failed again:
    test_logging test_socketserver

Example buildbot failure:

https://buildbot.python.org/all/#/builders/562/builds/79/steps/5/logs/stdio

PR #13902, opened 2019-06-07
bpo-37201: fix test_distutils failures for Windows ARM64

paulmon:
There are a few places where ARM64 is not correctly specified in order for distutils to work for on-target builds using Visual Studio (32-bit x86 emulation).

@zooba 

https://bugs.python.org/issue37201

### DIFF
--- a/Lib/distutils/tests/test_bdist_wininst.py
+++ b/Lib/distutils/tests/test_bdist_wininst.py
@@ -7,7 +7,7 @@ from test.support import run_unittest
 from distutils.command.bdist_wininst import bdist_wininst
 from distutils.tests import support
 
-@unittest.skipIf(sys.platform == 'win32' and platform.machine() == 'ARM64', 
+@unittest.skipIf(sys.platform == 'win32' and platform.machine() == 'ARM64',
     'bdist_wininst is not supported in this install')
 @unittest.skipIf(getattr(bdist_wininst, '_unsupported', False),
     'bdist_wininst is not supported in this install')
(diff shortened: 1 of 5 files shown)


Lib/distutils/tests/test_bdist_wininst.py
```diff
@@ -11,6 +12,8 @@ class BuildWinInstTestCase(support.TempdirManager,
                            support.LoggingSilencer,
                            unittest.TestCase):
 
+    @unittest.skipIf(sys.platform == 'win32' and platform.machine() == 'ARM64', 
+                    "No wininst-14.0-arm64.exe")
```
paulmon: I'm not sure what wininst-*.exe does. 
I'm not sure whether it's better to skip this test, or find the sources for wininst-*.exe, build it, and check the .exe in.

Lib/distutils/tests/test_bdist_wininst.py
```diff
@@ -1,5 +1,6 @@
 """Tests for distutils.command.bdist_wininst."""
 import unittest
+import sys, platform
```
zooba: Separate lines please

Lib/distutils/tests/test_bdist_wininst.py
zooba: No, let's just declare `bdist_wininst` unsupported on ARM. It's for generating self-extracting package installers.

paulmon: If we just declare it unsupported on ARM does that mean no change to this file?
It looks like the stub with _unsupported attribute only gets copied in by layout.  
Which seems to mean that the _unsupported attribute is not set when buildbot runs.
Or am I missing something?

zooba:
"Declaring" it unsupported and "giving users a helpful message when they try to use it on an unsupported platform" are two different things ;)

paulmon:
Thanks!

PR #13906, opened 2019-06-08
[2.7] bpo-37188: Fix a divide-by-zero in arrays of size-0 objects (GH-13881)

eric-wieser:
`itemsize` can be zero when `_type_` is an empty struct or length-0 array.

This line used to be `length * itemsize < 0`, which didn't exhibit this problem.

https://bugs.python.org/issue37188


eric-wieser:
Not done using `cherry_picker` because `pip3.5 install cherry_picker` installs code that does not run on python 3.5.

skrah:
@eric-wieser Thanks, not using cherry_picker is of course fine!

PR #13908, opened 2019-06-08
bpo-18299: extending a kill_on_os_error() in Popen

shihai1991:
Using context manager kill subprocess if a python's exception raised.

https://bugs.python.org/issue18299

### DIFF
--- a/Doc/library/subprocess.rst
+++ b/Doc/library/subprocess.rst
@@ -723,6 +723,11 @@ Instances of the :class:`Popen` class have the following methods:
    On Windows :meth:`kill` is an alias for :meth:`terminate`.
 
 
+.. method:: Popen.kill_on_error()
+
+   Kills the child when a python error raised.
+
+
 The following attributes are also available:
 
 .. attribute:: Popen.args
--- a/Doc/library/subprocess.rst
+++ b/Doc/library/subprocess.rst
@@ -723,9 +723,15 @@ Instances of the :class:`Popen` class have the following methods:
    On Windows :meth:`kill` is an alias for :meth:`terminate`.
 
 
-.. method:: Popen.kill_on_error()
+.. method:: Popen.kill_on_os_error()
 
-   Kills the child when a python error raised.
+   Kills the subprocess when the parent process touch off a race condition and
+   subprocess hanged. One of practical case can be found in:
+   https://bugs.python.org/issue25122::
+
+      proc = subprocess.Popen(...)
+      with proc.kill_on_os_error():
+          proc.communicate()
 
 
 The following attributes are also available:
--- a/Lib/subprocess.py
+++ b/Lib/subprocess.py
@@ -1886,7 +1886,7 @@ class Popen(object):
             if subprocess touch off a race condition.
             One of practical case can be found in:
             https://bugs.python.org/issue25122
-            
+
             For example:
             proc = subprocess.Popen(...)
             with proc.kill_on_os_error():
@@ -1904,7 +1904,7 @@ class Popen(object):
                 try:
                     self.kill()
                 except:
-                    # catch all possible errors. 
+                    # catch all possible errors.
                     pass
                 self.wait()
                 raise
--- a/Lib/subprocess.py
+++ b/Lib/subprocess.py
@@ -1879,3 +1879,25 @@ class Popen(object):
             """Kill the process with SIGKILL
             """
             self.send_signal(signal.SIGKILL)
+ 
+        @contextlib.contextmanager
+        def kill_on_error(self):
+            """Using context manager to kill subprocess if
+            a python exception is raised   
+            """
+            try:
+                yield self
+            except:
+                if self.stdin:
+                    self.stdin.close()
+                if self.stdout:
+                    self.stdout.close()
+                if self.stderr:
+                    self.stderr.close()
+                try:
+                    self.kill()
+                except OSError:
+                    # process already terminated
+                    pass
+                self.wait()
+                raise
--- a/Lib/subprocess.py
+++ b/Lib/subprocess.py
@@ -1879,12 +1879,11 @@ class Popen(object):
             """Kill the process with SIGKILL
             """
             self.send_signal(signal.SIGKILL)
- 
+
         @contextlib.contextmanager
         def kill_on_error(self):
             """Using context manager to kill subprocess if
-            a python exception is raised   
-            """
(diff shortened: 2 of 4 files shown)


asvetlov (changes requested):
Tests for this new feature are required

shihai1991:
> Tests for this new feature are required
Andrew, thanks for review. I will replenish it.

Lib/test/test_subprocess.py
```diff
@@ -2814,6 +2814,11 @@ def test_stopped(self):
 
         self.assertEqual(returncode, -3)
 
+    def test_kill_on_error(self):
```
shihai1991: Looks it only can add positive testcase due to the core code just only `yield self`
If somebody have a good idea about negatigve testcase, pls ping me;)

shihai1991:

Doc/library/subprocess.rst
```diff
@@ -723,6 +723,11 @@ Instances of the :class:`Popen` class have the following methods:
    On Windows :meth:`kill` is an alias for :meth:`terminate`.
 
 
+.. method:: Popen.kill_on_error()
+
+   Kills the child when a python error raised.
```
gpshead: When a python error is raised from where and when?

Lib/subprocess.py
```diff
@@ -1879,3 +1879,24 @@ def kill(self):
             """Kill the process with SIGKILL
             """
             self.send_signal(signal.SIGKILL)
+
+        @contextlib.contextmanager
+        def kill_on_error(self):
+            """Using context manager to kill subprocess if
```
gpshead: How does using a context manager kill the process and why would someone reading this understand that to be what it does?  Does it matter to the docstring reader if a context manager is used?

similar to the documentation comment: if an python exception is raised is non-specific.  raised from where and when?  this could easily be misinterpreted to have something to do with the exit status of the child process or even exceptions raised within it.

"on error" is a very ambiguious method name.  someone reading code calling this won't intuitively understand what it does other than hopefully kill the process.

Lib/subprocess.py
```diff
@@ -1879,3 +1879,24 @@ def kill(self):
...
+                if self.stdin:
+                    self.stdin.close()
+                if self.stdout:
+                    self.stdout.close()
+                if self.stderr:
+                    self.stderr.close()
+                try:
+                    self.kill()
+                except OSError:
+                    # process already terminated
```
gpshead: FYI - There is another reason why self.kill() can fail: PermissionError.  When a setuid process is launched (or potentially other system configurations enforcing process policies) you may get a PermissionError trying to signal that process.

When a PermissionError happens, the wait() call will block until the process dies on its own, if ever.  https://bugs.python.org/issue37091 covers this one.  It isn't clear if we should do anything for this situation.

Lib/test/test_subprocess.py
gpshead: you could create a Popen subclass that overrides either `__enter__` or `__exit__` to raise an exception.  (or do the equivalent using unittest.mock)

gpshead (changes requested):
It is not entirely clear to me what this method is intended to do.  Its name doesn't make that obvious to me, "on error" is ambiguous.

Lib/test/test_subprocess.py
shihai1991: Both of them can not work in here. for the reason:
1) We use contextmanager in the new function and it can not work with `__enter__` or `__exit__`.
2) Looks like `yield proc` can not be mocked to raise a error.

Doc/library/subprocess.rst
shihai1991: Hi, Gregory, thank for your elaborative review:).
one practical case would be found in https://bugs.python.org/issue25122(due to some race condition on FreeBSD) and i updated the subprocess.rst too.

Lib/subprocess.py
shihai1991: Using context manager can use `try...except...` statement to trap the error. I udpated the docstring too.
I use on_os_error to replace on_error.

Lib/subprocess.py
shihai1991: Looks try all possible exceptions or remove all possible exceptions is fine.

shihai1991:

gpshead:
I'm rejecting this PR.  This is not an API we need on the `subprocess.Popen` class.

There may be utility on a context manager that tries to terminate a process when an exception is escaping the context, but that belongs in its own library.  It isn't something users of subprocess need on a regular basis.

The PR references bpo-18299 but I see no relation to that bug.  If something like this would've been used in that bug maybe this belongs in `test.script_helper`.

_Misc review notes for posterity here rather than trying to connect them with parts of the diffs:_

The description in the documentation still makes no sense.  We don't protect anyone from race conditions or hangs.  And the method name is still misleading.  It appears to attempt to kill the process if any exception escapes the context manager followed by waiting for it regardless of kill succeeding or not (that wait may hang if not).  Yet the method name is confusing as a call to Popen.kill() tries to kill the process immediately while a call to Popen.kill_on_os_error() does nothing unless the caller uses it as a context manager.  A better name if this _were_ to be a method at all would be something more along the lines of .get_killing_context.

It _is_ possible to write unittests for this functionality, tests would be required.

shihai1991:
> I'm rejecting this PR. This is not an API we need on the `subprocess.Popen` class.
> 
> There may be utility on a context manager that tries to terminate a process when an exception is escaping the context, but that belongs in its own library. It isn't something users of subprocess need on a regular basis.
> 
> The PR references [bpo-18299](https://bugs.python.org/issue18299) but I see no relation to that bug. If something like this would've been used in that bug maybe this belongs in `test.script_helper`.
> 
> _Misc review notes for posterity here rather than trying to connect them with parts of the diffs:_
> 
> The description in the documentation still makes no sense. We don't protect anyone from race conditions or hangs. And the method name is still misleading. It appears to attempt to kill the process if any exception escapes the context manager followed by waiting for it regardless of kill succeeding or not (that wait may hang if not). Yet the method name is confusing as a call to Popen.kill() tries to kill the process immediately while a call to Popen.kill_on_os_error() does nothing unless the caller uses it as a context manager. A better name if this _were_ to be a method at all would be something more along the lines of .get_killing_context.
> 
> It _is_ possible to write unittests for this functionality, tests would be required.

Copy that.

PR #13932, opened 2019-06-09
bpo-37111: Added encoding keyword parameter to basicConfig.

vsajip:
https://bugs.python.org/issue37111

### DIFF
--- a/Doc/howto/descriptor.rst
+++ b/Doc/howto/descriptor.rst
@@ -145,7 +145,7 @@ print a message for each get or set.  Overriding :meth:`__getattribute__` is
 alternate approach that could do this for every attribute.  However, this
 descriptor is useful for monitoring just a few chosen attributes::
 
-    class RevealAccess(object):
+    class RevealAccess:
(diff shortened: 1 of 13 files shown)


vsajip:
@zooba Please review just to ensure I haven't missed anything ... thanks.

zooba:
@vsajip We normally add `encoding` and `errors` parameters as pairs (at least, this is what I was asked to do when adding them to `subprocess`)

zooba:
To be clear, I mean you should add an `errors` parameter as well. But the change overall looks good 👍

vsajip:
Whoops, I made a mistake by doing a rebase - should I just kill this PR and start again?

zooba:
@vsajip Yeah, looks like you undid some other changes, so probably best to restart.

FWIW, once you submit a PR and people start looking it at, you're best to just do regular merges and we'll squash it all at the end. That way you don't break the review workflow.

vsajip:
That's what I normally do, I just had a brain-fade this time :-(

PR #13935, opened 2019-06-10
[3.8] bpo-37206: fix docstring of dict.pop()

methane:
Stop using Argument Clinic for dict.pop().
It reuses code generated by clinic, but docstring is reverted.

https://bugs.python.org/issue37206


ned-deily:
(See discussion on the bug report.)

PR #13937, opened 2019-06-10
[2.7] bpo-36742: Fix urlparse.urlsplit() error message for Unicode URL

vstinner:
If urlparse.urlsplit() detects an invalid netloc according to NFKC
normalization, the error message type is now str rather than unicode,
and use repr() to format the URL, to prevent <exception str() failed>
when display the error message.

### DIFF
--- a/Lib/test/test_urlparse.py
+++ b/Lib/test/test_urlparse.py
@@ -649,7 +649,7 @@ class UrlParseTestCase(unittest.TestCase):
         self.assertEqual(str(cm.exception),
                          "netloc u'\u30d7:80' contains invalid characters "
                          "under NFKC normalization")
-        self.assertIsInstance(cm.exception.message, str)
+        self.assertIsInstance(cm.exception.args[0], str)
 
         for scheme in [u"http", u"https", u"ftp"]:
             for netloc in [u"netloc{}false.netloc", u"n{}user@netloc"]:
--- a/Lib/test/test_urlparse.py
+++ b/Lib/test/test_urlparse.py
@@ -644,12 +644,8 @@ class UrlParseTestCase(unittest.TestCase):
         # bpo-36742: Verify port separators are ignored when they
         # existed prior to decomposition
         urlparse.urlsplit(u'http://\u30d5\u309a:80')
-        with self.assertRaises(ValueError) as cm:
+        with self.assertRaises(ValueError):
             urlparse.urlsplit(u'http://\u30d5\u309a\ufe1380')
-        self.assertEqual(str(cm.exception),
-                         "netloc u'\u30d7:80' contains invalid characters "
-                         "under NFKC normalization")
-        self.assertIsInstance(cm.exception.args[0], str)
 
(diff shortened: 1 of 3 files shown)


vstinner:
This change fix the bug described at: https://bugs.python.org/issue36742#msg344981

Script:

```
import urlparse as urllib_parse
url = u'example.com\uFF03@bing.com'
url = u'http://\u30d5\u309a\ufe1380'
urllib_parse.urlsplit(url)
```

Output without the fix:

```
ValueError: <exception str() failed>
```

Output with the fix:

```
ValueError: netloc u'\u30d7:80' contains invalid characters under NFKC normalization
```

vstinner:
cc @stratakis @encukou

Lib/test/test_urlparse.py
```diff
@@ -644,8 +644,12 @@ def test_urlsplit_normalization(self):
         # bpo-36742: Verify port separators are ignored when they
         # existed prior to decomposition
         urlparse.urlsplit(u'http://\u30d5\u309a:80')
-        with self.assertRaises(ValueError):
+        with self.assertRaises(ValueError) as cm:
             urlparse.urlsplit(u'http://\u30d5\u309a\ufe1380')
+        self.assertEqual(str(cm.exception),
+                         "netloc u'\u30d7:80' contains invalid characters "
+                         "under NFKC normalization")
+        self.assertIsInstance(cm.exception.message, str)
```
tirkarthi: There is a `DeprecationWarning` in this line in CI https://travis-ci.org/python/cpython/jobs/543658580#L1921

```
/home/travis/build/python/cpython/Lib/test/test_urlparse.py:652: DeprecationWarning: BaseException.message has been deprecated as of Python 2.6
  self.assertIsInstance(cm.exception.message, str)
```

orsenthil (approved):
lgtm.

Lib/test/test_urlparse.py
vstinner: fixed

vstinner:
cc @tirkarthi @zooba @gpshead

Lib/urlparse.py
```diff
@@ -180,8 +180,9 @@ def _checknetloc(netloc):
         return
     for c in '/?#@:':
         if c in netloc2:
-            raise ValueError(u"netloc '" + netloc + u"' contains invalid " +
-                             u"characters under NFKC normalization")
+            raise ValueError("netloc %r contains invalid characters "
+                             "under NFKC normalization"
+                             % netloc2)
```
tirkarthi: Should `netloc` be reported in the error message? Python 3 and previous error messages report `netloc`.

vstinner: Oh. You're right. I also fixed this bug and I updated the unit test to test this case.

tirkarthi (approved):
LGTM. Thanks.

orsenthil (approved):

vstinner:
Thanks for review @tirkarthi and @orsenthil!

PR #13938, opened 2019-06-10
[WIP] bpo-37214: Add a new EncodingWarning warning category

vstinner:
open() now emits a EncodingWarning if called with encoding=None.

https://bugs.python.org/issue37214


vstinner:
@methane: What do you think of this approach? Does it make any sense to have an opt-in warning?

methane:
> @methane: What do you think of this approach? Does it make any sense to have an opt-in warning?

No.  Warning everytime when encoding is omitted was @zooba 's suggestion.  But he opposite to change default encoding strongly.
Without planned breaking change, adding warning which everyone ignore it doesn't make sense to me.

In PEP 597 (2nd), I am proposing adding more option.  macOS user can test their code easily with `PYTHONTEXTENCODING=ascii`.  They can notice if their code is depending on default encoding is UTF-8.

Anyway, I'm thinking about advertising your UTF-8 mode.
Windows user can run code which assumes default text encoding is UTF-8.

vstinner:
Ok. I may keep this PR open to show that always emit a warning can be a bad idea :-)

vstinner:
I wrote this PR to discuss https://www.python.org/dev/peps/pep-0597/ but it's unrelated to the implementation of this PR :-)

vstinner:
I abandon this PR.

PR #13939, opened 2019-06-10
bpo-37215: Fix compilation with dtrace

kulikjak:
After the integration of https://bugs.python.org/issue36842, build with dtrace is broken on systems where files that reference DTrace probes need to be modified in-place by dtrace (e.g., Solaris).

The reason for that is that Python/sysmodule.o, which newly contains some dtrace references, was not added to DTRACE_DEPS.

https://bugs.python.org/issue37215


kulikjak:
The problem was fixed by another pull request (#13940) so I am closing this one...

PR #13944, opened 2019-06-10
bpo-26806: add 30 to the recursion limit in IDLE's shell

taleinat:
This is done to compensate for the extra stack frames added by IDLE's shell, which can cause problems when setting the recursion limit to low values.

This wraps `sys.setrecursionlimit()` and `sys.getrecursionlimit()`, adding a note about this to their doc-strings.

https://bugs.python.org/issue26806


taleinat:
@terryjreedy, I couldn't figure out where you'd like this mentioned in which docs. I'm happy to add appropriate wording if you point me in the right direction.

Lib/idlelib/run.py
```diff
@@ -474,6 +475,44 @@ def __init__(self, rpchandler):
...
+
+            @wraps(sys.getrecursionlimit)
+            def getrecursionlimit():
+                return getrecursionlimit.__wrapped__() - 25
+            sys.getrecursionlimit = getrecursionlimit
+
+            del sys
+            del wraps
+            """)
+        exec(code, self.locals)
```
terryjreedy: I believe that this code should be executed in the run module on import, NOT in the main module.  Calling a module function from main, as I suggested on the issue, should do that.  Then the deletions are not needed.

terryjreedy (changes requested):
Also see new comment on the issue, which includes some manual tests.

terryjreedy:
Our messages crossed.  I worry about the docstring and doc and blurb when we agree on everything else.  I considered the PRs you and Cheryl worked on last week a higher priority.

taleinat:

taleinat:
Ping, @terryjreedy. I'd like to get this into 3.8.0.

BTW, should we backport this to 3.7.4?

Lib/idlelib/run.py
terryjreedy: Done, except that call to install_recursionlimit_wrappers is in MyHandler.handle rather than main().  I presume the rationale is that 4 other sys attributes are replaced here.  Good enough.

Lib/idlelib/run.py
```diff
@@ -303,6 +305,60 @@ def fix_scaling(root):
...
+    def setrecursionlimit(*args, **kwargs):
+        # mimic the original sys.setrecursionlimit()'s input handling
+        if kwargs:
+            raise TypeError(
+                "setrecursionlimit() takes no keyword arguments")
+        try:
+            limit, = args
+        except ValueError:
+            raise TypeError("setrecursionlimit() takes exactly one "
+                            "argument ({} given)".format(len(args)))
```
terryjreedy: Use f-string here also. 
<DONE, and added bad call tests>

Messages verified.  It is somewhat annoying that C and python functions with same signature (here, 'limit, /') give different error messages for the same error.  (A separate issue for the tracker.)  So we cannot, even for 3.8+, when '/' is valid, match behavior by just giving the proper signature.

Lib/idlelib/run.py
```diff
@@ -303,6 +305,60 @@ def fix_scaling(root):
...
+
+    # add the delta to the default recursion limit, to compensate
+    sys.setrecursionlimit(sys.getrecursionlimit() + RECURSIONLIMIT_DELTA)
+
+    sys.setrecursionlimit = setrecursionlimit
+    sys.getrecursionlimit = getrecursionlimit
+
+
+def uninstall_recursionlimit_wrappers():
+    """Uninstall the recursion limit wrappers from the sys module."""
```
terryjreedy: This is (only) needed for tests.  Add line "# Needed for tests." so function is never deleted as 'unused'.  (I have made similar notes for other test-only code.)
<Done.>

Misc/NEWS.d/next/IDLE/2019-06-10-22-48-50.bpo-26806.Zltkum.rst
```diff
@@ -0,0 +1,2 @@
+Always add 30 to the recursion limit in IDLE's shell to compensate for
+stack frames added by IDLE.
```
terryjreedy: Suggestion:
"To compensate for stack frames added by IDLE and avoid possible problems with low recursion limits, add 30 to limits in the user code execution process.  Subtract 30 when reporting recursion limits to make this addition mostly transparent."
<Done.>
Note: we are not doing the same in the IDLE process when -n (deprecated) is passed.

terryjreedy (commented):
When I initially pulled this PR and ran test_idle, I got 4 failures.  They disappeared when I merged in current master.  We need to remember to do this before final testing.

I have made the minor changes I suggested and added a test method.  I will add the idlelib news item and probably a doc change and then push everything for your final approval.

terryjreedy (approved):
Please merge if changes seem OK.

Lib/idlelib/run.py
taleinat: Why would it be deleted if it is used in test code? Even if it were deleted, the tests would immediately break!

Also, there's no telling who might find other uses for this in the future.

Doc/library/idle.rst
```diff
@@ -713,6 +713,10 @@ or ``print`` or ``write`` to sys.stdout or sys.stderr,
 IDLE should be started in a command line window.  The secondary subprocess
 will then be attached to that window for input and output.
 
+The IDLE code running in the execution process adds frames to the call stack
+that would not be there otherwise.  IDLE wraps ``sys.getrecursionlimit`` and
+``sys.setrecursionlimit`` to reduce their visibility.
```
taleinat: "their" is ambiguous here; I suggest "to reduce the visibility of the additional stack frames."

Also, perhaps we should mention that this is not just about visibility? E.g. "to reduce the effect of the additional stack frames".

taleinat:
@terryjreedy, see my two comments on your recent changes.

Lib/idlelib/run.py
terryjreedy: Let me start over.  I initially wasted time looking for where this was used in run.py.   I like to mark test code (code we/IDLE only use as such).  So I moved the comment, reworded, into the docstring *and* added a note that users could also use this.  (Yeah, one of use might want to do so sometime.)  I think that this is an improvement.

terryjreedy:
I have made changes (improvements) in response to both comments.  I'd like to merge this and move on to other issues.

taleinat:
Looks good to me, just one minor typo. I'll fix it and merge later today
even I'm at a computer.

terryjreedy:
> 2.7.4?
This will be in 2.7.5.  I don't think a cherry-pick request would be worth the cost.

taleinat:
> 2.7.4?

No, I did really mean to ask about 3.7.4, but we've already backported to 3.7, so I'll take that as a yes :)

> This will be in 2.7.5.  I don't think a cherry-pick request would be worth the cost.

I agree.

ZackerySpytz:
@taleinat @terryjreedy It seems that IDLE fails to start after fcf1d003bf4f0100c9d0921ff3d70e1127ca1b71 when CPython is configured without docstrings (`./configure --without-doc-strings`). When I run `./python -m idlelib.idle`, I see the following:

```

----------------------------------------
Unhandled server exception!
Thread: SockThread
Client Address:  ('127.0.0.1', 40895)
Request:  <socket.socket fd=3, family=AddressFamily.AF_INET, type=SocketKind.SOCK_STREAM, proto=0, laddr=('127.0.0.1', 40832), raddr=('127.0.0.1', 40895)>
  File "/home/lubuntu2/cpython/Lib/socketserver.py", line 316, in _handle_request_noblock
    self.process_request(request, client_address)
  File "/home/lubuntu2/cpython/Lib/socketserver.py", line 347, in process_request
    self.finish_request(request, client_address)
  File "/home/lubuntu2/cpython/Lib/socketserver.py", line 360, in finish_request
    self.RequestHandlerClass(request, client_address, self)
  File "/home/lubuntu2/cpython/Lib/idlelib/rpc.py", line 514, in __init__
    socketserver.BaseRequestHandler.__init__(self, sock, addr, svr)
  File "/home/lubuntu2/cpython/Lib/socketserver.py", line 720, in __init__
    self.handle()
  File "/home/lubuntu2/cpython/Lib/idlelib/run.py", line 511, in handle
    install_recursionlimit_wrappers()
  File "/home/lubuntu2/cpython/Lib/idlelib/run.py", line 332, in install_recursionlimit_wrappers
    setrecursionlimit.__doc__ += "\n\n" + textwrap.fill(textwrap.dedent(f"""\
TypeError: unsupported operand type(s) for +=: 'NoneType' and 'str'

*** Unrecoverable, server exiting!
----------------------------------------
  File "/home/lubuntu2/cpython/Lib/idlelib/rpc.py", line 432, in pollresponse
    message = self.pollmessage(wait)
  File "/home/lubuntu2/cpython/Lib/idlelib/rpc.py", line 384, in pollmessage
    packet = self.pollpacket(wait)
  File "/home/lubuntu2/cpython/Lib/idlelib/rpc.py", line 363, in pollpacket
    raise EOFError
EOFError

  File "/home/lubuntu2/cpython/Lib/runpy.py", line 192, in _run_module_as_main
    return _run_code(code, main_globals, None,
  File "/home/lubuntu2/cpython/Lib/runpy.py", line 85, in _run_code
    exec(code, run_globals)
  File "/home/lubuntu2/cpython/Lib/idlelib/idle.py", line 14, in <module>
    main()
  File "/home/lubuntu2/cpython/Lib/idlelib/pyshell.py", line 1521, in main
    shell = flist.open_shell()
  File "/home/lubuntu2/cpython/Lib/idlelib/pyshell.py", line 330, in open_shell
    if not self.pyshell.begin():
  File "/home/lubuntu2/cpython/Lib/idlelib/pyshell.py", line 1050, in begin
    client = self.interp.start_subprocess()
  File "/home/lubuntu2/cpython/Lib/idlelib/pyshell.py", line 463, in start_subprocess
    self.transfer_path(with_cwd=True)
  File "/home/lubuntu2/cpython/Lib/idlelib/pyshell.py", line 553, in transfer_path
    self.runcommand("""if 1:
  File "/home/lubuntu2/cpython/Lib/idlelib/pyshell.py", line 756, in runcommand
    self.rpcclt.remotequeue("exec", "runcode", (code,), {})
  File "/home/lubuntu2/cpython/Lib/idlelib/rpc.py", line 224, in remotequeue
    return self.asyncreturn(seq)
  File "/home/lubuntu2/cpython/Lib/idlelib/rpc.py", line 248, in asyncreturn
    response = self.getresponse(seq, wait=0.05)
  File "/home/lubuntu2/cpython/Lib/idlelib/rpc.py", line 291, in getresponse
    response = self._getresponse(myseq, wait)
  File "/home/lubuntu2/cpython/Lib/idlelib/rpc.py", line 311, in _getresponse
    response = self.pollresponse(myseq, wait)
  File "/home/lubuntu2/cpython/Lib/idlelib/rpc.py", line 436, in pollresponse
    self.handle_EOF()
  File "/home/lubuntu2/cpython/Lib/idlelib/pyshell.py", line 388, in handle_EOF
    raise EOFError
EOFError
```

PR #13946, opened 2019-06-10
[WIP] bpo-36402: Fix test_threading.test_threads_join_2()

vstinner:
Fix test_threading.test_threads_join_2(): run the test into a
subprocess, rather than in a subinterpreter, to avoid a race
condition.


vstinner:
I'm not sure that this fix is correct. The purpose of the test is to ensure that Py_EndInterpreter() joins properly threads: commit 7b4769937fb612d576b6829c3b834f3dd31752f1.

vstinner:
Nah, this change is plain wrong. https://bugs.python.org/issue36402 is a real bug, this change only hides it. PR #13948 is the good fix.

PR #13948, opened 2019-06-10
bpo-36402: Fix threading._shutdown() race condition

vstinner:
Fix a race condition at Python shutdown when waiting for threads.
Wait until the Python thread state of threads get deleted, rather
than just wait until the Python thread completes.

* Add threading._shutdown_locks set: set of Thread._tstate_lock locks
  of non-daemon threads used by _shutdown() wait until all Python
  thread states get deleted (see Thread._set_tstate_lock()).
* Add also threading._shutdown_locks_lock to protect access to
  threading._shutdown_locks.

### DIFF
--- a/Lib/test/test_threading.py
+++ b/Lib/test/test_threading.py
@@ -583,6 +583,41 @@ class ThreadTests(BaseTestCase):
         self.assertEqual(data.splitlines(),
                          ["GC: True True True"] * 2)
 
+    def test_finalization_shutdown(self):
+        # bpo-36402: Py_Finalize() calls threading._shutdown() which must wait
+        # until Python thread states of all non-daemon threads get deleted.
+        #
+        # Test similar to SubinterpThreadingTests.test_threads_join_2(), but
+        # test the finalization of the main interpreter.
+        code = """if 1:
+            import os
+            import threading
+            import time
+            import random
+
+            def random_sleep():
+                seconds = random.random() * 0.010
+                time.sleep(seconds)
+
+            class Sleeper:
+                def __del__(self):
+                    random_sleep()
+
+            tls = threading.local()
+
+            def f():
+                # Sleep a bit so that the thread is still running when
+                # Py_Finalize() is called.
+                random_sleep()
+                tls.x = Sleeper()
+                random_sleep()
+
+            threading.Thread(target=f).start()
+            random_sleep()
+        """
+        rc, out, err = assert_python_ok("-c", code)
+        self.assertEqual(err, b"")
+
     def test_tstate_lock(self):
         # Test an implementation detail of Thread objects.
         started = _thread.allocate_lock()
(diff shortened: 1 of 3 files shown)


vstinner:
This PR changes critical code, so I would appreciate multiple reviews :-)

---

test_threading: test_threads_join_2() was added by commit 7b4769937fb612d576b6829c3b834f3dd31752f1 in 2013, but the test failed randomly since it was added. It's just that failures were ignored until I created https://bugs.python.org/issue36402 last March.

In fact, when the test failed randomly on buildbot (with tests run in parallel), it was fine since test_threading was re-run alone and then the test passed. The buildbot build was seen overall as a success.

The test shows the bug using subinterpreters (Py_EndInterpreter), but I'm quite sure that Py_Finalize() has the same race condition since it also calls threading._shutdown().

pitrou (approved):
Nice find, and the fix looks correct. +1.

Lib/threading.py
```diff
@@ -1398,6 +1417,10 @@ def _after_fork():
     new_active = {}
     current = current_thread()
     _main_thread = current
+
+    with _shutdown_locks_lock:
```
gpshead: We could have forked while this lock was held. You'll need to reset it above. There can be no other threads running in our new process yet.

gpshead (changes requested):
I haven't entirely digested this change yet. It looks good but I don't know if it is enough, I'll need to dig further to convince myself of that. Meanwhile, one issue noted.

Lib/threading.py
vstinner: Oops. You're right. It's now fixed.

vstinner:
> I haven't entirely digested this change yet.

Sadly, these things are badly documented. The main issue is that there are 2 state changes:

* theading.Thread._bootstrap_inner() adds the thread into _active and then removes it from _active
* theading.Thread._set_sentinel() creates a lock registers a callback which release the lock, so .join() can wait until the internal Python thread state is released for real

Currently, _shutdown() rely on _active. But for Python finalization: Py_Finalize() and Py_EndInterpreter(), we really want to wait until the Python thread state get delete, not only when theading.Thread._bootstrap_inner() exits.

test_threads_join_2() demonstrates the bug: "Fatal Python error: Py_EndInterpreter: not the last thread".

Lib/test/test_threading.py
```diff
@@ -583,6 +583,41 @@ def __del__(self):
...
+
+            class Sleeper:
+                def __del__(self):
+                    random_sleep()
+
+            tls = threading.local()
+
+            def f():
+                # Sleep a bit so that the thread is still running when
+                # Py_EndInterpreter is called.
```
pablogsal: Should it be `PyFinalize` is called?

vstinner: fixed

pablogsal (approved):
LGTM

I checked locally also with the patch in https://bugs.python.org/file48409/py_finalize.patch to test `test_finalization_shutdown` manually.

Good catch! :tada:

vstinner:
@gpshead: Would yo mind to review the updated PR?

vstinner:
I rebased my PR and squashed commits. I clarified comments and the NEWS enty to specify that only non-daemon threads are joined. I reviewed my own PR and it LGTM :-) Since it already two approvals, I'm now comfortable enough to merge it.

wmanley:
This caused a regression for people overriding `Thread.join` to implement custom thread interruption.  See https://stackoverflow.com/questions/50486083/ending-non-daemon-threads-when-shutting-down-an-interactive-python-session

vstinner:
> This caused a regression for people overriding Thread.join to implement custom thread interruption. See https://stackoverflow.com/questions/50486083/ending-non-daemon-threads-when-shutting-down-an-interactive-python-session

I guess that you are talking about this regression: https://bugs.python.org/issue37788

PR #13949, opened 2019-06-10
bpo-36402: Add test_threading.test_finalization_shutdown()

vstinner:


vstinner:
Test added to PR #13948.

PR #13955, opened 2019-06-11
bpo-36918, test_urllib: Silence FakeHTTPConnection destructor errors

vstinner:
Implement FakeHTTPConnection.close() to silence destructor errors.

Co-Authored-By: Xtreak <tir.karthi@gmail.com>

https://bugs.python.org/issue36918


vstinner:
Oh. It was too easy: test_urllib2 fails with my change :-(

tirkarthi:
Yes, this was my first approach as well but `test_urllib2` explicitly checks Fakesocket is closed.

vstinner:
This PR is wrong. I wrote PR #13996 which is correct.

PR #13957, opened 2019-06-11
bpo-37220: Fix 2.7 test -R test_IDLE crash on Windows.

terryjreedy:
Patch might make test more stable on 3.x also.

https://bugs.python.org/issue37220

### DIFF
--- a/Lib/idlelib/idle_test/test_searchbase.py
+++ b/Lib/idlelib/idle_test/test_searchbase.py
@@ -48,7 +48,6 @@ class SearchDialogBaseTest(unittest.TestCase):
         self.dialog.default_command = None
 
         toplevel = Toplevel(self.root)
-        self.addCleanup(toplevel.destroy)
         text = Text(toplevel)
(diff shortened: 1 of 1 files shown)


terryjreedy:
The manual test is that `test -R3:3 -ugui test_idle` continues to pass.

Lib/idlelib/idle_test/test_searchbase.py
```diff
@@ -48,7 +48,6 @@ def test_open_and_close(self):
         self.dialog.default_command = None
 
         toplevel = Toplevel(self.root)
-        self.addCleanup(toplevel.destroy)
```
taleinat, line 51: I suggest:
```python
def cleanup():
    toplevel.update()
    toplevel.destroy()
self.addCleanup(cleanup)
```

terryjreedy: See my comment on the cli issue.

PR #13958, opened 2019-06-11
[2.7] bpo-37220: Fix 2.7 test -R test_IDLE failure on Windows

terryjreedy:
Cherry-picked from 66d47da.

https://bugs.python.org/issue37220

### DIFF
--- a/Lib/idlelib/idle_test/test_searchdialogbase.py
+++ b/Lib/idlelib/idle_test/test_searchdialogbase.py
@@ -46,7 +46,6 @@ class SearchDialogBaseTest(unittest.TestCase):
         self.dialog.default_command = None
 
         toplevel = Toplevel(self.root)
-        self.addCleanup(toplevel.destroy)
         text = Text(toplevel)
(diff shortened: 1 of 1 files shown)


terryjreedy:
The manual test is that `test -R3:3 -ugui test_idle` passes (on Windows) or continues to pass (where it already did).

terryjreedy:
Before merging, need to add 'Cherry-picked from ##### where ##### will be merge hash for #13957.

Lib/idlelib/idle_test/test_searchdialogbase.py
```diff
@@ -46,7 +46,6 @@ def test_open_and_close(self):
         self.dialog.default_command = None
 
         toplevel = Toplevel(self.root)
-        self.addCleanup(toplevel.destroy)
```
taleinat, line 49: I suggest the following:

```python
def cleanup():
    toplevel.update()
    toplevel.destroy()
self.addCleanup(cleanup)
```

PR #13962, opened 2019-06-11
bpo-37216: Fix filename in mac using document

MakDon:
The python folder name is wrong in mac using document. 

$ ls /Applications | grep Python
Python 2.7
Python 3.5
Python 3.6
Python 3.7

And my system version is macOS Mojave Version 10.14.5(18F132)
$uname -v
Darwin Kernel Version 18.6.0: Thu Apr 25 23:16:27 PDT 2019; root:xnu-4903.261.4~2/RELEASE_X86_64

This commit should be backport to 2.7, 3.5, 3.6. 
DO NOT backport to 3.7 and 3.8 because i  create another PR to fix the filename and version in branch 3.7 and 3.8.
See: 
[#13963 for 3.7](https://github.com/python/cpython/pull/13963), 
[#13964 for 3.8](https://github.com/python/cpython/pull/13964)
[#13966 for 3.9](https://github.com/python/cpython/pull/13966)

https://bugs.python.org/issue37216

PR #13966, opened 2019-06-11
bpo-37216: Fix version and filename in Mac using document

MakDon:
FIx folder name and version for mac using document on branch 3.9(master branch now).
See: [#13962](https://github.com/python/cpython/pull/13962)
WARNING: merge [#13962](https://github.com/python/cpython/pull/13962) first, and then merge this PR after resolving the conflict, or the version would be overwritten by [#13962](https://github.com/python/cpython/pull/13962). Just FYI. :D 

The title of this PR begins with [3.9] because this PR is used to update the version on document.

https://bugs.python.org/issue37216

### DIFF
--- a/Doc/using/mac.rst
+++ b/Doc/using/mac.rst
@@ -25,7 +25,7 @@ there.
 
 What you get after installing is a number of things:
 
-* A :file:`MacPython 3.6` folder in your :file:`Applications` folder. In here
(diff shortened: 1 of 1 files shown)


mangrisano (approved):

ned-deily (approved):
LGTM.  Thanks!

PR #13973, opened 2019-06-11
bpo-37231: optimize calls of special methods

jdemeyer:
https://bugs.python.org/issue37231


methane (approved):
LGTM.  Have you run some microbench?
If there is measurable speedup, please write it in NEWS entry.

jdemeyer:
> LGTM. Have you run some microbench?

Yes. With this code
```python
class X:
    def __mul__(self, other):
        pass

x = X()
```
and benchmarking `x * x` I get a speedup from
```
Mean +- std dev: 103 ns +- 1 ns
```
to
```
Mean +- std dev: 98.6 ns +- 2.2 ns
```

> If there is measurable speedup, please write it in NEWS entry.

Done.

jdemeyer:
I'm thinking that this `vectorcall_method` is useful enough to become official API. But that will be for a different PR.

PR #13977, opened 2019-06-11
closes bpo-33758: Skip test_get_type_hints_modules_forwardref.

benjaminp:
This test "works" if things are run in the right order, so it's better to use @skip than @expectedFailure here.

https://bugs.python.org/issue33758

### DIFF
--- a/Lib/test/test_typing.py
+++ b/Lib/test/test_typing.py
@@ -3,7 +3,7 @@ import collections
 import pickle
 import re
 import sys
-from unittest import TestCase, main, skipUnless, SkipTest, expectedFailure
+from unittest import TestCase, main, skipUnless, SkipTest, skip
 from copy import copy, deepcopy
 
 from typing import Any, NoReturn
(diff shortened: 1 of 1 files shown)


ilevkivskyi (approved):
Thanks!

Lib/test/test_typing.py
```diff
@@ -2654,7 +2654,7 @@ def test_get_type_hints_modules(self):
         self.assertEqual(gth(ann_module2), {})
         self.assertEqual(gth(ann_module3), {})
 
-    @expectedFailure
+    @skip("known bug")
```
tirkarthi, line 2657:
```suggestion
    @skip("known bug : issue #33758")
```

Can the issue number be added as part of this message ?

ilevkivskyi: The test body below has a better explanation, while the issue is more about the "expected failure" being flaky.

benjaminp: That doesn't make sense because that's not the issue to fix the bug exposed by the test.

tirkarthi: Got it, sorry.

PR #13983, opened 2019-06-11
bpo-37236: pragma optimize off for _Py_c_quot on Windows arm64

paulmon:
There is a compiler optimization error on Windows ARM64 that causes test_truediv (test.test_complex.ComplexTest) to fail with ZeroDivisionError: complex division by zero.

Adding a pragma optimize around the affected function fixes the issue.  I am also submitting a report to the compiler team

@zooba 

https://bugs.python.org/issue37236

### DIFF
--- a/Objects/complexobject.c
+++ b/Objects/complexobject.c
@@ -55,6 +55,7 @@ _Py_c_prod(Py_complex a, Py_complex b)
     return r;
 }
 
+/* Avoid bad optimization on Windows ARM64 until the compiler is fixed */
 #ifdef _M_ARM64
 #pragma optimize("", off)
 #endif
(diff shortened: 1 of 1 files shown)


Objects/complexobject.c
```diff
@@ -55,6 +55,9 @@ _Py_c_prod(Py_complex a, Py_complex b)
     return r;
 }
 
+#ifdef _M_ARM64
```
zooba, line 59:
```suggestion
/* Avoid bad optimization on Windows ARM64 until the compiler is fixed */
#ifdef _M_ARM64
```

paulmon:
I merged the suggested comment

paulmon:
Thanks!

PR #13984, opened 2019-06-11
bpo-35766: Whats new in the ast and typing modules

gvanrossum:
Lumping these together because they both come from the static type checking world.


vstinner (approved):
LGTM.

I looked for "3.8" in https://docs.python.org/dev/library/ast.html and https://docs.python.org/dev/library/typing.html and it seems like all changes are documented.

ilevkivskyi (approved):
Looks great!

PR #13986, opened 2019-06-11
gh-70368: Fix multiple registration of ForkAwareLocal atfork cleaner

orivej:
`threading.local` instance `__init__` method is called multiple times, once per
thread where the instance is used, but ForkAwareLocal atfork handler needs to be
registered just once when its instance is created. If it is registered many
times, the registrations bloat `_afterfork_registry` as long as the instance is
alive, and can exhaust all memory.

https://bugs.python.org/issue26180

* Issue: gh-70368

### DIFF
--- a/Lib/multiprocessing/util.py
+++ b/Lib/multiprocessing/util.py
@@ -348,8 +348,11 @@ class ForkAwareThreadLock(object):
 
 
 class ForkAwareLocal(threading.local):
-    def __init__(self):
-        register_after_fork(self, lambda obj : obj.__dict__.clear())
+    def __new__(cls):
+        self = threading.local.__new__(cls)
+        register_after_fork(self, lambda obj: obj.__dict__.clear())
+        return self
+
     def __reduce__(self):
         return type(self), ()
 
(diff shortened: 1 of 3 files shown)


asvetlov:
A test for the change would be awesome

orivej:
I have added a test `TestForkAwareLocal` modeled (and placed) after `TestForkAwareThreadLock`.

orivej:
Added a NEWS blurb. Feel free to edit or squash commits as you see fit.

jdemeyer (approved):
The change and test look good to me. The only thing that I'm not sure about is whether this is the best fix for the problem: as I mentioned on bpo-26810 I wonder if we could ensure that `__init__` is not called multiple times.

orivej:
`__init__` is called multiple times by design of `threading.local`: https://github.com/python/cpython/blob/v3.7.3/Lib/_threading_local.py#L65-L67 . Since https://docs.python.org/3.7/library/threading.html#threading.local refers to this docstring, I guess this is a documented behaviour and changing it would be breaking.

jdemeyer:
OK, you convinced me.

orivej:
What shall happen next?

jdemeyer:
We need a core developer to review your PR.

orivej:
Monthly ping.

PR #13991, opened 2019-06-11
bpo-29505: Fuzz json module, enforce size limit on int(x) fuzz

ammaraskar:
Since CPython seems to be successfully building and being fuzzed in OSS-Fuzz, it's time to add some more interesting fuzzing targets. This PR does the following:

* Add a size limit on the `int(x)` fuzz target. Since python supports arbitrarily large integers, there is a rather
  boring case that is essentially `int("9" * 100000)` causing timeouts in this path.
* Submit `json.loads(x)` for fuzzing.

---
Potential reviewers: @gpshead @alex (skip-news label can be applied)


gpshead (approved):
nice, let the fuzzy fun begin!

PR #13994, opened 2019-06-12
bpo-35766: compile(): rename feature_version parameter

vstinner:
Rename compile() feature_version parameter to _feature_version and
convert it to a keyword-only parameter.

Update also test_type_comments to pass feature_version as a tuple.

### DIFF
--- a/Lib/ast.py
+++ b/Lib/ast.py
@@ -45,7 +45,7 @@ def parse(source, filename='<unknown>', mode='exec', *,
         feature_version = -1
     # Else it should be an int giving the minor version for 3.x.
     return compile(source, filename, mode, flags,
-                   feature_version=feature_version)
+                   _feature_version=feature_version)
 
 
 def literal_eval(node_or_string):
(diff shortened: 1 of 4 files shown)


vstinner:
I proposed this change to make the further changes that I proposed in PR #13992. The PR #13992 introduced an inconsistency between ast.parse() and compile() API for feature_version. So I propose to make compile() feature_version private. Anyway, the parameter only makes sense for ast.parse() anyway, and it's undocumented on purpose.

I skipped the NEWS entry since feature_version is not documented on purpose:
https://docs.python.org/dev/library/functions.html#compile

ilevkivskyi (approved):
Again, as I said I am fine with this.

vstinner:
Note: I saw that importlib has a source_to_code() which has a keyword-only "private" parameter: parameter called "_optimize", and the "_" prefix in Python stands for "private thing". I reused the idea here ;-)

We might make the parameter public later if needed, but I would prefer to accept the feature_version=(3, 6) format in that case. Right now, it's not needed, so I just made it private ;-)

PR #13999, opened 2019-06-12
bpo-37223: io.BufferedWrite.close() don't call flush() if already closed

vstinner:
The close() method of io buffered files no longer call flush()
if the file is already closed.

https://bugs.python.org/issue37223

### DIFF
--- a/Lib/_pyio.py
+++ b/Lib/_pyio.py
@@ -1279,7 +1279,7 @@ class BufferedWriter(_BufferedIOMixin):
             self._flush_unlocked()
 
     def _flush_unlocked(self):
-        if self.closed:
+        if self.closed or self._write_buf is None:
             raise ValueError("flush on closed file")
         while self._write_buf:
             try:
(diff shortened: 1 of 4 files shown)


vstinner:
This PR fix the following 2 errors when running test_io:

```
test_writer_close_error_on_close (test.test_io.CBufferedRWPairTest) ... Exception ignored in: <_io.BufferedWriter>
  File "/home/vstinner/prog/python/master/Lib/unittest/case.py", line 611, in _callTestMethod
    method()
ValueError: flush of closed file
ok

test_destructor (test.test_io.CBufferedRandomTest) ... Exception ignored in: <_io.BufferedRWPair object at 0x7fc9d1c230b0>
  File "/home/vstinner/prog/python/master/Lib/test/support/__init__.py", line 1627, in gc_collect
    gc.collect()
ValueError: flush of closed file
Exception ignored in: <_io.BufferedWriter>
  File "/home/vstinner/prog/python/master/Lib/test/support/__init__.py", line 1627, in gc_collect
    gc.collect()
ValueError: flush of closed file
ok
```

vstinner:
I rebased my PR, added an unit test and I also fixed the pure Python implementation.

vstinner:
This PR changes the behavior of the io module for a corner case. I merged PR #14028 instead to only fix an error when running test_io. IMHO the current behavior makes sense and ensures that text, buffered and fileio states are consistent, especially for the closed attribute. I close this PR.

PR #14007, opened 2019-06-12
bpo-36901: Fix some possible memory leaks in PC/bdist_wininst/install.c

ZackerySpytz:
Co-Authored-By: Tiago Filipe Silva <tiagosilva29@sapo.pt>

https://bugs.python.org/issue36901

### DIFF
--- a/PC/bdist_wininst/install.c
+++ b/PC/bdist_wininst/install.c
@@ -1117,6 +1117,7 @@ static BOOL ExtractInstallData(char *data, DWORD size, int *pexe_size,
         return FALSE;
     if (!GetTempPath(sizeof(tempdir), tempdir)
         || !GetTempFileName(tempdir, "~du", 0, ini_file)) {
+        free(ini_file);
         SystemError(GetLastError(),
                      "Could not create temporary file");
         return FALSE;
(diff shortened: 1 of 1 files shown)


PC/bdist_wininst/install.c
```diff
@@ -2549,11 +2562,16 @@ int DoUninstall(int argc, char **argv)
...
+            char **tmp;
             lines_buffer_size += 10;
-            lines = (char **)realloc(lines,
-                                     sizeof(char *) * lines_buffer_size);
-            if (!lines)
+            tmp = (char **)realloc(lines,
+                                   sizeof(char *) * lines_buffer_size);
+            if (!tmp) {
+                fclose(logfile);
+                free(lines);
```
vstinner: lines is an array of strdup() strings. strdup() should be released as well, no?

zware:
Considering that bdist_wininst is deprecated in 3.8, doesn't actually build with the current project file, and has not been regenerated since a few changes ago anyway, I'm going to go ahead and close this as "won't fix".  We aren't likely to regenerate the checked-in exes in light of the deprecation, so this would never actually be released.

Thanks for the patch anyway!

PR #14008, opened 2019-06-12
bpo-37111: Add 'encoding' and 'errors' parameters to logging.basicCon…

vsajip:
…fig.

Add 'errors' parameter to FileHandler and its derived classes (they already
have an 'encoding' parameter).

https://bugs.python.org/issue37111

### DIFF
--- a/Doc/howto/logging.rst
+++ b/Doc/howto/logging.rst
@@ -128,10 +128,18 @@ look at that next. Be sure to try the following in a newly-started Python
 interpreter, and don't just continue from the session described above::
 
    import logging
-   logging.basicConfig(filename='example.log',level=logging.DEBUG)
+   logging.basicConfig(filename='example.log', encoding='utf-8', level=logging.DEBUG)
    logging.debug('This message should go to the log file')
    logging.info('So should this')
    logging.warning('And this, too')
+   logging.error('And non-ASCII stuff, too, like Øresund and Malmö')
+
+.. versionchanged:: 3.9
+   The *encoding* argument was added. In earlier Python versions, or if not
+   specified, the encoding used is the default value used by :func:`open`. While
+   not shown in the above example, an *errors* argument can also now be passed,
+   which determines how encoding errors are handled. For available values and
+   the default, see the documentation for :func:`open`.
 
 And now if we open the file and look at what we have, we should find the log
 messages:
@@ -141,6 +149,7 @@ messages:
    DEBUG:root:This message should go to the log file
    INFO:root:So should this
    WARNING:root:And this, too
+   ERROR:root:And non-ASCII stuff, too, like Øresund and Malmö
 
 This example also shows how you can set the logging level which acts as the
 threshold for tracking. In this case, because we set the threshold to
--- a/Doc/library/logging.handlers.rst
+++ b/Doc/library/logging.handlers.rst
@@ -89,17 +89,21 @@ sends logging output to a disk file.  It inherits the output functionality from
 :class:`StreamHandler`.
 
 
-.. class:: FileHandler(filename, mode='a', encoding=None, delay=False)
+.. class:: FileHandler(filename, mode='a', encoding=None, delay=False, errors=None)
 
    Returns a new instance of the :class:`FileHandler` class. The specified file is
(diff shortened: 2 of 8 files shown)


vsajip:
@zooba I added the `errors` parameter after your last comment on the other, now closed, PR. Do you want to take another look? This PR touches a few more places, as I had to add the `errors` parameter to all the file-based handlers.

Doc/howto/logging.rst
```diff
@@ -128,10 +128,18 @@ look at that next. Be sure to try the following in a newly-started Python
...
    logging.info('So should this')
    logging.warning('And this, too')
+   logging.error('And non-ASCII stuff, too, like Øresund and Malmö')
+
+.. versionchanged:: 3.9
+   The *encoding* argument was added. In earlier Python versions, or if not
+   specified, the encoding used is the default value used by :func:`open`. While
+   not shown in the above example, an *errors* argument can also now be passed,
+   which determines how encoding errors are handled. For available values and
+   the default, see the documentation for :func:`open`.
```
zooba, line 142: Do we normally refer to `open` as the best example of these? Or do we point to `TextIOWrapper`?

Doc/howto/logging.rst
```diff
@@ -141,6 +149,7 @@ messages:
    DEBUG:root:This message should go to the log file
    INFO:root:So should this
    WARNING:root:And this, too
+   ERROR:root:And non-ASCII stuff, too, like Øresund and Malmö
```
zooba, line 152: We've had problems in the past with non-ASCII characters in the docs (breaking the CHM build in particular). I don't remember exactly how we dealt with it, but we might need some kind of escape (or we may have it already...)

Doc/library/logging.handlers.rst
```diff
@@ -89,18 +89,22 @@ sends logging output to a disk file.  It inherits the output functionality from
 :class:`StreamHandler`.
 
 
-.. class:: FileHandler(filename, mode='a', encoding=None, delay=False)
+.. class:: FileHandler(filename, mode='a', encoding=None, delay=False, errors=None)
```
zooba, line 92: Perhaps make it keyword only (add `delay=False, *, errors=None`)?

zooba (commented):
A few quick comments, which I know I could have looked up myself, but can't right now and maybe you can dismiss them before I get a chance to.

Doc/howto/logging.rst
vsajip: Well, `open` makes more sense to me here because many users (especially novice users) will be more familiar with that than with `TextIOWrapper`. Also, the values specified here are passed directly to `open`, so the connection is clearer.

Doc/howto/logging.rst
vsajip: The [Unicode HOWTO](https://raw.githubusercontent.com/python/cpython/3.7/Doc/howto/unicode.rst) appears to contain Unicode characters without any special escaping: for example,

   265E    '♞': BLACK CHESS KNIGHT
   265F    '♟': BLACK CHESS PAWN

zooba: You're right, and they're rendering properly, so I guess it's fine.

zooba (approved):
Up to you whether to make `errors` a keyword-only argument or set a default other than `strict`. But feel free to merge whenever you're ready.

vsajip:
Thanks. I'll change the `errors` default to `backslashreplace`, as that's a reasonable suggestion, but won't make it a keyword-only parameter. Not because that's a bad idea, but because all of the _other_ parameters which would have been keyword-only parameters aren't, because that feature wasn't available when I wrote the original code. I don't want to make the `errors` parameter seem somehow different in kind to all the others. Ideally it would appear next to `encoding`, but of course I can't do that.

vsajip:
@zooba I changed the `errors` default to `backslashreplace` and added some tests. There was one surprise - if `errors` is specified and a file is opened in binary mode, you get an error (though not if `encoding` is specified, even though it's not applicable). You might want to have a quick look at that last commit. If all the tests pass and I don't hear from you in a day or two, I'll merge this.

PR #14009, opened 2019-06-12
bpo-37250: define tp_print for backwards compatibility

jdemeyer:
bugfix for 3.8, so needs backport.

https://bugs.python.org/issue37250


vstinner (commented):
Cython cannot be fixed instead? I am not sure that I like reintroducing tp_print, it is unused for 10 years...

jdemeyer:
> Cython cannot be fixed instead?

This is being discussed at bpo-37221, please discuss there.

vstinner:
Please open a separated issue, https://bugs.python.org/issue37221 is about PyCode_New

vstinner:
I created https://bugs.python.org/issue37250

jdemeyer:
Rejected on bpo-37250

PR #14010, opened 2019-06-12
bpo-37243: Skip test_asyncio.test_streams.test_sendfile unless hasattr(os, 'sendfile')

aixtools:
https://bugs.python.org/issue37243


asvetlov:
Please no.
`stream.sendfile()` should work even without `os.sendfile()` support.
We have a fallback mode for this.
Unfortunately, the fallback doesn't work on AIX :(

Today I'll make a PR that converts test hangs on AIX into a regular failure.

asvetlov:
#14025 should prevent hangs, the actual fix needs more work.

PR #14011, opened 2019-06-12
bpo-35545: Skip `test_asyncio.test_create_connection_ipv6_scope` on AIX

aixtools:
because "getaddrinfo()" behaves different on AIX

https://bugs.python.org/issue35545

### DIFF
--- a/Lib/test/test_asyncio/test_base_events.py
+++ b/Lib/test/test_asyncio/test_base_events.py
@@ -1298,7 +1298,8 @@ class BaseEventLoopWithSelectorTests(test_utils.TestCase):
             t.close()
             test_utils.run_briefly(self.loop)  # allow transport to close
 
-    @unittest.skipIf(sys.platform.startswith('aix'), "getaddrinfo() different on AIX")
+    @unittest.skipIf(sys.platform.startswith('aix'),
+                    "bpo-25545: IPv6 scope id and getaddrinfo() behave differently on AIX")
     @patch_socket
     def test_create_connection_ipv6_scope(self, m_socket):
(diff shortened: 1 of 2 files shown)


Lib/test/test_asyncio/test_base_events.py
```diff
@@ -1298,6 +1298,7 @@ def _test_create_connection_ip_addr(self, m_socket, allow_inet_pton):
             t.close()
             test_utils.run_briefly(self.loop)  # allow transport to close
 
+    @unittest.skipIf(sys.platform.startswith('aix'), "getaddrinfo() different on AIX")
```
vstinner: I suggest:

```
     @unittest.skipIf(sys.platform.startswith('aix'), 
                      "bpo-35545: IPv6 scope id and getaddrinfo() behave differently on AIX") 
```

Misc/NEWS.d/next/Tests/2019-06-12-11-02-06.bpo-35545.s63eHE.rst
```diff
@@ -0,0 +1,4 @@
+Skip `test_asyncio.test_create_connection_ipv6_scope` on AIX 
+because "getaddrinfo()" behaves different on AIX
+
+Patch by M. Felt
```
vstinner: Honestly, it's not really worth it to document when a test is skipped on a specific platform. I suggest to remove the NEWS entry. I added the "skipped news" label.

asvetlov:
The test was added in Python 3.8, backporting to 3.7 is not needed.

asvetlov (approved):

PR #4153, opened 2017-10-27
bpo-4032: Cygwin: Add .dll.a to UnixCCompiler for searching libraries

ma8ma:
Allow the UnixCCompiler.find_library_file to correctly find DLL import libs (.dll.a).

This is an alternative PR for python/cpython#4136:
> This simple patch allows UnixCCompiler.find_library_file, which searching library directories (e.g. /usr/lib), to recognize .dll.a as matches for the library. This is necessary on Cygwin where actual DLLs are not typically found under <prefix>/lib, but rather <prefix>/bin (because Windows searches PATH for DLLs). However one does .dll.a import libs under <prefix>/lib (these are simple static archives that contain stubs for symbols actually found in the DLL, which can be used in lieu of the DLL itself for linking). This satisfies that the library being searched for is typically found.
>
> This is especially necessary now that the _ctypes module can only be built with --with-system-ffi; without this patch the system libffi cannot be found.

The main difference to the original is the use of the `implib_lib_extension` variable. The original patch will overwrite the `dylib_lib_extension` variable which previously only had meaning on OSX, but is a bit of a misnomer for searching for DLL import libs on Cygwin.

https://bugs.python.org/issue4032

PR #14013, opened 2019-06-12
bpo-28459: Fix _pyio on Cygwin where the msvcrt module is not built

embray:
Patch for [bpo-28459](https://bugs.python.org/issue28459), a long-standing issue on Cygwin especially for running some of the test suite since it uses `_pyio`.

This PR is based on top of #4153 which needs to be merged before this issue can be fixed, as the fix relies on ctypes, and the ctypes module can't be built without #4153.

https://bugs.python.org/issue28459

### DIFF
--- a/Lib/_pyio.py
+++ b/Lib/_pyio.py
@@ -10,8 +10,13 @@ import stat
 import sys
 # Import _thread instead of threading to reduce startup cost
 from _thread import allocate_lock as Lock
-if sys.platform in {'win32', 'cygwin'}:
+if sys.platform == 'win32':
     from msvcrt import setmode as _setmode
+elif sys.platform == 'cygwin':
+    import ctypes
+    _cygwin1 = ctypes.CDLL('cygwin1.dll')
+    def _setmode(fd, mode):
+        return _cygwin1._setmode(ctypes.c_int(fd), ctypes.c_int(mode))
 else:
     _setmode = None
 
--- a/Lib/distutils/ccompiler.py
+++ b/Lib/distutils/ccompiler.py
@@ -875,9 +875,10 @@ main (int argc, char **argv) {
     def library_filename(self, libname, lib_type='static',     # or 'shared'
                          strip_dir=0, output_dir=''):
         assert output_dir is not None
-        if lib_type not in ("static", "shared", "dylib", "xcode_stub"):
-            raise ValueError(
-                  "'lib_type' must be \"static\", \"shared\", \"dylib\", or \"xcode_stub\"")
+        if lib_type not in ("static", "shared", "dylib", "xcode_stub",
+                            "implib"):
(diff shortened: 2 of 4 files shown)


mangrisano (approved):

embray:
Pinging core devs on this and #4153.  Also I fixed the networking problems with my Cygwin buildbot worker so it's actually producing meaningful results again, and this will fix a number of tests.

Lib/_pyio.py
```diff
@@ -10,8 +10,13 @@
 import sys
 # Import _thread instead of threading to reduce startup cost
 from _thread import allocate_lock as Lock
-if sys.platform in {'win32', 'cygwin'}:
+if sys.platform == 'win32':
     from msvcrt import setmode as _setmode
+elif sys.platform == 'cygwin':
+    import ctypes
```
benjaminp, line 16: We should avoid `ctypes` in the `io` module. There probably should cygwin equivalent for `msvcrt` instead.

benjaminp (changes requested):
There seems to be an extra distutils change mixed in here.

embray:
@benjaminp 
> There seems to be an extra distutils change mixed in here.

See issue description

> This PR is based on top of #4153 which needs to be merged before this issue can be fixed, as the fix relies on ctypes, and the ctypes module can't be built without #4153.

Lib/_pyio.py
embray: @benjaminp This is the _pyio module used primarily only in testing.  There is no "cygwin equivalent for `msvcrt`"--why should there be?  The closest I can think of is my own [PyCygwin](https://pycygwin.readthedocs.io/en/latest/) package, and it's not part of the standard lib, though I'd be willing to make it so; I'm hardly interested in fighting for a new stdlib package just to fix one bug though.

As was discussed [on the issue](https://bugs.python.org/issue28459#msg280203) it would be possible to expose Cygwin's `setmode()` as `os.setmode()` but it could be confusing since it does not function the same the FreeBSD `setmode()`.

embray:
Thank you for looking at this issue, though #4153 is the one that really needs attention from a core dev.

phdye:
This bug does not seem exist any longer.  At least I could not replicate it with 3.9 on Cygwin 10.0.

```console
$ uname -a
CYGWIN_NT-10.0-22631 xps-ne 3.5.4-1.x86_64 2024-08-25 16:52 UTC x86_64 Cygwin

$ python pyio-example.py
io from '/usr/lib/python3.9/_pyio.py'
Hello, world!
```

`pyio-example.py` :
```python
import _pyio as io
with io.StringIO() as f:
    f.write("Hello, world!")
    f.seek(0)
    content = f.read()

print(f"io from '{io.__file__}'")
print(content)  # Output: Hello, world!
```

PR #14014, opened 2019-06-12
bpo-37247: Swap build_ext and build_py commands

jlvandenhout:
Currently when building an extension which lists a SWIG interface (.i) file as a source file, SWIG creates the files correctly in the build_ext command. Unfortunately this command is run after the build_py command, so the python files (.py) created in the build_ext command are never copied to the installation.

Swapping the build_ext and build_py commands in the sub_commands attribute of the build command fixes this issue.

https://bugs.python.org/issue37247

### DIFF
--- a/Lib/distutils/command/build.py
+++ b/Lib/distutils/command/build.py
@@ -150,8 +150,8 @@ class build(Command):
         return self.distribution.has_scripts()
 
 
-    sub_commands = [('build_py',      has_pure_modules),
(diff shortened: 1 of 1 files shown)


the-knights-who-say-ni:

csabella:
Since this is from an unknown repository and there was discussion on the bug tracker about the best way to implement this, I'm going to close this pull request.  Once consensus is reached on the bug tracker, then a new pull request can be created.  Thank you.

PR #14015, opened 2019-06-12
bpo-37249: add declaration of _PyObject_GetMethod

jdemeyer:
https://bugs.python.org/issue37249

### DIFF
--- a/Include/cpython/object.h
+++ b/Include/cpython/object.h
@@ -318,6 +318,9 @@ PyAPI_FUNC(int) _PyObject_HasAttrId(PyObject *, struct _Py_Identifier *);
 */
 PyAPI_FUNC(int) _PyObject_LookupAttr(PyObject *, PyObject *, PyObject **);
 PyAPI_FUNC(int) _PyObject_LookupAttrId(PyObject *, struct _Py_Identifier *, PyObject **);
+
+PyAPI_FUNC(int) _PyObject_GetMethod(PyObject *obj, PyObject *name, PyObject **method);
+
 PyAPI_FUNC(PyObject **) _PyObject_GetDictPtr(PyObject *);
 PyAPI_FUNC(PyObject *) _PyObject_NextNotImplemented(PyObject *);
 PyAPI_FUNC(void) PyObject_CallFinalizer(PyObject *);
--- a/Objects/call.c
+++ b/Objects/call.c
@@ -1211,9 +1211,6 @@ object_vacall(PyObject *base, PyObject *callable, va_list vargs)
 }
 
 
-/* Private API for the LOAD_METHOD opcode. */
-extern int _PyObject_GetMethod(PyObject *, PyObject *, PyObject **);
-
 PyObject *
 PyObject_CallMethodObjArgs(PyObject *obj, PyObject *name, ...)
 {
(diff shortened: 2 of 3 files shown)


Include/internal/pycore_object.h
```diff
@@ -10,6 +10,10 @@ extern "C" {
 
 #include "pycore_pystate.h"   /* _PyRuntime */
 
+/* Private API for the LOAD_METHOD opcode. */
```
methane: I think this comment is not needed.
We have detailed comment in function definition.

Include/internal/pycore_object.h
```diff
@@ -10,6 +10,10 @@ extern "C" {
...
+PyAPI_FUNC(int)
+_PyObject_GetMethod(PyObject *obj, PyObject *name, PyObject **method);
```
methane: Other API definitions in this file are single line format.

methane: BTW, I'm not sure we need `PyAPI_FUNC` here.
It is used to expose symbol.  But this API is really private.  This is used only in python executable or python dll.
Even extension modules in stdlib must not use it.

jdemeyer: > BTW, I'm not sure we need `PyAPI_FUNC` here.

It seems that most functions in the internal API do use `PyAPI_FUNC`. So it seems better to keep it. If you insist, I will remove the `PyAPI_FUNC`.

jdemeyer:
I'm not actually convinced that `_PyObject_GetMethod` should be super-internal. For example, Cython literally copied the code of `_PyObject_GetMethod` from CPython because CPython does not expose it. So maybe we shouldn't actually make it internal.

methane:
I expect further optimization which requires changing this API completely.

For general use, I want more high level API for calling method with vectercall.

So, if expected user is only Cython, please keep copying.

jdemeyer:
> So, if expected user is only Cython, please keep copying.

OK, got it.

msullivan:
> So, if expected user is only Cython, please keep copying.

I'm probably *also* going to copy it for mypyc at some point, but I think that falls in the same bucket.

msullivan (approved):

methane:
If there are at least two users (mypyc and Cython), I'm OK to move it to private/.

jdemeyer:
> I'm OK to move it to private/.

What do you mean here? There is a directory `cpython/` for the default non-limited API and a directory `internal/` for the API that is only defined when `Py_BUILD_CORE` is defined. Which of the two do you mean?

methane:
I'm sorry, there is no `private/` directory.  As @vstinner said:

```
Python 3.8 now has a better separation between "private" and "internal" 
APIs:

* private are "_Py" symbols which are exported in the python DLL: they 
should not be used, but can be used techically

* internal are symbols defined in internal header files 
(Include/internal/). Some symbols use PyAPI_FUNC()/PyAPI_DATA(), some 
only use "extern".
```

So I meant put this API in `Include/cpython/object.h`, near to `_PyObject_GetAttrId`.

I will add new really internal API to implement per-opcode cache for LOAD_METHOD,
instead of changing existing `_PyObject_GetMethod`.

jdemeyer:
> So I meant put this API in `Include/cpython/object.h`, near to `_PyObject_GetAttrId`.

OK, done.

PR #14016, opened 2019-06-12
bpo-37210: Fix pure Python pickle: proto 5 requires _pickle.PickleBuffer

vstinner:
Fix pure Python implementation of pickle: it failed with
ModuleNotFoundError on importing _pickle.PickleBuffer. If
_pickle.PickleBuffer is missing, change HIGHEST_PROTOCOL from 5 to 4
and disable PickleBuffer support.

https://bugs.python.org/issue37210

### DIFF
--- a/Lib/pickle.py
+++ b/Lib/pickle.py
@@ -36,10 +36,16 @@ import io
 import codecs
 import _compat_pickle
 
-from _pickle import PickleBuffer
-
 __all__ = ["PickleError", "PicklingError", "UnpicklingError", "Pickler",
-           "Unpickler", "dump", "dumps", "load", "loads", "PickleBuffer"]
+           "Unpickler", "dump", "dumps", "load", "loads"]
+
+try:
+    from _pickle import PickleBuffer
+    __all__.append("PickleBuffer")
+    _HAVE_PICKLE_BUFFER = True
+except ImportError:
+    _HAVE_PICKLE_BUFFER = False
+
 
 # Shortcut for use in isinstance testing
 bytes_types = (bytes, bytearray)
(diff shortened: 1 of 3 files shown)


vstinner:
@pitrou: Would you mind to review my change? Does it sound  reasonable to you?

pitrou:
This sounds both fine but almost pointless to me. Is it for `performance`?

vstinner:
> This sounds both fine but almost pointless to me. Is it for performance?

It repairs pyperformance, right: to measure performances of the pure Python implementation of pickle.

But other Python implementations (PyPy, MicroPython, RustPython, etc.) should also benefit of this change to reuse CPython pickle.py if they miss the _pickle module.

pitrou:
Ok. I'm not sure you need to downgrade the protocol to 4. You can just disable the code paths which depend on PickleBuffer.

vstinner:
> I'm not sure you need to downgrade the protocol to 4. You can just disable the code paths which depend on PickleBuffer.

Oh, I wasn't sure. I tested test_pickle with `import sys; sys.modules['_pickle'] = None`: I confirm that test_pickle pass if I leave HIGHEST_PROTOCOL unchanged (equal to 5).

I rebased my PR, I reverted my HIGHEST_PROTOCOL change and I made another tiny fix in test_pickle when _pickle is missing.

vstinner:
@pitrou: Does it look better now?

pitrou (approved):
+1.

vstinner:
@pitrou: Thanks for the review, I updated the commit message using your NEWS entry ;-)

PR #14017, opened 2019-06-12
bpo-37252: Fix devpoll tests

kulikjak:
test_devpoll currently ends with two failures with Python 3.8 on Solaris.

First one is wrong number of arguments to devpoll.register function (which thrown the same error as expected in 3.7 but now acts differently).

Second one is that register and modify no longer throw OverflowError when negative number is given as second argument, but rather a ValueError. I am not sure whether this is just a small semantics change or some bigger problem (documentation doesn't mention what error should be thrown).

So I fixed it by changing the expected thrown error but there might be other problem as well.

https://bugs.python.org/issue37252

### DIFF
--- a/Lib/test/test_devpoll.py
+++ b/Lib/test/test_devpoll.py
@@ -109,7 +109,7 @@ class DevPollTests(unittest.TestCase):
         # operations must fail with ValueError("I/O operation on closed ...")
         self.assertRaises(ValueError, devpoll.modify, fd, select.POLLIN)
         self.assertRaises(ValueError, devpoll.poll)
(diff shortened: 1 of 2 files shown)


asvetlov:
Would you point on commits that produced test failures?
`git bisect` can help

kulikjak:
I bisected Python and found out following:

Test was first broken with 6dc57e2 - it ends with `ModuleNotFoundError: No module named 'select'` error. It was later followed with 0e6e7a1 which did fixed the not found error, but now there are above mentioned problems.

benjaminp:
We don't have a Solaris buildbot any more, so we'll have to take your word for it.

PR #14018, opened 2019-06-12
bpo-37253: Add _PyCompilerFlags_INIT macro

vstinner:
Add a new _PyCompilerFlags_INIT macro to initialize PyCompilerFlags
variables, rather than initializing cf_flags and cf_feature_version
explicitly in each variable.


vstinner:
Since the added macro is private, I chose to not document it and omit a NEWS entry.

PR #14019, opened 2019-06-12
bpo-37253: Document PyCompilerFlags.cf_feature_version

vstinner:
* Update PyCompilerFlags structure documentation.
* Document the new cf_feature_version field in the Changes in the C
  API section of the What's New in Python 3.8 doc.


vstinner:
I'm not sure if the new field must be documented, since it's ignored by default and used if and only if PyCF_ONLY_AST flag is used.

But PyCF_ONLY_AST flag must not be used in functions documented at:
https://docs.python.org/dev/c-api/veryhigh.html

PR #14020, opened 2019-06-12
bpo-37253: Remove PyAST_obj2mod_ex() function

vstinner:
PyAST_obj2mod_ex() is similar to PyAST_obj2mod() with an additional
'feature_version' parameter which is unused.


vstinner:
@gvanrossum, @ilevkivskyi: I understand that PyAST_obj2mod_ex() used its feature_version parameter, but it doesn't anymore, and so it's now safe to remove the function. Would you mind to have a look?

gvanrossum (approved):
👍

PR #14022, opened 2019-06-12
bpo-37130: Path('..').name now returns ''

NophKe:
To be consistent with the behavior of Path('.').name that returns an
empty string, the name property will check for the case of '..' dir.
Consequently .with_suffix() and .with_name() will raise 'has an empty name'
exception.
The .stem property test already handling it,  '..' becomes a hard-coded
special case.

- updated documentation of .name
- added '..' case in .name()
- added '..' cases in related tests
- added special case for .stem()

https://bugs.python.org/issue37130

### DIFF
--- a/Doc/library/pathlib.rst
+++ b/Doc/library/pathlib.rst
@@ -381,6 +381,12 @@ Pure paths provide the following methods and properties:
       >>> PureWindowsPath('//some/share').name
       ''
 
+   Trailing dots are not considered as filenames::
+
+      >>> PurePosixPath('..').name
+      ''
+
+
 
 .. data:: PurePath.suffix
 
--- a/Doc/library/pathlib.rst
+++ b/Doc/library/pathlib.rst
@@ -381,11 +381,10 @@ Pure paths provide the following methods and properties:
       >>> PureWindowsPath('//some/share').name
       ''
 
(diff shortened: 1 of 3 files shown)


Doc/library/pathlib.rst
```diff
@@ -381,6 +381,12 @@ Pure paths provide the following methods and properties:
       >>> PureWindowsPath('//some/share').name
       ''
 
+   Trailing dots are not considered as filenames::
+
+      >>> PurePosixPath('..').name
+      ''
+
+
```
mangrisano, line 389: I think there is one more newline here. What do you think about it?

mangrisano (commented):
Hi, thank you for the Pull Request.
I've left a suggestion about the changes.

Thank you. :)

Doc/library/pathlib.rst
NophKe: Yep, I missed it.... 
And by reading it again i chose an other phrasing that seems more appropriate....
Thank you,

mangrisano (approved):

mangrisano:
LGTM. Thank you.

brettcannon:
FYI for core devs: the best way to handle the issue is still being discussed on b.p.o.

PR #14028, opened 2019-06-12
bpo-37223: Second call to BufferedWrite.close() does nothing

vstinner:
When io.BufferedWrite.close() is called twice, the second call now
does nothing. Previously, the flush() method and raw.close() were
called at the second call.

Moreover, even if close() raises an exception, the file is now
considered as closed.

https://bugs.python.org/issue37223

### DIFF
--- a/Lib/_pyio.py
+++ b/Lib/_pyio.py
@@ -786,6 +786,7 @@ class _BufferedIOMixin(BufferedIOBase):
 
     def __init__(self, raw):
         self._raw = raw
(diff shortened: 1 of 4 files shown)


vstinner:
This PR changes the behavior of the io module for a corner case. I merged PR #14028 instead to only fix an error when running test_io. IMHO the current behavior makes sense and ensures that text, buffered and fileio states are consistent, especially for the closed attribute. I close this PR.

PR #14030, opened 2019-06-12
bpo-17535: IDLE editor line numbers

taleinat:
Line numbers for IDLE editor windows!

This is based on Saimadhav Heblikar's last patch posted to the related issue.

https://bugs.python.org/issue17535


tebeka:
I'd love to see line number finally landing in IDLE. I use it a lot for teaching and it'll be a great help referencing parts of the code.

taleinat:
I've made some significant improvements in several aspects:

1. Scrolling: performance and behavior
2. Selection entire lines by dragging over line numbers
3. Color config: a separate highlighting setting for line numbers, which takes effect immediately

terryjreedy:
Tal, thank you for reviving the improving the old patch.  This seems to work better than the 2014 version, enough so I want to get this into b3 (scheduled in 3 weeks, 7/29).  If we can coordinate, I will help with the needed changes.  (I believe some of them amount to fixing parts of the old patch that are now obsolete.)

I would like to immediately follow up with a shell sidebar, which will, among other things, separate prompt and code and allow properly spaced indents instead of the wonky tabbing.  (we might do something else immediately, but I want it eventually.)  I will try to post something for discussion tomorrow.

Issues (couple of which would be separate).

0. New names: Because the sidebar is not just about line numbers, the new file should be sidebar.py and the highlights sidebar-foreground and -background.

1. Startup
```
>>> import idlelib.idle

 Warning: config.IdleConf.GetThemeDict -
 problem retrieving theme element 'linenumber-foreground'
 from theme 'Custom Dark'.
 returning default color: '#000000'

 Warning: config.IdleConf.GetThemeDict -
 problem retrieving theme element 'linenumber-background'
 from theme 'Custom Dark'.
 returning default color: '#ffffff'
```
This happens with custom themes every time we add theme elements.  We immediately need to extend previous workaround, but should do a more permanent fix that adds elements, with default values, to custom themes.

2. Default status: When I opened the editor, line numbers appeared.  I believe the consensus on the issue is that the default should be off.  

3. Menu: Location good, but as I said on the issue, I want to copy the code context solution, which works, instead of keeping the check mark.  Besides aesthetics and consistency, the line number mark is buggy in a way similar to the way that the old code context check was: changing the mark in one window affects the mark in other windows, possibly resulting in the menu checkmark for a window becoming out of sync with the window state.  When a window with a check does not have line numbers, they sometimes cannot be added.

4. Scrolling.  While working on code context, Cheryl and I fixed scrolling to be by lines, and intended that there should always be an even line at the top.  I wonder if this helped this patch work better.

The default size of a Text widget with all lines the same height (as in IDLE) is an even number of lines, and when this is true, there are no partial lines to worry about.  When a user changes the window height to have a fractional size, IDLE (*without* this patch) puts a whole line at the top and the fractional line at the bottom.  I intend that this remain true.

While using  the patch, I discovered that when one scrolls to the bottom of the file, the fractional line moves to the top.  It stays there until one scrolls to the top or until the widget just switches back.  This looks even worse with line numbers, so a fix would be good if it can be done without making scrolling on macOS worse.  (1? adjust window size to an integral number of line after a resize?  2? display a partial unnumbered blank line, like Notepad++.)

5. Selection: With Notepad++, one can select a whole line by clicking on a line number.  To select multiple lines, one drags the cursor to the last line one wants to include.  Very intuitive to me.  With this patch, clicking does nothing and one must drag to the first line to be excluded.  Much worse.  This slice-like behavior would look better if the cursor were a horizontal line, so one could click between line numbers and drag to another slice position, between numbers.  But even then, click -- select would be more useful, and eventually I would like to be able to select a multiline statement with a double click.  (And in Shell, double click select an entire traceback, where the sidebar might have 'ERR' opposite the 'Traceback...' line.)

6.  Font resizing:  This seems to now work the same as without line numbers, with none of the vertical jumps I remember.  The main issue now is that with 4 windows, at least 1 with a few 1000 lines, resizing takes about a second to complete, mostly because of re-colorizing the files.  It naively seems that changing the font attribute should be enough.  Do you know whether it is tk or IDLE that removes the tags?

terryjreedy:
8. Code Context.  The CC box extends over the side bar.  Result is displayed lines offset from text below.  Not good.  If we grid, which we should, row 0, column 0 should be blank.  Whether and  how color background to be worked out.

terryjreedy (changes requested):
See separate comments.  Will look at code later.

taleinat:
> 6. Font resizing: This seems to now work the same as without line numbers, with none of the vertical jumps I remember. The main issue now is that with 4 windows, at least 1 with a few 1000 lines, resizing takes about a second to complete, mostly because of re-colorizing the files. It naively seems that changing the font attribute should be enough. Do you know whether it is tk or IDLE that removes the tags?

I believe this happens without this change as well. Unless this somehow makes things worse—which would be surprising—I'd prefer not to address this as part of this PR. I'm happy to look into it separately!

taleinat:
> 7. Implementation: Implementing builtin features as extensions is obsolete since Sept 2017. Issue 27099.

This is not implemented as an extension. (The addition to `Lib/idlelib/config-extensions.def` is commented out and does nothing; I just forgot to remove it.)

taleinat:
@terryjreedy 

> 1. This happens with custom themes every time we add theme elements. We immediately need to extend previous workaround, but should do a more permanent fix that adds elements, with default values, to custom themes.

What is the previous workaround? What exactly do you suggest we do in the context of this PR?

taleinat:
> 4. While using the patch, I discovered that when one scrolls to the bottom of the file, the fractional line moves to the top. It stays there until one scrolls to the top or until the widget just switches back. This looks even worse with line numbers, so a fix would be good if it can be done without making scrolling on macOS worse. (1? adjust window size to an integral number of line after a resize? 2? display a partial unnumbered blank line, like Notepad++.)

Other editors, e.g. PyCharm, just allow the lines and line numbers to be partially shown/hidden. Personally I don't see any issue with this. I would greatly prefer not to deal with scrolling further as part of this PR; I've already spent hours getting it working well.

taleinat:
> 5. Selection: With Notepad++, one can select a whole line by clicking on a line number. To select multiple lines, one drags the cursor to the last line one wants to include. Very intuitive to me. With this patch, clicking does nothing and one must drag to the first line to be excluded. Much worse. This slice-like behavior would look better if the cursor were a horizontal line, so one could click between line numbers and drag to another slice position, between numbers. But even then, click -- select would be more useful, and eventually I would like to be able to select a multiline statement with a double click. (And in Shell, double click select an entire traceback, where the sidebar might have 'ERR' opposite the 'Traceback...' line.)

I've just pushed an implementation of selection behavior as suggested: Clicking a line number selects the whole line; dragging on line numbers selects a range of lines, inclusively.

Note that before this change, clicking on a line number would be like clicking at the beginning of a line in the editor: The insertion cursor would be placed at the line start, and the selection would be cleared. Now, with the recent change, clicking at the beginning of a line is difficult, and it may be common to accidentally select a line instead of moving the cursor to its beginning.

terryjreedy:
7 does not apply, 4 and 6 are possible other issues, definitely for later.  I will try to do any of 1, 2, 3 not already done sometime tomorrow, and also look what you did for 5.  I will then retest in use, look at the unittests, and consider whether I think anything more is truly *required* for a first merge. 

If you are only testing on Mac, I will eventually ask Cheryl to try out on Ubuntu.

taleinat:
> 8. Code Context. The CC box extends over the side bar. Result is displayed lines offset from text below. Not good. If we grid, which we should, row 0, column 0 should be blank. Whether and how color background to be worked out.

In general, I agree. I suggest that the (0,0) cell should be the same color as the sidebar background. The (0,2) cell (above the scrollbar) should likely have the code context background color.

However, I'm having trouble working on this on macOS, since scrolling appears to be rather broken on macOS when using code context, including on 3.7 and 3.8. I'll see if I can get it working properly on Windows, and open a separate macOS-specific issue about scrolling with code context.

terryjreedy:
I don't want macOS to block merging when this is otherwise ready to merge.  If we have to, we can make code context and line number mutually exclusive on macOS.  The quickest fix would be to disable the one that is less important, which I think is code context.

Your code context patch #14675 overlaps with this.  I might merge it first, in which case
9.? Update to accommodate other idlelib changes.

I am going to work on 1-3 now.

terryjreedy:
11. Single clicking selects the line as desired.  Double clicking sends TclError to the console.
```
Exception in Tkinter callback
  File "F:\dev\3x\lib\tkinter\__init__.py", line 1883, in __call__
    return self.func(*args)
  File "F:\dev\3x\lib\idlelib\linenumbers.py", line 67, in redirect_mousebutton_event
    self.text.event_generate(event_name or event, x=0, y=event.y)
  File "F:\dev\3x\lib\tkinter\__init__.py", line 1849, in event_generate
    self.tk.call(args)
_tkinter.TclError: Double or Triple modifier not allowed
```
and then selects.  Not good, besides wanting to eventually recognize double clicks.  In the meanwhile, they should act same a click or two clicks.

Drag select sometimes works, but is sometimes nearly unusable as the selection flashes all over the screen while dragging.  Even though click moves the insertion cursor, there seems to be a ghost memory of 1 or 2 previous positions.  I have looked carefully at cursor movement and behavior of both left margin selection and sidebar selection on both IDLE and Notepad++.  I will check other programs and see if anythings helps understand what is going on.

terryjreedy:
I pushed what I've done so far.  The workaround needed for 1. was the one previously used for new key-bindings.  'context' and 'linenumber' are the only newish theme elements.  My comment about a better solution later applies to both.  I decided to stick with 'linenumber' since we might not use those elements for a shell sidebar.

taleinat:
> Double clicking sends TclError to the console.

Fixed.

Lib/idlelib/config.py
```diff
@@ -345,11 +349,11 @@ def GetThemeDict(self, type, themeName):
...
                 'console-foreground':'#000000',
                 'console-background':'#ffffff',
-                'context-foreground':'#000000',
-                'context-background':'#ffffff',
                 }
         for element in theme:
-            if not cfgParser.has_option(themeName, element):
+            if (not cfgParser.has_option(themeName, element) and
+                # Skip warning for new elements.
+                element[:element.index('-')] not in ['context', 'sidebar']):
```
taleinat:
```suggestion
                not element.startswith(('context-', 'linenumber-')):
```

taleinat:
> Drag select sometimes works, but is sometimes nearly unusable as the selection flashes all over the screen while dragging.

I've seen this happen only with code-context active. Does this also happen without it? If so, can you describe how to reproduce?

Lib/idlelib/config.py
terryjreedy: I agree, but cannot commit thru this interface.  Says "You do not have permission to push to this repository.", which is wrong for this branch.  Can you accept your own suggestion?

terryjreedy:
Code context not needed.  Open a file such as run.py.  Cursor at top of file.  Sidebar selection works great as long as cursor never moves from initial position.  Selections leave visible cursor at top of file.  I think this is a bug, because clicking and dragging in the text widget move the cursor.  Leaving the cursor alone only works if is has never moved.

Sidebar selection in Notepad++ results in a text state, including cursor position, that is the same as if the selection were done in the text widget.  So click on a line and the cursor is at the beginning of the next line (if there is one).  This is same result as click at beginning of line and drag down (with either IDLE or N++).  Imitating in-text selection is what I would try. 

Anyway, click in the file, say at (blank) line 18, moving the cursor.  Click another line, selecting it.  Cursor remains on 18.  Click elsewhere and drag.  Cursor visibly disappears, but the part of the text that flashes is the part (or a portion thereof) between the true selection and the ghost cursor.  So click text on line 2 and sidebar around 40 and there are flashes all over.  Sometimes parts of the correct selection flash, especially when dragging up.

taleinat:
> I don't want macOS to block merging when this is otherwise ready to merge. If we have to, we can make code context and line number mutually exclusive on macOS. The quickest fix would be to disable the one that is less important, which I think is code context.

I agree. It's not really blocking work, since scrolling works well without code context active, and scrolling is already rather broken. I'm just not test scrolling with both active.

The root of the problem is that the code context window changes its size according to the number of context lines. This in turn resizes the main editor text widget and its scrollbar, which causes issues while scrolling, especially by dragging the scrollbar and now also by dragging on the line numbers. I'm going to open a separate issue so we can discuss this there, since it is really almost entirely unrelated.

> Your code context patch #14675 overlaps with this. I might merge it first, in which case
9.? Update to accommodate other idlelib changes.

I'm happy to do the merging after one of these PRs is merged into master.

taleinat:
I've addressed points 2 & 3, making the line numbers menu option behave exactly like the code context menu option. I've also added a config variable, with appropriate config UI, to set whether line numbers are initially displayed for new windows.

And with that I think we've addressed all of the pending issues! 🥳

I image you'll want more tests, @terryjreedy, for new functionality such as the new config option. Also, we've still to merge this PR with the recent code context PR, which I've volunteered to do in any order that you would prefer.

taleinat:
> The root of the problem is that the code context window changes its size according to the number of context lines. This in turn resizes the main editor text widget and its scrollbar, which causes issues while scrolling, especially by dragging the scrollbar and now also by dragging on the line numbers. I'm going to open a separate issue so we can discuss this there, since it is really almost entirely unrelated.

See bpo-37559.

taleinat:
I've added more tests, bringing coverage up to 97%.

The drag-scrolling can't be initiated by generating Tk events (I tried!). I added an htest for it. The 97% coverage is achieved by the automated tests and the htest together.

The only thing I haven't managed to test properly is the focus redirection, but that is very minor.

taleinat:
> Anyway, click in the file, say at (blank) line 18, moving the cursor. Click another line, selecting it. Cursor remains on 18. Click elsewhere and drag. Cursor visibly disappears, but the part of the text that flashes is the part (or a portion thereof) between the true selection and the ghost cursor. So click text on line 2 and sidebar around 40 and there are flashes all over. Sometimes parts of the correct selection flash, especially when dragging up.

Fixed! Give it a spin, I think you'll like it :)

taleinat:
@csabella, @tebeka, I think this is now ready for wider review. I'd love to get your input!

csabella:
Hi @taleinat, great work on this!  I've run at it locally and it looks awesome!  Unfortunately, I won't be able to spend too much time with it for a few weeks.  I keep trying to get back to all the recent PR activity and I just can't find enough time.  Sorry about that.

terryjreedy (changes requested):
I fixed pending merge conflicts.  Both CI and my manual tests of code context pass.  Cell00 looks fine, as does unfilled cell 0, 2, above scroller.  As for line numbers, line selection works as requested and crazy (additional) flashing is gone, at least on Windows.

Minor issue.  Line numbers are shifted up one pixel, at least on my machine.  I thought they seemed off, but verified and measured.  Open a new file with line numbers.  On lines 2 and 7, enter multiple 2s and 7s.  Use paper to trace underbars and overbars to sidebar.  Is there a difference in padding between the two text widgets?

Major issue: open settings; change font, apply changes.  Usually text changes, numbers do not; this is a bug.  Hit Ok.  Usually, number catch up, but one can instead exit with Cancel, which only cancels changes since apply.  'Bold []' changes act differently from size and face changes (and the latter can affect size).   The result is that after closing the config dialog, the number font may not match the text font.  This is especially bad when the spacing is different. I will try to investigate.

terryjreedy:
The spacing difference is due to `borderwidth=0` in the sidebar init.  Replacing that with `relief='flat'` might have worked, but I am fine with the narrow x margins.  Given that 0 is 2 less than the default border width of 2, and that the default pady is 1, I tried both 2 and 3 for pady, but 2 is definitely better for me.  

I did not get to 'apply' issue tonight.

taleinat:
> Minor issue. Line numbers are shifted up one pixel, at least on my machine. ... Is there a difference in padding between the two text widgets?

Yes, there was. I've fixed this using a similar method to that I implemented in CodeContext to ensure proper alignment with the main text widget.

> Major issue: open settings; change font, apply changes. Usually text changes, numbers do not; this is a bug. Hit Ok. Usually, number catch up, but one can instead exit with Cancel, which only cancels changes since apply. 'Bold []' changes act differently from size and face changes (and the latter can affect size). The result is that after closing the config dialog, the number font may not match the text font. This is especially bad when the spacing is different. I will try to investigate.

This was indeed a bug, now fixed. I went ahead and refactored the (similar) config update mechanisms of both CodeContext and LineNumbers, while keeping useful default for the base SideBar class.

Lib/idlelib/sidebar.py
```diff
@@ -0,0 +1,323 @@
...
+
+        for button in [2, 3, 4, 5]:
+            for event_name in (f'<Button-{button}>',
+                               f'<ButtonRelease-{button}>',
+                               f'<B{button}-Motion>',
+                               ):
+                bind_mouse_event(event_name, target_event_name=event_name)
+
+            # Convert double- and triple-click events to normal click events,
+            # since event_generate() doesn't allow generating such events.
```
terryjreedy, line 200: This does not mean that we cannot use double-clicks, just that we cannot test their effect, at least not easily.  But I think that this is the best thing to do until we do use them.

terryjreedy (approved):
Numbers are still lined up properly.  On my screen they appear to be a bit shorter.  I believe that this is the result of fg/bg color differences resulting in subpixel shading.  So it is not a bug (but might someday get reported as such).

Font changes work for me.  With this, and the doc changes I am about to push*, I am ready for this to be merged.  I have configured line numbers on so I will see them daily as soon as it is.  I will review the sidebar code and tests more closely when I try making another subclass.

* idle.rst , help.html, and the blurb.  My commit got squashed when I re-pulled to get the test_sidebar change.

taleinat:
I've made a few more touch-ups to the docs, added testing for font and color config updates, and made a small fix to the htest. I also think that this is ready to go. @terryjreedy, care for a final look, especially at my latest doc changes?

Re. coverage, it is still at 93%. This points to two future improvements we should make (IMO after this goes in):

1. Extract the `get_widget_padding()` utility function to a separate module and add tests for it.
2. The base SideBar class should be exercised more by tests when it has additional sub-classes; until then it's not worth the effort IMO.

taleinat:
I've also added "What's New" entries for 3.8 and 3.7.5.

terryjreedy:
The doc looks even better.

PR #14039, opened 2019-06-13
bpo-37257:  obmalloc:  stop simple arena thrashing

tim-one:
[bpo-37257](https://bugs.python.org/issue37257):  obmalloc:  stop simple arena thrashing

https://bugs.python.org/issue37257

### DIFF
--- a/Objects/obmalloc.c
+++ b/Objects/obmalloc.c
@@ -1758,7 +1758,12 @@ pymalloc_free(void *ctx, void *p)
     /* All the rest is arena management.  We just freed
      * a pool, and there are 4 cases for arena mgmt:
      * 1. If all the pools are free, return the arena to
-     *    the system free().
+     *    the system free().  Except if this is the last
+     *    arena in the list, keep it to avoid trashing:
+     *    keeping one wholly free arena in the list avoids
+     *    pathological cases where a simple loop would
+     *    otherwise provoke needing to allocate and free an
+     *    arena on every iteration.  See bpo-37257.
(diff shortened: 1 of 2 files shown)


Objects/obmalloc.c
```diff
@@ -1758,7 +1758,12 @@ pymalloc_free(void *ctx, void *p)
     /* All the rest is arena management.  We just freed
      * a pool, and there are 4 cases for arena mgmt:
      * 1. If all the pools are free, return the arena to
-     *    the system free().
+     *    the system free().  Except if this is the last
+     *    arena in the list, keep it to avoid trashing:
```
methane:
```suggestion
     *    arena in the list, keep it to avoid thrashing:
```

Objects/obmalloc.c
```diff
@@ -1767,7 +1772,7 @@ pymalloc_free(void *ctx, void *p)
      *    nfreepools.
      * 4. Else there's nothing more to do.
      */
-    if (nf == ao->ntotalpools) {
+    if (nf == ao->ntotalpools && ao->nextarena!= NULL) {
```
methane:
```suggestion
    if (nf == ao->ntotalpools && ao->nextarena != NULL) {
```

methane:
LGTM except two nitpicks.

Objects/obmalloc.c
tim-one: @methane, thanks for that!  Those are kinds of silly typos I usually don't make .  Gladly accepted the changes :smile:.

PR #14049, opened 2019-06-13
bpo-37266: Daemon threads are now denied in subinterpreters

vstinner:
In a subinterpreter, spawning a daemon thread now raises an
exception. Daemon threads were never supported in subinterpreters.
Previously, the subinterpreter finalization crashed with a Pyton
fatal error if a daemon thread was still running.

* Add _thread._is_main_interpreter()
* threading.Thread.start() now raises RuntimeError if the thread is a
  daemon thread and the method is called from a subinterpreter.
* The _thread module now uses Argument Clinic the new function.
* Use textwrap.dedent() in test_threading.SubinterpThreadingTests

https://bugs.python.org/issue37266

### DIFF
--- a/Doc/library/threading.rst
+++ b/Doc/library/threading.rst
@@ -272,6 +272,8 @@ since it is impossible to detect the termination of alien threads.
    base class constructor (``Thread.__init__()``) before doing anything else to
    the thread.
 
+   Daemon threads must not be used in subinterpreters.
+
    .. versionchanged:: 3.3
       Added the *daemon* argument.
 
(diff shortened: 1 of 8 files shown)


vstinner:
Oops. Typo in my PR title: "bpo-37266: Daemon threads are now **denied** in subinterpreters".

vstinner:
> Add _thread._is_main_interpreter()

I chose to add the new function the _thread module to not add a dependency (import) between thread and <another module>. For example, I would prefer that threading doesn't import _xxsubintepreters. The module is private, its API is a work-in-progress and it may pull other things which are not needed by threads.

PR #14052, opened 2019-06-13
bpo-37261: Fix support.catch_unraisable_exception()

vstinner:
The __exit__() method of test.support.catch_unraisable_exception
context manager now ignores unraisable exception raised when clearing
self.unraisable attribute.

https://bugs.python.org/issue37261

### DIFF
--- a/Lib/test/support/__init__.py
+++ b/Lib/test/support/__init__.py
@@ -3070,7 +3070,7 @@ class catch_unraisable_exception:
         self._old_hook = None
 
     def _hook(self, unraisable):
-        # Storing unraisable.object can resurect an object which is being
+        # Storing unraisable.object can resurrect an object which is being
         # finalized. Storing unraisable.exc_value creates a reference cycle.
         self.unraisable = unraisable
 
--- a/Lib/test/support/__init__.py
+++ b/Lib/test/support/__init__.py
@@ -3040,6 +3040,19 @@ class catch_unraisable_exception:
     """
     Context manager catching unraisable exception using sys.unraisablehook.
 
+    The the *object* attribute of the unraisable hook is set and the object is
+    being finalized, the object is resurected because the context manager
+    stores a strong reference to it (cm.unraisable.object).
+
+    Storing the exception value (cm.unraisable.exc_value) creates a reference
+    cycle. The reference cycle is broken explicitly when the context manager
+    exists.
+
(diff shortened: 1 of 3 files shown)


Lib/test/support/__init__.py
```diff
@@ -3040,6 +3040,19 @@ class catch_unraisable_exception:
     """
     Context manager catching unraisable exception using sys.unraisablehook.
 
+    The the *object* attribute of the unraisable hook is set and the object is
```
tirkarthi:
```suggestion
    The *object* attribute of the unraisable hook is set and the object is
```

Lib/test/support/__init__.py
```diff
@@ -3040,6 +3040,19 @@ class catch_unraisable_exception:
...
+    being finalized, the object is resurected because the context manager
```
tirkarthi:
```suggestion
    being finalized, the object is resurrected because the context manager
```

resurected -> resurrected in below content too.

Lib/test/support/__init__.py
```diff
@@ -3040,6 +3040,19 @@ class catch_unraisable_exception:
...
+    stores a strong reference to it (cm.unraisable.object).
+
+    Storing the exception value (cm.unraisable.exc_value) creates a reference
+    cycle. The reference cycle is broken explicitly when the context manager
+    exists.
```
tirkarthi: Should "exists" be "exits" as I understand that reference cycle is broken explicitly when `__exit__` of context manager is called

tirkarthi (commented):
Minor typo suggestions. Is this worth updating `Doc/library/test.rst` too?

vstinner:
> Minor typo suggestions. Is this worth updating Doc/library/test.rst too?

Done. Thanks for the review ;-)

vstinner:
@tirkarthi: Would you mind to review my update PR?

Lib/test/support/__init__.py
```diff
@@ -3058,6 +3070,8 @@ def __init__(self):
         self._old_hook = None
 
     def _hook(self, unraisable):
+        # Storing unraisable.object can resurect an object which is being
```
tirkarthi:
```suggestion
        # Storing unraisable.object can resurrect an object which is being
```

tirkarthi (approved):
One minor typo in code comment that is same as doc change to use resurrect. Doc changes look good to me. Thanks.

PR #14061, opened 2019-06-13
Update concurrent.futures.rst

geryogam:
This PR adds missing details in the [`concurrent.futures`](https://docs.python.org/3/library/concurrent.futures.html) documentation:

* the mention that `Future.cancel` also returns `False` if the call finished running;
* the mention of the states for `Future` that did not complete: pending or running.

### DIFF
--- a/Doc/library/concurrent.futures.rst
+++ b/Doc/library/concurrent.futures.rst
@@ -306,9 +306,10 @@ The :class:`Future` class encapsulates the asynchronous execution of a callable.
 
     .. method:: cancel()
 
-       Attempt to cancel the call.  If the call is currently being executed and
-       cannot be cancelled then the method will return ``False``, otherwise the
-       call will be cancelled and the method will return ``True``.
+       Attempt to cancel the call.  If the call is currently being executed or
+       finished running and cannot be cancelled then the method will return
+       ``False``, otherwise the call will be cancelled and the method will
+       return ``True``.
 
     .. method:: cancelled()
 
--- a/Doc/library/concurrent.futures.rst
+++ b/Doc/library/concurrent.futures.rst
@@ -423,8 +423,9 @@ Module Functions
    Wait for the :class:`Future` instances (possibly created by different
    :class:`Executor` instances) given by *fs* to complete.  Returns a named
    2-tuple of sets.  The first set, named ``done``, contains the futures that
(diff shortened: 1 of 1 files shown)


Doc/library/concurrent.futures.rst
```diff
@@ -306,9 +306,10 @@ The :class:`Future` class encapsulates the asynchronous execution of a callable.
 
     .. method:: cancel()
 
-       Attempt to cancel the call.  If the call is currently being executed and
-       cannot be cancelled then the method will return ``False``, otherwise the
-       call will be cancelled and the method will return ``True``.
+       Attempt to cancel the call.  If the call is currently being executed or
+       finished running and cannot be cancelled then the method will return
```
asvetlov, line 310:
```suggestion
       finished and cannot be cancelled then the method will return
```
I'm not a native speaker but maybe my suggestion is a little more readable?

asvetlov (changes requested):

Doc/library/concurrent.futures.rst
geryogam: @asvetlov

I am not a native speaker myself, but I am not sure that removing "running" makes the sentence more readable, because to finish doing what? I used the same "finish running" phrasing as in the existing [done](https://docs.python.org/3/library/concurrent.futures.html#concurrent.futures.Future.done) and [add_done_callback](https://docs.python.org/3/library/concurrent.futures.html#concurrent.futures.Future.add_done_callback) descriptions.

But maybe what bothered you was that you wanted to use the exact names of the five internal states of `Future` listed in the [concurrent.futures._base module](https://github.com/python/cpython/blob/master/Lib/concurrent/futures/_base.py#L25), which are PENDING, RUNNING, CANCELLED, CANCELLED_AND_NOTIFIED and FINISHED. So in this case the sentence:

> If the call is currently being executed or finished running and cannot be cancelled then…

could be replaced by:

> If the call is in the running state or the cancelled state then…

and likewise in all the page.

asvetlov: Ok, if `done()` uses "finished running" I can live with it.

asvetlov (approved):

asvetlov:
Thanks!

maggyero:
@asvetlov Thank you for the review!

PR #14065, opened 2019-06-13
Implement Windows release builds in Azure Pipelines

zooba:
This automates most of the Windows Release build in a single (multi-stage) Azure Pipelines job.

You can see a sample run at https://dev.azure.com/python/cpython/_build/results?buildId=45280&view=results

It still requires my custom build machine to be running, but only when `SigningCertificate` is set. Without this, and without `UsePGO`, it should all run on hosted machines.

The final publish steps are currently not implemented, as they weren't previously anyway (other than Nuget). I expect to leave them unimplemented for now, and just won't use them, but eventually may come around to automating it the whole way through.

Once you get past the .yml files, there are actual changes to the build files. I found it convenient to generate more files at build (e.g. `LICENSE.txt`) rather than doing it as part of setup. There shouldn't be anything dramatic in here, and it obviously all works with the release build process or I wouldn't be posting yet :)


asvetlov:
Side question.
Does azure look on `.azure-piplenes` folder or `.azure-pipelines.yml`?
What files are used as entrypoints for the folder case?

I've tried to search using google and azure pipelines documentation site embedded searcher but found nothing

zooba:
@asvetlov It's totally customizable. We have multiple build definitions configured for this repo - one using `.azure-pipelines/ci.yml`, one for `.azure-pipelines/pr.yml` and now one for `.azure-pipelines/windows-release.yml`.

If you use `azure-pipelines.yml` in the root directory, then it will be automatically detected when you add the repo. But it's only a default.

zooba:
You should be able to see all the builds at https://dev.azure.com/python/cpython/_build

I also have a few manually created ones for building the OpenSSL and libffi bits that we use.

asvetlov:
@zooba thanks.
Now I figured out that I can use a path to pipeline in azure pipelines settings UI.
I've missed that point.

PCbuild/python.vcxproj
```diff
@@ -144,4 +145,22 @@ $(_PGOPath)
     </PropertyGroup>
     <WriteLinesToFile File="$(PySourcePath)python.bat" Lines="$(_Content)" Overwrite="true" Condition="'$(_Content)' != '$(_ExistingContent)'" />
   </Target>
+  <Target Name="GenerateLicense" AfterTargets="AfterBuild">
+    <ItemGroup>
+      <LicenseFiles Include="$(PySourcePath)LICENSE;
+                             $(PySourcePath)PC\crtlicense.txt;
+                             $(bz2Dir)LICENSE;
+                             $(opensslOutDir)LICENSE;
+                             $(tcltkDir)tcllicense.terms;
```
paulmon, line 154: @zooba This change breaks building with --no-tkinter.  Which breaks the Windows arm32 buildbot

PR #14068, opened 2019-06-14
bpo-37271: Optimize bytecode multiple times until it cannot be optimized further

pablogsal:
https://bugs.python.org/issue37271


nedbat:
@pablogsal Any chance you want to help get #13600 finished?

pablogsal:
@nedbat Sure, I will talk with Victor to see if we can push it forward.

pablogsal:
@serhiy-storchaka What do you think about this approach?

Python/peephole.c
```diff
@@ -536,3 +527,45 @@ PyCode_Optimize(PyObject *code, PyObject* consts, PyObject *names,
...
+                Py_DECREF(old_bytecode);
+            }
+            return NULL;
+        }
+        compare_bytecode = PyObject_RichCompareBool(old_bytecode, bytecode, Py_NE);
+        if (old_bytecode != code) {
+            Py_DECREF(old_bytecode);
+        }
+        if (compare_bytecode == -1) {
+            break;
```
ZackerySpytz: I don't think the `PyObject_RichCompareBool()` call is being handled properly. A return value of `-1` means that an error occurred, so `PyCode_Optimize()` should return `NULL` in that case.

Python/peephole.c
```diff
@@ -536,3 +527,45 @@ PyCode_Optimize(PyObject *code, PyObject* consts, PyObject *names,
...
+
+PyObject *
+PyCode_Optimize(PyObject *code, PyObject* consts, PyObject *names,
+                PyObject *lnotab_obj)
+{
+    PyObject* bytecode = NULL;
+    PyObject* old_bytecode = code;
+    int compare_bytecode;
+    do {
+        bytecode = old_bytecode;
```
ZackerySpytz: This assignment is unneeded.

Python/peephole.c
```diff
@@ -536,3 +527,45 @@ PyCode_Optimize(PyObject *code, PyObject* consts, PyObject *names,
...
+   single basic block.  All transformations keep the code size the same or
+   smaller.  For those that reduce size, the gaps are initially filled with
+   NOPs.  Later those NOPs are removed and the jump addresses retargeted in
+   a single pass. */
+
+PyObject *
+PyCode_Optimize(PyObject *code, PyObject* consts, PyObject *names,
+                PyObject *lnotab_obj)
+{
+    PyObject* bytecode = NULL;
```
ZackerySpytz: The declaration of `bytecode` can be moved into the loop.

Python/peephole.c
```diff
@@ -536,3 +527,45 @@ PyCode_Optimize(PyObject *code, PyObject* consts, PyObject *names,
...
+    PyObject* old_bytecode = code;
```
ZackerySpytz:
```suggestion
    PyObject *old_bytecode = code;
```

Python/peephole.c
```diff
@@ -536,3 +527,45 @@ PyCode_Optimize(PyObject *code, PyObject* consts, PyObject *names,
...
+   encoding for gaps >= 255.
+
+   Optimizations are restricted to simple transformations occurring within a
+   single basic block.  All transformations keep the code size the same or
+   smaller.  For those that reduce size, the gaps are initially filled with
+   NOPs.  Later those NOPs are removed and the jump addresses retargeted in
+   a single pass. */
+
+PyObject *
+PyCode_Optimize(PyObject *code, PyObject* consts, PyObject *names,
```
ZackerySpytz:
```suggestion
PyCode_Optimize(PyObject *code, PyObject *consts, PyObject *names,
```

Python/peephole.c
pablogsal: I am returning bytecode outside the loop, so that would need to be placed inside as well and I prefer not to do it in case the function needs some cleanup in the future.

Python/peephole.c
pablogsal: Yup, but it needs a change in the initialization of `bytecode` as `optimize_bytecode_once` cannot receive `NULL`.

pablogsal:
I have rebased and address the feedback. Also, I have added a max cap of iterations (although is not needed, it can be proven that the while loop will always finish) to make sure the time expended in the loop is always bounded.

Python/peephole.c
```diff
@@ -21,6 +21,7 @@
         (ABSOLUTE_JUMP(_Py_OPCODE(arr[i])) ? 0 : i+1))
 #define ISBASICBLOCK(blocks, start, end) \
     (blocks[start]==blocks[end])
+#define MAX_BYTECODE_OPT_ITERS 5
```
markshannon: This seems arbitrary, and quite low.
What the rationale for it being 5?

Python/peephole.c
```diff
@@ -474,10 +475,14 @@ PyCode_Optimize(PyObject *code, PyObject* consts, PyObject *names,
         unsigned int offset_delta, new_offset;
         cum_orig_offset += lnotab[i];
         assert(cum_orig_offset % sizeof(_Py_CODEUNIT) == 0);
-        new_offset = blocks[cum_orig_offset / sizeof(_Py_CODEUNIT)] *
-                sizeof(_Py_CODEUNIT);
-        offset_delta = new_offset - last_offset;
-        assert(offset_delta <= 255);
+        index = cum_orig_offset / sizeof(_Py_CODEUNIT);
```
markshannon, line 485: Why is this change necessary?

markshannon (commented):
Looks good in general, just a couple of questions.

Python/peephole.c
pablogsal: This is the average change in the biggest files when compiling the CPython test suite and code and is a number lower enough so it does not take a time increment that is big enough to notice.

Technically you can leave it unbounded because the change will always make the bytecode smaller, but this will assure that is small enough to not be a difference.

Python/peephole.c
pablogsal: Because the current line processing produces temporary indexes that violate that the `offset_delta < 255` when you run the optimizer multiple times. You can comment this change and experiment for yourself. This is the smaller and less invasive change to the peephole optimizer itself that I can think of, but I welcome any idea around this,

Python/peephole.c
markshannon: Can you add a brief comment to the code to this effect.

Python/peephole.c
markshannon: Again, can you add a brief comment to the code. thanks

Python/peephole.c
pablogsal: Will do :)

Python/peephole.c
pablogsal: Will do :)

pablogsal:
@markshannon I have added some comments as requested, check if they look good to you :)

pablogsal:
CC: @Yhg1s

pablogsal:
This is the code that is failing the assert currently:

```
def f(cond1, cond2):
    while 1:
        return 3
    while 1:
        return 5
    return 6
```

Yhg1s:
GH-15970 should fix the lnotab issue.

PR #14074, opened 2019-06-14
bpo-37278: Fix test_asyncio ProactorLoopCtrlC

vstinner:
Join the thread to prevent leaking a running thread and leaking a
reference.

Cleanup also the test:

* asyncioWindowsProactorEventLoopPolicy became the default policy,
  there is no need to set it manually.
* Only start the thread once the loop is running.
* Use a shorter sleep in the thread (100 ms rather than 1 sec).
* Use close_loop(loop) rather than loop.close().
* Use longer variable names.

https://bugs.python.org/issue37278

### DIFF
--- a/Lib/test/test_asyncio/test_windows_events.py
+++ b/Lib/test/test_asyncio/test_windows_events.py
@@ -45,20 +45,21 @@ class ProactorLoopCtrlC(test_utils.TestCase):
     def test_ctrl_c(self):
 
         def SIGINT_after_delay():
-            time.sleep(1)
+            time.sleep(0.1)
             signal.raise_signal(signal.SIGINT)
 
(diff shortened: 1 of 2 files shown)


asvetlov (approved):
Nice catch, thanks!

vstinner:
I tested manually on Windows that the test no longer leaks:

```
vstinner@WIN C:\vstinner\python\master>python -m test -R 3:3 test_asyncio -m test.test_asyncio.test_windows_events.ProactorLoopCtrlC.test_ctrl_c
Running Debug|x64 interpreter...
Run tests sequentially
0:00:00 load avg: 0.00 [1/1] test_asyncio
beginning 6 repetitions
123456
......

== Tests result: SUCCESS ==

1 test OK.

Total duration: 8 sec 31 ms
Tests result: SUCCESS
```

PR #14075, opened 2019-06-14
bpo-37279: Fix asyncio sendfile support when  extra data are sent in fallback mode.

asvetlov:
https://bugs.python.org/issue37279


aixtools:
Many Thanks!!

AIX (which lacks os.sendfile():

test_asyncio passed in 1 min 25 sec

== Tests result: SUCCESS ==

1 test OK.

PR #14079, opened 2019-06-14
bpo-28805: document METH_FASTCALL

jdemeyer:
Documentation-only patch, so should be safe for backporting to 3.7 and 3.8

CC @methane @vstinner 

https://bugs.python.org/issue28805

### DIFF
--- a/Doc/c-api/structures.rst
+++ b/Doc/c-api/structures.rst
@@ -114,10 +114,20 @@ the definition of all other Python objects.
 
 .. c:type:: PyCFunctionWithKeywords
 
-   Type of the functions used to implement Python callables in C that take
-   keyword arguments: they take three :c:type:`PyObject\*` parameters and return
-   one such value.  See :c:type:`PyCFunction` above for the meaning of the return
-   value.
+   Type of the functions used to implement Python callables in C
+   with signature :const:`METH_VARARGS | METH_KEYWORDS`.
+
+
+.. c:type:: _PyCFunctionFast
+
+   Type of the functions used to implement Python callables in C
+   with signature :const:`METH_FASTCALL`.
+
+
+.. c:type:: _PyCFunctionFastWithKeywords
+
+   Type of the functions used to implement Python callables in C
+   with signature :const:`METH_FASTCALL | METH_KEYWORDS`.
 
 
 .. c:type:: PyMethodDef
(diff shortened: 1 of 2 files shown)


Doc/c-api/structures.rst
```diff
@@ -164,13 +175,38 @@ can be combined with a binding flag.
...
+.. data:: METH_VARARGS | METH_KEYWORDS
 
    Methods with these flags must be of type :c:type:`PyCFunctionWithKeywords`.
-   The function expects three parameters: *self*, *args*, and a dictionary of
-   all the keyword arguments.  The flag must be combined with
-   :const:`METH_VARARGS`, and the parameters are typically processed using
-   :c:func:`PyArg_ParseTupleAndKeywords`.
+   The function expects three parameters: *self*, *args*, *kwargs* where
+   *kwargs* is a dictionary of all the keyword arguments or possibly *NULL*
+   if there are no keyword arguments. The parameters are typically processed
```
mangrisano: Double space after a full stop.

mangrisano (changes requested):

mangrisano (changes requested):
Hi, thank you for the pull request.
I've left a suggestion about the change. 

Thank you. :)

Doc/c-api/structures.rst
```diff
@@ -114,10 +114,20 @@ the definition of all other Python objects.
...
 
-   Type of the functions used to implement Python callables in C that take
-   keyword arguments: they take three :c:type:`PyObject\*` parameters and return
-   one such value.  See :c:type:`PyCFunction` above for the meaning of the return
-   value.
+   Type of the functions used to implement Python callables in C
+   with signature :const:`METH_VARARGS | METH_KEYWORDS`.
+
+
+.. c:type:: _PyCFunctionFast
```
methane, line 121: (just a comment) Maybe, we should remove the `_` prefix in 3.9.

methane (approved):

Doc/c-api/structures.rst
jdemeyer: > Maybe, we should remove the `_` prefix in 3.9.

Probably yes, together with doing the same for the vectorcall API.

jdemeyer:
Thanks for the merge. Please remember to backport to 3.7 and 3.8.

PR #14080, opened 2019-06-14
bpo-35998: Avoid TimeoutError in test_asyncio: test_start_tls_server_1()

asvetlov:
The test was not robust because it used to rely on `sock.shutdown(socket.SHUT_RDWR)` before `sock.close()`.

Actually, sometimes these socket state changes are merged into the just transition into the socket-closed state on the peer side. I don't want to add sleep(N) between calls, it is the easiest way to add new test flakiness.

The `start_tls` test doesn't need to check `proto.eof_received()` callback receiving, there are other explicit tests for it.

The test is changed to receive TSL encrypted data and send an acknowledgment back before client socket closing.

Now everything works smoothly, I don't see any timeout exceptions even under very heavy testing.

https://bugs.python.org/issue35998

### DIFF
--- a/Lib/test/test_asyncio/test_sslproto.py
+++ b/Lib/test/test_asyncio/test_sslproto.py
@@ -566,7 +566,7 @@ class BaseStartTLS(func_tests.FunctionalTestCaseMixin):
             addr = server.sockets[0].getsockname()
 
             with self.tcp_client(lambda sock: client(sock, addr),
-                                 timeout=self.TIMEOUT) as threaded_client:
+                                 timeout=self.TIMEOUT):
                 await asyncio.wait_for(
                     main(proto, on_con, on_con_lost),
                     timeout=self.TIMEOUT)
--- a/Lib/test/test_asyncio/test_sslproto.py
+++ b/Lib/test/test_asyncio/test_sslproto.py
@@ -494,17 +494,14 @@ class BaseStartTLS(func_tests.FunctionalTestCaseMixin):
 
     def test_start_tls_server_1(self):
         HELLO_MSG = b'1' * self.PAYLOAD_SIZE
+        ANSWER = b'answer'
 
         server_context = test_utils.simple_server_sslcontext()
         client_context = test_utils.simple_client_sslcontext()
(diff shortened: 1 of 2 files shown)


Lib/test/test_asyncio/test_sslproto.py
```diff
@@ -550,29 +550,30 @@ def connection_lost(self, exc):
...
-            proto = ServerProto(on_con, on_eof, on_con_lost)
+            proto = ServerProto(on_con, on_con_lost)
 
             server = await self.loop.create_server(
                 lambda: proto, '127.0.0.1', 0)
             addr = server.sockets[0].getsockname()
 
             with self.tcp_client(lambda sock: client(sock, addr),
-                                 timeout=self.TIMEOUT):
+                                 timeout=self.TIMEOUT) as threaded_client:
```
vstinner: the new "as" is not used

vstinner (approved):
LGTM, I just have a minor comment.

On Fedora 30, reproducing the take takes me 5 seconds using (the test hangs for a few minutes and then fail):

```
./python -m test test_asyncio -m test_start_tls_server_1 -F
```

With this change, the test no longer hangs.

I can `./python -m test test_asyncio -m test_start_tls_server_1 -F` 6 times in parallel for 5 minutes, my system load is around 7.0  (the system is very busy), and the test still doesn't hang. So I can confirm that this change fix the issue for me.

Lib/test/test_asyncio/test_sslproto.py
asvetlov: Sorry.
This is an artifact of my tries to fix this test in several ways

asvetlov: removed

vstinner (approved):
LGTM.

asvetlov:
I'll make backport myself

vstinner:
@asvetlov: I just made it :-) I really want this bug to go away, it annoy me on buildbots for 6 months at least. It was a stupid conflict on additional "loop=self.loop" in 3.7.

asvetlov:
Yes, I see.
Thank you!

PR #14081, opened 2019-06-14
bpo-19865: ctypes.create_unicode_buffer() fails on non-BMP strings on Windows

ZackerySpytz:
The unit test is based on some of the tests in test_unicode.py that check for `sizeof(c_wchar) == 2`.

https://bugs.python.org/issue19865

### DIFF
--- a/Lib/ctypes/__init__.py
+++ b/Lib/ctypes/__init__.py
@@ -274,11 +274,15 @@ def create_unicode_buffer(init, size=None):
     """
     if isinstance(init, str):
         if size is None:
-            size = len(init)+1
             if sizeof(c_wchar) == 2:
-                for c in init:
-                    if ord(c) > 0xFFFF:
-                        size += 1
+                # UTF-16 requires a surrogate pair (2 wchar_t) for non-BMP
+                # characters (outside [U+0000; U+FFFF] range). +1 for trailing
+                # NUL character.
+                size = sum(2 if ord(c) > 0xFFFF else 1 for c in init) + 1
+            else:
+                # 32-bit wchar_t (1 wchar_t per Unicode character). +1 for
+                # trailing NUL character.
+                size = len(init) + 1
         buftype = c_wchar * size
         buf = buftype()
         buf.value = init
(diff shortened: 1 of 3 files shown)


Lib/ctypes/__init__.py
```diff
@@ -275,6 +275,10 @@ def create_unicode_buffer(init, size=None):
     if isinstance(init, str):
         if size is None:
             size = len(init)+1
+            if sizeof(c_wchar) == 2:
```
vstinner, line 277: Would you mind to add a short comment to explain the purpose of this loop?

Alternative proposal:

```
if sizeof(c_wchar) == 2: 
    # UTF-16 requires a surrogate pair (2 wchar_t) for non-BMP characters
    # (outside [U+0000; U+FFFF] range). +1 for trailing NUL character
    size = sum(2 if ord(c) > 0xFFFF else 1 for c in init) + 1
else:
    # 32 bit wchar_t (1 wchar_t per Unicode character). +1 for trailing NUL character
    size = len(init) + 1 
```

Lib/ctypes/test/test_buffers.py
```diff
@@ -60,5 +60,13 @@ def test_unicode_conversion(self):
         self.assertEqual(b[::2], "ac")
         self.assertEqual(b[::5], "a")
 
+    @need_symbol('c_wchar')
+    def test_create_unicode_buffer_non_bmp(self):
+        b = create_unicode_buffer('\U00010000\U00100000')
```
vstinner: I suggest to test also the largest Unicode code point (currently):

'\U00010000\U0010ffff'

Misc/NEWS.d/next/Library/2019-06-14-08-30-16.bpo-19865.FRGH4I.rst
```diff
@@ -0,0 +1,2 @@
+:func:`ctypes.create_unicode_buffer()` no longer fails on non-BMP strings on
+Windows.
```
vstinner: According to my notes, AIX also use 16-bit wchar_t:
https://unicodebook.readthedocs.io/programming_languages.html#character-api-wchar-t

You might express it in a more positive way :-)

:func:`ctypes.create_unicode_buffer` now also supports non-BMP characters (outside [U+0000; U+FFFF] range) on platforms using 16-bit :c:type:`wchar_t` (ex: Windows, AIX).

vstinner (changes requested):

ZackerySpytz:
Thank you, @vstinner, for the review.

PR #14090, opened 2019-06-14
bpo-36707: Document "m" removed from sys.abiflags

vstinner:
https://bugs.python.org/issue36707

### DIFF
--- a/Doc/whatsnew/3.8.rst
+++ b/Doc/whatsnew/3.8.rst
@@ -962,8 +962,8 @@ Build and C API Changes
   Example of changes:
 
   * Only ``python3.8`` program is installed, ``python3.8m`` program is gone.
-  * Only ``python3.8-config`` script is installed, no ``python3.8m-config`` is
-    gone.
+  * Only ``python3.8-config`` script is installed, ``python3.8m-config`` script
+    is gone.
   * The ``m`` flag has been removed from the suffix of dynamic library
     filenames: extension modules in the standard library as well as those
     produced and installed by third-party packages, like those downloaded from
--- a/Doc/whatsnew/3.8.rst
+++ b/Doc/whatsnew/3.8.rst
@@ -965,7 +965,9 @@ Build and C API Changes
   * Only ``python3.8-config`` script is installed, no ``python3.8m-config`` is
     gone.
   * The ``m`` flag has been removed from the suffix of dynamic library
-    filenames. On Linux, for example, the Python 3.7 suffix
+    filenames: extension modules in the standard library as well as those
+    produced and installed by third-party packages, like those downloaded from
+    PyPI. On Linux, for example, the Python 3.7 suffix
     ``.cpython-37m-x86_64-linux-gnu.so`` became
     ``.cpython-38-x86_64-linux-gnu.so`` in Python 3.8.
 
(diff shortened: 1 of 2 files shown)


vstinner:
@hroncok @ned-deily: Would you mind to review this doc change?

Doc/whatsnew/3.8.rst
```diff
@@ -955,6 +955,20 @@ Optimizations
...
+  Example of changes:
+
+  * Only ``python3.8`` program is installed, ``python3.8m`` program is gone.
+  * Only ``python3.8-config`` script is installed, no ``python3.8m-config`` is
+    gone.
+  * The ``m`` flag has been removed from the suffix of dynamic library
+    filenames. On Linux, for example, the Python 3.7 suffix
+    ``.cpython-37m-x86_64-linux-gnu.so`` became
+    ``.cpython-38-x86_64-linux-gnu.so`` in Python 3.8.
+
```
ned-deily, line 973: It should also be noted that this affects extension modules in the standard library as well as those produced and installed by third-party packages, like those downloaded from PyPI.

vstinner: I updated my PR. Does it look better now?

vstinner:
I rebased my PR to fix a conflict.

ned-deily (approved):
My preference remains to not make the flags change at all but, if we must, this documents it adequately, I think.  :). Thanks!

Doc/whatsnew/3.8.rst
```diff
@@ -955,6 +955,22 @@ Optimizations
...
 
+* Default :data:`sys.abiflags` became an empty string: the ``m`` flag for
+  pymalloc became useless (builds with and without pymalloc are ABI compatible)
+  and so has been removed. (Contributed by Victor Stinner in :issue:`36707`.)
+
+  Example of changes:
+
+  * Only ``python3.8`` program is installed, ``python3.8m`` program is gone.
+  * Only ``python3.8-config`` script is installed, no ``python3.8m-config`` is
+    gone.
```
yan12125: > no ``python3.8m-config`` is gone.

Should be?

> ``python3.8m-config`` is gone.

vstinner:
> My preference remains to not make the flags change at all

Well, "m" in ABI flags became useless around Python 3.4. I don't see the point of pretending that the "m" ABI is different and then provide two names for the same thing:

* python3.8 **and** python3.8m
* python3.8-config **and** python3.8m-config
* man python3.8 **and** man python3.8m
* etc.

I'm not sure why, but on Fedora /usr/bin/python3.7 and /usr/bin/python3.7m are two separated files... with the same content (ex: same MD5 sum). Well, it's just 17 KiB each :-) /usr/bin/python3.7-config is a symlink to python3.7m-config.

The change itself is painful for distributors like Red Hat (Fedora, RHEL), but things should be simpler once the change is done.

My long-term goal is to get a single ABI for everything: CPython, PyPy, RustPython, MicroPython, etc. :-) Maybe it's not possible, but at least I would like to move towards this goal!

Overall project: https://pythoncapi.readthedocs.io/

vstinner:
Thanks @ned-deily and @yan12125 for the review ;-)

hroncok (commented):
looks good, sorry for not responding faster

PR #14109, opened 2019-06-15
Doc: Deprecation header: More precise wording.

JulienPalard:


willingc (approved):
Good change @JulienPalard. Thanks.

PR #14117, opened 2019-06-15
bpo-37251: MagicMock with function spec used as a spec in mock.patch should return a MagicMock

tirkarthi:
MagicMock with a function spec will have `__code__` object. When it is used a spec to `mock.patch` then `__code__` object which is a MagicMock is used to detect if it's a coroutine. `MagicMock.__code__ & 128` returns another MagicMock which is evaluated to True and returned as a coroutine incorrectly. Hence check `__code__` to be a `CodeType` to make sure MagicMock with `__code__` is not detected as coroutine.

https://bugs.python.org/issue37251

### DIFF
--- a/Lib/unittest/mock.py
+++ b/Lib/unittest/mock.py
@@ -46,7 +46,8 @@ FILTER_DIR = True
 _safe_super = super
 
 def _is_async_obj(obj):
-    if getattr(obj, '__code__', None):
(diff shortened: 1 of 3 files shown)


mangrisano (approved):

tirkarthi:
Closing in favor of #15830 .

PR #14133, opened 2019-06-16
gh-81489: Fix possible mojibake in mmap.mmap() when using the tagname param

ZackerySpytz:
Switch the type of *tagname* to wchar_t *. Use CreateFileMappingW().

https://bugs.python.org/issue37308

* Issue: gh-81489


ZackerySpytz:
Unfortunately, a high-quality test for this issue cannot be created, mainly because the `tagname` cannot be accessed directly through the mmap object.

Modules/mmapmodule.c
```diff
@@ -703,7 +703,7 @@ mmap__sizeof__method(mmap_object *self, void *unused)
 
     res = _PyObject_SIZE(Py_TYPE(self));
     if (self->tagname)
-        res += strlen(self->tagname) + 1;
+        res += wcslen(self->tagname) + 1;
```
zooba: The `__sizeof__` method is calculating the size in bytes of the object, but `wcslen` will return the number of characters.
```suggestion
        res += (wcslen(self->tagname) + 1) * sizeof(self->tagname[0]);
```

ZackerySpytz: I'm sorry. You're right.

Modules/mmapmodule.c
```diff
@@ -1272,7 +1272,7 @@ new_mmap_object(PyTypeObject *type, PyObject *args, PyObject *kwdict)
     DWORD off_lo;       /* lower 32 bits of offset */
     DWORD size_hi;      /* upper 32 bits of size */
     DWORD size_lo;      /* lower 32 bits of size */
-    const char *tagname = "";
+    PyObject *tagname = Py_None;
```
eryksun: This should be `NULL`, not an uncounted reference to `None`.

Modules/mmapmodule.c
```diff
@@ -1282,7 +1282,7 @@ new_mmap_object(PyTypeObject *type, PyObject *args, PyObject *kwdict)
                                 "tagname",
                                 "access", "offset", NULL };
 
-    if (!PyArg_ParseTupleAndKeywords(args, kwdict, "in|ziL", keywords,
+    if (!PyArg_ParseTupleAndKeywords(args, kwdict, "in|OiL", keywords,
```
eryksun, line 1409: Use "U" instead of "O". It does the type check for you.

eryksun:
3.8 has `_winapi.OpenFileMapping`, which we can use in a test that tries to open a section with a non-ASCII name. For example:

    name_non_ascii = f'python_test_mmap_ŝƥāɱ_{os.getpid()}'
    m = mmap.mmap(-1, 4096, tagname=name_non_ascii)
    try:
        h = _winapi.OpenFileMapping(_winapi.FILE_MAP_READ, False,
                name_non_ascii)
    finally:
        m.close()
    _winapi.CloseHandle(h)

---

The existing `test_tagname` test should be improved to use names that are more unique, such as `f'python_test_mmap_foo_{pid}'` instead of just `'foo'`. We've been lucky that most applications don't use such generic names for session kernel objects.

Modules/mmapmodule.c
ZackerySpytz: No, "U" should not be used here. `None` must be accepted as an argument for `tagname` (as [documented](https://docs.python.org/3.9/library/mmap.html#mmap.mmap)).

Modules/mmapmodule.c
ZackerySpytz: No, it should be `Py_None` (please see my other comment). Also, for what it's worth, using an uncounted reference to `Py_None` in this way is safe (it's also done in many places in the interpreter core).

Modules/mmapmodule.c
eryksun: Okay, right you are. 

Using `tagname=""` should be equivalent. Maybe map the empty string case to `m_obj->tagname = NULL`, as the current code does. But, in practice passing an empty string is effectively the same. The kernel only assigns an object name if it has non-zero length.

zooba (changes requested):
I'd like to have this fix, but I would also like the tagnames in the tests to be improved (as per Eryk Sun's last comment).

csabella:
@ZackerySpytz, please address the review comments.  Thanks!

Modules/mmapmodule.c
```diff
@@ -32,9 +32,7 @@
 #  include <unistd.h>             // close()
 #endif
 
-// to support MS_WINDOWS_SYSTEM OpenFileMappingA / CreateFileMappingA
-// need to be replaced with OpenFileMappingW / CreateFileMappingW
-#if !defined(MS_WINDOWS) || defined(MS_WINDOWS_DESKTOP) || defined(MS_WINDOWS_GAMES)
+#if !defined(MS_WINDOWS_SYSTEM)
```
erlend-aasland: @eryksun, @zooba: Is my change here correct? I pulled in main, resolved the conflicts, and did what I assumed the now removed comment meant :)

zooba: I think the entire ifdef can be removed now - it's excluding the entire implementation for those builds, which shouldn't be necessary with the updated APIs. I _believe_ this module is now supported for all Windows API set variants.

There's also going to be a reference in PC\config.c I assume, since that should have a reference to `init_mmap`.

erlend-aasland: Removed; let's see what the CI says.

zooba (approved):

PR #14142, opened 2019-06-17
bpo-37189: Export old PyRun_XXX() functions

vstinner:
Many PyRun_XXX() functions like PyRun_String() were no longer
exported in libpython38.dll by mistake. Export them again to fix the
ABI compatibiliy.

https://bugs.python.org/issue37189


vstinner:
I tested manually my fix: https://bugs.python.org/issue37189#msg345915

PR #14143, opened 2019-06-17
bpo-37312: Remove _dummy_thread and dummy_threading modules

vstinner:
Remove _dummy_thread and dummy_threading modules. These modules
were deprecated since Python 3.7 which requires threading support.

https://bugs.python.org/issue37312


tirkarthi (commented):
There are two more references. It seems there is a fixer in 2to3 to convert `dummy_thread` to `_dummy_thread`. Since it's being removed I guess its good to remove the import related fixes too.

```
Lib/lib2to3/fixes/fix_imports.py:35:           'dummy_thread': '_dummy_thread',
Lib/_compat_pickle.py:31:    'dummy_thread': '_dummy_thread',
```

2to3 renames `dummy_thread` to `_dummy_thread` but running the fixed program with this PR throws `ImportError`

```
➜  cpython git:(pr_14143) cat /tmp/foo.py
import dummy_thread
➜  cpython git:(pr_14143) ./python.exe Tools/scripts/2to3 /tmp/foo.py
RefactoringTool: Skipping optional fixer: buffer
RefactoringTool: Skipping optional fixer: idioms
RefactoringTool: Skipping optional fixer: set_literal
RefactoringTool: Skipping optional fixer: ws_comma
RefactoringTool: Refactored /tmp/foo.py
--- /tmp/foo.py	(original)
+++ /tmp/foo.py	(refactored)
@@ -1 +1 @@
-import dummy_thread
+import _dummy_thread
RefactoringTool: Files that need to be modified:
RefactoringTool: /tmp/foo.py
```

vstinner:
> There are two more references. It seems there is a fixer in 2to3 to convert dummy_thread to _dummy_thread.

Yep, I kept them on purpose. At least, it helps to make code compatible with Python 3.7.

vstinner:
PR rebased to handle a conflict.

methane (approved):

vstinner:
Thanks for the review @methane. Tests pass, I merged my PR.

PR #14148, opened 2019-06-17
bpo-35031, test_asycio: disable TLS 1.3 in test_start_tls_server_1()

vstinner:
Reintroduce workaround on Windows and FreeBSD in
test_start_tls_server_1() of test_asyncio: disable TLS v1.3 on the
client context.

https://bugs.python.org/issue35031

### DIFF
--- a/Lib/test/test_asyncio/test_sslproto.py
+++ b/Lib/test/test_asyncio/test_sslproto.py
@@ -498,6 +498,12 @@ class BaseStartTLS(func_tests.FunctionalTestCaseMixin):
 
         server_context = test_utils.simple_server_sslcontext()
         client_context = test_utils.simple_client_sslcontext()
+        if sys.platform.startswith('freebsd') or sys.platform.startswith('win'):
+            # bpo-35031: Some FreeBSD and Windows buildbots fail to run this test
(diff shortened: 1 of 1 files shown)


vstinner:
test_asyncio restarted to fail on Windows buildbots after the commit f0749da9a535375f05a2015e8960e8ae54877349 backported to 3.7 and 3.8 branches:
https://bugs.python.org/issue35998#msg345644

This PR doesn't really make the situation way worse: we use this workaround since October 2018.
https://bugs.python.org/issue35031#msg328871

If someone can have a look at the root issue, it would be great, but in the meanwhile, we need the CI to be green again (and reliable).

cc @tiran @asvetlov @1st1 @pablogsal

asvetlov (approved):
If it can fix buildbots -- I'm ok with suppressing the test

vstinner:
> If it can fix buildbots -- I'm ok with suppressing the test

Well, this PR doesn't remove the test, it only works around a failure on 2 platforms: only test TLS 1.2 on Windows and FreeBSD.

I heard that asyncio and ssl have issues on TLS 1.3 which handles TLS shutdown differently, but it seems like nobody has time to work on this issue.

PR #9747, opened 2018-10-07
bpo-34903: For datetime.datetime.strptime(), the leading zero for some two-digit formats is optional

mikegleen:
https://bugs.python.org/issue34903

PR #14149, opened 2019-06-17
bpo-34903: For datetime.datetime.strptime(), the leading zero for some two-digit formats is optional.

mikegleen:
This pull request replaces #9747 which I am closing. Sorry, couldn't figure out how to cleanly update the original.

I am adding tests to datetimetester.py following discussion with Paul Ganssle. Please let me know if this causes a problem.

https://bugs.python.org/issue34903


Misc/NEWS.d/next/Documentation/2019-06-17-09-36-46.bpo-34903.r_wGRc.rst
```diff
@@ -0,0 +1,3 @@
+For datetime.datetime.strptime(), the leading zero for some two-digit formats is optional.
```
pganssle: You are encouraged to add a "Patch by" message to your NEWS entries, to give you the well-deserved credit for the patch.

And since I am suggesting that you do this anyway, I feel less bad about bikeshedding on the content of the message: I think it can be reworded to be a bit more concise:

"Documented that in :meth:`datetime.datetime.strptime`, the leading zero in some two-digit formats is optional. Patch by Mike Gleen."

Lib/test/datetimetester.py
```diff
@@ -2501,6 +2501,47 @@ def test_strptime(self):
...
+             '6/4/2003', '%u/%V/%G', '2003-01-25T00:00:00'),
+        ]
+        def time_helper(reason, string, format, target):
+            reason = 'test single digit ' + reason
+            with self.subTest(reason=reason,
+                              string=string,
+                              format=format,
+                              target=target):
+                newdate = strptime(string, format)
+                self.assertEqual(newdate.isoformat(timespec='seconds'), target,
```
pganssle: Is there a particular reason you are converting the datetime to a string representation and then comparing that, rather than just comparing the datetimes directly?

It adds another thing that can fail that is unrelated to what you are testing here, so I think it would be best avoided unless there's something about datetime equality semantics that makes comparing the strings a better test.

Instead, I recommend something like:

```python
dt1 = self.theclass(2003, 1, 2, 4, 5, 6)
dt2 = self.theclass(2003, 2, 1)

inputs = [
    ('%d', '1/02/03 4:5:6', '%d/%m/%y %H:%M:%S', dt1),
    ('%m', '01/2/03 4:5:6', '%d/%m/%y %H:%M:%S', dt1),
    ...
    ('%V', '6/4/2003', '%u/%V/%G', dt2),
    ...
```

Then you can change the assertion to: `self.assertEqual(newdate, target, msg=reason)`.

Lib/test/datetimetester.py
```diff
@@ -2501,6 +2501,47 @@ def test_strptime(self):
...
+                                 msg=reason)
+        for testcase in inputs:
```
pganssle: Using a closure for this seems unusual to me, did you have a particular reason for doing it this way? I can see for example that it does help to isolate the individual test cases by creating a new scope for each subtest, *but* I do find it a bit harder to read.

If you didn't have any particular reason, I'm inclined to say you should just move the body of `time_helper` into this for loop. If you had a particular reason or feel strongly about it, I'm OK with leaving the closure in place, but I think you can make the test more readable by changing the following things:

1. Move the `subtest` call out of the function, so that the function body isn't as deeply nested.

2. Rename `time_helper` to something like `run_strptime_test_case` or `assert_strptime`.

Then the loop would be something like this:

```python
for testcase in inputs:
    with self.subTest(reason=testcase[0]):
        run_strptime_test_case(*testcase)
```

pganssle (dismissed):
These changes look mostly pretty good, I've commented inline with some nitpicks.

One thing I don't want to get lost in the migration from #9747 is that Brett suggested that the tests be moved to `test_strptime`. I gave my reasoning why I think `datetimetester.py` [in this comment on the original issue](https://github.com/python/cpython/pull/9747#discussion_r294086307), but I would like @brettcannon to weigh in on this before just overriding his suggestion.

Finally, I should note that once the changes are approved, I'd like to run these tests against the buildbots *before* merging. `strptime` and `strftime` are particularly sensitive to weird edge cases on the "long tail" platforms, so changes like this are very likely to need adjustment after they've been run on the buildbots.

mikegleen:
Paul,

I am ok with your nitpicks except the change to the blurb should have
parens after strptime to avoid Brett's wrath ;-). Also I'm afraid I'm at a
loss as to how to update a blurb.

I agree that using the datetime rather than a string is better.

I used the time_helper subroutine just because it was easier to write it
that way. If you think it's easier to read inline, that's fine—you're the
guy who has to read it.

I'll update and commit the changes to datetimetester.py and maybe will have
figured out how to update the blurb by then.

Regards,
Mike

mikegleen:

Misc/NEWS.d/next/Documentation/2019-06-17-09-36-46.bpo-34903.r_wGRc.rst
```diff
@@ -0,0 +1 @@
+Documented that in :meth:datetime.datetime.strptime(), the leading zero in some two-digit formats is optional. Patch by Mike Gleen.
```
pganssle:
```suggestion
Documented that in :meth:`datetime.datetime.strptime()`, the leading zero in some two-digit formats is optional. Patch by Mike Gleen.
```

This needs backticks so Sphinx can turn it into a link to the relevant method.

pganssle:
@mikegleen Thanks Mike, these changes look good to me! I made one last edit suggestion inline to fix an issue with the What's New. I'm finding it a bit hard to interpret the CI failures, but it's just the documentation builds that are failing, so it's possible that that's just a linter detecting the missing backticks.

mikegleen:
Paul,

Ok, backticks added.

Rgds,
Mike

brettcannon:
No wrath from me. 😉 At this point my review is "whatever @pganssle says".

pganssle (approved):
I ran this through the buildbots yesterday and it did not create any new failures. The only semi-related failure was on [Alpine Linux](https://buildbot.python.org/all/#/builders/149/builds/43), but that is related to an earlier PR, so it seems like this documentation is accurate for all supporting platforms.

Thanks @mikegleen!

Lib/test/datetimetester.py
```diff
@@ -2501,6 +2501,39 @@ def test_strptime(self):
                 self.assertEqual(expected, got)
 
         strptime = self.theclass.strptime
+
+        # bpo-34903: Check that single digit dates and times are allowed.
```
vstinner: If you have a comment to delim start/end of your block, it means that you have to add a new method, like:

```
def test_strptime_single_digit():
  # bpo-34903: Check that single digit dates and times are allowed.
  ...
```

vstinner (changes requested):

Lib/test/datetimetester.py
mikegleen: I'm happy to make this change but note that the code immediately preceding mine has a similar comment for bpo-34482. Could you (@vstinner) and @pganssle please confirm?

vstinner: In that case, move also the previous lines into a subtest (new method).

pganssle: @mikegleen It's fairly common for there to be stylistic and organizational inconsistencies in a file, because the codebase is large, old and has many contributors with different priorities (and different reviewers with different standards).

I tend to agree with Victor here. The main reason I didn't ask for this to be in a new test to start with was because, as you said, the current test suite tends towards these single monolithic tests and I prefer smaller tests and I wasn't sure if monolithic tests was a general policy or not. Given that Victor has much more experience than me with the project's conventions, I'm happy that he asked for this, because it will make me feel more confident in making smaller tests in the future. :smile:

Also, as an aside, I'd like to thank you for your ongoing efforts in this PR, I think we're in the home stretch :wink:

mikegleen:

Thanks to the core devs and administrators (and bots) who patiently walked me through this really minor fix. I am sure that I've benefited more learning about the dev process than Python will benefit from the actual code change.

vstinner (approved):
LGTM.

pganssle:
Thanks for the review @vstinner. I can merge this when the CI passes unless there are any other objections.

pganssle:
There was a failure in the Mac PR that looked unrelated, seemed like the multiprocessing tests hung for about 56 minutes. Assuming it wasn't just some other kind of weird fluke, I suppose it could be a race condition in the multiprocessing tests, but I don't think it's related to this PR, so I'm going to go ahead and merge.

I'll try to dig up the broken log a bit later and see if it's worth reporting.

ZackerySpytz:
@pganssle This PR should also be backported to 3.8, no?

pganssle:
@ZackerySpytz Oh jeez, I completely forgot that master isn't 3.8 anymore! Thanks!

vstinner:
The macOS issue has been reported but we have no clue so far:
https://bugs.python.org/issue37245

-- 
Night gathers, and now my watch begins. It shall not end until my death.

PR #14154, opened 2019-06-17
Fix assertions regarding magic methods function body that was not executed

tirkarthi:
The context manager object supplied to `MagicMock` has the magic methods `__aenter__` and `__aexit__` called with mock call registered but the actual function body was not executed. Since `MagicMock` is always True the attributes entered and exited returned a `MagicMock` that asserted to True in the test.

### DIFF
--- a/Lib/unittest/test/testmock/testasync.py
+++ b/Lib/unittest/test/testmock/testasync.py
@@ -361,17 +361,14 @@ class AsyncArguments(unittest.TestCase):
 
 
 class AsyncContextManagerTest(unittest.TestCase):
+
     class WithAsyncContextManager:
-        def __init__(self):
-            self.entered = False
-            self.exited = False
 
         async def __aenter__(self, *args, **kwargs):
-            self.entered = True
             return self
 
         async def __aexit__(self, *args, **kwargs):
-            self.exited = True
+            pass
 
     def test_magic_methods_are_async_mocks(self):
         mock = MagicMock(self.WithAsyncContextManager())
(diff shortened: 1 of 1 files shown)


tirkarthi:
@lisroach @mariocj89 Would be helpful to have a review of this change . This is a test only change and doesn't require a NEWS entry or an issue. I have removed the assertions as per PR description and also removed the constructor since it's not needed.

mariocj89:
I am not sure why the test is wrong, are you saying that the original code that is being passed to the mock is not being called?

tirkarthi:
> Sorry, I am not sure why the test is wrong, are you saying that the original code that is being passed to the mock is not being called? 

Yes, when an object is presented as spec to `MagicMock` then the calls to magic methods are set to MagicProxy at https://github.com/python/cpython/blob/0a28f8d379544eee897979da0ce99f0b449b49dd/Lib/unittest/mock.py#L1957

So in these tests `__enter__` and `__exit__` implementations are not called and accessing `entered` and `exited` returns MagicMock which is always True. The tests are trying to assert that entered attribute is True in boolean context with `assertTrue`. If `assertIs` is used then the below error is displayed : 

```
./python.exe -m unittest Lib.unittest.test.testmock.testasync.AsyncContextManagerTest.test_mock_supports_async_context_manager
F
FAIL: test_mock_supports_async_context_manager (Lib.unittest.test.testmock.testasync.AsyncContextManagerTest)
  File "/Users/karthikeyansingaravelan/stuff/python/cpython/Lib/unittest/test/testmock/testasync.py", line 396, in test_mock_supports_async_context_manager
    self.assertIs(mock_instance.entered, True)
AssertionError: <MagicMock name='mock.entered' id='4319777968'> is not True

Ran 1 test in 0.064s

FAILED (failures=1)
```

lisroach (approved):
You are right, thanks for catching this!

lisroach:
Closed and reopened to try to re-trigger Azure

ned-deily:
@lisroach This PR seems to have stalled.  Are you planning to merge it?

lisroach:
Yes sorry for the delay, I've gotten really busy but plan to loop back
around to this during the sprint at the latest.

tirkarthi:
Thanks Lisa :)

lisroach:
Buildbot looks like a False positive.

PR #14158, opened 2019-06-17
[3.7] bpo-37194: Complete PyObject_CallXXX() docs (GH-14156)

vstinner:
Mention explicitly that PyObject_CallXXX() functions raise an
exception an failure.

(cherry picked from commit 1ce2656f13e726b3b99d4c968926908cff1f460a)


wbolster:
@vstinner it seems the new functions are missing from `data/refcounts.dat`, causing docs [like this one for PyObject_CallNoArgs](https://docs.python.org/3.9/c-api/object.html#c.PyObject_CallNoArgs) to not have the

> Return value: New reference.

... note which is also shown for the older `PyObject_Call*` functions.

PR #14159, opened 2019-06-17
bpo-37194: Add PyObject_CallNoArgs() rationale

vstinner:
Explain in the doc why PyObject_CallNoArgs() should be preferred over
other existing ways to call a function without any arguments.


vstinner:
> It is the most efficient way to call a callable Python object without any argument.

To be fair, _PyObject_CallNoArg() should a little bit better, but it's not exposed in the public API :-D

PR #14163, opened 2019-06-17
bpo-34631: Updated OpenSSL to 1.1.1c in Windows installer

zooba:
https://bugs.python.org/issue34631


zooba:
I messed up the OpenSSL binaries build, so these bits are still 1.1.1b. I'll rerun CI once I've updated them.

PR #10596, opened 2018-11-19
bpo-25988: Do not expose abstract collection classes in the collections module.

serhiy-storchaka:
https://bugs.python.org/issue25988

PR #14171, opened 2019-06-17
bpo-37324: Remove ABC aliases from collections

vstinner:
Remove deprecated aliases to Abstract Base Classes from the
collections module.

https://bugs.python.org/issue37324

### DIFF
--- a/Doc/library/collections.rst
+++ b/Doc/library/collections.rst
@@ -33,11 +33,6 @@ Python's general purpose built-in containers, :class:`dict`, :class:`list`,
 :class:`UserString`     wrapper around string objects for easier string subclassing
 =====================   ====================================================================
 
-.. deprecated-removed:: 3.3 3.9
-    Moved :ref:`collections-abstract-base-classes` to the :mod:`collections.abc` module.
-    For backwards compatibility, they continue to be visible in this module through
-    Python 3.8.
-
 
 :class:`ChainMap` objects
 -------------------------
--- a/Doc/whatsnew/3.9.rst
+++ b/Doc/whatsnew/3.9.rst
@@ -122,9 +122,13 @@ Deprecated
 Removed
 =======
 
-``_dummy_thread`` and ``dummy_threading`` modules have been removed. These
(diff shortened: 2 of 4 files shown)


vstinner:
pip is broken by this change, causing test_venv to fail: https://bugs.python.org/issue37324#msg345906

tirkarthi:
Vendored html5lib in pip was fixed upstream but needs a new release to fix thjs. See also discussion at https://github.com/python/cpython/pull/10596

vstinner:
Oh. I wasn't aware of the previous attempt: https://github.com/python/cpython/pull/10596 there is a long discussion there.

> Vendored html5lib in pip was fixed upstream but needs a new release to fix thjs. See also discussion at #10596

I also found some info about that: https://bugs.python.org/issue37324#msg345908

vstinner:
It seems like jinja2 should be updated:

```
/home/vstinner/prog/python/master/env/lib/python3.9/site-packages/jinja2/utils.py:485: DeprecationWarning: Using or importing the ABCs from 'collections' instead of from 'collections.abc' is deprecated since Python 3.3,and in 3.9 it will stop working
  from collections import MutableMapping

/home/vstinner/prog/python/master/env/lib/python3.9/site-packages/jinja2/runtime.py:318: DeprecationWarning: Using or importing the ABCs from 'collections' instead of from 'collections.abc' is deprecated since Python 3.3,and in 3.9 it will stop working
  from collections import Mapping

/home/vstinner/prog/python/master/env/lib/python3.9/site-packages/jinja2/sandbox.py:82: DeprecationWarning: Using or importing the ABCs from 'collections' instead of from 'collections.abc' is deprecated since Python 3.3,and in 3.9 it will stop working
  from collections import MutableSet, MutableMapping, MutableSequence
```

vstinner:
This PR cannot be merged because it breaks the Python CI because of html5lib (used by pip): https://bugs.python.org/issue37324#msg348011 I prefer to close this PR right now. I will not be available next  weeks, so I prefer to close the PR. I may reopen it later, once html5lib and pip are fixed.

PR #14179, opened 2019-06-18
bpo-35360: Update Windows builds to use SQLite 3.28.0

ghost:
https://bugs.python.org/issue35360


animalize:
`macOS PR Tests` has 1 test failed:  `test_multiprocessing_spawn`

It should be unrelated to this patch.

PR #14180, opened 2019-06-18
bpo-35360: Update macOS installer to use SQLite 3.28.0

ghost:
https://bugs.python.org/issue35360

### DIFF
--- a/Misc/NEWS.d/next/Build/2019-06-18-08-58-30.bpo-35360.-CWbfy.rst
+++ b/Misc/NEWS.d/next/Build/2019-06-18-08-58-30.bpo-35360.-CWbfy.rst
@@ -1,0 +1,1 @@
+Update OS X installer to use SQLite 3.28.0.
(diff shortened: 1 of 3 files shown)


Misc/NEWS.d/next/Build/2019-06-18-08-58-30.bpo-35360.-CWbfy.rst
```diff
@@ -0,0 +1 @@
+Update OS X installer to use SQLite 3.28.0.
```
ned-deily: Apple now uses "macOS" instead of "OS X" and we should, too.  Also, please put the NEWS entry in the "macOS" section rather than the "Build" section.

ned-deily (changes requested):
Thanks for the PR.  If you could make the requested changes for the NEWS entry, we can merge this in.

animalize:

ned-deily (approved):

PR #14186, opened 2019-06-18
bpo-37328: remove deprecated HTMLParser.unescape

methane:
It is deprecated since Python 3.4

https://bugs.python.org/issue37328

### DIFF
--- a/Lib/html/parser.py
+++ b/Lib/html/parser.py
@@ -9,7 +9,6 @@
 
 
 import re
-import warnings
 import _markupbase
 
 from html import unescape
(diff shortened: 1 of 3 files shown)


methane:
Oh! setuptools still depends on this, accidentally.

https://github.com/pypa/setuptools/pull/1788

methane:
https://github.com/pypa/setuptools/pull/1788 is fixed.
I wait new setuptools is released and bundled setuptools is upgraded to >=41.0.2

Lib/html/parser.py
```diff
@@ -9,7 +9,6 @@
 
 
 import re
-import warnings
 import _markupbase
 
 from html import unescape
```
ezio-melotti, line 14: I think this could be removed too if it's no longer needed (even though removing it might break some code if people are doing `from html.parser import unescape` instead of `from html import unescape`).

methane: `unescape` is used in this module several times.

pradyunsg:
#15483 bumps the setuptools version, so that'll unblock this once merged.

PR #14193, opened 2019-06-18
[3.8] bpo-37250: put back tp_print for backwards compatibility

jdemeyer:
PR for 3.8 only, since this is supposed to be a temporary compatibility measure.

https://bugs.python.org/issue37250


ncoghlan:
+1, but let's refrain from merging until there's clearer consensus on the tracker issue for the pragmatic Py3.8-only source compatibility fix.

Include/cpython/object.h
```diff
@@ -256,6 +256,9 @@ typedef struct _typeobject {
     destructor tp_finalize;
     vectorcallfunc tp_vectorcall;
 
+    /* Unused, kept for backwards compatibility in CPython 3.8 only */
+    int (*tp_print)(PyObject *, FILE *, int);
```
ncoghlan: Does adding the `Py_DEPRECATED(3.8)` macro here cause the `tp_print` reference in the test to generate a compiler warning?

If it does, that would be a good thing to add (as long as we also have a way to disable the deprecation warning in the test suite, which I presume we do)

Reference: https://docs.python.org/3.8/c-api/intro.html#c.Py_DEPRECATED

Misc/NEWS.d/next/C API/2019-06-12-11-45-36.bpo-37221.RhP1E7.rst
```diff
@@ -0,0 +1,3 @@
+``tp_print`` is put back at the end of the ``PyTypeObject`` structure
+to restore support for old code (in particular generated by Cython)
+setting ``tp_print = 0``.
```
encukou, line 3:
```suggestion
setting ``tp_print = 0``.
Note that ``tp_print`` will be removed entirely in Python 3.9.
```

Include/cpython/object.h
```diff
@@ -256,6 +256,9 @@ typedef struct _typeobject {
     destructor tp_finalize;
     vectorcallfunc tp_vectorcall;
 
+    /* Unused, kept for backwards compatibility in CPython 3.8 only */
```
vstinner: Please mention that it will be removed from Python 3.8. You may add "bpo-37250: " prefix.

Modules/_testcapimodule.c
```diff
@@ -6008,6 +6008,10 @@ PyInit__testcapi(void)
     Py_INCREF(&MyList_Type);
     PyModule_AddObject(m, "MyList", (PyObject *)&MyList_Type);
 
+    /* Old Cython code sets tp_print to 0, we check that
+     * this doesn't break anything. */
```
vstinner, line 6012: Please mention "bpo-37250: " here.

Include/cpython/object.h
jdemeyer: That works but then the deprecation warning will be shown when compiling `_testcapi.c`. I'll add a guard for that. If that's controversial, I'll gladly remove it.

(off-topic: since this is the only such warning shown, it seems that we never actually test deprecated functionality of the C API, which is worrying)

Include/pyport.h
```diff
@@ -510,6 +510,7 @@ extern "C" {
  *    Py_DEPRECATED(3.4) typedef int T1;
  *    Py_DEPRECATED(3.8) PyAPI_FUNC(int) Py_OldFunction(void);
  */
+#ifndef Py_DEPRECATED
```
vstinner: Why do you modify pyport.h?

vstinner: IMHO it's fine if compiling _testcapimodule.c emits a deprecation warning. If it becomes an issue, we can just remove the test.

jdemeyer: > IMHO it's fine if compiling _testcapimodule.c emits a deprecation warning.

The more warnings are shown, the less value there is in those warnings. A single warning during compilation stands out, an additional warning in between many not. So we should strive for compiling CPython without any warnings.

> If it becomes an issue, we can just remove the test.

I think that's the worst solution. We typically test deprecated functionality in pure Python modules, why shouldn't we test deprecated C API?

ncoghlan: Reading https://stackoverflow.com/questions/13459602/how-can-i-get-rid-of-deprecated-warnings-in-deprecated-functions-in-gcc/ suggests there are a few ways we can go here:

1. Mark the _testcapi init function itself as deprecated, so compilers won't care that it accesses deprecated symbols.
2. Start defining compiler specific `_Py_PUSH_PRAGMA_IGNORE_DEPRECATION_WARNINGS` and `_Py_POP_PRAGMA` macros and use those to selectively silence compiler warnings
3. Use the testcapi warnings to check that deprecated C APIs actually *are* deprecated (doing that long term might involve pulling out a separately compiled module for those though, so we can more easily distinguish them from actual warnings in the main interpreter and standard library build process)

For right now, I'd suggest seeing if the warning goes away if you mark the testcapi module's init function as deprecated.

scoder (approved):

willingc (approved):
Thanks @jdemeyer. This looks good to me. @encukou @ncoghlan Do one of you wish to review and merge?

encukou (approved):

encukou:
Will do. Thanks to everyone involved!

PR #11500, opened 2019-01-10
bpo-35704: Prevent test_shutil fail result when AIX is 32-bit and MAXDATA < 0x20000000

aixtools:
https://bugs.python.org/issue35704

PR #14194, opened 2019-06-18
[3.7] bpo-35704: Fix test_shutil result for 32-bit AIX when MAXDATA < 0x20000000 [GH-11500]

aixtools:
Forgot to request backport previously.

https://bugs.python.org/issue35704


ned-deily:
Thanks for preparing this PR but I think the changes proposed are not appropriate at this late stage in 3.7's lifecycle as there have been a number of subsequent unrelated merges into this module since the PR was proposed. Let's focus on 3.8.

PR #9687, opened 2018-10-03
bpo-34711: Return HTTPStatus.NOT_FOUND if path.endswith(/) and not a directory

aixtools:
Please provide guidelines on how much, if any comments, to have in Lib/http/server.py.

I assume there is too much now, while my gut says no comments may only make people wonder why this is being done.

https://bugs.python.org/issue34711

PR #14197, opened 2019-06-18
[3.7] bpo-34711: Return HTTPStatus.NOT_FOUND if path.endswith(/) and not a directory (GH-9687)

aixtools:
https://bugs.python.org/issue34711


ned-deily:
Thanks for preparing this PR but I think the changes proposed are not appropriate at this late stage in 3.7's lifecycle. Let's focus on 3.8.

PR #14202, opened 2019-06-18
[2.7] bpo-37329: valgrind: ignore _PyWarnings_Init false alarms

vstinner:
_PyWarnings_Init() only allocates memory once at startup but it is
not released at exit. Ignore this issue to be able to catch other
bugs more easily.

https://bugs.python.org/issue37329

### DIFF
--- a/Misc/valgrind-python.supp
+++ b/Misc/valgrind-python.supp
@@ -104,7 +104,7 @@
 }
 
 {
-   bpo-37329: _PyWarnings_Init allocates memory at startup, but does't release it at exit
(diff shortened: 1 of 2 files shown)


methane (approved):

Misc/valgrind-python.supp
```diff
@@ -103,6 +103,14 @@
    fun:COMMENT_THIS_LINE_TO_DISABLE_LEAK_WARNING
 }
 
+{
+   bpo-37329: _PyWarnings_Init allocates memory at startup, but does't release it at exit
```
stratakis: Typo: does't -> doesn't

PR #14209, opened 2019-06-18
bpo-37325: Fix focus traversal for 2 IDLE dialogs

terryjreedy:
Tab now moves focus across and down for Help Source and Custom Run.

https://bugs.python.org/issue37325


terryjreedy:
A simple change works for me and this should not depend on OS, so I will merge when CI passes.

PR #14221, opened 2019-06-19
bpo-36710: Remove PyImport_Cleanup() function

vstinner:
* Rename PyImport_Cleanup() to _PyImport_Cleanup() and move it to the
  internal C API. Add 'tstate' parameters.
* Remove documentation of _PyImport_Init(), PyImport_Cleanup(),
  _PyImport_Fini(). All three were documented as "For internal use
  only.".

### DIFF
--- a/Doc/c-api/import.rst
+++ b/Doc/c-api/import.rst
@@ -223,21 +223,6 @@ Importing Modules
    Return a new reference to the finder object.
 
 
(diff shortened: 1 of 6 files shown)


vstinner:
> https://buildbot.python.org/all/#builders/141/builds/2029

https://bugs.python.org/issue37338

PR #14228, opened 2019-06-19
bpo-37337: Add _PyObject_VectorcallMethod()

jdemeyer:

### DIFF
--- a/Doc/c-api/object.rst
+++ b/Doc/c-api/object.rst
@@ -395,6 +395,9 @@ Object Protocol
    argument 1 (not 0) in the allocated vector.
    The callee must restore the value of ``args[-1]`` before returning.
 
+   For :c:func:`_PyObject_VectorcallMethod`, this flag means instead that
+   ``args[0]`` may be changed.
+
    Whenever they can do so cheaply (without additional allocation), callers
    are encouraged to use :const:`PY_VECTORCALL_ARGUMENTS_OFFSET`.
    Doing so will allow callables such as bound methods to make their onward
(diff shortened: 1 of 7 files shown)


jdemeyer:
CC @vstinner since this is increasing the size of the C API.

Doc/c-api/object.rst
```diff
@@ -353,6 +353,15 @@ Object Protocol
    *NULL* on failure.
 
 
+.. c:function:: PyObject* PyObject_CallMethodObjNoArgs(PyObject *obj, PyObject *name)
```
methane: Please note that `ObjArgs` means "called with a variable number of `PyObject*` arguments".
This suffix is added because `PyObject_CallMethod` uses `Py_BuildValue` style arguments.

On the other hand, this API doesn't accept any arguments.  `CallMethodNoArgs` makes more sense.

jdemeyer: OK, I changed the name.

vstinner (changes requested):
This PR adds too many functions at once, it's hard to see how each function is used. Would it be possible to start by adding _PyObject_VectorcallMethod() and its _PyObject_VectorcallMethodId() variant.

Only once it's merged, propose a new PR to only add the private _PyObject_CallMethodNoArgs() and its _PyObject_CallMethodIdNoArgs() variant.

Adding a public PyObject_CallMethodNoArgs() should be the last PR. You should elaborate "efficiency" rationale: measure how many times it's called, the performance benefit (if any), effect on the stack usage, etc.

jdemeyer:
> Would it be possible to start by adding _PyObject_VectorcallMethod() and its _PyObject_VectorcallMethodId() variant.

Sure. It's just that sometimes you actually need to use the function in more places to see which effect it has. If I would have made a PR adding `_PyObject_VectorcallMethodId` but never using it, people might complain about adding a new function which isn't used anywhere.

But now that we see that these functions would certainly be used, we can indeed split this up.

I assume that by asking this, you agree to add the functions `_PyObject_VectorcallMethod()` and `_PyObject_VectorcallMethodId()`.

vstinner:
I suggest to only add _PyObject_VectorcallMethod() and its _PyObject_VectorcallMethodId() variant, but also use them. My mean: remove the 3 other functions.

jdemeyer:
So what should I do with all the changes of the form
```diff
-    loop = _PyObject_CallMethodId(policy, &PyId_get_event_loop, NULL);
+    loop = _PyObject_CallMethodIdNoArgs(policy, &PyId_get_event_loop);

jdemeyer:
> I suggest to only add _PyObject_VectorcallMethod() and its _PyObject_VectorcallMethodId() variant, but also use them.

Done.

methane (approved):

jdemeyer:
You set this to `awaiting merge`, so you plan to merge it? Or are you waiting for more reviews?

methane:
Bot changed the label, not l.
It looks good to me.  But I'm waiting for other core-dev's review.

@vstinner Would you check this again?

jdemeyer:
`I have made the requested changes; please review again`

PR #8954, opened 2018-08-27
bpo-11192: Fix test_socket.py for AIX

aixtools:
Add "SkipIf" statements for unsupported features
Use run-time platform.system() to establish active platform rather than build-time sys.platform()

https://bugs.python.org/issue11192

PR #14237, opened 2019-06-19
[3.7] bpo-11192: Fix test_socket.py for AIX (GH-8954)

aixtools:
https://bugs.python.org/issue11192


ned-deily:
Thanks for preparing this PR but I think the changes proposed are not appropriate at this late stage in 3.7's lifecycle. Let's focus on 3.8.

PR #14241, opened 2019-06-19
bpo-37342: Fix the incorrect nb_index's type in typeobj documentation

shihai1991:
It was listed as `binaryfunc`. It should be `unaryfunc`.

### DIFF
--- a/Doc/c-api/typeobj.rst
+++ b/Doc/c-api/typeobj.rst
@@ -285,7 +285,7 @@ sub-slots
    +---------------------------------------------------------+-----------------------------------+--------------+
    | :c:member:`~PyNumberMethods.nb_inplace_true_divide`     | :c:type:`binaryfunc`              | __truediv__  |
    +---------------------------------------------------------+-----------------------------------+--------------+
-   | :c:member:`~PyNumberMethods.nb_index`                   | :c:type:`unaryfunc`              | __index__    |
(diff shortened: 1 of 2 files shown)


tirkarthi (approved):
LGTM. It's listed as unaryfunc ".. c:member:: unaryfunc PyNumberMethods.nb_index" in https://github.com/python/cpython/blob/master/Doc/c-api/typeobj.rst line 2069.

tirkarthi:
Please fix the CI error. I think the last column with "__index__" at https://travis-ci.org/python/cpython/jobs/547863960#L399 needs a space so that it's aligned with the rest of columns in the table as shown in the CI error log.

shihai1991:
> Please fix the CI error. I think the last column with "**index**" at https://travis-ci.org/python/cpython/jobs/547863960#L399 needs a space so that it's aligned with the rest of columns in the table as shown in the CI error log.
> 
> Edit: I think NEWS entry is not needed for this doc change

Good catch. Let us try it ;)

Mariatta (approved):
Thanks. I confirmed it is `unaryfunc` in https://github.com/python/cpython/blob/af41c567af81de7c4408e2e2617f1d3747408434/Include/cpython/object.h#L136. News file is not needed.

tirkarthi:
Thank you @shihai1991

PR #14250, opened 2019-06-19
bpo-36710: Add tstate parameter in _warnings.c

vstinner:
_Warnings_GetState() now requires tstate to be non-NULL.


vstinner:
I'm not sure that this change is really useful. Maybe I gone too far in https://bugs.python.org/issue36710 :-) I reject my PR.

PR #14251, opened 2019-06-19
bpo-36511 add arm32 buildbot diagnostic messages

paulmon:
I'm trying figure out why ssh.exe doesn't work on Windows when the buildbot worker is running.

The path includes `C:\WINDOWS\System32\OpenSSH\` and the location of `ssh.exe` on my machine is `C:\WINDOWS\System32\OpenSSH\ssh.exe`

Previously the script had `ssh.exe` expecting to find it on the path. But at runtime there was this error message: `'ssh' is not recognized as an internal or external command`

So I tried adding `if "%SSH%"=="" if EXIST %WINDIR%\System32\OpenSSH\ssh.exe (set SSH=%WINDIR%\System32\OpenSSH\ssh.exe)` and calling `%SSH%` in the script

Now it echos `SSH = ` meaning that the test for `EXIST` failed.  From the same console window that I'm launching the worker from ssh.exe is on the path and works fine.

Any ideas?  @zware @zooba

EDIT: Steve pointed out in the comments of #13454 that it might be 32-bit vs 64-bit issue.  I had 32-bit python installed and ssh.exe appears to be 64-bit only. Trying to update the worker to python 64-bit

### DIFF
--- a/Tools/buildbot/remoteDeploy.bat
+++ b/Tools/buildbot/remoteDeploy.bat
@@ -2,7 +2,7 @@
 rem Used by the buildbot "remotedeploy" step.
 setlocal
 
-set PATH=%PATH%;%SystemRoot%\SysNative\OpenSSH
+set PATH=%PATH%;%SystemRoot%\SysNative\OpenSSH;%SystemRoot%\System32\OpenSSH
 set here=%~dp0
 set arm32_ssh=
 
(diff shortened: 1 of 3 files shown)


paulmon:
I tried skimming the buildbot documentation before posting this at https://devguide.python.org/buildworker/ and https://docs.buildbot.net/current/manual/index.html and I don't see anything obvious that needs to be configured to allow calling ssh.exe

zooba:
It looks like `C:\Windows\System32\OpenSSH` exists, but not `C:\Windows\SysWOW64\OpenSSH`, which means if you hit 32-bit file system redirection then it's not going to find it by that path.

Maybe add fallback checks for `C:\Windows\SysNative\OpenSSH`. You can probably also simplify the code a bit by just extending PATH with both of these and then assuming that `ssh` and `scp` are available.

zooba:
You can test this by launching `C:\Windows\SysWOW64\cmd.exe` explicitly and running from there. That'll give you the 32-bit command prompt

paulmon:
still testing these changes on the buildbot worker machine

Tools/buildbot/remoteDeploy.bat
```diff
@@ -2,6 +2,7 @@
 rem Used by the buildbot "remotedeploy" step.
 setlocal
 
+set PATH=%PATH%;%SystemRoot%\SysNative\OpenSSH
```
zooba: Unfortunately, `SysNative` only works in 32-bit command prompt. Add both `SysNative` and `System32` here to ensure it works even if we switch to a 64-bit process one day.

paulmon: System32\OpenSSH is always on the path right now, whether I run C:\Windows\SysWOW64\cmd.exe or C:\Windows\System32\cmd.exe.  I can add it again, but it seemed redundant

zooba (approved):

PR #14255, opened 2019-06-20
bpo-29505: Add fuzzing for re.compile, re.load and csv.reader

ammaraskar:
@gpshead This should finish up the original bpo ticket, adds fuzzing for all the high-risk functions likely to be exposed to user input. Please let me know if I missed anything you intended to have fuzzed in the original ticket. Current list:

- `float(x)`
- `int(x)`
- `str(x)`
- `json.loads(x)`
- `re.compile(x)`
- `re.match(..., x)`
- `csv.reader(x)`

I would have liked to combine the `re.compile` and `re.match` tests but what ended up happening was we would generate some catastrophically slow regex like `^(A+)*B` and then `"A" * 25` will cause a timeout.


Modules/_xxtestfuzz/fuzzer.c
```diff
@@ -115,6 +115,155 @@ static int fuzz_json_loads(const char* data, size_t size) {
...
+const size_t NUM_PATTERNS = sizeof(regex_patterns) / sizeof(regex_patterns[0]);
+PyObject** compiled_patterns = NULL;
+/* Fuzz re.match(x) */
+static int fuzz_sre_match(const char* data, size_t size) {
+    if (size < 1 || size > MAX_RE_TEST_SIZE) {
+        return 0;
+    }
+    /* Use the first byte as a uint8_t specifying the index of the
+       regex to use */
+    uint8_t idx = ((uint8_t*) data)[0];
```
drj11: This is unsafe. Safer would be:

    uint8_t idx = (uint8_t)(data[0]);

(at any rate, this is defined). But if you want an unsigned type that is guaranteed to hold all of the values of a char, then you want `unsigned char` instead of `uint8_t`.

ammaraskar: Thank you, `unsigned char` makes way more sense, done.

Out of curiosity, why is accessing the element and then casting safer than casting the pointer and then accessing the element?

Modules/_xxtestfuzz/fuzzer.c
```diff
@@ -115,6 +115,155 @@ static int fuzz_json_loads(const char* data, size_t size) {
...
+
+    Py_DECREF(pattern_bytes);
+    Py_DECREF(flags_obj);
+    Py_XDECREF(compiled);
+    return 0;
+}
+
+/* Some random patterns used to test re.match.
+   Be careful not to add catostraphically slow regexes here, we want to
+   excercise the matchign code without causing timeouts.*/
```
gpshead: matching

Modules/_xxtestfuzz/fuzzer.c
```diff
@@ -171,7 +314,56 @@ int LLVMFuzzerTestOneInput(const uint8_t *data, size_t size) {
...
+    /* Import json.loads */
+    if (json_loads_method == NULL) {
+        PyObject* json_module = PyImport_ImportModule("json");
+        json_loads_method = PyObject_GetAttrString(json_module, "loads");
+    }
+
     rv |= _run_fuzz(data, size, fuzz_json_loads);
+#endif
+#if !defined(_Py_FUZZ_ONE) || defined(_Py_FUZZ_fuzz_sre_compile)
+    /* Impore sre_compile.compile and sre.error */
```
gpshead: Import

Modules/_xxtestfuzz/fuzzer.c
```diff
@@ -171,7 +314,56 @@ int LLVMFuzzerTestOneInput(const uint8_t *data, size_t size) {
...
+        SRE_FLAG_DEBUG = PyLong_AsLong(
+            PyObject_GetAttrString(sre_constants, "SRE_FLAG_DEBUG"));
+    }
+
+    rv |= _run_fuzz(data, size, fuzz_sre_compile);
+#endif
+#if !defined(_Py_FUZZ_ONE) || defined(_Py_FUZZ_fuzz_sre_match)
+    /* Precompile all the regex patterns on the first run for faster fuzzing */
+    if (compiled_patterns == NULL) {
+        PyObject* re_module = PyImport_ImportModule("re");
```
gpshead: being pedantic: check re_module for NULL.

Modules/_xxtestfuzz/fuzzer.c
```diff
@@ -171,7 +314,56 @@ int LLVMFuzzerTestOneInput(const uint8_t *data, size_t size) {
...
+        PyObject* json_module = PyImport_ImportModule("json");
+        json_loads_method = PyObject_GetAttrString(json_module, "loads");
+    }
+
     rv |= _run_fuzz(data, size, fuzz_json_loads);
+#endif
+#if !defined(_Py_FUZZ_ONE) || defined(_Py_FUZZ_fuzz_sre_compile)
+    /* Impore sre_compile.compile and sre.error */
+    if (sre_compile_method == NULL) {
+        PyObject* sre_compile_module = PyImport_ImportModule("sre_compile");
```
gpshead: pedantic: check sre_compile_module for NULL

Modules/_xxtestfuzz/fuzzer.c
```diff
@@ -171,7 +314,56 @@ int LLVMFuzzerTestOneInput(const uint8_t *data, size_t size) {
...
+        sre_compile_method = PyObject_GetAttrString(sre_compile_module, "compile");
+
+        PyObject* sre_constants = PyImport_ImportModule("sre_constants");
```
gpshead: NULL check

Modules/_xxtestfuzz/fuzzer.c
gpshead: the same goes for all of your GetAttrString return values.

Modules/_xxtestfuzz/fuzzer.c
```diff
@@ -171,7 +314,56 @@ int LLVMFuzzerTestOneInput(const uint8_t *data, size_t size) {
...
+            compiled_patterns[i] = compiled;
+        }
+    }
+
+    rv |= _run_fuzz(data, size, fuzz_sre_match);
+#endif
+#if !defined(_Py_FUZZ_ONE) || defined(_Py_FUZZ_fuzz_csv_reader)
+    /* Import csv and csv.Error */
+    if (csv_module == NULL) {
+        csv_module = PyImport_ImportModule("csv");
```
gpshead: NULL checks

gpshead (commented):
overall, looks good.  just some pedantry around NULL checks after CPython API calls even if you think they're unlikely to be taken.

Modules/_xxtestfuzz/fuzzer.c
```diff
@@ -115,6 +124,215 @@ static int fuzz_json_loads(const char* data, size_t size) {
...
+        Py_DECREF(pattern_bytes);
+        return 0;
+    }
+
+    /* compiled = _sre.compile(data[2:], data[0:2] */
+    PyObject* compiled = PyObject_CallFunctionObjArgs(
+        sre_compile_method, pattern_bytes, flags_obj, NULL);
+    /* Ignore ValueError as the fuzzer will more than likely
+       generate some invalid combination of flags */
+    if (compiled == NULL && PyErr_ExceptionMatches(PyExc_ValueError)) {
```
gpshead, line 189: this series if if's with PyErr_ExceptionMatches calls seems somewhat redundant (the comments for each match type are useful).  It can also call PyErr_ExceptionMatches after PyErr_Clear has been called as these are sequential if's not else if's.  perhaps write it as:

```
if (compiled == NULL) {
  if (PyErr_ExceptionMatches(xxx) ||
      PyErr_ExceptionMatches(yyy) ||
      ...) {
  }
}
```
with the comments interspersed.

ammaraskar: Agreed, the comments were primarily why I went with this style but I'll see if I can put them between the ORs.

ammaraskar: Done, kinda iffy about the formatting. Got any suggestions?

```c
if (PyErr_ExceptionMatches(xxx) ||
    PyErr_ExceptionMatches(yyy)
    ) {
    PyErr_Clear();
    }
}
```
and
```c
if (PyErr_ExceptionMatches(xxx) ||
    PyErr_ExceptionMatches(yyy)) {
    PyErr_Clear();
    }
}
```
both look really confusing, so I went with some alternative style.

PR #14258, opened 2019-06-20
bpo-37345: Add formal UDPLITE support

LivInTheLookingGlass:
At the moment you can definitely use UDPLITE sockets on Linux systems, but it would be good if this support were formalized such that you can detect support at runtime easily.

At the moment, to make and use a UDPLITE socket requires something like the following code:

```
>>> import socket
>>> a = socket.socket(socket.AF_INET, socket.SOCK_DGRAM, 136)
>>> b = socket.socket(socket.AF_INET, socket.SOCK_DGRAM, 136)
>>> a.bind(('localhost', 44444))
>>> b.sendto(b'test'*256, ('localhost', 44444))
>>> b.setsockopt(136, 10, 16)
>>> b.sendto(b'test'*256, ('localhost', 44444))
>>> b.setsockopt(136, 10, 32)
>>> b.sendto(b'test'*256, ('localhost', 44444))
>>> b.setsockopt(136, 10, 64)
>>> b.sendto(b'test'*256, ('localhost', 44444))
```

If you look at this through Wireshark, you can see that the packets are different in that the checksums and checksum coverages change.

With the pull request that I am submitting momentarily, you could do the following code instead:

```
>>> import socket
>>> a = socket.socket(socket.AF_INET, socket.SOCK_DGRAM, socket.IPPROTO_UDPLITE)
>>> b = socket.socket(socket.AF_INET, socket.SOCK_DGRAM, socket.IPPROTO_UDPLITE)
>>> a.bind(('localhost', 44444))
>>> b.sendto(b'test'*256, ('localhost', 44444))
>>> b.set_send_checksum_coverage(16)
>>> b.sendto(b'test'*256, ('localhost', 44444))
>>> b.set_send_checksum_coverage(32)
>>> b.sendto(b'test'*256, ('localhost', 44444))
>>> b.set_send_checksum_coverage(64)
>>> b.sendto(b'test'*256, ('localhost', 44444))
```

One can also detect support for UDPLITE just by checking

```
>>> hasattr(socket, 'IPPROTO_UDPLITE')
```

https://bugs.python.org/issue37345


Doc/library/socket.rst
```diff
@@ -200,6 +200,16 @@ created.  Socket addresses are represented as follows:
 
   .. versionadded:: 3.8
 
+- :const:`IPPROTO_UDPLITE` is a variant on UDP which allows you to specify
+  what portion of a packet to cover with the checksum. It adds the methods
+  :meth:`socket.set_send_checksum_coverage` to change what portion of
+  outgoing packets are covered and :meth:`socket.set_recv_checksum_coverage`
+  to filter out packets which cover too little of their data.
+
+  ..availability:: Linux >= 2.6.20
```
mangrisano: It should be `.. availability::`. Therefore the tests don't pass :)

mangrisano (changes requested):
Hi, thank you for the pull request.
I've left a suggestion about the changes.

Thank you :)

Doc/library/socket.rst
LivInTheLookingGlass: Should be fixed in most recent commits

Modules/socketmodule.c
```diff
@@ -7737,6 +7737,17 @@ PyInit__socket(void)
 #else
     PyModule_AddIntConstant(m, "IPPROTO_UDP", 17);
 #endif
+#ifdef  IPPROTO_UDPLITE
+    PyModule_AddIntMacro(m, IPPROTO_UDPLITE);
+    #ifndef UDPLITE_SEND_CSCOV
+        #define UDPLITE_SEND_CSCOV 10
```
asvetlov, line 7743: I believe if `IPPROTO_UDPLITE` is defined then `UDPLITE_SEND_CSCOV` / `UDPLITE_RECT_CSCOV` are defined too.
What is the reason to use custom numeric constants?

asvetlov (changes requested):
Please add NEWS entry: https://devguide.python.org/committing/?highlight=blurb#what-s-new-and-news-entries

Doc/library/socket.rst
```diff
@@ -200,6 +200,23 @@ created.  Socket addresses are represented as follows:
 
   .. versionadded:: 3.8
 
+- :const:`IPPROTO_UDPLITE` is a variant on UDP which allows you to specify
```
asvetlov:
```suggestion
- :const:`IPPROTO_UDPLITE` is a variant of UDP which allows you to specify
```
on -> of maybe?

Doc/library/socket.rst
```diff
@@ -200,6 +200,23 @@ created.  Socket addresses are represented as follows:
...
+  what portion of a packet to cover with the checksum. It adds the methods
```
asvetlov:
```suggestion
  what portion of a packet is covered with the checksum. It adds the methods
```
Not sure about the proposed text but it looks a little more coherent to me.

asvetlov: Please rephrase.
`IPPROTO_UDPLITE` doesn't `add the methods` but allows to use `UDPLITE_SEND_CSCOV` / `UDPLITE_RECV_CSCOV` in `setsockopt`

Doc/library/socket.rst
LivInTheLookingGlass: Sorry, I think that was a regionalism on my part. Fixed.

Doc/library/socket.rst
LivInTheLookingGlass: Does the new version make this better in your eyes?

LivInTheLookingGlass: Fixed

Modules/socketmodule.c
LivInTheLookingGlass: When I tried to compile on my system it hit an error unless I added that condition. I'm not sure how common that is, but it happened to me on Ubuntu 18.04.02. If it happened to me, I assume there are other systems where it will happen. From reading the manpage it seems that there's varying glibc support for those macros, and they had a section on how they should be defined if they weren't present.

gappleto97:

asvetlov (approved):

asvetlov:
Thanks!

PR #14264, opened 2019-06-20
bpo-37348 : add _PyUnicode_FROM_ASCII

methane:
It is a convenient wrapper for _PyUnicode_FromASCII().
It can be used like PyUnicode_FromString, without a length of the string.


skrah:
I wonder if we need a new macro throughout the code base.  In the case of `_decimal`, the paths are not performance sensitive, so I'd appreciate if you take it out.

If this really saves a measurable amount of time, we should have an external `PyUnicode_FromASCIIString()`, so other people can use it, too.

methane:
> In the case of `_decimal`, the paths are not performance sensitive, so I'd appreciate if you take it out.

OK, I reverted it.

PR #14266, opened 2019-06-20
bpo-37344 - Allow plistlib to parse data with leading whitespaces

Kriyszig:
`xml` detection fails if the `header` contains leading whitespaces as mentioned in the [issue](https://bugs.python.org/issue37344#msg346090)
Applying an `lstrip()` on `header` while checking. 

https://bugs.python.org/issue37344

### DIFF
--- a/Lib/plistlib.py
+++ b/Lib/plistlib.py
@@ -996,7 +996,7 @@ def loads(value, *, fmt=None, use_builtin_types=True, dict_type=dict):
     """Read a .plist file from a bytes object.
     Return the unpacked root object (which usually is a dictionary).
     """
-    fp = BytesIO(value)
+    fp = BytesIO(value.lstrip())
     return load(
         fp, fmt=fmt, use_builtin_types=use_builtin_types, dict_type=dict_type)
 
--- a/Lib/test/test_plistlib.py
+++ b/Lib/test/test_plistlib.py
@@ -486,6 +486,11 @@ class TestPlistlib(unittest.TestCase):
         self.assertRaises(ValueError, plistlib.loads,
                           b"<plist><integer>not real</integer></plist>")
 
+    def test_leadingwhitspaces(self):
+        self.assertEqual(plistlib.loads(b"   <plist><integer>1</integer></plist>"), 1)
+        data = plistlib.dumps(1, fmt=plistlib.FMT_XML)
+        self.assertEqual(plistlib.loads(b"  " + data), 1)
+
     def test_xml_encodings(self):
         base = TESTDATA[plistlib.FMT_XML]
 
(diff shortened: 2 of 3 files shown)


the-knights-who-say-ni:

Kriyszig:
I have signed the CLA and it's reflected on my account at <bugs.python.org> but the Heroku app hasn't synced yet

Lib/plistlib.py
```diff
@@ -518,7 +518,7 @@ def _is_fmt_xml(header):
     prefixes = (b'<?xml', b'<plist')
 
     for pfx in prefixes:
-        if header.startswith(pfx):
+        if header.lstrip().startswith(pfx):
```
tirkarthi: Please add a test in `test_plistlib.py` that is something similar to https://github.com/python/cpython/blob/a1952122a3a20272e4db9e3c6205009a91c73ccf/Lib/test/test_plistlib.py#L481 but with valid data and with a space before " <plist" so that it should decode correctly. Also a similar test for xml tags is needed I guess.

Before PR :

```
>>> plistlib.loads(b" <plist><integer>1</integer></plist>")
  File "/Users/karthikeyansingaravelan/stuff/python/cpython/Lib/plistlib.py", line 1000, in loads
    return load(
  File "/Users/karthikeyansingaravelan/stuff/python/cpython/Lib/plistlib.py", line 986, in load
    raise InvalidFileException()
plistlib.InvalidFileException: Invalid file
```

After PR :

```
>>> plistlib.loads(b" <plist><integer>1</integer></plist>")
1
```

tirkarthi:
Please also add a [news entry](https://devguide.python.org/committing/?highlight=news#what-s-new-and-news-entries) regarding this change.

Lib/plistlib.py
Kriyszig: On it!

pachoo:
I'd be happy to try giving a PR a whirl.  The initially proposed solution (lstrip the header) has a few issues:

* The expat XML parser doesn't seem to like it when the XML data doesn't start at a tag (so the parser would fail).
* If the plist had > 32 bytes of whitespace the _is_fmt_xml() would also fail.

Kriyszig:
@pachoo You are correct. I ran into the same troubles while writing tests.
There are two ways to solve this from where I can see this:
* lstrip the data before `BytesIO`
* Modify parser to allow data with leading white space

The first one will solve both the issues. The latter is a bit complicated as once we have data in `BytesIO` dropping the leading characters are process intensive as the data after a drop has to be copied back

Kriyszig:
7cfebbb implements the former solution - `lstrip` the argument  before checking for validity

Lib/plistlib.py
Kriyszig: Added tests in 7cfebbb

PR #14268, opened 2019-06-20
bpo-37347: fix refcount problem in sqlite3

gescheit:
fix refcount problem in sqlite3

https://bugs.python.org/issue37347

### DIFF
--- a/Lib/sqlite3/test/regression.py
+++ b/Lib/sqlite3/test/regression.py
@@ -25,7 +25,6 @@ import datetime
 import unittest
 import sqlite3 as sqlite
 import weakref
-import sys
 import functools
 from test import support
 
@@ -395,11 +394,7 @@ class RegressionTests(unittest.TestCase):
                        self.con.set_authorizer]:
             printer_instance = Printer()
             method(printer_instance.log)
-            log = printer_instance.log
-            refcnt_before = sys.getrefcount(log)
-            method(log)
-            refcnt_after = sys.getrefcount(log)
-            self.assertEqual(refcnt_after - refcnt_before, 1, "%r must increase reference-count of arg" % method)
+            method(printer_instance.log)
             self.con.execute("select 1")  # trigger seg fault
             method(None)
 
--- a/Lib/sqlite3/test/regression.py
+++ b/Lib/sqlite3/test/regression.py
@@ -25,6 +25,8 @@ import datetime
 import unittest
 import sqlite3 as sqlite
 import weakref
+import sys
+import functools
 from test import support
 
 class RegressionTests(unittest.TestCase):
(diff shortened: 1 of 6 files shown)


Modules/_sqlite/connection.c
```diff
@@ -1003,13 +1010,15 @@ static PyObject* pysqlite_connection_set_authorizer(pysqlite_Connection* self, P
...
     }
 
-    if (PyDict_SetItem(self->function_pinboard, authorizer_cb, Py_None) == -1) {
-        return NULL;
-    }
     rc = sqlite3_set_authorizer(self->db, _authorizer_callback, (void*)authorizer_cb);
     if (rc != SQLITE_OK) {
         PyErr_SetString(pysqlite_OperationalError, "Error setting authorizer callback");
+        Py_INCREF(Py_None);
+        Py_XSETREF(self->function_pinboard_authorizer_cb, Py_None);
```
benjaminp: Why are we doing this?

Lib/sqlite3/test/regression.py
```diff
@@ -383,72 +385,30 @@ def CheckDelIsolation_levelSegfault(self):
...
-    __hash__ = None
+        for method in [self.con.set_trace_callback,
+                       functools.partial(self.con.set_progress_handler, n=1),
+                       self.con.set_authorizer]:
+            printer_instance = Printer()
+            method(printer_instance.log)
+            log = printer_instance.log
+            refcnt_before = sys.getrefcount(log)
+            method(log)
+            refcnt_after = sys.getrefcount(log)
```
benjaminp: We should try to avoid writing tests that use `sys.getrefcount`. Does this test fail if you pass `-R ::` to `regrtest.py`?

Modules/_sqlite/connection.c
```diff
@@ -1033,12 +1042,13 @@ static PyObject* pysqlite_connection_set_progress_handler(pysqlite_Connection* s
     if (progress_handler == Py_None) {
         /* None clears the progress handler previously set */
         sqlite3_progress_handler(self->db, 0, 0, (void*)0);
+        Py_INCREF(Py_None);
+        Py_XSETREF(self->function_pinboard_progress_handler, Py_None);
```
benjaminp: Why not just use `NULL`?

Modules/_sqlite/connection.h
```diff
@@ -89,7 +89,9 @@ typedef struct
      * create_function/create/aggregate, use these as dictionary keys, so we
      * can keep the total system refcount constant by clearing that dictionary
      * in connection_dealloc */
```
benjaminp: Comments needs an update.

Modules/_sqlite/connection.c
gescheit: I am new this business, so I mimic current behavior like in cursor.c:
```c
Py_INCREF(Py_None);
Py_XSETREF(self->lastrowid, Py_None);
...
Py_DECREF(self->lastrowid);
....
self->lastrowid = _pysqlite_long_from_int64(lastrowid);
```
I will change it to NULL.

Modules/_sqlite/connection.h
gescheit: Fixed.

Modules/_sqlite/connection.c
gescheit: Fixed.

Modules/_sqlite/connection.c
gescheit: Fixed.

Lib/sqlite3/test/regression.py
gescheit: Nope. But it can be rewritten to fall to segmentation fault.

benjaminp: Sounds good to me. Not crashing is the behavior we want to test anyway.

gescheit: Then it is done in d1b57ff.

PR #14273, opened 2019-06-20
bpo-37348: optimize PyUnicode_FromString

methane:

### DIFF
--- a/Objects/unicodeobject.c
+++ b/Objects/unicodeobject.c
@@ -2163,12 +2163,27 @@ PyUnicode_FromStringAndSize(const char *u, Py_ssize_t size)
 PyObject *
 PyUnicode_FromString(const char *u)
 {
-    size_t size = strlen(u);
-    if (size > PY_SSIZE_T_MAX) {
-        PyErr_SetString(PyExc_OverflowError, "input too long");
-        return NULL;
+    /* Most calls of PyUnicode_FromString is for short ASCII string.
+     * So we optimize this function for them.
+     */
+    size_t len = 0;
+    int is_ascii = 1;
+
+    while (u[len] != '\0') {
+        if (u[len] > 127) {
+            is_ascii = 0;
+        }
+        len++;
+        if (len > PY_SSIZE_T_MAX) {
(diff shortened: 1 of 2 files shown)


Objects/unicodeobject.c
```diff
@@ -2163,12 +2163,27 @@ PyUnicode_FromStringAndSize(const char *u, Py_ssize_t size)
...
-        return NULL;
+    /* Most calls of PyUnicode_FromString is for short ASCII string.
+     * So we optimize this function for them.
+     */
+    size_t len = 0;
+    int is_ascii = 1;
+
+    while (u[len] != '\0') {
+        if (u[len] > 127) {
+            is_ascii = 0;
```
vstinner, line 2174: Can't you reuse fast _Py_bytes_isascii() somehow here? It works on unsigned long words (64 bits) rather than working on bytes (8 bits): it should be 8x faster.

Objects/unicodeobject.c
```diff
@@ -2163,12 +2163,27 @@ PyUnicode_FromStringAndSize(const char *u, Py_ssize_t size)
...
+            PyErr_SetString(PyExc_OverflowError, "input too long");
+            return NULL;
+        }
     }
-    return PyUnicode_DecodeUTF8Stateful(u, (Py_ssize_t)size, NULL, NULL);
+
+    if (is_ascii) {
+        return _PyUnicode_FromASCII(u, (Py_ssize_t)len);
+    }
+    return PyUnicode_DecodeUTF8Stateful(u, (Py_ssize_t)len, NULL, NULL);
```
vstinner, line 2186: Why not modifying PyUnicode_DecodeUTF8Stateful() to detect if the input string is ASCII, or your heuristic is faster? Is it because it make non-ASCII string decoding slower?

vstinner (commented):
I don't understand why this change is needed (why it makes the code faster): unicode_decode_utf8() already starts by decoding from ASCII. Is it a performance issue in ascii_decode()? Does _PyUnicodeWriter adds a small overhead which is significant here?

cc @serhiy-storchaka who knows well such hardcores micro-optimizations ;-)

PR #14282, opened 2019-06-21
[3.8] bpo-32924: Fix the Show Source url in 3.8 documentation.

Mariatta:
The "Show Source" link in Python 3.8 docs is pointing
to the master branch. It should point to the 3.8 branch.

https://bugs.python.org/issue32924

### DIFF
--- a/Doc/tools/templates/customsourcelink.html
+++ b/Doc/tools/templates/customsourcelink.html
@@ -4,7 +4,7 @@
     <ul class="this-page-menu">
       <li><a href="{{ pathto('bugs') }}">{% trans %}Report a Bug{% endtrans %}</a></li>
       <li>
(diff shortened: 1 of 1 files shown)


tirkarthi (approved):
LGTM. Thanks. Given that it was reported for 3.7 release too in https://bugs.python.org/issue32924 perhaps this could be added as a step in release process for major version like 3.9? cc: @JulienPalard

brettcannon:
@Mariatta it looks like the review from tirkarthi broke automerge because the `awaiting` label changed off of `awaiting merge`?

Mariatta:
Awaiting core review is correct state here. All core devs PRs now require other core dev's review since https://github.com/python/bedevere/pull/158.

ned-deily (approved):

ned-deily:
Thanks, @Mariatta, I'll make sure that step is included in PEP 101.

PR #14283, opened 2019-06-21
bpo-37348: optimize decoding ASCII string

methane:
Use `_PyUnicode_Writer` only after `ascii_decode` is failed.

### DIFF
--- a/Objects/unicodeobject.c
+++ b/Objects/unicodeobject.c
@@ -265,7 +265,7 @@ unicode_fill(enum PyUnicode_Kind kind, void *data, Py_UCS4 value,
 /* Forward declaration */
 static inline int
 _PyUnicodeWriter_WriteCharInline(_PyUnicodeWriter *writer, Py_UCS4 ch);
-static inline int
+static inline void
 _PyUnicodeWriter_InitWithBuffer(_PyUnicodeWriter *writer, PyObject *buffer);
 static PyObject *
 unicode_encode_utf8(PyObject *unicode, _Py_error_handler error_handler,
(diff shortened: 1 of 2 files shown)


Objects/unicodeobject.c
```diff
@@ -4900,13 +4892,32 @@ unicode_decode_utf8(const char *s, Py_ssize_t size,
...
+
+    // Use _PyUnicodeWriter after fast path is failed.
+    _PyUnicodeWriter writer;
     _PyUnicodeWriter_Init(&writer);
     writer.min_length = size;
-    if (_PyUnicodeWriter_Prepare(&writer, writer.min_length, 127) == -1)
-        goto onError;
+    writer.buffer = u;
+    writer.pos = s - starts;
+    _PyUnicodeWriter_Update(&writer);
```
vstinner: IMHO this pattern would deserve a new _PyUnicodeWriter_InitFromBuffer(writer, buffer) constructor.

methane:
```
$ ./python -m pyperf timeit --compare-to=./python-master -s 'b=b"foo"' -- 'b.decode()'
python-master: ..................... 83.3 ns +- 1.2 ns
python: ..................... 72.4 ns +- 1.1 ns

Mean +- std dev: [python-master] 83.3 ns +- 1.2 ns -> [python] 72.4 ns +- 1.1 ns: 1.15x faster (-13%)
```

`PyUnicode_FromString("aaaaaaaaaaaaaaaaaaaaaaaaaaaaaa");`: about 38ns -> 28ns

PR #14284, opened 2019-06-21
bpo-37358: Use vectorcall for functools.partial

jdemeyer:
https://bugs.python.org/issue37358

### DIFF
--- a/Misc/NEWS.d/next/Library/2019-06-21-14-54-02.bpo-37358.RsASpn.rst
+++ b/Misc/NEWS.d/next/Library/2019-06-21-14-54-02.bpo-37358.RsASpn.rst
@@ -1,0 +1,1 @@
+Optimized ``functools.partial`` by using vectorcall.
--- a/Modules/_functoolsmodule.c
+++ b/Modules/_functoolsmodule.c
@@ -141,21 +141,6 @@ partial_vectorcall_fallback(partialobject *pto, PyObject *const *args,
 }
 
 static PyObject *
-partial_vectorcall_no_args(partialobject *pto, PyObject *const *args,
-                           size_t nargsf, PyObject *kwnames)
-{
-    /* pto->kw is mutable, so need to check every time */
-    if (PyDict_GET_SIZE(pto->kw)) {
-        return partial_vectorcall_fallback(pto, args, nargsf, kwnames);
-    }
-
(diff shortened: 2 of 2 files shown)


encukou:
Is `partial_vectorcall_no_args` useful? It kicks in only if no arguments (positional nor keyword) are added at all. Are such no-op partial objects worth optimizing?

jdemeyer:
> Are such no-op partial objects worth optimizing?

To be honest, I have no idea. I just notices that this could be optimized, so I did it. The behaviour of `partial(func)` would be just like `staticmethod(func)` but callable itself.

Do you want me to remove that special case?

encukou:
> To be honest, I have no idea. I just notices that this could be optimized, so I did it.

That sounds like premature optimization.
I wonder how much this PR itself helps; did you run any benchmarks?

> The behaviour of `partial(func)` would be just like `staticmethod(func)` but callable itself.

If that is an actual use case for that, just make `staticmethod` callable. AFAIK, in the discussions around that [⁽⁰⁾](https://mail.python.org/pipermail/python-ideas/2018-June/051585.html) [⁽¹⁾](https://bugs.python.org/issue20309), the only arguments against were not enough actual use cases.
We shouldn't optimize for people repurposing `partial` for this.

> Do you want me to remove that special case?

Yes, unless there is a need for the complication.

jdemeyer:
I tried 3 simple micro-benchmarks (without PGO):

1. `f = partial(len)` and then benchmarking `f(())`
**before**: Mean +- std dev: 70.3 ns +- 0.2 ns
**after**: Mean +- std dev: 47.3 ns +- 0.1 ns
2. `f = partial(len,  [])` and then benchmarking `f()`
**before**: Mean +- std dev: 50.2 ns +- 0.2 ns
**after**: Mean +- std dev: 36.1 ns +- 0.3 ns
3. `f = partial(list.count, [])` and then benchmarking `f(0)`
**before**: Mean +- std dev: 84.4 ns +- 1.3 ns
**after**: Mean +- std dev: 47.9 ns +- 2.2 ns

So there is a clear performance improvement in all cases.

jdemeyer:
Failed Travis CI build is due to `OSError: [Errno 113] No route to host` clearly not caused by this PR.

encukou (approved):

encukou:
Looks good! Thank you!

PR #11195, opened 2018-12-17
bpo-37335: Add alias for ASCII encoding

kulikjak:
Solaris uses '646' as an alias for ASCII encoding and thus all the locale coercion tests are failing. This simple patch fixes the problem.

https://bugs.python.org/issue37335

PR #14285, opened 2019-06-21
[3.7] bpo-37335: Add alias for ASCII encoding

kulikjak:
Solaris uses '646' as an alias for ASCII encoding and thus all the locale coercion tests are failing. This simple patch fixes the problem.

This is a second PR into a correct branch (this is 3.7 only change). Original one is here #11195 


ncoghlan (approved):

kulikjak:
The issue was fixed with a different PR.

PR #14286, opened 2019-06-21
bpo-36888: Increase timeout in test_parent_process

pierreglaser:
Attempt to fix the buildbot failures reported in the original bpo issue.

We (@tomMoral and I) suspect that the `AssertionError` comes from the child process and the grand-child processes taking more than 5 seconds to get up and running, making the following calls to `rconn.poll(timeout=5)` return `False`. 

The proposed fix here is therefore to give more time for the child processes to start by increasing the timeout in the `rconn.poll` calls.

We also increase the life-span of the child process to make sure the it is not dead when the grand-child process starts executing.

cc @vstinner :)

https://bugs.python.org/issue36888

### DIFF
--- a/Lib/test/_test_multiprocessing.py
+++ b/Lib/test/_test_multiprocessing.py
@@ -301,7 +301,7 @@ class _TestProcess(BaseTestCase):
             target=self._test_create_grandchild_process, args=(wconn, ))
         p.start()
 
-        if not rconn.poll(timeout=5):
+        if not rconn.poll(timeout=60):
             raise AssertionError("Could not communicate with child process")
         parent_process_status = rconn.recv()
         self.assertEqual(parent_process_status, "alive")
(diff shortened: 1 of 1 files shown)


tomMoral:
I agree this looks like an issue due to timeout that are too short. This PR should fix this.

vstinner (changes requested):
test_parent_process() seems to be designed around time. That's a very weird synchronization primitation, it's not reliable at all. Can't you use another more reliable synchronization primitive? Something like https://docs.python.org/dev/library/multiprocessing.html#multiprocessing.Event?

cc @pablogsal

tomMoral:
The synchronization here does not rely on time bu is done thru the `multiprocessing.Pipe` object which is used as a communication channel. The timeout are only here to safeguard against hanged interpreters. Using `mp.Event` would lead to very similar code, as to avoid frozen interpreters, we would need to rely on `Event.wait` with non-zero timeout, which would be equivalent to the `poll` with non-zero timeout.

vstinner (approved):

PR #14287, opened 2019-06-21
bpo-37362: test_gdb now ignores stderr

vstinner:
test_gdb no longer fails if it gets an "unexpected" message on
stderr: it now ignores stderr. The purpose of test_gdb is to test
that python-gdb.py commands work as expected, not to test gdb.

https://bugs.python.org/issue37362

### DIFF
--- a/Misc/NEWS.d/next/Tests/2019-06-21-15-47-33.bpo-37362.D3xppx.rst
+++ b/Misc/NEWS.d/next/Tests/2019-06-21-15-47-33.bpo-37362.D3xppx.rst
@@ -1,0 +1,3 @@
+test_gdb no longer fails if it gets an "unexpected" message on stderr: it now
+ignores stderr. The purpose of test_gdb is to test that python-gdb.py commands
+work as expected, not to test gdb.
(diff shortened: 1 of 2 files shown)


Misc/NEWS.d/next/Tests/2019-06-21-15-47-33.bpo-37362.D3xppx.rst
```diff
@@ -0,0 +1,3 @@
+test_gdb no longer fails if it gets an "unexpected" message on stderr: it now
+ignores stderr. The purpose of test_gdb is to test that python-gdd.py commands
```
davidmalcolm: > python-gdd.py
"python-gdb.py", presumably

vstinner: oops, right, it's a typo. fixed.

vstinner:
Oh, I didn't expect that I would have the honor of getting a review of the author of the original code author :-) Dave Malcolm! He wrote "(...) your approach looks good to me": https://bugs.python.org/issue37362#msg346221

PR #14288, opened 2019-06-21
bpo-37244: Increase robustness of test_resource_tracker

pierreglaser:
Attempt to fix the buildbot failures reported in the original bpo issue.
From the failure tracebacks, I suspect it is only because the cleanup time given to the `resource_tracker` was not high enough (2 seconds). I increased it to to 60 seconds.

cc @vstinner :) 

https://bugs.python.org/issue37244

### DIFF
--- a/Lib/test/_test_multiprocessing.py
+++ b/Lib/test/_test_multiprocessing.py
@@ -5013,7 +5013,7 @@ class TestResourceTracker(unittest.TestCase):
 
                 deadline = time.monotonic() + 60
                 while time.monotonic() < deadline:
-                    time.sleep(2.0)
+                    time.sleep(.5)
                     try:
                         _resource_unlink(name2, rtype)
                     except OSError as e:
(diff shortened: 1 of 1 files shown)


Lib/test/_test_multiprocessing.py
```diff
@@ -5010,12 +5010,21 @@ def create_and_register_resource(rtype):
...
-                time.sleep(2.0)
-                with self.assertRaises(OSError) as ctx:
-                    _resource_unlink(name2, rtype)
-                # docs say it should be ENOENT, but OSX seems to give EINVAL
-                self.assertIn(
-                    ctx.exception.errno, (errno.ENOENT, errno.EINVAL))
+
+                deadline = time.monotonic() + 60
+                while time.monotonic() < deadline:
+                    time.sleep(2.0)
```
vstinner: Would it be possible to use a shorter sleep to make the test faster? Like sleep(.5).

PR #14291, opened 2019-06-21
bpo-37348: optimize decoding ASCII string

methane:
gcc and clang can not optimize this code well:

    _PyUnicodeWriter_Init(&writer);
    writer.min_length = size;
    _PyUnicodeWriter_Prepare(&writer, size, maxchar);

Add _PyUnicodeWriter_InitPrepare() which does same thing, but
more compiler friendly.


methane:
#14283 is faster.  _PyUnicode_Writer is a relatively large struct.  Skip using it for a simple case is faster than initializing _PyUnicode_Writer.

PR #14292, opened 2019-06-21
gh-81483: Add an "onerror" callback parameter to the tempfile.TemporaryDirectory member functions

websurfer5:
This PR adds an "onerror" callback parameter to the tempfile.TemporaryDirectory() member functions so that the caller can perform special handling for directory items that it can't automatically delete.  The caller created the undeletable directory entries, so it is reasonable to believe the caller may know how to unmake what they made.

The provided test cases create a filesystem image and mount it within the temporary directory to force an error in the underlying shutil.rmtree() call during cleanup that triggers calls to the new onerror parameters.  The test cases are provided for macOS because Windows and Linux require special user privileges to create and/or mount/unmount a filesystem image.  This method was chosen to simulate the cleanup problems described in [bpo-36422](https://bugs.python.org/issue36422).

I am open to suggestions on how to force errors on those systems or in a portable manner.  The usual trick of changing directory entry permissions won't work because tempfile.TemporaryDirectory() has heuristics to handle that case.

https://bugs.python.org/issue37302

* Issue: gh-81483

### DIFF
--- a/Doc/library/tempfile.rst
+++ b/Doc/library/tempfile.rst
@@ -114,7 +114,7 @@ The module defines the following user-callable items:
       Added *errors* parameter.
 
 
-.. function:: TemporaryDirectory(suffix=None, prefix=None, dir=None)
+.. function:: TemporaryDirectory(suffix=None, prefix=None, dir=None, onerror=None)
 
    This function securely creates a temporary directory using the same rules as :func:`mkdtemp`.
    The resulting object can be used as a context manager (see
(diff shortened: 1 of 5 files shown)


mangrisano (approved):

iritkatriel (changes requested):
See recent changes to rmtree's onerror arg: https://github.com/python/cpython/pull/102829

erlend-aasland:
@websurfer5 could you please resolve the conflicts and update the PR?

erlend-aasland:
The feature request was rejected by @giampaolo in https://github.com/python/cpython/issues/80603#issuecomment-1093819196. Closing.

PR #14299, opened 2019-06-21
bpo-37364: Use io.open_code() to read .pth files

zooba:
https://bugs.python.org/issue37364


tiran (approved):

tiran:
As you request, Miss Islington. I'm not going to argue with a bot. :)

PR #14310, opened 2019-06-22
bpo-37212: Preserve keyword argument order in unittest.mock.call and error messages

tirkarthi:
Since Python 3.6 the order of keyword arguments is preserved. Change `mock.call` repr and error messages to print the original order instead of sorting it.

https://bugs.python.org/issue37212

### DIFF
--- a/Lib/unittest/mock.py
+++ b/Lib/unittest/mock.py
@@ -2281,7 +2281,7 @@ def _format_call_signature(name, args, kwargs):
     formatted_args = ''
     args_string = ', '.join([repr(arg) for arg in args])
     kwargs_string = ', '.join([
-        '%s=%r' % (key, value) for key, value in sorted(kwargs.items())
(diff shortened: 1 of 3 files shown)


dhilst (approved):
Looks good to me!

mariocj89 (approved):

ammaraskar (approved):
LGTM, preserving the call order is far more intuitive.

And for reference this is just a visual change useful for users introspecting test failures and such,
```python
>>> m = Mock()
>>> m(b=1, a=2)
>>> m.assert_called_with(a=2, b=1)
```
works and will continue to work.

zware:
Thanks for the patch, @tirkarthi, and for the reviews @dhilst and @ammaraskar!  Due to the (admittedly very slim) chance that this could break things for someone who shouldn't be relying on repr orders anyway, I won't backport this further than 3.8.

PR #14312, opened 2019-06-22
bpo-37374: Do not escape quotes in minidom inside text segments

mitar:
Addressing the issue: https://bugs.python.org/issue37374

https://bugs.python.org/issue37374

### DIFF
--- a/Lib/test/test_minidom.py
+++ b/Lib/test/test_minidom.py
@@ -5,6 +5,7 @@ import pickle
 import io
 from test import support
 import unittest
+from xml.etree import ElementTree
 
 import xml.dom.minidom
 
@@ -1662,6 +1663,15 @@ class MinidomTest(unittest.TestCase):
         self.checkWholeText(dom1.getElementsByTagName('node')[0].firstChild, '</data>')
         dom2 = parseString(dom1.toprettyxml())
         self.checkWholeText(dom2.getElementsByTagName('node')[0].firstChild, '</data>')
+
+    def testQuoteEscape(self):
+        text = ElementTree.Element('text')
+        text.text = 'f&oo"b<a>r'
+        xml_string = ElementTree.tostring(text)
+        xml_tree = parseString(xml_string)
+        output = xml_tree.toprettyxml(indent='  ')
+        self.assertEqual(output.splitlines()[1], xml_string.decode('utf8'))
+
 
 if __name__ == "__main__":
     unittest.main()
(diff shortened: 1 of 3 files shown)


mangrisano:
Hi and thank you for the pull request. 
Just one question: did you test the change?

mitar:
Yes, I tested by running:

```python
from xml.etree import ElementTree
text = ElementTree.Element('text')
text.text = 'foo"bar'
xml_string = ElementTree.tostring(text)
import xml.dom.minidom as minidom
xml_tree = minidom.parseString(xml_string)
xml_tree.toprettyxml(indent='  ')
```

which now returns:

```
'<?xml version="1.0" ?>\n<text>foo"bar</text>\n'
```

I also ran whole Python test suite.

mangrisano:
Perhaps you should provide it in the test_xml module.
I'm not a core-dev so feel free to not follow my suggestion. :)

mitar:
You are right. I should add the test.

This PR in fact makes minidom behave exactly the same as `ElementTree.tostring`. `ElementTree.tostring` is also not escaping quotes.

Lib/test/test_minidom.py
```diff
@@ -1663,5 +1664,14 @@ def test_cdata_parsing(self):
         dom2 = parseString(dom1.toprettyxml())
         self.checkWholeText(dom2.getElementsByTagName('node')[0].firstChild, '</data>')
 
+    def testQuoteEscape(self):
+        text = ElementTree.Element('text')
+        text.text = 'f&oo"b<a>r'
+        xml_string = ElementTree.tostring(text)
```
scoder, line 1670: We should not rely on an unrelated library providing the same text string serialisation. Instead, the test should use a plain string to compare against. Otherwise, changes in ElementTree could make this test fail without need.

Lib/xml/dom/minidom.py
```diff
@@ -1110,7 +1117,7 @@ def splitText(self, offset):
         return newText
 
     def writexml(self, writer, indent="", addindent="", newl=""):
-        _write_data(writer, "%s%s%s" % (indent, self.data, newl))
+        _write_text_data(writer, "%s%s%s" % (indent, self.data, newl))
```
scoder, line 1120: I haven't checked deeply, but my gut feeling would prefer a special case for attribute values, not for text content. Can't say if that's similarly easy to achieve.

scoder (commented):
Thank you for the PR. However, I've rejected the ticket, so I have to reject the PR, too.

I'll leave my review comments anyway, in case it gets reconsidered at some later point.

PR #14139, opened 2019-06-17
bpo-35224: Reverse evaluation order of key: value in dict comprehensions

joernheissler:
… as proposed in PEP 572; key is now evaluated before value.

https://bugs.python.org/issue35224

PR #14313, opened 2019-06-22
bpo-35224: Bump the pyc magic number after the change in MAP_ADD

pablogsal:
CC: @joernheissler 


gvanrossum:
This will need a backport to 3.8 once https://github.com/python/cpython/pull/14314 has been merged (that's the backport of https://github.com/python/cpython/pull/14139).

pablogsal:
Then I will change the version of the number to Python 3.8b1 instead of Python3.9a0

PR #14316, opened 2019-06-23
bpo-26967: fix flag grouping with allow_abbrev=False

Zac-HD:
The `allow_abbrev` option for ArgumentParser is documented and intended to disable support for unique prefixes of --options, which may sometimes be ambiguous due to deferred parsing.

However, the initial implementation also broke parsing of grouped short flags, such as `-ab` meaning `-a -b` (or `-a=b`).  Checking the argument for a leading `--` before rejecting it fixes this.

This was prompted by pytest-dev/pytest#5469, so a backport to at least 3.8 would be great :smile:  
And this is my first PR to CPython, so please let me know if I've missed anything!

https://bugs.python.org/issue26967

### DIFF
--- a/Doc/library/argparse.rst
+++ b/Doc/library/argparse.rst
@@ -182,6 +182,10 @@ ArgumentParser objects
    .. versionchanged:: 3.5
       *allow_abbrev* parameter was added.
 
+   .. versionchanged:: 3.8
+      In previous versions, *allow_abbrev* also disabled grouping of short
+      flags such as ``-vv`` to mean ``-v -v``.
+
 The following sections describe how each of these are used.
 
 
--- a/Lib/argparse.py
+++ b/Lib/argparse.py
@@ -2130,7 +2130,7 @@ class ArgumentParser(_AttributeHolder, _ActionsContainer):
                 action = self._option_string_actions[option_string]
                 return action, option_string, explicit_arg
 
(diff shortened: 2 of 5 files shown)


mangrisano (approved):
LGTM. Thanks a lot for providing the tests. They are very useful. :)

blueyed (approved):

Meinersbur:
Thank you for taking care of the issue.

Doc/library/argparse.rst
```diff
@@ -182,6 +182,10 @@ ArgumentParser objects
    .. versionchanged:: 3.5
       *allow_abbrev* parameter was added.
 
+   .. versionchanged:: 3.8
```
encukou, line 185:
```suggestion
   .. versionchanged:: 3.9
```

Zac-HD: I'm still hoping that this will be merged into the 3.8 branch before 3.8.0, but I guess we can adjust the number later.

encukou:
I talked to a more experienced core dev, and we agreed that this a bugfix appropriate for 3.8.
Sorry for the noise!

Zac-HD:
No worries!  I'm just glad I can make a contribution, and getting it into 3.8 is even better :smile:

encukou (approved):

PR #14320, opened 2019-06-23
bpo-35224: Bump the pyc magic number by 1 instead of by 10 in last modification

pablogsal:


gvanrossum (approved):
If Serhiy says so. :-)

PR #14329, opened 2019-06-24
bpo-36661: Improved code snippets in dataclasses.rst

Kriyszig:
Issue: [36661](https://bugs.python.org/issue36661)

The imports are consistent with those in the [tests for dataclass](https://github.com/python/cpython/blob/master/Lib/test/dataclass_module_1.py)

https://bugs.python.org/issue36661

### DIFF
--- a/Doc/library/dataclasses.rst
+++ b/Doc/library/dataclasses.rst
@@ -19,6 +19,8 @@ in :pep:`557`.
 The member variables to use in these generated methods are defined
 using :pep:`526` type annotations.  For example this code::
 
+  from dataclasses import dataclass
+
   @dataclass
   class InventoryItem:
       '''Class for keeping track of an item in inventory.'''


ericvsmith (commented):
I'm not sure this is an improvement. I haven't done an exhaustive survey, but most usages I've seen are `from dataclasses import dataclass`, not `import dataclasses`. I honestly don't know what the best style is, or even if what we use here should influence the recommended style.

In any event, all other module-level names (like `field`) need the same treatment, if we're going to make this change.

Kriyszig:
Do you think using `from dataclasses import dataclass` would be a better alternative?
[I used the patter used while writing tests and hence used `@dataclasses.dataclass`]

Kriyszig:
cc/ @ericvsmith
 Added `from dataclasses import dataclass` to the first example. Should help in clearing things up for the confused users.
In case you think the old doc is just fine, I'll go ahead and close this PR  (^_^)
Thank you for your time

PR #14330, opened 2019-06-24
bpo-20177: use builtin_dir to test METH_VARARGS

jdemeyer:
CC @ZackerySpytz

https://bugs.python.org/issue20177


vstinner:
Maybe add a function to testcapi to not have to update tests each time the tested function uses a different calling convention.

pganssle (changes requested):
I agree with @vstinner, these tests are fragile because it seems they're testing incidental implementation details of these builtins that are not guaranteed.

Unless these properties *are* guaranteed for backwards-compatibility, I think replacing these with dedicated test functions with a guaranteed calling method is the right way to test it.

PR #14331, opened 2019-06-24
bpo-37382: check for debug info before testing GDB

jdemeyer:
https://bugs.python.org/issue37382

### DIFF
--- a/Doc/library/test.rst
+++ b/Doc/library/test.rst
@@ -414,6 +414,11 @@ The :mod:`test.support` module defines the following functions:
    Return ``True`` if Python was not built with ``-O0`` or ``-Og``.
 
 
+.. function:: python_has_debug_info()
+
+   Return ``True`` if Python was built with ``-g``.
+
+
 .. function:: with_pymalloc()
 
    Return :data:`_testcapi.WITH_PYMALLOC`.
--- a/Lib/test/support/__init__.py
+++ b/Lib/test/support/__init__.py
@@ -1643,6 +1643,16 @@ def disable_gc():
             gc.enable()
 
 
+def python_has_debug_info():
+    """Find if Python was built with debug information."""
+    cflags = sysconfig.get_config_var('PY_CFLAGS') or ''
+    final_opt = ""
+    for opt in cflags.split():
+        if opt.startswith('-g'):
+            final_opt = opt
+    return final_opt == '-g'
+
+
 def python_is_optimized():
     """Find if Python was built with optimizations."""
     cflags = sysconfig.get_config_var('PY_CFLAGS') or ''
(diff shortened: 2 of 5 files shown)


jdemeyer:
CC @encukou

jdemeyer:
Bugfix, so should be backported to all earlier versions.

vstinner (changes requested):
The tests should be skipped if Python is optimized: see the issue.

jdemeyer:
So how would you suggest to fix bpo-37382?

vstinner:
https://bugs.python.org/issue37382 is not a bug.

jdemeyer:
At least part of it is a bug: `test_gdb` fails when compiling CPython with `-O0 -g0`.

But the real issue for me is that `test_gdb` is not run by default. This PR addresses that.

Lib/test/support/__init__.py
```diff
@@ -1643,6 +1643,16 @@ def disable_gc():
             gc.enable()
 
 
+def python_has_debug_info():
+    """Find if Python was built with debug information."""
+    cflags = sysconfig.get_config_var('PY_CFLAGS') or ''
+    final_opt = ""
+    for opt in cflags.split():
+        if opt.startswith('-g'):
+            final_opt = opt
```
vstinner, line 1652: Why not just returning True here? (and "return False" outside the loop)

Lib/test/support/__init__.py
```diff
@@ -1643,6 +1643,16 @@ def disable_gc():
...
+    return final_opt == '-g'
```
vstinner, line 1653: Why do you reject "-ggdb"? I heard that it's even better than "-g".

Doc/library/test.rst
```diff
@@ -414,6 +414,11 @@ The :mod:`test.support` module defines the following functions:
    Return ``True`` if Python was not built with ``-O0`` or ``-Og``.
 
 
+.. function:: python_has_debug_info()
+
+   Return ``True`` if Python was built with ``-g``.
```
vstinner, line 419: Add ".. versionadded:: 3.9".

vstinner (commented):
Please don't backport this change to Python 3.8 :-) test_gdb is giving me too many headaches :-) For example, it fails if Python is built with -Os.

encukou:
Sorry for missing the ping back in 2019.
This PR has been superseded by [#110026](https://github.com/python/cpython/pull/110026/files#diff-65a7e7fedffb3771932da22df1aa9c74f8ec3781ed33775db45ac011be17e8a0) (and others). 
The current settings probably skip more than they need to, so the issue might still be relevant -- but a PR for that should now skip fewer tests rather than more.

PR #14340, opened 2019-06-24
[2.7] bpo-37124: Fix reference leak in test_msilib (GH-13750)

vstinner:
(cherry picked from commit c0295dba259accc4b247beb22a0b2cc2f31d9850)

https://bugs.python.org/issue37124

### DIFF
--- a/Lib/test/test_msilib.py
+++ b/Lib/test/test_msilib.py
@@ -45,6 +45,7 @@ class MsiDatabaseTestCase(unittest.TestCase):
 
     def test_directory_start_component_keyfile(self):
         db, db_path = init_database()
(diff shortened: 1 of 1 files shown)


vstinner:
Compared to the change in master, I also added:

```
        self.addCleanup(unlink, db_path)
        self.addCleanup(db.Close)
```

vstinner:
`self.addCleanup(db.Close)` doesn't work: `AttributeError: '_msi.Database' object has no attribute 'Close'`.

I removed:

```
        self.addCleanup(unlink, db_path)
        self.addCleanup(db.Close)
```

PR #14351, opened 2019-06-24
bpo-37394: Fix pure Python implementation of queue

pablogsal:
https://bugs.python.org/issue37394

### DIFF
--- a/Lib/test/test_queue.py
+++ b/Lib/test/test_queue.py
@@ -1,7 +1,6 @@
 # Some simple queue module tests, plus some failure conditions
 # to ensure the Queue locks remain stable.
 import itertools
-import queue
 import random
 import threading
 import time
@@ -9,11 +8,9 @@ import unittest
 import weakref
 from test import support
 
-
-try:
-    import _queue
-except ImportError:
-    _queue = None
+py_queue = support.import_fresh_module('queue', blocked=['_queue'])
+c_queue = support.import_fresh_module('queue', fresh=['_queue'])
+need_c_queue = unittest.skipUnless(c_queue, "No _queue module found")
 
 QUEUE_SIZE = 5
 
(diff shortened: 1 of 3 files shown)


vstinner:
test_queue didn't catch this issue, so test_queue should be fixed somehow.

IMHO test_queue should be rewritten using something like that:

```
py_functools = support.import_fresh_module('functools', blocked=['_functools'])
c_functools = support.import_fresh_module('functools', fresh=['_functools'])
```

Lib/test/test_queue.py
```diff
@@ -232,45 +228,88 @@ def test_shrinking_queue(self):
...
+
+    def setUp(self):
+        self.type2test = self.queue.Queue
+        super().setUp()
+
+class PyQueueTest(QueueTest, unittest.TestCase):
+    queue = py_queue
+
+
+@unittest.skipUnless(c_queue, "No _queue module found")
```
vstinner: Maybe you can use:

```
need_c_queue = unittest.skipUnless(c_queue, "No _queue module found")

@need_c_queue
class CQueueTest(QueueTest, unittest.TestCase):  ...
```

Lib/test/test_queue.py
```diff
@@ -232,45 +228,88 @@ def test_shrinking_queue(self):
...
-        return queue.Queue._get(self)
-
-class FailingQueueTest(BlockingTestMixin, unittest.TestCase):
+class FailingQueueException(Exception): pass
+
+class FailingQueueTest(BlockingTestMixin):
+
+    def setUp(self):
+
+        queue_module = self.queue
```
vstinner: nitpick: maybe you can use there: Queue = self.queue.Queue

pablogsal: Done in 
2175091

Lib/test/test_queue.py
pablogsal: Done in 2175091

vstinner (approved):
LGTM. I suggest to rephrase the commit title, mayby: "Fix pure Python implementation of queue".

I may be interesting to add NEWS entry, and maybe backport this change to 3.7 and 3.8. You have to open a bpo issue in this case (to get a bpo number).

PR #14355, opened 2019-06-24
bpo-37392: Remove sys.setcheckinterval()

vstinner:
Remove sys.getcheckinterval() and sys.setcheckinterval() functions.
They were deprecated since Python 3.2. Use sys.getswitchinterval()
and sys.setswitchinterval() instead.

Remove also check_interval field of the PyInterpreterState structure.

https://bugs.python.org/issue37392

### DIFF
--- a/Doc/library/sys.rst
+++ b/Doc/library/sys.rst
@@ -551,14 +551,6 @@ always available.
    .. versionadded:: 3.7
 
 
-.. function:: getcheckinterval()
-
-   Return the interpreter's "check interval"; see :func:`setcheckinterval`.
-
-   .. deprecated:: 3.2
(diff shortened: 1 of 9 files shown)


vstinner:
macOS failed in Azure Pipelines, it's a known and unrelated issue: https://bugs.python.org/issue37245

serhiy-storchaka (approved):

vstinner:
I prefer to remove deprecated functions early in the 3.9 dev cycle, so we have more time for feedback. If something goes wrong, we can discuss reverting this removal before 3.9.0 final release.

vstinner:
Thanks for the review @serhiy-storchaka ;-)

PR #14356, opened 2019-06-24
bpo-37386: Conditionally skip test_large_file_ops

potomak:
Skip the test if `write()` fails with `errno.ENOSPC`: "OSError: [Errno
28] No space left on device".

https://bugs.python.org/issue37386

### DIFF
--- a/Lib/test/test_io.py
+++ b/Lib/test/test_io.py
@@ -604,10 +604,15 @@ class IOTest(unittest.TestCase):
             support.requires(
                 'largefile',
                 'test requires %s bytes and a long time to run' % self.LARGE)
-        with self.open(support.TESTFN, "w+b", 0) as f:
-            self.large_file_ops(f)
-        with self.open(support.TESTFN, "w+b") as f:
-            self.large_file_ops(f)
+        try:
+            with self.open(support.TESTFN, "w+b", 0) as f:
+                self.large_file_ops(f)
+            with self.open(support.TESTFN, "w+b") as f:
+                self.large_file_ops(f)
+        except OSError as e:
+            if e.errno == errno.ENOSPC:
+                self.skipTest("No space left on device")
+            raise
 
     def test_with_open(self):
         for bufsize in (0, 100):


mangrisano (approved):
LGTM.

Lib/test/test_io.py
```diff
@@ -604,10 +604,15 @@ def test_large_file_ops(self):
...
-            self.large_file_ops(f)
+        try:
+            with self.open(support.TESTFN, "w+b", 0) as f:
+                self.large_file_ops(f)
+            with self.open(support.TESTFN, "w+b") as f:
+                self.large_file_ops(f)
+        except OSError as e:
+            if e.errno == errno.ENOSPC:
+                self.skipTest("No space left on device")
+            raise
```
dhilst, line 615: Do this raise stop the test? Or it just print in the stdout!?

potomak: `raise` will re-raise the current exception, effectively stopping the single test.

I added a line `raise RuntimeException` in the test body to test this behavior, this is the result:

```
ERROR: test_large_file_ops (test.test_io.CIOTest)
  File "/home/gcappellotto/src/cpython/Lib/test/test_io.py", line 608, in test_large_file_ops
    raise RuntimeError
RuntimeError

ERROR: test_large_file_ops (test.test_io.PyIOTest)
  File "/home/gcappellotto/src/cpython/Lib/test/test_io.py", line 608, in test_large_file_ops
    raise RuntimeError
RuntimeError

Ran 606 tests in 38.472s

FAILED (errors=2, skipped=8)
test test_io failed
test_io failed in 38 sec 661 ms

== Tests result: FAILURE ==

1 test failed:
    test_io

Total duration: 38 sec 714 ms
Tests result: FAILURE
```

vstinner (approved):
LGTM. But I would prefer to get a second review from another core dev.

vstinner:
@serhiy-storchaka: would you mind to review this change? I think you have some experience with largefile/bigmem tests.

vstinner:
https://bugs.python.org/issue37386 was closed as "not a bug".

PR #14359, opened 2019-06-25
Make StreamHandler.terminator more discoverable

andresdelfino:
I think the StreamHandler.terminator attribute should be marked as such, and linked in emit() to make it more discoverable.


tirkarthi:
cc: @vsajip

PR #14360, opened 2019-06-25
bpo-37380: subprocess: skip _cleanup and __del__ on win

efiop:
As noted by @eryksun in [1] and [2], using _cleanup and _active(in
__del__) is not necessary on Windows, since:

> Unlike Unix, a process in Windows doesn't have to be waited on by
> its parent to avoid a zombie. Keeping the handle open will actually
> create a zombie until the next _cleanup() call, which may be never
> if Popen() isn't called again.

This patch simply defines `subprocess._active` as `None`, for which we already
have the proper logic in place in `subprocess.Popen.__del__`, that prevents it
from trying to append the process to the `_active`. This patch also defines
`subprocess._cleanup` as a noop for Windows.

[1] https://bugs.python.org/issue37380#msg346333
[2] https://bugs.python.org/issue36067#msg336262

Signed-off-by: Ruslan Kuprieiev <ruslan@iterative.ai>

https://bugs.python.org/issue37380

### DIFF
--- a/Lib/subprocess.py
+++ b/Lib/subprocess.py
@@ -218,22 +218,38 @@ else:
         _PopenSelector = selectors.SelectSelector
 
 
-# This lists holds Popen instances for which the underlying process had not
-# exited at the time its __del__ method got called: those processes are wait()ed
-# for synchronously from _cleanup() when a new Popen object is created, to avoid
-# zombie processes.
-_active = []
+if _mswindows:
+    # On Windows we just need to close `Popen._handle` when we no longer need
+    # it, so that the kernel can free it. `Popen._handle` gets closed
+    # implicitly when the `Popen` instance is finalized (see `Handle.__del__`,
+    # which is calling `CloseHandle` as requested in [1]), so there is nothing
+    # for `_cleanup` to do.
+    #
+    # [1] https://docs.microsoft.com/en-us/windows/desktop/ProcThread/
+    # creating-processes
+    _active = None
 
-def _cleanup():
-    for inst in _active[:]:
-        res = inst._internal_poll(_deadstate=sys.maxsize)
-        if res is not None:
-            try:
-                _active.remove(inst)
-            except ValueError:
-                # This can happen if two threads create a new Popen instance.
-                # It's harmless that it was already removed, so ignore.
-                pass
+    def _cleanup():
+        pass
+else:
+    # This lists holds Popen instances for which the underlying process had not
+    # exited at the time its __del__ method got called: those processes are
+    # wait()ed for synchronously from _cleanup() when a new Popen object is
+    # created, to avoid zombie processes.
+    _active = []
+
+    def _cleanup():
+        if _active is None:
+            return
+        for inst in _active[:]:
+            res = inst._internal_poll(_deadstate=sys.maxsize)
+            if res is not None:
+                try:
+                    _active.remove(inst)
+                except ValueError:
+                    # This can happen if two threads create a new Popen instance.
+                    # It's harmless that it was already removed, so ignore.
+                    pass
 
 PIPE = -1
 STDOUT = -2
--- a/Lib/test/test_subprocess.py
+++ b/Lib/test/test_subprocess.py
@@ -59,10 +59,14 @@ class BaseTestCase(unittest.TestCase):
         support.reap_children()
 
     def tearDown(self):
-        for inst in subprocess._active:
-            inst.wait()
-        subprocess._cleanup()
-        self.assertFalse(subprocess._active, "subprocess._active not empty")
+        if not mswindows:
+            # subprocess._active is not used on Windows and is set to None.
+            for inst in subprocess._active:
+                inst.wait()
+            subprocess._cleanup()
+            self.assertFalse(
+                subprocess._active, "subprocess._active not empty"
+            )
         self.doCleanups()
         support.reap_children()
 
@@ -2679,8 +2683,12 @@ class POSIXProcessTestCase(BaseTestCase):
         with support.check_warnings(('', ResourceWarning)):
             p = None
 
-        # check that p is in the active processes list
-        self.assertIn(ident, [id(o) for o in subprocess._active])
+        if mswindows:
+            # subprocess._active is not used on Windows and is set to None.
+            self.assertIsNone(subprocess._active)
+        else:
+            # check that p is in the active processes list
+            self.assertIn(ident, [id(o) for o in subprocess._active])
 
     def test_leak_fast_process_del_killed(self):
         # Issue #12650: on Unix, if Popen.__del__() was called before the
@@ -2701,8 +2709,12 @@ class POSIXProcessTestCase(BaseTestCase):
             p = None
 
         os.kill(pid, signal.SIGKILL)
-        # check that p is in the active processes list
-        self.assertIn(ident, [id(o) for o in subprocess._active])
+        if mswindows:
+            # subprocess._active is not used on Windows and is set to None.
+            self.assertIsNone(subprocess._active)
+        else:
+            # check that p is in the active processes list
+            self.assertIn(ident, [id(o) for o in subprocess._active])
 
         # let some time for the process to exit, and create a new Popen: this
         # should trigger the wait() of p
@@ -2714,7 +2726,11 @@ class POSIXProcessTestCase(BaseTestCase):
                 pass
         # p should have been wait()ed on, and removed from the _active list
         self.assertRaises(OSError, os.waitpid, pid, 0)
-        self.assertNotIn(ident, [id(o) for o in subprocess._active])
+        if mswindows:
+            # subprocess._active is not used on Windows and is set to None.
+            self.assertIsNone(subprocess._active)
+        else:
+            self.assertNotIn(ident, [id(o) for o in subprocess._active])
 
     def test_close_fds_after_preexec(self):
         fd_status = support.findfile("fd_status.py", subdir="subprocessdata")
--- a/Misc/NEWS.d/next/Windows/2019-06-25-04-15-22.bpo-37380.tPxjuz.rst
+++ b/Misc/NEWS.d/next/Windows/2019-06-25-04-15-22.bpo-37380.tPxjuz.rst
@@ -1,0 +1,2 @@
+Don't collect unfinished processes with ``subprocess._active`` on Windows to
+cleanup later. Patch by Ruslan Kuprieiev.


the-knights-who-say-ni:

Misc/NEWS.d/next/Windows/2019-06-25-04-15-22.bpo-37380.tPxjuz.rst
```diff
@@ -0,0 +1,2 @@
+Don't use ``subprocess._collect`` and ``subprocess.Popen.__del__`` to
```
eryksun: It's `_cleanup`, not `_collect`.

Lib/subprocess.py
```diff
@@ -908,6 +913,8 @@ def __exit__(self, exc_type, value, traceback):
             self.wait()
 
     def __del__(self, _maxsize=sys.maxsize, _warn=warnings.warn):
+        if _mswindows:
+            return
```
eryksun: Maybe explain why nothing needs to be done in Windows by repeating or referencing the comment in `__init__`.

efiop:
Thanks for the review @eryksun ! Addressed your comments in the patch.

Lib/subprocess.py
```diff
@@ -908,6 +913,10 @@ def __exit__(self, exc_type, value, traceback):
             self.wait()
 
     def __del__(self, _maxsize=sys.maxsize, _warn=warnings.warn):
+        if _mswindows:
+            # We don't need to do anything in Windows. See `_cleanup` comment
+            # in `Popen.__init__` for more info.
+            return
```
vstinner: That's wrong: you must also wait for process completion on Windows.

Lib/subprocess.py
```diff
@@ -712,7 +712,12 @@ def __init__(self, args, bufsize=-1, executable=None,
                  restore_signals=True, start_new_session=False,
                  pass_fds=(), *, encoding=None, errors=None, text=None):
         """Create new Popen instance."""
-        _cleanup()
+        if not _mswindows:
+            # Unlike Unix, a process in Windows doesn't have to be waited on
+            # by its parent to avoid a zombie. Keeping the handle open will
+            # actually create a zombie until the next _cleanup() call, which
+            # may be never if Popen() isn't called again.
+            _cleanup()
```
vstinner: Can't you modify _cleanup() to do nothing if _active is None, rather than modify the code here?

Just initialize _active to None on Windows, to make Popen code look the same on all platforms.

vstinner (changes requested):

Lib/subprocess.py
efiop: Hi @vstinner ! Thanks for the review! On Windows we just need to close the handle when we no longer need it, so that the kernel can free it. `CloseHandle` is called as a `__del__` in `class Handle` defined above, so there are no additional actions required here, and so we can just skip the whole `__del__` logic here. From [1]

```
When you no longer need these handles, close them by using the CloseHandle function.
```

[1] https://docs.microsoft.com/en-us/windows/desktop/ProcThread/creating-processes

Lib/subprocess.py
efiop: Great point! Adjusted my patch. Thanks!

Lib/subprocess.py
```diff
@@ -908,6 +915,16 @@ def __exit__(self, exc_type, value, traceback):
             self.wait()
 
     def __del__(self, _maxsize=sys.maxsize, _warn=warnings.warn):
+        if _mswindows:
```
zooba: Let's just skip defining `__del__` completely. Python is happier without it

Lib/subprocess.py
```diff
@@ -222,9 +222,16 @@ def __repr__(self):
...
-_active = []
+#
+# Unlike Unix, a process in Windows doesn't have to be waited on by its parent
+# to avoid a zombie. Keeping the handle open will actually create a zombie
+# until the next _cleanup() call, which may be never if Popen() isn't called
+# again.
+_active = None if _mswindows else []
 
 def _cleanup():
+    if _active is None:
```
zooba: Define `_cleanup` as an empty function on Windows.

Lib/subprocess.py
zooba: @vstinner If we're in `__del__`, presumably that means that nobody has a reference to the `Popen` object, right? In that case, we can let it be cleaned up. Closing the handle does not terminate the process - it'll just keep running and we can't get back to it through that particular handle (but we could use `OpenProcess` with the PID to get a new handle later).

Lib/test/test_subprocess.py
```diff
@@ -59,6 +59,8 @@ def setUp(self):
         support.reap_children()
 
     def tearDown(self):
+        if mswindows:
```
zooba: I don't care so much about whether we early-exit from `tearDown` or just don't define it, but it's probably easier to read the code if we skip the definition completely. It might also help prevent people from adding code that _does_ matter on Windows at the end of the function without noticing that it's not going to run.

vstinner:
Steve:
> @vstinner If we're in __del__, presumably that means that nobody has a reference to the Popen object, right? In that case, we can let it be cleaned up. Closing the handle does not terminate the process - it'll just keep running and we can't get back to it through that particular handle (but we could use OpenProcess with the PID to get a new handle later).

My worry is that a badly written program may "leak" programs which continue to run in the background which can cause different issues. For example, you may not be able to remove the directory used by these background tasks since they have a handle open on these directories. It can cause various different issues.

The idea of adding a detatch() method to Popen was proposed once but didn't convince enough people to be implemented: https://bugs.python.org/issue27068

The purpose of the ResourceWarning is to help the developer to write more reliable softwares. It's just a matter of a handle which remains open longer than expected.

I'm talking about this warning:

```
$ cat x.py 
import subprocess, sys, gc
proc = subprocess.Popen([sys.executable, '-c', 'import time; time.sleep(60)'])
proc = None
gc.collect()

$ python3 -Wd x.py 
/usr/lib64/python3.7/subprocess.py:858: ResourceWarning: subprocess 12920 is still running
  ResourceWarning, source=self)
```

I would like to also get this warning on Windows.

efiop:
Guys, big thanks for the reviews! I've adjusted my patch, so that it only re-defines `_active` and `_cleanup` for windows. I didn't notice right away that `__del__` is actually already able to deal with `_active` being `None`, so that we don't need to introduce any changes to `__del__` at all :tada: So we will still get the ResourceWarning for Windows.

Lib/test/test_subprocess.py
```diff
@@ -58,13 +58,16 @@ def setUp(self):
...
-    def tearDown(self):
-        for inst in subprocess._active:
-            inst.wait()
-        subprocess._cleanup()
-        self.assertFalse(subprocess._active, "subprocess._active not empty")
-        self.doCleanups()
-        support.reap_children()
+    if not mswindows:
+        # No need to do anything for Windows. See a note for
+        # subprocess._cleanup() for more info.
```
vstinner: Please only skip the code which rely on _active: calling support.reap_children()  on Windows is still useful.

Lib/test/test_subprocess.py
```diff
@@ -2662,6 +2665,7 @@ def test_select_unbuffered(self):
         finally:
             p.wait()
 
+    @unittest.skipIf(mswindows, "POSIX specific tests")
```
vstinner: This test (ResourceWarning) is still relevant on Windows, no? Just skip the test on _active on Windows.

Lib/test/test_subprocess.py
```diff
@@ -2682,6 +2686,7 @@ def test_zombie_fast_process_del(self):
         # check that p is in the active processes list
         self.assertIn(ident, [id(o) for o in subprocess._active])
 
+    @unittest.skipIf(mswindows, "POSIX specific tests")
```
vstinner:
```suggestion
    @unittest.skipIf(mswindows, "Windows doesn't use subprocess._active]")
```

Lib/subprocess.py
```diff
@@ -218,22 +218,39 @@ def __repr__(self):
...
-                # This can happen if two threads create a new Popen instance.
-                # It's harmless that it was already removed, so ignore.
-                pass
+if _mswindows:
+    # On Windows we just need to close `Popen._handle` when we no longer need
+    # it, so that the kernel can free it. Keeping the handle open will actually
+    # create a zombie until the next `_cleanup()` call, which may be never if
+    # `Popen()` isn't called again. `CloseHandle` is called as a `__del__` in
+    # `class Handle` defined above, so there is no need to collect it as
+    # `_active` in `Popen.__del__` if it didn't finish yet.
```
vstinner: I'm now confused. This comment explains that __del__() must be called to avoid zombie process. But what is a process exit shortly, before __del__()? This comment basically explains that _active is needed, but then set it to None. I'm confused :-)

vstinner (changes requested):

vstinner:
I don't see where this PR explicitly closes the handle: https://bugs.python.org/issue37380#msg346568

Lib/subprocess.py
```diff
@@ -218,22 +218,39 @@ def __repr__(self):
...
-                # It's harmless that it was already removed, so ignore.
-                pass
+if _mswindows:
+    # On Windows we just need to close `Popen._handle` when we no longer need
+    # it, so that the kernel can free it. Keeping the handle open will actually
+    # create a zombie until the next `_cleanup()` call, which may be never if
+    # `Popen()` isn't called again. `CloseHandle` is called as a
+    # `Popen._handle.__del__`, so there is no need to collect it as `_active`
+    # in `Popen.__del__` if it didn't finish yet.
+    #
```
eryksun, line 227: I don't think it helps to talk about zombies here. It's not a visible zombie in the task list, unlike Unix. The process has already been rundown in the kernel. What remains, as long as it's kept alive by a handle or pointer reference, is the kernel Process object that has information such as the exit status. What we may need to clarify for the reader is that `Popen._handle` gets closed implicitly when the `Popen` instance is finalized, so there is nothing for `_cleanup` to do in Windows.

eryksun:
As to `Popen.__del__`, I don't understand why we force scripts to always have to `poll()` in order to avoid a resource warning when warnings are enabled. Why can't we instead check whether `self._internal_poll() is None` in order to quietly handle cases where the program has exited already? For example:

    >>> p = subprocess.Popen('python -c ""')
    >>> del p
    C:\Program Files\Python37\lib\subprocess.py:858: ResourceWarning: subprocess 3180 is still running
      ResourceWarning, source=self)

The process exited. It is not running. The warning is wrong.

eryksun:
In a related issue to waiting on the child process, if someone wants to open an easy issue for this, we should not be creating `self._waitpid_lock = threading.Lock()` in Windows. It isn't used.

Lib/test/test_subprocess.py
```diff
@@ -59,10 +59,14 @@ def setUp(self):
...
+        if not mswindows:
+            # subprocess._active is not used on Windows and is set to None.
+            for inst in subprocess._active:
+                inst.wait()
+            subprocess._cleanup()
+            self.assertFalse(
+                subprocess._active, "subprocess._active not empty"
+            )
         self.doCleanups()
         support.reap_children()
```
eryksun, line 71: Calling `support.reap_children` is pointless but harmless in Windows. It has nothing to do:

    # Need os.waitpid(-1, os.WNOHANG): Windows is not supported
    if not (hasattr(os, 'waitpid') and hasattr(os, 'WNOHANG')):
        return

vstinner:
> As to Popen.__del__, I don't understand why we force scripts to always have to poll() in order to avoid a resource warning when warnings are enabled. Why can't we instead check whether self._internal_poll() is None instead of self.returncode is None, in order to quietly handle cases where the program has exited already?

The purpose is to promote to handle explicitly resources, rather than relying on Python garbage collector to magically handle them: the garbage collection behaves differently on PyPy, and may not collect objects "immediately" depending if they are part of a reference cycle or not.

subprocess.Popen should "release" its resources when the process exit: close pipes, "close" the child processes, etc. I suggest to use "with popen: ..." since __exit__() closes pipes and waits until the child process completes.

vstinner:
In the case of Windows, I would expect that the handle can be closed as soon as the process completes. But it seems to be the case now: https://bugs.python.org/issue37380#msg346568

Lib/test/test_subprocess.py
efiop: Sorry, I don't quite understand why this is harmless. Could you please elaborate?

efiop: Ah, sorry, misread harm**less** as harm**ful**.

efiop:
Adjusted the `_cleanup` comment. Please take a look.

Lib/test/test_subprocess.py
eryksun: Victor had mentioned that calling `support.reap_children` was useful in Windows, but the snippet of code above shows that it doesn't do anything in Windows.

Lib/subprocess.py
```diff
@@ -218,22 +218,38 @@ def __repr__(self):
...
-                # This can happen if two threads create a new Popen instance.
-                # It's harmless that it was already removed, so ignore.
-                pass
+if _mswindows:
+    # On Windows we just need to close `Popen._handle` when we no longer need
+    # it, so that the kernel can free it. `Popen._handle` gets closed
+    # implicitly when the `Popen` instance is finalized (see `Handle.__del__`,
+    # which is calling `CloseHandle` as requested in [1]), so there is nothing
+    # for `_cleanup` to do.
+    #
```
eryksun, line 227: LGTM. 

FYI, if [bpo37410](https://bugs.python.org/issue37410) is resolved, we'll also be explicitly closing `_handle` when `returncode` is set. This would be a significant change. Currently we can assume `_handle` is valid, but with the proposed change we'll have to acquire a lock and check whether `self.returncode is None` before code that sets it in `_internal_poll`, `_wait`, and `terminate`.

PR #14365, opened 2019-06-25
bpo-37392: Remove getcheckinterval and setcheckinterval from doc output

tirkarthi:
Remove `getcheckinterval` and `setcheckinterval` from doc output since they were removed with 36456df13843c5b8a1fb5a6022ab9ed1fe2a11c5 .

https://bugs.python.org/issue37392

### DIFF
--- a/Doc/tutorial/modules.rst
+++ b/Doc/tutorial/modules.rst
@@ -306,23 +306,27 @@ defines.  It returns a sorted list of strings::
    >>> dir(fibo)
    ['__name__', 'fib', 'fib2']
    >>> dir(sys)  # doctest: +NORMALIZE_WHITESPACE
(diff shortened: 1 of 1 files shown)


vstinner:
Well, I saw it that I don't think it hurts too much to have outdated examples :-) If you care, I would prefer to update the whole list rather than just remove old ones: sys got new symbols in the meanwhile.

tirkarthi:
I updated it with the latest list. I just came across this while grepping and thought this was missed. But as you have mentioned I am okay with this to be closed too since the output was outdated for long time and also that it's in tutorial so it might not be synced regularly with the changes.

vstinner:
I rewrote the commit message ;-) I don't think that it's worth it to backport the change.

tirkarthi:
Thanks Victor.

PR #14374, opened 2019-06-25
bpo-37400: Fix test_os.test_chown()

vstinner:
Use os.getgroups() rather than pwd.getpwall() to get groups.
Rename also test_chown() to test_chown_gid().

https://bugs.python.org/issue37400


vstinner:
Oh, grp.getgrall() was used, not pwd.getpwall(). I updated my PR.

PR #14379, opened 2019-06-25
Improve Windows commands in tutorial

zooba:


zooba:
This should be backported to 3.8 and 3.7, but I'll do it manually to update the referenced version numbers.

PR #14385, opened 2019-06-25
bpo-37388: Add PyUnicode_Decode(str, 0) fast-path

vstinner:
Add a fast-path to PyUnicode_Decode() for size equals to 0.

https://bugs.python.org/issue37388


vstinner:
PyUnicode_FromEncodedObject() already has the same fast-path.

PR #14391, opened 2019-06-26
bpo-37410: subprocess closes the process handle when done

vstinner:
On Windows, subprocess.Popen now closes the process handle when the
process completes. Previously, it was only closed when the Popen
object was destroyed.

https://bugs.python.org/issue37410

### DIFF
--- a/Lib/subprocess.py
+++ b/Lib/subprocess.py
@@ -1300,6 +1300,12 @@ class Popen(object):
             self.pid = pid
             _winapi.CloseHandle(ht)
 
+
+        def _set_returncode(self, exitcode):
+            self.returncode = exitcode
+            self._handle.Close()
+            self._handle = None
+
         def _internal_poll(self, _deadstate=None,
                 _WaitForSingleObject=_winapi.WaitForSingleObject,
                 _WAIT_OBJECT_0=_winapi.WAIT_OBJECT_0,
@@ -1313,11 +1319,9 @@ class Popen(object):
             """
             if self.returncode is None:
                 if _WaitForSingleObject(self._handle, 0) == _WAIT_OBJECT_0:
-                    self.returncode = _GetExitCodeProcess(self._handle)
-                    self._handle.Close()
-                    self._handle = None
+                    exitcode = _GetExitCodeProcess(self._handle)
+                    self._set_returncode(exitcode)
             return self.returncode
-
 
         def _wait(self, timeout):
             """Internal implementation of wait() on Windows."""
@@ -1331,7 +1335,8 @@ class Popen(object):
                                                      timeout_millis)
                 if result == _winapi.WAIT_TIMEOUT:
                     raise TimeoutExpired(self.args, timeout)
-                self.returncode = _winapi.GetExitCodeProcess(self._handle)
+                exitcode = _winapi.GetExitCodeProcess(self._handle)
+                self._set_returncode(exitcode)
             return self.returncode
 
 
@@ -1419,7 +1424,7 @@ class Popen(object):
                 rc = _winapi.GetExitCodeProcess(self._handle)
                 if rc == _winapi.STILL_ACTIVE:
                     raise
-                self.returncode = rc
+                self._set_returncode(rc)
 
         kill = terminate
 
@@ -1640,7 +1645,7 @@ class Popen(object):
                 try:
                     pid, sts = os.waitpid(self.pid, 0)
                     if pid == self.pid:
-                        self._handle_exitstatus(sts)
+                        self._set_returncode(sts)
                     else:
                         self.returncode = sys.maxsize
                 except ChildProcessError:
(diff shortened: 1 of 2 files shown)


vstinner:
cc @efiop @eryksun @zooba

vstinner:
See also https://bugs.python.org/issue37380

vstinner:
cc @gpshead

Lib/subprocess.py
```diff
@@ -1314,6 +1314,8 @@ def _internal_poll(self, _deadstate=None,
             if self.returncode is None:
                 if _WaitForSingleObject(self._handle, 0) == _WAIT_OBJECT_0:
                     self.returncode = _GetExitCodeProcess(self._handle)
+                    self._handle.Close()
+                    self._handle = None
```
eryksun: `__exit__` calls `wait`, which calls `_wait`, which does not call `_internal_poll`. This should be implemented in `_wait` first and foremost. It can also be implemented here, or modify `_internal_poll` to call `_wait(0)`.

vstinner: Oh right, I forgot wait() and terminate(). It's now fixed.

I don't want to bother with a timeout exception in _internal_poll(), so I didn't modify it to reuse wait().

vstinner:
Oh, a test now fails on Windows:

```
ERROR: test_threadsafe_wait (test.test_subprocess.ProcessTestCase)
Issue21291: Popen.wait() needs to be threadsafe for returncode.
  File "C:\projects\cpython\lib\test\test_subprocess.py", line 1284, in test_threadsafe_wait
    proc.wait(timeout=20)
  File "C:\projects\cpython\lib\subprocess.py", line 1050, in wait
    return self._wait(timeout=timeout)
  File "C:\projects\cpython\lib\subprocess.py", line 1340, in _wait
    exitcode = _winapi.GetExitCodeProcess(self._handle)
TypeError: GetExitCodeProcess() argument must be int, not None
```

It seems like _wait() is not longer thread-safe with my change... but I'm surprised that it was thread-safe before my change.

The Unix implementation uses a lock to make _wait() thread-safe. Extract:

```
                    with self._waitpid_lock:
                        if self.returncode is not None:
                            break  # Another thread waited.
                        (pid, sts) = self._try_wait(0)
                        # Check the pid and loop as waitpid has been known to
                        # return 0 even without WNOHANG in odd situations.
                        # http://bugs.python.org/issue14396.
                        if pid == self.pid:
                            self._set_returncode(sts)
```

Maybe the Windows implementation should also use a lock?

The failing test comes from https://bugs.python.org/issue21291

eryksun:
> It seems like _wait() is not longer thread-safe with my change... but I'm surprised that it was thread-safe before my change.

Before we could rely on `_handle` being valid. Now I think we need to acquire a lock before checking and setting the return code. The following is an untested version of `_wait` that acquires a lock:

```python
    def _wait(self, timeout):
        """Internal implementation of wait() on Windows."""
        if self.returncode is not None:
            return self.returncode

        if timeout is None:
            self._status_lock.acquire()
        else:
            if timeout < 0:
                # Disallow the -1 default timeout value of acquire.
                raise ValueError('timeout value must be positive')
            endtime = _time() + timeout
            acquired = self._status_lock.acquire(timeout=timeout)
            remaining = endtime - _time() if timeout != 0 else 0
            if not acquired or remaining < 0:
                raise TimeoutExpired(self.args, timeout)
        try:
            if self.returncode is None:
                # API note: returns immediately if remaining is 0.
                result = _winapi.WaitForSingleObject(self._handle,
                    _winapi.INFINITE if timeout is None else
                    int(remaining * 1000))
                if result == _winapi.WAIT_TIMEOUT:
                    raise TimeoutExpired(self.args, timeout)
                exitcode = _winapi.GetExitCodeProcess(self._handle)
                self._set_returncode(exitcode)
        finally:
            self._status_lock.release()
        return self.returncode
```

efiop:
@eryksun It seems like we could re-use `_waitpid_lock`(it was requested for removal in https://github.com/python/cpython/pull/14360#issuecomment-505665385) instead of introducing new `_status_lock`, right?

vstinner:
The Unix code path has the same issue and it was already been fixed. I suggest to factorize the code between Windows and Unix: reuse the same lock and the same locking code, but add sub-methods for Windows or Unix specific code.

I plan to write such PR.

eryksun:
> The Unix code path has the same issue and it was already been fixed. I suggest to factorize the code between Windows and Unix: reuse the same lock and the same locking code, but add sub-methods for Windows or Unix specific code.

The POSIX code uses a busy loop. I see no reason for that in Windows.

vstinner:
My PR doesn't work: more locking is needed. Code should be shared with Unix, or copied from the Unix implementation, to add locking. I will not be available next weeks, so I prefer to close the issue. I may rewrite it properly later.

PR #14393, opened 2019-06-26
Use "kwargs" argument name in documentation

johnthagen:
I've seen the vast majority of documentation and examples use `**kwargs` to name the keyword arguments. This change brings this part of the docs in line with what most users should expect.

### DIFF
--- a/Doc/library/functools.rst
+++ b/Doc/library/functools.rst
@@ -544,9 +544,9 @@ The :mod:`functools` module defines the following functions:
       >>> from functools import wraps
       >>> def my_decorator(f):
       ...     @wraps(f)
-      ...     def wrapper(*args, **kwds):
+      ...     def wrapper(*args, **kwargs):
       ...         print('Calling decorated function')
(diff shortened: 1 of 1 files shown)


jdemeyer (approved):
Makes sense for me.

merwok:
Quick doc search:
- we see kwargs, kw, kwds, keywords
- also options, params, _params
- more specific: env_vars, fmtparams, x509

IMO there is no problem to fix here, and the change wouldn’t really improve the doc.  The name kwargs is just a convention; seeing `kwds` may be the trigger for a beginner Pythonista to understand that the name is not special but `**` makes var kwargs.

rhettinger:
I don't think this needs to be changed.  There is no standard here.  You will see `**d`, `**kwds`, and many other variants.  They are all allowed and though inconsistent do have the minor virtue of showing the word is not a keyword and that any valid identifier will suffice.

Thank you for the suggestion though.

PR #14394, opened 2019-06-26
bpo-37411: Rewrite test_wsgiref.testEnviron()

vstinner:
Fix test_wsgiref.testEnviron() to no longer depend on the environment
variables (don't fail if "X" variable is set).

testEnviron() now overrides os.environ to get a deterministic
environment. Test full TestHandler.environ content: not only a few
selected variables.

https://bugs.python.org/issue37411

### DIFF
--- a/Lib/test/test_wsgiref.py
+++ b/Lib/test/test_wsgiref.py
@@ -548,32 +548,62 @@ class TestHandler(ErrorHandler):
 
 
 class HandlerTests(TestCase):
-
-    def checkEnvironAttrs(self, handler):
-        env = handler.environ
-        for attr in [
-            'version','multithread','multiprocess','run_once','file_wrapper'
-        ]:
-            if attr=='file_wrapper' and handler.wsgi_file_wrapper is None:
-                continue
-            self.assertEqual(getattr(handler,'wsgi_'+attr),env['wsgi.'+attr])
-
-    def checkOSEnviron(self,handler):
-        empty = {}; setup_testing_defaults(empty)
-        env = handler.environ
-        from os import environ
-        for k,v in environ.items():
-            if k not in empty:
-                self.assertEqual(env[k],v)
-        for k,v in empty.items():
-            self.assertIn(k, env)
+    # testEnviron() can produce long error message
+    maxDiff = 80 * 50
 
     def testEnviron(self):
-        h = TestHandler(X="Y")
-        h.setup_environ()
-        self.checkEnvironAttrs(h)
-        self.checkOSEnviron(h)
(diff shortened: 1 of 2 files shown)


vstinner:
@tirkarthi: Would you be interested to review my PR?

Lib/test/test_wsgiref.py
```diff
@@ -548,32 +548,62 @@ def handle_error(self):
...
+            'HOME': "/override/home",
+
+            # overriden by setup_testing_defaults()
+            "SCRIPT_NAME": "",
+            "SERVER_NAME": "127.0.0.1",
+
+            # set by BaseHandler.setup_environ()
+            'wsgi.input': handler.get_stdin(),
+            'wsgi.errors': handler.get_stderr(),
+            'wsgi.version': (1, 0),
```
tirkarthi, line 599: Clarification : Is it okay to let these values to be taken from handler's attributes so that the tests don't have to worry about these default values being changed in future? Since we just need to check `handler.wsgi_version` is set correctly with https://github.com/python/cpython/blob/c6a2320e876354ee62cf8149b849bcff9492d38a/Lib/wsgiref/handlers.py#L160 in setup_environ. With this change the below assertion checks inside for loops are also not needed. Anything I am missing with setting these values?

Something like below : 

```
'wsgi.input': handler.get_stdin(),
'wsgi.errors': handler.get_stderr(),
'wsgi.version': handler.wsgi_version,
'wsgi.run_once': handler.wsgi_run_once,
'wsgi.url_scheme': handler.get_scheme(),
'wsgi.multithread': handler.wsgi_multithread,
'wsgi.multiprocess': handler.wsgi_multiprocess,
'wsgi.file_wrapper': util.FileWrapper,
```

vstinner: I made a deliberate choice of copy directly the expected value rather than "reimplement" the logic creating the environ dictionary. I don't expect that default values will change often. If they change, well, the CI will tell you that and it will be trivial to update the tests.

vstinner: @tirkarthi: the test already checks that wsgi_xxx attributes are copied to wsgi.xxx env vars. I exchanged the two tests in testEnviron() so attributes are checked before hardcoded values. Are you fine with hardcoded values? Or do you prefer that I explicitly copy wsgi_xxx attributes?

tirkarthi (approved):
LGTM. Thanks Victor for the details.

vstinner:
test_asyncio failed on AppVeyor. I manually scheduled a rebuild.

PR #14395, opened 2019-06-26
bpo-33926: enable GDB tests on Travis CI

jdemeyer:
Fixes https://bugs.python.org/issue33926

See also https://bugs.python.org/issue37382

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -14,8 +14,7 @@ env:
     - OPENSSL=1.1.1c
     - OPENSSL_DIR="$HOME/multissl/openssl/${OPENSSL}"
     - PATH="${OPENSSL_DIR}/bin:$PATH"
-    # Use -Og because we want to run all GDB tests
-    - CFLAGS="-I${OPENSSL_DIR}/include -Og"
+    - CFLAGS="-I${OPENSSL_DIR}/include"
     - LDFLAGS="-L${OPENSSL_DIR}/lib"
     # Set rpath with env var instead of -Wl,-rpath linker flag
     # OpenSSL ignores LDFLAGS when linking bin/openssl
--- a/.travis.yml
+++ b/.travis.yml
@@ -14,8 +14,8 @@ env:
     - OPENSSL=1.1.1c
     - OPENSSL_DIR="$HOME/multissl/openssl/${OPENSSL}"
     - PATH="${OPENSSL_DIR}/bin:$PATH"
-    # Use -O3 because we don't use debugger on Travis-CI
-    - CFLAGS="-I${OPENSSL_DIR}/include -O3"
+    # Use -Og because we want to run all GDB tests
+    - CFLAGS="-I${OPENSSL_DIR}/include -Og"
     - LDFLAGS="-L${OPENSSL_DIR}/lib"
     # Set rpath with env var instead of -Wl,-rpath linker flag
     # OpenSSL ignores LDFLAGS when linking bin/openssl
@@ -43,6 +43,7 @@ matrix:
       addons:
         apt:
           packages:
+            - gdb
             - xvfb
     - name: "Documentation build"
       os: linux


.travis.yml
```diff
@@ -14,8 +14,8 @@ env:
     - OPENSSL=1.1.1c
     - OPENSSL_DIR="$HOME/multissl/openssl/${OPENSSL}"
     - PATH="${OPENSSL_DIR}/bin:$PATH"
-    # Use -O3 because we don't use debugger on Travis-CI
-    - CFLAGS="-I${OPENSSL_DIR}/include -O3"
```
methane, line 18: Please don't change here.  This environment variable is used for all build, including document test.

tirkarthi:
I guess this also resolves https://bugs.python.org/issue33926 . Thanks.

.travis.yml
```diff
@@ -14,8 +14,8 @@ env:
...
+    # Use -Og because we want to run all GDB tests
+    - CFLAGS="-I${OPENSSL_DIR}/include -Og"
```
methane:
```suggestion
    - CFLAGS="-I${OPENSSL_DIR}/include"
```

.travis.yml
```diff
@@ -14,8 +14,8 @@ env:
     - OPENSSL=1.1.1c
     - OPENSSL_DIR="$HOME/multissl/openssl/${OPENSSL}"
     - PATH="${OPENSSL_DIR}/bin:$PATH"
-    # Use -O3 because we don't use debugger on Travis-CI
-    - CFLAGS="-I${OPENSSL_DIR}/include -O3"
+    # Use -Og because we want to run all GDB tests
```
methane: remove this comment line.

methane (changes requested):

jdemeyer:
@methane: we need to use `-O0` or `-Og` as optimization flag. This is because gdb tests are not run when CPython is compiled with the default `-O3` level (IMHO, this is in itself a bug, however fixing that is more controversial that I expected, see #14331).

jdemeyer:
> I guess this also resolves https://bugs.python.org/issue33926 . Thanks.

In fact, it fixes exactly that. So I'm changing the bpo number here.

.travis.yml
jdemeyer: So what would you suggest instead?

Note that this affects only the Linux Travis CI build. On other platforms and with other CI services, we still use `-O3`.

methane: > So what would you suggest instead?

Just remove '-O3`.

> Note that this affects only the Linux Travis CI build. 

Affects *all* of Linux Travis CI build.

methane:
> @methane: we need to use `-O0` or `-Og` as optimization flag.

When `configure --with-pydebug` is used, it is set automatically.  No need to set it explicitly.

tirkarthi:
@jdemeyer Can you please change the issue number in PR description from 32926 to 33926? I guess the bot keeps linking this PR to the issue 32926 for every comment made in this PR.

jdemeyer:
> When `configure --with-pydebug` is used, it is set automatically.

OK, let's test that.

.travis.yml
methane: https://travis-ci.org/python/cpython/jobs/551191274

Actually, it affects even macOS build.

jdemeyer:
OK, this seems to work. `I have made the requested changes; please review again`

tirkarthi:
Does this change have any noticeable impact on CI timing? The merge run happened for [20 mins](https://travis-ci.org/python/cpython/jobs/551215123#L2423) which used to be 10-12 mins earlier. The backport PR also completes in around 20 minutes.

methane:
> Does this change have any noticeable impact on CI timing? The merge run happened for [20 mins](https://travis-ci.org/python/cpython/jobs/551215123#L2423) which used to be 10-12 mins earlier. The backport PR also completes in around 20 minutes.

Yes, it affects.  I will compare the time of `-Og` and `-O0`.

methane:
```
-Og
real    5m3.236s
user    16m29.552s
sys     0m57.023s

-O0
real    6m56.909s
user    30m5.529s
sys     1m0.605s
```

vstinner:
> Does this change have any noticeable impact on CI timing? The merge run happened for 20 mins which used to be 10-12 mins earlier. The backport PR also completes in around 20 minutes.

* On PR #14390: "CPython tests" job took 15 min 39 sec
* On this PR, it took 22 min 23 sec: +6 min 44 sec, it's quite significant

vstinner:
* PR #14390: "Total duration: 11 min 8 sec" just for tests (regrtest)
* On this PR: "Total duration: 20 min 30 sec": +9 min 22 sec

It's 1.8x slower.

vstinner:
Previously, we used -O3:

```
commit 8ff53564730f7ba503f5ad94418c309c48e8516d
Author: INADA Naoki <methane@users.noreply.github.com>
Date:   Sat Feb 10 20:35:17 2018 +0900

    travis: Use -O3 option (GH-5599)
    
    We don't use debugger on Travis.
```

Is it really useful to make the pre-commit CI almost 2x slower just because of test_gdb? We already have a large fleet of post-commit buildbots which run test_gdb.

methane:
https://travis-ci.org/python/cpython/builds/551298808
16min with -Og

PR #14396, opened 2019-06-26
bpo-37412: os.getcwdb() now uses UTF-8 on Windows

vstinner:
The os.getcwdb() function now uses the UTF-8 encoding on Windows,
rather than the ANSI code page: see PEP 529 for the rationale. The
function is no longer deprecated on Windows.

os.getcwd() and os.getcwdb() detect integer overflow on memory
allocations. On Unix, these functions properly report MemoryError on
memory allocation failure.

### DIFF
--- a/Doc/library/os.rst
+++ b/Doc/library/os.rst
@@ -1730,6 +1730,11 @@ features:
 
    Return a bytestring representing the current working directory.
 
+   .. versionchanged:: 3.8
+      The function now uses the UTF-8 encoding on Windows, rather than the ANSI
+      code page: see :pep:`529` for the rationale. The function is no longer
+      deprecated on Windows.
+
 
 .. function:: lchflags(path, flags)
 
--- a/Modules/posixmodule.c
+++ b/Modules/posixmodule.c
@@ -506,17 +506,6 @@ void _Py_attribute_data_to_stat(BY_HANDLE_FILE_INFORMATION *,
                                             ULONG, struct _Py_stat_struct *);
 #endif
 
-#ifdef MS_WINDOWS
-static int
-win32_warn_bytes_api()
-{
-    return PyErr_WarnEx(PyExc_DeprecationWarning,
-        "The Windows bytes API has been deprecated, "
-        "use Unicode filenames instead",
-        1);
-}
-#endif
-
 
 #ifndef MS_WINDOWS
 PyObject *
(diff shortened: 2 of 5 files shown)


Doc/library/os.rst
```diff
@@ -1730,6 +1730,11 @@ features:
 
    Return a bytestring representing the current working directory.
 
+   .. versionchanged:: 3.9
```
zooba: Why not fix it in 3.8 as well? It's not a new feature

Modules/posixmodule.c
```diff
@@ -3334,83 +3323,104 @@ os_lchown_impl(PyObject *module, path_t *path, uid_t uid, gid_t gid)
...
-#endif
+    }
+
+    if (use_bytes) {
+        PyObject *bytes = PyUnicode_EncodeFSDefault(resobj);
+        Py_DECREF(resobj);
+        if (bytes == NULL) {
+            return NULL;
+        }
+        resobj = bytes;
```
zooba: I quite like `Py_SETREF(resobj, PyUnicode_EncodeFSDefault(resobj))` as a replacement for this entire block, but up to you.

zooba (approved):

Doc/library/os.rst
vstinner: It's a backward incompatible change.

Modules/posixmodule.c
vstinner: Done

vstinner:
I retargeted my PR 14396 to Python 3.8.

> Why not fix it in 3.8 as well? It's not a new feature

Sorry, I was confused about 3.8 status. It's not released yet, I'm fine with targeting 3.8. It's better to get this bug fixed ASAP.

Modules/posixmodule.c
```diff
@@ -3334,83 +3323,99 @@ os_lchown_impl(PyObject *module, path_t *path, uid_t uid, gid_t gid)
...
+    wchar_t *wbuf2 = wbuf;
+    DWORD len;
+
+    Py_BEGIN_ALLOW_THREADS
+    len = GetCurrentDirectoryW(Py_ARRAY_LENGTH(wbuf), wbuf);
+    /* If the buffer is large enough, len does not include the
+       terminating \0. If the buffer is too small, len includes
+       the space needed for the terminator. */
+    if (len >= Py_ARRAY_LENGTH(wbuf)) {
+        if (len >= PY_SSIZE_T_MAX / sizeof(wchar_t)) {
```
eryksun, line 3337: `PyMem_RawMalloc` already checks this and returns `NULL`   if  `size > (size_t)PY_SSIZE_T_MAX`. Also, `>=`  is the wrong comparison operator, so this is causing a `MemoryError` on long paths:

```
>>> p = 'C:/Temp/longpath' + ('/' + 'a' * 255) * 9
>>> os.chdir(p)
>>> len(os.getcwd())
MemoryError
```

vstinner: > PyMem_RawMalloc already checks this and returns NULL if size > (size_t)PY_SSIZE_T_MAX.

In CPython, we attempt to avoid undefined behaviors (UB). The C language does not define what happens in case of integer overflow. Maybe PyMem_RawMalloc() check will be enough, maybe not. In case of doubt, we prefer to check *before* the multiply.

I wrote PR #14424 to fix my regression and add a test for it.

eryksun: Right, I didn't think that through clearly. In this case it's academic, though.  In practice the current directory is limited to 32,767 characters since the kernel uses counted strings with a `USHORT` length in bytes.

PR #14397, opened 2019-06-26
bpo-36889: Replace deprecation warning with RuntimeError

asvetlov:
asyncio.Stream is a new class, there is no reason for deprecation period


eamanu (approved):
LGTM

tirkarthi (approved):
LGTM. Thanks.

PR #14398, opened 2019-06-26
bpo-37414: Remove sys.callstats()

vstinner:
Remove the undocumented sys.callstats() function. Since Python 3.7,
it was deprecated and always returned None. It required a special
build option CALL_PROFILE which was already removed in Python 3.7.

https://bugs.python.org/issue37414


vstinner:
cc @jdemeyer

PR #14406, opened 2019-06-26
bpo-37390: Add audit event table to documentations

zooba:
Also updates some (unreleased) event names for consistency.

https://bugs.python.org/issue37390

### DIFF
--- a/Doc/library/persistence.rst
+++ b/Doc/library/persistence.rst
@@ -21,4 +21,3 @@ The list of modules described in this chapter is:
    marshal.rst
    dbm.rst
    sqlite3.rst
-   audit_events.rst
--- a/Doc/library/persistence.rst
+++ b/Doc/library/persistence.rst
@@ -21,3 +21,4 @@ The list of modules described in this chapter is:
    marshal.rst
    dbm.rst
    sqlite3.rst
+   audit_events.rst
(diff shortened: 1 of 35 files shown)


Doc/library/persistence.rst
```diff
@@ -21,3 +21,4 @@ The list of modules described in this chapter is:
    marshal.rst
    dbm.rst
    sqlite3.rst
+   audit_events.rst
```
tirkarthi: I added it here in persistence page since Sphinx kept complaining with a warning that audit_events.rst should be present under a toctree like below. This can be moved to a more appropriate section.

```
Warning, treated as error:
/Users/karthikeyansingaravelan/stuff/python/cpython/Doc/library/audit_events.rst:document isn't included in any toctree
```

tirkarthi:
There is note about purging env in the tutorial I used for initial prototype. Not sure if this is needed for this page. Reference : http://www.sphinx-doc.org/en/stable/extdev/tutorial.html#the-event-handlers

Doc/library/persistence.rst
zooba: 👍 I'll move it under `library/python.rst`

zooba:
I think the env will be updated fine because all the source files have changes now. The important builds are done from a clean state anyway.

Doc/library/persistence.rst
zooba: I actually moved it to `library/debug.rst`

PR #14409, opened 2019-06-26
bpo-37419: Fix possible segfaults when passing large sequences to os.posix_spawn()

ZackerySpytz:
Switch the type of i to Py_ssize_t.

https://bugs.python.org/issue37419


ZackerySpytz:
A `bigmem` test could be added for this, but I don't think it's worth it.

serhiy-storchaka (approved):

pablogsal (approved):

pablogsal:
Good catch!

PR #14414, opened 2019-06-26
bpo-37420: os.sched_setaffinity does not handle errors during iteration.

brandtbucher:
This is related to [bpo-37417](https://bugs.python.org/issue37417): `os.sched_setaffinity` doesn't properly handle errors that arise during iteration of the `mask` argument:

```py
Python 3.9.0a0 (heads/master:d52a83a, Jun 26 2019, 15:13:41) 
[GCC 5.4.0 20160609] on linux
>>> import os
>>> bad_iter = map(int, "0X")
>>> os.sched_setaffinity(0, bad_iter)
ValueError: invalid literal for int() with base 10: 'X'

SystemError: <built-in function sched_setaffinity> returned a result with an error set
```

It looks like this bug is also present on all versions of Python 3. I've attached a patch with a fix and a regression test.

https://bugs.python.org/issue37420

### DIFF
--- a/Lib/test/test_os.py
+++ b/Lib/test/test_os.py
@@ -2560,6 +2560,12 @@ class PidTests(unittest.TestCase):
         status = os.waitpid(pid, 0)
         self.assertEqual(status, (pid, 0))
 
+class AffinityTests(unittest.TestCase):
+
+    @unittest.skipUnless(hasattr(os, 'sched_setaffinity'), "test needs os.sched_setaffinity")
+    def test_os_sched_setaffinity_error(self):
+        bad_iter = map(int, "0X")
+        self.assertRaises(ValueError, os.sched_setaffinity, 0, bad_iter)
 
 class SpawnTests(unittest.TestCase):
     def create_args(self, *, with_env=False, use_bytes=False):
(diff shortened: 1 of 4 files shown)


Lib/test/test_os.py
```diff
@@ -2560,6 +2560,12 @@ def test_waitpid(self):
         status = os.waitpid(pid, 0)
         self.assertEqual(status, (pid, 0))
 
+class AffinityTests(unittest.TestCase):
```
benjaminp: Can you put this in `test_posix` with the other `setaffinity` tests?

Modules/posixmodule.c
```diff
@@ -6415,6 +6415,10 @@ os_sched_setaffinity_impl(PyObject *module, pid_t pid, PyObject *mask)
     }
     Py_CLEAR(iterator);
 
+    if (PyErr_Occurred()) {
```
benjaminp, line 6416: seems like this could be before the `Py_CLEAR`. Closer to the loop is better I think.

Lib/test/test_os.py
brandtbucher: Thanks for pointing that out... I hadn't noticed `test_posix.py`!

Modules/posixmodule.c
brandtbucher: I agree. Done.

mangrisano (approved):
LGTM.

PR #14416, opened 2019-06-26
bpo-37421: Fix test_shutil: don't leak temporary files

vstinner:
* Fix typo in supports_file2file_sendfile(); ensure that dst is
  removed
* Fix test_copytree_custom_copy_function(): remove dst tree.
  Use support.rmtree() rather than shutil.rmtree() to remove
  temporary directories: support tries harder.


vstinner:
I checked: Python 3.7 is not affected, it doesn't contain the code which had the two bugs.

PR #14419, opened 2019-06-27
bpo-37366: add onitem callback argument to shutil.rmtree()

websurfer5:
Add an "onitem" callback paramter to shutil.rmtree() that, if provided, gets called for each directory entry as it is encountered.  This allows the caller to perform any required special handling of individual directory entries (e.g. unmounting a mount point, closing a file, shutting down a named pipe, etc.) before rmtree() attempts to remove them.

https://bugs.python.org/issue37366


kumaraditya303:
Closing as bpo is rejected.

PR #14421, opened 2019-06-27
bpo-37428: Don't set PHA verify flag on client side

tiran:
SSLContext.post_handshake_auth = True no longer sets
SSL_VERIFY_POST_HANDSHAKE verify flag for client connections. Although the
option is documented as ignored for clients, OpenSSL implicitly enables cert
chain validation when the flag is set.

https://bugs.python.org/issue37428

### DIFF
--- a/Lib/test/test_ssl.py
+++ b/Lib/test/test_ssl.py
@@ -4434,6 +4434,37 @@ class TestPostHandshakeAuth(unittest.TestCase):
                 s.write(b'PHA')
                 self.assertIn(b'WRONG_SSL_VERSION', s.recv(1024))
 
+    def test_bpo37428_pha_cert_none(self):
+        # verify that post_handshake_auth does not implicitly enable cert
+        # validation.
+        hostname = SIGNED_CERTFILE_HOSTNAME
+        client_context = ssl.SSLContext(ssl.PROTOCOL_TLS_CLIENT)
+        client_context.post_handshake_auth = True
+        client_context.load_cert_chain(SIGNED_CERTFILE)
+        # no cert validation and CA on client side
+        client_context.check_hostname = False
+        client_context.verify_mode = ssl.CERT_NONE
+
+        server_context = ssl.SSLContext(ssl.PROTOCOL_TLS_SERVER)
+        server_context.load_cert_chain(SIGNED_CERTFILE)
+        server_context.load_verify_locations(SIGNING_CA)
+        server_context.post_handshake_auth = True
+        server_context.verify_mode = ssl.CERT_REQUIRED
+
+        server = ThreadedEchoServer(context=server_context, chatty=False)
+        with server:
+            with client_context.wrap_socket(socket.socket(),
+                                            server_hostname=hostname) as s:
+                s.connect((HOST, server.port))
+                s.write(b'HASCERT')
+                self.assertEqual(s.recv(1024), b'FALSE\n')
+                s.write(b'PHA')
+                self.assertEqual(s.recv(1024), b'OK\n')
+                s.write(b'HASCERT')
+                self.assertEqual(s.recv(1024), b'TRUE\n')
+                # server cert has not been validated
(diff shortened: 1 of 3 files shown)


ned-deily:
FWIW, I tested a manual cherrypick of this to 3.7 on macOS with OpenSSL 1.1.1c and it didn't cause any new failures in `test_ssl` and it wasn't failing before in this environment.  However, FTR, the following non-fatal traceback consistently shows up both before and after this change and both with and without test -j:

```
[...]
test_pha_required_nocert (test.test_ssl.TestPostHandshakeAuth) ... Exception in thread Thread-242:
  File "/Library/Frameworks/Python.framework/Versions/3.7/lib/python3.7/test/test_ssl.py", line 2293, in run
    msg = self.read()
  File "/Library/Frameworks/Python.framework/Versions/3.7/lib/python3.7/test/test_ssl.py", line 2270, in read
    return self.sslconn.read()
  File "/Library/Frameworks/Python.framework/Versions/3.7/lib/python3.7/ssl.py", line 920, in read
    return self._sslobj.read(len)
ssl.SSLError: [SSL: PEER_DID_NOT_RETURN_A_CERTIFICATE] peer did not return a certificate (_ssl.c:2508)

  File "/Library/Frameworks/Python.framework/Versions/3.7/lib/python3.7/threading.py", line 926, in _bootstrap_inner
    self.run()
  File "/Library/Frameworks/Python.framework/Versions/3.7/lib/python3.7/test/test_ssl.py", line 2379, in run
    raise ssl.SSLError('tlsv13 alert certificate required')
ssl.SSLError: ('tlsv13 alert certificate required',)

ok
test_pha_setter (test.test_ssl.TestPostHandshakeAuth) ... ok
test_get_server_certificate_ipv6 (test.test_ssl.NetworkedTests) ... skipped "Resource 'ipv6.google.com' is not available"
test_timeout_connect_ex (test.test_ssl.NetworkedTests) ... ok

Ran 153 tests in 76.861s

OK (skipped=10)
test_ssl passed in 1 min 16 sec
```

ned-deily:
Anything we can do to expedite this?  It's currently blocking 3.7.4 final and could shortly block 3.8.0 b2. @alex, would you be able to review this?   Thanks!

alex (approved):
This looks right to me -- but I confess this code has changed a lot since I last looked at it, so I'd love if another person had eyes on it.

Lib/test/test_ssl.py
```diff
@@ -4434,6 +4434,35 @@ def test_pha_not_tls13(self):
                 s.write(b'PHA')
                 self.assertIn(b'WRONG_SSL_VERSION', s.recv(1024))
 
+    def test_bpo37337_pha_cert_none(self):
```
orsenthil: Is `bpo37337` correct here? The issue addressed seems to be: https://bugs.python.org/issue37428 

This might be more of a question.
I thouhght, the test should have two parts.

1. Server verifying `SSL_VERIFY_POST_HANDSHAKE` flag set.
2. Client which asserts `SSL_VERIFY_POST_HANDSHAKE` not set.

I could not figure out if the test covers both these scenarios.

tiran: I fixed the BPO number in the name of the test.

The new test verifies that ``client_context.post_handshake_auth`` does not implicitly enable server cert validation on the client side. There are already several other tests that verify several combinations of PHA.

* the test case ``test_pha_required`` verifies that ``SSL_VERIFY_POST_HANDSHAKE`` on a server context works as expected. The server does request a client during handshake.
* ``SSL_VERIFY_POST_HANDSHAKE`` must not be set on a client context. This was the original bug. The ``_set_verify_mode`` helper prevents this already.

orsenthil (approved):
LGTM. Thank you!

Lib/test/test_ssl.py
orsenthil: Thanks for answering these, @tiran. LGTM.

PR #14423, opened 2019-06-27
bpo-30345: travis: use -Og cflags with --with-pydebug

methane:
https://bugs.python.org/issue30345

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -166,7 +166,8 @@ install:
 
 # Travis provides only 2 cores, so don't overdo the parallelism and waste memory.
 before_script:
-  - ./configure --with-pydebug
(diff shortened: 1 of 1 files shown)


vstinner (changes requested):
There are 3 ./configure calls in .travis.yml. Either update them all, or set -Og in global CFLAGS.

methane:
> There are 3 ./configure calls in .travis.yml. Either update them all, or set -Og in global CFLAGS.

-Og is useful only when running test_gdb.
In other places, -O3 is used automatically because there is no --with-pydebug.  It is much faster than -Og.

vstinner (approved):
Almost, sorry. I misread .travis.yml. So your change LGTM.

PR #14428, opened 2019-06-27
bpo-37369: Fix initialization of sys members when launched via an app container

zooba:

### DIFF
--- a/Include/cpython/initconfig.h
+++ b/Include/cpython/initconfig.h
@@ -373,10 +373,11 @@ typedef struct {
                                        module_search_paths_set is equal
                                        to zero. */
 
-    wchar_t *executable;    /* sys.executable */
-    wchar_t *prefix;        /* sys.prefix */
-    wchar_t *base_prefix;   /* sys.base_prefix */
-    wchar_t *exec_prefix;   /* sys.exec_prefix */
+    wchar_t *executable;        /* sys.executable */
+    wchar_t *base_executable;   /* sys.base_executable */
+    wchar_t *prefix;            /* sys.prefix */
+    wchar_t *base_prefix;       /* sys.base_prefix */
+    wchar_t *exec_prefix;       /* sys.exec_prefix */
     wchar_t *base_exec_prefix;  /* sys.base_exec_prefix */
 
     /* --- Parameter only used by Py_Main() ---------- */
--- a/Include/cpython/initconfig.h
+++ b/Include/cpython/initconfig.h
@@ -374,7 +374,7 @@ typedef struct {
                                        to zero. */
 
     wchar_t *executable;        /* sys.executable */
-    wchar_t *base_executable;   /* sys.base_executable */
+    wchar_t *base_executable;   /* sys._base_executable */
     wchar_t *prefix;            /* sys.prefix */
     wchar_t *base_prefix;       /* sys.base_prefix */
     wchar_t *exec_prefix;       /* sys.exec_prefix */
--- a/Lib/test/test_venv.py
+++ b/Lib/test/test_venv.py
@@ -25,6 +25,13 @@ try:
 except ImportError:
     ctypes = None
 
+# Platforms that set sys._base_executable can create venvs from within
+# another venv, so no need to skip tests that require venv.create().
+requireVenvCreate = unittest.skipUnless(
+    hasattr(sys, '_base_executable')
+    or sys.prefix == sys.base_prefix,
+    'cannot run venv.create from within a venv on this platform')
+
 def check_output(cmd, encoding=None):
     p = subprocess.Popen(cmd,
         stdout=subprocess.PIPE,
@@ -50,7 +57,7 @@ class BaseTest(unittest.TestCase):
             self.bindir = 'bin'
             self.lib = ('lib', 'python%d.%d' % sys.version_info[:2])
             self.include = 'include'
-        executable = sys.base_executable
+        executable = getattr(sys, '_base_executable', sys.executable)
         self.exe = os.path.split(executable)[-1]
 
     def tearDown(self):
@@ -146,6 +153,7 @@ class BasicTest(BaseTest):
             with patch('venv.subprocess.check_call', pip_cmd_checker):
                 builder.upgrade_dependencies(fake_context)
 
+    @requireVenvCreate
     def test_prefixes(self):
         """
         Test that the prefix values are as expected.
@@ -281,6 +289,7 @@ class BasicTest(BaseTest):
     # run the test, the pyvenv.cfg in the venv created in the test will
     # point to the venv being used to run the test, and we lose the link
     # to the source build - so Python can't initialise properly.
+    @requireVenvCreate
     def test_executable(self):
         """
         Test that the sys.executable value is as expected.
@@ -324,6 +333,7 @@ class BasicTest(BaseTest):
         )
         self.assertEqual(out.strip(), '0')
 
+    @requireVenvCreate
     def test_multiprocessing(self):
         """
         Test that the multiprocessing is able to spawn.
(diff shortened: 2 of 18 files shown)


Lib/test/test_venv.py
```diff
@@ -25,13 +25,6 @@
 except ImportError:
     ctypes = None
 
-# Platforms that set sys._base_executable can create venvs from within
-# another venv, so no need to skip tests that require venv.create().
-requireVenvCreate = unittest.skipUnless(
-    hasattr(sys, '_base_executable')
-    or sys.prefix == sys.base_prefix,
-    'cannot run venv.create from within a venv on this platform')
-
```
zooba: I need to do some work to set `base_executable` properly on Linux to make this work, so I may just undo it and leave nested-venv creation as unsupported.

Lib/venv/__init__.py
```diff
@@ -251,7 +251,7 @@ def setup_python(self, context):
 
             for suffix in suffixes:
                 src = os.path.join(dirname, suffix)
-                if os.path.exists(src):
+                if os.path.lexists(src):
```
zooba, line 273: Because the contents of the `WindowsApps\Python...` folder are symlinks, we need `lexists` here to see them.

PC/python_uwp.cpp
```diff
@@ -46,170 +49,158 @@ set_user_base()
...
-            return NULL;
-        }
-    } else {
-        if (wcscpy_s(buffer, len, argv0)) {
-            Py_FatalError("failed to copy argv[0]");
-            return NULL;
+static int
+set_process_name(PyConfig *config)
+{
+    std::wstring executable, home;
```
zooba: I got sick of memory management... these will always be available on platforms where this file is built.

PC/python_uwp.cpp
```diff
@@ -46,170 +49,158 @@ set_user_base()
...
-        } else if (len == bufferLen &&
-                   GetLastError() == ERROR_INSUFFICIENT_BUFFER) {
-            free(r);
-            r = NULL;
-            bufferLen *= 2;
+    if (!home.empty()) {
+        PyConfig_SetString(config, &config->home, home.c_str());
+        // XXX: Why doesn't setting prefix and exec_prefix work?
+        //PyConfig_SetString(config, &config->prefix, home.c_str());
+        //PyConfig_SetString(config, &config->exec_prefix, home.c_str());
```
zooba: @vstinner I expected to be able to just override prefix and exec_prefix, but they don't seem to flow through properly?

vstinner:
> @vstinner I expected to be able to just override prefix and exec_prefix, but they don't seem to flow through properly?

Hum, I tried the following change:

```
diff --git a/Modules/main.c b/Modules/main.c
index 853afedd7b..471dd0803c 100644
--- a/Modules/main.c
+++ b/Modules/main.c
@@ -65,6 +65,11 @@ pymain_init(const _PyArgv *args)
         return status;
     }
 
+    status = PyConfig_SetString(&config, &config.exec_prefix, L"/victor");
+    if (_PyStatus_EXCEPTION(status)) {
+        return status;
+    }
+
     /* pass NULL as the config: config is read from command line arguments,
        environment variables, configuration files */
     if (args->use_bytes_argv) {
```

It works for me:

```
$ ./python -c 'import sys; print(sys.exec_prefix)'
/victor
```

PC/python_uwp.cpp
```diff
@@ -46,170 +49,158 @@ set_user_base()
...
-        if (len == 0) {
-            free((void *)r);
-            return NULL;
-        } else if (len == bufferLen &&
-                   GetLastError() == ERROR_INSUFFICIENT_BUFFER) {
-            free(r);
-            r = NULL;
-            bufferLen *= 2;
+    if (!home.empty()) {
+        PyConfig_SetString(config, &config->home, home.c_str());
```
vstinner: It can fail, you have to check its return status:
https://docs.python.org/dev/c-api/init_config.html#c.PyConfig_SetString

PC/python_uwp.cpp
```diff
@@ -46,170 +49,158 @@ set_user_base()
...
 
-    new_argc = argc;
-    new_argv = (const wchar_t**)malloc(sizeof(wchar_t *) * (argc + 2));
-    if (new_argv == NULL) {
-        Py_FatalError("out of memory");
-        return -1;
+    PyPreConfig preconfig;
+    PyConfig config;
+
+    PyPreConfig_InitPythonConfig(&preconfig);
```
vstinner, line 166: FYI if you don't modify the default preconfig, you can simply on implicit pre-configuration:
https://docs.python.org/dev/c-api/init_config.html#c.PyConfig_SetArgv
"Preinitialize Python if needed."

Here it's fine since PyConfig_SetArgv() is called first.

(Or you can keep explicit pre-configuration, it's up to you.)

PC/python_uwp.cpp
vstinner: I don't know: it works for me (see my other comment). config_calculate_pathconfig() should leave config->exec_prefix unmodified if it's already set explicitly.

Later, pycore_init_import_warnings() calls _PyConfig_SetPathConfig() which copies config->exec_prefix into _Py_path_config.exec_prefix.

Include/cpython/initconfig.h
```diff
@@ -373,10 +373,11 @@ typedef struct {
                                        module_search_paths_set is equal
                                        to zero. */
 
-    wchar_t *executable;    /* sys.executable */
-    wchar_t *prefix;        /* sys.prefix */
-    wchar_t *base_prefix;   /* sys.base_prefix */
-    wchar_t *exec_prefix;   /* sys.exec_prefix */
+    wchar_t *executable;        /* sys.executable */
+    wchar_t *base_executable;   /* sys.base_executable */
```
vstinner: It's sys._base_executable, no?

PC/getpathp.c
```diff
@@ -537,14 +537,25 @@ get_program_full_path(const PyConfig *config,
...
+    if (!GetModuleFileNameW(NULL, program_full_path, MAXPATHLEN)) {
+        /* GetModuleFileName should never fail when passed NULL */
+        return _PyStatus_ERR("Cannot determine program path");
+    }
+
     /* The launcher may need to force the executable path to a
      * different environment, so override it here. */
     pyvenv_launcher = _wgetenv(L"__PYVENV_LAUNCHER__");
     if (pyvenv_launcher && pyvenv_launcher[0]) {
+        /* If overridden, preserve the original full path */
```
vstinner, line 549: It seems like macOS code should also be updated in config_init_program_name(), but that can be done later.

Note: I'm not sure if venv() code of Lib/site.py is redundant or not.

zooba:
I think I was expecting that updating the prefix would be enough to locate the files under the prefix, but that doesn't work - it still comes up with the wrong `sys.path` default. So it's probably best to set `home` and let it infer `prefix` and `exec_prefix`.

Thanks for your other comments!

PC/getpathp.c
zooba: Honestly, I can't figure out exactly what the macOS logic is supposed to be... I don't think it's the same as what I have for Windows, even though theoretically it is fixing the same kind of issue.

Include/cpython/initconfig.h
zooba: It is now, yes :) (I started this change by making it public, but I gave up getting it to work on Linux so I'm just making it private again)

vstinner:
> I think I was expecting that updating the prefix would be enough to locate the files under the prefix, but that doesn't work - it still comes up with the wrong sys.path default. So it's probably best to set home and let it infer prefix and exec_prefix.

Python Path Configuration is a black box. Nobody knows how it works! :-D

zooba:
I'm going to give this one more review in the morning with fresh eyes (and coffee), but it should basically match the fix that's gone in for 3.7 already. This one defines more new API rather than just using a hack though (but I still want it in for the next 3.8.0 beta).

PR #14430, opened 2019-06-27
bpo-37432: Fix a small param type in allocation.rst

shihai1991:
https://bugs.python.org/issue37432

### DIFF
--- a/Objects/dictobject.c
+++ b/Objects/dictobject.c
@@ -1427,7 +1427,7 @@ PyDict_GetItemWithError(PyObject *op, PyObject *key)
 {
     Py_ssize_t ix;
     Py_hash_t hash;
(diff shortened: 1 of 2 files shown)


Objects/dictobject.c
```diff
@@ -1427,7 +1427,7 @@ PyDict_GetItemWithError(PyObject *op, PyObject *key)
 {
     Py_ssize_t ix;
     Py_hash_t hash;
-    PyDictObject*mp = (PyDictObject *)op;
```
methane: revert unrelating change

shihai1991: ok, this is a code style problem.
Hi, Inada. I have a question. Looks like pure code style can not accepted. How can I improve the code style problem or just ignored(such as: https://bugs.python.org/issue37375 can not accepted.)

PR #14431, opened 2019-06-27
bpo-36511: clean up python process before deploy

paulmon:
If the previous buildbot test times out, there will be orphaned python processes that prevent deploy from working.  Kill python.exe processes before trying to clean remote test directory

@zooba @zware 


zware (approved):
LGTM, provided that the `kill` command exists :).  I may be operating from out of date information, but I thought Windows only offered `taskkill /IM python.exe`?

zooba:
Agreed with Zachary - "kill" isn't a Windows cmd.exe command (it is an alias in PowerShell), and I'm pretty sure you've only got cmd.exe commands at the other end of the SSH connection (since `dir /b` won't work in PowerShell).

Unless it's a new/extra addition on the buildbot?

paulmon:
kill.exe is a command on Windows IoT Core which is what the buildbot is running.
There is no taskkill.exe on Windows IoT Core.

zware:
Ok, we'll give this a try then.

PR #14433, opened 2019-06-27
[bpo-37433] Fix `SyntaxError` indicator printing too many spaces for multi-line strings

asottile:
https://bugs.python.org/issue37433

### DIFF
--- a/Lib/test/test_cmd_line_script.py
+++ b/Lib/test/test_cmd_line_script.py
@@ -613,6 +613,20 @@ class CmdLineTest(unittest.TestCase):
             self.assertNotIn("\f", text)
             self.assertIn("\n    1 + 1 = 2\n    ^", text)
 
+    def test_syntaxerror_multi_line_fstring(self):
+        script = 'foo = f"""{}\nfoo"""\n'
+        with support.temp_dir() as script_dir:
+            script_name = _make_test_script(script_dir, 'script', script)
+            exitcode, stdout, stderr = assert_python_failure(script_name)
+            self.assertEqual(
+                stderr.splitlines()[-3:],
+                [
+                    b'    foo = f"""{}',
+                    b'          ^',
+                    b'SyntaxError: f-string: empty expression not allowed',
+                ],
+            )
+
     def test_consistent_sys_path_for_direct_execution(self):
         # This test case ensures that the following all give the same
         # sys.path configuration:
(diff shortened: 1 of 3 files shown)


Lib/test/test_cmd_line_script.py
```diff
@@ -613,6 +613,20 @@ def test_syntaxerror_indented_caret_position(self):
             self.assertNotIn("\f", text)
             self.assertIn("\n    1 + 1 = 2\n    ^", text)
 
+    def test_syntaxerror_multi_line_fstring(self):
```
terryjreedy, line 616: Perhaps
```suggestion
    def test_syntaxerror_multi_line_fstring(self):  # bpo-37433
```

terryjreedy (commented):
Using the example in the new test, I verified the problem and fix on Win 10.

I am not familiar enough with C and CPython C code to be sure that fix is *exactly* what is needed.

Lib/test/test_cmd_line_script.py
asottile: in previous patches I was told to remove the `bpo` comment as it's essentially noise (and `git blame` can retrieve the same information) -- let me know if you'd like to follow that suggestion or this one 👍

asottile:
> 
> 
> Using the example in the new test, I verified the problem and fix on Win 10.
> 
> I am not familiar enough with C and CPython C code to be sure that fix is _exactly_ what is needed.

to explain a little better in words what's happening in this block:

- initially there is a buffer which is allocated, `multi_line_start` and `line_start` point at the beginning of that lines
- at some point, the tokenizer realises it needs to allocate more and read more to fill out the buffer for a multi line token
- this causes a reallocation of the original buffer into a new buffer, all of the pointers are copied over and adjusted for the new positions and offset -- however the `mutli_line_start` was not being copied and adjusted -- this patch fixes that bit

Lib/test/test_cmd_line_script.py
terryjreedy: Leave it alone for now.

vstinner (approved):
LGTM. But I would prefer that at least another core dev reviews the change.

pablogsal (approved):
LGTM

I checked the examples provided and `tok->multi_line_start` does not point anymore to random memory (the one that is there before the `PyMem_REALLOC` happens).

I would suggest adding more tests for this that do not involve f-strings, maybe in a different PR.

pablogsal:
@asottile  Does this affect 3.7?

asottile:
@pablogsal nope, only 3.8+

pablogsal:
Thank you very much @asottile for the fix :)

PR #14434, opened 2019-06-27
bpo-37412: pythoninfo: add Windows long paths

vstinner:
On Windows, test.pythoninfo now checks if support for long paths is
enabled using ntdll.RtlAreLongPathsEnabled() function.

Co-Authored-By: Eryk Sun <eryksun@gmail.com>

### DIFF
--- a/Lib/test/pythoninfo.py
+++ b/Lib/test/pythoninfo.py
@@ -651,6 +651,29 @@ def collect_subprocess(info_add):
     copy_attributes(info_add, subprocess, 'subprocess.%s', ('_USE_POSIX_SPAWN',))
 
 
+def collect_windows(info_add):
+    try:
+        import ctypes
+    except ImportError:
+        return
+
+    if not hasattr(ctypes, 'WinDLL'):
+        return
+
+    ntdll = ctypes.WinDLL('ntdll')
+    BOOLEAN = ctypes.c_ubyte
+
+    try:
+        RtlAreLongPathsEnabled = ntdll.RtlAreLongPathsEnabled
+    except AttributeError:
+        res = '<function not available>'
+    else:
+        RtlAreLongPathsEnabled.restype = BOOLEAN
+        RtlAreLongPathsEnabled.argtypes = ()
+        res = bool(RtlAreLongPathsEnabled())
+    info_add('windows.RtlAreLongPathsEnabled', res)
+
+
 def collect_info(info):
     error = False
     info_add = info.add
@@ -684,6 +707,7 @@ def collect_info(info):
         collect_testcapi,
         collect_time,
         collect_tkinter,
+        collect_windows,
         collect_zlib,
 
         # Collecting from tests should be last as they have side effects.


vstinner:
The code comes from @eryksun, so I added him as a co-author:
https://github.com/python/cpython/pull/14424#issuecomment-506517744

vstinner:
@eryksun: Would you mind to review this change?

I tested it manually on my Windows VM:

```
windows.long_paths_enabled: True
```

Maybe it was the Python installer who proposed me to enable this Windows feature, I don't recall.

vstinner:
* AppVeyor: windows.long_paths_enabled: **False**
* Azure Pipelines, win32: windows.long_paths_enabled: **True**
* Azure Pipelines, win64: windows.long_paths_enabled: **True**

Lib/test/pythoninfo.py
```diff
@@ -651,6 +651,27 @@ def collect_subprocess(info_add):
...
+        return
+
+    try:
+        import ctypes
+    except ImportError:
+        return
+
+    ntdll = ctypes.WinDLL('ntdll')
+    if not hasattr(ntdll, 'RtlAreLongPathsEnabled'):
+        return
```
eryksun: In this case set `info_add('windows.long_paths_enabled', False)`.

vstinner:
Oh, there is also a RtlIsLongPathAwareProcessByManifest() function which may be interesting to check:
https://stackoverflow.com/questions/40983329/are-long-path-behavior-per-app-can-be-enable-via-the-manifest

I updated my PR to also call this function. Result on my Windows 10 VM:

```
sys.windowsversion: sys.getwindowsversion(major=10, minor=0, build=17763, platform=2, service_pack='')
windows.RtlAreLongPathsEnabled: True
windows.RtlIsLongPathAwareProcessByManifest: <function not available>
```

Oh, RtlIsLongPathAwareProcessByManifest is not function not available in ntdll on my VM? :-(

vstinner:
Results on AppVeyor:

```
platform.platform: Windows-10-10.0.14393-SP0
sys.windowsversion: sys.getwindowsversion(major=10, minor=0, build=14393, platform=2, service_pack='')
windows.RtlAreLongPathsEnabled: False
windows.RtlIsLongPathAwareProcessByManifest: True
```

Results on Azure Pipeline, win32 and win64 (same values):

```
sys.windowsversion: sys.getwindowsversion(major=10, minor=0, build=14393, platform=2, service_pack='')
windows.RtlAreLongPathsEnabled: True
windows.RtlIsLongPathAwareProcessByManifest: True
```

vstinner:
RtlIsLongPathAwareProcessByManifest may be interesting when I will backport this change to old Python versions. I'm trying to keep the pythoninfo.py code the same in all branches.

eryksun:
> RtlIsLongPathAwareProcessByManifest may be interesting when I will backport this change to old Python versions. I'm trying to keep the pythoninfo.py code the same in all branches.

I wouldn't worry about `RtlIsLongPathAwareProcessByManifest`. It only exists in older versions of Windows 10, and IIRC it's called only once at process startup. I'm pretty sure that `RtlIsLongPathAwareProcessByManifest` was removed because the scope of the long-path policy is the Windows base API, not the NT API. Thus the initial configuration should be, and now is, implemented completely in kernelbase.dll, using private functions.

`KernelbasePostInit` evaluates whether long paths are enabled, and if so it sets the `IsLongPathAwareProcess` flag in the Process Environment Block (PEB). This flag value is what `RtlAreLongPathsEnabled` checks. 

`KernelbasePostInit` first calls `BasepIsProcessLongPathAwareByPolicy` to check the "LongPathsEnabled" registry value. If the policy is enabled, it calls `BasepIsProcessLongPathAwareByGlobalOverride` to check the "LPGO" registry value, which is a relatively new feature that globally overrides the application manifest. (I would recommend against setting this value.) If the global override isn't set, it calls `BasepIsProcessLongPathAwareByManifest`, which calls `RtlQueryActivationContextApplicationSettings` to check whether "longPathAware" is enabled in the application manifest. If so, it sets the `IsLongPathAwareProcess` flag in the PEB.

vstinner:
> I wouldn't worry about RtlIsLongPathAwareProcessByManifest

Ok ok. Anyway, this info can be retrieved indirectly from the Python version. I rebased my PR and squashed commits.

@eryksun: Would you mind to review the updated PR?

vstinner:
I merged my PR. Thanks @eryksun for reviews!

PR #14443, opened 2019-06-28
 bpo-37335: Fix unexpected ASCII aliases in locale coercion tests

kulikjak:
Solaris uses '646' as an alias for ASCII encoding and thus all the locale coercion tests are failing. This patch fixes the problem by using codecs lookup to normalize encoding names and thus prevent these problems.

### DIFF
--- a/Lib/test/test_c_locale_coercion.py
+++ b/Lib/test/test_c_locale_coercion.py
@@ -100,11 +100,11 @@ _EncodingDetails = namedtuple("EncodingDetails", _fields)
 class EncodingDetails(_EncodingDetails):
     # XXX (ncoghlan): Using JSON for child state reporting may be less fragile
     CHILD_PROCESS_SCRIPT = ";".join([
-        "import sys, os",
-        "print(sys.getfilesystemencoding())",
-        "print(sys.stdin.encoding + ':' + sys.stdin.errors)",
-        "print(sys.stdout.encoding + ':' + sys.stdout.errors)",
-        "print(sys.stderr.encoding + ':' + sys.stderr.errors)",
+        "import sys, os, codecs",
+        "print(codecs.lookup(sys.getfilesystemencoding()).name)",
+        "print(codecs.lookup(sys.stdin.encoding).name + ':' + sys.stdin.errors)",
+        "print(codecs.lookup(sys.stdout.encoding).name + ':' + sys.stdout.errors)",
+        "print(codecs.lookup(sys.stderr.encoding).name + ':' + sys.stderr.errors)",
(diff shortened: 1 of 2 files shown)


mangrisano (approved):
LGTM.

Lib/test/test_c_locale_coercion.py
```diff
@@ -100,11 +100,11 @@ def _set_locale_in_subprocess(locale_name):
...
-        "import sys, os",
-        "print(sys.getfilesystemencoding())",
-        "print(sys.stdin.encoding + ':' + sys.stdin.errors)",
-        "print(sys.stdout.encoding + ':' + sys.stdout.errors)",
-        "print(sys.stderr.encoding + ':' + sys.stderr.errors)",
+        "import sys, os, codecs",
+        "print(codecs.lookup(sys.getfilesystemencoding()).name)",
+        "print(codecs.lookup(sys.stdin.encoding).name + ':' + sys.stdin.errors)",
+        "print(codecs.lookup(sys.stdout.encoding).name + ':' + sys.stdout.errors)",
+        "print(codecs.lookup(sys.stderr.encoding).name + ':' + sys.stderr.errors)",
```
vstinner, line 107: Oh wait, I added _PyUnicode_InitEncodings() in Python 3.8: filesystem and stdio encodings are now always normalized. https://bugs.python.org/issue36775 :

```
commit 43fc3bb7cf0278735eb0010d7b3043775a120cb5
Date:   Thu May 2 11:54:20 2019 -0400

    bpo-36775: Add _PyUnicode_InitEncodings() (GH-13057)
    
    Move get_codec_name() and initfsencoding() from pylifecycle.c to
    unicodeobject.c.
    
    Rename also "init" functions in pylifecycle.c.
```

So _handle_output_variations() can simply be removed from 3.8, without having to add codecs.lookup().name here.

This change makes sense to Python 3.7. Would you mind to first create a new PR for master, backport it to 3.8, and then retarget this PR to 3.7?

vstinner (changes requested):

Lib/test/test_c_locale_coercion.py
kulikjak: Yes, that will be the reason why Python 3.8 no longer has this problem. I will look into it.

Is there a way how to retarget this to 3.7? When we tried it previously, Github tried to backport the whole master into 3.7... Or do I have to create new PR for that as well?

vstinner: Well, maybe creating a new PR for 3.7 is simpler :-)

kulikjak:
I opened new other PRs for this change - closing this one.

PR #14444, opened 2019-06-28
bpo-36763: Add PyConfig_SetWideStringList()

vstinner:

### DIFF
--- a/Doc/c-api/init_config.rst
+++ b/Doc/c-api/init_config.rst
@@ -25,6 +25,7 @@ Functions:
 * :c:func:`PyConfig_SetBytesArgv`
 * :c:func:`PyConfig_SetBytesString`
 * :c:func:`PyConfig_SetString`
+* :c:func:`PyConfig_SetWideStringList`
 * :c:func:`PyPreConfig_InitIsolatedConfig`
 * :c:func:`PyPreConfig_InitPythonConfig`
 * :c:func:`PyStatus_Error`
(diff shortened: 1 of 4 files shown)


vstinner:
@zooba, @methane, @Yhg1s: What do you think of adding PyConfig_SetWideStringList to https://www.python.org/dev/peps/pep-0587/ ? IMHO this function is missing to make the API "complete". Even if technically, PyWideStringList_Append() should be enough for most usage :-)

While I wrote PR #14445 to cleanup old code of the implementation of the PEP 587, I failed to convert test_init_from_config() in Programs/_testembed.c to dynamically allocated memory. PyConfig API doesn't provide any function to set a wide string list, whereas the test sets warnoptions and xoptions.

While PyWideStringList_Append() could be used, I prefer to be able to provide a list created in pure C without PyConfig API "at once".

First, I wrote:

```
PyStatus PyConfig_SetWideStringList(PyConfig *config, PyWideStringList *list, const PyWideStringList *list2)
```

But there is no way to build a wide string list: I didn't add any API for that on purpose, to minimize the size of the API. There is no function to create an empty PyWideStringList, no function to clear a list, no function to copy a list, etc.

Maybe tomorrow we may want to add such functions, but it can be done later. I prefer to start with a smaller API.

vstinner:
> While PyWideStringList_Append() could be used, I prefer to be able to provide a list created in pure C without PyConfig API "at once".

PyConfig_SetWideStringList() is different because first it clears the list, so options set previously are ignored. And I would like this behavior: replace existing list with my own list :-)

vstinner:
Hum, I decided to merge this PR soon to try to get PR #14445 fix into Python 3.8 beta2, to have more time to test it before 3.8.0 final. We still have time to revisit the PyConfig_SetWideStringList API before 3.8.0 final if sometimes dislikes it :-)

PR #14445, opened 2019-06-28
bpo-36763: Use PyConfig_Clear()

vstinner:
Stop using "static PyConfig", PyConfig must now always use
dynamically allocated strings: use PyConfig_SetString(),
PyConfig_SetArgv() and PyConfig_Clear().

### DIFF
--- a/Lib/test/test_embed.py
+++ b/Lib/test/test_embed.py
@@ -695,10 +695,19 @@ class InitConfigTests(EmbeddingTestsMixin, unittest.TestCase):
 
             'pycache_prefix': 'conf_pycache_prefix',
             'program_name': './conf_program_name',
-            'argv': ['-c', 'arg2'],
+            'argv': ['-c', 'arg2', ],
             'parse_argv': 1,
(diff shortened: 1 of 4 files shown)


vstinner:
This PR depends on PR #14444 to add PyConfig_SetWideStringList().

brettcannon:
The title says this is about `PyConfig_Clear()` but most of the changes in the PR revolve around `PyConfig_SetWideStringList()`. Is the title accurate?

vstinner:
> The title says this is about PyConfig_Clear() but most of the changes in the PR revolve around PyConfig_SetWideStringList(). Is the title accurate?

To be able to use PyConfig_Clear(), as "required" by the PEP 587, you have to use dynamically allocated memory, and so the code has to be modified to use PyConfig_SetWideStringList() instead of static strings.

This change depends on PR #14444: we need a PyConfig_SetWideStringList() function, otherwise, it's harder to use set warnoptions and xoptions fields of PyConfig.

vstinner:
This change should be backported to 3.8 once PR #14523 is merged.

PR #14446, opened 2019-06-28
bpo-20443: Fix calculate_program_full_path() warning

vstinner:
Don't call _Py_isabs() with a bytes string (char*), the function
expects as wide string.


vstinner:
cc @tirkarthi

vstinner:
I don't see any compiler warning on getpath.c compilation in the macOS job of Travis CI, whereas I saw it in a previous job.

tirkarthi:
Thanks Victor, I can confirm that this fixes the compiler warning.

PR #14447, opened 2019-06-28
bpo-37335: Remove unnecessary code from c locale coercion tests

kulikjak:
Python 3.8+ encodings are always normalized and thus no output variations handling is necessary (the code is no longer necessary).

### DIFF
--- a/Lib/test/test_c_locale_coercion.py
+++ b/Lib/test/test_c_locale_coercion.py
@@ -128,19 +128,6 @@ class EncodingDetails(_EncodingDetails):
         env_info = expected_lang, expected_lc_ctype, expected_lc_all
         return dict(cls(fs_encoding, *stream_info, *env_info)._asdict())
 
-    @staticmethod
-    def _handle_output_variations(data):
-        """Adjust the output to handle platform specific idiosyncrasies
-
-        * Some platforms report ASCII as ANSI_X3.4-1968
-        * Some platforms report ASCII as US-ASCII
-        * Some platforms report UTF-8 instead of utf-8
-        """
-        data = data.replace(b"ANSI_X3.4-1968", b"ascii")
-        data = data.replace(b"US-ASCII", b"ascii")
-        data = data.lower()
-        return data
-
     @classmethod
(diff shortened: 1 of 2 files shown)


Lib/test/test_c_locale_coercion.py
```diff
@@ -160,7 +147,7 @@ def get_child_details(cls, env_vars):
         if not result.rc == 0:
             result.fail(py_cmd)
         # All subprocess outputs in this test case should be pure ASCII
-        adjusted_output = cls._handle_output_variations(result.out)
+        adjusted_output = result.out.lower()
```
vstinner: What is the purpose of .lower()?

kulikjak: It was in the removed function (under those replaces) because according to the comment: "_Some platforms report UTF-8 instead of utf-8_". But maybe that is also fixed by guaranteed normalization?

Also when I remove it, tests start failing, because 'os.environ.get()' functions return C and the suite expects c (as everything was lowered before), but I can fix that as well.

vstinner: > It was in the removed function (under those replaces) because according to the comment: "Some platforms report UTF-8 instead of utf-8". But maybe that is also fixed by guaranteed normalization?

That's the purpose of normalization: always get the same string :-)

```
>>> codecs.lookup("UTF8").name
'utf-8'
>>> codecs.lookup("UTF-8").name
'utf-8'
>>> codecs.lookup("UTF_8").name
'utf-8'
>>> codecs.lookup("utf8").name
'utf-8'
```

> Also when I remove it, tests start failing, because 'os.environ.get()' functions return C and the suite expects c (as everything was lowered before), but I can fix that as well.

Please use "C" instead of "c" :-)

kulikjak: > That's the purpose of normalization: always get the same string :-)

Well, I could have thought of that...

> Please use "C" instead of "c" :-)

Fixed

vstinner:
Thanks, I merged your PR. I enhanced the commit message to better elaborate the change.

PR #14448, opened 2019-06-28
bpo-37440: Enable TLS 1.3 post-handshake auth in http.client

tiran:
Post-handshake authentication is required for conditional client cert authentication with TLS 1.3.

https://bugs.python.org/issue37440

### DIFF
--- a/Doc/library/http.client.rst
+++ b/Doc/library/http.client.rst
@@ -94,6 +94,11 @@ The module provides the following classes:
       :func:`ssl._create_unverified_context` can be passed to the *context*
       parameter.
 
+   .. versionchanged:: 3.8
+      This class now enables TLS 1.3
+      :attr:`ssl.SSLContext.post_handshake_auth` for the default *context* or
+      when *cert_file* is passed with a custom *context*.
+
    .. deprecated:: 3.6
 
        *key_file* and *cert_file* are deprecated in favor of *context*.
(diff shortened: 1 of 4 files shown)


ned-deily:
Anything we can do to expedite this?  It's currently blocking 3.7.4 final and could shortly block 3.8.0 b2. @alex, would you be able to review this?   Thanks!

alex:
I don't think I understand this well enough to really review it -- why don't we always set `post_handshake_auth` when using TLS 1.3 and client certs?

tiran:
@alex OpenSSL disables PHA by default because clients must be able to handle it. An application protocol must deal with fact that there is an additional TCP roundtrip involved.

orsenthil:
Hi @tiran  - I reviewed after understanding the relevant context here:
* https://bugs.python.org/issue34670 
and 
* https://github.com/python/cpython/commit/9fb051f032c36b9f6086b79086b4d6b7755a3d70#diff-452aa9cf86b07b29d66846b6d0a86687 

Especially this part for `SSLContext.post_handshake_auth`

```
>    .. note::
      Only available with OpenSSL 1.1.1 and TLS 1.3 enabled. Without TLS 1.3
      support, the property value is None and can't be modified
```

* The changes look good to me, and tests are helpful to understand the change. 
* The review discussion with Alex was helpful to understand why the value is not to True by default.

I am giving my approval, but if you desire additional review+approval, please do seek out to other core-devs.

Thanks!

orsenthil (approved):

PR #14449, opened 2019-06-28
[3.7] bpo-37335: Fix unexpected ASCII aliases in locale coercion tests

kulikjak:
Solaris uses '646' as an alias for ASCII encoding and thus all the locale coercion tests are failing. This patch fixes the problem by using codecs lookup to normalize encoding names and thus prevent these problems.

### DIFF
--- a/Lib/test/test_c_locale_coercion.py
+++ b/Lib/test/test_c_locale_coercion.py
@@ -119,12 +119,12 @@ class EncodingDetails(_EncodingDetails):
         stream_info = 2*[_stream.format("surrogateescape")]
         # stderr should always use backslashreplace
         stream_info.append(_stream.format("backslashreplace"))
-        expected_lang = env_vars.get("LANG", "not set").lower()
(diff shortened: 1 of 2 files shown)


Lib/test/test_c_locale_coercion.py
```diff
@@ -160,7 +147,7 @@ def get_child_details(cls, env_vars):
         if not result.rc == 0:
             result.fail(py_cmd)
         # All subprocess outputs in this test case should be pure ASCII
-        adjusted_output = cls._handle_output_variations(result.out)
+        adjusted_output = result.out.lower()
```
vstinner: What's the purpose of .lower()? Does the test still pass without it?

kulikjak: Tests did not pass, because of C vs c returned by `os.environ.get()`. It is fixed now.

vstinner:
Thanks, I merged your PR. I modified the commit message to better describe it.

PR #14450, opened 2019-06-28
[3.7] bpo-37369: Fixes path for sys.executable when running from the Microsoft Store

zooba:

### DIFF
--- a/Lib/site.py
+++ b/Lib/site.py
@@ -458,13 +458,12 @@ def venv(known_paths):
     env = os.environ
     if sys.platform == 'darwin' and '__PYVENV_LAUNCHER__' in env:
         executable = sys._base_executable = os.environ['__PYVENV_LAUNCHER__']
-    elif sys.platform == 'win32' and '__PYVENV_LAUNCHER__' in env:
+    elif sys.platform == 'win32':
         executable = sys.executable
-        import _winapi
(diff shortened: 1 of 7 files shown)


zooba:
@ned-deily This is the fix I'm considering for today. Still need to run some testing on it, but wanted to give you a look since I figure you'll be paying attention right now :)

Lib/site.py
```diff
@@ -458,13 +458,12 @@ def venv(known_paths):
...
         executable = sys.executable
-        import _winapi
-        sys._base_executable = _winapi.GetModuleFileName(0)
-        # bpo-35873: Clear the environment variable to avoid it being
-        # inherited by child processes.
-        del os.environ['__PYVENV_LAUNCHER__']
+        if '__PYVENV_LAUNCHER__' in env:
+            # bpo-35873: Clear the environment variable to avoid it being
+            # inherited by child processes.
+            del os.environ['__PYVENV_LAUNCHER__']
```
zooba: Going to move this bit into `getpathp.c`, which is probably where it should have been the whole time.

zooba:
@ned-deily I need to step away for an hour, but if this passes CI then I'm happy with it.

zooba:
Okay, _this_ time it should be good

PR #14455, opened 2019-06-28
bpo-37438: Added default argument for set in configparser doc string

Kriyszig:
Issue: [37438](https://bugs.python.org/issue37438)
Code: [Definition of `set` in configparser](https://github.com/python/cpython/blob/db4d7ddb012ef8f087a8eb2a5b8a672d04a48e1a/Lib/configparser.py#L889)

In the code mentioned above default value of argument `value` is None. This will use `dict_type` as a default value.

https://bugs.python.org/issue37438

### DIFF
--- a/Lib/configparser.py
+++ b/Lib/configparser.py
@@ -129,7 +129,7 @@ ConfigParser -- responsible for parsing a list of
     remove_option(section, option)
         Remove the given option from the given section.
 
-    set(section, option, value)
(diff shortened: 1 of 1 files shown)


tirkarthi:
The issue also noted about updating signature in docs at https://docs.python.org/3/library/configparser.html#configparser.ConfigParser.set and adding a note on the behavior when allow_no_value is set.

PR #14456, opened 2019-06-28
[3.7] bpo-37369: Fix venv and test symlinking

zooba:

### DIFF
--- a/Lib/test/test_platform.py
+++ b/Lib/test/test_platform.py
@@ -16,14 +16,24 @@ class PlatformTest(unittest.TestCase):
 
     @support.skip_unless_symlink
     def test_architecture_via_symlink(self): # issue3762
+        if sys.platform == "win32" and not os.path.exists(sys.executable):
+            # App symlink appears to not exist, but we want the
+            # real executable here anyway
+            import _winapi
+            real = _winapi.GetModuleFileName(0)
+        else:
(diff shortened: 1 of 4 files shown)


zooba:
@ned-deily As I worried, there are flow on effects. Because `sys.executable` is no longer a real file _or_ a real symlink, some operations break down.

There has to be a trade-off between "can I use `sys.executable` later to execute Python" vs "can I `open(sys.executable)` or make a link to sys.executable", and I think the former should win. Hopefully this is narrow enough that it only impacts Store users, and we can give it as another reason you may need to switch to the regular installer.

ned-deily:
FWIW, I agree that the main purpose of `sys.executable` should be to facilitate executing another instance of the current Python so that sounds like the right tradeoff.  Also, FWIW, in macOS framework builds, `sys.executable` points to the so-called stub launcher executable, not to the actual interpreter binary, and the path is often not within `sys.prefix`.  (That's true for Python 3; in Python 2, that behavior was never changed and still causes problems here and there with tests and the like.)  So there is some precedent for divergence here.

zooba:
I'm watching https://dev.azure.com/Python/cpython/_build/results?buildId=46366&view=results before declaring this totally done - the CI build has additional tests in it that will cover these scenarios.

PR #14457, opened 2019-06-28
bpo-37404: Raising value error if an SSLSocket is passed to asyncio functions

blackblood:
https://bugs.python.org/issue37404

### DIFF
--- a/Lib/asyncio/selector_events.py
+++ b/Lib/asyncio/selector_events.py
@@ -341,7 +341,7 @@ class BaseSelectorEventLoop(base_events.BaseEventLoop):
         self._ensure_fd_no_transport(fd)
         return self._remove_writer(fd)
 
-    async def sock_recv(self, sock, n, _warn=warnings.warn):
+    async def sock_recv(self, sock, n):
         """Receive data from the socket.
 
         The return value is a bytes object representing the data received.
@@ -349,7 +349,7 @@ class BaseSelectorEventLoop(base_events.BaseEventLoop):
         nbytes.
         """
         if isinstance(sock, ssl.SSLSocket):
-            _warn("the socket must not be an SSLSocket instance", DeprecationWarning, source=self)
+            warnings.warn("the socket must not be an SSLSocket instance", DeprecationWarning, source=self)
         if self._debug and sock.gettimeout() != 0:
             raise ValueError("the socket must be non-blocking")
         try:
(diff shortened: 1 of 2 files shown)


the-knights-who-say-ni:

tirkarthi (commented):
This introduces a change in behavior which would need a test for respective methods where to assert `ValueError` to be raised. The below program with this PR would raise `ValueError` . It's already documented at https://docs.python.org/3/library/asyncio-eventloop.html#asyncio.loop.sock_recv that sock should be a non-blocking socket and perhaps need a note about `ValueError`.

```python
import socket
import ssl
import asyncio

async def main():
    with socket.socket() as sock:
        with ssl.wrap_socket(sock) as ssl_socket:
            loop = asyncio.get_event_loop()
            await loop.sock_recv(ssl_socket, 10)

asyncio.run(main())
```

```
./python /tmp/bpo37404.py
  File "/tmp/bpo37404.py", line 11, in <module>
    asyncio.run(main())
  File "/home/karthi/cpython/Lib/asyncio/runners.py", line 43, in run
    return loop.run_until_complete(main)
  File "/home/karthi/cpython/Lib/asyncio/base_events.py", line 608, in run_until_complete
    return future.result()
  File "/tmp/bpo37404.py", line 9, in main
    await loop.sock_recv(ssl_socket, 10)
  File "/home/karthi/cpython/Lib/asyncio/selector_events.py", line 352, in sock_recv
    raise ValueError("the socket must not be an SSLSocket instance")
ValueError: the socket must not be an SSLSocket instance
```

asvetlov (changes requested):
I believe we need at least one or two Python release with *deprecation warning* before raising a error.
Otherwise, we break existing code without any notice.

blackblood:
Aren't deprecation warnings for features that are working right now but won't work in the future? wouldn't sending a deprecation warning give the user a false impression that the sock_recv() function currently works with SSLSocket.

blackblood:

blackblood:

Lib/asyncio/selector_events.py
```diff
@@ -341,13 +341,15 @@ def remove_writer(self, fd):
         self._ensure_fd_no_transport(fd)
         return self._remove_writer(fd)
 
-    async def sock_recv(self, sock, n):
+    async def sock_recv(self, sock, n, _warn=warnings.warn):
```
ZackerySpytz: Why are you adding new `_warn` parameters to methods in this file? That's not necessary.

blackblood:

Lib/asyncio/selector_events.py
blackblood: Fixed

blackblood:

brettcannon:
Thanks for the PR, but closing as the CLA has not been signed within the last month. If you do decide to sign the CLA we can re-open this PR.

PR #14460, opened 2019-06-28
bpo-37552 strptime/strftime return invalid results with UCRT version 17763.615

paulmon:
Skip tests that fail because of bug in MSVC UCRT failure when the default code page is cp65001.

@zooba 


zooba:
These issues ought to be fixed in the next Windows update (end of year, about the time that 3.8 releases), though we may get a redistributable ucrtbase.dll sooner.

Lib/test/test_strptime.py
```diff
@@ -135,6 +135,8 @@ def test_pattern_escaping(self):
                       "%s does not have re characters escaped properly" %
                       pattern_string)
 
+    @unittest.skipIf(sys.platform == 'win32' and locale.getdefaultlocale()[1] == 'cp65001',
```
pganssle: Can you add a comment referencing bpo-36511 here, so people in the future will know when it is safe to stop skipping this?

It might even be worth opening a separate BPO issue for tracking this specific thing, because issue 36511 seems like more of an omnibus issue.

@zooba You probably know more about the lifetime of Windows support here - what is the best way to communicate that it will be safe to remove this `skipIf`?

pganssle (dismissed):

Lib/test/test_strptime.py
zooba: Agreed, a specific issue for this particular failure would be good. There was one already for the same thing (to do with passing `de_DE` instead of `de-DE` IIRC, which triggered the experimental/broken UTF-8 support in the UCRT) that might be okay to reuse.

I don't know how best to communicate a future time for this, especially since the UCRT may get updated down-level (in earlier releases) and certain fixes may release out of band. Probably best to just leave a comment with the current UCRT version (@paulmon?) and a reminder to verify as new updates are made.

paulmon: @zooba, Is this the same issue? https://bugs.python.org/issue36792
These tests aren't crashing as indicated in bpo-36792, but they do cause the Windows ARM32 buildbot to fail

paulmon: I created a new issue: https://bugs.python.org/issue37552 and added comments.

paulmon:

zooba:
I believe this has been sufficiently handled, but I'll let @pganssle retract his review when he's ready. (And I'll poke him in person today to make sure he's ready ;) )

Lib/test/test_strptime.py
```diff
@@ -135,6 +135,9 @@ def test_pattern_escaping(self):
                       "%s does not have re characters escaped properly" %
                       pattern_string)
 
+    # bpo-37552 [Windows] strptime/strftime return invalid results with UCRT version 17763.615
+    @unittest.skipIf(sys.platform == 'win32' and locale.getdefaultlocale()[1] == 'cp65001',
```
zware: Rather than repeating the whole condition and comment several times, I'd rather see it factored out to a simple `@skip_if_buggy_ucrt` decorator.

Separately, I would suggest that that decorator should actually check for one of the known bugs in the UCRT, and only skip if that bug is still present; we *will* forget to remove these skips later otherwise :)

paulmon: Those are both great ideas :)

paulmon:
If the Azure Pipelines PR check passes, I think this is ready for additional feedback and/or merging.

Lib/test/support/__init__.py
```diff
@@ -2501,6 +2502,27 @@ def skip_unless_symlink(test):
...
+    """
+    Skip decorator for tests that use buggy strptime/strftime
+
+    If the UCRT bugs are present time.localtime().tm_zone will be
+    an empty string, otherwise we assume the UCRT bugs are fixed
+
+    See bpo-37552 [Windows] strptime/strftime return invalid
+    results with UCRT version 17763.615
+    """
+    global _buggy_ucrt
```
pganssle, line 2516: This is not a particularly expensive check, right? Why cache it instead of checking each time?

zware: The caching is simple and easy to follow; I'd rather keep it than assume it's not worthwhile :)

Lib/test/support/__init__.py
```diff
@@ -2501,6 +2502,27 @@ def skip_unless_symlink(test):
     msg = "Requires functional symlink implementation"
     return test if ok else unittest.skip(msg)(test)
 
+_buggy_ucrt = None
+def skip_if_buggy_ucrt(test):
```
zware: Since this now lives in `test.support`, it might be worthwhile to mention 'time' in the name (`skip_if_buggy_ucrt_strfptime` or similar).

zware (approved):
LGTM.

Lib/test/support/__init__.py
paulmon: I added strftime/strptime to the skip message and skip decorator name

pganssle (approved):

pganssle:
> @pganssle: Please replace `#` with `GH-` in the commit message next time. Thanks!

Noooo, I forgot I left that in the PR title. Serves me right for not using automerge.

pganssle:
Thanks for the PR and for promptly addressing comments, @paulmon!

paulmon:
Thanks! all of the buildbot tests for Windows ARM32 have passed passed with only warnings since this was merged.

PR #14461, opened 2019-06-28
[3.7] bpo-37369: Fix path handling when python.exe is used as a symlink

zooba:


.azure-pipelines/pr.yml
```diff
@@ -151,3 +151,7 @@ jobs:
   - template: ./windows-steps.yml
     parameters:
       targetBranch: $(System.PullRequest.TargetBranch)
+  - template: ./windows-layout-steps.yml
+    parameters:
+      kind: appx
+      fulltest: true
```
zooba: This part is just to run the full tests before I merge, to save more post-merge fixups. I'll remove it

PR #14462, opened 2019-06-29
bpo-37445: Include FORMAT_MESSAGE_IGNORE_INSERTS in FormatMessageW() calls

ZackerySpytz:
If FormatMessageW() is passed the FORMAT_MESSAGE_FROM_SYSTEM flag
without FORMAT_MESSAGE_IGNORE_INSERTS, it will fail if there are
insert sequences in the message definition.

https://bugs.python.org/issue37445


ZackerySpytz:
A test similar to `test_winerror()` in `ctypes/test/test_win32.py` could be added for this (to test `ctypes.FormatError()`), though it would be flimsy to test the content of the error message.

encukou:
@mangrisano Not sure what I can do here; I'm not familiar with Windows.

PR #14474, opened 2019-06-29
bpo-37448: Add radix tree implementation for obmalloc address_in_range().

nascheme:
The radix tree approach is a relatively simple and memory sanitary
alternative to the current (slightly) unsanitary address_in_range().
The radix tree is currently only implemented for 64-bit platforms.
Adding a 32-bit version would be relatively easy.

https://bugs.python.org/issue37448

### DIFF
--- a/Objects/obmalloc.c
+++ b/Objects/obmalloc.c
@@ -2955,15 +2955,18 @@ arena_map_mark_used(uintptr_t arena_base, int is_used)
          * again (do the full tree traversal).
          */
         n_hi->arenas[i3].tail_hi = is_used ? tail : 0;
-        uintptr_t arena_base_next = arena_base + ARENA_SIZE;
-        arena_map3_t *n_lo = arena_map_get((block *)arena_base_next, is_used);
-        if (n_lo == NULL) {
-            assert(is_used); /* otherwise should already exist */
-            n_hi->arenas[i3].tail_hi = 0;
-            return 0; /* failed to allocate space for node */
+        uintptr_t arena_next = arena_base + ARENA_SIZE;
+        /* check for overflow of arena_next */
+        if (arena_next > arena_base) {
+            arena_map3_t *n_lo = arena_map_get((block *)arena_next, is_used);
+            if (n_lo == NULL) {
+                assert(is_used); /* otherwise should already exist */
+                n_hi->arenas[i3].tail_hi = 0;
+                return 0; /* failed to allocate space for node */
+            }
+            int i3_next = MAP3_INDEX(arena_next);
+            n_lo->arenas[i3_next].tail_lo = is_used ? tail : 0;
         }
-        int i3_next = MAP3_INDEX(arena_base_next);
-        n_lo->arenas[i3_next].tail_lo = is_used ? tail : 0;
     }
     return 1;
 }
--- a/Objects/obmalloc.c
+++ b/Objects/obmalloc.c
@@ -907,8 +907,8 @@ static int running_on_valgrind = -1;
  * Arenas are allocated with mmap() on systems supporting anonymous memory
  * mappings to reduce heap fragmentation.
  */
-#define ARENA_BITS              18
-#define ARENA_SIZE              (1 << ARENA_BITS)     /* 256 KiB */
+#define ARENA_BITS              20
+#define ARENA_SIZE              (1 << ARENA_BITS)     /* 1 MiB */
 #define ARENA_SIZE_MASK         (ARENA_SIZE - 1)
 
 #ifdef WITH_MEMORY_LIMITS
@@ -918,8 +918,8 @@ static int running_on_valgrind = -1;
 /*
  * Size of the pools used for small blocks.  Must be a power of 2.
  */
-#define POOL_BITS               12                  /* 4 KiB */
-#define POOL_SIZE               (1 << POOL_BITS)    /* 4 KiB */
+#define POOL_BITS               14                  /* 16 KiB */
+#define POOL_SIZE               (1 << POOL_BITS)
 #define POOL_SIZE_MASK          (POOL_SIZE - 1)
 
 #define MAX_POOLS_IN_ARENA  (ARENA_SIZE / POOL_SIZE)
--- a/Objects/obmalloc.c
+++ b/Objects/obmalloc.c
@@ -1272,7 +1272,7 @@ _Py_GetAllocatedBlocks(void)
    bit allocation for keys
 
    64-bit pointers and 2^20 arena size:
-     16 -> ignored (BITS - PHYSICAL_BITS)
+     16 -> ignored (POINTER_BITS - ADDRESS_BITS)
      10 -> MAP_TOP
      10 -> MAP_MID
       8 -> MAP_BOT
@@ -1291,21 +1291,21 @@ _Py_GetAllocatedBlocks(void)
 #if SIZEOF_VOID_P == 8
 
 /* number of bits in a pointer */
-#define BITS 64
+#define POINTER_BITS 64
 
 /* Current 64-bit processors are limited to 48-bit physical addresses.  For
  * now, the top 17 bits of addresses will all be equal to bit 2**47.  If that
  * changes in the future, this must be adjusted upwards.
  */
-#define PHYSICAL_BITS 48
+#define ADDRESS_BITS 48
 
 /* use the top and mid layers of the radix tree */
 #define USE_INTERIOR_NODES
 
 #elif SIZEOF_VOID_P == 4
 
-#define BITS 32
-#define PHYSICAL_BITS 32
+#define POINTER_BITS 32
+#define ADDRESS_BITS 32
 
 #else
 
@@ -1321,7 +1321,7 @@ _Py_GetAllocatedBlocks(void)
 
 #ifdef USE_INTERIOR_NODES
 /* number of bits used for MAP_TOP and MAP_MID nodes */
-#define INTERIOR_BITS ((PHYSICAL_BITS - ARENA_BITS + 2) / 3)
+#define INTERIOR_BITS ((ADDRESS_BITS - ARENA_BITS + 2) / 3)
 #else
 #define INTERIOR_BITS 0
 #endif
@@ -1334,7 +1334,7 @@ _Py_GetAllocatedBlocks(void)
 #define MAP_MID_LENGTH (1 << MAP_MID_BITS)
 #define MAP_MID_MASK (MAP_MID_LENGTH - 1)
 
-#define MAP_BOT_BITS (PHYSICAL_BITS - ARENA_BITS - 2*INTERIOR_BITS)
+#define MAP_BOT_BITS (ADDRESS_BITS - ARENA_BITS - 2*INTERIOR_BITS)
 #define MAP_BOT_LENGTH (1 << MAP_BOT_BITS)
 #define MAP_BOT_MASK (MAP_BOT_LENGTH - 1)
 
@@ -1347,10 +1347,13 @@ _Py_GetAllocatedBlocks(void)
 #define MAP_MID_INDEX(p) ((AS_UINT(p) >> MAP_MID_SHIFT) & MAP_MID_MASK)
 #define MAP_TOP_INDEX(p) ((AS_UINT(p) >> MAP_TOP_SHIFT) & MAP_TOP_MASK)
 
-#if PHYSICAL_BITS > BITS
-/* Return non-physical bits of pointer.  Should be same for all valid
- * pointers if PHYSICAL_BITS set correctly. */
-#define HIGH_BITS(p) (AS_UINT(p) >> PHYSICAL_BITS)
+#if ADDRESS_BITS > POINTER_BITS
+/* Return non-physical address bits of a pointer.  Those bits should be same
+ * for all valid pointers if ADDRESS_BITS set correctly.  Linux has support for
+ * 57-bit address space (Intel 5-level paging) but will not currently give
+ * those addresses to user space.
+ */
+#define HIGH_BITS(p) (AS_UINT(p) >> ADDRESS_BITS)
 #else
 #define HIGH_BITS(p) 0
 #endif
@@ -1400,7 +1403,7 @@ static arena_map_bot_t *
 arena_map_get(block *p, int create)
 {
 #ifdef USE_INTERIOR_NODES
-    /* sanity check that PHYSICAL_BITS is correct */
+    /* sanity check that ADDRESS_BITS is correct */
     assert(HIGH_BITS(p) == HIGH_BITS(&arena_map_root));
     int i1 = MAP_TOP_INDEX(p);
     if (arena_map_root.ptrs[i1] == NULL) {
@@ -1460,7 +1463,7 @@ arena_map_get(block *p, int create)
 static int
 arena_map_mark_used(uintptr_t arena_base, int is_used)
 {
-    /* sanity check that PHYSICAL_BITS is correct */
+    /* sanity check that ADDRESS_BITS is correct */
     assert(HIGH_BITS(arena_base) == HIGH_BITS(&arena_map_root));
     arena_map_bot_t *n_hi = arena_map_get((block *)arena_base, is_used);
     if (n_hi == NULL) {
--- a/Objects/obmalloc.c
+++ b/Objects/obmalloc.c
@@ -2967,22 +2967,20 @@ arena_map_mark_used(uintptr_t arena_base, int is_used)
          * again (do the full tree traversal).
          */
         n_hi->arenas[i3].tail_hi = is_used ? tail : 0;
-        uintptr_t arena_next = arena_base + ARENA_SIZE;
-        /* If arena_base is a legit arena address, so is arena_next - 1
-         * (last address in arena).  If arena_next overflows then it
+        uintptr_t arena_base_next = arena_base + ARENA_SIZE;
+        /* If arena_base is a legit arena address, so is arena_base_next - 1
+         * (last address in arena).  If arena_base_next overflows then it
          * must overflow to 0.  However, that would mean arena_base was
          * "ideal" and we should not be in this case. */
-        assert(arena_base < arena_next);
-        if (arena_next > arena_base) {
-            arena_map3_t *n_lo = arena_map_get((block *)arena_next, is_used);
-            if (n_lo == NULL) {
-                assert(is_used); /* otherwise should already exist */
-                n_hi->arenas[i3].tail_hi = 0;
-                return 0; /* failed to allocate space for node */
-            }
-            int i3_next = MAP3_INDEX(arena_next);
-            n_lo->arenas[i3_next].tail_lo = is_used ? tail : 0;
+        assert(arena_base < arena_base_next);
+        arena_map3_t *n_lo = arena_map_get((block *)arena_base_next, is_used);
+        if (n_lo == NULL) {
+            assert(is_used); /* otherwise should already exist */
+            n_hi->arenas[i3].tail_hi = 0;
+            return 0; /* failed to allocate space for node */
         }
+        int i3_next = MAP3_INDEX(arena_base_next);
+        n_lo->arenas[i3_next].tail_lo = is_used ? tail : 0;
     }
     return 1;
 }
--- a/Objects/obmalloc.c
+++ b/Objects/obmalloc.c
@@ -834,15 +834,6 @@ static int running_on_valgrind = -1;
  * -- Main tunable settings section --
  */
 
-/* If defined, use radix tree to find if address is controlled by
- * obmalloc.  Otherwise, we use a slightly memory unsanitary scheme that
- * has the advantage of performing very well.
- */
-#if SIZEOF_VOID_P == 8
-#define WITH_RADIX_TREE
-#endif
-#define WITH_RADIX_TREE
-
 /*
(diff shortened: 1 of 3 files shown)


Objects/obmalloc.c
```diff
@@ -2739,4 +2802,195 @@ _PyObject_DebugMallocStats(FILE *out)
...
+ * _PyObject_Free() calls this so it needs to be very fast. */
+static int
+arena_map_is_marked(block *p)
+{
+    arena_map3_t *n = arena_map_get(p, 0);
+    if (n == NULL) {
+        return 0;
+    }
+    int i3 = MAP3_INDEX(p);
+    /* in order to fit tail into 32-bits, ARENA_BITS must be <= 32 */
```
tim-one: `<= 32` is off by 1.  Suggest instead, e.g.,

    /* ARENA_BITS must be < 32 so that the tail is a non-negative int32_t. */

Objects/obmalloc.c
```diff
@@ -2739,4 +2802,195 @@ _PyObject_DebugMallocStats(FILE *out)
...
+    }
+    else {
+        /* arena_base address is not ideal (aligned to arena size) and
+         * so it potentially covers two MAP3 nodes.  Get the MAP3 node
+         * for the next arena.  Note that it might be in different MAP1
+         * and MAP2 nodes as well so we need to call arena_map_get()
+         * again (do the full tree traversal).
+         */
+        n_hi->arenas[i3].tail_hi = is_used ? tail : 0;
+        uintptr_t arena_base_next = arena_base + ARENA_SIZE;
```
tim-one: Alas, overflow is theoretically possible here.  Although it may happen only once in the next 50 years ☹️  If it does, there's no possibility for this arena to "spill over" into the next ideal arena address, so guarding the rest of this block with, e.g.,

`if (arena_base_next > arena_base) {  // no overflow`

should be fine.  Since this code is executed relatively rarely, I don't care about speed here.

tim-one:
Food for thought:  consider making L1 fatter?  My understanding is that Intel and AMD chips are limited to at most 48 physical address bits now, and they agreed that, for now, the top 16 bits of virtual addresses must be copies of bit 2**47.  Which was intentional, to stop "clever" OS authors and users from using the higher-order bits for something else, thus making future expansion of physical address bits essentially impossible.  Jerks - we could have used the sign bit to mean "address in range" :wink:. 

So the top 17 bits will all be 0 or all be 1, and any L1 width <= 17 will never see more than those 2 values.  On Windows, I expect they'll only see "all 0", because the OS reserves the "all 1" patterns for its own use - don't know about Linux.

I bet the next step will be 56-bit physical addresses.

Objects/obmalloc.c
tim-one: On second thought, if `arena_base` is a legit arena base address, then so is `arena_base + ARENA_SIZE - 1`, the last address in the arena.  So if 1 more than that overflows, it must overlow to 0, in which case `arena_base` was arena-aligned to begin with, and so would have taken the earlier "is ideal arena address" branch instead.  So I'd be happy here just to assert:

    assert(arena_base < arena_base_next); // no overflow

tim-one:
Note that on a 32-bit box, with 256K arenas there are only 2**(32-18) = 2**14 = 16K ideal arena addresses.  So a static 1-level tree would be fine.

nascheme:
Regarding making the root node larger, that sounds like a good idea.  I'm not sure how large is good though.  It increases the virtual set size of the process.  I compared VSS and RSS of the base Python (current address_in_range()), small root radix node (MAP1_BITS = 16) and MAP1_BITS = 20.

| Program | RSS [MiB] | VSS [MiB] |
|----------|-----|------|
| Base | 11.4 | 12.9 |
| 16 bits | 11.4 | 14.0 |
| 20 bits | 11.4 | 21.2 |

Below is my quick and dirty change to use 20 bits for the root node.

I'm hesitant to increase the VSS.  On Linux it is not much of a problem because Linux typically is very willing to overcommit on RAM.  However, other operating systems might be different.  E.g. I think FreeBSD is more conservative.  Not sure about Windows.  The idea is to avoid having to kill programs due to an out-of-memory situation.  If they don't overcommit, they will refuse to start new programs if the VSS is too large.  At least, that's my understanding.


```
--- a/Objects/obmalloc.c
+++ b/Objects/obmalloc.c
@@ -2831,20 +2831,27 @@ _PyObject_DebugMallocStats(FILE *out)
 #   error "arena size must be < 2^32"
 #endif
 
-/* bits used for MAP1 and MAP2 nodes */
-#define INTERIOR_BITS ((BITS - ARENA_BITS + 2) / 3)
-
-#define MAP1_BITS INTERIOR_BITS
+/* Current 64-bit processors are limited to 48-bit physical addresses.  For
+ * now, the top 17 bits of addresses will all be equal to bit 2**47.  That is
+ * one reason to make MAP1_LENGTH large.  Another reason is that we initialize
+ * the root node in the BSS.  For modern operating systems with virtual memory,
+ * it will only consume actual RAM for the OS pages that are touched.
+ */
+#define MAP1_BITS 20
 #define MAP1_LENGTH (1 << MAP1_BITS)
 
-#define MAP2_BITS INTERIOR_BITS
+#define MAP2_BITS ((BITS - MAP1_BITS - ARENA_BITS)/2)
 #define MAP2_LENGTH (1 << MAP2_BITS)
 #define MAP2_MASK (MAP2_LENGTH - 1)
 
-#define MAP3_BITS (BITS - ARENA_BITS - 2*INTERIOR_BITS)
+#define MAP3_BITS (BITS - MAP1_BITS - MAP2_BITS - ARENA_BITS)
 #define MAP3_LENGTH (1 << MAP3_BITS)
 #define MAP3_MASK (MAP3_LENGTH - 1)
 
+#if MAP3_BITS < 8
+#   error "MAP3_BITS too small, adjust MAP1_BITS"
+#endif
+
 #define MAP3_SHIFT ARENA_BITS
 #define MAP2_SHIFT (MAP3_BITS + MAP3_SHIFT)
 #define MAP1_SHIFT (MAP2_BITS + MAP2_SHIFT)

```

nascheme:
Regarding a 32-bit version, I'm not sure what to do there.  We should at least do some benchmarking on common 32-bit platforms.  I would guess most common would be 32-bit Windows on an amd64 processor or Linux on a 32-bit ARM processor.  In those cases, maybe we could keep POOL_SIZE <= 4096 and keep the existing address_in_range()?

tim-one:
I had in mind more widening L1 by narrowing it ;-)  That is, throw away the top 16 address bits entirely (in debug mode verifying they're the same as the top bit that remains).  Then we only have 48 - 18 = 30 arena prefix bits for the tree levels to resolve.

Yes, that's aggressive.  But then, best I can tell, the entire useful effect of widening L1 to 20 bits as-is could be gotten by shrinking it to 4 bits instead.  Of course the tree would have to be reworked a bit when physical address bits increase.  Keep it 3 levels, and perhaps it could just be a matter of changing a `PHYSICAL_ADDRESS_BITS` `#define`.

FYI, there is no overcommit on Windows, and no OOM killer.  While doing a "commit" with `VirtualAlloc` doesn't assign physical RAM (same as Linux, that's assigned as pages are later touched), the amount is charged against a system-wide total of physical RAM + swapfile size.  The commit fails if that total would be exceeded.

So increasing (the moral equivalent of) VSS there is even less atractive.  But there's something to be said for writing code that exploits hardware realities:  there are only 48 address bits :wink:.  Then VSS can shrink.

nascheme:
> So increasing (the moral equivalent of) VSS there is even less atractive. But there's something to be said for writing code that exploits hardware realities: there are only 48 address bits wink. Then VSS can shrink

Ah, that is very pragmatic.  Thank you for explaining it to me. ;-P

I have made a patch and done some testing.  Here is an updated table.  Obviously the radix tree is tiny in the case that it only uses 30 bits.  I have root node using 10 bits.

| Program | RSS [MiB] | VSS [MiB] |
|----------|-----|------|
| Base | 11.4 | 12.9 |
| 16 bits | 11.4 | 14.0 |
| 20 bits | 11.4 | 21.2 |
| 10 bits | 11.2 | 12.9 |

tim-one:
Cool!  OTOH ... we're making much stronger assumptions now, and it's unclear how long they'll last.  Or even if they all apply even now.  For example, I know that the 64-bit PowerPC architecture still lives on in some very high-end systems, but don't know whether Python is used on them, or what their virtual address conventions are.

FYI, looks like "the next step" in x64-land will be 57-bit physical addresses rather than 56:

https://en.wikipedia.org/wiki/Intel_5-level_paging

And Linux has apparently already been patched for that, but in anticipation - no chips can use it yet.

Prudent:  yes, leave 32-bit Python entirely alone.  But also make it very easy to revert 64-bit Python to the status quo too (e.g., if `PY38OBMALLOC` is defined).

Objects/obmalloc.c
```diff
@@ -2739,4 +2802,214 @@ _PyObject_DebugMallocStats(FILE *out)
...
+         * again (do the full tree traversal).
+         */
+        n_hi->arenas[i3].tail_hi = is_used ? tail : 0;
+        uintptr_t arena_next = arena_base + ARENA_SIZE;
+        /* If arena_base is a legit arena address, so is arena_next - 1
+         * (last address in arena).  If arena_next overflows then it
+         * must overflow to 0.  However, that would mean arena_base was
+         * "ideal" and we should not be in this case. */
+        assert(arena_base < arena_next);
+        if (arena_next > arena_base) {
```
tim-one: We don't need _both_ an assert and an `if` on the same thing :wink:.  I'd be happy to have the assert alone, but don't really care.  If you want to keep both, at least spell them the same way so it's dead obvious they're testing exactly the same thing?

nascheme:
> Cool! OTOH ... we're making much stronger assumptions now, and it's unclear how long they'll last. Or even if they all apply even now. For example, I know that the 64-bit PowerPC architecture still lives on in some very high-end systems, but don't know whether Python is used on them, or what their virtual address conventions are.

I found this recent bug fix:

[powerpc/mm/64s/hash: Reallocate context ids on fork](https://git.kernel.org/pub/scm/linux/kernel/git/torvalds/linux.git/commit/?id=ca72d88378b2f2444d3ec145dd442d449d3fefbc)

I think that means with certain setups, user mode programs on POWER9 processors can see addresses in the 4 PB range (52 bits?).
> 
> FYI, looks like "the next step" in x64-land will be 57-bit physical addresses rather than 56:
> 
> https://en.wikipedia.org/wiki/Intel_5-level_paging
> 
> And Linux has apparently already been patched for that, but in anticipation - no chips can use it yet.

Yes, the patch is in the current release:
https://www.kernel.org/doc/html/latest/x86/x86_64/5level-paging.html

You will be amused to read that you are not the first to think of hiding flags in the high bits of pointers.
  
> Prudent: yes, leave 32-bit Python entirely alone. But also make it very easy to revert 64-bit Python to the status quo too (e.g., if `PY38OBMALLOC` is defined).

There is a define for that: WITH_RADIX_TREE.  Easy to turn off.

nascheme:
The branch has been updated to add 32-bit support.  For 32-bit pointers, two levels of the tree have been eliminated vs the 64-bit version.

Now the radix tree is enabled unconditionally for all CPython platforms.  If desired, we can restore the `WITH_RADIX_TREE` conditional and leave the old `address_in_range()` logic in there.  However, I think it's cleaner just to remove it.  The only reason to keep it is if there is some platform that the radix tree doesn't work on or it is (materially) slower than the radix tree version.

I went ahead an also increased the arena and pool sizes by 4x.  That's based on advice from Tim Peters.  He says:

> I'm all in favor of merging this if it included, say, quadrupling
arena and pool sizes. Then the radix tree is a much better technical
approach then the "big pool" hack I coded. 

Tim's big-pool patch is here: https://bugs.python.org/issue37211

I've run pyperformance benchmarks.  Without increasing pool and arena sizes, the radix version of obmalloc is 1.01x or 1.02x slower on some of the benchmarks.  With the size increases, it is about that much faster.  So, I would say basically a wash.

nascheme:
Small update: reduce arena and pool sizes on 32-bit platforms.  Sizes match Tim's big-pool patch.  Below are memory usage stats based on testing on Linux 32-bit container:

before change:
```
RSS 6.68 MB
VmPeak:	    9376 kB
VmSize:	    9364 kB
VmHWM:	    6676 kB
VmRSS:	    6676 kB
VmData:	    1876 kB
VmStk:	     132 kB
VmExe:	    2072 kB
VmLib:	    2336 kB
VmPTE:	      44 kB
```

after change:
```
RSS 6.72 MB
VmPeak:	    9504 kB
VmSize:	    9492 kB
VmRSS:	    6716 kB
VmData:	    2004 kB
VmStk:	     132 kB
VmExe:	    2072 kB
VmLib:	    2336 kB
VmPTE:	      40 kB
```

Objects/obmalloc.c
```diff
@@ -1233,6 +1243,259 @@ _Py_GetAllocatedBlocks(void)
...
+*/
+
+#if SIZEOF_VOID_P == 8
+
+/* number of bits in a pointer */
+#define BITS 64
+
+/* Current 64-bit processors are limited to 48-bit physical addresses.  For
+ * now, the top 17 bits of addresses will all be equal to bit 2**47.  If that
+ * changes in the future, this must be adjusted upwards.
```
methane, line 1298: Is the 57-bit virtual address space considered?
https://en.wikipedia.org/wiki/Intel_5-level_paging

nascheme: It is only considered in as far as having the PHYSICAL_BITS define.  On systems that implement more address space, we would need some conditional logic to set PHYSICAL_BITS to the appropriate value.  There is a bit of discussion above about it, e.g. PowerPC systems.

nascheme: Thinking about this more, I could add a configure option like `--enable-virtual-addr-size=57` that would allow people to override the PHYSICAL_BITS define.  The calculations for the sizes of the radix tree levels should adapt.

A downside to that approach is if you build for 48-bits and run on a 57-bits addr platform, Python could just crash.   An alternative would be to somehow detect the size of the virtual address space at runtime and then swap out the `address_in_range` function as needed.  I.e. have two implementations of the radix tree and choose the right one.

It seems Go just hard-codes it at 48 bits:
https://github.com/golang/go/blob/6cc8aa7ece96aca282db19f08aa5c98ed13695d9/src/runtime/lfstack_64bit.go#L26
```
	// On s390x, virtual addresses are 64-bit. There's not much we
	// can do about this, so we just hope that the kernel doesn't
	// get to really high addresses and panic if it does.
	addrBits = 48
```

Objects/obmalloc.c
```diff
@@ -1233,6 +1243,259 @@ _Py_GetAllocatedBlocks(void)
...
+     10 -> MAP_TOP
+     10 -> MAP_MID
+      8 -> MAP_BOT
+     20 -> ideal aligned arena
+   ----
+     64
+
+   32-bit pointers:
+     12 -> MAP_BOT
+     20 -> ideal aligned arena
```
methane: 18?

nascheme: I believe the numbers are correct for 2^20 size pools.  However, I changed to code to use 2^18 sized pools if SIZEOF_VOID_P <= 4, on 32-bit platform.  So, maybe I should change the example to match, e.g.
```
   64-bit pointers, for 2^20 arena size:
     16 -> ignored (BITS - PHYSICAL_BITS)
     10 -> MAP_TOP
     10 -> MAP_MID
      8 -> MAP_BOT
     20 -> ideal aligned arena
   ----
     64

   32-bit pointers, for 2^18 arena size:
     14 -> MAP_BOT
     18 -> ideal aligned arena
   ----
     32
```

Objects/obmalloc.c
methane: OK. I found "To mitigate this, we are not going to allocate virtual address space above 47-bit by default." in the [linux doc](https://www.kernel.org/doc/html/latest/x86/x86_64/5level-paging.html).

methane (approved):
LGTM, overall. But I'm bit afraid of not having compile switch to use old behavior.

vstinner:
> The address_in_range() function must be fast.

If the intent of this change is to optimize pymalloc, can you please run a microbenchmark to prove it?

nascheme:
> LGTM, overall. But I'm bit afraid of not having compile switch to use old behavior.

It's a bit more code to leave the old version in but not too much.  I can do that.  Should I also add a "configure" command line option?  E.g. `--disable-obmalloc-radix-tree`.

methane:
> It's a bit more code to leave the old version in but not too much. I can do that. Should I also add a "configure" command line option? E.g. `--disable-obmalloc-radix-tree`.

Configure option is not required. If some Python user find memory usage regression, they can build Python with `make -D PYMALLOC_RADIXTREE=0`.

gvanrossum:
@nascheme So what's stopping us from landing this?

nascheme:
Only my own caution, I think.  It is easy to back out or disable if we decide it's not good.  I have a small cleanup to variable names to make but I can merge after that.

gvanrossum:
Should probably get a review from @pablogsal (Pablo, you can also delegate to someone else).

eduardo-elizondo:
Hey everyone, wanted to add an extra data point here. We tested this change with the Instagram workload and all the health metrics looked fine (i.e no weird crashes / behavior). Also, performance-wise, this looks like a net-neutral change.

pablogsal:
> Hey everyone, wanted to add an extra data point here. We tested this change with the Instagram workload and all the health metrics looked fine (i.e no weird crashes / behavior). Also, performance-wise, this looks like a net-neutral change.

Thanks a lot @eduardo-elizondo, that will help us a lot when making a review and a final decision. Seems that the main thing we need to consider is maintenance cost and possible semantic change, but I am optimistic about both. I still need to find time to review it in detail.

pablogsal:
One question: does valgrind complains the same way it used to with the old `address_in_range()` with this change? Is this situation better with the radix tree?

tim-one:
Pablo, I never ran valgrind myself, but a primary _point_ of this different approach is that it never references uninitialized memory (well, assuming users don't pass insane addresses to `free()/realloc()`). The current approach does, which is why valgrind legitimately (but spuriously) complains. valgrind "should be" 100% happy with what the PR does instead.

pablogsal (approved):

nascheme:
Buildbot failure on S390x seems to be unrelated to this change.  See https://bugs.python.org/issue37368 .

Objects/obmalloc.c
glaubitz: 48 bits is not correct. At least sparc64 and ia64 use larger virtual address spaces. And FWIW, this change causes Python to segfault on ia64.

I have not verified yet whether bumping this to 64 will fix the problem.

Objects/obmalloc.c
```diff
@@ -1233,6 +1265,264 @@ _Py_GetAllocatedBlocks(void)
...
+     32
+
+*/
+
+#if SIZEOF_VOID_P == 8
+
+/* number of bits in a pointer */
+#define POINTER_BITS 64
+
+/* Current 64-bit processors are limited to 48-bit physical addresses.  For
```
jrtc27, line 1296: You're talking about *virtual* addresses here, Python does not see physical addresses unless run bare-metal.

But that's not true, you're only thinking of x86, but even then LA57 support is a thing that's been floating around for half a decade now. AArch64, depending on hardware support and configuration, can go up to 52-bit addresses. Some 64-bit SPARC machines have full 64-bit virtual address space support (though the 4-level Linux kernel implementation can only use up to 53). Itanium is complicated; it seems to use up to 44 of the low bits, at least on this specific machine, but the _top three bits_ are also significant, as they are a region number, within which the other bits are interpreted (i.e. you get 8 separate address spaces you can configure differently).

jrtc27: (RISC-V also has only Sv39 and Sv48 currently but Sv57 is on track to be ratified very soon)

nascheme: It sounds like we need to make the comment more accurate and add some extra ifdefs for these machines.  The easiest fix is to set `PYMALLOC_RADIXTREE=0` for those machines.

jrtc27: Or perhaps better to have the throwing-bits-away version opt-in rather than opt-out, so some architecture you've never heard of, or doesn't yet exist, works, albeit perhaps a little less efficiently, rather than breaking?

Objects/obmalloc.c
```diff
@@ -1233,6 +1265,264 @@ _Py_GetAllocatedBlocks(void)
...
+#ifdef USE_INTERIOR_NODES
+/* number of bits used for MAP_TOP and MAP_MID nodes */
+#define INTERIOR_BITS ((ADDRESS_BITS - ARENA_BITS + 2) / 3)
+#else
+#define INTERIOR_BITS 0
+#endif
+
+#define MAP_TOP_BITS INTERIOR_BITS
+#define MAP_TOP_LENGTH (1 << MAP_TOP_BITS)
+#define MAP_TOP_MASK (MAP_BOT_LENGTH - 1)
```
methane, line 1331: MAP_TOP_LENGTH?

nascheme: Yup it's a bug, it should be TOP not BOT.  On 64-bit machines, MAP_BOT_LENGTH is 256 and MAP_TOP_LENGTH is 1024.

PR #14480, opened 2019-06-30
bpo-37199: Fix test failures when IPv6 is unavailable or disabled

ZackerySpytz:
https://bugs.python.org/issue37199

### DIFF
--- a/Lib/test/test_asyncio/test_base_events.py
+++ b/Lib/test/test_asyncio/test_base_events.py
@@ -91,6 +91,9 @@ class BaseEventTests(test_utils.TestCase):
         self.assertIsNone(
             base_events._ipaddr_info('1.2.3.4', 1, UNSPEC, 0, 0))
 
+        if not support.IPV6_ENABLED:
+            return
+
         # IPv4 address with family IPv6.
         self.assertIsNone(
             base_events._ipaddr_info('1.2.3.4', 1, INET6, STREAM, TCP))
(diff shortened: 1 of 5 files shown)


asvetlov (approved):

Lib/test/test_asyncio/test_base_events.py
```diff
@@ -91,6 +91,9 @@ def test_ipaddr_info(self):
         self.assertIsNone(
             base_events._ipaddr_info('1.2.3.4', 1, UNSPEC, 0, 0))
 
+        if not support.IPV6_ENABLED:
+            return
```
tiran, line 95: I don't like this approach, because it's hard to see that the rest of the test is skipped. Please either split the test function into two or move the rest of the code into the ``if not supported.IPV6_ENABLED`` block.

ZackerySpytz: Okay. I am willing to make such changes, but I would like to hear what @asvetlov thinks.

asvetlov: @ZackerySpytz please make a PR with changes suggested by @tiran 

My first reaction was opening the source code to make sure that the `return` doesn't make harm.
On the other hand, I've committed the PR as is because you proposed a minimal possible diff.

ZackerySpytz: Okay. I have created GH-14535.

PR #14483, opened 2019-06-30
bpo-37451: remove redundant code in _testcapimodule.c

shihai1991:
https://bugs.python.org/issue37451

### DIFF
--- a/Modules/_testcapimodule.c
+++ b/Modules/_testcapimodule.c
@@ -1860,8 +1860,6 @@ unicode_aswidechar(PyObject *self, PyObject *args)
 
     if (size < buflen)
         buflen = size + 1;
-    else
-        buflen = size;
     result = PyUnicode_FromWideChar(buffer, buflen);
     PyMem_Free(buffer);
     if (result == NULL)


mangrisano:
Hi. Thank you for the pull request. I'm not an expert, therefore, would you mind telling me why that is redundant? Just for learning. Thank you :)

jdemeyer (changes requested):
I don't see why this code should be removed. It seems to be used for testing that `PyUnicode_FromWideChar` can deal with NULL-terminated and non-NULL-terminated buffers.

So at least, this PR needs better justification.

shihai1991:
> I don't see why this code should be removed. It seems to be used for testing that `PyUnicode_FromWideChar` can deal with NULL-terminated and non-NULL-terminated buffers.
> 
> So at least, this PR needs better justification.

shihai1991:
> I don't see why this code should be removed. It seems to be used for testing that `PyUnicode_FromWideChar` can deal with NULL-terminated and non-NULL-terminated buffers.
> 
> So at least, this PR needs better justification.

Jeroen, thanks for review.

`PyUnicode_FromWideChar` can return two kind of values of three scenarios:
1) `widechar_size + 1` if buffer is NULL or buflen > widechar_size + 1
    1.1 buffer is NULL can not happen in `PyUnicode_FromWideChar` because it have been judged in 
          https://github.com/python/cpython/blob/master/Modules/_testcapimodule.c#L1852
    1.2 buflen > widechar_size + 1 is judged in _testcapimodule.c is ok.
2) `buflen`
     2.1 the latest return value is buflen  or size(buflen==size) because of 
           https://github.com/python/cpython/blob/master/Objects/unicodeobject.c#L3164.

So I think the else statement in _testcapimodule.c is redundant.

shihai1991:
> Hi. Thank you for the pull request. I'm not an expert, therefore, would you mind telling me why that is redundant? Just for learning. Thank you :)

Hi, Michele. Discussing process over result ;). I have answered downstairs.

shihai1991:
> I don't see why this code should be removed. It seems to be used for testing that `PyUnicode_FromWideChar` can deal with NULL-terminated and non-NULL-terminated buffers.
> 
> So at least, this PR needs better justification.

jdemeyer (commented):
OK, I see what you mean now. But given that it took me a while to understand why your PR was correct, I'm not convinced that we should merge it as-is.

shihai1991:
> OK, I see what you mean now. But given that it took me a while to understand why your PR was correct, I'm not convinced that we should merge it as-is.

it's a small change and don't effect any functions, if core teams think it is not necessary, pls help me close this PR.
Don't waste too much time on it ;)

zhangyangyu:
Thanks for the contribution and I could understand your logic. It actually is from my point of view but it's quite tiny and doesn't affect anything so I think it's not necessary to change it.

PR #14488, opened 2019-06-30
bpo-36889: Document Stream class and add docstrings

tirkarthi:
* This just copies the docs from `StreamWriter` and `StreamReader`.
* Add docstring for asyncio functions.

### DIFF
--- a/Doc/library/asyncio-api-index.rst
+++ b/Doc/library/asyncio-api-index.rst
@@ -132,17 +132,42 @@ High-level APIs to work with network IO.
     :widths: 50 50
     :class: full-width-table
 
+    * - ``await`` :func:`connect`
+      -  Establish a TCP connection to send and receive data.
+
     * - ``await`` :func:`open_connection`
       -  Establish a TCP connection.
+
+    * - ``await`` :func:`connect_unix`
+      -  Establish a Unix socket connection to send and receive data.
 
     * - ``await`` :func:`open_unix_connection`
       -  Establish a Unix socket connection.
 
+    * - :class:`StreamServer`
+      - Start a TCP server.
+
     * - ``await`` :func:`start_server`
       - Start a TCP server.
 
+    * - :class:`UnixStreamServer`
+      - Start a Unix socket server.
+
     * - ``await`` :func:`start_unix_server`
       - Start a Unix socket server.
+
+    * - :func:`connect_read_pipe`
+      - Establish a connection to :term:`file-like object <file object>` *pipe*
+        to receive data.
+
+    * - :func:`connect_write_pipe`
+      - Establish a connection to :term:`file-like object <file object>` *pipe*
+        to send data.
+
+    * - :class:`Stream`
+      - High-level async/await object to send and receive network data.
+        It includes the API provided by :class:`StreamReader`
+        and :class:`StreamWriter`.
 
     * - :class:`StreamReader`
       - High-level async/await object to receive network data.
--- a/Doc/library/asyncio-api-index.rst
+++ b/Doc/library/asyncio-api-index.rst
@@ -136,25 +136,25 @@ High-level APIs to work with network IO.
       -  Establish a TCP connection to send and receive data.
 
     * - ``await`` :func:`open_connection`
-      -  Establish a TCP connection.
+      -  Establish a TCP connection. (Deprecated in favor of :func:`connect`)
 
     * - ``await`` :func:`connect_unix`
       -  Establish a Unix socket connection to send and receive data.
 
     * - ``await`` :func:`open_unix_connection`
-      -  Establish a Unix socket connection.
+      -  Establish a Unix socket connection. (Deprecated in favor of :func:`connect_unix`)
 
     * - :class:`StreamServer`
       - Start a TCP server.
 
     * - ``await`` :func:`start_server`
-      - Start a TCP server.
+      - Start a TCP server. (Deprecated in favor of :class:`StreamServer`)
 
     * - :class:`UnixStreamServer`
       - Start a Unix socket server.
 
     * - ``await`` :func:`start_unix_server`
-      - Start a Unix socket server.
+      - Start a Unix socket server. (Deprecated in favor of :class:`UnixStreamServer`)
 
     * - :func:`connect_read_pipe`
       - Establish a connection to :term:`file-like object <file object>` *pipe*
@@ -165,15 +165,14 @@ High-level APIs to work with network IO.
         to send data.
 
     * - :class:`Stream`
-      - High-level async/await object to send and receive network data.
-        It includes the API provided by :class:`StreamReader`
-        and :class:`StreamWriter`.
+      - Stream is a single object combining APIs of :class:`StreamReader` and
+        :class:`StreamWriter`.
 
     * - :class:`StreamReader`
-      - High-level async/await object to receive network data.
+      - High-level async/await object to receive network data. (Deprecated in favor of :class:`Stream`)
 
     * - :class:`StreamWriter`
-      - High-level async/await object to send network data.
+      - High-level async/await object to send network data. (Deprecated in favor of :class:`Stream`)
 
 
 .. rubric:: Examples
--- a/Doc/library/asyncio-stream.rst
+++ b/Doc/library/asyncio-stream.rst
@@ -377,7 +377,7 @@ Stream
 
    .. attribute:: mode
 
-      Returns the mode of the stream which is a :class:`StreamMode`.
+      Returns the mode of the stream which is a :class:`StreamMode` value.
 
    .. method:: set_exception(exc)
 
@@ -387,6 +387,21 @@ Stream
    .. coroutinemethod:: abort()
 
       Aborts the :ref:`Transport <asyncio-transport>` instance used by the Stream.
+
+   .. coroutinemethod:: drain()
+
+      Wait until it is appropriate to resume writing to the stream.
+      Example::
+
+          stream.write(data)
+          await stream.drain()
+
+      This is a flow control method that interacts with the underlying
+      IO write buffer.  When the size of the buffer reaches
+      the high watermark, *drain()* blocks until the size of the
+      buffer is drained down to the low watermark and writing can
+      be resumed.  When there is nothing to wait for, the :meth:`drain`
+      returns immediately.
 
    .. coroutinemethod:: sendfile(file, offset=0, count=None, *, fallback=True)
 
@@ -443,16 +458,6 @@ Stream
       buffer is reset.  The :attr:`IncompleteReadError.partial` attribute
       may contain a portion of the separator.
 
-   .. method:: at_eof()
-
-      Return ``True`` if the buffer is empty and :meth:`feed_eof`
-      was called.
-
-   .. method:: feed_data(data)
-
-      Adds ``data`` to the write buffer. It raises an :exc:`AssertionError` when
-      it was called after :meth:`feed_eof`.
-
    .. method:: write(data)
 
       The method attempts to write the *data* to the underlying socket immediately.
@@ -491,11 +496,6 @@ Stream
       The ``await`` pauses the current coroutine until the stream and the underlying
       socket are closed (and SSL shutdown is performed for a secure connection).
 
-      Below is an equivalent code that works with Python <= 3.7::
-
-         stream.close()
-         await stream.wait_closed()
-
    .. method:: can_write_eof()
 
       Return *True* if the underlying transport supports
@@ -506,6 +506,16 @@ Stream
       Close the write end of the stream after the buffered write
       data is flushed.
 
+   .. method:: at_eof()
+
+      Return ``True`` if the buffer is empty and :meth:`feed_eof`
+      was called.
+
+   .. method:: feed_data(data)
+
+      Adds ``data`` to the write buffer. It raises an :exc:`AssertionError` when
+      it was called after :meth:`feed_eof`.
+
    .. attribute:: transport
 
       Return the underlying asyncio transport.
@@ -514,21 +524,6 @@ Stream
 
       Access optional transport information; see
       :meth:`BaseTransport.get_extra_info` for details.
-
-   .. coroutinemethod:: drain()
-
-      Wait until it is appropriate to resume writing to the stream.
-      Example::
-
-          stream.write(data)
-          await stream.drain()
-
-      This is a flow control method that interacts with the underlying
-      IO write buffer.  When the size of the buffer reaches
-      the high watermark, *drain()* blocks until the size of the
-      buffer is drained down to the low watermark and writing can
-      be resumed.  When there is nothing to wait for, the :meth:`drain`
-      returns immediately.
 
    .. method:: is_closing()
 
--- a/Doc/library/asyncio-stream.rst
+++ b/Doc/library/asyncio-stream.rst
@@ -46,7 +46,6 @@ and work with streams:
    Connect to TCP socket on *host* : *port* address and return a :class:`Stream`
    object of mode :attr:`StreamMode.READWRITE`.
 
-
    *limit* determines the buffer size limit used by the returned :class:`Stream`
    instance. By default the *limit* is set to 64 KiB.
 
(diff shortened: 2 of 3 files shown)


Doc/library/asyncio-stream.rst
```diff
@@ -367,13 +366,177 @@ Stream
...
+   :meth:`readline` returns empty data.
 
    Do not instantiate *Stream* objects directly; use API like :func:`connect`
    and :class:`StreamServer` instead.
 
    .. versionadded:: 3.8
 
+   .. attribute:: mode
+
+      Returns the mode of the stream which is a :class:`StreamMode` value.
```
1st1: I think we need to elaborate on this.  Add a sentence to list the enum values & explain what they mean.

Doc/library/asyncio-stream.rst
```diff
@@ -367,13 +366,177 @@ Stream
...
+
+   .. method:: set_exception(exc)
+
+      Sets the exception to be raised when :meth:`read`, :meth:`readexactly` and
+      :meth:`readuntil` are called.
```
1st1: Ouch.  @asvetlov Why is this a public API?

Doc/library/asyncio-stream.rst
```diff
@@ -367,13 +366,177 @@ Stream
```
1st1, line 366: Please update *high level API index* with the new APIs.

https://docs.python.org/3.8/library/asyncio-api-index.html

Doc/library/asyncio-stream.rst
```diff
@@ -367,13 +366,177 @@ Stream
...
+
+   .. coroutinemethod:: sendfile(file, offset=0, count=None, *, fallback=True)
+
+      Calls :meth:`Stream.drain()` and rest of the arguments are passed directly to
+      :meth:`loop.sendfile`.
+
+   .. coroutinemethod:: start_tls(sslcontext, *, server_hostname=None, \
+                                  ssl_handshake_timeout=None)
+
+      Calls :meth:`Stream.drain()` and rest of the arguments are passed directly to
```
1st1: I'd actually elaborate a bit on this method - why we have it, ideally, with an example.

Lib/asyncio/streams.py
```diff
@@ -70,6 +70,14 @@ def connect(host=None, port=None, *,
             server_hostname=None,
             ssl_handshake_timeout=None,
             happy_eyeballs_delay=None, interleave=None):
+    """Connect to TCP socket on *host* : *port* address and return a `Stream`
```
asvetlov: Minor: in docstrings for CPython we usually avoid markup. The docstring is for human reading, the documentation is generated from hand-written Sphinx pages.
`help()` doesn't render embedded markup too.
I know, older docstrings for the module contains back tickets at least.
Should we add new markup in this PR or avoid it in new updates?
@1st1 I need your opinion here

Lib/asyncio/streams.py
```diff
@@ -108,6 +116,11 @@ def connect(host=None, port=None, *,
 
 
 def connect_read_pipe(pipe, *, limit=_DEFAULT_LIMIT):
+    """Takes a file-like object *pipe* to return a Stream object of the mode
```
1st1: Please make sure that all docstrings follow this: First paragraph is 80 characters long or less.  It should provide a quick summary of what the method/function is for.  Later paragraphs can elaborate on that further.

1st1 (changes requested):

Doc/library/asyncio-stream.rst
asvetlov: Agree, mentioning *read*, *write* and *read-write* modes make sense.
*Read* and *write* are used for pipes only, sockets are always read-write.

Doc/library/asyncio-stream.rst
asvetlov: Hard question.
It was public in the previous stream implementation.
We have two options:
1. Rename the method to `_set_exception()`.
2. Leave it as is but document that users should never touch it. Useless thing, I think (and a point of API misusage, like `Task.set_exception()`)

1st1: I would like to rename the method to `_set_exception()` in 3.8.  If people come up with useful scenarios for that we can make it public in 3.9.  cc @ambv

Lib/asyncio/streams.py
1st1: Using `*` for parameters improves readability IMO, using backticks isn't necessary (again, IMO).

Doc/library/asyncio-stream.rst
asvetlov: Preparing a PR.

asvetlov: @tirkarthi please remove deprecated by #15762 methods and properties.
`asyncio.Stream` is a new class but it should support backward compatibility for the sake of asyncio subprocess subsystem.

There is no need to introduce deprecated-from-very-beginning API in docs.

Lib/asyncio/streams.py
asvetlov: Ok, I have no strong opinion here.

tirkarthi: I have kept them and some of the other docstrings also have backticks to represent class names.

Doc/library/asyncio-stream.rst
tirkarthi: Thanks, I added a bullet list of the values and a sentence about them.

Doc/library/asyncio-stream.rst
tirkarthi: Thanks, removed `set_exception` and references to deprecated methods too in the docs as per the linked PR.

Doc/library/asyncio-stream.rst
tirkarthi: Thanks, I have added the methods and classes added to the index. Can we also recommend in the deprecated methods about new replacements? Like `open_connection` could have a sentence that `connect` is recommended. I have used pretty much same sentences since they do the same and recommending would avoid confusion over duplicate functions.

Doc/library/asyncio-stream.rst
tirkarthi: I have added a sentence over upgrading the TLS. I have also added similar sentence around `sendfile` above. I couldn't come up with a useful example than calling `connect` and then calling `start_tls` to upgrade SSL in next statement. If there is a more practical one please add in.

Lib/asyncio/streams.py
tirkarthi: Done to here and other docstrings I have added to make sure first sentence have a short description.

Doc/library/asyncio-stream.rst
1st1: Yes! Please do that.

Doc/library/asyncio-api-index.rst
```diff
@@ -132,18 +132,43 @@ High-level APIs to work with network IO.
...
+      -  Establish a TCP connection to send and receive data.
+
     * - ``await`` :func:`open_connection`
       -  Establish a TCP connection.
 
+    * - ``await`` :func:`connect_unix`
+      -  Establish a Unix socket connection to send and receive data.
+
     * - ``await`` :func:`open_unix_connection`
       -  Establish a Unix socket connection.
```
1st1: You can add "(deprecated in favor of the XXX function)" for old APIs.

Doc/library/asyncio-api-index.rst
```diff
@@ -132,18 +132,43 @@ High-level APIs to work with network IO.
...
+      - Establish a connection to :term:`file-like object <file object>` *pipe*
+        to receive data.
+
+    * - :func:`connect_write_pipe`
+      - Establish a connection to :term:`file-like object <file object>` *pipe*
+        to send data.
+
+    * - :class:`Stream`
+      - High-level async/await object to send and receive network data.
+        It includes the API provided by :class:`StreamReader`
```
1st1: I'd rephrase to "Stream is a single object combining APIs of StreamReader and StreamWriter"

Doc/library/asyncio-stream.rst
```diff
@@ -367,13 +366,167 @@ Stream
...
 
    Do not instantiate *Stream* objects directly; use API like :func:`connect`
    and :class:`StreamServer` instead.
 
    .. versionadded:: 3.8
 
+   .. attribute:: mode
+
+      Returns the mode of the stream which is a :class:`StreamMode` value. It could
+      be one of the below :
```
1st1:
```suggestion
      be one of the below:
```

Doc/library/asyncio-stream.rst
```diff
@@ -367,13 +366,167 @@ Stream
...
+
+      * :attr:`StreamMode.READ` - Connection can receive data.
+      * :attr:`StreamMode.WRITE` - Connection can send data.
+      * :attr:`StreamMode.READWRITE` - Connection can send and receive data.
+
+   .. coroutinemethod:: abort()
+
+      Aborts the :ref:`Transport <asyncio-transport>` instance used by the Stream.
```
1st1: This is a relatively low-level description of what's happening here.  I'd use a more high-level description, such as "Aborts the connection immediately, without waiting for the send buffer to drain." (please double check that :))

Doc/library/asyncio-stream.rst
```diff
@@ -367,13 +366,167 @@ Stream
...
+
+          stream.write(data)
+          await stream.drain()
+
+      This is a flow control method that interacts with the underlying
+      IO write buffer.  When the size of the buffer reaches
+      the high watermark, *drain()* blocks until the size of the
+      buffer is drained down to the low watermark and writing can
+      be resumed.  When there is nothing to wait for, the :meth:`drain`
+      returns immediately.
```
1st1, line 424: This is a bit more subtle.  `drain()` is only needed when you do a few `stream.write()` calls.  When you have only one `stream.write()` you can just say `await stream.write()` and not bother with `drain()`.  @asvetlov am I right?

Doc/library/asyncio-api-index.rst
tirkarthi: Thanks, done.

Doc/library/asyncio-api-index.rst
tirkarthi: Thanks, reads better and combines both sentences. Reworded as suggested.

Doc/library/asyncio-stream.rst
tirkarthi: I checked the method and it goes to call the abort method of the transport and waits for the close. I think it doesn't wait for data to be drained.

asvetlov: Yes, `transport.abort()` clears pending write buffer if any,

Doc/library/asyncio-stream.rst
asvetlov: From my perspective, people always should use `await stream.write()` form and never `drain()` explicitly. There is no reason for late buffer overrun detection (and draining) if it can be done earlier.

The only reason why the method exists is backward compatibility for subprocess stdin/stdout/stderr pipes usage.

asvetlov: I think we should reflect it in docs but sorry, I cannot propose exact wording.

Doc/library/asyncio-stream.rst
```diff
@@ -367,13 +366,167 @@ Stream
...
+
+   .. coroutinemethod:: sendfile(file, offset=0, count=None, *, fallback=True)
+
+      Calls :meth:`Stream.drain()` to write to the connection and uses :meth:`loop.sendfile`
```
asvetlov: `drain()` is implementation detail here.

```
Sends a *file* over the stream using an optimized syscall if available.

For other parameters meaning please see :meth:`AbstractEventloop.sendfile`.
```
or something like this.

Doc/library/asyncio-stream.rst
```diff
@@ -367,13 +366,167 @@ Stream
...
+      to send *file* over *transport*.
+
+   .. coroutinemethod:: start_tls(sslcontext, *, server_hostname=None, \
+                                  ssl_handshake_timeout=None)
+
+      Calls :meth:`Stream.drain()` to write to the connection and uses :meth:`loop.start_tls`
```
asvetlov: The same as for `sendfile`: the method upgrades a connection to TLS encrypted one.

Doc/library/asyncio-stream.rst
```diff
@@ -367,13 +366,167 @@ Stream
...
+      is left in the internal buffer and can be read again.
+
+      If EOF is reached before the complete separator is found,
+      an :exc:`IncompleteReadError` exception is raised, and the internal
+      buffer is reset.  The :attr:`IncompleteReadError.partial` attribute
+      may contain a portion of the separator.
+
+   .. method:: write(data)
+
+      The method attempts to write the *data* to the underlying socket immediately.
```
asvetlov, line 514: I think here and below for methods that can be used both with `await` and without it we should focus on `await` version first.  E.g. `await stream.write(data)` is more idiomatic than `stream.write(data); await stream.drain()`.

The first documentation line can be `Write *data* to the underlying socket; wait until the data is sent.`.

Thinking more, I believe we need to document two functions with the same name:
```
.. comethod:: write(data)

   Write *data* to the underlying socket; wait until the data is sent, e.g.::

      await stream.write(b'data')

.. method:: write(data)

   Long description here that teaches how combite `write()` with `drain()` and mentions that usually people should prefer `await stream,write()` version.
```
I'm curious is it supported by sphinx well or not? What version is a target for `:meth:Stream.write` expression?

tirkarthi: It is supported. Here is a first pass version of this. The `Stream.write` target is set to the coroutine and the normal write method cannot be hyperlinked. We also do some explanation on the drain method so I have the docs as below : 

```rst
   .. coroutinemethod:: write(data)

      Write *data* to the underlying socket; wait until the data is sent, e.g.::

         await stream.write(data)

   .. method:: write(data)

      The method attempts to write the *data* to the underlying socket immediately.
      If that fails, the data is queued in an internal write buffer until it can be
      sent. :meth:`drain` can be used to flush the underlying buffer once writing is
      available::

         stream.write(data)
         await stream.drain()

      It is recommended to directly await on the `write()` method instead::

         await stream.write(data)
```

Screen Shot 2019-09-11 at 10 59 30 am

asvetlov: Looks nice!

tirkarthi:
I have pushed the changes to make sure `await stream.write(data)` as recommended with `stream.write(data); await stream.drain()` also documented with a method. I think the only detail pending now is over how multiple `stream.write` can be buffered to use `stream.drain` later as below that should be noted in the `stream.drain` docs.

https://github.com/python/cpython/pull/14488#discussion_r322740985

> This is a bit more subtle. drain() is only needed when you do a few stream.write() calls. When you have only one stream.write() you can just say await stream.write() and not bother with drain(). @asvetlov am I right?

asvetlov:
> I think the only detail pending now is over how multiple stream.write can be buffered to use stream.drain later as below that should be noted in the stream.drain docs.

The interesting part is that multiple sync writes and single drain after doesn't help with buffering. Instead, it may consume more memory that expected.  

asyncio streams already have low and high watermarks for write buffer size control. If kernel write buffer is full the data is collected in asyncio internal write buffer. If the internal buffer size is greater than high watermark the writing is paused until the size of the buffer doesn't fall to less than the low watermark. After that write is resumed and keeps resumed until the high watermark is reached again.

So, data is effectively buffered even if `await stream.write()` notation is always used, `drain()` after muptiple sync `write()` calls doesn't make any value.

I doubt if we need document it here, in a method reference.  Maybe a separate asyncio section about buffering makes sense? 
The section is not related to streams itself, low-level `protocol.pause_writing()` / `protocol.resume_writing()` supports the same logic.

1st1:
Andrew, reading all of this makes me think about soft-deprecating non-await `write()` and `drain()` in the documentation (it will probably take us many releases before we can actually deprecate them, if ever).  This will actually encourage people to just write `await stream.write()` and not touch `drain()` etc.

asvetlov:
> Andrew, reading all of this makes me think about soft-deprecating non-await write() and drain() in the documentation (it will probably take us many releases before we can actually deprecate them, if ever). This will actually encourage people to just write await stream.write() and not touch drain() etc.

@1st1 I told you at US PyCon that non-awaitable methods are useless and should be deprecated eventually. You had doubts, so we decided to support both.

I totally agree with the deprecation strategy.

Does soft-deprecation means `..deprecated:` section or completely removal these API from docs?
The second is more aggressive, sure. But it simplifies documentation writing and (much more important) reading a lot.

@1st1 what do you prefer?

1st1:
> Does soft-deprecation means ..deprecated: section or completely removal these API from docs?
The second is more aggressive, sure. But it simplifies documentation writing and (much more important) reading a lot.

I'd do it in a very soft way in 3.8/3.9; basically we just warn users that they are deprecated in the docs and that's it.

asvetlov:
Ok, I agree with any decision.
@tirkarthi please add `.. deprecation::` sections.

Sorry for the endless list of notes. Writing a new stream API took me weeks, documenting it takes a long as well.

tirkarthi:
Added deprecated notes to `Stream.drain` and method `Stream.write` to recommending usage of `await Stream.write` as a coroutine to be awaited. There is also `StreamWriter.write` and `StreamWriter.writelines` but they are already recommended to use `stream.write` and contain examples on using it in 3.7 where await is not supported so I have not added deprecations there.

No problem, I got to know better about the API while documentation and also closed some old and outdated issues related to Streams on the tracker with the PR. Appreciate your help on word formation and care to perfect user facing documentation.

asvetlov (changes requested):
I like the current text, looks good.
The last nitpick: please provide an order to Stream methods.
I think alphabetical sorting is better than the current grouping.

tirkarthi:
Good, done. The current order is alphabetic sort as below with attribute mode at first

```
mode

abort
at_eof
can_write_eof
close
drain
get_extra_info
is_closing
read
readexactly
readline
readuntil
sendfile
start_tls
wait_closed
write coroutine
write method
writelines
write_eof

``

asvetlov (approved):
Well done!
Thank you very much!

asvetlov:
@1st1 please review again. The PR is ready for merging from my perspective

tirkarthi:
Thank you very much Andrew and Yury. I can create a separate PR if there are additional improvements.

asvetlov:
Oops. Sorry, @1st1 
I've added automerge label with the hope that the merge will be blocked until your approval.
@miss-islington has own logic for this situation.

Anyway, please feel free to express your objection if any, @tirkarthi and I can fix them in the following PR quickly.

PR #14489, opened 2019-06-30
Separate import for each module

hansrajdas:
Used separate imports for os and sys. This change as per PEP8 import guidelines: https://www.python.org/dev/peps/pep-0008/#imports


asvetlov:
We usually don't accept PRs for style changing only.
It pollutes git history without strong positive effect.

PR #14490, opened 2019-07-01
bpo-37424: Avoid a hang in subprocess.run timeout output capture

gpshead:
Fixes a possible hang when using a timeout on `subprocess.run()` while
capturing output.  If the child process spawned its own children or otherwise
connected its stdout or stderr handles with another process, we could hang
after the timeout was reached and our child was killed when attempting to read
final output from the pipes.

https://bugs.python.org/issue37424

### DIFF
--- a/Lib/subprocess.py
+++ b/Lib/subprocess.py
@@ -752,11 +752,6 @@ class Popen(object):
         if not isinstance(bufsize, int):
             raise TypeError("bufsize must be an integer")
 
-        # If True, kill/terminate/send_signal will send the signal to
-        # os.getpgid(self.pid) on platforms with that concept IF the
-        # group is not our own process group.
-        self.__signal_process_group = shell and hasattr(os, 'getpgid')
-
         if _mswindows:
             if preexec_fn is not None:
                 raise ValueError("preexec_fn is not supported on Windows "
@@ -1933,15 +1928,7 @@ class Popen(object):
             """Send a signal to the process."""
             # Skip signalling a process that we know has already died.
             if self.returncode is None:
-                if self.__signal_process_group:
-                    pgid = os.getpgid(self.pid)
-                    if pgid == os.getpgid(os.getpid()):
-                        # Never killpg our own process group.
-                        os.kill(self.pid, sig)
-                    else:
-                        os.killpg(pgid, sig)
-                else:
-                    os.kill(self.pid, sig)
+                os.kill(self.pid, sig)
 
         def terminate(self):
             """Terminate the process with SIGTERM
--- a/Lib/subprocess.py
+++ b/Lib/subprocess.py
@@ -491,7 +491,13 @@ def run(*popenargs,
             stdout, stderr = process.communicate(input, timeout=timeout)
         except TimeoutExpired:
             process.kill()
-            stdout, stderr = process.communicate()
+            # The timeout here is to avoid waiting around _forever_ if
+            # the kill failed to deal with all processes holding the
+            # output file handles open.  Otherwise we could hang until
+            # those processes terminate before continuing with our timeout.
+            # See https://bugs.python.org/issue37424.
+            stdout, stderr = process.communicate(
+                    timeout=_get_cleanup_timeout(timeout))
             raise TimeoutExpired(process.args, timeout, output=stdout,
                                  stderr=stderr)
         except:  # Including KeyboardInterrupt, communicate handled that.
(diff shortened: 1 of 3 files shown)


gpshead:
If an API change is made to make the default behavior easier for people in the future, that part won't get backported to 3.7.  3.8 only.

Lib/subprocess.py
```diff
@@ -505,6 +511,11 @@ def run(*popenargs,
     return CompletedProcess(process.args, retcode, stdout, stderr)
 
 
+def _get_cleanup_timeout(timeout):  # For test injection.
+    if timeout is not None:
+        return timeout/20
```
giampaolo: why timeout / 20? (instead of just returning original timeout)

gpshead:
The original full timeout has already elapsed and expired, SIGKILL was
sent, we don't want to wait much longer.

Lib/subprocess.py
gpshead: The original full timeout has already elapsed and expired, SIGKILL was
sent, we don't want to wait much longer.

Lib/subprocess.py
```diff
@@ -491,7 +491,13 @@ def run(*popenargs,
             stdout, stderr = process.communicate(input, timeout=timeout)
         except TimeoutExpired:
             process.kill()
```
vstinner, line 493: Why don't we close all pipes here? Closing pipes + .wait() should be enough, no? SIGKILL cannot be ignored on Unix (I don't know for Windows).

gpshead: Agreed this implementation is too complex.  That sounds tempting and simple.  Trying that, I found that it caused the capture_output_timeout tests to fail as we no longer collected the output and attached it to the Timeout exception.  But that led to some code inspection... we're already gathering the output within the original _communicate() on POSIX and can attach everything we have in the _original_ Timeout exception so that we don't need to call communicate() again after kill() at all in this path.

Windows collects output in a different manner so it still needs it, but this bug is POSIX only so I've reworked this to simplify it.  The communicate call for output gathering becomes Windows only.  I'm relying on the CI systems to do my Windows testing.

Lib/subprocess.py
```diff
@@ -1910,7 +1933,15 @@ def send_signal(self, sig):
             """Send a signal to the process."""
             # Skip signalling a process that we know has already died.
             if self.returncode is None:
-                os.kill(self.pid, sig)
+                if self.__signal_process_group:
```
gpshead: Should I also undo the addition of this magic when shell=True process group killing logic?

The bug was about a hang after the timeout should've expired.  That hang will no longer happen and we will return.  But this does potentially leave the shell's child processes running on some posix platforms and child process configurations.  Likely not what the caller ever intended.

People running on stable Python releases who need to ensure the process group's children have died already need to do their own post-timeout getpgid() + killpg() logic dance similar to what is in here no matter what.  _This_ logic feels like a feature that I'd do for 3.8+ only.

Adding explicit process group killing features to the kill APIs as some of the dup'ed issues have desired is also squarely a feature request, likely a good idea, but belongs in its own set of PRs as it won't be backported as a bugfix.

gpshead: okay, i've convinced myself.  process group related logic belongs in its own PR.  people need to manually specify start_new_session for this logic to be meaningful regardless.  and for uniformity there's a question of what makes sense on windows as it has its own notion of process groups but *not* a start_new_session concept.

Lib/subprocess.py
```diff
@@ -489,11 +489,20 @@ def run(*popenargs,
...
+            if _mswindows:
+                # Windows accumulates the output in a single blocking
+                # read() call run on child threads, with the timeout
+                # being done in a join() on those threads.  communicate()
+                # _after_ kill() is required to collect that and add it
+                # to the exception.
+                exc.stdout, exc.stderr = process.communicate()
+            else:
+                # POSIX _communicate already populated the output so
+                # far into the TimeourExpired exception.
```
Yhg1s: Timeour -> Timeout

Lib/test/test_subprocess.py
```diff
@@ -1564,6 +1565,22 @@ def test_stderr_with_capture_output_arg(self):
...
+        before_secs = time.monotonic()
+        try:
+            subprocess.run('sleep 3', shell=True, timeout=0.1,
+                           capture_output=True)  # New session unspecified.
+        except subprocess.TimeoutExpired as exc:
+            after_secs = time.monotonic()
+            stacks = traceback.format_exc()  # assertRaises doesn't give this.
+        else:
+            self.fail("TimeoutExpired not raised.")
+        self.assertLess(after_secs - before_secs, 1.5,
```
Yhg1s, line 1584: This feels like it would be sensitive to heavily loaded machines... Maybe add a comment to it to make it easier to identify if that is the case?

PR #14491, opened 2019-07-01
bpo-32631: IDLE: Enable zzdummy extension module

csabella:
* Complete functionality to add or remove prefix string to each line.
* Add comments and docstrings to zzdummy.
* Add unittests in test_zzdummy.
* Add help section for extensions tab to configdialog.

https://bugs.python.org/issue32631


terryjreedy (approved):
Although the behavior is not exactly what I intended (see issue), the example now works as an example.  Approved with my minor doc changes.  Will merge when retest passes.

terryjreedy:
The Travis and Pipelines failures are only due to bad markup in the News item for bpo-40810, just fixed by #24110 and backports.  I restarted Travis and it passed.  Cannot restart Pipelines, so ignoring.

PR #14492, opened 2019-07-01
Docs: Clarify BNF description in "introduction.rst"

aeros:
The text below describes the change proposal to the "Notation" section of "Doc/reference/introduction.rst".

In reference to the section describing the first line of BNF notation, "name: lc_letter (lc_letter | "\_")*", the section states "a name is an lc_letter followed by a sequence of zero or more lc_letters and underscores" there is an issue with the usage of "and" in "lc_letters and underscores". While this usage of "and" is grammatically correct in English context, if the reader interprets it literally in the context of BNF notation or computer science, "lc_letters and underscores" would translate to (lc_letter & "\_"), which would entirely change the meaning of that line. 

Therefore, I would suggest changing "lc_letters and underscores" to "lc_letters or underscores". This ensures the reader clearly understands the logical disjunction operator is being utilized and does not misunderstand "and" to be the conjunction operator since the English word "and" can mean either of them, depending on the context.

### DIFF
--- a/Doc/reference/introduction.rst
+++ b/Doc/reference/introduction.rst
@@ -98,7 +98,7 @@ notation.  This uses the following style of definition:
    lc_letter: "a"..."z"
 
 The first line says that a ``name`` is an ``lc_letter`` followed by a sequence
-of zero or more ``lc_letter``\ s or underscores.  An ``lc_letter`` in turn is
+of zero or more ``lc_letter``\ s and/or underscores.  An ``lc_letter`` in turn is
 any of the single characters ``'a'`` through ``'z'``.  (This rule is actually
 adhered to for the names defined in lexical and grammar rules in this document.)
 
--- a/Doc/reference/introduction.rst
+++ b/Doc/reference/introduction.rst
@@ -98,7 +98,7 @@ notation.  This uses the following style of definition:
    lc_letter: "a"..."z"
 
 The first line says that a ``name`` is an ``lc_letter`` followed by a sequence
-of zero or more ``lc_letter``\ s and underscores.  An ``lc_letter`` in turn is
+of zero or more ``lc_letter``\ s or underscores.  An ``lc_letter`` in turn is
 any of the single characters ``'a'`` through ``'z'``.  (This rule is actually
 adhered to for the names defined in lexical and grammar rules in this document.)
 


taleinat:
Thanks for looking into this!

"lc_letters or underscores" could also be potentially confusing, couldn't it? One might perhaps think that there must be only lc_letters or only underscores, but not a mix of both.

Perhaps we should use "lc_letters and/or underscores" for maximum clarity and preciseness?

aeros167:
@taleinat Thanks for the feedback, I was thinking primarily in the context of the logical operators rather than the English usage of "or". The usage of "and/or" would probably be the most accurate way to simultaneously satisfy both of the contexts. I'll add your suggestion to the PR.

aeros167:
@taleinat Would you mind reviewing this submission? I made the changes that you recommended.

taleinat (approved):
LGTM, but this will wait for another core dev review before going in.

aeros167:
From the documentation section of the [experts index](https://devguide.python.org/experts/#miscellaneous), cc @ezio-melotti, @merwok, @JulienPalard, @willingc.

aeros167:
@taleinat Since this PR still hasn't received any reviews a month after I pinged the documentation experts, could we go forward with an auto-merge label? According to the [devguide](https://devguide.python.org/pullrequest/#reviewing) the appropriate procedure is:

> If you don’t get a response within a week after pinging the issue, then you can try emailing python-dev@python.org to ask for someone to review your pull request.

However, I think this PR might be a bit too minimal to justify creating a new thread on the ML.

I don't at all blame anyone in particular for this issue, but in the current system, it can be very difficult to get minor documentation PRs approved which make improvements to existing sections. I've tried to assist with this issue myself by providing feedback on other documentation PRs, but that doesn't always solve the problem.

Would it be possible to recruit a new core developer specifically for reviewing and approving smaller documentation PRs? Carol is quite busy as a SC member and the other three listed documentation experts are only sporadically active or have not been recently active, see [ezio-melotti](https://github.com/ezio-melotti?tab=overview&from=2019-07-01&to=2019-07-31&org=python), [JulienPalard](https://github.com/JulienPalard?tab=overview&from=2019-07-01&to=2019-07-31&org=python), and  [merwork](https://github.com/merwok?tab=overview&from=2019-07-01&to=2019-07-31&org=python).

merwok:
>  While this usage of "and" is grammatically correct in English context

IMO this is the important point.

> if the reader interprets it literally in the context of BNF notation or computer science

Such a reader would read the BNF directly and not be confused by the plain text exegesis, no?  :slightly_smiling_face: 

I’m -0, I don’t think there is a problem here (unlike `or` which depending on the language is inclusive or exclusive so benefits from clarification, I don’t think people will read the `and` and think they MUST include from both categories), but the PR doesn’t make the text more cumbersome or worse, so fine if someone merges it.

(Sorry you have to wait and discuss so much for such a small thing!  Good that you have other PRs merged for more substantive changes)

aeros167:
Thanks for the feedback. Also, I have to admit, that timing was quite ironic. (:

>Such a reader would read the BNF directly and not be confused by the plain text exegesis, no? slightly_smiling_face

Good point, I may have been taking the "and" a bit more literally than most readers would. Also, I think at the time I made this PR, I might have been looking a bit too hard to find meaningful improvements. I've gained a decent amount of experience since then, and have found more impactful issues to work on. But, I've always had the mentality that it's far better to take the risk of making an unnecessary suggestion than it is to potentially miss out on a meaningful one. 

>(Sorry you have to wait and discuss so much for such a small thing! Good that you have other PRs merged for more substantive changes)

No problem. So far from my experiences thus far in both submitting and reviewing PRs, there are far more PRs than there are people to review them.

However, it does seem that this issue disproportionately affects PRs that exclusively involve documentation changes. I think there might be a significantly greater deficit of active documentation experts in comparison to other categories, especially when the volume of PRs involving documentation changes is taken into consideration.

aeros167:
Thanks again for the feedback @taleinat and @merwok. Since it looks like this change probably isn't needed, I'll close the PR.

PR #14494, opened 2019-07-01
[3.8] bpo-37428: Don't set PHA verify flag on client side (GH-14494)

tiran:
SSLContext.post_handshake_auth = True no longer sets
SSL_VERIFY_POST_HANDSHAKE verify flag for client connections. Although the
option is documented as ignored for clients, OpenSSL implicitly enables cert
chain validation when the flag is set.

https://bugs.python.org/issue37428
(cherry picked from commit f0f5930ac88482ef896283db5be9b8d508d077db)

https://bugs.python.org/issue37428


tiran:
GH limits the title to 72 chars and didn't let me merge this PR with commit message ``[3.8] bpo-37428: Don't set PHA verify flag on client side (GH-14421) (GH-14493)``.

PR #14497, opened 2019-07-01
bpo-37462: Default for "top" argument in "os.walk()"

aeros:
Currently, os.walk() can't be utilized without specifying a starting directory. This change uses the current working directory as the starting point for os.walk() when one is not specified. Reasoning for this change is described in further detail on the issue page. 

https://bugs.python.org/issue37462

### DIFF
--- a/Lib/os.py
+++ b/Lib/os.py
@@ -275,7 +275,7 @@ def renames(old, new):
 
 __all__.extend(["makedirs", "removedirs", "renames"])
 
-def walk(top, topdown=True, onerror=None, followlinks=False):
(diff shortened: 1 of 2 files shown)


taleinat:
Marking with `DO-NOT-MERGE` since the opinion on the bug tracker is currently that this change shouldn't be made.

aeros167:
@taleinat After reading over the criticisms on the bug tracker, I've decided to close the PR, as I'm finding myself in agreement with them. Thanks again for the constructive feedback over there.

PR #14499, opened 2019-07-01
bpo-37463: match_hostname requires quad-dotted IPv4

tiran:
ssl.match_hostname() no longer accepts IPv4 addresses with additional text
after the address and only quad-dotted notation without trailing
whitespaces. Some inet_aton() implementations ignore whitespace and all data
after whitespace, e.g. '127.0.0.1 whatever'.

Short notations like '127.1' for '127.0.0.1' were already filtered out.

The bug was initially found by Dominik Czarnota and reported by Paul Kehrer.

https://bugs.python.org/issue37463

### DIFF
--- a/Lib/ssl.py
+++ b/Lib/ssl.py
@@ -327,12 +327,22 @@ def _inet_paton(ipname):
     Supports IPv4 addresses on all platforms and IPv6 on platforms with IPv6
     support.
     """
-    # inet_aton() also accepts strings like '1'
-    if ipname.count('.') == 3:
-        try:
-            return _socket.inet_aton(ipname)
-        except OSError:
-            pass
+    # inet_aton() also accepts strings like '1', '127.1', some also trailing
+    # data like '127.0.0.1 whatever'.
+    try:
+        addr = _socket.inet_aton(ipname)
+    except OSError:
+        # not an IPv4 address
+        pass
+    else:
+        if _socket.inet_ntoa(addr) == ipname:
+            # only accept injective ipnames
+            return addr
+        else:
+            # refuse for short IPv4 notation and additional trailing data
+            raise ValueError(
+                "{!r} is not a quad-dotted IPv4 address.".format(ipname)
+            )
 
     try:
         return _socket.inet_pton(_socket.AF_INET6, ipname)
@@ -346,14 +356,15 @@ def _inet_paton(ipname):
     raise ValueError("{!r} is not an IPv4 address.".format(ipname))
 
 
-def _ipaddress_match(ipname, host_ip):
+def _ipaddress_match(cert_ipaddress, host_ip):
     """Exact matching of IP addresses.
 
     RFC 6125 explicitly doesn't define an algorithm for this
     (section 1.7.2 - "Out of Scope").
     """
-    # OpenSSL may add a trailing newline to a subjectAltName's IP address
-    ip = _inet_paton(ipname.rstrip())
+    # OpenSSL may add a trailing newline to a subjectAltName's IP address,
+    # commonly woth IPv6 addresses. Strip off trailing \n.
+    ip = _inet_paton(cert_ipaddress.rstrip())
     return ip == host_ip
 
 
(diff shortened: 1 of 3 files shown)


Lib/ssl.py
```diff
@@ -327,12 +327,20 @@ def _inet_paton(ipname):
...
-        except OSError:
-            pass
+    # inet_aton() also accepts strings like '1', '127.1', or
+    # '127.0.0.1 whatever'.
+    try:
+        addr = _socket.inet_aton(ipname)
+    except OSError:
+        pass
+    else:
+        if _socket.inet_ntoa(addr) == ipname:
```
zooba, line 338: Any reason not to `ipname.rstrip()` here for the comparison? Does whitespace matter?

zooba: I just notice that you're explicitly testing trailing whitespace below.

Lib/ssl.py
```diff
@@ -327,12 +327,20 @@ def _inet_paton(ipname):
...
+            # only accept injective ipnames
+            return addr
+        else:
+            raise ValueError(
+                "{!r} is not a quad-dotted IPv4 address.".format(ipname)
```
zooba, line 344: FWIW, on Windows if you pass an invalid address you get `OSError: illegal IP address string passed to inet_aton`. Not sure if that's a Python error

Lib/ssl.py
```diff
@@ -327,12 +327,20 @@ def _inet_paton(ipname):
     Supports IPv4 addresses on all platforms and IPv6 on platforms with IPv6
     support.
```
vstinner, line 328: _ipaddress_match() is documented as "Exact matching of IP addresses." but is explicitly strips trailing spaces. Maybe just add a sentence to explain that in its docstring?

_inet_paton() disallow leading and trailing whitespaces: maybe explain that in its docstring?

_inet_paton() allows short form of IPv6 addresses, but not for IPv4 address. Is it a deliberate choice? Again, maybe document it in the docstring.

Lib/ssl.py
vstinner, line 344: nitpick: usually, Python error messages don't end with a dot. (Same comment for the 3 error messages.)

Lib/ssl.py
tiran: ``rstrip()`` would not deal with invalid input like ``127.0.0.1 whatever``, ``127.0.0.1\tinvalid``, or other embedded white spaces.

Lib/ssl.py
tiran: Some implementations of ``inet_aton`` are strict and don't allow trailing data. glibc's ``inet_aton`` is more lean and ignores trailing data if the first character is some sort of white space.

Lib/ssl.py
tiran: > I just notice that you're explicitly testing trailing whitespace below.

The ``rstrip`` call in ``_ipaddress_match`` is necessary because OpenSSL includes a trailing new line for IPv6 addresses. The code removes trailing white space in SAN ``IP Address`` items of decoded certs, not of the user supplied ``hostname`` argument.

Lib/ssl.py
tiran: I don't think it is useful to add more doc strings to internal helper methods for a deprecated feature. ``ssl.match_hostname`` is deprecated and I plan to drop it in 3.9 or 3.10.

The ``strip`` in ``_ipaddress_match()`` is an implementation detail that works around a quirk in OpenSSL. The IP address may sometimes contain a newline. IIRC IPv6 addresses only.

Yes, that is deliberate. Short notation of IPv4 addresses is an uncommon and rarely used feature. However short notation of IPv6 is pretty much the default. The behavior also reflects the behavior of the old implementation with ipaddress module.

Lib/ssl.py
tiran: I'm following the style of the surrounding code. The exception messages are full sentences and end with a full stop.

tiran:
I have added a few more comments to the code. I also changed one argument to make it more obvious what the function is doing.

@vstinner The other exceptions use the same style: full sentence with a full stop.

https://github.com/python/cpython/blob/7cb9204ee1cf204f6f507d99a60f7c5bb359eebb/Lib/ssl.py#L275-L346

ned-deily (approved):

PR #14503, opened 2019-07-01
bpo-37209: Add pickle entry for 3.8 whatsnew

pierreglaser:
@pitrou any re-wording is welcome.

https://bugs.python.org/issue37209


pitrou (approved):
LGTM

PR #14504, opened 2019-07-01
bpo-37467: Fix PyErr_Display() for bytes filename

vstinner:
Fix sys.excepthook() and PyErr_Display() if a filename is a bytes
string. For example, for a SyntaxError exception where the filename
attribute is a bytes string.

https://bugs.python.org/issue37467

### DIFF
--- a/Lib/test/test_sys.py
+++ b/Lib/test/test_sys.py
@@ -1,5 +1,7 @@
-import unittest, test.support
+import unittest
+import test.support
 from test.support.script_helper import assert_python_ok, assert_python_failure
+from test import support
 import sys, io, os
 import struct
 import subprocess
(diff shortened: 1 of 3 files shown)


vstinner:
The test fails on the CI, but it pass locally on my Linux laptop and in my Windows VM. I don't understand why.

Example of failure on AppVeyor:

```
FAIL: test_excepthook_bytes_filename (test.test_sys.SysModuleTest)
  File "C:\projects\cpython\lib\test\test_sys.py", line 91, in test_excepthook_bytes_filename
    self.assertIn("""  File "b'bytes_filename'", line 123\n""", err)
AssertionError: '  File "b\'bytes_filename\'", line 123\n' not found in 'Traceback (most recent call last):\n  File "C:\\projects\\cpython\\lib\\test\\test_sys.py", line 85, in test_excepthook_bytes_filename\n    raise SyntaxError("msg", (b"bytes_filename", 123, 0, "text"))\n\n'
```

PR #14506, opened 2019-07-01
bpo-10945: Drop support for bdist_wininst on non-Windows systems

hroncok:
bdist_wininst depends on MBCS codec, unavailable on non-Windows,
and bdist_wininst have not worked since at least Python 3.2, possibly
never on py3k.

Here we document that bdist_wininst is only supported on Windows
and we mark it _unsupported otherwise to skip tests.

Distributors of Python 3 can now safely drop the bdist_wininst .exe files
without the need to skip bdist_wininst related tests.

https://bugs.python.org/issue10945

### DIFF
--- a/Doc/distutils/builtdist.rst
+++ b/Doc/distutils/builtdist.rst
@@ -315,7 +315,8 @@ or the :command:`bdist` command with the :option:`!--formats` option::
 
 If you have a pure module distribution (only containing pure Python modules and
 packages), the resulting installer will be version independent and have a name
-like :file:`foo-1.0.win32.exe`.
+like :file:`foo-1.0.win32.exe`. Note that creating ``wininst`` binary
+distributions in only supported on Windows systems.
 
 If you have a non-pure distribution, the extensions can only be created on a
 Windows platform, and will be Python version dependent. The installer filename
@@ -340,9 +341,6 @@ version number.  This can be changed to another text by using the
 
 The installer file will be written to the "distribution directory" --- normally
 :file:`dist/`, but customizable with the :option:`!--dist-dir` option.
-
-   .. versionchanged:: 3.9
-      Creating ``wininst`` binary distributions in only supported on Windows.
 
 .. _cross-compile-windows:
 
--- a/Doc/distutils/builtdist.rst
+++ b/Doc/distutils/builtdist.rst
@@ -315,8 +315,7 @@ or the :command:`bdist` command with the :option:`!--formats` option::
 
 If you have a pure module distribution (only containing pure Python modules and
 packages), the resulting installer will be version independent and have a name
-like :file:`foo-1.0.win32.exe`.  These installers can even be created on Unix
-platforms or Mac OS X.
+like :file:`foo-1.0.win32.exe`.
 
 If you have a non-pure distribution, the extensions can only be created on a
 Windows platform, and will be Python version dependent. The installer filename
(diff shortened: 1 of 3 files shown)


Doc/distutils/builtdist.rst
```diff
@@ -342,6 +341,9 @@ version number.  This can be changed to another text by using the
 The installer file will be written to the "distribution directory" --- normally
 :file:`dist/`, but customizable with the :option:`!--dist-dir` option.
 
+   .. versionchanged:: 3.9
```
vstinner: Please remove this note.

I don't see why this change should be Python 3.9 only: mbcs encoding was never available on platforms other than Windows. If we make a change, I suggest to make it in 3.7, 3.8 and master branches.

Lib/distutils/command/bdist_wininst.py
```diff
@@ -55,6 +56,8 @@ class bdist_wininst(Command):
     boolean_options = ['keep-temp', 'no-target-compile', 'no-target-optimize',
                        'skip-build']
 
+    _unsupported = platform.system() != "Windows"
```
vstinner: I suggest:

```
     # bpo-10945: bdist_wininst requires mbcs encoding which is only available on Windows
     _unsupported = (sys.platform != "win32")
```

Misc/NEWS.d/next/Windows/2019-07-01-12-38-48.bpo-10945.s0YBHG.rst
```diff
@@ -0,0 +1,2 @@
+Officially drop support for creating bdist_wininst installers on non-Windows
+systems.
```
vstinner, line 2: Technically, it wasn't supported previously, so I don't think that we even have to document the change. But I'm fine with keeping this changelog entry if you prefer.

vstinner (changes requested):

Doc/distutils/builtdist.rst
hroncok: Gladly :rocket:

hroncok:

hroncok:
An idea: Don't install the relevant exe files on non-Windows?

Lib/distutils/command/bdist_wininst.py
```diff
@@ -4,6 +4,7 @@
 exe-program."""
 
 import sys, os
+import platform
```
vstinner: You can now removed this import which became unused.

hroncok: good catch

Doc/distutils/builtdist.rst
```diff
@@ -315,8 +315,8 @@ or the :command:`bdist` command with the :option:`!--formats` option::
 
 If you have a pure module distribution (only containing pure Python modules and
 packages), the resulting installer will be version independent and have a name
-like :file:`foo-1.0.win32.exe`.  These installers can even be created on Unix
-platforms or Mac OS X.
+like :file:`foo-1.0.win32.exe`. Note that creating ``wininst`` binary
+distributions in only supported on Windows systems.
```
vstinner, line 319: That sounds better to me, thanks for the update!

Lib/distutils/command/bdist_wininst.py
hroncok: done

vstinner:
> An idea: Don't install the relevant exe files on non-Windows?

I would prefer to have a separated PR for that.

vstinner (approved):
LGTM. Thanks for the update.

PR #14513, opened 2019-07-01
bpo-37471: mmap module adding FreeBSD specific flag into the constants

ghost:
https://bugs.python.org/issue37471

### DIFF
--- a/Doc/library/mmap.rst
+++ b/Doc/library/mmap.rst
@@ -328,6 +328,7 @@ MADV_* Constants
           MADV_NOHUGEPAGE
           MADV_DONTDUMP
           MADV_DODUMP
+          MADV_CONCEAL
           MADV_FREE
           MADV_NOSYNC
           MADV_AUTOSYNC
--- a/Doc/library/mmap.rst
+++ b/Doc/library/mmap.rst
@@ -328,7 +328,6 @@ MADV_* Constants
           MADV_NOHUGEPAGE
           MADV_DONTDUMP
           MADV_DODUMP
-          MADV_CONCEAL
           MADV_FREE
           MADV_NOSYNC
           MADV_AUTOSYNC
--- a/Doc/library/mmap.rst
+++ b/Doc/library/mmap.rst
@@ -81,7 +81,11 @@ To map anonymous memory, -1 should be passed as the fileno along with the length
    private copy-on-write mapping, so changes to the contents of the mmap
    object will be private to this process, and :const:`MAP_SHARED` creates a
    mapping that's shared with all other processes mapping the same areas of
-   the file.  The default value is :const:`MAP_SHARED`.
+   the file. :const:`MAP_ALIGNED_SUPER` (FreeBSD only) uses super pages
+   feature, it is enabled by default in most of architectures and exists
+   since 2008.
+   The default value is :const:`MAP_SHARED`.
+   See :ref:`MAP_* constants <mmap-constants>`.
 
    *prot*, if specified, gives the desired memory protection; the two most
    useful values are :const:`PROT_READ` and :const:`PROT_WRITE`, to specify
@@ -307,6 +311,24 @@ To map anonymous memory, -1 should be passed as the fileno along with the length
       the mmap was created with :const:`ACCESS_READ`, then writing to it will
       raise a :exc:`TypeError` exception.
 
+.. _mmap-constants:
+
+MAP_* Constants
++++++++++++++++
+
+.. data:: MAP_ANONYMOUS
+          MAP_ANON
+          MAP_SHARED
+          MAP_PRIVATE
+          MAP_DENYWRITE
+          MAP_EXECUTABLE
+          MAP_ALIGNED_SUPER
+          MAP_CONCEAL
+
+   .. versionadded:: 3.9
+      Add :data:`MAP_ALIGNED_SUPER` constant.
+      Add :data:`MAP_CONCEAL` constant,
+
 .. _madvise-constants:
 
 MADV_* Constants
@@ -341,3 +363,4 @@ MADV_* Constants
    Availability: Systems with the madvise() system call.
 
    .. versionadded:: 3.8
+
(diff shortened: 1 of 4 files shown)


vstinner (commented):
Would it be possible to document the constant in https://docs.python.org/dev/library/mmap.html ?

It seems like other MAP_xxx constants are not documented yet and so should be documented as well.

dcarlier-afilias:
They are (shortly) in the flags section. Ok for adding the new constant too.

vstinner:
The minimum doc should be a list of contants, like: https://docs.python.org/dev/library/os.html#os.O_RDONLY

dcarlier-afilias:
Fair enough ! ought to be before madvise's I think.

Doc/library/mmap.rst
```diff
@@ -81,7 +81,11 @@ To map anonymous memory, -1 should be passed as the fileno along with the length
    private copy-on-write mapping, so changes to the contents of the mmap
    object will be private to this process, and :const:`MAP_SHARED` creates a
    mapping that's shared with all other processes mapping the same areas of
-   the file.  The default value is :const:`MAP_SHARED`.
+   the file. :const:`MAP_ALIGNED_SUPER` (FreeBSD only) allows to map to
+   pages of 2MB (x86) or so, it is enabled by default, at boot time,
```
vstinner:
```suggestion
   pages of 2 MiB (x86) or so, it is enabled by default, at boot time,
```

Is it x86 (32-bit) or x86-64 (64-bit)?

Misc/NEWS.d/next/Library/2019-07-01-14-27-17.bpo-37471._EFODL.rst
```diff
@@ -0,0 +1 @@
+Adding new 'MAP_ALIGNED_SUPER' FreeBSD constant.
```
vstinner:
```suggestion
Adding new :data:`mmap.MAP_ALIGNED_SUPER` FreeBSD constant.
```

Doc/library/mmap.rst
```diff
@@ -307,6 +311,19 @@ To map anonymous memory, -1 should be passed as the fileno along with the length
...
+MAP_* Constants
++++++++++++++++
+
+.. data:: MAP_ANONYMOUS
+          MAP_ANON
+          MAP_SHARED
+          MAP_PRIVATE
+          MAP_DENYWRITE
+          MAP_EXECUTABLE
+          MAP_ALIGNED_SUPER
```
vstinner, line 325: You have to document that MAP_ALIGNED_SUPER is new, example:

```
.. versionadded:: 3.9
   Add :data:`MAP_ALIGNED_SUPER` constant.
```

Doc/library/mmap.rst
ghost: both.

vstinner: Write maybe (x86, x86-64) in this case... But I'm not sure if Python documentation should go too far in the description, since FreeBSD may decide to add support for MAP_ALIGNED_SUPER for other platforms.

Doc/library/mmap.rst
```diff
@@ -81,7 +81,11 @@ To map anonymous memory, -1 should be passed as the fileno along with the length
    private copy-on-write mapping, so changes to the contents of the mmap
    object will be private to this process, and :const:`MAP_SHARED` creates a
    mapping that's shared with all other processes mapping the same areas of
-   the file.  The default value is :const:`MAP_SHARED`.
+   the file. :const:`MAP_ALIGNED_SUPER` (FreeBSD only) allows to map to
+   pages of 2MB (x86) or so, it is enabled by default, at boot time,
+   in most of architectures and exists since 2008.
+   :ref:`MAP_* constants <mmap-constants>`.
```
vstinner: "MAP_* constants" is not really a sentence. Maybe move it after the default, and write something like:

```
The default value is :const:`MAP_SHARED`. 
See :ref:`MAP_* constants <mmap-constants>`.
```

Doc/library/mmap.rst
ghost: It supports more than x86 already in fact but you re right I ll try to find more concise wording

Doc/library/mmap.rst
```diff
@@ -341,3 +359,6 @@ MADV_* Constants
    Availability: Systems with the madvise() system call.
 
    .. versionadded:: 3.8
+
+   .. versionadded:: 3.9
+      Add :data:`MAP_ALIGNED_SUPER` constant.
```
vstinner: Can you move this to the your new  MAP_* Constants  section?

Misc/NEWS.d/next/Library/2019-07-05-20-54-46.bpo-35711.xYuDRM.rst
```diff
@@ -0,0 +1 @@
+Adding new :data:`mmap.MAP_CONCEAL` OpenBSD constant.
```
vstinner: Can you try to merge the 2 changelog entries? Put content of the 2nd entry in the first one, remove the second file.

Doc/library/mmap.rst
```diff
@@ -307,6 +311,20 @@ To map anonymous memory, -1 should be passed as the fileno along with the length
...
++++++++++++++++
+
+.. data:: MAP_ANONYMOUS
+          MAP_ANON
+          MAP_SHARED
+          MAP_PRIVATE
+          MAP_DENYWRITE
+          MAP_EXECUTABLE
+          MAP_ALIGNED_SUPER
+          MAP_CONCEAL
```
vstinner, line 326: You have to document that MAP_CONCEAL is new.

Doc/library/mmap.rst
```diff
@@ -81,7 +81,11 @@ To map anonymous memory, -1 should be passed as the fileno along with the length
    private copy-on-write mapping, so changes to the contents of the mmap
    object will be private to this process, and :const:`MAP_SHARED` creates a
    mapping that's shared with all other processes mapping the same areas of
-   the file.  The default value is :const:`MAP_SHARED`.
+   the file. :const:`MAP_ALIGNED_SUPER` (FreeBSD only) to use
+   super pages feature, enabled by default in most of architectures
+   and exists since 2008.
```
vstinner: I don't understand this sentence :-/ I suggest:

```
:const:`MAP_ALIGNED_SUPER` (FreeBSD only) uses super pages feature,
it is enabled by default in most of architectures and exists since 2008.
```

PR #14516, opened 2019-07-01
bpo-37472: Remove Lib/test/outstanding_bugs.py

vstinner:
https://bugs.python.org/issue37472


jdemeyer:
I would remove the `NEWS` entry:
- it's not useful for users since it doesn't change CPython in any way
- it's not useful for developers since clearly nobody uses it

vstinner:
> it's not useful for developers since clearly nobody uses it

It's unclear if a developer still uses it or not, so I prefer to document its removal.

vstinner:
I made the NEWS entry way shorter, to not pollute the change.

zware (approved):

vstinner:
@zware: Thanks for your review. It makes me confident to remove this file ;-)

PR #14518, opened 2019-07-01
Remove unused imports

vstinner:


vstinner:
I'm not sure of test_case_sensitivity.py change. In case of doubt, I prefer to remove it :-)

vstinner:
I backported the change to Python 3.7 and 3.8 to reduce conflicts on backports.

PR #14522, opened 2019-07-01
[3.7] Remove unused imports in tests (GH-14518)

vstinner:
(cherry picked from commit 8f4ef3b019ce380022018587571b0f970e668de3)

### DIFF
--- a/Lib/test/libregrtest/main.py
+++ b/Lib/test/libregrtest/main.py
@@ -587,7 +587,6 @@ class Regrtest:
 
     def cleanup(self):
         import glob
-        import shutil
 
         path = os.path.join(self.tmp_dir, 'test_python_*')
         print("Cleanup %s directory" % self.tmp_dir)
--- a/Lib/test/support/__init__.py
+++ b/Lib/test/support/__init__.py
@@ -5,7 +5,6 @@ if __name__ != 'test.support':
 
 import collections.abc
 import contextlib
-import datetime
 import errno
 import faulthandler
 import fnmatch
@@ -13,7 +12,6 @@ import functools
 import gc
 import importlib
 import importlib.util
-import io
 import logging.handlers
 import nntplib
 import os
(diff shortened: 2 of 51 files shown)


ned-deily:
Why is this being backported into a [maintenance branch](https://devguide.python.org/devcycle/#maintenance-branches)?  What bug is this fixing?  What bugs are being introduced by it?

vstinner:
> Why is this being backported into a maintenance branch? What bug is this fixing? What bugs are being introduced by it?

Oh sorry, I only explained it in the main PR (for master):
"I backported the change to Python 3.7 and 3.8 to reduce conflicts on backports."
https://github.com/python/cpython/pull/14518#issuecomment-507354764

Otherwise, automated backports will more likely fail with a dummy conflict. I prefer when the bot is able to automatically backport changes to 3.7 and 3.8 ;-) I'm trying to backport all bugfixes to 2.7, 3.7 and 3.8: all maintained branches.

Note: I chose to not backport this change to 2.7, there are likely too many conflicts, and 2.7 is getting less and less changes.

ned-deily:
I don't want to make a big deal out of this but, In general, reducing the potential for future backport conflicts is **not** an acceptable reason for changing a maintenance branch:

> The only changes allowed to occur in a maintenance branch without debate are bug fixes.

To do otherwise is to jeopardize the stability and compatibility guarantees that we make to our downstream users (i.e. that they can upgrade from one micro release to any higher micro release without anything breaking), solely for our own convenience.  Sure, there is not a hard and fast rule: factors like where a particular release is in its life cycle should be taken into account.  So backporting such a change from master to a branch that is still in a pre-release phase (like 3.8 is at the moment) is one thing.  But 3.7 has been released for a year and is halfway through its maintenance phase.  We should not be planning to backport lots of changes to it at this point.  And changes to tests are still changes to the codebase of a release.  Changes to the tests can hide errors and makes it more difficult to see regressions.  What are the chances that some new bug has been introduced in the changes to over 40 files - for no functional benefit?

vstinner:
@ned-deily: Ok, I created PR #14555 to revert it. I will merge it as soon as the CI tests pass.

vstinner:
I reverted this change ;-)

PR #14524, opened 2019-07-01
bpo-37363: Add audit events on startup for the run commands

zooba:
https://bugs.python.org/issue37363


tiran:
Fantastic! This simplifies one of my PoCs for EP a lot.

Are these all entry points or is there another entry point for embedded Python interpreters?

zooba:
There are definitely more entry points, but if someone else is embedding it then they have full control and audit hooks won't help :)

Doc/library/sys.rst
```diff
@@ -905,6 +905,12 @@ always available.
    read, so that you can set this hook there.  The :mod:`site` module
    :ref:`sets this <rlcompleter-config>`.
 
+   .. audit-event:: cpython.run_interactivehook hook sys.__interactivehook__
```
tiran, line 908: I don't understand the audit-event syntax. The event takes one argument (hook), but the audit-event has two arguments. What's going on here?

zooba: I need to document the syntax somewhere...

Basically, it's:
```
.. audit-event:: event-name list,of,comma,separated,args anchor
```

Where the anchor is used for the backlink from the generated table (which appears to [not be generating right now](https://docs.python.org/3.8/library/audit_events.html)...).

zooba:
I can't repro the Azure Pipelines failures locally (on Windows or Ubuntu), so I'm guessing it's environment related. Making one more guess before I start adding lots of diagnostic info :)

PR #14531, opened 2019-07-01
bpo-37476: Adding a unit test of unicode in test_unicode.py

shihai1991:
https://bugs.python.org/issue37476

### DIFF
--- a/Lib/test/test_unicode.py
+++ b/Lib/test/test_unicode.py
@@ -2833,6 +2833,20 @@ class CAPITest(unittest.TestCase):
         self.assertEqual(unicode_asutf8(nonbmp), b'\xf4\x8f\xbf\xbf')
         self.assertRaises(UnicodeEncodeError, unicode_asutf8, 'a\ud800b\udfffc')
 
+    # Test PyUnicode_AsUTF8()
+    @support.cpython_only
+    def test_asutf8andsize(self):
+        from _testcapi import unicode_asutf8andsize
+
+        bmp = '\u0100'
+        bmp2 = '\uffff'
+        nonbmp = chr(0x10ffff)
+
+        self.assertEqual(unicode_asutf8andsize(bmp), (b'\xc4\x80', 2))
+        self.assertEqual(unicode_asutf8andsize(bmp2), (b'\xef\xbf\xbf', 3))
+        self.assertEqual(unicode_asutf8andsize(nonbmp), (b'\xf4\x8f\xbf\xbf', 4))
+        self.assertRaises(UnicodeEncodeError, unicode_asutf8andsize, 'a\ud800b\udfffc')
+
     # Test PyUnicode_FindChar()
     @support.cpython_only
     def test_findchar(self):
--- a/Lib/test/test_unicode.py
+++ b/Lib/test/test_unicode.py
@@ -2824,9 +2824,13 @@ class CAPITest(unittest.TestCase):
     def test_asutf8(self):
         from _testcapi import unicode_asutf8
 
-        self.assertEqual(unicode_asutf8('abc'), 'abc')
-        self.assertEqual(unicode_asutf8('abc\0'), 'abc')
-        self.assertEqual(unicode_asutf8('abc\0abc'), 'abc')
+        bmp = '\u0100'
+        bmp2 = '\uffff'
+        nonbmp = chr(0x10ffff)
+
+        self.assertEqual(unicode_asutf8(bmp), b'\xc4\x80')
+        self.assertEqual(unicode_asutf8(bmp2), b'\xef\xbf\xbf')
+        self.assertEqual(unicode_asutf8(nonbmp), b'\xf4\x8f\xbf\xbf')
         self.assertRaises(UnicodeEncodeError, unicode_asutf8, 'a\ud800b\udfffc')
 
     # Test PyUnicode_FindChar()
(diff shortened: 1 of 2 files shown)


Modules/_testcapimodule.c
```diff
@@ -1921,6 +1921,25 @@ unicode_asucs4(PyObject *self, PyObject *args)
     return result;
 }
 
+static PyObject *
+unicode_asutf8(PyObject *self, PyObject *args)
+{
+    PyObject *unicode, *result;
```
zhangyangyu: Add Curly brackets for all `if` please.

Lib/test/test_unicode.py
```diff
@@ -2819,6 +2819,16 @@ def test_asucs4(self):
...
             self.assertEqual(unicode_asucs4(s, len(s), 0), s+'\uffff')
 
+    # Test PyUnicode_AsUTF8()
+    @support.cpython_only
+    def test_asutf8(self):
+        from _testcapi import unicode_asutf8
+
+        self.assertEqual(unicode_asutf8('abc'), 'abc')
+        self.assertEqual(unicode_asutf8('abc\0'), 'abc')
+        self.assertEqual(unicode_asutf8('abc\0abc'), 'abc')
```
zhangyangyu: It's more like you are testing `decodeUTF8`. I'd suggest add more encode samples, BMP, non-BMP, just take from other tests.

Lib/test/test_unicode.py
```diff
@@ -2819,6 +2819,16 @@ def test_asucs4(self):
             self.assertEqual(unicode_asucs4(s, len(s), 1), s+'\0')
             self.assertEqual(unicode_asucs4(s, len(s), 0), s+'\uffff')
 
+    # Test PyUnicode_AsUTF8()
+    @support.cpython_only
+    def test_asutf8(self):
+        from _testcapi import unicode_asutf8
+
+        self.assertEqual(unicode_asutf8('abc'), 'abc')
```
vstinner: The function encodes to UTF-8, so the result should be a bytes string, not Unicode. Use PyString_FromString() rather than PyUnicode_FromString() in _testcapi.

vstinner (commented):
If you add tests, please test also PyUnicode_AsUTF8AndSize() which would allow embedded NUL characters/bytes.

Lib/test/test_unicode.py
zhangyangyu: The others similar C API tests also return unicodes other than bytes which in some sense also tests the counter part decoding function, though I don't very understand why designed like this.

vstinner: Either fix other tests, or leave them unchanged. But I would prefer to not add new buggy tests :-)

shihai1991: So returning the decoded bytes string is right way? I am not sure I understand it clearly.

Lib/test/test_unicode.py
shihai1991: np, xiang core.

Modules/_testcapimodule.c
shihai1991: done

shihai1991:
> If you add tests, please test also PyUnicode_AsUTF8AndSize() which would allow embedded NUL characters/bytes.

np, I would continue to add test.

tiran:
FYI, I cancelled the Travis CI run to free some resources for upcoming 3.8b2 and 3.7.4rc2 releases.

Lib/test/test_unicode.py
zhangyangyu: Return bytes, not unicodes. Test decode functionality separately.

Modules/_testcapimodule.c
```diff
@@ -1921,6 +1921,28 @@ unicode_asucs4(PyObject *self, PyObject *args)
...
+        return NULL;
+    }
+
+    buffer = PyUnicode_AsUTF8(unicode);
+    if (buffer == NULL) {
+        return NULL;
+    }
+
+    result = PyUnicode_FromString(buffer);
+    if (result == NULL) {
```
ZackerySpytz, line 1959: This explicit check for `result == NULL` is not necessary.
Just do `return func_name(buffer);`.

shihai1991: Zackery, good catch. As Victor said,  I would use `PyBytes_FromString` :)

Lib/test/test_unicode.py
```diff
@@ -2819,6 +2819,34 @@ def test_asucs4(self):
...
+        bmp = '\u0100'
+        bmp2 = '\uffff'
+        nonbmp = chr(0x10ffff)
+
+        self.assertEqual(unicode_asutf8(bmp), b'\xc4\x80')
+        self.assertEqual(unicode_asutf8(bmp2), b'\xef\xbf\xbf')
+        self.assertEqual(unicode_asutf8(nonbmp), b'\xf4\x8f\xbf\xbf')
+        self.assertRaises(UnicodeEncodeError, unicode_asutf8, 'a\ud800b\udfffc')
+
+    # Test PyUnicode_AsUTF8()
```
zhangyangyu: PyUnicode_AsUTF8AndSize()

zhangyangyu (approved):

shihai1991:
@vstinner hi, victor. Pls help me review this patch again, thanks ;)

PR #14535, opened 2019-07-01
bpo-37199: Replace the early returns added in c2cda63

ZackerySpytz:
https://bugs.python.org/issue37199


aeros (approved):
Using a conditional for an early return seems to be particularly unconventional across the rest of Python, and can somewhat reduce the readability of the code when it is done too many times, approved. 

This is not specifically related to this issue, but what are your thoughts on the usage of the single letter variable names ``t`` and ``p`` in on lines 1275 and 1282? Changing the variable name ``t`` to ``task`` for example seems like it would improve the readability, particularly for anyone who is not as intimately familiar with async. I'm not arguing for ``t`` to be named ``task`` specifically if there's another preferred name. I would be in favor of any name that accurately summarizes the purpose of the variable within the method.

benjaminp:
merging because it addresses review comments on GH-14480.

PR #14537, opened 2019-07-01
bpo-19696: Move threaded import tests to dir test_importlib/

aeros:
This is the final component of issue 19696. This PR includes both test_threaded_import.py and threaded_import_hangers.py because the hangers are required for the threaded import test, so the two files are directly related. I'm not sure as to why the changes can't be automatically merged, the previous file move I had done had no issues with that. Is it due to more than one file being moved in a single PR?

There's a minor typo in one of the commit messages ("theaded" instead of "threaded"), but there were no issues with the file moves. Is it possible to edit the commit messages? If this typo is an issue and there's no way of editing the commit messages, I can create a new branch, redo the changes, and submit another PR.

Since this is just a file move, a "skip news" label would likely be appropriate. 

https://bugs.python.org/issue19696


aeros167:
I believe I found the cause of the issue, on line 184, ``import test.threaded_import_hangers`` probably needs to be changed to ``import test.test_importlib.threaded_import_hangers`` to reflect the new location of the file.

aeros167:
After performing a code search of all references to ``threaded_import_hangers``, it appears there is another reference to it in ``PCbuild/lib.pyproj``. I'll update that reference as well.

aeros167:
@brettcannon Do I need to resubmit the PR for the tests to be ran again, or is there something else that I'm missing? If I need to resubmit the PR, I'll instead break it up into two. ``test_threaded_import.py`` was able to be moved without issue, it seems that moving ``threaded_import_hangers.py`` has been the primary cause of the conflicts.

aeros167:
Closing this PR, I'll create a new one which exclusively moves ``test_threaded_import.py`` and attempt to move ``test_import_hangers.py`` in a separate one later tonight or tomorrow. Hopefully that will resolve the merge conflicts.

PR #14538, opened 2019-07-02
bpo-37480: add ptpython to alternate interpreters

zachvalenta:
Despite personally using bpython, ptpython seems both mature and popular enough to merit inclusion in the docs. Plus, bpython recommends them :)

https://bugs.python.org/issue37480

### DIFF
--- a/Doc/tutorial/interactive.rst
+++ b/Doc/tutorial/interactive.rst
@@ -45,10 +45,11 @@ quotes, etc., would also be useful.
 One alternative enhanced interactive interpreter that has been around for quite
 some time is IPython_, which features tab completion, object exploration and
 advanced history management.  It can also be thoroughly customized and embedded
-into other applications.  Another similar enhanced interactive environment is
-bpython_.
+into other applications.  Other similar enhanced interactive environments are
+bpython_ and ptpython_.
 
 
 .. _GNU Readline: https://tiswww.case.edu/php/chet/readline/rltop.html
 .. _IPython: https://ipython.org/
 .. _bpython: https://www.bpython-interpreter.org/
+.. _ptpython: https://github.com/prompt-toolkit/ptpython


the-knights-who-say-ni:

zachvalenta:
Unsure if this PR would qualify for the `skip news` label; checked out some other recent PRs (like [14538](https://github.com/python/cpython/pull/14568)) and couldn't tell how that call was made. Just let me know and I'll be happy to add.

aeros (changes requested):
@zachvalenta Thanks for the contribution! To answer your question, I'd say this doc change could warrant a `skip news`, as the section itself still provides the same purpose (describing alternative interpreters). A brief news blurb wouldn't hurt though, such as "Added ptpython to list of alternative interpreters".

As to whether or not ptpython should actually be added to this section, I'm not certain that the docs are the appropriate place for maintaining a list of alternative interpreters. From my understanding, this section was not meant to provide a list of all up to date alternatives. Instead, it seems to just be there to provide a couple of examples. My concern is that if we add this to the docs, there will be an argument to continually add and remove others, which further adds to documentation maintenance costs. As @terryjreedy had suggested on the bug tracker page, maintaining a list of interpreters seems better suited to the wiki (which is much easier to periodically edit).

For the most part, worthwhile maintenance costs on the docs primarily revolve around the version of the Python language itself. However, I do think that it might be appropriate to link to the wiki page itself in the docs, so that users can easily find a list of current interpreters. You can find that here: https://wiki.python.org/moin/PythonEditors#Enhanced_Python_shells. Feel free to add ptpython to the list on there.

Depending on what the others think, it might be worthwhile to also refactor the paragraph of alternative interpreters to remove any specific mention of individuals ones in the docs, with the only direct link being to the python wiki page section. Something like this might be appropriate:

```
For a maintained list of enhanced interactive environments, visit the python wiki:
https://wiki.python.org/moin/PythonEditors#Enhanced_Python_shells
```

This would likely be the most fair to all of the alternative options, so it does not appear that the docs favor any of them in particular. The docs really should not be the place to mention specific alternative applications when we have a wiki.

zachvalenta:
@aeros167 I agree with all this. The wiki is a more natural home for that info, and might as well move the `bpython` reference over there while we're at it. I'll update my PR and make an edit to the wiki to add as well. Harried at work this week so will get to it this weekend!

matrixise:
As indicated in the comments, the right place for this info is the wiki. Thank you again for your proposal.

PR #14550, opened 2019-07-02
bpo-37233: optimize case totalargs == 0 in method_vectorcall

jdemeyer:
Because of [bpo-37138](https://bugs.python.org/issue37138), we need to check the case `totalargs == 0` anyway. So it doesn't hurt to move that check earlier and add a fast path for the common case of calling a method without arguments.

CC @methane 

https://bugs.python.org/issue37233

### DIFF
--- a/Objects/classobject.c
+++ b/Objects/classobject.c
@@ -62,9 +62,13 @@ method_vectorcall(PyObject *method, PyObject *const *args,
     }
     else {
         Py_ssize_t nkwargs = (kwnames == NULL) ? 0 : PyTuple_GET_SIZE(kwnames);
-        PyObject **newargs;
         Py_ssize_t totalargs = nargs + nkwargs;
+        if (totalargs == 0) {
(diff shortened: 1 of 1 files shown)


methane:
I'm not sure this is really optimization.
Isn't `(nargsf & PY_VECTORCALL_ARGUMENTS_OFFSET)` true on major case?

We need `if (totalargs) {` at C level.  But it doesn't mean this check is needed in assembly level.
Compilers may optimize out this check if it is safe.

Would you show microbenchmark result for this?

jdemeyer:
> Compilers may optimize out this check if it is safe.

I checked and they don't optimize away this check. So my original reasoning is still correct: if this check has to be done anyway, we should put it as early as possible.

jdemeyer:
If you want to backport this, you should also backport #13974

methane:
OK, I want to keep 3.8 and 3.9 same to ease future backporting.
#13974 seems safe enough to backport too.

PR #14555, opened 2019-07-02
[3.7] Revert "[3.7] Remove unused imports in tests (GH-14518)"

vstinner:
Reverts python/cpython#14522


vstinner:
Rationale for the revert: https://github.com/python/cpython/pull/14522#issuecomment-507496216

PR #14557, opened 2019-07-02
bpo-37484: Optimize __exit__

jdemeyer:
CC @methane @markshannon 

https://bugs.python.org/issue37484

### DIFF
--- a/Python/ceval.c
+++ b/Python/ceval.c
@@ -3314,7 +3314,6 @@ main_loop:
 
                Finally we push the result of the call.
             */
-            PyObject *stack[3];
             PyObject *exit_func;
             PyObject *exc, *val, *tb, *res;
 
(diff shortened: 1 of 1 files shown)


Python/ceval.c
```diff
@@ -3351,10 +3350,13 @@ _PyEval_EvalFrameDefault(PyFrameObject *f, int throwflag)
...
             }
 
+            PyObject *_stack[4];
+            PyObject **stack = _stack + 1;  /* For PY_VECTORCALL_ARGUMENTS_OFFSET */
             stack[0] = exc;
             stack[1] = val;
             stack[2] = tb;
-            res = _PyObject_FastCall(exit_func, stack, 3);
+            res = _PyObject_Vectorcall(exit_func, stack,
+                    3 | PY_VECTORCALL_ARGUMENTS_OFFSET, NULL);
```
methane, line 3355: How about this?

```c
PyObject *stack[4] = {NULL, exc, val, tb};
res = _PyObject_Vectorcall(exit_func, stack + 1, 3 | PY_VECTORCALL_ARGUMENTS_OFFSET, NULL);
```

jdemeyer: Fine for me. It's only slightly less efficient since you need to assign that `NULL`.

jdemeyer: > How about this?

OK, done.

PR #14558, opened 2019-07-02
bpo-37483: add _PyObject_CallOneArg() function

jdemeyer:

### DIFF
--- a/Doc/c-api/object.rst
+++ b/Doc/c-api/object.rst
@@ -264,6 +264,17 @@ Object Protocol
    .. versionadded:: 3.9
 
 
+.. c:function:: PyObject* _PyObject_CallOneArg(PyObject *callable, PyObject *arg)
+
+   Call a callable Python object *callable* with exactly 1 positional argument
+   *arg* and no keyword arguments.
+
+   Return the result of the call on success, or raise an exception and return
+   *NULL* on failure.
+
+   .. versionadded:: 3.9
+
+
 .. c:function:: PyObject* PyObject_Call(PyObject *callable, PyObject *args, PyObject *kwargs)
 
    Call a callable Python object *callable*, with arguments given by the
--- a/Include/cpython/abstract.h
+++ b/Include/cpython/abstract.h
@@ -62,6 +62,7 @@ PyVectorcall_NARGS(size_t n)
 static inline vectorcallfunc
 _PyVectorcall_Function(PyObject *callable)
 {
+    assert(callable != NULL);
     PyTypeObject *tp = Py_TYPE(callable);
     if (!PyType_HasFeature(tp, _Py_TPFLAGS_HAVE_VECTORCALL)) {
         return NULL;
@@ -132,6 +133,17 @@ _PyObject_FastCall(PyObject *func, PyObject *const *args, Py_ssize_t nargs)
 static inline PyObject *
 _PyObject_CallNoArg(PyObject *func) {
     return _PyObject_Vectorcall(func, NULL, 0, NULL);
+}
+
+static inline PyObject *
+_PyObject_CallOneArg(PyObject *func, PyObject *arg)
+{
+    assert(arg != NULL);
+    PyObject *_args[2];
+    PyObject **args = _args + 1;  // For PY_VECTORCALL_ARGUMENTS_OFFSET
+    args[0] = arg;
+    return _PyObject_Vectorcall(func, args,
+                                1 | PY_VECTORCALL_ARGUMENTS_OFFSET, NULL);
 }
 
 PyAPI_FUNC(PyObject *) _PyObject_Call_Prepend(
--- a/Misc/NEWS.d/next/C API/2019-07-02-15-42-37.bpo-37483.vftT4f.rst
+++ b/Misc/NEWS.d/next/C API/2019-07-02-15-42-37.bpo-37483.vftT4f.rst
@@ -1,0 +1,2 @@
+Add new function ``_PyObject_CallOneArg`` for calling an object with one
+positional argument.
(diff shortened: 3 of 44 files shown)


aeros (commented):
@jdemeyer On lines 145 and 146 of abstract.h: 
``return _PyObject_Vectorcall(func, args, 1 | PY_VECTORCALL_ARGUMENTS_OFFSET, NULL);`` 

What is the purpose of using the constant ``PY_VECTORCALL_ARGUMENTS_OFFSET`` in a function that is made specifically for single arguments? Unless I am misunderstanding something, ``return _PyObject_Vectorcall(func, args, 1, NULL);`` would effectively have the same functionality. Is there any circumstance where this function is used and ``1 | PY_VECTORCALL_ARGUMENTS_OFFSET`` does not result in ``1``?

jdemeyer:
> What is the purpose of using the constant `PY_VECTORCALL_ARGUMENTS_OFFSET` in a function that is made specifically for single arguments?

It's meant for optimizing calls to bound methods. If you're calling a bound method with 1 argument, you're effectively calling a function with 2 arguments when you add `self`. So `args[-1]` is meant to store `self`. See https://www.python.org/dev/peps/pep-0590/#py-vectorcall-arguments-offset

methane (approved):

Modules/_testbuffer.c
```diff
@@ -788,7 +788,7 @@ get_itemsize(PyObject *format)
     PyObject *tmp;
     Py_ssize_t itemsize;
 
-    tmp = PyObject_CallFunctionObjArgs(calcsize, format, NULL);
+    tmp = _PyObject_CallOneArg(calcsize, format);
```
skrah: Please take private functions out of this test module.

Modules/_xxsubinterpretersmodule.c
```diff
@@ -1642,7 +1642,7 @@ _channel_from_cid(PyObject *cid, int end)
     if (cls == NULL) {
         return NULL;
     }
-    PyObject *chan = PyObject_CallFunctionObjArgs(cls, cid, NULL);
+    PyObject *chan = _PyObject_CallOneArg(cls, cid);
```
skrah: Private functions are definitely out of place here.

Modules/_decimal/_decimal.c
```diff
@@ -2514,7 +2514,7 @@ dec_from_float(PyObject *type, PyObject *pyfloat)
     CURRENT_CONTEXT(context);
     result = PyDecType_FromFloatExact(&PyDec_Type, pyfloat, context);
     if (type != (PyObject *)&PyDec_Type && result != NULL) {
-        Py_SETREF(result, PyObject_CallFunctionObjArgs(type, result, NULL));
+        Py_SETREF(result, _PyObject_CallOneArg(type, result));
```
skrah: This is a code path that is practically never taken. Additionally, `PyDecType_FromFloatExact()` will completely dominate the run time, so any marginal speedup will not be noticed.

Please leave such changes to module authors/maintainers.

skrah (changes requested):

aeros167:
@jdemeyer Wouldn't the value of ``1 | PY_VECTORCALL_ARGUMENTS_OFFSET`` always resolve to the same value in the case of this function though? The primary use for the it seems to be calculating the offset when the number of args is unknown: 
``nargs | PY_VECTORCALL_ARGUMENTS_OFFSET`` 

or for the example provided in PEP 590: 
``nargs & PY_VECTORCALL_ARGUMENTS_OFFSET``

I suppose a better question then is: what is the purpose of using ``PY_VECTORCALL_ARGUMENTS_OFFSET`` when the the number of arguments for the function is static? As far as I can tell, the result of ``1 | PY_VECTORCALL_ARGUMENTS_OFFSET`` will always be ``1`` any time ``_PyObject_CallOneArg()`` is utilized. Is this incorrect?

Also, apologies for the questions if these answers are fairly obvious. I have some understanding of C fundamentals, but I'm significantly more experienced with Python. I'm just trying to get a solid understanding of both so that I can be more useful as a contributor.

jdemeyer:

jdemeyer:
> Wouldn't `1 | PY_VECTORCALL_ARGUMENTS_OFFSET` always resolve to the same value in the case of this function though?

`PY_VECTORCALL_ARGUMENTS_OFFSET` is a constant flag, so of course `1 | PY_VECTORCALL_ARGUMENTS_OFFSET` is a constant value. But that's not the point.

The `PY_VECTORCALL_ARGUMENTS_OFFSET` flag is meant for the called function. It indicates that the value `args[-1]` may temporarily be changed as scratch space. That's the reason for the array `_args` of length two in this PR: `_args[0]` is the same as `args[-1]` which can be used as scratch space by the called function.

aeros167:
@jdemeyer Oh that makes sense, I see. I hadn't realized the purpose of it within this function was solely to be an indicating flag, in general I'm far more used to working on much smaller scale projects and codebases where almost everything has a specific functional purpose. Seeing the two constant values being used for the nargs argument of ``PyObject_Vectorcall`` had seemed redundant to me initially, but now I understand the reasoning. 

Thank you for having the patience to fully explain it.

skrah (approved):

skrah:
Thanks for the changes, the "approval" was a GitHub thing, I'm actually -0 on such large replacement PRs.

jdemeyer:
It's not doing replacement for the sake of replacement. As I explained on bpo-37483, it gives a nice performance gain (in terms of runtime and stack usage).

skrah:
I mean **large** replacement.  Note that I did not protest against the changes in `memoryobject.c`.

PR #14561, opened 2019-07-02
bpo-37482: Fix email address name with encoded words and special chars

bsiem:
Special characters in email address header display names are normally
put within double quotes. However, encoded words (=?charset?x?...?=) are
not allowed withing double quotes. When the header contains a word with
special characters and another word that must be encoded, the first one
must also be encoded.

In the next example, the display name in the From header is quoted and
therefore the comma is allowed; in the To header, the comma is not
within quotes and not encoded, which is not allowed and therefore
rejected by some mail servers.

From: "Foo Bar, France" <foo@example.com>
To: Foo Bar, =?utf-8?q?Espa=C3=B1a?= <foo@example.com>

https://bugs.python.org/issue37482

### DIFF
--- a/Misc/NEWS.d/next/Library/2019-07-09-11-20-21.bpo-37482.auzvev.rst
+++ b/Misc/NEWS.d/next/Library/2019-07-09-11-20-21.bpo-37482.auzvev.rst
@@ -1,0 +1,1 @@
+Fix serialization of display name in originator or destination address fields with both encoded words and special chars.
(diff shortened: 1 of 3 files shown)


the-knights-who-say-ni:

Misc/NEWS.d/next/Library/2019-07-09-11-20-21.bpo-37482.auzvev.rst
```diff
@@ -0,0 +1 @@
+Fixed email address display name with both encoded words and special chars.
```
maxking: nit:
```suggestion
Fix serialization of display name in orignator or destination address fields with both encoded words and special chars.
```

maxking (changes requested):
Thanks for the PR, it looks good to me. 

I have a very tiny nit, but otherwise I think we can merge this.

bsiem:

maxking (approved):

PR #14564, opened 2019-07-02
bpo-37421: Fix test_distutils.test_build_ext()

vstinner:
test_distutils.test_build_ext() is now able to remove the temporary
directory on Windows: don't import the newly built C extension ("xx")
in the current process, but test it in a separated process.

### DIFF
--- a/Lib/distutils/tests/support.py
+++ b/Lib/distutils/tests/support.py
@@ -6,6 +6,7 @@ import tempfile
 import unittest
 import sysconfig
 from copy import deepcopy
+import test.support
 
 from distutils import log
 from distutils.log import DEBUG, INFO, WARN, ERROR, FATAL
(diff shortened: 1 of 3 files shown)


zware (approved):

aeros (approved):
@vstinner Thanks for changing the variable ``d`` to ``tmpdir`` on line 67 of ``support.py``. In my opinion, non-descriptive variables can be the bane of readability.

Is there a particular reason or advantage in this case for wrapping the tests within a whitespace-stripped string instead of writing the Tests class within another file? I have no issues with it, I just haven't seen it done before.

vstinner:
> @vstinner Thanks for changing the variable d to tmpdir on line 67 of support.py. In my opinion, non-descriptive variables can be the bane of readability.

You're welcome.

> Is there a particular reason or advantage in this case for wrapping  the tests within a whitespace-stripped string instead of writing the  Tests class within another file? I have no issues with it, I just  haven't seen it done before.

It's more convenient to put everything at the same place. It's a very common pattern if you look at the Python test suite.

aeros167:
@vstinner Thanks for the explanation. I'm a newer contributor, so I'm not overly familiar with the common patterns within the Python test suite. However, I definitely am a fan of how compact it is, I'll definitely keep the format in mind. After finishing up reorganization of ``test_importlib``, Brett had suggested that it'd be helpful to work on modernizing the tests over there. If possible, I'll use this format as a template.

PR #14572, opened 2019-07-03
bpo-37421: Fix multiprocessing get_temp_dir() finalizer

vstinner:
Fix multiprocessing.util.get_temp_dir() finalizer: clear also the
'tempdir' configuration of the current process, so next call to
get_temp_dir() will create a new temporary directory, rather than
reusing the removed temporary directory.

### DIFF
--- a/Lib/multiprocessing/util.py
+++ b/Lib/multiprocessing/util.py
@@ -106,6 +106,11 @@ def log_to_stderr(level=None):
 # Function returning a temp directory which will be removed on exit
 #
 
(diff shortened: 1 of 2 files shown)


vstinner:
Oh, this fix is not enough to repair ``./python -u -m test test_multiprocessing_spawn -R 3:3`` regression, two other tests are still failing:

```
ERROR: test_context (test.test_multiprocessing_spawn.TestStartMethod)
ERROR: test_set_get (test.test_multiprocessing_spawn.TestStartMethod)
```

PR #14575, opened 2019-07-03
bpo-37493: use _PyObject_CallNoArg in more places

jdemeyer:
CC @vstinner @methane 

https://bugs.python.org/issue37493

### DIFF
--- a/Doc/extending/newtypes.rst
+++ b/Doc/extending/newtypes.rst
@@ -104,7 +104,7 @@ done.  This can be done using the :c:func:`PyErr_Fetch` and
            /* This saves the current exception state */
            PyErr_Fetch(&err_type, &err_value, &err_traceback);
 
-           cbresult = PyObject_CallObject(self->my_callback, NULL);
+           cbresult = PyObject_CallNoArgs(self->my_callback);
            if (cbresult == NULL)
                PyErr_WriteUnraisable(self->my_callback);
            else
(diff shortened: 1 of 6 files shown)


skrah (dismissed):
I'm a bit surprised --- we discussed on bpo to use fewer places, but you add more.

So please take out the decimal changes.  I'll add new functions as I see fit.

jdemeyer:
> we discussed on bpo to use fewer places, but you add more.

I'm not sure what you mean. The only thing that we discussed (on bpo-37483 which relates to a different PR) is that maybe we should not use private functions in `_decimal.c`. However, in this case I'm using a documented public function, belonging to the limited API `PyObject_CallNoArgs`.

So please let me know whether the changes to `_decimal.c` should be reverted anyway.

skrah:
The PR title mentions `_PyObject_CallNoArg()`.  I see now that there are other functions in the `_decimal.c` diff, but again not in speed sensitive locations.

So perhaps there are too many new functions.  I'll decide later if to use any of these, so yes, please take them out.

jdemeyer:
> not in speed sensitive locations.

Even if it's not speed sensitive, I see no reason to prefer `PyObject_CallObject(func, NULL)` over `PyObject_CallNoArgs(func)`.

jdemeyer:

skrah (commented):
Thanks for the changes.

PR #14576, opened 2019-07-03
[3.6] bpo-35605: Fix documentation build for sphinx<1.6

asottile:
https://bugs.python.org/issue35605


ned-deily (approved):
I'll accept this but we're not going to keep supporting the oldest versions of Sphinx forever since it's so easy to upgrade.

PR #14577, opened 2019-07-03
bpo-27452: IDLE: Cleanup config.py code

csabella:
* Remove `RemoveFile` method.
* Simplify `CreateConfigHandlers` handling of `__file__`.

Patch by Terry Reedy.

https://bugs.python.org/issue27452


terryjreedy (approved):
I tested running config.py from repository idlelib with both ../../python config.py and -m config.

PR #14578, opened 2019-07-03
gh-59705: Export threading.Thread() names to the OS

ZackerySpytz:
Call pthread_setname_np() et al. with the name argument passed to
threading.Thread().

https://bugs.python.org/issue15500

* Issue: gh-59705

### DIFF
--- a/Lib/threading.py
+++ b/Lib/threading.py
@@ -41,6 +41,11 @@ try:
     __all__.append('get_native_id')
 except AttributeError:
     _HAVE_THREAD_NATIVE_ID = False
+try:
+    _set_thread_name = _thread._set_thread_name
+    _HAVE_SET_THREAD_NAME = True
+except AttributeError:
+    _HAVE_SET_THREAD_NAME = False
 ThreadError = _thread.error
 try:
     _CRLock = _thread.RLock
@@ -924,6 +929,8 @@ class Thread:
             self._set_tstate_lock()
             if _HAVE_THREAD_NATIVE_ID:
                 self._set_native_id()
+            if _HAVE_SET_THREAD_NAME:
+                _set_thread_name(self.name)
             self._started.set()
             with _active_limbo_lock:
                 _active[self._ident] = self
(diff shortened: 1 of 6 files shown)


ZackerySpytz:
Unfortunately, the string passed to `pthread_setname_np()` must be less than 16 chars on Linux, so the name may be truncated.

debauer:
Will this ever happen?! The Issue has this year the 10th anniversary. In the Issue @eryksun also provides Code for Windows... so let this happen.

Lib/threading.py
```diff
@@ -924,6 +929,8 @@ def _bootstrap_inner(self):
             self._set_tstate_lock()
             if _HAVE_THREAD_NATIVE_ID:
                 self._set_native_id()
+            if _HAVE_SET_THREAD_NAME:
+                _set_thread_name(self.name)
```
pitrou, line 933: This should probably be more sophisticated, e.g. `f"{sys.executable} [thread {self.name!r}]"`

Python/thread_pthread.h
```diff
@@ -343,6 +344,37 @@ PyThread_get_thread_native_id(void)
...
+    size_t len = Py_ARRAY_LENGTH(buf) - 1;
+    strncpy(buf, name, len);
+    buf[len] = '\0';
+#ifndef __APPLE__
+    unsigned long ident = PyThread_get_thread_ident();
+#endif
+    int ret = 0;
+#ifdef __APPLE__
+    // On macOS, pthread_setname_np() can only set the calling thread's name.
+    ret = pthread_setname_np(buf);
```
encukou, line 365: It seems (from the man page) that macOS doesn't have the 16-character limitation. Could we pass the original buffer here?

Lib/threading.py
encukou: That wouldn't work well for platforms with the 16-character limitation.

encukou (commented):
This needs tests, which in turn will need a getter function -- at least a private one.

vstinner:
New attempt to implement this feature: https://github.com/python/cpython/pull/127338

vstinner:
https://github.com/python/cpython/pull/127338 was merged instead. Thanks for your contribution anyway!

PR #14579, opened 2019-07-04
gh-81678: Add inspect.Signature.from_text()

ericsnowcurrently:
https://bugs.python.org/issue37497

* Issue: gh-81678

### DIFF
--- a/Doc/library/inspect.rst
+++ b/Doc/library/inspect.rst
@@ -550,7 +550,7 @@ The Signature object represents the call signature of a callable object and its
 return annotation.  To retrieve a Signature object, use the :func:`signature`
 function.
 
-.. function:: signature(callable, \*, follow_wrapped=True)
+.. function:: signature(callable_or_text, \*, follow_wrapped=True)
 
    Return a :class:`Signature` object for the given ``callable``::
 
@@ -570,7 +570,8 @@ function.
       <class 'int'>
 
    Accepts a wide range of Python callables, from plain functions and classes to
-   :func:`functools.partial` objects.
+   :func:`functools.partial` objects.  Text (string) signature are also
+   supported.  See :func:`Signature.from_text`.
 
    Raises :exc:`ValueError` if no signature can be provided, and
    :exc:`TypeError` if that type of object is not supported.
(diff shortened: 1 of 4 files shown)


Lib/inspect.py
```diff
@@ -1976,7 +1990,9 @@ def _signature_fromstr(cls, obj, s, skip_bound_arg=True):
         module = None
 
     if not isinstance(module, ast.Module):
-        raise ValueError("{!r} builtin has invalid signature".format(obj))
+        exc = ValueError(f"invalid signature {text!r}")
```
tirkarthi, line 1993: Commenting the if block on master and in this PR has no test failure so I guess this code was not tested. Perhaps a test could be added where `__text_signature__` has a invalid syntax as signature to see if `ValueError` is raised like below : 

```python
def test_python_function_invalid_text_signature(self):
    def func(*args, **kwargs):
        pass
    func.__text_signature__ = '($self, **kwargs, a)'
    with self.assertRaises(ValueError) as e:
        sig = inspect.signature(func)
```

Doc/library/inspect.rst
```diff
@@ -550,7 +550,7 @@ The Signature object represents the call signature of a callable object and its
 return annotation.  To retrieve a Signature object, use the :func:`signature`
 function.
 
-.. function:: signature(callable, \*, follow_wrapped=True)
+.. function:: signature(callable_or_text, \*, follow_wrapped=True)
```
1st1, line 553: Sorry, I don't like the `signature()` function to accept both callables and text. This introduces extra complexity to the API for a very narrow use case. 

It also makes the function too dynamic: currently, if a text value is passed to `signature()` by mistake the user would get an exception saying that a callable was expected. With this patch they'll likely get a ParseError.

IMO a public documented `Signature.from_text()` API is enough.

Lib/inspect.py
```diff
@@ -1955,18 +1955,32 @@ def _signature_strip_non_python_syntax(signature):
...
     # Lazy import ast because it's relatively heavy and
     # it's not used for other than this function.
     import ast
 
     Parameter = cls._parameter_cls
 
-    clean_signature, self_parameter, last_positional_only = \
-        _signature_strip_non_python_syntax(s)
+    (clean_signature, self_parameter, last_positional_only
+     ) = _signature_strip_non_python_syntax(text)
```
1st1, line 1983: nit: I'd keep the `\` :)

1st1 (changes requested):

PR #14583, opened 2019-07-04
Modify helper class in pickletest, to be a context manager

pradyunsg:
This allows for cleaner code style and minimizes the possibility of using the helper incorrectly.

Not sure if we'd want a b.p.o. issue for this.


tirkarthi:
cc: @pitrou

pradyunsg:
Closing since I don't have time to finish this.

PR #14587, opened 2019-07-04
Update failing URL for ThreadPoolExecutor example

yeraydiazdiaz:
The ThreadPoolExecutor example includes a previously failing URL that actually returns content from a parked domain now, so the exception handling path is never executed.

### DIFF
--- a/Doc/library/concurrent.futures.rst
+++ b/Doc/library/concurrent.futures.rst
@@ -182,7 +182,7 @@ ThreadPoolExecutor Example
            'http://www.cnn.com/',
            'http://europe.wsj.com/',
            'http://www.bbc.co.uk/',
-           'http://some-made-up-domain.com/']
+           'http://some-made-up-domain-that-definitely-does-not-exist.com/']
 
    # Retrieve a single page and report the URL and contents
    def load_url(url, timeout):


the-knights-who-say-ni:

mangrisano (approved):
LGTM.

Doc/library/concurrent.futures.rst
```diff
@@ -182,7 +182,7 @@ ThreadPoolExecutor Example
            'http://www.cnn.com/',
            'http://europe.wsj.com/',
            'http://www.bbc.co.uk/',
-           'http://some-made-up-domain.com/']
+           'http://some-made-up-domain-that-definitely-does-not-exist.com/']
```
tirkarthi, line 185: There is a chance that someone might claim this domain too in future. Given that the example is for illustration purpose and not a hard doctest failure I am not sure of this change.

benjaminp: It would be better to use something explicitly reserved like [`domain.invalid`](https://en.wikipedia.org/wiki/.invalid).

yeraydiazdiaz: Sure, I was simply respecting the original spirit of the example.

PR #14589, opened 2019-07-04
bpo-29312: use METH_FASTCALL for dict.update

jdemeyer:
https://bugs.python.org/issue29312

### DIFF
--- a/Lib/test/test_dict.py
+++ b/Lib/test/test_dict.py
@@ -142,9 +142,11 @@ class DictTest(unittest.TestCase):
         d.update({2:20})
         d.update({1:1, 2:2, 3:3})
         self.assertEqual(d, {1:1, 2:2, 3:3})
+        d.update(([i,i] for i in range(4,6)), key="value")
+        self.assertEqual(d, {1: 1, 2: 2, 3: 3, 4: 4, 5: 5, 'key': 'value'})
 
         d.update()
-        self.assertEqual(d, {1:1, 2:2, 3:3})
+        self.assertEqual(d, {1: 1, 2: 2, 3: 3, 4: 4, 5: 5, 'key': 'value'})
 
         self.assertRaises((TypeError, AttributeError), d.update, None)
 
--- a/Objects/dictobject.c
+++ b/Objects/dictobject.c
@@ -2340,16 +2340,6 @@ dict_update_common(PyObject *self, PyObject *args, PyObject *kwds,
     return result;
 }
 
-/* Note: dict.update() uses the METH_VARARGS|METH_KEYWORDS calling convention.
-   Using METH_FASTCALL|METH_KEYWORDS would make dict.update(**dict2) calls
-   slower, see the issue #29312. */
-static PyObject *
-dict_update(PyObject *self, PyObject *args, PyObject *kwds)
-{
-    if (dict_update_common(self, args, kwds, "update") != -1)
-        Py_RETURN_NONE;
-    return NULL;
-}
 
 /* Update unconditionally replaces existing items.
    Merge has a 3rd argument 'override'; if set, it acts like Update,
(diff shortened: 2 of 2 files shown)


jdemeyer:
Here are some benchmarks:
1. `./python -m perf timeit --duplicate 200 -s 'L = [("x",None)]' 'dict(L)'`
**before**: Mean +- std dev: 221 ns +- 1 ns
**after**: Mean +- std dev: 186 ns +- 1 ns
2. `./python -m perf timeit --duplicate 200 -s 'L = [("x",None)]' '{}.update(L)'`
**before**: Mean +- std dev: 189 ns +- 2 ns
**after**: Mean +- std dev: 161 ns +- 1 ns
3. `./python -m perf timeit --duplicate 200 -s 'D = {"x":None}' 'dict(D)'`
**before**: Mean +- std dev: 196 ns +- 1 ns
**after**: Mean +- std dev: 126 ns +- 1 ns
4. `./python -m perf timeit --duplicate 200 -s 'D = {"x":None}' '{}.update(D)'`
**before**: Mean +- std dev: 174 ns +- 1 ns
**after**: Mean +- std dev: 116 ns +- 0 ns
5. `./python -m perf timeit --duplicate 200 'dict(x=None)'`
**before**: Mean +- std dev: 200 ns +- 1 ns
**after**: Mean +- std dev: 119 ns +- 1 ns
6. `./python -m perf timeit --duplicate 200 '{}.update(x=None)'`
**before**: Mean +- std dev: 201 ns +- 1 ns
**after**: Mean +- std dev: 120 ns +- 1 ns
7. `./python -m perf timeit --duplicate 200 -s 'D = {"x":None}' 'dict(**D)'`
**before**: Mean +- std dev: 149 ns +- 1 ns
**after**: Mean +- std dev: 171 ns +- 1 ns
8. `./python -m perf timeit --duplicate 200 -s 'D = {"x":None}' '{}.update(**D)'`
**before**: Mean +- std dev: 139 ns +- 1 ns
**after**: Mean +- std dev: 171 ns +- 1 ns
9. `./python -m perf timeit --duplicate 200 -s 'D = {f"x{i}":None for i in range(10000)}' 'dict(**D)'`
**before**: Mean +- std dev: 321 us +- 1 us
**after**: Mean +- std dev: 393 us +- 3 us
10. `./python -m perf timeit --duplicate 200 -s 'D = {f"x{i}":None for i in range(10000)}' '{}.update(**D)'`
**before**: Mean +- std dev: 321 us +- 3 us
**after**: Mean +- std dev: 394 us +- 4 us

So the calls `dict(**D)` and `dict.update(**D)` got slower but the rest got faster.

vstinner (commented):
Your PR title says "use METH_FASTCALL for dict.update" but you modified PyDictType.tp_init as well. Can you please move your tp_init change in a separated PR?

jdemeyer:
I changed the branch to match the title, reverting the changes to `__init__`.

Note that this PR has more code duplication: the code paths for `dict.update` and `dict.__init__` used to be the same but that's no longer the case with this PR. When `dict.__init__` is also changed (as in my original PR), the code paths will be the same again.

Lib/test/test_dict.py
```diff
@@ -141,10 +141,11 @@ def test_update(self):
         d.update({1:100})
         d.update({2:20})
         d.update({1:1, 2:2, 3:3})
-        self.assertEqual(d, {1:1, 2:2, 3:3})
```
vstinner: Can you please keep self.assertEqual(d, {1:1, 2:2, 3:3})?

Lib/test/test_dict.py
```diff
@@ -212,6 +213,15 @@ def __next__(self):
 
         self.assertRaises(ValueError, {}.update, [(1, 2, 3)])
 
+    def test_init(self):
```
vstinner: That's unrelated to this PR, right?

Objects/dictobject.c
```diff
@@ -2591,6 +2581,102 @@ dict_merge(PyObject *a, PyObject *b, int override)
...
+dict_merge_from_any(PyObject *self, PyObject *obj)
+{
+    /* Fast path for dict subclasses */
+    if (PyDict_Check(obj)) {
+        return dict_merge(self, obj, 1);
+    }
+
+    _Py_IDENTIFIER(keys);
+    PyObject *func;
+    if (_PyObject_LookupAttrId(obj, &PyId_keys, &func) < 0) {
```
vstinner, line 2598: That sounds redundant with dict_merge() which calls PyMapping_Keys(). Maybe dict_merge() could be modified to dispatch to PyDict_MergeFromSeq2() if PyMapping_Keys() keys with an attribute error, rather than trying to get the attribute twice?

It's not directly related to your change, it can be done/discussed in a separated PR.

Objects/dictobject.c
```diff
@@ -2591,6 +2581,102 @@ dict_merge(PyObject *a, PyObject *b, int override)
...
+    if (USABLE_FRACTION(mp->ma_keys->dk_size) < n) {
+        if (dictresize(mp, ESTIMATE_SIZE(mp->ma_used + n))) {
+           return -1;
+        }
+    }
+
+    for (Py_ssize_t i = 0; i < n; i++) {
+        PyObject *key = keys[i];
+        if (!PyUnicode_Check(key)) {
+            PyErr_SetString(PyExc_TypeError, "keywords must be strings");
```
vstinner: Do we really have to check here that the FASTCALL protocol is used properly? Maybe use an assertion instead? What do you think @methane?

Objects/dictobject.c
```diff
@@ -2591,6 +2581,102 @@ dict_merge(PyObject *a, PyObject *b, int override)
...
+     * that there will be no (or few) overlapping keys.
+     */
+    if (USABLE_FRACTION(mp->ma_keys->dk_size) < n) {
+        if (dictresize(mp, ESTIMATE_SIZE(mp->ma_used + n))) {
+           return -1;
+        }
+    }
+
+    for (Py_ssize_t i = 0; i < n; i++) {
+        PyObject *key = keys[i];
```
vstinner, line 2626: Would you mind to add `PyObject *value = values[i];` to better show that we iterate on zip(keys, values)?

Objects/dictobject.c
jdemeyer: > Do we really have to check here that the FASTCALL protocol is used properly?

This is currently the only place where this is checked. If we remove this check here, then we must check it somewhere else (unless you are suggesting that `dict.update(**d)` should be allowed even if `d` has non-string keys).

There is a test for this in `test_dict.py`:
```python
    def test_invalid_keyword_arguments(self):
        class Custom(dict):
            pass
        for invalid in {1 : 2}, Custom({1 : 2}):
            with self.assertRaises(TypeError):
                dict(**invalid)
            with self.assertRaises(TypeError):
                {}.update(**invalid)
```

Lib/test/test_dict.py
jdemeyer: It's mostly unrelated but I assumed that these tests are good to have anyway. Should I remove them? (when you ask a question like that, it's not always easy to understand whether it's just a question or whether you want something to be fixed).

Objects/dictobject.c
jdemeyer: I agree, I also noticed this. There is some room for improvement here. Let's do it in a new PR after this is merged.

Objects/dictobject.c
vstinner: For METH_VARARGS and METH_VARARGS|METH_KEYWORDS, we don't check arguments type at runtime in release mode. But there are plenty of assertions. Example:

```
PyObject *
PyObject_Call(PyObject *callable, PyObject *args, PyObject *kwargs)
{
    assert(!PyErr_Occurred());
    assert(PyTuple_Check(args));
    assert(kwargs == NULL || PyDict_Check(kwargs));
```

METH_VARARGS|METH_KEYWORDS should only get strings for kwargs keys, but that isn't checked neither.

If you misuse the Python C API, you are on your own: Python is likely to crash, and I consider that it's the expected behavior.

jdemeyer: I don't understand what you're trying to suggest here. Simply removing the check is not an option, as it causes testsuite failures.

jdemeyer: Interestingly, `OrderedDict` does allow non-string keys:
```
>>> from collections import OrderedDict
>>> OrderedDict(**{1:2})
OrderedDict([(1, 2)])
```
This is probably a bug, but it shows that nothing in the core interpreter prevents non-string keys.

jdemeyer: I just posted to python-dev about this issue.

jdemeyer: Fixed by #14682

jdemeyer:
Rebased to latest master, using #14682.

vstinner:
@methane: I let you decide what do to do with this change.

Objects/dictobject.c
```diff
@@ -2591,6 +2581,100 @@ dict_merge(PyObject *a, PyObject *b, int override)
...
+ */
+static int
+dict_merge_from_any(PyObject *self, PyObject *obj)
+{
+    /* Fast path for dict subclasses */
+    if (PyDict_Check(obj)) {
+        return dict_merge(self, obj, 1);
+    }
+
+    _Py_IDENTIFIER(keys);
```
methane, line 2596: Please move this identifier to global static variable.  I don't like two same identifiers in single source file.

iritkatriel:
https://bugs.python.org/issue29312 is closed. What is that status of this PR?

methane:
Yes. This PR is also rejected.

PR #14599, opened 2019-07-05
bpo-37483: fix refleak

methane:

### DIFF
--- a/Python/codecs.c
+++ b/Python/codecs.c
@@ -175,9 +175,11 @@ PyObject *_PyCodec_Lookup(const char *encoding)
         Py_DECREF(result);
         goto onError;
     }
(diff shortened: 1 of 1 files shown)


jdemeyer (changes requested):
You need to initialize `v = NULL` and you need to handle the case `v == NULL` in `onError`. See #14600

methane:
> You need to initialize `v = NULL` and you need to handle the case `v == NULL` in `onError`. See #14600

Why?

```
    v = normalizestring(encoding);
    if (v == NULL)
        goto onError;
    PyUnicode_InternInPlace(&v);
```

methane:
I'm sorry, I found it myself.

PR #14600, opened 2019-07-05
bpo-37483: fix reference leak in _PyCodec_Lookup

jdemeyer:
CC @vstinner 

### DIFF
--- a/Python/codecs.c
+++ b/Python/codecs.c
@@ -99,40 +99,38 @@ PyObject *normalizestring(const char *string)
 
 PyObject *_PyCodec_Lookup(const char *encoding)
 {
-    PyObject *result, *v;
-    Py_ssize_t i, len;
-
     if (encoding == NULL) {
         PyErr_BadArgument();
-        goto onError;
+        return NULL;
     }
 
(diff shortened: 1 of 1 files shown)


Python/codecs.c
```diff
@@ -127,8 +127,7 @@ PyObject *_PyCodec_Lookup(const char *encoding)
         return result;
     }
     else if (PyErr_Occurred()) {
-        Py_DECREF(v);
-        return NULL;
+        goto onError;
     }
```
vstinner, line 130: I would prefer to clear v reference just after the PyDict_GetItemWithError() call. There is no need to keep alive.

I also suggest to move "v" variable declaration to where it's defined ("v = ...").

jdemeyer: > I would prefer to clear v reference just after the PyDict_GetItemWithError() call.

That doesn't work, the string `v` is used in multiple places. We really need to keep it alive until the end.

vstinner: > That doesn't work, the string v is used in multiple places. We really need to keep it alive until the end.

Oh ok, I didn't notice.

PR #14603, opened 2019-07-05
bpo-37151: remove _PyMethodDef_RawFastCall* functions

jdemeyer:
Remove some dead code after PEP 590 refactoring.

CC @encukou 


jdemeyer:
Should this have a `NEWS` entry? It's removing private undocumented functions.

encukou:
I don't think that's necessary, especially for functions only introduced in 3.7.

PR #14606, opened 2019-07-05
bpo-37504: Fix documentation build with texinfo builder

mitya57:
In the table model used by docutils, the `cols` attribute of `tgroup` nodes is [mandatory][1]. It is used in texinfo builder [here][2].

Here is a screenshot proving that this change works:
Screenshot of `info` showing the Python documentation

[1]: https://www.oasis-open.org/specs/tm9901.htm#AEN348
[2]: https://github.com/sphinx-doc/sphinx/blob/v2.1.2/sphinx/writers/texinfo.py#L1129

https://bugs.python.org/issue37504

### DIFF
--- a/Doc/tools/extensions/pyspecific.py
+++ b/Doc/tools/extensions/pyspecific.py
@@ -536,6 +536,7 @@ def process_audit_events(app, doctree, fromdocname):
         nodes.colspec(colwidth=30),
         nodes.colspec(colwidth=55),
         nodes.colspec(colwidth=15),
+        cols=3,
     )
     head = nodes.thead()
     body = nodes.tbody()
--- a/Doc/Makefile
+++ b/Doc/Makefile
@@ -16,7 +16,7 @@ SPHINXERRORHANDLING = -W
 ALLSPHINXOPTS = -b $(BUILDER) -d build/doctrees -D latex_elements.papersize=$(PAPER) \
                 $(SPHINXOPTS) $(SPHINXERRORHANDLING) . build/$(BUILDER) $(SOURCES)
 
-.PHONY: help build html htmlhelp latex text changes linkcheck \
+.PHONY: help build html htmlhelp latex text texinfo changes linkcheck \
 	suspicious coverage doctest pydoc-topics htmlview clean dist check serve \
 	autobuild-dev autobuild-stable venv
 
@@ -29,6 +29,7 @@ help:
 	@echo "  htmlhelp   to make HTML files and a HTML help project"
 	@echo "  latex      to make LaTeX files, you can set PAPER=a4 or PAPER=letter"
 	@echo "  text       to make plain text files"
+	@echo "  texinfo    to make Texinfo file"
 	@echo "  epub       to make EPUB files"
 	@echo "  changes    to make an overview over all changed/added/deprecated items"
 	@echo "  linkcheck  to check all external links for integrity"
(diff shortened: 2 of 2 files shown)


aeros (changes requested):
@mitya57 Thanks for the contribution! In order to fully verify that the change is working appropriately since this is addressing a specific bug (Documentation fails to build when using Sphinx' textinfo builder), I would advise advise updating the tests (which would be included in the PR). A screenshot does provide a solid visual example, but is usually not reliable enough to show that it fully functions as described in the initial issue. This also helps to ensure that if your addition works as intended, no other changes will unintentionally break it. 

The above suggestions might not be necessary in order for your PR to be merged, but in general adding tests is always helpful, especially when it involves a change in functionality or fixing a bug that was under the radar of the existing tests.

mitya57:
Now the texinfo file is built successfully during Travis test, see <https://travis-ci.org/python/cpython/jobs/555099128#L283>.

aeros (approved):
Thanks for updating the tests.

Doc/tools/extensions/pyspecific.py
```diff
@@ -536,6 +536,7 @@ def process_audit_events(app, doctree, fromdocname):
         nodes.colspec(colwidth=30),
         nodes.colspec(colwidth=55),
         nodes.colspec(colwidth=15),
+        cols=3
```
asottile: nice, I had almost the same patch for deadsnakes (ubuntu packaging) but opted for a trailing comma here: https://github.com/deadsnakes/python3.8/blob/dc6234da93ec2b4be6535209a9b63a4541217c8c/debian/patches/0034-fix-docs-build.patch

asottile (commented):
sorry for the nitpick / drive-by -- looks good!

Doc/tools/extensions/pyspecific.py
mitya57: OK, added a comma.

asottile (approved):

mitya57:
@zooba @tirkarthi As you were the author and reviewer of #14406 which added this `nodes.tgroup` construct, please review this too.

The Travis build failed with a temporary error, I guess it can be retried.

As your original PR is not part of any stable release, I think the “skip news” label can be added.

aeros167:
> The Travis build failed with a temporary error, I guess it can be retried.

You can retry the travis build by adding and removing an empty comment `#` to ``pyspecific.py``, either at the end of an existing line or a new one.

tirkarthi (approved):
Thanks, I didn't know about texinfo and reading it on Emacs is nice. I am okay with the changes as per the reference and running locally also builds fine. I am not sure about running `make texinfo` on primary CI like Travis. I would defer to @JulienPalard  on that.

mitya57:
> I am not sure about running `make texinfo` on primary CI like Travis.

It was not my idea, but requested by @aeros167.

aeros167:
> I would defer to @JulienPalard on that.
> It was not my idea, but requested by @aeros167.

From what I recall, that suggestion was just to add further testing, I'm not certain as to the best way to do it with regards to texinfo or if it should even be done through travis. Julien would definitely be far better to ask, also @willingc since she's a [documentation expert](https://devguide.python.org/experts/#miscellaneous).

willingc:
Thanks @mitya57 for the pull request.  A few thoughts on the changes:

- I don't think it is necessary to add the texinfo to Travis since we do not currently build LaTeX or EPUB on Travis. @JulienPalard, do you see any reason to add all of the formats? 
- On the Makefile, adding `texinfo` to the `.PHONY` line would be a good practice
- Also, the Makefile has commands toward the end to archive the different format. Probably a good idea to add that for texinfo as well.

willingc (commented):
Please update as suggested in my earlier message. Thanks!

mitya57:
I have now rebased on latest master and updated as suggested by @willingc.

PR #14609, opened 2019-07-05
bpo-37511: mmap module add OpenBSD madvise flag.

ghost:
Similar role as MADV_NOCORE.

https://bugs.python.org/issue37511

### DIFF
--- a/Doc/library/mmap.rst
+++ b/Doc/library/mmap.rst
@@ -328,6 +328,7 @@ MADV_* Constants
           MADV_NOHUGEPAGE
           MADV_DONTDUMP
           MADV_DODUMP
+          MADV_CONCEAL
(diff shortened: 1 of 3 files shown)


jdemeyer:
What does `MADV_CONCEAL` do? I googled it and the only hits came from the Python bug and PR.

jdemeyer:
And it's not listed on https://man.openbsd.org/madvise.2 so there is something fishy with this PR.

dcarlier-afilias:
Ah I made a typo (and out of place too). Going to fix it.

jdemeyer (approved):
Looks good to me, but maybe add it to #14513 instead?

dcarlier-afilias:
Ok if preferred I ll move it there.

PR #14615, opened 2019-07-06
bpo-37513: Fix a type error in ctypes.rst

shihai1991:
https://bugs.python.org/issue37513


aeros (approved):
@shihai1991 Thanks for the contribution! Verified that ``POINT(1, 2, 3)`` causes a ``TypeError`` instead of a ``ValueError``:
image

shihai1991:
> @shihai1991 Thanks for the contribution! Verified that `POINT(1, 2, 3)` causes a `TypeError` instead of a `ValueError`:
> image

Thanks for review and verify :)

zhangyangyu (approved):

PR #14627, opened 2019-07-06
bpo-37456: Slash ('/') is now part of syntax

terryjreedy:
https://bugs.python.org/issue37456

### DIFF
--- a/Doc/faq/programming.rst
+++ b/Doc/faq/programming.rst
@@ -800,10 +800,6 @@ an error::
      File "<stdin>", line 1, in <module>
    TypeError: pow() takes no keyword arguments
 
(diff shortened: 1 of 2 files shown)


aeros (commented):
I think this would be worthwhile to backport to versions prior to 3.8 as well. I'm not as to the exact version positional-only keyword arguments have been around, but haven't they been around for a significant period of time now?

terryjreedy:
No, / has only be valid in code since 3.8.0b1.  The note is correct for 3.7.

aeros167:
My apologies, I misunderstood the purpose of the PR. I definitely should have referred to the section above it in ``programming.rst`` before submitting the review.

PR #14631, opened 2019-07-07
bpo-37478: Add missing 'and'.

terryjreedy:
https://bugs.python.org/issue37478


terryjreedy:
Followup to #14611.

aeros (commented):
@terryjreedy Is "and" appropriate in this case? Based on the discussion in the PR which added the exception description, it seemed like `os.chdir()` could not not directly raise `OSError` itself, only its subclasses.

terryjreedy:
Yes, Eryk's says it *can* raise OSError is a situation where none of the subclasses fits.

PR #14634, opened 2019-07-07
bpo-21120: Exclude Python-ast.h, ast.h and asdl.h from the limited API

ZackerySpytz:
The PyArena type is not part of the limited API, so these headers
shouldn't be part of it either.

https://bugs.python.org/issue21120


jdemeyer (approved):
This is really just documenting what's already the case (this functionality is already not part of the limited API by PEP 384), so looks good to me.

PR #14640, opened 2019-07-07
Doc: Fix missing negation.

JulienPalard:
Reported by Hug Capella on docs@.

Automerge-Triggered-By: @matrixise

### DIFF
--- a/Doc/library/re.rst
+++ b/Doc/library/re.rst
@@ -540,8 +540,8 @@ character ``'$'``.
    Matches any character which is not a word character. This is
    the opposite of ``\w``. If the :const:`ASCII` flag is used this
    becomes the equivalent of ``[^a-zA-Z0-9_]``.  If the :const:`LOCALE` flag is
-   used, matches characters considered alphanumeric in the current locale
-   and the underscore.
+   used, matches characters neither considered alphanumeric in the current locale
+   nor the underscore.
 
 .. index:: single: \Z; in regular expressions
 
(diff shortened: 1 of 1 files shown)


aeros (changes requested):
As a minor phrasing improvement suggestion, I would recommend changing it from:

> ...  If the :const:`LOCALE` flag is used, matches characters considered alphanumeric in the current locale and the underscore.

to:
> ...  If the :const:`LOCALE` flag is used, matches characters which are considered alphanumeric in the current locale and the underscore.

eamanu (approved):
LGTM

JulienPalard:
@aeros167 So you mean you're suggestion is to change:

> If the	:const:`LOCALE` flag is
   used, matches characters neither considered alphanumeric in the current locale
   nor the underscore.

to:

> If the	:const:`LOCALE` flag is
   used, matches characters which are neither considered alphanumeric in the current locale
   nor the underscore.

?

aeros167:
@JulienPalard I'll submit the actual suggestion instead of just leaving it as a comment so it's easier to see

Doc/library/re.rst
```diff
@@ -540,8 +540,8 @@ character ``'$'``.
    Matches any character which is not a word character. This is
    the opposite of ``\w``. If the :const:`ASCII` flag is used this
    becomes the equivalent of ``[^a-zA-Z0-9_]``.  If the :const:`LOCALE` flag is
-   used, matches characters considered alphanumeric in the current locale
-   and the underscore.
+   used, matches characters neither considered alphanumeric in the current locale
```
aeros:
```suggestion
   used, matches characters which are neither alphanumeric in the current locale
```
Also removed "considered", unnecessary filler word in this case.

aeros (changes requested):

matrixise (approved):

PR #14646, opened 2019-07-08
bpo-37055: fix warnings in _blake2 module

methane:
https://bugs.python.org/issue37055


methane:
@tiran Could you review this?

tiran (approved):

PR #14647, opened 2019-07-08
gh-81368: Update PyCField_get_size

shihai1991:
https://bugs.python.org/issue37187

* Issue: gh-81368

### DIFF
--- a/Misc/NEWS.d/next/Core and Builtins/2019-07-08-17-11-20.bpo-37187.-KLvmv.rst
+++ b/Misc/NEWS.d/next/Core and Builtins/2019-07-08-17-11-20.bpo-37187.-KLvmv.rst
@@ -1,1 +1,1 @@
-Print bit width in :c:func:`PyCField_get_size`
+Return bit width in ``PyCField_get_size``
--- a/Misc/NEWS.d/next/Core and Builtins/2019-07-08-17-11-20.bpo-37187.-KLvmv.rst
+++ b/Misc/NEWS.d/next/Core and Builtins/2019-07-08-17-11-20.bpo-37187.-KLvmv.rst
@@ -1,0 +1,1 @@
+Print bit width in :c:func:`PyCField_get_size`
--- a/Modules/_ctypes/cfield.c
+++ b/Modules/_ctypes/cfield.c
@@ -247,7 +247,6 @@ PyCField_get_offset(PyObject *self, void *data)
 static PyObject *
 PyCField_get_size(PyObject *self, void *data)
 {
-    PyObject *result;
     Py_ssize_t bits = ((CFieldObject *)self)->size >> 16;
     if (bits) {
         return PyLong_FromSsize_t(bits); 
--- a/Modules/_ctypes/cfield.c
+++ b/Modules/_ctypes/cfield.c
@@ -250,11 +250,9 @@ PyCField_get_size(PyObject *self, void *data)
     PyObject *result;
     Py_ssize_t bits = ((CFieldObject *)self)->size >> 16;
     if (bits) {
-        result = PyLong_FromSsize_t(bits); 
-    } else {
-        result = PyLong_FromSsize_t(((CFieldObject *)self)->size);
+        return PyLong_FromSsize_t(bits); 
     }
-    return result;
+    return PyLong_FromSsize_t(((CFieldObject *)self)->size);
 }
 
 static PyGetSetDef PyCField_getset[] = {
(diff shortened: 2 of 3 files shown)


eric-wieser:
It's not clear to me that this is the right fix. With this change, it's now impossible to get the bit offset of a struct field.

Modules/_ctypes/cfield.c
```diff
@@ -247,7 +247,14 @@ PyCField_get_offset(PyObject *self, void *data)
 static PyObject *
 PyCField_get_size(PyObject *self, void *data)
 {
-    return PyLong_FromSsize_t(((CFieldObject *)self)->size);
+    PyObject *result;
+    Py_ssize_t bits = ((CFieldObject *)self)->size >> 16;
+    if (bits) {
+        result = PyLong_FromSsize_t(bits); 
+    } else {
+        result = PyLong_FromSsize_t(((CFieldObject *)self)->size);
```
eamanu: You could return without create the *result pointer:

```
if(bits){
    return PyLong_FromSsize_t(bits);
}
...
```

shihai1991:
> It's not clear to me that this is the right fix. With this change, it's now impossible to get the bit offset of a struct field.

Getting offset should be use https://github.com/python/cpython/blob/master/Modules/_ctypes/cfield.c#L254

Misc/NEWS.d/next/Core and Builtins/2019-07-08-17-11-20.bpo-37187.-KLvmv.rst
```diff
@@ -0,0 +1 @@
+Print bit width in :c:func:`PyCField_get_size`
```
ZackerySpytz: I believe `c:func` should only be used for the documented functions of the C API.

eric-wieser:
> Getting offset should be use https://github.com/python/cpython/blob/master/Modules/_ctypes/cfield.c#L254

What if I want the bit offset and not the byte offset? Consider:
```python
class S(ctypes.Structure):
    _fields_ = [
        ('a', ctypes.c_uint8),
        ('b', ctypes.c_uint8, 6),
        ('c', ctypes.c_uint8, 2)
    ]
```
Previously, I could do:
```
>>> f = S.c
>>> f.size >> 16  # number of bits
2
>>> f.size & 0xFFFF  # bit offset
6
>>> f.offset  # byte offset
1
```

How would I calculate those three values with your patch?

Misc/NEWS.d/next/Core and Builtins/2019-07-08-17-11-20.bpo-37187.-KLvmv.rst
shihai1991: en, looks somebody use this tag too. I am not sure about it.
https://github.com/python/cpython/blob/master/Misc/NEWS.d/next/Core%20and%20Builtins/2019-06-17-03-53-16.bpo-37316.LytDX_.rst

shihai1991:
> > Getting offset should be use https://github.com/python/cpython/blob/master/Modules/_ctypes/cfield.c#L254
> 
> What if I want the bit offset and not the byte offset? Consider:
> 
> ```python
> class S(ctypes.Structure):
>     _fields_ = [
>         ('a', ctypes.c_uint8),
>         ('b', ctypes.c_uint8, 6),
>         ('c', ctypes.c_uint8, 2),
>     ]
> ```
> 
> Previously, I could do:
> 
> ```python
> >>> S.c.size >> 16  # number of bits
> 2
> >>> S.c.size & 0xFFFF  # bit offset
> 6
> >>> S.c.offset  # byte offset
> 1
> ```
> 
> Obviously this API didn't make much sense from the current documentation, but it was at least possible to extract this information. How would I calculate those three values with your patch?

en, I think it should be a inner behavior. Why the user should know this detail?

Misc/NEWS.d/next/Core and Builtins/2019-07-08-17-11-20.bpo-37187.-KLvmv.rst
zooba: Zackery is correct, the tag is only for documented functions (the other example - `PySys_Audit` - is documented).
```suggestion
Return bit width in ``PyCField_get_size``
```

shihai1991: Thanks for review, Steve. Done ;)

PR #14648, opened 2019-07-08
bpo-27679: Remove set_bitfields()

shihai1991:
https://bugs.python.org/issue27679

### DIFF
--- a/Modules/_ctypes/_ctypes_test.c
+++ b/Modules/_ctypes/_ctypes_test.c
@@ -425,30 +425,6 @@ struct BITS {
 #endif
 };
 
-EXPORT(void) set_bitfields(struct BITS *bits, char name, int value)
-{
-    switch (name) {
-    case 'A': bits->A = value; break;
-    case 'B': bits->B = value; break;
-    case 'C': bits->C = value; break;
-    case 'D': bits->D = value; break;
-    case 'E': bits->E = value; break;
-    case 'F': bits->F = value; break;
-    case 'G': bits->G = value; break;
-    case 'H': bits->H = value; break;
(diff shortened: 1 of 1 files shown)


shihai1991:
Looks like nobody need this `set_bitfields()` function.
If I was wrong , pls help me close this PR, thanks.

eamanu (approved):
LGTM

aeros167:
> Looks like nobody need this `set_bitfields()` function.
> If I was wrong , pls help me close this PR, thanks.

In order to close a PR, simply enter in a final comment such as "Closing PR because ..." and instead of selecting the `Comment` option, choose the option to the left, `Close and comment`.

shihai1991:
> > Looks like nobody need this `set_bitfields()` function.
> > If I was wrong , pls help me close this PR, thanks.
> 
> In order to close a PR, simply enter in a final comment such as "Closing PR because ..." and instead of selecting the `Comment` option, choose the option to the left, `Close and comment`.

Thanks, got it;)

aeros (commented):
In a search of the cpython repository for `set_bitfields`, this is the only occurrence. This is the initial message in the commit which first created the function 12 years ago: "Make _ctypes_test.c compile on Windows" without any further details or documentation. 

If the integrations tests are still fully passing without it and it is not used in any other location, as far as I can tell, it can be safely removed. However, since this is not at all my area of expertise, I don't want to explicitly approve this PR, in case this function is still useful in some way that I'm unaware of. @vstinner would probably have a more definitive answer.

vstinner:
As soon as tests still pass, the change LGTM :-) Thanks.

PR #14657, opened 2019-07-09
bpo-26806: IDLE should run without docstrings

terryjreedy:
After fcf1d00, IDLE startup failed with python compiled without docstrings.

https://bugs.python.org/issue26806

### DIFF
--- a/Lib/idlelib/idle_test/test_run.py
+++ b/Lib/idlelib/idle_test/test_run.py
@@ -292,6 +292,14 @@ class TestSysRecursionLimitWrappers(unittest.TestCase):
         new_reclimit = sys.getrecursionlimit()
         self.assertEqual(new_reclimit, orig_reclimit)
 
+    def test_fixdoc(self):
+        def func(): "docstring"
+        run.fixdoc(func, "more")
+        self.assertEqual(func.__doc__, "docstring\n\nmore")
+        func.__doc__ = None
+        run.fixdoc(func, "more")
+        self.assertEqual(func.__doc__, "more")
+
 
 if __name__ == '__main__':
     unittest.main(verbosity=2)
--- a/Lib/idlelib/run.py
+++ b/Lib/idlelib/run.py
@@ -307,7 +307,12 @@ def fix_scaling(root):
                 font['size'] = round(-0.75*size)
 
 
+def fixdoc(fun, text):
+    tem = (fun.__doc__ + '\n\n') if fun.__doc__ is not None else ''
+    fun.__doc__ = tem + textwrap.fill(textwrap.dedent(text))
+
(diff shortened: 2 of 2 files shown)


terryjreedy:
@ZackerySpytz  I added `if setrecursionlimit.__doc__ is not None:` and code still works with normal python.  I cannot imagine that it does not work with 'reduced' python, but it would still be nice if you test before I merge.
@taleinat I removed `.strip()` as it empirically is not needed.

taleinat (approved):
LGTM.

My apologies for this regression; I wasn't aware of the option to build without doc-strings.

ZackerySpytz:
Thank you for creating this fix so quickly. I have tested this PR locally, and I can confirm that IDLE now runs again when CPython is configured without docstrings.

However, I see the following in the terminal when I close the IDLE shell:
```
Exception ignored in: <idlelib.run.PseudoInputFile object at 0x7efe0b1a0c30>
  File "/home/lubuntu2/cpython/Lib/idlelib/run.py", line 488, in close
  File "/home/lubuntu2/cpython/Lib/idlelib/pyshell.py", line 1017, in close
  File "/home/lubuntu2/cpython/Lib/idlelib/editor.py", line 1019, in close
  File "/home/lubuntu2/cpython/Lib/idlelib/outwin.py", line 94, in maybesave
  File "/home/lubuntu2/cpython/Lib/idlelib/editor.py", line 952, in get_saved
AttributeError: 'NoneType' object has no attribute 'get_saved'
```

I don't know much about IDLE's internals, but this seems to be an unrelated issue.

taleinat:
> However, I see the following in the terminal when I close the IDLE shell:

@ZackerySpytz, indeed that seems like a separate issue. It would be great if you would open a new issue with details how to reproduce it. Feel free to nosy me on the issue and I'll take a look.

terryjreedy:
Tal, I was not either.  Some years ago, the test_calltips then existing were guarded against -OO, but I had not idea that builtins like list and its methods could ever disappear.  Hence the failure being fixed in #14592.

After working on that, I want to try a different fix here, adding instead
`if sys.setrecursionlimit.__doc__ is None: sys.setrecursionlimit.__doc__ = ''`
Then our wrapper will still obviously be a wrapper.

Zachery: I see that newish closing bug also, on Windows, and need to fix it.

terryjreedy:
I added a test that would have failed, with __doc__ == None, without patch.

terryjreedy:
I opened bpo 37424 for bad closing and nosied both of you.  Zachery, please check if specifc conditions are the same for you.

terryjreedy:
This problem was reported by Zachery Spitz on #13944, the PR with the flawed merge.

PR #14658, opened 2019-07-09
bpo-37523: Raise ValueError for I/O operations on a closed zipfile.ZipExtFile

danifus:
Raises `ValueError`when calling the following on a closed `zipfile.ZipExtFile`:
- `read`
- `readable`
- `seek`
- `seekable`
- `tell`

https://bugs.python.org/issue37523

### DIFF
--- a/Lib/test/test_zipfile.py
+++ b/Lib/test/test_zipfile.py
@@ -567,6 +567,20 @@ class StoredTestsWithSourceFile(AbstractTestsWithSourceFile,
             with open(TESTFN, "rb") as f:
                 self.assertEqual(zipfp.read(TESTFN), f.read())
 
+    def test_io_on_closed_zipextfile(self):
+        fname = "somefile.txt"
+        with zipfile.ZipFile(TESTFN2, mode="w") as zipfp:
+            zipfp.writestr(fname, "bogus")
+
+        with zipfile.ZipFile(TESTFN2, mode="r") as zipfp:
+            with zipfp.open(fname) as fid:
+                fid.close()
+                self.assertRaises(ValueError, fid.read)
+                self.assertRaises(ValueError, fid.seek, 0)
+                self.assertRaises(ValueError, fid.tell)
+                self.assertRaises(ValueError, fid.readable)
+                self.assertRaises(ValueError, fid.seekable)
+
     def test_write_to_readonly(self):
         """Check that trying to call write() on a readonly ZipFile object
         raises a ValueError."""
--- a/Lib/zipfile.py
+++ b/Lib/zipfile.py
@@ -877,12 +877,16 @@ class ZipExtFile(io.BufferedIOBase):
         return self._readbuffer[self._offset: self._offset + 512]
 
     def readable(self):
+        if self.closed:
+            raise ValueError("I/O operation on closed file.")
         return True
 
     def read(self, n=-1):
         """Read and return up to n bytes.
         If the argument is omitted, None, or negative, data is read and returned until EOF is reached.
         """
+        if self.closed:
+            raise ValueError("read from closed file.")
         if n is None or n < 0:
             buf = self._readbuffer[self._offset:]
             self._readbuffer = b''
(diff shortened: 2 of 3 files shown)


mangrisano:
I think the _news_ can be skipped.

danifus:
Ok, I've removed the news but I'm unsure of how to apply the "skip news" label. Thanks for the review

mangrisano:
Unfortunately we can't because we don't have the right permissions on the repo. Anyway, after removing the news file, the tests that before this operation failed now are ok. 

Thank you :)

mangrisano (approved):
LGTM.

danifus:
Ah that explains why I couldn't find the add label button!

Thanks again for the review :)

aeros167:
>Ok, I've removed the news but I'm unsure of how to apply the "skip news" label. Thanks for the review

The `skip news` label can currently only be applied by the core devs. There's a proposal in place to allow for the triagers to also be able to do this, but contributors can't add those labels. If you don't think your particular PR needs the news item, just leave it out and a core dev will let you know whether or not it's needed when they review it.

aeros (changes requested):
@danifus Thanks for the contribution! I like the idea of adding tests to ensure that `ValueError`s are appropriately raised for zipfiles, but I had some suggestions.

To make this change a bit more scale-able (in case other exceptions are added to the tests), I would recommend changing lines 575-582 to something like this:

```python
with zipfile.ZipFile(TESTFN2, mode="r") as zipfp:
    with zipfp.open(fname) as fid:
        fid.close()
        for function, *args in [fid.read, fid.tell, fid.readable, fid.seekable, 
        fid.extract, fid.testzip, fid.write, (fid.seek, 0)]:
            self.assertRaises(ValueError, function, *args)
```
Functionally, this operates the same, but is a bit more readable and easy to add additional functions to test simply by adding another one at the end of the list. ~~Also, `seek()` already defaults to `0`, so there's no need to specify it explicitly~~ (https://docs.python.org/3.9/library/io.html#io.IOBase.seek).

Edit: Corrected `seek()` and added other functions, mostly as scale-ability example, not for specifically adding each of those functions to the PR.

Edit2: Remembered a way to include a variable number of arguments to unpack, added it to the example.

danifus:
Thanks for the suggestions @aeros167. `seek` does require the `offset` parameter to be supplied (`seek(offset[, whence])`) and I prefer having them written out explicitly in this instance.

aeros167:
> Thanks for the suggestions @aeros167. `seek` does require the `offset` parameter to be supplied (`seek(offset[, whence])`) and I prefer having them written out explicitly in this instance.

@danifus 
Ah yes you're correct, the `whence` argument is the one with a default of `0` that doesn't need to be assigned, not `offset`. My mistake on that one.

With regards to having them explicitly written out, I'd probably agree with you that it's entirely up to preference if it were just using 2 or so consecutive repetitions, or if this were C instead of Python. However, this is using `assertRaises()` 5 times consecutively with the same exact exception, `ValueError`. Scenarios such as those are the perfect place to use iteration for, otherwise you end up breaking DRY and ultimately reducing the readability of the code. Readability doesn't just involve the simplicity of the code, it also involves *effectively* conveying the meaning in fewer lines.

Although this one section by itself isn't a big deal at all, when you consider the sheer volume of cpython, the extra needless lines of code can scale nearly indefinitely. As an example, let's say 10K PRs each added a needless 5-10 lines of code. 50K-100k lines of code is no joke at all, that could easily be a massive project worth of excess code to read through, adding significantly to the maintainability costs.

So, my opinion is that maintenance costs and readability supersede personal preference in this case. The scale-ability is also especially important for the unit tests. For the record, in a personal project for my own tests, I probably would have written it out explicitly like you did (since the few lines of extra code don't matter in that case). 

It's perfectly fine though if we disagree, the final verdict doesn't fall upon me. That's ultimately up to the core devs. Thanks again for your contribution and welcome to cpython!

serhiy-storchaka (approved):

PR #14659, opened 2019-07-09
bpo-18374: fix tests to check the correct thing about line numbers

cfbolz:
@ilevkivskyi here's the promised fix.

Should I also try to do the backport to 3.7? (the automated merging fails because the `end_lineno` and  `end_col_offset` didn't exist in 3.7).

https://bugs.python.org/issue18374


ilevkivskyi (approved):
Thanks!

ilevkivskyi:
@cfbolz 
> Should I also try to do the backport to 3.7?

I don't think it is worth spending some manual work on the backport. Fixing it in 3.8 is already good enough.

PR #14661, opened 2019-07-09
bpo-37473: Don't import importlib ASAP in tests

vstinner:
[bpo-15386](https://bugs.python.org/issue15386), [bpo-37473](https://bugs.python.org/issue37473): test_import, regrtest and libregrtest no longer
import importlib as soon as possible, as the first import, "to test
[bpo-15386](https://bugs.python.org/issue15386)".

Sort test_import imports.

https://bugs.python.org/issue37473

### DIFF
--- a/Lib/test/libregrtest/__init__.py
+++ b/Lib/test/libregrtest/__init__.py
@@ -1,5 +1,2 @@
-# We import importlib *ASAP* in order to test #15386
-import importlib
-
 from test.libregrtest.cmdline import _parse_args, RESOURCE_NAMES, ALL_RESOURCES
 from test.libregrtest.main import main
--- a/Lib/test/regrtest.py
+++ b/Lib/test/regrtest.py
@@ -6,9 +6,6 @@ Script to run Python regression tests.
 Run this script with -h or --help for documentation.
 """
 
-# We import importlib *ASAP* in order to test #15386
-import importlib
-
 import os
 import sys
 from test.libregrtest import main
--- a/Lib/test/test_import/__init__.py
+++ b/Lib/test/test_import/__init__.py
@@ -1,24 +1,22 @@
-# We import importlib *ASAP* in order to test #15386
-import importlib
+import builtins
+import contextlib
+import errno
+import glob
 import importlib.util
 from importlib._bootstrap_external import _get_sourcefile
-import builtins
 import marshal
 import os
 import py_compile
 import random
 import shutil
-import subprocess
 import stat
+import subprocess
 import sys
+import textwrap
 import threading
 import time
 import unittest
-import unittest.mock as mock
-import textwrap
-import errno
-import contextlib
-import glob
+from unittest import mock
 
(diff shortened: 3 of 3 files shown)


vstinner:
@brettcannon, @ncoghlan: Are you ok to remove these hacks?

It seems like https://bugs.python.org/issue15386 is already tested by test_import.test_there_can_be_only_one(). "import imp" was replaced with "import _imp" incommit be7e49fd820318509cd8b4dbde479c552f74ef62:

```
commit be7e49fd820318509cd8b4dbde479c552f74ef62
Author: Nick Coghlan <ncoghlan@gmail.com>
Date:   Fri Jul 20 23:40:09 2012 +1000

    Close #15386: There was a loophole that meant importlib.machinery and imp would sometimes reference an uninitialised copy of importlib._bootstrap
```

At least, I can say that importlib breaks quickly with the following change:

```
diff --git a/Lib/importlib/__init__.py b/Lib/importlib/__init__.py
index 0c73c505f9..130bc19537 100644
--- a/Lib/importlib/__init__.py
+++ b/Lib/importlib/__init__.py
@@ -9,7 +9,7 @@ __all__ = ['__import__', 'import_module', 'invalidate_caches', 'reload']
 # modules would get an uninitialised copy of the source version, instead
 # of a fully initialised version (either the frozen one or the one
 # initialised below if the frozen one is not available).
-import _imp  # Just the builtin component, NOT the full Python module
+import imp  # Just the builtin component, NOT the full Python module
 import sys
 
 try:
```

brettcannon:
As long as there's another test then I'm fine with the change.

aeros (commented):
After reading through the issues and doing some testing on the latest version, I checked to see if the issue is still occurring:

>I haven't worked out how yet, but importlib.machinery is managing to bypass the replacement of importlib._bootstrap with _frozen_importlib:

Since `FileFinder` was moved from `_bootstrap` to `_bootstrap_external` I replaced the reference when I ran through the process in the interpreter:

image

Based upon the outputs, it looks like the issue was solved at some point by having `importlib._bootstrap_external.FileFinder` and `importlib.machinery.FileFinder` refer to the same class, `_frozen_importlib_external.FileFinder`. To ensure this is happening, an assertion such as this could be used: 

```python3
# ...
self.assertEqual(importlib.machinery.FileFinder, 
importlib._bootstrap_external.FileFinder, 
msg = "Both locations should refer to the same class.")
```

It might be a bit bandage solution still since it does not explain why the override was occurring in the first place, but at least it no longer happens with `FileFinder`.

vstinner:
I merged your PR. If someone considers than test_there_can_be_only_one() is not enough, please add a new test.

ncoghlan:
importlib initialisation changed quite a bit when the external import bootstrapping was split out to a separate step, so it's quite plausible that the previous odd behaviour was a result of FileFinder being defined *in* the core import system bootstrap module.

PR #14662, opened 2019-07-09
bpo-37322: Fix test_ssl.test_pha_required_nocert() ResourceWarning

vstinner:
Close the TLS connection in test_pha_required_nocert() of test_ssl to
fix a ResourceWarning.

https://bugs.python.org/issue37322


vstinner:
The test failed on Windows, I cancelled the backport to 3.8.

PR #14667, opened 2019-07-09
bpo-37322: Fix again test_ssl.test_pha_required_nocert() ResourceWarning

vstinner:
test_ssl.test_pha_required_nocert(): only close the TLS connection at
the server side once the handler completed.

https://bugs.python.org/issue37322

### DIFF
--- a/Lib/test/test_ssl.py
+++ b/Lib/test/test_ssl.py
@@ -2382,7 +2382,6 @@ class ThreadedEchoServer(threading.Thread):
                         if self.server.chatty and support.verbose:
                             sys.stdout.write(err.args[1])
                         # test_pha_required_nocert is expecting this exception
-                        self.close()
                         raise ssl.SSLError('tlsv13 alert certificate required')
                 except OSError:
                     if self.server.chatty:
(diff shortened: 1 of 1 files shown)


vstinner:
It seems like the test fails sometimes on Windows. I don't have time right now to investigate, so I will just revert my change instead, PR #14662.

aeros (commented):
Perhaps instead of explicitly using `handler.close()`, you could use a with statement to guarantee that the resource is safely closed:

```python
with self.ConnectionHandler(self, newconn, connaddr) as handler:
    handler.start()
    handler.join()
```
This only requires that the object has an `__enter__` and `__exit__` method, which `ConnectionHandler` does.

Also as a minor readability change, I would recommend changing this (lines 2457 & 2458):
```python
sys.stdout.write(' server:  new connection from ' 
    + repr(connaddr) + '\n')
```
to:
```python
sys.stdout.write(f"server:  new connection from {conaddr!r}\n")
```
This isn't at all necessary, but I figured since it was right near by it would be a good time it clean it up a little. I find the f string formatting a lot easier to quickly read than the old string concatenation.

vstinner:
@aeros167: You are commenting a closed PR :-) I abandoned it since the test was still failing randomly with this change when combined with commit 73ea54620a6f91c3f2e53880373dd47813691a21.

aeros167:
@vstinner Oops, I started typing it while it was still open, my apologies. Do you think this would be worth reopening the issue over to test if it fixes the `ResourceWarning` and issue with windows compatibility? I wouldn't mind opening a draft PR and just to experiment with it if you're otherwise occupied.

PR #14674, opened 2019-07-09
bpo-37543: optimize pymalloc

methane:
PyObject_Malloc() and PyObject_Free() inlines pymalloc_alloc and
pymalloc_free partially.
But when PGO is not used, the compiler doesn't know where is the hot part
in pymalloc_alloc and pymalloc_free.

To help compilers:

* use LIKELY/UNLIKELY macros
* mark pymalloc_alloc and pymalloc_free as inline function
* split slow / rare parts into other functions

```
# bm_spectral_norm
Mean +- std dev: [master] 199 ms +- 1 ms -> [patched] 176 ms +- 1 ms: 1.13x faster (-11%)
```

https://bugs.python.org/issue37543


Objects/obmalloc.c
```diff
@@ -1407,96 +1409,43 @@ address_in_range(void *p, poolp pool)
...
-        return 0;
-    }
-#endif
-
-    if (nbytes == 0) {
-        return 0;
-    }
-    if (nbytes > SMALL_REQUEST_THRESHOLD) {
-        return 0;
+    if (LIKELY(pool->nextoffset <= pool->maxnextoffset)) {
```
tim-one: Unfortunately, this is likely for lightly used size classes, but unlikely for heavily used size classes (pools for those get entirely carved into blocks, and the bump allocator never gets used again).  So, overall, it's probably better to say this one is UNLIKELY.

Objects/obmalloc.c
```diff
@@ -1407,96 +1409,43 @@ address_in_range(void *p, poolp pool)
...
+    /* Pool is full, unlink from used pools. */
+    poolp next;
+    next = pool->nextpool;
+    pool = pool->prevpool;
+    next->prevpool = pool;
+    pool->nextpool = next;
+}
 
+static void*
+allocate_from_new_pool(size_t size)
```
tim-one: 1. A brief comment before to explain what the arguments and return value are would be very helpful.

2. The `size` argument here isn't a byte count, but a pymalloc size class.  `uint` is  more appropriate than `size_t`.

tim-one (approved):
Very nice!  It's lovely to see the main functions cut down to something understandable :smile:  Just have some minor comments.

jdemeyer:
Did you putting the macros `LIKELY` and `UNLIKELY` in a public header file (renaming them of course)? There are surely other places in CPython that could benefit from them.

methane:
I did not it, but I will do it when I want to use it in other places.

PR #14675, opened 2019-07-09
bpo-37530: simplify, optimize and clean up IDLE code context

taleinat:
* Only create `CodeContext` instances for "real" editors windows, but not e.g. shell or output windows.
* Remove configuration update Tk event fired every second, by having the editor window ask its code context widget to update when necessary, i.e. upon font or highlighting updates.
* When code context isn't being shown, avoid having a Tk event fired every 100ms to check whether the code context needs to be updated.
* Use the editor window's `getlineno()` method where applicable.
* Fix a bare `except:`.

https://bugs.python.org/issue37530

### DIFF
--- a/Lib/idlelib/editor.py
+++ b/Lib/idlelib/editor.py
@@ -799,6 +799,8 @@ class EditorWindow(object):
         # Called from configdialog.py
 
         self.text['font'] = idleConf.GetFont(self.root, 'main','EditorWindow')
+        if self.codecontext is not None:
+            self.codecontext.update_font()
 
     def RemoveKeybindings(self):
         "Remove the keybindings before they are changed."
--- a/Lib/idlelib/editor.py
+++ b/Lib/idlelib/editor.py
@@ -62,6 +62,8 @@ class EditorWindow(object):
     filesystemencoding = sys.getfilesystemencoding()  # for file names
     help_url = None
 
+    allow_codecontext = True
+
     def __init__(self, flist=None, filename=None, key=None, root=None):
         # Delay import: runscript imports pyshell imports EditorWindow.
         from idlelib.runscript import ScriptBinding
@@ -247,6 +249,7 @@ class EditorWindow(object):
         self.good_load = False
         self.set_indentation_params(False)
         self.color = None # initialized below in self.ResetColorizer
+        self.codecontext = None
         if filename:
             if os.path.exists(filename) and not os.path.isdir(filename):
                 if io.loadfile(filename):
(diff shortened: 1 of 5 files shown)


taleinat:
I've tested this on Windows, macOS and Ubuntu Linux. Seems working well.

@terryjreedy, would you like a NEWS entry for this?

terryjreedy:
1. Does that mean you think this is ready to commit?  
2. Yes.  "Optimize code context to reduce unneeded background activity" seems to be the essence of the improvement (including not creating instances for non-editors) in words that users should be able to appreciate.  

An update loop is only needed for an active editor with code context, rather all editors with such.  Whatever this PR does not do (have not read details yet) can be another PR.

taleinat:
1. Yes.
2. Great wording! Done.

terryjreedy:
Then I will look at this this evening.

Lib/idlelib/editor.py
```diff
@@ -791,6 +799,8 @@ def ResetFont(self):
         # Called from configdialog.py
 
         self.text['font'] = idleConf.GetFont(self.root, 'main','EditorWindow')
+        if self.codecontext is not None:
```
terryjreedy, line 804: It might be better (later) to change context font first, since effect of font change on text depends on effect of font change on context, but not vice versa.  See review comment.  Possible code:
```
        newfont = idleConf.GetFont(self.root, 'main','EditorWindow')
        if self.codecontext is not None:
           self.codecontext.update_font(newfont)  # change sig
        self.text['font'] = newfont
        if self.linenumber is not None: ...
```

terryjreedy (approved):
I added a sentence to blurb.  Edit if you want.  Please merge and update line number patch.

After adding line numbers, we might want to experiment with the order in which changes are applied.  See example.

Lib/idlelib/editor.py
terryjreedy: I was OK with leaving this for a later PR, but it may actually make changes even snappier.  In reading ResetFont, I noticed that it is ResetColorizer, above this, also called from configdialog and somewhere else, that deletes and restores colors.  We should try reconfiguring tags instead on another issue.  Names should sometime be lower-cased.

terryjreedy:
I presume you are working on fixing test_font.  It might be doing more than is needed, such as restoring the font of the temporary text instance.

Travis also had an unrelated doctest failure.  I will open a new issue for that if not one already.

taleinat:
@terryjreedy 

> Please merge and update line number patch.

Done.

PR #14679, opened 2019-07-09
bpo-37531: Fix regrtest timeout for subprocesses

nanjekyejoannah:
Fix regrtest timeout for subprocesses.

Naming things is the hardest thing in computer science. Interested in any reviews in a better name If any for the new function I introduced. cc @vstinner 


ammaraskar:
Hey @nanjekyejoannah, did you accidentally attach the wrong bpo ticket for this? 3753 seems really old and unrelated.

ZackerySpytz:
It seems that you have changed the file mode for several files under `Lib/test`. Why? I don't think those changes should be made.

ZackerySpytz:
It seems that you have changed the file mode of several files under `Lib/test`. Why? I don't think those changes should be made.

Lib/test/libregrtest/runtest.py
```diff
@@ -35,6 +36,7 @@
     INTERRUPTED: '%s interrupted',
     CHILD_ERROR: '%s crashed',
     TEST_DID_NOT_RUN: '%s run no tests',
+    TIMEOUT: '%s test did not finish',
```
vstinner: I suggest:

```suggestion
    TIMEOUT: '%s timed out',
```

Lib/test/libregrtest/runtest.py
```diff
@@ -179,6 +181,7 @@ def runtest(ns, test_name):
         FAILED           test failed
         PASSED           test passed
         EMPTY_TEST_SUITE test ran no subtests.
+        TIMEOUT         test failed due to timeout
```
vstinner:
```suggestion
         TIMEOUT         test timed out 
```

Misc/NEWS.d/next/Library/2019-07-09-19-38-26.bpo-3753.GX7s8S.rst
```diff
@@ -0,0 +1 @@
+Fix `regrtest` timeout for subprocesses.
```
vstinner: You should better explain the change. The change only impacts "python3 -m test -jN --timeout=TIMEOUT", it ensures that a worker process is stopped with a timeout if it runs longer than TIMEOUT seconds.

vstinner (commented):
It seems like you modified Lib/test/test_os.py permissions somehow, please revert this change.

nanjekyejoannah:

nanjekyejoannah:
> 
> 
> Hey @nanjekyejoannah, did you accidentally attach the wrong bpo ticket for this? 3753 seems really old and unrelated.

Yes, thanks for catching this. I made the necessary changes.

nanjekyejoannah:
> 
> 
> It seems that you have changed the file mode of several files under `Lib/test`. Why? I don't think those changes should be made.

oh yes, I have reverted this.

vstinner:
Overall, the change is great and seems to work as expected, but I tested it and it doesn't work as I expected:

* it doesn't report stdout/stderr on timeout
* regression: it ignores stdout/stderr on "CHILD_ERROR" whereas it was properly passed to the main process previously
* faulthandler  and Popen.communicate() use the same timeout, so it's unclear which one which trigger first: I tuned timeouts to give more time to faulthandler. I prefer faulthandler since it dumps the traceback where Python was stuck, whereas Popen.communicate() timeout only returns "truncated" output with no traceback
* the timeout is reported as an exception, not as as "result"

I tried to add a commit to this PR but I wasn't allowed, so I forked this PR and created PR #15072 to add my changes on top of it.

nanjekyejoannah:
Closing this then in favor of PR #15072

PR #14680, opened 2019-07-10
bpo-37537: Compute allocated blocks in _Py_GetAllocatedBlocks()

nascheme:
Keeping an account of allocated blocks slows down _PyObject_Malloc() and _PyObject_Free() by a measureable amount.  Have _Py_GetAllocatedBlocks() iterate over the arenas to sum up the allocated blocks for pymalloc.

https://bugs.python.org/issue37537

### DIFF
--- a/Objects/obmalloc.c
+++ b/Objects/obmalloc.c
@@ -1206,12 +1206,29 @@ static size_t ntimes_arena_allocated = 0;
 /* High water mark (max value ever seen) for narenas_currently_allocated. */
 static size_t narenas_highwater = 0;
 
-static Py_ssize_t _Py_AllocatedBlocks = 0;
+static Py_ssize_t raw_allocated_blocks;
 
 Py_ssize_t
 _Py_GetAllocatedBlocks(void)
 {
-    return _Py_AllocatedBlocks;
+    Py_ssize_t n = raw_allocated_blocks;
+    /* add up allocated blocks for used pools */
+    for (uint i = 0; i < maxarenas; ++i) {
+        uintptr_t base = arenas[i].address;
+
+        /* Skip arenas which are not allocated. */
+        if (arenas[i].address == (uintptr_t)NULL)
+            continue;
+
+        /* visit every pool in the arena */
+        assert(base <= (uintptr_t) arenas[i].pool_address);
+        for (uint j = 0; base < (uintptr_t) arenas[i].pool_address;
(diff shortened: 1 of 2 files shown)


methane (approved):

Objects/obmalloc.c
```diff
@@ -1206,12 +1206,29 @@ static size_t ntimes_arena_allocated = 0;
...
+static Py_ssize_t raw_allocated_blocks;
 
 Py_ssize_t
 _Py_GetAllocatedBlocks(void)
 {
-    return _Py_AllocatedBlocks;
+    Py_ssize_t n = raw_allocated_blocks;
+    /* add up allocated blocks for used pools */
+    for (uint i = 0; i < maxarenas; ++i) {
+        uintptr_t base = arenas[i].address;
```
tim-one: In theory, `base` may not be pool-aligned.  So it needs another line here, like,

    base = (uintptr_t)_Py_ALIGN_UP(base, SYSTEM_PAGE_SIZE);

Objects/obmalloc.c
```diff
@@ -1206,12 +1206,29 @@ static size_t ntimes_arena_allocated = 0;
...
+
+        /* Skip arenas which are not allocated. */
+        if (arenas[i].address == (uintptr_t)NULL)
+            continue;
+
+        /* visit every pool in the arena */
+        assert(base <= (uintptr_t) arenas[i].pool_address);
+        for (uint j = 0; base < (uintptr_t) arenas[i].pool_address;
+             ++j, base += POOL_SIZE) {
```
tim-one: `j` isn't used in the loop, so no need for `j = 0` or `++j`.   I'm guessing you copied most of this loop from `_PyObject_DebugMallocStats()`, where `j` isn't actually used either :wink:

tim-one (changes requested):
Surprising!  Good catch.  I doubt this function is ever used outside of memory-leak detectors, so its speed is hard to care about.  I'm just asking for a change to round the arena base address up to the first pool's actual address (which probably isn't really needed on Linux or Windows, but will be needed if pool sizes ever increase beyond the OS page size).

Objects/obmalloc.c
tim-one: Oops - that should round up to `POOL_SIZE`, not page size.

nascheme:

tim-one (approved):

PR #14682, opened 2019-07-10
bpo-37540: vectorcall: keyword names must be strings

jdemeyer:
The fact that keyword names are strings is now part of the vectorcall and `METH_FASTCALL` protocols. The biggest concrete change is that `_PyStack_UnpackDict` now checks that and raises `TypeError` if not.

CC @markshannon @vstinner 

https://bugs.python.org/issue37540

### DIFF
--- a/Lib/test/test_unpack_ex.py
+++ b/Lib/test/test_unpack_ex.py
@@ -256,7 +256,7 @@ Overridden parameters
     >>> f(**{1: 3}, **{1: 5})
     Traceback (most recent call last):
       ...
-    TypeError: f() keywords must be strings
+    TypeError: f() got multiple values for keyword argument '1'
 
 Unpacking non-sequence
 
--- a/Doc/c-api/object.rst
+++ b/Doc/c-api/object.rst
@@ -388,8 +388,8 @@ Object Protocol
    :c:func:`PyVectorcall_NARGS(nargsf) <PyVectorcall_NARGS>`.
 
    *kwnames* can be either NULL (no keyword arguments) or a tuple of keyword
-   names. In the latter case, the values of the keyword arguments are stored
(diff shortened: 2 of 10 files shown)


encukou:
Instead of "string", I'd say "\`str` or a subclass" explicitly.

jdemeyer:
There are many instances in the documentation (unrelated to vectorcall) where the word "string" is used. I don't see the problem here.

Concrete proposal: we leave this doc as is and I will reboot #13844 (I always planned to do that, but *after* all other PRs that affect documentation). In the section describing the "vectorcall protocol", I will say explicitly what "string" means for vectorcall.

jdemeyer:
Is this PR okay now or do you want me to change things?

encukou:
@markshannon, you said you were looking at this PR. Is that still the case? Any progress?

jdemeyer:
Ping?

Lib/test/test_unpack_ex.py
```diff
@@ -256,7 +256,7 @@
     >>> f(**{1: 3}, **{1: 5})
     Traceback (most recent call last):
       ...
-    TypeError: f() keywords must be strings
```
encukou, line 259: I see `f(**{1:2})` is still tested in test_excall.py.

encukou (approved):
@markshannon seems unresponsive. I don't see an issue with this PR.

Lib/test/test_unpack_ex.py
jdemeyer: Indeed. This is testing two errors at the same time (duplicate keyword and non-string keyword). Which of the two errors you get seems arbitrary to me and this PR changes the error.

jdemeyer:
Thanks! This will help to make progress on some other PRs.

PR #14683, opened 2019-07-10
bpo-29548: no longer use PyEval_Call* functions

jdemeyer:
The `PyEval_Call*` functions should not be used anymore. This fixes the last few places where they are called within CPython.

CC @methane 

https://bugs.python.org/issue29548


zooba (approved):

zooba:
LMK if you'd like me to merge, or if you want additional review from someone else.

PR #14684, opened 2019-07-10
bpo-37151: remove special case for PyCFunction from PyObject_Call

jdemeyer:
The function `PyObject_Call` contains a special case for calling instances of `PyCFunction` with `METH_VARARGS`. This is to avoid a double recursion check: `cfunction_call_varargs` already calls `Py_EnterRecursiveCall`, so `PyObject_Call` should not. We can simplify this by removing the recursion check from `cfunction_call_varargs` and removing the special case from `PyObject_Call`.

In addition, we fold `cfunction_call_varargs` into `PyCFunction_Call` and move it back to `Objects/methodobject.c` which already contains the vectorcall functions of `PyCFunction`.

### DIFF
--- a/Include/methodobject.h
+++ b/Include/methodobject.h
@@ -39,7 +39,7 @@ PyAPI_FUNC(int) PyCFunction_GetFlags(PyObject *);
 #define PyCFunction_GET_FLAGS(func) \
         (((PyCFunctionObject *)func) -> m_ml -> ml_flags)
 #endif
-PyAPI_FUNC(PyObject *) PyCFunction_Call(PyObject *, PyObject *, PyObject *);
+Py_DEPRECATED(3.9) PyAPI_FUNC(PyObject *) PyCFunction_Call(PyObject *, PyObject *, PyObject *);
 
 struct PyMethodDef {
     const char  *ml_name;   /* The name of the built-in function/method */
--- a/Objects/call.c
+++ b/Objects/call.c
@@ -5,9 +5,6 @@
 #include "frameobject.h"
 
 
-static PyObject *
-cfunction_call_varargs(PyObject *func, PyObject *args, PyObject *kwargs);
-
 static PyObject *const *
 _PyStack_UnpackDict(PyObject *const *args, Py_ssize_t nargs, PyObject *kwargs,
                     PyObject **p_kwnames);
@@ -236,11 +233,6 @@ PyObject_Call(PyObject *callable, PyObject *args, PyObject *kwargs)
     if (_PyVectorcall_Function(callable) != NULL) {
         return PyVectorcall_Call(callable, args, kwargs);
     }
-    else if (PyCFunction_Check(callable)) {
-        /* This must be a METH_VARARGS function, otherwise we would be
-         * in the previous case */
-        return cfunction_call_varargs(callable, args, kwargs);
-    }
     else {
         call = callable->ob_type->tp_call;
         if (call == NULL) {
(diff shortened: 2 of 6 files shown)


Objects/call.c
```diff
@@ -364,60 +356,6 @@ _PyFunction_Vectorcall(PyObject *func, PyObject* const* stack,
...
-    assert(!PyErr_Occurred());
-    assert(kwargs == NULL || PyDict_Check(kwargs));
-
-    PyCFunction meth = PyCFunction_GET_FUNCTION(func);
-    PyObject *self = PyCFunction_GET_SELF(func);
-    PyObject *result;
-
-    assert(PyCFunction_GET_FLAGS(func) & METH_VARARGS);
-    if (PyCFunction_GET_FLAGS(func) & METH_KEYWORDS) {
-        if (Py_EnterRecursiveCall(" while calling a Python object")) {
```
scoder, line 381: The recursion check is gone now. Was that intentional? Any reason why you think we don't need it any more?
Since you're dealing with a public C-API function here, I would advise to keep the current semantics.

jdemeyer: The recursion check is gone from `PyCFunction_Call` but it is in `PyObject_Call`.

Note that `PyCFunction_Call` is not a documented C API function, so I assumed that this minor change in behaviour (no longer doing the recursion check) would be acceptable.

vstinner:
> Note that PyCFunction_Call is not a documented C API function, so I assumed that this minor change in behaviour (no longer doing the recursion check) would be acceptable.

That's the kind of function that projects like Cython call directly.

jdemeyer:
> That's the kind of function that projects like Cython call directly.

What point are you trying to make? It that just a comment or is it saying something about this PR? Or are you just complaining about Cython? Or is it a joke somehow?

vstinner:
I suggest to ensure that any PyCFunction_Call behavior change will not break anything, especially in Cython.

encukou:
It is undocumented, so the only documentation available to people who wanted to use this function was the code itself.
It would be OK to remove/rename `PyCFunction_Call`, if we needed to do that, but I don't think we should subtly change its behavior.

Perhaps CPython could use `cfunction_call_varargs` internally, and that would have the check removed; `PyCFunction_Call` could then be a wrapper that does the check?

jdemeyer:
I checked and Cython doesn't use `PyCFunction_Call`. I'm not sure if that function was ever intended to be public. It's quite special, there are no such functions for other classes. It's not documented. And what's the use case? You can only use `PyCFunction_Call` if you somehow know in advance that some object is a `builtin_function_or_method`, which you typically don't know. And if you somehow *do* know that it's a `builtin_function_or_method` and care about speed, you will want to use the vectorcall calling convention instead of the slow `tp_call` convention used by `PyCFunction_Call`.

jdemeyer:
> `PyCFunction_Call` could then be a wrapper that does the check?

At that point, probably best would be to just alias it to `PyObject_Call`.

encukou:
Right, it's quite useless as public API. Remove it in 3.9. But:
* Don't keep the same name if the semantics changed.
* Mention the removal in the news file, so anyone who did use it may know the removal was intentional.

It was probably never intended to be public (and probably predates discussions on what should be public), but it *is* an exported symbol.

jdemeyer:
We need to keep it for the stable ABI, but that's easily solved by making it an alias of `PyObject_Call`.

I updated my branch to do this.

Include/methodobject.h
```diff
@@ -39,7 +39,7 @@ PyAPI_FUNC(int) PyCFunction_GetFlags(PyObject *);
 #define PyCFunction_GET_FLAGS(func) \
         (((PyCFunctionObject *)func) -> m_ml -> ml_flags)
 #endif
-PyAPI_FUNC(PyObject *) PyCFunction_Call(PyObject *, PyObject *, PyObject *);
+Py_DEPRECATED(3.9) PyAPI_FUNC(PyObject *) PyCFunction_Call(PyObject *, PyObject *, PyObject *);
```
encukou, line 42: @vstinner, is this the right way to deprecate an undocumented part of the stable ABI?

(I'll merge regardless, but I wanted to bring this to your attention.)

encukou (approved):
Thanks. I'm now comfortable merging this, especially since it's for 3.9.

Include/methodobject.h
vstinner: The stable ABI is supposed to remain unchanged :-) "Py_DEPRECATED(3.9) PyAPI_FUNC(PyObject *) PyCFunction_Call" is the right usage of Py_DEPRECATED().

I don't know why PyCFunction_Call "leaked" into the stable ABI. I'm in favor of removing it. Start with a deprecation is a good start ;)

PR #14685, opened 2019-07-10
bpo-37547: add _PyObject_CallMethodOneArg

jdemeyer:
https://bugs.python.org/issue37547

### DIFF
--- a/Doc/c-api/object.rst
+++ b/Doc/c-api/object.rst
@@ -375,6 +375,18 @@ Object Protocol
    .. versionadded:: 3.9
 
 
+.. c:function:: PyObject* _PyObject_CallMethodOneArg(PyObject *obj, PyObject *name, PyObject *arg)
+
+   Call a method of the Python object *obj* with a single positional argument
+   *arg*, where the name of the method is given as a Python string object in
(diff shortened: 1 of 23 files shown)


aeros (commented):
In a similar theme to the proposed `CallMethodOneArg()` and existing `CallNoArg()`, `CallOneArg()`, and `CallMethodNoArgs()`, would it make sense to also have a  `VectorcallNoKws()`? There's several instances where `Vectorcall()` is being used with a null value for the kwnames argument. 

Edit: This would probably be in it's own separate PR and issue, not this one

jdemeyer:
> There's several instances where `Vectorcall()` is being used with a null value for the kwnames argument.

There are only very few such use cases. It we would add another function in similar spirit, I think that `_PyObject_CallTwoArgs()` would make the most sense and that was rejected.

methane:
> would it make sense to also have a `VectorcallNoKws()`?

No.  Passing `NULL` is not so hard, compared to `(PyObject*[]){self, arg}`.

aeros167:
@jdemeyer @methane Thanks for the clarification.

PR #11273, opened 2018-12-21
bpo-21478: Autospec functions should propagate mock calls to parent

tirkarthi:
When a function is passed to `create_autospec` then a function object is returned where the mock is attached to the object with the attribute name `mock` . Set parent against the mock attribute so that calls to children are propagated to the parent.

https://bugs.python.org/issue21478

PR #14688, opened 2019-07-10
bpo-21478: Record calls to parent when autospecced objects are used as child with attach_mock

tirkarthi:
When autospecced object is passed to `attach_mock` the mock object is stored in the `mock` attribute that has `_mock_name` and `_mock_parent` set. For function types passed check for the mock attribute and clear name and parent of it so that `attach_mock` updates the child with correct name and parent in `_check_and_set_parent`.

https://bugs.python.org/issue21478

### DIFF
--- a/Lib/unittest/mock.py
+++ b/Lib/unittest/mock.py
@@ -467,10 +467,19 @@ class NonCallableMock(Base):
         Attach a mock as an attribute of this one, replacing its name and
         parent. Calls to the attached mock will be recorded in the
         `method_calls` and `mock_calls` attributes of this one."""
-        mock._mock_parent = None
-        mock._mock_new_parent = None
-        mock._mock_name = ''
-        mock._mock_new_name = None
+
+        # When autospecced object is passed to attach_mock then clear
+        # name and parent of the mock attribute which holds the actual
+        # mock object.
+        if isinstance(mock, FunctionTypes) and hasattr(mock, 'mock'):
+            inner_mock = mock.mock
+        else:
+            inner_mock = mock
+
+        inner_mock._mock_parent = None
+        inner_mock._mock_new_parent = None
+        inner_mock._mock_name = ''
+        inner_mock._mock_new_name = None
 
         setattr(self, attribute, mock)
 
--- a/Lib/unittest/mock.py
+++ b/Lib/unittest/mock.py
@@ -72,6 +72,15 @@ def _is_exception(obj):
     )
 
 
+def _extract_mock(obj):
+    # Autospecced functions will return a FunctionType with "mock" attribute
+    # which is the actual mock object that needs to be used.
+    if isinstance(obj, FunctionTypes) and hasattr(obj, 'mock'):
+        return obj.mock
+    else:
+        return obj
+
+
 def _get_signature_object(func, as_instance, eat_self):
     """
     Given an arbitrary, possibly callable object, try to create a suitable
@@ -346,13 +355,7 @@ class _CallList(list):
 
 
 def _check_and_set_parent(parent, value, name, new_name):
-    # function passed to create_autospec will have mock
-    # attribute attached to which parent must be set
-    if isinstance(value, FunctionTypes):
-        try:
-            value = value.mock
-        except AttributeError:
-            pass
+    value = _extract_mock(value)
 
     if not _is_instance_mock(value):
         return False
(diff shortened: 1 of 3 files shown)


iforapsy (approved):

Lib/unittest/test/testmock/testmock.py
```diff
@@ -1808,6 +1811,21 @@ def test_attach_mock_return_value(self):
...
 
+    def test_attach_mock_patch_autospec(self):
+        # bpo-21478: attach_mock used with autospecced object should have
+        # should use child's name and record calls to parent
+        parent = Mock()
+
+        with mock.patch(f'{__name__}.something', autospec=True) as mock_func:
+            self.assertEqual(mock_func.mock._extract_mock_name(), 'something')
+            parent.attach_mock(mock_func, 'child')
+            parent.child()
```
iforapsy:
```suggestion
            something()
```
In my use case for attaching the child mock to the parent, I call the child mock directly instead of through the parent mock. The only reason I attach the child to a parent is so that I can check the order of calls to different functions, à la https://stackoverflow.com/a/27306920/2407644.

tirkarthi: Thanks @iforapsy for the review. I have changed the test to ensure call as a child, standalone function call and mock object call are all recorded to both parent and the respective mock object so that all cases are covered.

tirkarthi:
@cjw296 @mariocj89 Would be helpful to have a review of this. This is similar to https://github.com/python/cpython/pull/11273 where `create_autospec` has returns the original object with mock stored in the `mock` attribute that needs to be used.

Lib/unittest/test/testmock/testmock.py
```diff
@@ -1808,6 +1811,29 @@ def test_attach_mock_return_value(self):
                 self.assertEqual(m.mock_calls, call().foo().call_list())
 
 
+    def test_attach_mock_patch_autospec(self):
+        # bpo-21478: attach_mock used with autospecced object should have
+        # should use child's name and record calls to parent when being called
+        # as child and also as standalone functions.
```
cjw296: Not sure we need this comment, but the text could do with some tweaking.

Lib/unittest/test/testmock/testmock.py
```diff
@@ -1891,6 +1917,16 @@ def foo(a, b): pass
 
         self.assertRaises(TypeError, mock.child, 1)
         self.assertEqual(mock.mock_calls, [call.child(1, 2)])
+        self.assertIn('mock.child', repr(mock.child.mock))
+
+        parent = Mock()
```
cjw296, line 1923: This feels like it should be a separate test.

Lib/unittest/test/testmock/testmock.py
```diff
@@ -1891,6 +1917,16 @@ def foo(a, b): pass
...
+        parent.attach_mock(create_autospec(foo, name='bar'), 'child')
+        parent.child(1, 2)
```
cjw296, line 1925: Why are we calling `child` direct here?

Lib/unittest/mock.py
```diff
@@ -467,10 +467,19 @@ def attach_mock(self, mock, attribute):
...
         `method_calls` and `mock_calls` attributes of this one."""
-        mock._mock_parent = None
-        mock._mock_new_parent = None
-        mock._mock_name = ''
-        mock._mock_new_name = None
+
+        # When autospecced object is passed to attach_mock then clear
+        # name and parent of the mock attribute which holds the actual
+        # mock object.
+        if isinstance(mock, FunctionTypes) and hasattr(mock, 'mock'):
```
cjw296: If this is the way that we know an autospecced object is passed in, it feels like it should be abstracted out into a helper function with a sensible name, and used wherever else this pattern has cropped up.

Lib/unittest/test/testmock/testmock.py
tirkarthi: I removed it since I feel bug report gives the relevant context. Feel free to suggest wording if I need to incorporate a code comment. Since I am not a native speaker sometimes my language might be little difficult to grasp.

Lib/unittest/test/testmock/testmock.py
tirkarthi: Moved it to a separate test. Thanks.

Lib/unittest/mock.py
tirkarthi: I moved it to `_extract_mock` to avoid code duplication and added a code comment related to autospecced functions.

Lib/unittest/test/testmock/testmock.py
tirkarthi: Sorry, I don't get the comment here. The foo function is attached as a `child` attribute so I am calling `parent.child` to ensure parent records the call. I am not getting the direct call part here.

mariocj89 (approved):
Nice fix, looks quite nicer with the function extracted indeed.

Lib/unittest/mock.py
```diff
@@ -72,6 +72,15 @@ def _is_exception(obj):
...
 
 
+def _extract_mock(obj):
+    # Autospecced functions will return a FunctionType with "mock" attribute
+    # which is the actual mock object that needs to be used.
+    if isinstance(obj, FunctionTypes) and hasattr(obj, 'mock'):
+        return obj.mock
+    else:
+        return obj
+
```
cjw296, line 82: Nice!

Lib/unittest/test/testmock/testmock.py
cjw296: I guess I don't really understand the use case for `attach_mock` so happy to go with what you say.

PR #14696, opened 2019-07-11
bpo-37532: email.header.make_header with `ascii` out of range(128)

aldwinaldwin:
https://bugs.python.org/issue37532

### DIFF
--- a/Lib/email/header.py
+++ b/Lib/email/header.py
@@ -136,7 +136,14 @@ def decode_header(header):
     last_word = last_charset = None
     for word, charset in decoded_words:
         if isinstance(word, str):
-            word = bytes(word, 'raw-unicode-escape')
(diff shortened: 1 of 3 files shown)


mangrisano (approved):
LGTM. Thank you so much for providing the test :)

brettcannon:
@maxking

aldwinaldwin:
Rejected in bpo: 'The input header is not valid (non-ascii is not allowed in headers), so you shouldn't expect make_header to do anything sensible.  Note that this is the legacy API, which is a toolkit and does not hold your hand when it comes to RFC compliance.  Aside from any other concerns, this is long standing behavior (it is the same in python2), and it doesn't make sense to change the behavior of a legacy API.'

PR #14697, opened 2019-07-11
bpo-37545: 'elif args.verbosity == 1:' in argparse tutorial

aldwinaldwin:
https://bugs.python.org/issue37545

### DIFF
--- a/Doc/howto/argparse.rst
+++ b/Doc/howto/argparse.rst
@@ -530,7 +530,7 @@ Let's fix::
    # bugfix: replace == with >=
    if args.verbosity >= 2:
        print("the square of {} equals {}".format(args.square, answer))
-   elif args.verbosity >= 1:
+   elif args.verbosity == 1:
        print("{}^2 == {}".format(args.square, answer))
    else:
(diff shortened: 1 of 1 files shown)


Doc/howto/argparse.rst
```diff
@@ -646,7 +646,7 @@ to display *more* text instead::
    answer = args.x**args.y
    if args.verbosity >= 2:
        print("Running '{}'".format(__file__))
-   if args.verbosity >= 1:
+   if args.verbosity == 1:
```
zware: This example must not be changed; it is explicitly about showing how to use "verbosity level to display *more* text"

zware (changes requested):
As noted in the issue I don't think this should be done at all, but the last change especially must not be made as it goes against the point of the example.

Doc/howto/argparse.rst
aldwinaldwin: @zware : ok, now I see the difference in output (filename) compared to previous examples. Agreed.

aldwinaldwin:

PR #14698, opened 2019-07-11
Docs: Add reST directives to c-api/intro#Include_Files

aeros:
Adds the following reST directives to the "Include Files" section: code c, important, and note. Also improves the phrasing in the section where a note was added, on lines 99-101.

Since this change is quite minor and doesn't affect the actual message conveyed, I decided to go without creating an issue for it. It's mostly formatting related and an improvement to the phrasing (splits a run-on sentence into two individual ones). This could also go without a news entry, so the `skip news` and `skip issue` labels would probably be appropriate.


mangrisano (approved):

Doc/c-api/intro.rst
```diff
@@ -46,7 +46,9 @@ Include Files
 =============
 
 All function, type and macro definitions needed to use the Python/C API are
-included in your code by the following line::
+included in your code by the following line:
+
+.. code:: c
```
benjaminp: Is this block mishighlighted without this directive?

Doc/c-api/intro.rst
```diff
@@ -90,9 +94,11 @@ multi-platform builds since the platform independent headers under
 :envvar:`prefix` include the platform specific headers from
 :envvar:`exec_prefix`.
 
-C++ users should note that though the API is defined entirely using C, the
-header files do properly declare the entry points to be ``extern "C"``, so there
-is no need to do anything special to use the API from C++.
+.. note::
```
benjaminp: I don't think this needs its own block.

Doc/c-api/intro.rst
aeros: When I had first opened the PR, I thought this would be needed, but it looks like it's fine in for this page since the header defines ``.. highlight:: c`` so any c syntax is automatically highlighted. I wasn't aware of that feature of reST at the time. I'll remove this change.

Doc/c-api/intro.rst
aeros: I was basing this change on the [devguide](https://devguide.python.org/documenting/#paragraph-level-markup)'s summary for using the ``note`` directive:

> An especially important bit of information about an API that a user should be aware of when using whatever bit of API the note pertains to. The content of the directive should be written in complete sentences and include all appropriate punctuation.

This section seemed to fit that description, but I can remove it if you disagree.

aeros: A better candidate for the ``note`` directive though might be the area where I was previously was using ``important``.

aeros167:
Thanks for the review @benjaminp. 

I've removed the ``code`` directive, removed the ``note`` directive you recommended removing, and replaced the previous ``important`` directive I was using with ``note``, which seems to be more appropriate based on the devguide's recommended usage of it. Let me know what you think of the changes. 

When I had initially opened this PR, I was significantly less familiar with the usage of reST within the docs. I've gained some experience since then, and likely would not have opened a PR applying those initial changes.

Doc/c-api/intro.rst
benjaminp: Obviously, this is subjective. But I don't think information about C++ compatibility is any more important than anything else in this section.

Generally, we want to avoid turning the docs in a a series of multi-colored boxes.

Doc/c-api/intro.rst
```diff
@@ -69,10 +69,12 @@ standard headers) have one of the prefixes ``Py`` or ``_Py``.  Names beginning
 with ``_Py`` are for internal use by the Python implementation and should not be
 used by extension writers. Structure member names do not have a reserved prefix.
 
-**Important:** user code should never define names that begin with ``Py`` or
-``_Py``.  This confuses the reader, and jeopardizes the portability of the user
-code to future Python versions, which may define additional names beginning with
-one of these prefixes.
+.. note::
+
+   User code should never define names that begin with ``Py`` or
```
benjaminp: Since you removed the beginning of this line, the whole paragraph probably needs to be reflowed.

aeros: @benjaminp Done. I made sure the maximum number of words fit per line within the paragraph changed, within the 80 char width limit for reST.

aeros167:
Thanks @benjaminp!

PR #14699, opened 2019-07-11
bpo-32381: .pyc files with non-ASCII paths cannot be reopened on Windows

ZackerySpytz:
https://bugs.python.org/issue32381

### DIFF
--- a/Lib/test/test_cmd_line_script.py
+++ b/Lib/test/test_cmd_line_script.py
@@ -438,6 +438,15 @@ class CmdLineTest(unittest.TestCase):
             self.assertEqual(b"", out)
             self.assertEqual(b"", err)
 
+    def test_issue32381(self):
+        # On Windows, a .pyc file with a non-ASCII path could not be reopened.
+        with support.temp_dir() as script_dir:
+            script_name = _make_test_script(script_dir, 'ߢߢscript')
+            py_compile.compile(script_name, doraise=True)
+            pyc_file = support.make_legacy_pyc(script_name)
+            self._check_script(pyc_file, pyc_file, pyc_file, script_dir, None,
+                               importlib.machinery.SourcelessFileLoader)
+
     @contextlib.contextmanager
     def setup_test_pkg(self, *args):
         with support.temp_dir() as script_dir, \
--- a/Python/fileutils.c
+++ b/Python/fileutils.c
@@ -1365,7 +1365,7 @@ _Py_wfopen(const wchar_t *path, const wchar_t *mode)
     return f;
 }
 
-/* Wrapper to fopen().
+/* Open a file.
 
    The file descriptor is created non-inheritable.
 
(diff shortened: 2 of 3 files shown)


Python/fileutils.c
```diff
@@ -1377,14 +1377,13 @@ _Py_fopen(const char *pathname, const char *mode)
         return NULL;
     }
 
-    FILE *f = fopen(pathname, mode);
-    if (f == NULL)
-        return NULL;
-    if (make_non_inheritable(fileno(f)) < 0) {
-        fclose(f);
+    PyObject *path = PyUnicode_DecodeFSDefault(pathname);
```
eryksun, line 1380: The change should be conditionally defined just for Windows. In the Unix world a path is just a sequence of bytes, with only ASCII slash and null reserved.

Lib/test/test_cmd_line_script.py
```diff
@@ -438,6 +438,15 @@ def test_issue20884(self):
             self.assertEqual(b"", out)
             self.assertEqual(b"", err)
 
+    def test_issue32381(self):
```
aeros, line 441: Instead of `test_issue32381`, I would recommend renaming the method to something more descriptive of the issue itself, such as `test_windows_non_ascii_path`. 
```suggestion
    # tests bpo-32381
    def test_windows_non_ascii_path(self):
```

Python/fileutils.c
aeros: Agreed with @eryksun. Performing the change across all OS's wouldn't be necessary, since this is a windows specific issue.

aeros (changes requested):

Python/fileutils.c
ZackerySpytz: Thank you, @eryksun, for the comment. One issue is the fact that `_Py_fopen_obj()` calls `set_inheritable()` with `raise == 1`, but `_Py_fopen()` currently uses `make_non_inheritable()`, which calls `set_inheritable()` with `raise == 0`. I think, for non-Windows, the `make_non_inheritable()` call in `_Py_fopen()` should be replaced with `set_inheritable(fileno(f), 0, 1, NULL)`.

eryksun: It's a problem that `_Py_fopen` doesn't require the GIL to be held, but `_Py_fopen_obj` requires it. It may be better to modify `PyRun_SimpleFileExFlags` instead of tinkering with `_Py_fopen`.

(Another issue is with `pymain_run_startup`, which gets the ANSI encoded `PYTHONSTARTUP` filename via `_Py_GetEnv` and passes it to `_Py_fopen`. This is a problem for users with arbitrary Unicode usernames if the startup file is in their profile directory. In Windows we need to pass the filename from the native wide-character environment to `_Py_wfopen`.)

vstinner:
I'm not comfortable with changing `_Py_fopen()`, so I wrote PR #23642 to use `_Py_fopen_obj()` in pythonrun.c. Would you mind to review it?

vstinner (commented):
I dislike the idea of calling PyUnicode_DecodeFSDefault() on any platform. On Unix, fopen() is the *native* API. It should only be done on Windows.

I fixed https://bugs.python.org/issue32381 differently. In general, I would suggest avoiding _Py_fopen() and prefer the variant accept a wide string (wchar_t*) or a Python object (PyObject*).

Should we close this PR?

pymain_run_startup() calls _Py_fopen() on Unix and _Py_wfopen() on Windows.

PyErr_ProgramText() calls _Py_fopen(), but Python no longer calls PyErr_ProgramText() anymore.

vstinner:
I removed _Py_fopen() in PR #23711, so this fix is not longer needed.

Moreover, I fixed https://bugs.python.org/issue32381 encoding issue differently: commit b6d98c10fff6f320f8fdf595c3f9a05d8be4e31d.

PR #14700, opened 2019-07-11
bpo-37555: Update _CallList.__contains__ to respect ANY

ElizabethU:
Greetings! Long time fan of the language, first time submitting a pull request to it.

I have a test on another project that goes something like:

```
@patch('a.place.to.patch')
def test_a_thing_calls_what_it_should(self, my_mock):
    # Set up logic here
    my_mock.assert_has_calls([
        call(
            ANY,
            Decimal('20')
        ),
        call(
            ANY,
            Decimal('10')
        )
    ])
```

Which fails, where my_mock.call_args_list looks like

```
[(<A Django Model>, Decimal('20')), (<A Django Model>, Decimal('10'))]
```

This seems like wrong behavior. ANY should be happy to be compared to anything, even a random object. I've added a test showing the behavior that fails on master.

Doing some digging, I found that in [mock.py](https://github.com/python/cpython/blob/master/Lib/unittest/mock.py#L349) `_CallList` is overriding `__contains__` and comparing each item in the tuples with what I'd passed in to assert_has_calls on the right, which means that instead of using `ANY.__eq__`, it's calling the Django model's `__eq__` with `ANY` as an argument. Django first checks if the thing it's comparing to is another Django model, and returns False if not. So, `<DjangoModel> == ANY` is False, but `ANY == <DjangoModel>` is True. I know that this could also be considered a bug with Django, and I plan to file one with them too, but I don't see any downside to improving the mock library to be more defensive in honoring `ANY` over any other custom class's overridden `__eq__` method.

More digging and I realized the reason this was only a problem on a mock with a spec. Specifically, when a mock is instantiated with a spec, the `_spec_signature` attribute gets assigned, and when `_spec_signature` is not none, `_call_matcher`, which is used to prep the calls for comparison, returns a `tuple` of a string and a `BoundArguments` object instead of a `_Call` ([link here](https://github.com/python/cpython/blob/master/Lib/unittest/mock.py#L822)). When `_call_matcher` returns `_Call` objects, they use `_Call`'s `__eq__` method, which flips the compared calls around, putting `ANY` on the [left](https://github.com/python/cpython/blob/master/Lib/unittest/mock.py#L2412). When `_call_matcher` returns `tuples`, it doesn't use this logic, and `ANY` remains on the right, not being used for comparison. The simplest fix seemed to be to ensure `_call_matcher` always returned a `_Call` object.

I have verified that there are [several tests](https://github.com/python/cpython/blob/adbf178e49113b2de0042e86a1228560475a65c5/Lib/unittest/test/testmock/testhelpers.py#L338) covering the use of spec signature, which make sure that things continue to match regardless of if they are args or kwargs and break when that logic is broken, and none of them break with my change.

https://bugs.python.org/issue37555

### DIFF
--- a/Lib/unittest/mock.py
+++ b/Lib/unittest/mock.py
@@ -348,7 +348,7 @@ class _CallList(list):
         self_list = list(self)
         other_list = list(other)
         # checking equality both directions is necessary for ANY to work
-        return self_list == other_list or other_list == self_list
+        return self_list.__eq__(other_list) or other_list.__eq__(self_list)
 
 
 def _check_and_set_parent(parent, value, name, new_name):
@@ -2412,8 +2412,8 @@ class _Call(tuple):
         self_params = self_args, self_kwargs
         other_params = other_args, other_kwargs
         return (
-            self_params == other_params
-            or other_params == self_params
+            self_params.__eq__(other_params)
+            or other_params.__eq__(self_params)
         )
 
 
--- a/Lib/unittest/mock.py
+++ b/Lib/unittest/mock.py
@@ -814,7 +814,8 @@ class NonCallableMock(Base):
             else:
                 name, args, kwargs = _call
             try:
-                return call(name, sig.bind(*args, **kwargs))
+                bound_call = sig.bind(*args, **kwargs)
+                return call(name, bound_call.args, bound_call.kwargs)
             except TypeError as e:
                 return e.with_traceback(None)
         else:
--- a/Lib/unittest/mock.py
+++ b/Lib/unittest/mock.py
@@ -337,12 +337,18 @@ class _CallList(list):
 
         for i in range(0, len_self - len_value + 1):
             sub_list = self[i:i+len_value]
-            if sub_list == value:
+            if value == sub_list:
                 return True
         return False
 
     def __repr__(self):
         return pprint.pformat(list(self))
+
+    def __eq__(self, other):
+        self_list = list(self)
+        other_list = list(other)
+        # checking equality both directions is necessary for ANY to work
+        return self_list.__eq__(other_list) or other_list.__eq__(self_list)
 
 
 def _check_and_set_parent(parent, value, name, new_name):
@@ -2293,7 +2299,6 @@ def _format_call_signature(name, args, kwargs):
     return message % formatted_args
 
 
-
 class _Call(tuple):
     """
     A tuple for holding the results of a call to a mock, either in the form
@@ -2403,8 +2408,12 @@ class _Call(tuple):
         if self_name and other_name != self_name:
             return False
 
-        # this order is important for ANY to work!
-        return (other_args, other_kwargs) == (self_args, self_kwargs)
+        self_params = self_args, self_kwargs
+        other_params = other_args, other_kwargs
+        return (
+            self_params.__eq__(other_params)
+            or other_params.__eq__(self_params)
+        )
 
 
     __ne__ = object.__ne__
--- a/Lib/unittest/mock.py
+++ b/Lib/unittest/mock.py
@@ -337,7 +337,7 @@ class _CallList(list):
 
         for i in range(0, len_self - len_value + 1):
             sub_list = self[i:i+len_value]
-            if sub_list == value:
+            if value == sub_list:
                 return True
         return False
 
--- a/Lib/unittest/mock.py
+++ b/Lib/unittest/mock.py
@@ -348,7 +348,7 @@ class _CallList(list):
         self_list = list(self)
         other_list = list(other)
         # checking equality both directions is necessary for ANY to work
-        return self_list.__eq__(other_list) or other_list.__eq__(self_list)
+        return self_list == other_list or other_list == self_list
 
 
 def _check_and_set_parent(parent, value, name, new_name):
@@ -2411,8 +2411,8 @@ class _Call(tuple):
         self_params = self_args, self_kwargs
         other_params = other_args, other_kwargs
         return (
-            self_params.__eq__(other_params)
-            or other_params.__eq__(self_params)
+            self_params == other_params
+            or other_params == self_params
         )
 
 
--- a/Lib/unittest/mock.py
+++ b/Lib/unittest/mock.py
@@ -864,9 +864,9 @@ class NonCallableMock(Base):
         def _error_message():
             msg = self._format_mock_failure_message(args, kwargs)
             return msg
-        expected = self._call_matcher((args, kwargs))
+        expected = self._call_matcher(_Call((args, kwargs)))
         actual = self._call_matcher(self.call_args)
-        if expected != actual:
+        if actual != expected:
             cause = expected if isinstance(expected, Exception) else None
             raise AssertionError(_error_message()) from cause
 
@@ -926,10 +926,10 @@ class NonCallableMock(Base):
         The assert passes if the mock has *ever* been called, unlike
         `assert_called_with` and `assert_called_once_with` that only pass if
         the call is the most recent one."""
-        expected = self._call_matcher((args, kwargs))
+        expected = self._call_matcher(_Call((args, kwargs), two=True))
+        cause = expected if isinstance(expected, Exception) else None
         actual = [self._call_matcher(c) for c in self.call_args_list]
-        if expected not in actual:
-            cause = expected if isinstance(expected, Exception) else None
+        if cause or expected not in _AnyComparer(actual):
             expected_string = self._format_mock_call_signature(args, kwargs)
             raise AssertionError(
                 '%s call not found' % expected_string
@@ -981,6 +981,22 @@ class NonCallableMock(Base):
             return ""
         return f"\n{prefix}: {safe_repr(self.mock_calls)}."
 
+
+class _AnyComparer(list):
+    """A list which checks if it contains a call which may have an
+    argument of ANY, flipping the components of item and self from
+    their traditional locations so that ANY is guaranteed to be on
+    the left."""
+    def __contains__(self, item):
+        for _call in self:
+            if len(item) != len(_call):
+                continue
+            if all([
+                expected == actual
+                for expected, actual in zip(item, _call)
+            ]):
+                return True
+        return False
 
 
 def _try_iter(obj):
@@ -2133,9 +2149,9 @@ class AsyncMockMixin(Base):
             msg = self._format_mock_failure_message(args, kwargs, action='await')
             return msg
 
-        expected = self._call_matcher((args, kwargs))
+        expected = self._call_matcher(_Call((args, kwargs), two=True))
         actual = self._call_matcher(self.await_args)
-        if expected != actual:
+        if actual != expected:
             cause = expected if isinstance(expected, Exception) else None
             raise AssertionError(_error_message()) from cause
 
@@ -2154,9 +2170,9 @@ class AsyncMockMixin(Base):
         """
         Assert the mock has ever been awaited with the specified arguments.
         """
-        expected = self._call_matcher((args, kwargs))
+        expected = self._call_matcher(_Call((args, kwargs), two=True))
         actual = [self._call_matcher(c) for c in self.await_args_list]
-        if expected not in actual:
+        if expected not in _AnyComparer(actual):
             cause = expected if isinstance(expected, Exception) else None
             expected_string = self._format_mock_call_signature(args, kwargs)
             raise AssertionError(
--- a/Lib/unittest/mock.py
+++ b/Lib/unittest/mock.py
@@ -337,18 +337,12 @@ class _CallList(list):
 
         for i in range(0, len_self - len_value + 1):
             sub_list = self[i:i+len_value]
-            if value == sub_list:
+            if sub_list == value:
                 return True
         return False
 
     def __repr__(self):
         return pprint.pformat(list(self))
-
-    def __eq__(self, other):
-        self_list = list(self)
-        other_list = list(other)
-        # checking equality both directions is necessary for ANY to work
-        return self_list == other_list or other_list == self_list
 
 
 def _check_and_set_parent(parent, value, name, new_name):
--- a/Lib/unittest/mock.py
+++ b/Lib/unittest/mock.py
@@ -337,12 +337,18 @@ class _CallList(list):
 
         for i in range(0, len_self - len_value + 1):
             sub_list = self[i:i+len_value]
-            if sub_list == value:
+            if value == sub_list:
                 return True
         return False
 
     def __repr__(self):
         return pprint.pformat(list(self))
+
+    def __eq__(self, other):
+        self_list = list(self)
+        other_list = list(other)
+        # checking equality both directions is necessary for ANY to work
+        return self_list == other_list or other_list == self_list
 
 
 def _check_and_set_parent(parent, value, name, new_name):
--- a/Lib/unittest/mock.py
+++ b/Lib/unittest/mock.py
@@ -337,7 +337,7 @@ class _CallList(list):
 
         for i in range(0, len_self - len_value + 1):
             sub_list = self[i:i+len_value]
-            if value == sub_list:
+            if sub_list == value:
                 return True
         return False
 
--- a/Lib/unittest/mock.py
+++ b/Lib/unittest/mock.py
@@ -820,7 +820,7 @@ class NonCallableMock(Base):
             else:
                 name, args, kwargs = _call
             try:
-                return name, sig.bind(*args, **kwargs)
+                return call(name, sig.bind(*args, **kwargs))
             except TypeError as e:
                 return e.with_traceback(None)
         else:
--- a/Lib/unittest/mock.py
+++ b/Lib/unittest/mock.py
@@ -337,18 +337,12 @@ class _CallList(list):
 
         for i in range(0, len_self - len_value + 1):
             sub_list = self[i:i+len_value]
-            if value == sub_list:
+            if sub_list == value:
                 return True
         return False
 
     def __repr__(self):
         return pprint.pformat(list(self))
-
-    def __eq__(self, other):
-        self_list = list(self)
-        other_list = list(other)
-        # checking equality both directions is necessary for ANY to work
-        return self_list.__eq__(other_list) or other_list.__eq__(self_list)
 
 
 def _check_and_set_parent(parent, value, name, new_name):
@@ -2300,6 +2294,7 @@ def _format_call_signature(name, args, kwargs):
     return message % formatted_args
 
 
+
 class _Call(tuple):
     """
     A tuple for holding the results of a call to a mock, either in the form
@@ -2409,12 +2404,8 @@ class _Call(tuple):
         if self_name and other_name != self_name:
             return False
 
-        self_params = self_args, self_kwargs
-        other_params = other_args, other_kwargs
-        return (
-            self_params.__eq__(other_params)
-            or other_params.__eq__(self_params)
-        )
+        # this order is important for ANY to work!
+        return (other_args, other_kwargs) == (self_args, self_kwargs)
 
 
     __ne__ = object.__ne__
--- a/Lib/unittest/mock.py
+++ b/Lib/unittest/mock.py
@@ -864,7 +864,7 @@ class NonCallableMock(Base):
         def _error_message():
             msg = self._format_mock_failure_message(args, kwargs)
             return msg
-        expected = self._call_matcher(_Call((args, kwargs), two=True))
+        expected = self._call_matcher(_Call((args, kwargs)))
         actual = self._call_matcher(self.call_args)
         if actual != expected:
             cause = expected if isinstance(expected, Exception) else None
@@ -927,9 +927,9 @@ class NonCallableMock(Base):
         `assert_called_with` and `assert_called_once_with` that only pass if
         the call is the most recent one."""
         expected = self._call_matcher(_Call((args, kwargs), two=True))
+        cause = expected if isinstance(expected, Exception) else None
         actual = [self._call_matcher(c) for c in self.call_args_list]
-        if expected not in actual:
-            cause = expected if isinstance(expected, Exception) else None
+        if cause or expected not in _AnyComparer(actual):
             expected_string = self._format_mock_call_signature(args, kwargs)
             raise AssertionError(
                 '%s call not found' % expected_string
@@ -980,6 +980,23 @@ class NonCallableMock(Base):
         if not self.mock_calls:
             return ""
         return f"\n{prefix}: {safe_repr(self.mock_calls)}."
+
+
+class _AnyComparer(list):
+    """A list which checks if it contains a call which may have an
+    argument of ANY, flipping the components of item and self from
+    their traditional locations so that ANY is guaranteed to be on
+    the left."""
+    def __contains__(self, item):
+        for _call in self:
+            if len(item) != len(_call):
+                continue
+            if all([
+                expected == actual
+                for expected, actual in zip(item, _call)
+            ]):
+                return True
+        return False
 
 
 def _try_iter(obj):
@@ -2155,7 +2172,7 @@ class AsyncMockMixin(Base):
         """
         expected = self._call_matcher(_Call((args, kwargs), two=True))
         actual = [self._call_matcher(c) for c in self.await_args_list]
-        if expected not in actual:
+        if expected not in _AnyComparer(actual):
             cause = expected if isinstance(expected, Exception) else None
             expected_string = self._format_mock_call_signature(args, kwargs)
             raise AssertionError(
--- a/Lib/unittest/mock.py
+++ b/Lib/unittest/mock.py
@@ -864,7 +864,7 @@ class NonCallableMock(Base):
         def _error_message():
             msg = self._format_mock_failure_message(args, kwargs)
             return msg
-        expected = self._call_matcher(_Call((args, kwargs)))
+        expected = self._call_matcher(_Call((args, kwargs), two=True))
         actual = self._call_matcher(self.call_args)
         if actual != expected:
             cause = expected if isinstance(expected, Exception) else None
@@ -2171,9 +2171,9 @@ class AsyncMockMixin(Base):
         Assert the mock has ever been awaited with the specified arguments.
         """
         expected = self._call_matcher(_Call((args, kwargs), two=True))
+        cause = expected if isinstance(expected, Exception) else None
         actual = [self._call_matcher(c) for c in self.await_args_list]
-        if expected not in _AnyComparer(actual):
-            cause = expected if isinstance(expected, Exception) else None
+        if cause or expected not in _AnyComparer(actual):
             expected_string = self._format_mock_call_signature(args, kwargs)
             raise AssertionError(
                 '%s await not found' % expected_string
--- a/Lib/unittest/mock.py
+++ b/Lib/unittest/mock.py
@@ -864,7 +864,7 @@ class NonCallableMock(Base):
         def _error_message():
             msg = self._format_mock_failure_message(args, kwargs)
             return msg
-        expected = self._call_matcher(_Call((args, kwargs)))
+        expected = self._call_matcher(_Call((args, kwargs), two=True))
         actual = self._call_matcher(self.call_args)
         if actual != expected:
             cause = expected if isinstance(expected, Exception) else None
@@ -927,9 +927,9 @@ class NonCallableMock(Base):
         `assert_called_with` and `assert_called_once_with` that only pass if
         the call is the most recent one."""
         expected = self._call_matcher(_Call((args, kwargs), two=True))
-        cause = expected if isinstance(expected, Exception) else None
         actual = [self._call_matcher(c) for c in self.call_args_list]
-        if cause or expected not in _AnyComparer(actual):
+        if expected not in actual:
+            cause = expected if isinstance(expected, Exception) else None
             expected_string = self._format_mock_call_signature(args, kwargs)
             raise AssertionError(
                 '%s call not found' % expected_string
@@ -980,23 +980,6 @@ class NonCallableMock(Base):
         if not self.mock_calls:
             return ""
         return f"\n{prefix}: {safe_repr(self.mock_calls)}."
-
-
-class _AnyComparer(list):
-    """A list which checks if it contains a call which may have an
-    argument of ANY, flipping the components of item and self from
-    their traditional locations so that ANY is guaranteed to be on
-    the left."""
-    def __contains__(self, item):
-        for _call in self:
-            if len(item) != len(_call):
-                continue
-            if all([
-                expected == actual
-                for expected, actual in zip(item, _call)
-            ]):
-                return True
-        return False
 
 
 def _try_iter(obj):
@@ -2172,7 +2155,7 @@ class AsyncMockMixin(Base):
         """
         expected = self._call_matcher(_Call((args, kwargs), two=True))
         actual = [self._call_matcher(c) for c in self.await_args_list]
-        if expected not in _AnyComparer(actual):
+        if expected not in actual:
             cause = expected if isinstance(expected, Exception) else None
             expected_string = self._format_mock_call_signature(args, kwargs)
             raise AssertionError(
--- a/Lib/unittest/test/testmock/testasync.py
+++ b/Lib/unittest/test/testmock/testasync.py
@@ -2,8 +2,8 @@ import asyncio
 import inspect
 import unittest
 
-from unittest.mock import (call, AsyncMock, patch, MagicMock, create_autospec,
-                           _AwaitEvent)
+from unittest.mock import (ANY, call, AsyncMock, patch, MagicMock,
+                           create_autospec, _AwaitEvent)
 
 
 def tearDownModule():
@@ -598,6 +598,30 @@ class AsyncMockAssert(unittest.TestCase):
 
         asyncio.run(self._runnable_test('SomethingElse'))
         self.mock.assert_has_awaits(calls)
+
+    def test_awaits_asserts_with_any(self):
+        class Foo:
+            def __eq__(self, other): pass
+
+        asyncio.run(self._runnable_test(Foo(), 1))
+
+        self.mock.assert_has_awaits([call(ANY, 1)])
+        self.mock.assert_awaited_with(ANY, 1)
+        self.mock.assert_any_await(ANY, 1)
+
+    def test_awaits_asserts_with_spec_and_any(self):
+        class Foo:
+            def __eq__(self, other): pass
+
+        mock_with_spec = AsyncMock(spec=Foo)
+
+        async def _custom_mock_runnable_test(*args):
+            await mock_with_spec(*args)
+
+        asyncio.run(_custom_mock_runnable_test(Foo(), 1))
+        mock_with_spec.assert_has_awaits([call(ANY, 1)])
+        mock_with_spec.assert_awaited_with(ANY, 1)
+        mock_with_spec.assert_any_await(ANY, 1)
 
     def test_assert_has_awaits_ordered(self):
         calls = [call('NormalFoo'), call('baz')]
--- a/Lib/unittest/test/testmock/testasync.py
+++ b/Lib/unittest/test/testmock/testasync.py
@@ -184,6 +184,10 @@ class AsyncAutospecTest(unittest.TestCase):
         spec.assert_awaited_with(1, 2, c=3)
         spec.assert_awaited()
 
+        with self.assertRaises(AssertionError):
+            spec.assert_any_await(e=1)
+
+
     def test_patch_with_autospec(self):
 
         async def test_async():
--- a/Lib/unittest/test/testmock/testasync.py
+++ b/Lib/unittest/test/testmock/testasync.py
@@ -184,10 +184,6 @@ class AsyncAutospecTest(unittest.TestCase):
         spec.assert_awaited_with(1, 2, c=3)
         spec.assert_awaited()
 
-        with self.assertRaises(AssertionError):
-            spec.assert_any_await(e=1)
-
-
     def test_patch_with_autospec(self):
 
         async def test_async():
--- a/Lib/unittest/test/testmock/testhelpers.py
+++ b/Lib/unittest/test/testmock/testhelpers.py
@@ -322,16 +322,6 @@ class CallTest(unittest.TestCase):
         kall = call(1).method(2)(3).foo.bar.baz(4)(5).__int__()
         self.assertEqual(kall.call_list(), mock.mock_calls)
 
-    def test_call_list_with_any_comparison_order(self):
-        class Foo:
-            def __eq__(self, other): pass
-            def __ne__(self, other): pass
-
-        mock = MagicMock()
-        mock(Foo())
-
-        self.assertEqual(call(ANY).call_list(), mock.mock_calls)
-        self.assertEqual(mock.mock_calls, call(ANY).call_list())
 
     def test_call_any(self):
         self.assertEqual(call, ANY)
@@ -341,16 +331,6 @@ class CallTest(unittest.TestCase):
         self.assertEqual(m.mock_calls, [ANY])
         self.assertEqual([ANY], m.mock_calls)
 
-    def test_call_any_comparison_order(self):
-        class Foo:
-            def __eq__(self, other): pass
-            def __ne__(self, other): pass
-
-        m = MagicMock()
-        m(Foo())
-
-        self.assertEqual(m.mock_calls[0], call(ANY))
-        self.assertEqual(call(ANY), m.mock_calls[0])
 
     def test_two_args_call(self):
         args = _Call(((1, 2), {'a': 3}), two=True)
--- a/Lib/unittest/test/testmock/testhelpers.py
+++ b/Lib/unittest/test/testmock/testhelpers.py
@@ -63,7 +63,19 @@ class AnyTest(unittest.TestCase):
             ]
             self.assertEqual(expected, mock.mock_calls)
             self.assertEqual(mock.mock_calls, expected)
+            mock.assert_has_calls(expected)
 
+    def test_assert_has_calls_with_any_and_spec_set(self):
+        # This is a regression test for bpo-37555
+        class Foo(object):
+            def __eq__(self, other): pass
+            def __ne__(self, other): pass
+
+        mock = Mock(spec_set=Foo)
+        expected = [call(ANY)]
+        mock(Foo())
+
+        mock.assert_has_calls(expected)
 
 
 class CallTest(unittest.TestCase):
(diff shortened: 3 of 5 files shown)


the-knights-who-say-ni:

tirkarthi:
This might break below program I posted on the tracker. I am slightly surprised there are no CI failures with this change.

```python
from unittest.mock import call, ANY, Mock

class Foo:

    def __eq__(self, other):
        if not isinstance(other, Foo):
            return False
        return True

m = Mock()
obj = Foo()
m(obj, 1)
m.assert_has_calls([call(ANY, 1)])
```

Without patch

```
➜  cpython git:(master) ./python.exe ../backups/bpo37555_1.py
```

With patch

```
➜  cpython git:(master) git checkout pr_14700 Lib/unittest/mock.py
➜  cpython git:(master) ✗ ./python.exe ../backups/bpo37555_1.py
  File "/Users/karthikeyansingaravelan/stuff/python/cpython/../backups/bpo37555_1.py", line 14, in <module>
    m.assert_has_calls([call(ANY, 1)])
  File "/Users/karthikeyansingaravelan/stuff/python/cpython/Lib/unittest/mock.py", line 900, in assert_has_calls
    raise AssertionError(
Expected: [call(<ANY>, 1)]
Actual: [call(<__main__.Foo object at 0x10b4a5280>, 1)].
```

tirkarthi:
Okay, there is a test at https://github.com/python/cpython/blob/master/Lib/unittest/test/testmock/testhelpers.py#L46 . It's just that it doesn't use `assert_has_calls` where this code path is executed. Adding `mock.assert_has_calls(expected)` to the test fails with the PR that passes on master.

ElizabethU:
Thanks, that's really helpful. I'll see what I can do on this after work today.

ElizabethU:
Okay, I've reverted the original submission and instead added a regression test that shows the specific issue I'm having: When the mock has a spec, `assert_has_calls` is not using `ANY`'s `__eq__`.

I'm working on a fix.

pganssle:
@ElizabethU I can't think of any *obvious* downsides to the current fix, but I do think it's worth noting that this still changes the semantics more broadly than just making it work with `ANY`, because it means that when `(x == y) != (y == x)`, it *always* returns `True`.

In the BPO issue, I suggested that a more tightly scoped fix would be to special-case `mock.ANY` so that if one of the two objects is `mock.ANY`, it would be treated as equal, otherwise the old equality rules would apply. This is just food for thought, though (maybe after this is merged if someone complains about the changes to the semantics we can try the `x is ANY` mechanism).

pganssle:
@ElizabethU Would you mind adding a NEWS entry?

Assuming you would like credit by name, please add "Patch by Elizabeth Uselton." at the end of the blurb. You may also add yourself to [Misc/ACKS](https://github.com/python/cpython/blob/master/Misc/ACKS).

ElizabethU:
Hi @pganssle sorry, I missed your comments over here, since most of the action seemed to be happening on the BPO issue.

1. I agree, special casing `mock.ANY` would be the way to go, however I'm not quite sure I see the path to doing so. `ANY` already has `__eq__` overridden, so surely that's not what you mean. I thought maybe you meant wrapping the `self_list` and `other_list` args in a tuple subclass that would look at each object, and compare the first to the second unless the second was `ANY`, in which case it would return true. But then I realized the second arg could be `{'data': ANY}` or `{'data': {'name': ANY} }`, etc. And then I realized I was writing a recursive JSON parser, and that felt bad, so I stopped. What am I missing here? I'd really like to get to the more elegant solution.

2. I added another commit that I probably should have started with, wrapping the tuple that `_call_matcher` returns if there is a spec passed to the mock with `call()` so that function consistently  returns a `_Call` object. This entirely fixes the problem demonstrated by the first test I added where `assert_has_calls` fails to match `ANY` on badly formed objects with a spec, without affecting the ordering of comparisons on `_Call` or `_CallList`. I still think the ordering fix is an improvement, and that making the other two tests pass is desirable, but I understand that may be more contentious. If you or any other Python caretaker would prefer, I'm happy to cherry-pick the commits for this smaller fix and the first test, and open a second issue for the ordering issue where it can be discussed more.

3. I changed the __eq__ in the comparisons to == after a mentor pointed out to me that `bool(NotImplemented)` is truthy, so without that change a `_Call` and `_CallList` would return `True` instead of `False` if both sides had no way to compare to each other. 😱 I added his name to the acknowledgements too, because he's been a great sounding board for me through this process. I hope that's okay.

I'm happy to keep refining this, especially in regards to #1 or #2, but otherwise I feel good about this and would really like to ship my first (and hopefully not last) Python contribution.

Lib/unittest/mock.py
```diff
@@ -337,13 +337,19 @@ def __contains__(self, value):
 
         for i in range(0, len_self - len_value + 1):
             sub_list = self[i:i+len_value]
-            if sub_list == value:
+            if value == sub_list:
```
tirkarthi: I think this is not needed since order doesn't matter with comparing both ways below in `__eq__` using `(self_params == other_params or other_params == self_params)`.

Lib/unittest/mock.py
```diff
@@ -337,13 +337,19 @@ def __contains__(self, value):
...
         return False
 
     def __repr__(self):
         return pprint.pformat(list(self))
 
+    def __eq__(self, other):
+        self_list = list(self)
+        other_list = list(other)
+        # checking equality both directions is necessary for ANY to work
+        return self_list == other_list or other_list == self_list
```
tirkarthi: This `__eq__` doing similar check can be removed since `_CallList` contains `call` objects and there is `_Call.__eq__` that does `(self_params == other_params or other_params == self_params)`. I am not seeing any test failures on removing this, feel free to point me to tests if any.

Lib/unittest/mock.py
```diff
@@ -2403,8 +2408,12 @@ def __eq__(self, other):
         if self_name and other_name != self_name:
             return False
 
-        # this order is important for ANY to work!
-        return (other_args, other_kwargs) == (self_args, self_kwargs)
+        self_params = self_args, self_kwargs
+        other_params = other_args, other_kwargs
+        return (
+            self_params == other_params
```
tirkarthi: I agree with Paul's comment that this changes semantics. I am not sure if there is some third party code that depends on the order of equality for their own custom objects though as illustrated handles using `ANY` irrespective of order. I will wait for others opinion on this.

Lib/unittest/mock.py
```diff
@@ -814,7 +820,7 @@ def _call_matcher(self, _call):
             else:
                 name, args, kwargs = _call
             try:
-                return name, sig.bind(*args, **kwargs)
+                return call(name, sig.bind(*args, **kwargs))
```
tirkarthi: Below construction looks more correct to me.

```python
bound_call = sig.bind(*args, **kwargs)
_Call((name, bound_call.args, bound_call.kwargs)
```

Lib/unittest/mock.py
ElizabethU: Yup, I should have removed this once the other fixes were working.

Lib/unittest/mock.py
ElizabethU: You're totally right. Embarrassed I missed that one. Thanks.

Lib/unittest/mock.py
ElizabethU: Thank you, this feels much better. I was kind of abusing the idea of what a `_Call` was before, even if it was working.

Lib/unittest/mock.py
ElizabethU: Yeah, I agree this is the riskier of the two changes, and like I said, I'm happy to cherry-pick out the commits having to do with this if we just want to merge the other and hold off on this or think about it more. I'm having a hard time picturing someone using order of equality on calls for something purposeful, and an easy time imagining someone being confused that the two tests I added about this don't pass. But if someone was depending on this, it'd be really interesting to find out why. Maybe Paul will give me a hint on his more tightly scoped special casing `ANY` solution?

ElizabethU: I think I’ve got the special casing ANY solution, I’ll try to post it soon.

ElizabethU:
Hey, I've reverted all my commits related to the `_Call.__eq__` part of this problem, since that seemed to have more uncertainty around it that the original bug that brought me here of `assert_has_calls` not working with `ANY` when specced on a class that's `__eq__` doesn't return `NotImplemented`. That original bug is totally fixed by this patch.

I'm not quite ready to give up on it yet, but thought it would be better served by being a separate ticket we could all discuss more, and that hopefully there could be consensus around this smaller change first.

ElizabethU:
@tirkarthi @pganssle Have you had a chance to see these changes? I updated the initial comment on this to explain the bug and fix more accurately, and think this de-scoped solution has less opportunity for consequences. Happy to do any further work on it if you have concerns or critiques.

pganssle:
@ElizabethU Sorry for the delay, I will try to look more closely at this this weekend. It looks like a much more minimal change, which is encouraging, I just don't really understand the mock module well enough to understand how and why this works.

I would have thought that changing the return type of `_call_matcher` would need some corresponding change in the function that consumes its output, but from a cursory glance at the code it looks like mostly the outputs of `_call_matcher` just get compared to the outputs of other calls to `_call_matcher`.

ElizabethU:
@pganssle Thanks and no worries about the wait. I'm sorry, I definitely don't mean to be one of those folks demanding open source maintainers get back to them immediately, like this isn't a thing y'all do on your free time out of the goodness of your hearts, I'm just still figuring out the social norms on how much it's my responsibility to ping people to not just fall off the radar. 

It's really just changing the return type so that it's consistent for whether there's a spec or not. `_call_matcher` has always returned a `_Call` when there's no spec, and so I don't think it's as surprising that it handles them just fine. It looks like `_call_matcher` only started returning a tuple sometimes when the signature binding was added in 2013. I updated the initial comment on this PR with some links to lines I think help guide through the reasons this fixes it.

pganssle:
> @pganssle Thanks and no worries about the wait. I'm sorry, I definitely don't mean to be one of those folks demanding open source maintainers get back to them immediately, like this isn't a thing y'all do on your free time out of the goodness of your hearts, I'm just still figuring out the social norms on how much it's my responsibility to ping people to not just fall off the radar.

To be absolutely clear on this point, I think you have been doing everything right in this PR and I did not feel any undue demand on my time. I know it can be disheartening to push out changes for review and then just get silence so I wanted to just give a little status update as to why I haven't requested changes or accepted the PR.

Thank you for your perseverance in this and if I or someone else hasn't either accepted or requested changes on this PR in say 1 week feel free to ping me again and I will either make sure to review it or find and start badgering the right person to review it on your behalf :smiling_imp:. We've got a pretty bad backlog, but it doesn't help anything if contributors (particularly high quality contributors such as yourself) have a bad experience!

pganssle (approved):
OK, I have dug into this and I think it's a very elegant fix.

@tirkarthi Do you mind giving this another review? If either you or @mariocj89 approves this PR I'd be happy to merge it.

mariocj89 (approved):
Great!

Land it! Looks good :)

Even if the return type is changed all that that it is used for is for comparing different calls on the expected functions, this should just improve the situation, cannot think of any issue ^^.

tirkarthi:
Unfortunately, this seems to break `assert_called_with` and `assert_any_call` as below example works on master but breaks with this patch. This is also not caught by test suite I think there could be cases that depend on the return value of `_call_matcher` being tuple :( Before the PR they were tuple with `expected.__eq__(actual)` with expected falling on left side having ANY. With expected being a call object now it will become `actual == expected` internally since `actual` is passed as other to `call.__eq__`.

```python
from unittest.mock import call, ANY, Mock

class Foo:

    def __eq__(self, other):
        pass

m = Mock(spec=Foo)
obj = Foo()
m(obj, 1)
m.assert_called_with(ANY, 1)
m.assert_any_call(ANY, 1)
```

```python
  File "/Users/karthikeyansingaravelan/stuff/python/cpython/bpo37555_3.py", line 11, in <module>
    m.assert_called_with(ANY, 1)
  File "/Users/karthikeyansingaravelan/stuff/python/cpython/Lib/unittest/mock.py", line 871, in assert_called_with
    raise AssertionError(_error_message()) from cause
AssertionError: expected call not found.
Expected: mock(<ANY>, 1)
Actual: mock(<__main__.Foo object at 0x10fa73c80>, 1)
```

ElizabethU:
Oh no! Thank you so much for noticing this. In retrospect this makes perfect sense and I'll bet it's related to the commit I had earlier where I noted assertEqual is not order agnostic. I'll take a look at this after work, but I think I'll find that `m.assert_called_with(ANY, 1)` needs to be `m.assert_called_with(1, ANY)` when there's no spec (assuming the same `Foo` class), and that the order for all these methods depends on whether there's a spec or not. I'm pretty sure I know how to fix this, but thank you so much for not letting my first commit to python break things.

tirkarthi:
You're welcome. No worries, asserting with spec is a little grey area and there are already issues in tracker that are complicated. My first thoughts were to have `call` objects to be only returned for `assert_has_calls` by `_call_matcher` as below patch but it feels little hacky and special casing to me. Maybe there is a better way to fix it. Feel free to push the changes I will try to review it. Please add my example as unit test so that this case is also covered in the test suite.

```patch
diff --git a/Lib/unittest/mock.py b/Lib/unittest/mock.py
index b3dc640f8f..47148fba00 100644
--- a/Lib/unittest/mock.py
+++ b/Lib/unittest/mock.py
@@ -804,7 +804,7 @@ class NonCallableMock(Base):
         return message % (action, expected_string, actual_string)


-    def _call_matcher(self, _call):
+    def _call_matcher(self, _call, return_call=False):
         """
         Given a call (or simply an (args, kwargs) tuple), return a
         comparison key suitable for matching with other calls.
@@ -819,7 +819,11 @@ class NonCallableMock(Base):
             else:
                 name, args, kwargs = _call
             try:
-                return name, sig.bind(*args, **kwargs)
+                if return_call:
+                    bound_call = sig.bind(*args, **kwargs)
+                    return call((name, bound_call.args, bound_call.kwargs))
+                else:
+                    return name, sig.bind(*args, **kwargs)
             except TypeError as e:
                 return e.with_traceback(None)
         else:
@@ -897,9 +901,9 @@ class NonCallableMock(Base):

         If `any_order` is True then the calls can be in any order, but
         they must all appear in `mock_calls`."""
-        expected = [self._call_matcher(c) for c in calls]
+        expected = [self._call_matcher(c, return_call=True) for c in calls]
         cause = expected if isinstance(expected, Exception) else None
-        all_calls = _CallList(self._call_matcher(c) for c in self.mock_calls)
+        all_calls = _CallList(self._call_matcher(c, return_call=True) for c in self.mock_calls)
         if not any_order:
             if expected not in all_calls:
                 raise AssertionError(
stash@{1}: WIP on master: 2a3d4d9c53 Document default parameter of .seek() in the signature. (GH-14691)
```

ElizabethU:
@tirkarthi I refactored the test you suggested I add so that it's four tests covering all the functions using `_call_matcher` both with and without spec, so every use case would be covered. By doing this I found out that `assert_has_awaits` has the same bug as `assert_has_calls`, which makes sense given how similar they are.

Your patch with the extra parameter felt like special casing to me too, so I spent some time trying to really understand the reasons the different cases were failing, and the more I looked at it, the more I felt like the root of the problem is that `_call_matcher` sometimes takes `_Calls` and sometimes doesn't, and sometimes returns `_Calls` and sometimes doesn't, sometimes returning a different type than what was passed in. The solution I've implemented makes ` _call_matcher` consistently use and return `_Call` objects, which feels simpler to reason about or extend off of for future developers. This isn't the simplest change I could have made, so I'm open to doing something else. I'm not particularly happy about the `_AnyComparer` class, but as a list subclasses to deal with the flipped `item == self` at the end of `_Call.__eq__`, it's just following a precedent set by `_CallList`.

Here are some other options that I tried out and think would work, please let me know if you have a strong preference for one of them over what I ended up doing.

1. Convert anything coming back from `_call_matcher` to a `_Call` instance in `assert_has_calls` and `assert_has_awaits`. Has a smaller scope but leaves the problem with `_call_matcher` being used for inconsistent things.

2. Convert anything coming into `_CallList.__eq__` to a `_Call`, since `_CallList` should always be a list of `_Calls`. Same pros and cons as option 1 above, really feels better to assure `_CallList` gets the right arg types than to try to force it.

3. Check the type of arguments to `_call_matcher` and return something of the same type. Seems like adding logic and less explicit.

4. Special case `ANY` in `Call.__eq__` so it's respected whether it's on the left or right of a comparison, as Paul suggested earlier. I like this one the most out of all the other working ideas, and think it has some good side effects that might be a great case for it even independent of this bug, but this bug is caused by conflating types in `_call_matcher`, and it feels like leaving that unfixed is inviting more complexity to this code, even if we can find a way to get around it.

5. Bonus: My favorite idea that sadly doesn't work: If we could override a function ANY to say that it's a subclass of everything, similar to how we override __eq__ on it to say it's equal to everything. The python `==` algorithm pulls the subclass on the left, so `ANY.__eq__` would always get used in any comparison involving `ANY`. Unfortunately, this is not possible because `__subclasscheck__` would only let me make everything a subclass of `ANY`, not the other way around, and also rich comparison uses `PyType_IsSubtype` which actually walks the inheritance, not `PyObject_IsSubclass` which uses `__subclasscheck__`. Just a fun idea I got excited about before I realized it wouldn't work.

Those are my thoughts, but I'm very happy to change to a different approach if you, @pganssle @mariocj89 or anyone else have opinions.

ElizabethU:
@tirkarthi @pganssle Hey there, not trying to be impatient, just checking in because I haven't heard anything in a while. I'm sorry about the amount of iteration this has needed, is there anything I can do to make it easier to review?

ElizabethU:
@tirkarthi @pganssle @mariocj89 Hi again, I know you mentioned there's quite a backlog to get through, so hopefully it's just that, but after this long of not hearing anything, I'm worried reviewing this has fallen off the to do list, or that I've done something egregiously bad in this latest draft. I'd really appreciate it if someone could let me know this isn't abandoned, even if you don't have time for it today.

pganssle:
At this point I mostly want to defer to @tirkarthi, because he seems to understand the edge cases better than the rest of us. I am happy to do a final review and merge once he gives approval.

ElizabethU:
@pganssle Totally understand, @tirkarthi 's input has been really helpful, and after the last bug he caught, I'd feel worried about merging this without his approval too. Unfortunately, that still leaves me in the same position of worrying it's fallen off the to do list. Is there any way I could ask you to help (earlier you were mentioning helping track down the reviewer I need), or is the best thing just for me to ping him again if I don't hear back in another week or so?

Also, I want to be open to the idea that I'm just not used to the speed at which open source moves, if I should just wait patiently and trust that it'll get reviewed with time?

tirkarthi:
I will try to give a review of it in couple of weeks given that this is little complex area. Sorry, I will be having little time for open source given my day job workload :( I will get back to you by September mid and I haven't forgotten the issue. Friendly ping to @cjw26 who can also give some helpful input since they have experience with mock codebase.

Thanks

ElizabethU:
Thank for the update @tirkarthi Sorry if I've been pushy about it, that wasn't my intention. I'll put this on the back burner and maybe dig up something else to work on in the mean time, since you said there are other issues in the tracker around this.

voidspace:
This is a high priority issue and thanks for the work on it, especially the places that need to change.

I think explicitly checking for ANY rather than reversing the order of arguments will be clearer and less likely to have other, potentially subtle, effects.

voidspace:
No, we can't explicitly check for ANY because we're comparing data structures. ANY could appear anywhere! @tirkarthi has a slight change to this PR that we're looking on getting committed during the core-dev sprints this week.

ElizabethU:
Hi @voidspace I was reluctant to just check for ANY because while the documentation for ANY doesn't explicitly say ANY can be used in nested objects, it supports it now, and it would feel bad to take that away from people, which I think is what you're saying, too.

This is my first contribution, so I'm new to how this works. What does that mean that @tirkarthi has a slight change to this PR that you're looking to commit? Will that use some of my commits? Will I still be a contributor? I'm very excited and invested in this bug and would love to stay with it after all the work I've put in on it, if I can make any changes to get this merged.

tirkarthi:
@ElizabethU I have made some changes to this in my branch few weeks back . My change is at https://github.com/tirkarthi/cpython/commits/pr_14700_re and commit change is https://github.com/tirkarthi/cpython/commit/bc099891dd327e00bc0271da9ddb752ae9198ce6. We can use this PR. I will add in the suggestions tomorrow as PR comments and you can review and comment if my comments are okay. You will remain as the author and this will be the same like the review changes we made earlier. Thanks.

Lib/unittest/mock.py
```diff
@@ -925,10 +926,10 @@ def assert_any_call(self, /, *args, **kwargs):
         The assert passes if the mock has *ever* been called, unlike
         `assert_called_with` and `assert_called_once_with` that only pass if
         the call is the most recent one."""
-        expected = self._call_matcher((args, kwargs))
+        expected = self._call_matcher(_Call((args, kwargs), two=True))
+        cause = expected if isinstance(expected, Exception) else None
         actual = [self._call_matcher(c) for c in self.call_args_list]
-        if expected not in actual:
-            cause = expected if isinstance(expected, Exception) else None
+        if cause or expected not in _AnyComparer(actual):
```
tirkarthi, line 932: I think as per my commit we can remove `_AnyComparer`. It takes the expected value which could contain ANY and make sure it stays on the left while comparing with actual. The calllist `__contains__` already does the same check here with https://github.com/python/cpython/blob/455122a0094c8cfdf7e062eccc5e5b5885c75e8b/Lib/unittest/mock.py#L340. The list `__contains__` does an equal per item and hence it does the same check at `_Call.__eq__` https://github.com/python/cpython/blob/455122a0094c8cfdf7e062eccc5e5b5885c75e8b/Lib/unittest/mock.py#L2394

There are also no test failures and I see the behavior around comparison to be the same. We also can skip the check for cause as I go with the approach. @ElizabethU  Thoughts?

ElizabethU: @tirkarthi I'm all for removing `_AnyComparer` as I think it's pretty clunky, but I'm seeing test failures with [tirkarthi@bc09989]. First I cherry-picked your commit onto my branch and saw test failures, then I tried on the branch on your fork, thinking that you might have had some other changes or even that someone else might have made some changes that you'd have if you'd synced your fork since I'd opened this PR. I'm still getting the same 4 failures either way:

`unittest.test.testmock.testasync.AsyncMockAssert.test_awaits_asserts_with_any`
`unittest.test.testmock.testasync.AsyncMockAssert.test_awaits_asserts_with_spec_and_any`
`unittest.test.testmock.testhelpers.AnyTest.test_any_and_spec_set`
`unittest.test.testmock.testhelpers.AnyTest.test_any_no_spec`

Are those four tests working for you? If so, I'm pretty confused since these failures really make sense to me. In `assert_any_call` above, whether `actual` is just a regular list or a `_CallList`, it goes into the default list class's `__contains__` either way. And when it does, `call(<ANY>, 1)` is the `value`, and it gets a little confusing, but the way item by item comparison is done, when it enters `_Call.__eq__` https://github.com/python/cpython/blob/455122a0094c8cfdf7e062eccc5e5b5885c75e8b/Lib/unittest/mock.py#L2394
`self` is `call(<ANY>, 1)` and `other` is an item from the list, meaning that when it hits the reversed comparison at the end of the method
https://github.com/python/cpython/blob/455122a0094c8cfdf7e062eccc5e5b5885c75e8b/Lib/unittest/mock.py#L2444
`ANY` is on the right, and it doesn't work.

At least that's what I'm seeing on my machine. I'd be really excited if you've found a way around `_AnyComparer`, but it seems like we still need it.

tirkarthi: I don't see any failures on checking out your branch and cherry picking my changes on to your branch. My terminal output as below assuming I am doing it correct.

```
➜  cpython git:(pr_14700_re) git pr 14700
From github.com:python/cpython
 * [new ref]               refs/pull/14700/head -> pr_14700
➜  cpython git:(pr_14700_re) git checkout pr_14700
Updating files: 100% (506/506), done.
Switched to branch 'pr_14700'
➜  cpython git:(pr_14700) git cherry-pick bc099891dd327e00bc0271da9ddb752ae9198ce6
[pr_14700 5b7c21b188] Remove AnyCompare and use call objects everywhere.
 Date: Mon Aug 19 15:11:01 2019 +0530
 2 files changed, 8 insertions(+), 21 deletions(-)
➜  cpython git:(pr_14700) ./python.exe -m unittest Lib.unittest.test.testmock
...........................................................................................................................................................................................................................................................................................................................................................................................................................................................
Ran 443 tests in 1.470s

OK
```

tirkarthi: I also don't have any changes more to push too as I can see.

```
➜  cpython git:(pr_14700_re) git show --summary | cat
commit bc099891dd327e00bc0271da9ddb752ae9198ce6
Author: Xtreak <tir.karthi@gmail.com>
Date:   Mon Aug 19 15:11:01 2019 +0530

    Remove AnyCompare and use call objects everywhere.

➜  cpython git:(pr_14700_re) ./python.exe -m unittest Lib.unittest.test.testmock
.............................................................................................................................................................................................................................................................................................................................................................................................................................................................
Ran 445 tests in 1.449s

OK
➜  cpython git:(pr_14700_re) git push origin pr_14700_re
Everything up-to-date
```

Lib/unittest/mock.py
```diff
@@ -863,9 +864,9 @@ def assert_called_with(self, /, *args, **kwargs):
         def _error_message():
             msg = self._format_mock_failure_message(args, kwargs)
             return msg
-        expected = self._call_matcher((args, kwargs))
+        expected = self._call_matcher(_Call((args, kwargs)))
         actual = self._call_matcher(self.call_args)
-        if expected != actual:
+        if actual != expected:
```
asvetlov, line 869: What is the difference?

Lib/unittest/mock.py
tirkarthi: > self is call(<ANY>, 1) and other is an item from the list, meaning that when it hits the reversed comparison at the end of the method
> ANY is on the right, and it doesn't work.

It's a list and `list.__contains__` is used only when it's not a list. So it takes the other code path `sub_list == value` to where `value` is the tuple that contains ANY. ANY is on the left as `other_args` and `other_kwargs`. as I can see.

Lib/unittest/mock.py
tirkarthi: Comparison of ANY depends on the order unfortunately and how python `__eq__` works. So since we made everything as a call object we need `expected` to be on the right. Previously expected is a tuple and is on left like below with ANY.__eq__ used : 

`(ANY, ) == Foo()`

Now it's a call object so it should be on the right since `call.__eq__` reverses the items so that ANY is on right now but on comparison ends on the left side.

`call(Foo()) == call(ANY)`

Lib/unittest/mock.py
tirkarthi: FWIW, I rebased my branch on latest master and did a clean rebuild and don't see test failures. I also ran the tests as python code in file to make sure my assumption is correct.

Lib/unittest/mock.py
asvetlov: Got it, thanks for the clarification

Lib/unittest/mock.py
ElizabethU: > It's a list and `list.__contains__` is used only when it's not a list. So it takes the other code path `sub_list == value` to where `value` is the tuple that contains ANY. ANY is on the left as `other_args` and `other_kwargs`. as I can see.

Are we talking about the same thing? The failures I'm seeing are for `assert_any_call` and `assert_any_awaits`, where `_CallList` is not even used. So `list.__contains__` is used because it's a `actual` is a regular list, and `expected` is not a list, it's a single call.

I believe you, but I'm still getting these failures.
```
FAIL: test_awaits_asserts_with_any (unittest.test.testmock.testasync.AsyncMockAssert)
  File "/Users/elizabethuselton/projects/cpython/Lib/unittest/test/testmock/testasync.py", line 614, in test_awaits_asserts_with_any
    self.mock.assert_any_await(ANY, 1)
  File "/Users/elizabethuselton/projects/cpython/Lib/unittest/mock.py", line 2166, in assert_any_await
    raise AssertionError(
AssertionError: mock(<ANY>, 1) await not found

FAIL: test_awaits_asserts_with_spec_and_any (unittest.test.testmock.testasync.AsyncMockAssert)
  File "/Users/elizabethuselton/projects/cpython/Lib/unittest/test/testmock/testasync.py", line 628, in test_awaits_asserts_with_spec_and_any
    mock_with_spec.assert_any_await(ANY, 1)
  File "/Users/elizabethuselton/projects/cpython/Lib/unittest/mock.py", line 2166, in assert_any_await
    raise AssertionError(
AssertionError: mock(<ANY>, 1) await not found

FAIL: test_any_and_spec_set (unittest.test.testmock.testhelpers.AnyTest)
  File "/Users/elizabethuselton/projects/cpython/Lib/unittest/test/testmock/testhelpers.py", line 88, in test_any_and_spec_set
    mock.assert_any_call(ANY, 1)
  File "/Users/elizabethuselton/projects/cpython/Lib/unittest/mock.py", line 939, in assert_any_call
    raise AssertionError(
AssertionError: mock(<ANY>, 1) call not found

FAIL: test_any_no_spec (unittest.test.testmock.testhelpers.AnyTest)
  File "/Users/elizabethuselton/projects/cpython/Lib/unittest/test/testmock/testhelpers.py", line 76, in test_any_no_spec
    mock.assert_any_call(ANY, 1)
  File "/Users/elizabethuselton/projects/cpython/Lib/unittest/mock.py", line 939, in assert_any_call
    raise AssertionError(
AssertionError: mock(<ANY>, 1) call not found

Ran 445 tests in 3.917s

FAILED (failures=4)
Rover-ElizabethUselton:cpython elizabethuselton$ git log
commit bc099891dd327e00bc0271da9ddb752ae9198ce6 (HEAD -> pr_14700_re, tirkarthis-fork/pr_14700_re)
Author: Xtreak <tir.karthi@gmail.com>
Date:   Mon Aug 19 15:11:01 2019 +0530

    Remove AnyCompare and use call objects everywhere.

commit 7a89117166dd640efed0b322a6584f1497383b61
Author: Elizabeth Uselton <elizabeth.uselton@rover.com>
Date:   Mon Aug 5 00:51:24 2019 -0700

    bpo-37555: Ensure all assert methods using _call_matcher are actually passing calls
```

I'm going to push your cherry-picked commit up to this PR, so either CI will show you the failures I'm getting, or I'll learn the good news that it's just my local build that's somehow messed up and have an adventure figuring out why.

tirkarthi: > I'm going to push your cherry-picked commit up to this PR, so either CI will show you the failures I'm getting, or I'll learn the good news that it's just my local build that's somehow messed up and have an adventure figuring out why.

CI passes. But just in case you can try `git clean -xdf && ./configure -q --with-pydebug && make -s -j4` and it should give you a clean rebuild to test locally.

ElizabethU: Whoa! I'm extremely happy to be wrong.

I'm trying to do the clean rebuild you suggested, but getting 
```
Could not build the ssl module!
Python requires an OpenSSL 1.0.2 or 1.1 compatible libssl with X509_VERIFY_PARAM_set1_host().
LibreSSL 2.6.4 and earlier do not provide the necessary APIs, https://github.com/libressl-portable/portable/issues/381
```
And tests still failing.

I'd be really interested to stick a breakpoint in a working copy, right above the `in` check in `assert_any_call` 
https://github.com/python/cpython/blob/455122a0094c8cfdf7e062eccc5e5b5885c75e8b/Lib/unittest/mock.py#L968
running `unittest.test.testmock.testhelpers.AnyTest.test_any_no_spec`
and see what `expected` and `actual` are, and what `self` and `other` are after stepping into the `_Call.__eq__` function
https://github.com/python/cpython/blob/455122a0094c8cfdf7e062eccc5e5b5885c75e8b/Lib/unittest/mock.py#L2394
because I'm right now I'm getting
```
cpython elizabethuselton$ ./python.exe -m unittest -v unittest.test.testmock.testhelpers.AnyTest.test_any_no_spec
test_any_no_spec (unittest.test.testmock.testhelpers.AnyTest) ... > /Users/elizabethuselton/projects/cpython/Lib/unittest/mock.py(932)assert_any_call()
-> if expected not in actual:
(Pdb) expected
call(<ANY>, 1)
(Pdb) actual
[call(<unittest.test.testmock.testhelpers.AnyTest.test_any_no_spec.<locals>.Foo object at 0x10ca94780>, 1)]
(Pdb) s
--Call--
> /Users/elizabethuselton/projects/cpython/Lib/unittest/mock.py(2356)__eq__()
-> def __eq__(self, other):
(Pdb) self
call(<ANY>, 1)
(Pdb) other
call(<unittest.test.testmock.testhelpers.AnyTest.test_any_no_spec.<locals>.Foo object at 0x10ca94780>, 1)
```
Which then leads to `ANY` being in the `self_args`, and so on the right, and the test failing.

I don't think I can get my local working before I go to work for the day, but I'm happy for this to be a working solution and not to block it on me figuring out my local dev environment.

ElizabethU: Update: I have `cpython` working on a second machine, and the same tests failing there. 🤷‍♀️😟. I also had the SSL module build problem there, but it doesn't seem like those should be connected. That's a little worrying to me.

tirkarthi: SSL module build is optional. Sorry, I am not sure how to debug it there. I do `git clean -xdf` and do a full rebuild. I have done clean rebuilds on your branch, my branch and with both synced with latest master and don't see tests failing just like the CI. Would be happy to know your input on the patch so that I can do a final review and approve tomorrow.

ElizabethU: I don't suppose anyone at the core dev sprints has time to pull down this branch and report on what they see, or if you could let me know what you see with a breakpoint on the lines I mentioned above so I can at least get a sense of *why* it works for you and CI?

ElizabethU: > Would be happy to know your input on the patch so that I can do a final review and approve tomorrow.

I don't really know how to offer constructive input on the patch without being able to see it working locally. If it works consistently, I'm very happy with it and would love to not have `_AnyComparer`. But when I read it, I though it would break on `assert_any_call` and `assert_any_await`, and when I ran it, that's exactly where it broke. I've rebuilt, I've tried it on two different computers and got another developer to try it on theirs, and all had it fail on the four tests I mentioned above. 

But yours passes and the CI passes. This is the weirdest "works on my machine" I've ever been a part of.

What can we do that could make us reasonably sure that this is just some weird fluke on my (and another developer's) computers? Can you get this working on another developer's computer at the dev sprints? Could you use a breakpoint and tell me what `expected` and `actual` are on your local at
https://github.com/ElizabethU/cpython/blob/master/Lib/unittest/mock.py#L930
and what `self` and `other` are at https://github.com/ElizabethU/cpython/blob/master/Lib/unittest/mock.py#L2355
in `unittest.test.testmock.testhelpers.AnyTest.test_any_no_spec` just so I can see what's different?

I'm sorry to be so much work, I know I should be really excited to have my first change finally approved, but I want it to be safe, y'know? I'll get up early so if you have any more insight in the morning, I can get back to you by your afternoon.

tirkarthi: I am at the sprints for the last day too. I will probably have less time but will try. I am not sure why you are facing this error and in another computer. I have done rebuilds several times and also verified my code paths along with CI on different machines. I will later post the breakpoints you have requested as a gist with value of stack at the point.

tirkarthi: Okay, I am able to reproduce the failures in your branch without rebasing on master. On rebasing on master there is no error. Before rebasing the `other` argument has the actual object without ANY. After rebasing on master the `other` argument has the actual object with ANY (arguments are swapped). I guess you are missing some change done as part of the master regarding contains check. Can you please try rebasing on master and then try? Wild guess it's due to 18b711c5a7f90d88fb74748f18fa8ef49d8486c7 (https://bugs.python.org/issue37648) but I haven't verified it yet. If it is so then my fix cannot be backported to 3.8 since the original commit is also not present.

On CI it always merges with latest master and then runs the tests. Since your and mine were old branches I usually rebase and keep it up to date while compiling which probably explains this.

tirkarthi: Confirmed that this change in the order of comparison to be crucial to my patch over change in the arguments. That means we can either 

1. Have ANYCompare and apply this fix on 3.9 and 3.8 . I am not sure of backporting this to 3.7 at this point since it's in 3.7.5 but I will leave it upto the core dev.
2. Depend on the below order in list comparison and remove ANYCompare to use just contains and keep the fix only to 3.9 since the listobject patch is not on 3.8 and below.

Thoughts?

```
diff --git a/Objects/listobject.c b/Objects/listobject.c
index d012ab933a..cea9b24a3b 100644
--- a/Objects/listobject.c
+++ b/Objects/listobject.c
@@ -449,8 +449,7 @@ list_contains(PyListObject *a, PyObject *el)
     int cmp;

     for (i = 0, cmp = 0 ; cmp == 0 && i < Py_SIZE(a); ++i)
-        cmp = PyObject_RichCompareBool(el, PyList_GET_ITEM(a, i),
-                                           Py_EQ);
+        cmp = PyObject_RichCompareBool(PyList_GET_ITEM(a, i), el, Py_EQ);
     return cmp;
 }
```

tirkarthi: Thinking a bit further there is also mock backport which would need to run on 3.6+ so this would cause problems. Now I am more leaning towards to reverting my change to keep ANYCompare though the interpreter is fixed in 3.9+. At this point I think I depended on the patch in 3.9 during debugging and made some incorrect assumptions with respect to ordering, sorry about that. I would be okay with this landing on 3.8+ with ANYcompare and the backport can work on 3.6+ with the PR.

@ElizabethU Feel free to revert my patch.

ElizabethU: I am able to get it to work by rebasing 🎉 🎊 🎺 Thanks, I should have tried that earlier, but I thought since I was still having the problem on your branch and you had that passing, that it couldn't be a rebase. I should have tried anyway.

My input on the patch for 3.9 is that it's great and a big improvement over what I had alone and I would love to see it get reviewed whenever there's time. As for backporting, maybe we'd still need the `_AnyComparer` class for 3.8 then? Or maybe this just doesn't get backported? Not sure what's the norm, but I'm happy to learn and go along with it.

Thank you so, so much for sticking with me through this. I've learned so much, and your expertise has been a big part of that. I hope I can contribute again and take up a lot less of your time now that I know the routine and some of the codebase.

ElizabethU: @tirkarthi Ah, I didn't see your last comments. So reverting is the right thing to do? I do like the patch, and was hoping we could keep it for 3.9, but I'm unfamiliar with the mock backport situation, and why it means we couldn't use the patch in just 3.9?

I'm reverting the patch per your instructions.

cjw296: FWIW, anything that means things work on 3.6+ would be very much appreciated by me :-)

tirkarthi, line 932: One more case: Please do the check for cause to be exception `if cause or expected not in _AnyComparer(actual):` in `assert_any_await` too to catch the exception like you did in `assert_any_call`. This would cause below test to fail. I had a test in my patch for this case. Something like below test :

```
diff --git a/Lib/unittest/test/testmock/testasync.py b/Lib/unittest/test/testmock/testasync.py
index 1b1f9111d2..87f3cfc8fa 100644
--- a/Lib/unittest/test/testmock/testasync.py
+++ b/Lib/unittest/test/testmock/testasync.py
@@ -192,6 +192,10 @@ class AsyncAutospecTest(unittest.TestCase):
         spec.assert_awaited_with(1, 2, c=3)
         spec.assert_awaited()

+        with self.assertRaises(AssertionError):
+            spec.assert_any_await(e=1)
+
+
     def test_patch_with_autospec(self):

         async def test_async():
```

ElizabethU: Thank you @tirkarthi ! I also saw in your patch you added the `two=True` parameter to `assert_called_with` which strikes me as good for consistency, so I added that back in too.

tirkarthi (approved):
LGTM. Thank you very much Elizabeth for your patience and tenacity throughout the review process though the process unfortunately turned out to be long due to the complexity of the bug and internals of how `__eq__` works in CPython :) Your bug report also helped in the 3.9 change discussion and fixing an issue in datetime module. Apologies for the confusion from my patch which worked correctly due to the change in master which I was thinking to be fine along in 3.8 too.

Looking forward to your contributions :)

pganssle (approved):
Great work everyone, thanks again!

pganssle:
Ugh, wtf, I wrote a custom commit message and changed the commit title and github ate it and applied the default one :frowning:

matrixise:
@pganssle question, this PR is not backported to 3.8, is there a reason? Thanks

pganssle:
@matrixise My understanding is that we were unable to fix this without some very minor backwards compatibility breaking - the order of an `==` comparison is swapped, for example. I highly doubt anyone would notice the difference, honestly, but my instinct is to leave this for 3.9, particularly if it will be included in the backport.

@tirkarthi Can you comment as to whether or not the final version of this ended up with any backwards incompatibilities? Should we go ahead and backport to 3.8?

tirkarthi:
> @tirkarthi Can you comment as to whether or not the final version of this ended up with any backwards incompatibilities? Should we go ahead and backport to 3.8?

My version of the patch was backwards incompatible but the one on current master doesn't depend on the incompatible change and should work on mock's backport and 3.8 . I was initially okay with backporting to 3.8. But given the complexity of the issue and also that 3.8 RC1 is just in 2 weeks I don't want this to affect stability though I believe the fix is good. This will be on 3.9 and the mock backport which would give us more time and exposure to testing to address issues if any. My opinion is that as far as I know this issue wasn't noticed and reported for 3.6 and 3.7 I think we can keep it on master and mock's backport due to complexity.

ElizabethU:
Does that mean that this will eventually find it's way to 3.8, after it's had more time to be vetted as part of the mock backport to 3.6? It seems counterintuitive to me that it would be in 3.6 but not in 3.8, but I'm just learning about release schedules and don't have a great understanding of the mock backport.

tirkarthi:
mock has a rolling backport on pypi that now supports 2.7 and 3.5+. Plan would be to drop Python 2 support and have it 3.6+ only so essentially the current mock.py will be copied over. Benefit is that users from 3.6 (3.6 is in security fixes only mode) can install the backport from pypi and use 3.8 features like AsyncMock. Mock backport has several million downloads per month and could get some testing. If there is a regression found then we can fix it in master and again merge it to rolling backport. Meanwhile with 3.8 branch getting a backport now it will have a release candidate in two weeks like a code freeze. So if there is a regression then there would be only few weeks to fix it and get another RC. Given the complexity I would prefer the patch can get testing from mock backport that users can install and 3.9 release. If there are requests over this bug being crucial then it can be considered for a 3.8.x release as a bug fix.

merwok:
For future PRs, please keep the ordering (alphabetical by last name) in Misc/ACKS.  Thanks!

PR #14701, opened 2019-07-11
bpo-37556 Extend help to include latest overrides

GadgetSteve:
Modify the help in cpython/PC/launcher.c to show users that "latest" can be overridden by shebang, PY_PYTHON[n] or py.ini files. Also show that script [args] is optional by enclosing in square brackets.

https://bugs.python.org/issue37556

### DIFF
--- a/PC/launcher.c
+++ b/PC/launcher.c
@@ -1415,21 +1415,21 @@ usage:\n\
 %ls [launcher-args] [python-args] [script [script-args]]\n\n", argv[0]);
     fputws(L"\
 Launcher arguments:\n\n\
--2     : Launch the latest* Python 2.x version\n\
--3     : Launch the latest* Python 3.x version\n\
+-2     : Launch the latest Python 2.x version\n\
+-3     : Launch the latest Python 3.x version\n\
 -X.Y   : Launch the specified Python version\n", stdout);
     if (canDo64bit) {
         fputws(L"\
      The above all default to 64 bit if a matching 64 bit python is present*.\n\
 -X.Y-32: Launch the specified 32bit Python version\n\
--X-32  : Launch the latest* 32bit Python X version\n\
+-X-32  : Launch the latest 32bit Python X version\n\
 -X.Y-64: Launch the specified 64bit Python version\n\
--X-64  : Launch the latest* 64bit Python X version", stdout);
+-X-64  : Launch the latest 64bit Python X version", stdout);
     }
     fputws(L"\n-0  --list       : List the available pythons", stdout);
     fputws(L"\n-0p --list-paths : List with paths", stdout);
     fputws(L"\n\n If no script is specified the specified interpreter is opened.", stdout);
-    fputws(L"\n*If an exact version is not given latest can be overriden by:", stdout);
+    fputws(L"\nIf an exact version is not given latest can be overriden by:", stdout);
     fputws(L"\n An active virtual environment", stdout);
     fputws(L"\n A shebang line in the script (if present)", stdout);
     fputws(L"\n Matching PY_PYTHON[n] Enviroment variable(s)", stdout);
--- a/PC/launcher.c
+++ b/PC/launcher.c
@@ -1433,7 +1433,7 @@ Launcher arguments:\n\n\
     fputws(L"\nany of the following, (in priority order):", stdout);
     fputws(L"\n An active virtual environment", stdout);
     fputws(L"\n A shebang line in the script (if present)", stdout);
-    fputws(L"\n Matching PY_PYTHON2 or 3 Enviroment variable", stdout);
+    fputws(L"\n With -2 or -3 flag a matching PY_PYTHON2 or PY_PYTHON3 Enviroment variable", stdout);
     fputws(L"\n A PY_PYTHON Enviroment variable", stdout);
     fputws(L"\n From [defaults] in py.ini in your %LOCALAPPDATA%\\py.ini", stdout);
     fputws(L"\n From [defaults] in py.ini beside py.exe (use `where py` to locate)", stdout);
--- a/PC/launcher.c
+++ b/PC/launcher.c
@@ -1420,7 +1420,7 @@ Launcher arguments:\n\n\
 -X.Y   : Launch the specified Python version\n", stdout);
     if (canDo64bit) {
         fputws(L"\
-     The above all default to 64 bit if a matching 64 bit python is present*.\n\
+     The above all default to 64 bit if a matching 64 bit python is present.\n\
 -X.Y-32: Launch the specified 32bit Python version\n\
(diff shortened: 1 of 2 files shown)


PC/launcher.c
```diff
@@ -1412,22 +1412,29 @@ show_help_text(wchar_t ** argv)
...
     fwprintf(stdout, L"\
 usage:\n\
-%ls [launcher-args] [python-args] script [script-args]\n\n", argv[0]);
+%ls [launcher-args] [python-args] [script [script-args]]\n\n", argv[0]);
     fputws(L"\
 Launcher arguments:\n\n\
--2     : Launch the latest Python 2.x version\n\
--3     : Launch the latest Python 3.x version\n\
+-2     : Launch the latest* Python 2.x version\n\
+-3     : Launch the latest* Python 3.x version\n\
```
zooba: I think we can do without the footnote marker, as the comment below applies to practically all possibilities (and the text itself explains which cases are not covered).

PC/launcher.c
```diff
@@ -1412,22 +1412,29 @@ show_help_text(wchar_t ** argv)
...
--X-32  : Launch the latest 32bit Python X version\n\
+-X-32  : Launch the latest* 32bit Python X version\n\
 -X.Y-64: Launch the specified 64bit Python version\n\
--X-64  : Launch the latest 64bit Python X version", stdout);
+-X-64  : Launch the latest* 64bit Python X version", stdout);
     }
     fputws(L"\n-0  --list       : List the available pythons", stdout);
     fputws(L"\n-0p --list-paths : List with paths", stdout);
+    fputws(L"\n\n If no script is specified the specified interpreter is opened.", stdout);
+    fputws(L"\n*If an exact version is not given latest can be overriden by:", stdout);
```
eryksun: This line is missing a comma and "overridden" is misspelled. I suggest the following: "If an exact version is not given, using the latest version can be overridden by any of the following:".

PC/launcher.c
GadgetSteve: Addressed!

PC/launcher.c
```diff
@@ -1412,22 +1412,29 @@ show_help_text(wchar_t ** argv)
...
+    fputws(L"\n An active virtual environment", stdout);
+    fputws(L"\n A shebang line in the script (if present)", stdout);
+    fputws(L"\n Matching PY_PYTHON[n] Enviroment variable(s)", stdout);
```
eryksun: Do you think it's clear that `PY_PYTHON` sets the default and `PY_PYTHON2` and `PY_PYTHON3` are for the -2 and -3 options?

PC/launcher.c
```diff
@@ -1412,22 +1412,29 @@ show_help_text(wchar_t ** argv)
...
+    fputws(L"\n From [defaults] in py.ini in your %LOCALAPPDATA%\\py.ini", stdout);
+    fputws(L"\n From [defaults] in py.ini beside py.exe (use where py to locate)", stdout);
```
eryksun: Put \`where py\` in backticks to clarify the command here.

PC/launcher.c
```diff
@@ -1412,22 +1412,29 @@ show_help_text(wchar_t ** argv)
...
+     The above all default to 64 bit if a matching 64 bit python is present*.\n\
 -X.Y-32: Launch the specified 32bit Python version\n\
 -X-32  : Launch the latest 32bit Python X version\n\
 -X.Y-64: Launch the specified 64bit Python version\n\
 -X-64  : Launch the latest 64bit Python X version", stdout);
     }
     fputws(L"\n-0  --list       : List the available pythons", stdout);
     fputws(L"\n-0p --list-paths : List with paths", stdout);
+    fputws(L"\n\n If no script is specified the specified interpreter is opened.", stdout);
+    fputws(L"\nIf an exact version is not given latest can be overriden by:", stdout);
```
eryksun: This line is missing a comma and "overridden" is misspelled. I suggest the following: "If an exact version is not given, using the latest version can be overridden by any of the following:".

PC/launcher.c
GadgetSteve: Corrected and added that the overrides are in priority order.

PC/launcher.c
GadgetSteve: Reworded to make clearer

PC/launcher.c
GadgetSteve: Done

PC/launcher.c
```diff
@@ -1412,22 +1412,31 @@ show_help_text(wchar_t ** argv)
...
 -X-64  : Launch the latest 64bit Python X version", stdout);
     }
     fputws(L"\n-0  --list       : List the available pythons", stdout);
     fputws(L"\n-0p --list-paths : List with paths", stdout);
+    fputws(L"\n\n If no script is specified the specified interpreter is opened.", stdout);
+    fputws(L"\nIf an exact version is not given, using the latest version can be overridden by", stdout);
+    fputws(L"\nany of the following, (in priority order):", stdout);
+    fputws(L"\n An active virtual environment", stdout);
+    fputws(L"\n A shebang line in the script (if present)", stdout);
+    fputws(L"\n Matching PY_PYTHON2 or 3 Enviroment variable", stdout);
```
eryksun: Without reading the docs, I'd be inclined to think that `PY_PYTHON3` should override the latest Python 3 selection, but it only applies with the `-3` command-line option. The same for `PY_PYTHON2`. So maybe change this to 

> "A PY_PYTHON2 or PY_PYTHON3 environment variable in combination with the -2 or -3 argument, respectively."

GadgetSteve: Modified to:
 `With -2 or -3 flag a matching PY_PYTHON2 or PY_PYTHON3 Enviroment variable`

_(Shorter than suggested but clearer than before)_.

PC/launcher.c
```diff
@@ -1412,22 +1412,31 @@ show_help_text(wchar_t ** argv)
...
+%ls [launcher-args] [python-args] [script [script-args]]\n\n", argv[0]);
     fputws(L"\
 Launcher arguments:\n\n\
 -2     : Launch the latest Python 2.x version\n\
 -3     : Launch the latest Python 3.x version\n\
 -X.Y   : Launch the specified Python version\n", stdout);
     if (canDo64bit) {
         fputws(L"\
-     The above all default to 64 bit if a matching 64 bit python is present.\n\
+     The above all default to 64 bit if a matching 64 bit python is present*.\n\
```
zooba:
```suggestion
     The above all default to 64 bit if a matching 64 bit python is present.\n\
```

Missed one

GadgetSteve: Darn! Fixed.

zooba (approved):

PR #14702, opened 2019-07-11
bpo-36044: Reduce number of unit tests run for PGO build.

nascheme:
This speeds up the PGO profile generation by a factor of about 15X.
Running the full unit test suite is slow.  This may result in a
slightly less optimized build since not as many code branches will
be executed.  If you are willing to wait for the slower build, the
old behavior can be enabled as follows:

    ./configure [...] --with-profile-task='-m test --pgo'

https://bugs.python.org/issue36044


Lib/test/libregrtest/cmdline.py
```diff
@@ -264,7 +264,9 @@ def _create_parser():
                        help='only write the name of test cases that will be run'
                             ' , don\'t execute them')
     group.add_argument('-P', '--pgo', dest='pgo', action='store_true',
-                       help='enable Profile Guided Optimization training')
+                       help='enable Profile Guided Optimization (PGO) training')
+    group.add_argument('--pgo-extended', action='store_true',
```
vstinner, line 268: I don't understand what "extended" means. I suggest to rename the option to "--pgo-all-tests" and rewrite the help to explain that it runs all tests by default.

Lib/test/libregrtest/main.py
```diff
@@ -21,6 +21,55 @@
 from test import support
 
 
+# Set of tests run by default if --pgo is specified.  The tests below were
+# chosen based on the following criteria: either they exercise a commonly used
+# C extension module or type, or they run some relatively typical Python code.
+# Long running tests should be avoided because the PGO instrumented executable
+# runs slowly.
```
vstinner: I would prefer to keep main.py small. Would you mind to create libregrtest/pgo.py for this list please?

Lib/test/libregrtest/cmdline.py
nascheme: Think list.extend().  I.e. run more tests.  I didn't want to use --pgo-all because we don't actually run all the tests.  Some are excluded using the `support.PGO` flag.  However, I don't feel strongly so if you really like --pgo-all, I will change it.

Lib/test/libregrtest/main.py
nascheme: Ok.

Lib/test/libregrtest/cmdline.py
nascheme: Hmm, --pgo-harder?  :smile:

Lib/test/libregrtest/cmdline.py
```diff
@@ -264,7 +264,9 @@ def _create_parser():
...
+                       help='enable extended PGO training (slower training)')
```
gpshead, line 269: I'd add to the description of this that "User are encouraged to benchmark the interpreter resulting from a PROFILE_TASK with this flag to decide if it is meaningfully faster than those produced using --pgo for their use cases".

I don't care what the flag name is, --pgo-extended is fine by me.  as is --pgo-very-long-training-run (a flag who's length matches the build time seems... appropriate). ;)

gpshead (approved):
i updated the news entry wording a bit.

methane (approved):

nascheme:
This speeds up the task by a factor of about 15x.  Running the full unit test
suite is slow.  This change may result in a slightly less optimized build
since not as many code branches will be executed.  If you are willing to
wait for the much slower build, the old behavior can be restored using
'./configure [..] PROFILE_TASK="-m test --pgo-extended"'.  We make no
guarantees as to which PGO task set produces a faster build.  Users who
care should run their own relevant benchmarks as results can depend on
the environment, workload, and compiler tool chain.

ZackerySpytz:
@nascheme It seems that only 40 tests are run on Travis after this change (for example, https://travis-ci.org/python/cpython/jobs/562362678) instead of the entire test suite. Please take a look at this issue.

gpshead:
https://github.com/python/cpython/pull/14927

PR #14703, opened 2019-07-11
bpo-37558: Shared memory tests are failing due to double slashes

kulikjak:
With the addition of shared memory into Python 3.8, we now have three tests failing on Solaris, namely `test_multiprocessing_fork`, `test_multiprocessing_forkserver` and `test_multiprocessing_spawn`. The reason seems to be incorrect name handling which results in two slashes being prepended.

https://bugs.python.org/issue37558


pitrou (approved):
+1. Thanks @kulikjak !

PR #14709, opened 2019-07-11
bpo-37564: ArgumentParser support bool type according to truth values

zachben:
https://bugs.python.org/issue37564

### DIFF
--- a/Lib/argparse.py
+++ b/Lib/argparse.py
@@ -89,7 +89,6 @@ import shutil as _shutil
 import sys as _sys
 
 from gettext import gettext as _, ngettext
-from distutils.util import strtobool as _strtobool
 
 SUPPRESS = '==SUPPRESS=='
 
@@ -1603,6 +1602,22 @@ class _MutuallyExclusiveGroup(_ArgumentGroup):
         self._container._remove_action(action)
         self._group_actions.remove(action)
 
+def _strtobool (val):
+    """Convert a string representation of truth to true or false.
+
+    True values are 'y', 'yes', 't', 'true', 'on', and '1'; false values
+    are 'n', 'no', 'f', 'false', 'off', and '0'.  Raises ValueError if
+    'val' is anything else.
+
+    <Based on strtobool located in module distutils.util>
+    """
+    val = val.lower()
+    if val in ('y', 'yes', 't', 'true', 'on', '1'):
+        return True
+    elif val in ('n', 'no', 'f', 'false', 'off', '0'):
+        return False
+    else:
+        raise ValueError("invalid truth value %r" % (val,))
 
 class ArgumentParser(_AttributeHolder, _ActionsContainer):
     """Object for parsing command line strings into Python objects.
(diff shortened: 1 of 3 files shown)


the-knights-who-say-ni:

mangrisano (approved):
LGTM. Thank you for providing test :)

Lib/argparse.py
```diff
@@ -89,6 +89,7 @@
 import sys as _sys
 
 from gettext import gettext as _, ngettext
+from distutils.util import strtobool as _strtobool
```
zooba: Let's not take a dependency on distutils here - copy the function in (and add the forced `bool` conversion if necessary).

zachben: Done.
And instead of forcing the cast to bool, the function returns True/False of bool type.

hpaulj:
`bool` already exists as a Python function.  It should not be used as a registry key.  This proposal should not shadow a builtin function without clearly documenting it.  It should be clear to users (even novices) that they are using a documented function, 'strtobool' that converts some strings to boolean.  We should not use a slight-of-hand just to make input of 'True/False' more convenient.

zachben:
> `bool` already exists as a Python function. It should not be used as a registry key. This proposal should not shadow a builtin function without clearly documenting it. It should be clear to users (even novices) that they are using a documented function, 'strtobool' that converts some strings to boolean. We should not use a slight-of-hand just to make input of 'True/False' more convenient.

Hi Paul,
I appreciate the points of concern that you have raised, which are indeed solid and with undoubtful seniority, and I would like to express my (and many others) point of view.
As a matter of fact, I wanted to reply in the bpo-37564 issue but couldn't figure out how to do that, so I will respond here.
It is clear and true that the bool function is intended to work as described in the documentation you referred to, and as you have mentioned in the [StackOverflow thread](https://stackoverflow.com/a/19233287/901925).
But this has nothing to do with the notion of defining CLI that *should be interacted by a user*, and as such, it should provide a simple, logical, and intuitive UI. The Python developer is using ArgumentParser to implement such CLI, and he/she deserves an intuitive interface, to reflect the expected user experience. We could not expect millions of developers to duplicate code that wraps this utility, just to gain this simple behavior.

Telling this developer that 'type' parameter is anything else then the object-type of the retrieved argument value from the user is not intuitive, and misleading. 
We should be bold enough to admit historical mistakes, and fix them for the sake of millions of developers that spending precious time on StackOverflow, where they could have been avoid it.
The documentation for parameter 'type' in the __init__ method of class Action, proves that from the first place bool type has been neglected to be not useful, and it's a shame.

It is agreed that the 'type' parameter is supposed to be callable. This is to allow a user to provide custom conversion functions for his/her special needs. But when it comes to primitive types such as int, str, float, and bool - the 'type' parameter should simply define the type of which the value will be received from the (humane) user.

No humane user will think to provide an empty string to express Falae, or to provide the string 'False' in order to express True.
And no developer will think to configure type=bool and at the same time to expect the user to behave as described above.

Regarding the qualms that you have raised about the proposed solution:
- The 'store_true' and 'store_false' actions cannot replace a simplified true/false parameter, that in some cases it's value needed to be generated by a wrapping script.
- I removed the 'bool' wrapper by implementing strtobool locally with return type adjustment.
- I removed the extra import also by implementing strtobool locally.
- This type should definitely be registered as object bool and not as string 'bool' from the reasons explained above.

I also agree that this solution might break a consistent behavior, although I don't see any reason why someone will have in his/her legacy code type=bool, while it is obviously not serving their needs. For that reason, I suggest making it "half-backward-compatible", by keeping None and empty string as False values.
Your observation is super critical here, so please let me know what do you think now, after reading this, fairly long :), explanation.

PR #14710, opened 2019-07-11
bpo-27379: In order to keep backward compatibility, update SocketType's type

shihai1991:
https://bugs.python.org/issue27379


shihai1991:
@vadmium Hi, Martin. What's your opinion about this bugfix?

shihai1991:
cc @vadmium

PR #14716, opened 2019-07-11
bpo-37553: SendfileUsingSendTest tests timeout too short for Windows …

paulmon:
The timeout error reproduces very infrequently when test_socket tests are run manually on a Raspberry Pi 3 running Windows IoT Core.  I was able to determine that this timeout is the one causing the failure by making it smaller (0.01 seconds) which caused the tests in question to fail every time.  

I would like to increase the timeout only if the test machine is running Windows ARM32 (aka ARM).

@zooba

https://bugs.python.org/issue37553

### DIFF
--- a/Lib/test/test_socket.py
+++ b/Lib/test/test_socket.py
@@ -5769,7 +5769,8 @@ class SendfileUsingSendTest(ThreadedTCPSocketTest):
     FILESIZE = (10 * 1024 * 1024)  # 10 MiB
     BUFSIZE = 8192
     FILEDATA = b""
-    TIMEOUT = 2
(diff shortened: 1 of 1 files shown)


zooba:
I'll merge this, but if it doesn't help the buildbot then let's revert it @paulmon

zooba:
If it _does_ help, then we can backport to 3.8. But the buildbot is running against 3.x only now.

paulmon:
The first pass with this fix passed the test_socket tests.  We need more test passes to see if it's still randomly failing. I'll keep an eye on it.

paulmon:
It looks like increasing this timeout works.  The socket tests have all passed on the Windows ARM32 buildbot workers since making this change.

PR #14719, opened 2019-07-12
Typo fix in file Python/getopt.c

hansrajdas:
Typo fix in file Python/getopt.c
* improvment should be improvement
* Extra whitespece removed


mangrisano (approved):
Thank you.

benjaminp:
We should probably just delete this paragraph. It seems unlikely that David Gottner would want to hear about Python's hacked-up copy of his code.

hansrajdas:
@benjaminp and @mangrisano - thanks for your time, so shall i delete these 3 lines?

hansrajdas:
> We should probably just delete this paragraph. It seems unlikely that David Gottner would want to hear about Python's hacked-up copy of his code.

@benjaminp I have deleted the requested paragraph. Please take a look. Thank you!

PR #14721, opened 2019-07-12
bpo-37571: Add 'b' to prevent the TypeError exception.

mangrisano:
In Python3 is not be able to assign a _str_ to a _c_char_p_, so the 'b' is essential to prevent the TypeError exception.

https://bugs.python.org/issue37571

### DIFF
--- a/Doc/library/ctypes.rst
+++ b/Doc/library/ctypes.rst
@@ -1178,9 +1178,9 @@ the root-object's underlying buffer.
 Another example that may behave different from what one would expect is this::
 
    >>> s = c_char_p()
-   >>> s.value = "abc def ghi"
+   >>> s.value = b"abc def ghi"
    >>> s.value
-   'abc def ghi'
+   b'abc def ghi'
    >>> s.value is s.value
    False
    >>>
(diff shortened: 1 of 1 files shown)


aeros (approved):
I confirmed that the `TypeError` occurs when attempting to assign a string to the value of an object instantiated from `c_char_p`:

image

Doc/library/ctypes.rst
```diff
@@ -1178,9 +1178,9 @@ the root-object's underlying buffer.
...
 
    >>> s = c_char_p()
-   >>> s.value = "abc def ghi"
+   >>> s.value = b"abc def ghi"
    >>> s.value
-   'abc def ghi'
+   b'abc def ghi'
    >>> s.value is s.value
    False
    >>>
```
aeros: I thought of this suggestion after approving the PR. It's not entirely necessary, but it would be helpful.

In order to help users not make the same mistake that the docs initially did, I would suggest adding a note after the code example explaining that objects created from the `c_char_p` class can only have their value assigned to bytes or integers.
```suggestion
    >>>

.. note::

   Objects instantiated from :class:`c_char_p` can only have their value set to bytes
   or integers.
```

mangrisano:
You're right. Thanks a lot for reviewing this.

PR #14723, opened 2019-07-12
bpo-37521: No longer treat insertion into sys.modules as optional in importlib examples

brettcannon:
Fix importlib examples to insert any newly created modules via importlib.util.module_from_spec() immediately into sys.modules instead of after calling loader.exec_module().

Thanks to Benjamin Mintz for finding the bug.

https://bugs.python.org/issue37521


ericsnowcurrently (approved):
LGTM

PR #14729, opened 2019-07-13
bpo-37583: Update support.get_socket_conn_refused_errs()

shihai1991:
https://bugs.python.org/issue37583

### DIFF
--- a/Lib/test/support/__init__.py
+++ b/Lib/test/support/__init__.py
@@ -1491,6 +1491,10 @@ def get_socket_conn_refused_errs():
         # bpo-31910: socket.create_connection() fails randomly
         # with EADDRNOTAVAIL on Travis CI
         errors.append(errno.EADDRNOTAVAIL)
+    if hasattr(errno, 'EHOSTUNREACH'):
+        # bpo-37583: socket.create_connection() fails always
+        # with EHOSTUNREACH on Virtual Machine
+        errors.append(errno.EHOSTUNREACH)
     if not IPV6_ENABLED:
         errors.append(errno.EAFNOSUPPORT)
     return errors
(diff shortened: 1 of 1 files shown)


vstinner:
I'm not sure that it's a good idea to catch EHOSTUNREACH in get_socket_conn_refused_errs(). It seems to be a very specific setup ("Looks it's a problem of security group." says the issue). I would prefer to catch the error somewhere else.

Lib/test/support/__init__.py
```diff
@@ -1491,6 +1491,10 @@ def get_socket_conn_refused_errs():
         # bpo-31910: socket.create_connection() fails randomly
         # with EADDRNOTAVAIL on Travis CI
         errors.append(errno.EADDRNOTAVAIL)
+    if hasattr(errno, 'EHOSTUNREACH'):
+        # bpo-37583: socket.create_connection() fails always
+        # with EHOSTUNREACH on Virtual Machine
```
vstinner: This comment is wrong. The issue is specific to your VM config and it's unrelated to VMs in general. I suggest this comment instead:

```
# bpo-37583: The destination host cannot be reached
```

In the issue you wrote "Looks it's a problem of security group." maybe you can investigate to write a more accurate comment, but it's not required ;-)

shihai1991: Thanks, victor. Your comment is good enough:)

vstinner:
@shihai1991: It seems like get_socket_conn_refused_errs() it outdated in Python 3.7. Can you try to create a PR to copy the code from master to the 3.7 branch?

shihai1991:
> @shihai1991: It seems like get_socket_conn_refused_errs() it outdated in Python 3.7. Can you try to create a PR to copy the code from master to the 3.7 branch?

of course ;) Adding a PR-->https://github.com/python/cpython/pull/15259

PR #14736, opened 2019-07-13
bpo-18378: Recognize "UTF-8" as a valid name in locale._parse_localename

ronaldoussoren:
"UTF-8" is a valid value for LC_CTYPE on macOS, and the variable is set to that value during interpreter startup by the UTF-8 coercion code in pylifecycle.c. 

https://bugs.python.org/issue18378

### DIFF
--- a/Lib/locale.py
+++ b/Lib/locale.py
@@ -492,6 +492,10 @@ def _parse_localename(localename):
         return tuple(code.split('.')[:2])
     elif code == 'C':
         return None, None
+    elif code == 'UTF-8':
+        # On macOS "LC_CTYPE=UTF-8" is a valid locale setting
+        # for getting UTF-8 handling for text.
+        return None, 'UTF-8'
     raise ValueError('unknown locale: %s' % localename)
 
 def _build_localename(localetuple):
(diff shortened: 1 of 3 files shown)


ronaldoussoren:
The pull-request enables recognition of "UTF-8" as a valid value on all platforms. I've primarily done this because the value of LC_CTYPE will usually leak into the target system when you use SSH to log on those systems.

ronaldoussoren:
Victor, I've asked for your review because you're involved with the code that sets LC_CTYPE during interpreter startup.

vstinner:
locale.getpreferredencoding() is based on nl_langinfo(CODESET) which is reliable.

I never understood the purpose of the locale.getdefaultlocale() function, especially locale._parse_localename(). What is the use case for locale.getdefaultlocale()?

ronaldoussoren:
>What is the use case for locale.getdefaultlocale()?

I don't know what the use case is, but this is a public API that's currently broken on macOS. I've run into this issue in the past when using software that uses this API, but to be  honest I don't recall which software that was. The BPO issue mentions (older versions of?) docutils that use this API.

ned-deily (approved):
LGTM

PR #14738, opened 2019-07-13
bpo-35476: clear error in _imp_create_dynamic_impl

isidentical:
clear error in _imp_create_dynamic_impl

https://bugs.python.org/issue35476

### DIFF
--- a/Python/import.c
+++ b/Python/import.c
@@ -2249,6 +2249,9 @@ _imp_create_dynamic_impl(PyObject *module, PyObject *spec, PyObject *file)
 
     mod = _PyImport_FindExtensionObject(name, path);
     if (mod != NULL || PyErr_Occurred()) {
+        if (PyErr_Occurred()) {
+            PyErr_Clear();
+        }
         Py_DECREF(name);
         Py_DECREF(path);
         Py_XINCREF(mod);


Python/import.c
```diff
@@ -2249,6 +2249,9 @@ _imp_create_dynamic_impl(PyObject *module, PyObject *spec, PyObject *file)
 
     mod = _PyImport_FindExtensionObject(name, path);
     if (mod != NULL || PyErr_Occurred()) {
+        if (PyErr_Occurred()) {
```
brettcannon, line 2252: To avoid the overhead of calling `PyErr_Occurred()`, could you store the result in a variable and check that?

Python/import.c
```diff
@@ -2249,6 +2249,9 @@ _imp_create_dynamic_impl(PyObject *module, PyObject *spec, PyObject *file)
...
+            PyErr_Clear();
+        }
```
brettcannon, line 2254: This actually won't work because since you cleared out the error but are returning `NULL` you will trigger a failure in the interpreter for saying there's an exception when one isn't set.

brettcannon (changes requested):
I don't think this is quite doing what @ericsnowcurrently meant when he said "the error should be cleared" as an option in the issue. Instead, I believe what he was thinking was that there be an `elif (PyErr_Occurred())` clause to go after the `if (mod != NULL)` which then clears the exception.

brettcannon:
I'm going to kick off a discussion over on the issue which I think may change the approach this PR takes.

Python/import.c
```diff
@@ -2249,6 +2249,9 @@ _imp_create_dynamic_impl(PyObject *module, PyObject *spec, PyObject *file)
 
     mod = _PyImport_FindExtensionObject(name, path);
     if (mod != NULL || PyErr_Occurred()) {
+        if (PyErr_Occurred()) {
+            PyErr_Clear();
```
brettcannon, line 2253: I've started a discussion over on the issue where if I'm right and swallowing an exception is bad then this clearing of the exception should be removed.

brettcannon:
@isidentical please note the conversation on the issue is advocating for explicit exception raising, so that changes the approach you've taken here.

brettcannon:
CLosing as @ericsnowcurrently pointed out the current semantics cover the bug.

PR #14747, opened 2019-07-13
bpo-30588: document codecs.escape_decode

carlbordum:
I gave it a shot :)

https://bugs.python.org/issue30588

### DIFF
--- a/Doc/library/codecs.rst
+++ b/Doc/library/codecs.rst
@@ -242,6 +242,13 @@ wider range of codecs when working with binary files:
    :func:`iterencode`.
 
 
+.. function:: escape_decode(data, errors=None)
+
+   Decode the bytes-like object *data* and return a tuple (decoded object,
+   length consumed). This is useful for decoding ascii escape sequences mixed
+   with unicode characters.
+
+
 The module also provides the following constants which are useful for reading
 and writing to platform dependent files:
 
@@ -1313,7 +1320,10 @@ encodings.
 |                    |         | Latin-1 source code.      |
 |                    |         | Beware that Python source |
 |                    |         | code actually uses UTF-8  |
-|                    |         | by default.               |
+|                    |         | by default. This does not |
+|                    |         | work in the general case, |
+|                    |         | see:                      |
+|                    |         | :func:`escape_decode`.    |
 +--------------------+---------+---------------------------+
 
 .. versionchanged:: 3.8


zooba:
I believe this is okay, but if @serhiy-storchaka and @asvetlov want to block it (as discussed in the bug), happy to let them overrule me :)

doerwalter:
IMHO, **if we document this function**, the added description shouldn't describe what a generic `*_decode` function does, it should decribe what `escape_decode` specifically does. I.e as it is now the first sentence is redundant, the second too vague.

Doc/library/codecs.rst
```diff
@@ -242,6 +242,13 @@ wider range of codecs when working with binary files:
    :func:`iterencode`.
 
 
+.. function:: escape_decode(data, errors=None)
+
+   Decode the bytes-like object *data* and return a tuple (decoded object,
+   length consumed). This is useful for decoding ascii escape sequences mixed
+   with unicode characters.
```
gpshead, line 249: What is "mixed with unicode characters" supposed mean in this context?  data is a bytes-like object, it can't contain unicode runes.  We should include an example of what this does that is different than using one of the text encoding codecs.

csabella:
@carlbordum please address the review comments.  Thanks!

Doc/library/codecs.rst
carlbordum: As Matthieu Dartiailh described on bugs.python.org, an example of ascii decode characters mixed with unicode is `'Δ\nΔ'`. 

Here is the difference:

``` python
>>> codecs.unicode_escape_decode('Δ\nΔ')
('Î\x94\nÎ\x94', 5)
>>> codecs.escape_decode('Δ\nΔ')
(b'\xce\x94\n\xce\x94', 5)
>>> codecs.escape_decode('Δ\nΔ')[0].decode('utf-8')
'Δ\nΔ'
```

noyannus: A [real-world example](https://stackoverflow.com/questions/70457699/decode-individual-octal-characters-in-string-variable). We can assume that many more homegrown 'solutions' exist. 

Wouldn't it be great if kludgy, slow, error-prone workarounds people have come up with were replaced with something elegant and Python-worthy?

Please consider that this function is so rarely seen outside the Python developer world *because it is kept almost a secret.*

Doc/library/codecs.rst
```diff
@@ -242,6 +242,13 @@ wider range of codecs when working with binary files:
    :func:`iterencode`.
 
 
+.. function:: escape_decode(data, errors=None)
+
+   Decode the bytes-like object *data* and return a tuple (decoded object,
```
JelleZijlstra, line 247: "bytes-like object" seems incorrect; it accepts strings too.

JelleZijlstra:
Closing as it's been a few years and the feedback hasn't been addressed.

PR #4293, opened 2017-11-06
Replace KB unit with KiB

vstinner:
kB (*kilo* byte) unit means 1000 bytes, whereas KiB ("kibibyte")
means 1024 bytes. KB was misused: replace kB or KB with KiB when
appropriate.

Same change for MB and GB which become MiB and GiB.

Change the output of Tools/iobench/iobench.py.

Round also the size of the documentation from 5.5 MB to 5 MiB.

PR #14748, opened 2019-07-13
bpo-18049: Sync thread stack size to main thread size on macOS

ronaldoussoren:
This changeset increases the size of the stack
for threads on macOS to the size of the stack
of the main thread and reenables the relevant
recursion test.

### DIFF
--- a/Misc/NEWS.d/next/macOS/2019-07-13-15-58-18.bpo-18049.MklhQQ.rst
+++ b/Misc/NEWS.d/next/macOS/2019-07-13-15-58-18.bpo-18049.MklhQQ.rst
@@ -1,0 +1,2 @@
+Increase stack size of threads to 16MB on macOS, to match the stack size of
+the main thread. This avoids crashes on deep recursion in threads.
--- a/Misc/NEWS.d/next/macOS/2019-07-13-15-58-18.bpo-18049.MklhQQ.rst
+++ b/Misc/NEWS.d/next/macOS/2019-07-13-15-58-18.bpo-18049.MklhQQ.rst
@@ -1,2 +1,3 @@
-Increase stack size of threads to 16MB on macOS, to match the stack size of
-the main thread. This avoids crashes on deep recursion in threads.
+Increase the default stack size of threads from 5MB to 16MB on macOS, to
+match the stack size of the main thread. This avoids crashes on deep recursion
+in threads.
--- a/Lib/test/test_threading.py
+++ b/Lib/test/test_threading.py
@@ -1057,8 +1057,6 @@ class ThreadingExceptionTests(BaseTestCase):
         lock = threading.Lock()
         self.assertRaises(RuntimeError, lock.release)
 
-    @unittest.skipUnless(sys.platform == 'darwin' and test.support.python_is_optimized(),
-                         'test macosx problem')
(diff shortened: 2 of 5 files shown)


ronaldoussoren:
Is this something that can be back ported to 3.8 as it is still in beta?

Misc/NEWS.d/next/macOS/2019-07-13-15-58-18.bpo-18049.MklhQQ.rst
```diff
@@ -0,0 +1,2 @@
+Increase stack size of threads to 16MB on macOS, to match the stack size of
```
vstinner: Would you mind to mention the old value, 5 MiB?

You may specify that it's the *default* size: it can be overriden by _thread.stack_size(size).

vstinner (approved):
LGTM.

vstinner:
> Is this something that can be back ported to 3.8 as it is still in beta?

For me it's a bugfix and so can be backported.

ned-deily (approved):
LGTM. Sure, I think backporting to 3.8 is fine.  I will run more tests on older versions of macOS before the next beta.

Misc/NEWS.d/next/macOS/2019-07-13-15-58-18.bpo-18049.MklhQQ.rst
```diff
@@ -0,0 +1,3 @@
+Increase the default stack size of threads from 5MB to 16MB on macOS, to
```
vstinner, line 1: nitpick:

```suggestion
Increase the default stack size of threads from 5 MiB to 16 MiB on macOS, to
```

vstinner (approved):
LGTM. I just have a minor comment.

Misc/NEWS.d/next/macOS/2019-07-13-15-58-18.bpo-18049.MklhQQ.rst
ronaldoussoren: Is MiB something we use elsewhere? 

I'm probably old, but don't particularly like the MiB vs. MB distinction as it is often clear from the context what's meant.

vstinner:
> Is MiB something we use elsewhere? I'm probably old, but don't particularly like the MiB vs. MB distinction as it is often clear from the context what's meant.


Well, it's not a big deal, just a remark. I already approved your PR twice ;-) It's up to you.

---

For MiB, see my PR #4293: "kB (kilo byte) unit means 1000 bytes, whereas KiB ("kibibyte")
means 1024 bytes. KB was misused: replace kB or KB with KiB when
appropriate."

And my commit 8c663fd60ecba9c82aa4c404dbfb1aae69fe8553:

```
commit 8c663fd60ecba9c82aa4c404dbfb1aae69fe8553
Date:   Wed Nov 8 14:44:44 2017 -0800

    Replace KB unit with KiB (#4293)
```

5 MB means 5 000 000 bytes, not `5*1024*1024` = 5242880 bytes.

Sadly MB (megabyte) is used instead of MiB in many places (outside Python).

`#define THREAD_STACK_SIZE       0x1000000` is 16 MiB ;-)

https://en.wikipedia.org/wiki/Megabyte

PR #14751, opened 2019-07-13
bpo-36261: email example was misusing preamble field

carlbordum:
https://bugs.python.org/issue36261

### DIFF
--- a/Doc/includes/email-mime.py
+++ b/Doc/includes/email-mime.py
@@ -14,7 +14,7 @@ msg['Subject'] = 'Our family reunion'
 # family = the list of all recipients' email addresses
 msg['From'] = me
 msg['To'] = ', '.join(family)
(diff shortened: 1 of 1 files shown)


mangrisano (approved):
LGTM.

aeros (approved):
Thanks for the contribution. This correctly addresses the issue described and for consistency, using the same preamble found in `email-dir.py`. 

However, since this format is exactly the same, I wonder if it would better to create a pre-configured setup to use as part of the `EmailMessage` class (as a private method) instead of replicating these 5 lines in different locations:
```python
msg = EmailMessage()
msg['Subject'] = 'Contents of directory %s' % os.path.abspath(directory)
msg['To'] = ', '.join(args.recipients)
msg['From'] = args.sender
msg.preamble = 'You will not see this in a MIME-aware mail reader.\n'
```

PR #14755, opened 2019-07-13
bpo-35409: Ignore GeneratorExit in async_gen_athrow_throw

vxgmichel:
Ignore `GeneratorExit` exceptions when throwing an exception into the `aclose` coroutine of an asynchronous generator.

https://bugs.python.org/issue35409

### DIFF
--- a/Lib/test/test_asyncgen.py
+++ b/Lib/test/test_asyncgen.py
@@ -753,6 +753,33 @@ class AsyncGenAsyncioTest(unittest.TestCase):
         self.loop.run_until_complete(run())
         self.assertEqual(DONE, 10)
 
+    def test_async_gen_asyncio_aclose_12(self):
+        DONE = 0
+
+        async def target():
+            await asyncio.sleep(0.01)
+            1 / 0
+
+        async def foo():
+            nonlocal DONE
+            task = asyncio.create_task(target())
+            try:
+                yield 1
+            finally:
+                try:
+                    await task
+                except ZeroDivisionError:
+                    DONE = 1
+
+        async def run():
+            gen = foo()
+            it = gen.__aiter__()
+            await it.__anext__()
+            await gen.aclose()
+
+        self.loop.run_until_complete(run())
+        self.assertEqual(DONE, 1)
+
     def test_async_gen_asyncio_asend_01(self):
         DONE = 0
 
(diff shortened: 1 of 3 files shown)


vxgmichel:
Corresponding PR on pypy: https://bitbucket.org/pypy/pypy/pull-requests/655

1st1 (approved):

1st1 (commented):
Can you double check if regular generators have the same bahavior before we merge?

vxgmichel:
@1st1 
> Can you double check if regular generators have the same behavior before we merge?

Well, I don't think this behavior applies to regular generators. The closest thing I could think of is:
```python
>>> def gen():
...     try:
...         yield
...     finally:
...         yield
... 
>>> g = gen()
>>> g.send(None)
>>> g.close()
```
which produces:
```python
RuntimeError: generator ignored GeneratorExit
```

Also note that the chunk of code added to `async_gen_athrow_throw` is directly taken from [async_gen_athrow_send](https://github.com/python/cpython/blob/82494aa6d947c4a320c09c58fe0f100cdcf7af0b/Objects/genobject.c#L1880-L1893). A bit of refactoring might be required, although I don't feel comfortable enough to do it myself. A similar refactoring has been done in pypy: see the [handle_error](https://bitbucket.org/pypy/pypy/pull-requests/655/ignore-generatorexit-when-throwing-into/diff#Lpypy/interpreter/generator.pyT777) method which is used by both the `send` and `throw` methods of the `athrow` coroutine.

vxgmichel:
Just pinging, as it turns out to be a blocking issue for the [aiostream](https://github.com/vxgmichel/aiostream) library :)

1st1:
Andrew, can you take a look at this too?

asvetlov:
Is it urgent?
I'll have time next week only after finishing me bisuness trip.

1st1:
I guess since we missed the RC1 for this it's no longer urgent. But please look at this as soon as you can.  I'd like to hear your opinion.

vxgmichel:
Hi all, I'm just pinging again as I would really like to have this issue fixed :)

asvetlov:
LGTM

asvetlov:
Thanks!

PR #14756, opened 2019-07-13
gh-70186: Fix doctest badly handling unwrapable objects

BoboTiG:
[EuroPython 2019]

For some reason, an object could raise an error different from `AttributeError` when checking for its `__wrapped__` attribute.

This patch silences any errors coming from the underlying call to `inpect.unwrap()` and, if `verbose` is set to `True`, prints out the exception.

Thanks to @mpaolini for helping me with the reproduction test case.

---

As of the current patch state, I followed R. David Murray (I do not find his GH nick) sentence from his last comment on the BPO:

> it should be capturing that and probably other errors and reporting them, instead of just producing a traceback, I think.

---

Another solution would be to patch `inspect.unwrap()` like:

```diff
diff --git a/Lib/inspect.py b/Lib/inspect.py
index 99a580bd2f..f359b0558a 100644
--- a/Lib/inspect.py
+++ b/Lib/inspect.py
@@ -507,11 +507,16 @@ def unwrap(func, *, stop=None):
 
     """
     if stop is None:
-        def _is_wrapper(f):
+        def __is_wrapper(f):
             return hasattr(f, '__wrapped__')
     else:
-        def _is_wrapper(f):
+        def __is_wrapper(f):
             return hasattr(f, '__wrapped__') and not stop(f)
+    def _is_wrapper(f):
+        try:
+            return __is_wrapper(f)
+        except Exception:
+            return False
     f = func  # remember the original func for error reporting
     # Memoise by id to tolerate non-hashable objects, but store objects to
     # ensure they aren't destroyed, which would allow their IDs to be reused.
```

But we then loose details and nobody could know that there is an issue somewhere.
I did not want to raise a specific error from there only for doctest (but it is still totally doable).

---

Yet another solution would to raise from doctest instead of printing. Such as:

```python
raise ValueError(f"DocTestFinder.find: __wrapped__ threw {exc!r}: {type(val)!r}")
```

As it is done few lines below when doctest cannot deal with certain objects or certains conditions.
This would keep some uniformization in how doctest handle bad cases.

---

:eyes: I need external eyes to decide on what is the best solution to use :) (I had to leave the sprints before having time to speak with a core dev)

https://bugs.python.org/issue25998

* Issue: gh-70186


aeros167:
cc @bitdancer (R. David Murray), @stevendaprano

aeros (commented):
I'm significantly more in favor of raising an explicit exception message from the doctests. The usage of:

```python
try:
    ...
except Exception:
    return False
```
is generally frowned upon, as far as I'm aware.

aeros167:
Also, thanks for the contribution!

BoboTiG:
@aeros167 Changes done :+1:

aeros (approved):
Looks good as far as I can tell, thanks for making the suggested changes. Since there's no active core dev expert on the doctests in the [expert index](https://devguide.python.org/experts/#miscellaneous) I'll mention some others. Three of them are listed as other test experts and vstinner is experienced with writing tests in general. 

cc @ezio-melotti @voidspace @vstinner @rbtcollins

furkanonder (approved):

BoboTiG:
I finally got time to update the patch, and the CI is still green :tada: 

/cc @aeros @furkanonder

mpaolini:
@BoboTiG shall we land this :) ?

BoboTiG:
> @BoboTiG shall we land this :) ?

It's up to the team, the CI is green :)

PR #14757, opened 2019-07-13
gh-57336: Fix attr name setter behavior

potomak:
Update the implementation of attributes `name` setter to correctly
replace the attribute in the owner element.

The update is divided in three main steps:

1. update the implementation of `Attr._set_name` to behave more
   similarly to `Document.renameNode`

2. update the implementation of `Document.renameNode` to use
   `Attr._set_name` if the node is of type "attribute"

3. create a new function `_parseName`, that is used in both
   `Attr._set_name` and `Document.renameNode`, to reduce code
   duplication

https://bugs.python.org/issue13127

* Issue: gh-57336


serhiy-storchaka:
@potomak, please sign the CLA (just click the button and follow instructions). We need this to be able to merge this PR.

potomak:
@serhiy-storchaka done

PR #14776, opened 2019-07-14
bpo-37592: sysconfig no longer relies on sys.version

carlbordum:
https://bugs.python.org/issue37592


mangrisano (approved):

carlbordum:
I believe this is no longer relevant thanks to https://github.com/python/cpython/pull/24644.

PR #14777, opened 2019-07-14
bpo-34543: Fix SystemErrors and segfaults with uninitialized Structs

ZackerySpytz:
https://bugs.python.org/issue34543

### DIFF
--- a/Lib/test/test_struct.py
+++ b/Lib/test/test_struct.py
@@ -652,6 +652,17 @@ class StructTest(unittest.TestCase):
         s2 = struct.Struct(s.format.encode())
         self.assertEqual(s2.format, s.format)
 
+    def test_uninitialized_struct(self):
+        s = struct.Struct.__new__(struct.Struct)
+        with self.assertRaises(ValueError):
+            s.format
+        with self.assertRaises(ValueError):
+            s.size
+        for meth in s.iter_unpack, s.pack, s.unpack, s.unpack_from:
+            self.assertRaises(ValueError, meth, b'0')
+        self.assertRaises(ValueError, s.pack_into, bytearray(1), 0, b'0')
+        self.assertRaises(ValueError, s.__sizeof__)
+
 
 class UnpackIteratorTest(unittest.TestCase):
     """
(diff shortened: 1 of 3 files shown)


ZackerySpytz:
This patch uses a `CHECK_INITIALIZED()` macro (like what is done in `Modules/_io`), but there are other ways to fix this issue.

aeros (approved):
In order to test the changes, I attempted to recreate the segfault in the most current commit to cpython master and in your remote branch `bpo-34543-struct-crashes`.

cpython master:
image

image

PR branch:
image


For the version used for testing the latest cpython, I used the function `test_segfault()` on the second round in order to perform the test multiple times consecutively. The issue tracker reported similar problems with replication. The first round caused a TypeError, and the second one caused the segfault.

After performing `test_segfault()` 3 times consecutively in the PR's branch, ValueError was raised each time with the same message. As far as I can tell, this resolves the segfault issue and provides a significant improvement by raising a consistent exception each time.

Nicely done @ZackerySpytz, approved.

Lib/test/test_struct.py
```diff
@@ -652,6 +652,17 @@ def test_format_attr(self):
...
+    def test_uninitialized_struct(self):
+        s = struct.Struct.__new__(struct.Struct)
+        with self.assertRaises(ValueError):
+            s.format
+        with self.assertRaises(ValueError):
+            s.size
+        for meth in s.iter_unpack, s.pack, s.unpack, s.unpack_from:
+            self.assertRaises(ValueError, meth, b'0')
+        self.assertRaises(ValueError, s.pack_into, bytearray(1), 0, b'0')
+        self.assertRaises(ValueError, s.__sizeof__)
```
aeros, line 664: Also, realized this after submitting the approval. This is quite minor, and I still approve of the PR either way. Instead of using `s.__sizeof__`, I would recommend using `sys.getsizeof(s)`:
```suggestion
        self.assertRaises(ValueError, sys.getsizeof(s))
```

In general, it seems to be preferable to use functions instead of directly accessing the special object attributes when possible. If you had a specific reason for not using `sys.getsizeof()`, let me know. Here's the a link to the [function defintion](https://github.com/python/cpython/blob/59ad110d7a7784d53d0b502eebce0346597a6bef/Python/sysmodule.c#L1565) and the [docs](https://docs.python.org/3/library/sys.html#sys.getsizeof). I looked over it, but I'm not very experienced with the python c-api. I mostly rely on the docs when it comes to the modules implemented in c.

aeros167:
Also, this should probably be backported to previous versions. This issue was reported for 3.7 with no noticeable difference in behavior prior to this patch.

jdemeyer:
Wouldn't it make more sense to ensure that the invalid objects can't be created in the first place, by doing the initialization in `__new__` instead of `__init__`?

kumaraditya303:
Superseded by https://github.com/python/cpython/pull/94532

PR #14778, opened 2019-07-14
bpo-37593: Swap the positions of posonlyargs and args in the constructor of ast.parameters nodes

pablogsal:
https://bugs.python.org/issue37593


ilevkivskyi (approved):
Thanks! Makes sense to me.

PR #14782, opened 2019-07-15
[3.8] bpo-36974: separate vectorcall functions for each calling convention (GH-13781)

jdemeyer:
Backport #13781 (part of the PEP 590 implementation) to 3.8 to fix a performance regression, see [bpo-37562](https://bugs.python.org/issue37562).

CC @vstinner @pablogsal @ambv 


pablogsal:
I will run pyperformance today to confirm if this fixes the regression. Thanks for the PR @jdemeyer!

rhettinger:
I can confirm this fixes the performance regression.  Please apply this backport.

rhettinger (approved):

ambv:
One day I'll remember to replace # with GH-. It will be a glorious day.

rhettinger:
Thanks for fixing this :-)

PR #14786, opened 2019-07-15
bpo-37599: Remove a vague statement in documentation of Integer Objects

sgalal:
In current Python document (3.7 - 3.9) there is such statement in documentation of Integer Objects:

> The current implementation keeps an array of integer objects for all integers between ``-5`` and ``256``, when you create an int in that range you actually just get back a reference to the existing object. So it should be possible to change the value of ``1``.  I suspect the behaviour of Python in this case is undefined. :-)

The last sentence is vague. It is inappropriate to write "I suspect" in documentation. And as the statements ahead has already clarified the sense that this is "a current implementation", **the last sentence should be removed**.

https://bugs.python.org/issue37599

https://bugs.python.org/issue37599

---

Edit: Change "irresponsible" to "inappropriate" to avoid pejoratives, according to <https://bugs.python.org/msg348019>.

### DIFF
--- a/Doc/c-api/long.rst
+++ b/Doc/c-api/long.rst
@@ -42,7 +42,7 @@ distinguished from a number.  Use :c:func:`PyErr_Occurred` to disambiguate.
 
    The current implementation keeps an array of integer objects for all integers
    between ``-5`` and ``256``, when you create an int in that range you actually
-   just get back a reference to the existing object. So it should be possible to
+    just get back a reference to the existing object.
    change the value of ``1``.
 
 
--- a/Doc/c-api/long.rst
+++ b/Doc/c-api/long.rst
@@ -42,7 +42,7 @@ distinguished from a number.  Use :c:func:`PyErr_Occurred` to disambiguate.
 
    The current implementation keeps an array of integer objects for all integers
    between ``-5`` and ``256``, when you create an int in that range you actually
-    just get back a reference to the existing object.
+   just get back a reference to the existing object.
 
 
 .. c:function:: PyObject* PyLong_FromUnsignedLong(unsigned long v)
--- a/Doc/c-api/long.rst
+++ b/Doc/c-api/long.rst
@@ -43,7 +43,6 @@ distinguished from a number.  Use :c:func:`PyErr_Occurred` to disambiguate.
    The current implementation keeps an array of integer objects for all integers
    between ``-5`` and ``256``, when you create an int in that range you actually
     just get back a reference to the existing object.
-   change the value of ``1``.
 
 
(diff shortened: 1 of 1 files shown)


the-knights-who-say-ni:

Doc/c-api/long.rst
```diff
@@ -43,8 +43,7 @@ distinguished from a number.  Use :c:func:`PyErr_Occurred` to disambiguate.
    The current implementation keeps an array of integer objects for all integers
    between ``-5`` and ``256``, when you create an int in that range you actually
    just get back a reference to the existing object. So it should be possible to
```
aeros:
```suggestion
    just get back a reference to the existing object.
```
Edit: Removal of the next line is included in a comment below

aeros (approved):
@sgalal Welcome and thanks for the contribution! 

I completely agree with the removal of the vague statement. While cleaning up the summary for this function, I would also recommend removing the previous sentence as well. It is also quite vague and doesn't add significant value. A sentence starting with "So it should be possible..." shouldn't be in the docs either.

aeros167:
If you've already submitted your CLA, it's occasionally required to manually update the status on GitHub if it's been over 24 hours. Let me know if you have any questions.

Doc/c-api/long.rst
```diff
@@ -43,8 +43,7 @@ distinguished from a number.  Use :c:func:`PyErr_Occurred` to disambiguate.
...
-   change the value of ``1``.  I suspect the behaviour of Python in this case is
-   undefined. :-)
+   change the value of ``1``.
```
aeros: Meant to include the removal of this line in my previous review, part of the previous sentence I was referring to:
```suggestion
```

aeros167:
The latest build error from travis wasn't particularly helpful:

`nntplib.NNTPTemporaryError: 400 load at 16.92, try later`

I'd try adding some whitespace the last character of a sentence before the period to trigger the test again. For example, on line 45:

`just get back a reference to the existing object.`
to
`just get back a reference to the existing object .`

Occasionally the integration tests will fail without having a specific reason. If that doesn't work, one of the core devs can try to help troubleshoot it. I had a similar issue with one of my PRs: https://github.com/python/cpython/pull/14611.

aeros167:
If running the tests again doesn't work, another good way to troubleshoot the doc tests is to use the `patchcheck` utility, see the [devguide](https://devguide.python.org/pullrequest/#patchcheck) for more information. This will attempt to automatically correct any whitespace issues. I usually use it as the final step before submitting any PRs.

rhettinger (approved):

PR #14787, opened 2019-07-16
bpo-7202: Parse command global options

potomak:
Add the ability to parse command global options.

If a command has a global option (except help) it will be reapplied to
the distribtuion object.

If a command has a global option (except help) but also defines it as
a local command option, it will not be reapplied to the distribtuion
object.

https://bugs.python.org/issue7202

### DIFF
--- a/Lib/distutils/dist.py
+++ b/Lib/distutils/dist.py
@@ -26,6 +26,8 @@ from distutils.debug import DEBUG
 # to look for a Python module named after the command.
 command_re = re.compile(r'^[a-zA-Z]([a-zA-Z0-9_]*)$')
 
+longopt_revxlate = str.maketrans('_', '-')
+
 
 def _ensure_list(value, fieldname):
     if isinstance(value, str):
@@ -474,7 +476,6 @@ Common commands: (see '--help-commands' for more)
         parser.set_aliases({'licence': 'license'})
         args = parser.getopt(args=self.script_args, object=self)
         option_order = parser.get_option_order()
-        log.set_verbosity(self.verbose)
 
         # for display options we return immediately
         if self.handle_display_options(option_order):
@@ -483,6 +484,8 @@ Common commands: (see '--help-commands' for more)
             args = self._parse_command_opts(parser, args)
             if args is None:            # user asked for help (and got it)
                 return
+
+        log.set_verbosity(self.verbose)
 
         # Handle the cases of --help as a "global" option, ie.
         # "setup.py --help" and "setup.py --help command ...".  For the
@@ -597,11 +600,26 @@ Common commands: (see '--help-commands' for more)
             if help_option_found:
                 return
 
+        global_option_names = {n for n, *_ in self.global_options}
+        command_option_names = {n for n, *_ in cmd_class.user_options}
+
         # Put the options from the command-line into their official
         # holding pen, the 'command_options' dictionary.
         opt_dict = self.get_option_dict(command)
         for (name, value) in vars(opts).items():
-            opt_dict[name] = ("command line", value)
+            opt_name = name.translate(longopt_revxlate)
+            if (  opt_name in global_option_names
+                  and opt_name not in command_option_names):
+                alias = self.negative_opt.get(name)
+                try:
+                    if alias:
+                        setattr(self, alias, not value)
+                    else:
+                        setattr(self, name, value)
+                except ValueError as msg:
+                    raise DistutilsOptionError(msg)
+            else:
+                opt_dict[name] = ("command line", value)
 
         return args
 
--- a/Lib/distutils/dist.py
+++ b/Lib/distutils/dist.py
@@ -26,8 +26,6 @@ from distutils.debug import DEBUG
 # to look for a Python module named after the command.
 command_re = re.compile(r'^[a-zA-Z]([a-zA-Z0-9_]*)$')
 
-longopt_revxlate = str.maketrans('_', '-')
-
 
 def _ensure_list(value, fieldname):
     if isinstance(value, str):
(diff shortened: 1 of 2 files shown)


Lib/distutils/tests/test_dist.py
```diff
@@ -490,6 +490,14 @@ def test_show_help(self):
...
         self.assertTrue(output)
 
+    def test_parse_command_opts(self):
+        dist = Distribution()
+        sys.argv = ["setup.py", "build", "--quiet"]
+        dist.parse_command_line()
+        # Note: the `quiet` option is an alias that negatively affects the
+        # `verbose` option value, that is `True` by default
+        self.assertFalse(dist.verbose)
+        self.assertTrue("verbose" not in dist.get_option_dict("build"))
```
ZackerySpytz: You can use `self.assertNotIn()`.

Lib/distutils/dist.py
```diff
@@ -597,11 +600,26 @@ def _parse_command_opts(self, parser, args):
...
+        global_option_names = {n for n, *_ in self.global_options}
+        command_option_names = {n for n, *_ in cmd_class.user_options}
+
         # Put the options from the command-line into their official
         # holding pen, the 'command_options' dictionary.
         opt_dict = self.get_option_dict(command)
         for (name, value) in vars(opts).items():
-            opt_dict[name] = ("command line", value)
+            opt_name = name.translate(longopt_revxlate)
+            if (  opt_name in global_option_names
```
ZackerySpytz: Please remove the whitespace after the `(`.

Lib/distutils/dist.py
```diff
@@ -597,11 +600,26 @@ def _parse_command_opts(self, parser, args):
...
 
+        global_option_names = {n for n, *_ in self.global_options}
+        command_option_names = {n for n, *_ in cmd_class.user_options}
+
         # Put the options from the command-line into their official
         # holding pen, the 'command_options' dictionary.
         opt_dict = self.get_option_dict(command)
         for (name, value) in vars(opts).items():
-            opt_dict[name] = ("command line", value)
+            opt_name = name.translate(longopt_revxlate)
```
epicfaace: It would be good to add a comment explaining why you need to replace the underscores with hyphens.

Why not just use `name.replace`? Is there a better reason to use `str.maketrans` instead?

Lib/distutils/tests/test_dist.py
```diff
@@ -490,6 +490,14 @@ def test_show_help(self):
                   if line.strip() != '']
         self.assertTrue(output)
 
+    def test_parse_command_opts(self):
```
epicfaace, line 493: Can you add a test case that covers:
- If a command has a global option (except help) but also defines it as
a local command option, it will not be reapplied to the distribtuion
object.

Lib/distutils/tests/test_dist.py
```diff
@@ -490,6 +490,14 @@ def test_show_help(self):
...
+        dist = Distribution()
+        sys.argv = ["setup.py", "build", "--quiet"]
+        dist.parse_command_line()
+        # Note: the `quiet` option is an alias that negatively affects the
+        # `verbose` option value, that is `True` by default
+        self.assertFalse(dist.verbose)
```
epicfaace, line 499: Can you also assert that `dist.quiet` is `True`?

Lib/distutils/dist.py
```diff
@@ -484,6 +485,8 @@ def parse_command_line(self):
             if args is None:            # user asked for help (and got it)
                 return
 
+        log.set_verbosity(self.verbose)
```
epicfaace, line 486: I think this seems a good way to handle this case; you probably don't need to change it.

Lib/distutils/tests/test_dist.py
epicfaace: Additionally, can you add test cases for the --help option, in which both the points in the PR description do not apply?

epicfaace (changes requested):

potomak:
@ZackerySpytz @epicfaace thanks for reviewing. I will update the PR in the next couple of days.

Lib/distutils/dist.py
potomak: The main reason is because I wanted to make it a little bit more explicit (without success) that this is the "reverse" `longopt_xlate` translation that exists in `fancy_getopt`.

I don't think there's any other reason to use `translate` instead of `replace`. I will follow your suggestion and use `replace` instead.

Lib/distutils/dist.py
potomak: The reason why I had to move `log.set_verbosity(self.verbose)` here from line 477 is because `self.verbose` now can be set by `_parse_command_opts` if it is included as a command option instead of a global option.

In general I found it a little bit confusing the state management of this class and I think it can be improved, but for now this update was enough to let the test pass.

Lib/distutils/tests/test_dist.py
potomak: `dist.quiet` doesn't exist. As I explained in the comment `quiet` is just an option that negatively affects the `verbose` option.

Lib/distutils/tests/test_dist.py
potomak: @epicfaace your suggested test case surfaced a very interesting use case.

I tried to find global options also defined as command options:

```
grep -rE "verbose|quiet|dry\-run|no\-user\-cfg" Lib/distutils/command/*
```

The result of this command shows that there's only one command that includes one of the standard options as a command option, that is `bdist_rpm` with the `quiet` option.

This case is special because `quiet` is a (negative) alias for the global option `verbose`, but it's also a `bdist_rpm` command option. At this point of the command line parsing, options have already been translated to their associated Python identifiers, so the `quiet` option has been translated to the `verbose` Python identifier as a result of the global `{'quiet': 'verbose'}` alias translation. In order to match it with the command option `quiet` I would need to check for both `verbose` and its associated alias `quiet` to be present in the command options set.

Long story short: I think there's an issue, not introduced by this PR, in the way negative options aliases are interacting with the command options parsing (`bdist_rpm`'s `quiet` option case) and I can't find another case that is worth it testing.

iritkatriel:
Closing this PR following the closure of its bpo issue (https://bugs.python.org/issue7202)

PR #14788, opened 2019-07-16
Docs: Fix argument name for PyLong_FromLong()

aeros:
In C-API docs, the first positional argument's name for the function `PyLong_FromLong()` is shown as `v`, but in the [function's definition](https://github.com/python/cpython/blob/e8692818afd731c1b7e925c626ac8200b1f1c31e/Objects/longobject.c#L310), it's `ival`.

### DIFF
--- a/Doc/c-api/long.rst
+++ b/Doc/c-api/long.rst
@@ -36,9 +36,9 @@ distinguished from a number.  Use :c:func:`PyErr_Occurred` to disambiguate.
    :c:type:`PyLongObject`.
 
 
-.. c:function:: PyObject* PyLong_FromLong(long v)
+.. c:function:: PyObject* PyLong_FromLong(long ival)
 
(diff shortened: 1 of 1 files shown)


taleinat:
Unlike in Python, in C the names of function arguments are just internal details. Therefore, there's no need for the documentation to accurately reflect argument names used in C code.

aeros167:
@taleinat Oh okay thanks for the clarification. I'm less familiar the c-api docs, so I wasn't sure if the argument names actually mattered for the purposes of documentation. I know that it has no functional difference in C. Is there a guide for the c-api documentation conventions or is it more of a trial by error process?

taleinat:
@aeros167, to my knowledge there isn't a guide specifically about the C API (but do make sure to read the part of the dev guide about [documenting Python](https://devguide.python.org/documenting/)).

What guides me in this case is that the intended audience is the users of the Python language, rather than contributors to the codebase.

PR #14791, opened 2019-07-16
Fix typo in ssl.RAND_bytes documentation

zthompson47:
It looks like "cryptographically strong" is the preferred phrase from the surrounding documentation.


tiran (approved):
Thanks for your contribution!

PR #14794, opened 2019-07-16
bpo-37461: Fix infinite loop in parsing of specially crafted email headers

maxking:
Some crafted email header would cause the get_parameter method to run in an
infinite loop causing a DoS attack surface when parsing those headers. This
patch fixes that by making sure the DQUOTE character is handled to prevent
going into an infinite loop.

### DIFF
--- a/Lib/email/_header_value_parser.py
+++ b/Lib/email/_header_value_parser.py
@@ -2494,11 +2494,11 @@ def get_parameter(value):
         # Treat the rest of value as bare quoted string content.
         v = Value()
         while value:
-            if value[0] == '"':
+            if value[0] in WSP:
+                token, value = get_fws(value)
+            elif value[0] == '"':
                 token = ValueTerminal('"', 'DQUOTE')
                 value = value[1:]
-            elif value[0] in WSP:
-                token, value = get_fws(value)
             else:
                 token, value = get_qcontent(value)
             v.append(token)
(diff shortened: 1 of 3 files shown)


maxking:
working on adding tests and NEWS entry.

maxking:
I have added a slightly different test case than given in BPO and that fails with a different exception. I'll keep this PR around and convert it to a WIP, while I figure out what is going on. I may need to dig deeper than I originally thought and might need more time :)

aeros167:
@maxking Instead of using WIP, you can also start the PR as a draft and open it once the tests are passing. I don't think it's possible to do this for a PR that's already been opened, but just for any future PRs it's a useful feature that I've started using recently.

maxking:
@aeros167 Yeah, I added the tests after I opened the PR, which is why  I had to switch it to WIP. 

It is a feature in Gitlab, which I more frequently use, where you can switch between WIP PRs (a.k.a draft PR) and normal ones by just adding the WIP: prefix.

Thanks for the tip though, I should have opened a draft PR initially.

maxking:
So, I think my initial solution was correct, but my test case was wrong. That is a separate bug and I am working on a separate PR to fix that. Pushed a fix and I think this is ready for a review now.

Fingers crossed for the tests ;-)

warsaw (approved):

PR #14795, opened 2019-07-16
bpo-37499: add test functions for each calling convention

jdemeyer:
https://bugs.python.org/issue37499


jdemeyer:
CC @pganssle

Modules/_testcapimodule.c
```diff
@@ -5856,6 +5856,97 @@ static PyTypeObject Generic_Type = {
...
+}
+
+static PyObject *
+CallTest_fastcall(PyObject *self, PyObject *const *args, Py_ssize_t nargs)
+{
+    return _PyObject_Vectorcall(self, args, nargs, NULL);
+}
+
+static PyObject *
+CallTest_fastcall_keywords(PyObject *self, PyObject *const *args, Py_ssize_t nargs, PyObject *kwnames)
```
pganssle: PEP 7: Line too long,
```suggestion
CallTest_fastcall_keywords(PyObject *self, PyObject *const *args,
                           Py_ssize_t nargs, PyObject *kwnames)
```

Lib/test/test_call.py
```diff
@@ -321,6 +321,9 @@ def static_method():
 
 @cpython_only
 class FastCallTests(unittest.TestCase):
+    from _testcapi import CallTest as C
+    c = C()
```
pganssle: Hm, this is actually quite confusing, `c` and `C` look very similar, I did not even realize that you were alternating between instance methods and class methods.

Can we use more distinct-looking names for these? How about:

```python
from _testcapi import CallTest as CT
ci = CT() 
```

Something like this?

pganssle (changes requested):
Sorry for the long delay in reviewing this.

I am ever-so-slightly worried that we're moving these all over to being method calls instead of function calls, but I can't think of any good reason why it would undermine the point of the tests.

I personally like the namespacing of a method, so I say leave it as is.

Also, it's not causing any new failures, but `test_gdb` is now spitting out a bunch of errors or warnings or something that seem very related:

```
$ ./python -m test test_gdb
Run tests sequentially
0:00:00 load avg: 1.84 [1/1] test_gdb
Function "CallTest_varargs" not defined.
Function "CallTest_varargs" not defined.
Function "CallTest_varargs_keywords" not defined.
Function "CallTest_varargs_keywords" not defined.
Function "CallTest_fastcall" not defined.
Function "CallTest_fastcall" not defined.
Function "CallTest_fastcall_keywords" not defined.
Function "CallTest_fastcall_keywords" not defined.
Function "CallTest_noargs" not defined.
Function "CallTest_noargs" not defined.
Function "CallTest_onearg" not defined.
Function "CallTest_onearg" not defined.
Function "CallTest_call" not defined.
Function "CallTest_call" not defined.
Function "CallTest_call" not defined.
Function "CallTest_call" not defined.
test_gdb passed in 37 sec 821 ms

== Tests result: SUCCESS ==

1 test OK.

Total duration: 37 sec 828 ms
Tests result: SUCCESS
```

encukou:
I hope to look at this at this week's sprint. To make it faster, I'll push changes directly to this branch.
@jdemeyer, let me know if you want more back-and-forth; it's always possible to revert my changes.

Lib/test/test_call.py
encukou: I'd prefer `CallTest` and `calltest`.

pganssle: I still find it very hard to visually distinguish between a sea of references like this:

```python
calltest.blah
CallTest.blah
calltest.blah
```

Can it be something like `ct_inst` and `CallTest`?

```python
ct_inst.blah
CallTest.blah
ct_inst.blah
```

Even `call_test` and `CallTest` would be fine:

```python
call_test.blah
CallTest.blah
call_test.blah
```

jdemeyer:
> Also, it's not causing any new failures, but `test_gdb` is now spitting out a bunch of errors or warnings or something that seem very related:

This has to do with the fact that we're importing these functions from an external module (`_testcapimodule`) instead of being compiled-in functions. I don't know how to suppress these warnings.

encukou:
I did something *similar* – testing that C callables can be called fro Python code. See GH-15776
I have merged code from here into there, but used my code for the GDB test to cover the various functions/methods/classmethods.
If you have some time, could you there & see if it makes sense?

jdemeyer:
Regarding the formatting in `Lib/test/test_call.py`: I think this is a case where the wrongly-formatted original is much more readable than the changed variant.

encukou:
That's one point where I disagree. The aligned code looks neater, but IMO it doesn't help readability much, and makes larger diffs wen it needs to be updated.

pganssle:
Closing this one as superceded by #15776, which includes elements of this PR and some additional tests.

Thanks @jdemeyer!

PR #14804, opened 2019-07-17
bpo-29548: deprecate PyEval_Call* functions

jdemeyer:
CC @methane 

https://bugs.python.org/issue29548


methane (approved):
LGTM.

But I will wait for a while because other core devs may have a different opinion
about stable ABI and Python 4.

PR #14810, opened 2019-07-17
Expand comment explaining update_one_slot

jdemeyer:
This is just changing comments in the code, so `skip issue` and `skip news`.

CC @vstinner

### DIFF
--- a/Objects/typeobject.c
+++ b/Objects/typeobject.c
@@ -7189,8 +7189,9 @@ resolve_slotdups(PyTypeObject *type, PyObject *name)
         *pp = NULL;
     }
 
-    /* Look in all matching slots of the type; if exactly one of these has
-       a filled-in slot, return its value.      Otherwise return NULL. */
+    /* Look in all slots of the type matching the name. If exactly one of these
+       has a filled-in slot, return a pointer to that slot.
+       Otherwise, return NULL. */
     res = NULL;
     for (pp = ptrs; *pp; pp++) {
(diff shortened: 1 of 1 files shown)


Objects/typeobject.c
```diff
@@ -7203,12 +7204,61 @@ resolve_slotdups(PyTypeObject *type, PyObject *name)
...
+ * resulting slot is applied. If the results disagree, then a wrapper for
+ * the special methods is installed. This is always safe, but less efficient
+ * because it uses method lookup instead of direct C calls.
+ *
+ * There are some further special cases for specific slots, like supporting
+ * __hash__ = None for tp_hash and special code for tp_new.
+ *
+ * When done, return a pointer to the next slotdef with a different offset,
+ * because that's convenient for fixup_slot_dispatchers(). This function never
+ * sets an exception: if an internal error happens (unlikely), it's ignored. */
```
epicfaace, line 7261: Is it true that this function never sets an exception? #14836 seems to have an example in which it does set an exception.

jdemeyer: > Is it true that this function never sets an exception?

Yes.

> #14836 seems to have an example in which it does set an exception.

No, that's an example where the wrapper descriptor raises an error, not `update_one_slot()`.

epicfaace: Got it, thanks!

epicfaace (approved):

Yhg1s:
Thanks, this is a great summary of the code.

PR #14814, opened 2019-07-17
bpo-37337: Fix a GCC 9 warning in Objects/descrobject.c

ZackerySpytz:
Commit b1263d5a60d3f7ab02dd28409fff59b3815a3f67 causes GCC 9.1.0 to
give a warning in Objects/descrobject.c.

```
Objects/descrobject.c:1050:19: warning: cast between incompatible function types from ‘PyObject * (*)(mappingproxyobject *, PyObject * const*, Py_ssize_t)’ {aka ‘struct _object * (*)(struct <anonymous> *, struct _object * const*, long int)’} to ‘PyObject * (*)(PyObject *, PyObject *)’ {aka ‘struct _object * (*)(struct _object *, struct _object *)’} [-Wcast-function-type]
 1050 |     {"get",       (PyCFunction)mappingproxy_get,        METH_FASTCALL,
      |                   ^
```

### DIFF
--- a/Objects/descrobject.c
+++ b/Objects/descrobject.c
@@ -1047,7 +1047,7 @@ mappingproxy_copy(mappingproxyobject *pp, PyObject *Py_UNUSED(ignored))
             to the underlying mapping */
 
 static PyMethodDef mappingproxy_methods[] = {
-    {"get",       (PyCFunction)mappingproxy_get,        METH_FASTCALL,
+    {"get",       (PyCFunction)(void(*)(void))mappingproxy_get, METH_FASTCALL,
      PyDoc_STR("D.get(k[,d]) -> D[k] if k in D, else d."
                "  d defaults to None.")},
     {"keys",      (PyCFunction)mappingproxy_keys,       METH_NOARGS,


jdemeyer (approved):

ZackerySpytz:
@methane

ZackerySpytz:
Hi, @methane. Can you please merge this PR?

methane:
The cast seems too complex to just avoid the warning.  But I'm not sure there is a better way.
When I have spare time, I'll try to find simpler ways to fix this.  But I don't have spare time now.
That's why I don't merge this.

If you want to this merged, please provide all casts you tried and demonstrate this is the simplest way to fix the warning.

jdemeyer:
This is the standard way to avoid those warnings, there are many places in CPython with casts like that. It's also documented to do this: https://docs.python.org/3.8/extending/extending.html#keyword-parameters-for-extension-functions

jdemeyer:
See bpo-33012 for the discussion about those casts.

As far as I know, this is indeed the simplest way to cast between incompatible function pointers which does not generate warnings with GCC 9 and which does not rely on undefined behaviour.

methane:
OK, thank you.

PR #14815, opened 2019-07-17
bpo-37560: Add exception handler in FieldStorage cleanup

nsiregar:
Add exception handler on `FieldStorage` at `__exit__` method

https://bugs.python.org/issue37560

### DIFF
--- a/Lib/cgi.py
+++ b/Lib/cgi.py
@@ -483,7 +483,10 @@ class FieldStorage:
         return self
 
     def __exit__(self, *args):
-        self.file.close()
+        try:
+            self.file.close()
+        except AttributeError:
+            pass
 
     def __repr__(self):
         """Return a printable representation."""
--- a/Lib/test/test_cgi.py
+++ b/Lib/test/test_cgi.py
@@ -365,11 +365,8 @@ Larry
 
     def test_fieldstorage_exit_context(self):
         fs = cgi.FieldStorage()
-        try:
-            with fs:
-                fs.bytes_read = 0
-        except AttributeError:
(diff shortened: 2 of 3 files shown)


Lib/test/test_cgi.py
```diff
@@ -363,6 +363,14 @@ def test_fieldstorage_as_context_manager(self):
         with self.assertRaisesRegex(ValueError, 'I/O operation on closed file'):
             fs.file.read()
 
+    def test_fieldstorage_exit_context(self):
+        fs = cgi.FieldStorage()
+        try:
```
ZackerySpytz: I don't think there's a need to use `try` / `except` here in the unit test.

nsiregar: should I remove it? I was assuming it should assert the test.

Lib/cgi.py
```diff
@@ -483,7 +483,10 @@ def __enter__(self):
         return self
 
     def __exit__(self, *args):
-        self.file.close()
+        try:
+            self.file.close()
+        except AttributeError:
```
benjaminp, line 488: I know this comes from `__del__`, but can we instead check whether `self.file` is `None` before calling `close` on it?

nsiregar: I thought it was better to follow EAFP coding style rather than LBYL, do you have any consent why it should check whether `self.file` is `None`?

kumaraditya303:
Closing as cgi module is deprecated as per [PEP 594](https://peps.python.org/pep-0594/) and further no improvements or bug fixes will be made to the module.

PR #14821, opened 2019-07-17
bpo-33610: When toggling code-context on, immediately show the current context.

taleinat:
This avoids a delay of up to 100ms and the accompanying visual artifact.

https://bugs.python.org/issue33610

### DIFF
--- a/Lib/idlelib/codecontext.py
+++ b/Lib/idlelib/codecontext.py
@@ -112,12 +112,13 @@ class CodeContext:
                 padx=padx, border=border, relief=SUNKEN, state='disabled')
             self.update_highlight_colors()
             self.context.bind('<ButtonRelease-1>', self.jumptoline)
+            # Get the current context and initiate the recurring update event.
+            self.timer_event()
             # Pack the context widget before and above the text_frame widget,
             # thus ensuring that it will appear directly above text_frame.
             self.context.pack(side=TOP, fill=X, expand=False,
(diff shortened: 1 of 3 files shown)


terryjreedy:
This is a partial solution.  Open a file, such as editor.py.  Scroll down to where there would be at least 2 line of context. 'Show Code Context'.  Before: a double action; I believe a single blank line is opened and then after a deley, the context is expanded and filled in.  After: the context is opened full size and filled in.  Much better.

To continue: 'Hide Code Context'.  'Show Code Context'.  A 1 line blank context is shown.  ........ Seconds later, nothing is changed.   ... Until do something line scroll that triggers the context.  Diagnosis.  Hide is not restoring everything to where it was before the first Show.  Nor is it putting CC into a 'hidden' state that is properly restored by 'Show'.  I have not looked yet.

The new line is likely part of a full solution and should be merged if we cannot do better before b3.

taleinat:
@terryjreedy, I've pushed a fix for the issue of wrong context shown after toggling off and back on. This was actually an entirely separate issue.

I've also added tests for both issues.

terryjreedy (approved):
Problem fixed.  There is an issue with flashing (window manager clearing an area before overwriting it) when CC is in 'add' mode, versus 'sub' mode.  Another reason to force CC into 'sub' mode.  Nothing I tried made a definite improvement, so I am merging now, so we can move on.

PR #14822, opened 2019-07-17
bpo-33610: validate non-negative integer inputs in IDLE's config

taleinat:
This forces the entry widgets to always only contain digits or be empty.

https://bugs.python.org/issue33610

### DIFF
--- a/Lib/idlelib/configdialog.py
+++ b/Lib/idlelib/configdialog.py
@@ -9,6 +9,8 @@ Note that tab width in IDLE is currently fixed at eight due to Tk issues.
 Refer to comments in EditorWindow autoindent code for details.
 
 """
+import re
+
 from tkinter import (Toplevel, Listbox, Text, Scale, Canvas,
                      StringVar, BooleanVar, IntVar, TRUE, FALSE,
                      TOP, BOTTOM, RIGHT, LEFT, SOLID, GROOVE,
@@ -1764,8 +1766,18 @@ class GenPage(Frame):
 
     def __init__(self, master):
         super().__init__(master)
+
+        self.init_validators()
         self.create_page_general()
         self.load_general_cfg()
+
+    def init_validators(self):
+        digits_or_empty_re = re.compile(r'[0-9]*')
+        def is_digits_or_empty(s):
+            "Return 's is blank or contains only digits'"
+            return digits_or_empty_re.fullmatch(s) is not None
+        self.digits_or_empty_vldcmd = (
+            self.register(is_digits_or_empty), '%P',)
 
     def create_page_general(self):
         """Return frame of widgets for General tab.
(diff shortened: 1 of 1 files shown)


Lib/idlelib/configdialog.py
```diff
@@ -1764,9 +1766,19 @@ class GenPage(Frame):
...
+        self.init_validators()
         self.create_page_general()
         self.load_general_cfg()
 
+    def init_validators(self):
+        digits_or_empty_re = re.compile(r'[0-9]*')
+        def is_digits_or_empty(s):
+            "Return 's is blank or contains only digits'"
+            return digits_or_empty_re.fullmatch(s) is not None
+        self.digits_or_empty_vldcmd = (
```
terryjreedy: I would strongly prefer a simpler name, such as 'digits_only', or 'check_digits', or 'non_negative'.

Lib/idlelib/configdialog.py
```diff
@@ -1764,9 +1766,19 @@ class GenPage(Frame):
...
+
+        self.init_validators()
         self.create_page_general()
         self.load_general_cfg()
 
+    def init_validators(self):
+        digits_or_empty_re = re.compile(r'[0-9]*')
+        def is_digits_or_empty(s):
+            "Return 's is blank or contains only digits'"
+            return digits_or_empty_re.fullmatch(s) is not None
```
terryjreedy, line 1778: Possible immediate alternative:
```
try:
    return 0 <= int(s)
except ValueError:
     return False
```
Or, adding range check,
```
def rangecheck(widget, lo, hi):  # At global level, so can use on other tabs (and test).
    def inner(s):
        try:
            valid = lo <= int(s) <= hi
        except ValueError:
            valid = False
        if not valid: <showmessage: f"Entry must be int from {lo} to {hi}, inclusive">
        return valid
    return widget.register(inner, '%P')
# Example use.
        mincheck = rangecheck(frame_win_size, 10, 999)
        ...
                   validate='key', validatecommand=mincheck,
```

terryjreedy (commented):
Without the patch, I entered '15aa' for max context lines.  When I closed and reopened config dialog, it only showed 15.  'OK, no damage done.'  When I closed and restarted IDLE, 
" Warning: config.py - IdleConf.GetOption -
 invalid 'int' value for configuration option 'maxlines'
 from section 'CodeContext': '15aa'

So I highly approve of adding something to prevent this in the next week (but see name comment).

It would be nice to check that an entry is in an allowed range, with an error message.   I suspect that 0 for some entries could disable IDLE or at least have negative effects.  But I have not tested my suggestion, limits would need some discussion, and it might be a good idea to put them in a global dict.

I mentioned not restricting validation to the General tab because I want to replace the indent slider with an Entry box.  I would like to eventually move this to the first section of the General tab, but perhaps only when splitting that tab into two.

Lib/idlelib/configdialog.py
taleinat: > I mentioned not restricting validation to the General tab because I want to replace the indent slider with an Entry box. I would like to eventually move this to the first section of the General tab, but perhaps only when splitting that tab into two.

I'd like to leave that for later, then.

> Possible immediate alternative:
> 
> ```
> try:
>     return 0 <= int(s)
> except ValueError:
>      return False
> ```

I avoided this because `int()` allows leading '+' and '-' signs, which IMO we should disallow to keep things simple. (For example, `-0` is a valid non-negative integer.)

> It would be nice to check that an entry is in an allowed range, with an error message.

At a quick glance, the existing integer config options should either be non-negative, such as timeouts, or positive (i.e. non-zero), such as window width or minimum context lines. Following is a list with my suggested allowed ranges; let me know what you think, @terryjreedy.

1. Initial Window Width & Height: Positive
2. Completions Popup Wait (milliseconds): Non-negative
3. Paren-Match Time Displayed (zero = until next input): Non-negative
4. Format Paragraph Max. Width: Positive
5. Max. Context Lines: Positive
6. Auto-Squeeze Min. Lines: Non-negative (zero would mean the same as one, but allowing entering zero to have no minimum is harmless and could be intuitive to some)

terryjreedy: For a simple check, use the current code (with name change).
For a range check, check 1st char first, then int range.

1. Windows resize min is 10 (Command Prompt) or more -- enough to show '--  [_]  X' on the title bar.  A setting of 2 gives a window 10.x 11 point chars wide.  10-999 seem sensible to me for 1 and 4.

6. Someone requested (Stackoverflow, I think) 0 = infinity (large number).  This is similar to 3, which has explicit note about its convention.

taleinat:
The trouble with range checks is that we shouldn't run them on every change, rejected invalid changes, as we do with the "empty or digits only" check. Not allowing zero would be fine, because there's never a good reason to write a number with leading zeros in such a config dialog. But anything beyond that would make editing the value a hassle.

For example, if I wanted to change the default window width from 80 to 100, I couldn't delete the "80" and type "100" nor could I delete the "8" and add "10". I would have to do something like type "100" before the "80", resulting in "80100", and then delete the "80".

Assuming we agree that that is horrible UX, we'd need to validate the value on focus out and/or hitting the "apply" or "ok" buttons, and somehow inform the user which value(s) they've set are invalid and why. That's orders of magnitude more complex than what's needed for my suggestion above.

I suggest staying with the current method, making a few entries non-zero, and leaving the range checks for a separate issue.

terryjreedy:
You are right.  A range validation should wait for a later event than key press in the widget.  That includes checking for initial 0, as changing 80 to 0 to 90 would temporarily have a leading 0.  So keep the current check.

PR #14823, opened 2019-07-17
bpo-33408: Enable AF_UNIX support in Windows

paulmon:
These changes are incomplete, but this seems like a good point to get feedback on what I've learned so far.

Tests in test_asyncio are failing because create_unix_server and create_unix_connection are not implemented for Windows on the ProactorEventLoop class.  

Linux implements create_unix_server and create_unix_connection on the _UnixEventSelectorLoop class, and the implementation cannot be simply copied because Linux uses signals, which don't exist for Windows. Windows appears to uses IO completion ports to achieve the same results.

https://bugs.python.org/issue33408

### DIFF
--- a/Lib/socket.py
+++ b/Lib/socket.py
@@ -494,7 +494,12 @@ if hasattr(_socket, "socketpair"):
 else:
 
     # Origin: https://gist.github.com/4325783, by Geert Jansen.  Public domain.
-    def socketpair(family=AF_UNIX, type=SOCK_STREAM, proto=0):
+    def socketpair(family=None, type=SOCK_STREAM, proto=0):
+        if family is None:
+            try:
+                family = AF_UNIX
+            except NameError:
+                family = AF_INET
         if family == AF_INET:
             host = _LOCALHOST
         elif family == AF_INET6:
--- a/Lib/socket.py
+++ b/Lib/socket.py
@@ -494,11 +494,14 @@ if hasattr(_socket, "socketpair"):
 else:
 
     # Origin: https://gist.github.com/4325783, by Geert Jansen.  Public domain.
-    def socketpair(family=AF_INET, type=SOCK_STREAM, proto=0):
+    def socketpair(family=AF_UNIX, type=SOCK_STREAM, proto=0):
         if family == AF_INET:
             host = _LOCALHOST
         elif family == AF_INET6:
             host = _LOCALHOST_V6
+        elif family == AF_UNIX:
+            from uuid import uuid4
+            host = os.path.join(os.environ['TEMP'],  str(uuid4()))
         else:
             raise ValueError("Only AF_INET and AF_INET6 socket address families "
                              "are supported")
@@ -511,7 +514,10 @@ else:
         # setblocking(False) that prevents us from having to create a thread.
         lsock = socket(family, type, proto)
         try:
-            lsock.bind((host, 0))
+            if family == AF_UNIX:
+                lsock.bind(host)
+            else:
+                lsock.bind((host, 0))
             lsock.listen()
             # On IPv6, ignore flow_info and scope_id
             addr, port = lsock.getsockname()[:2]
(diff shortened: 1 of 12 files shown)


Lib/socket.py
```diff
@@ -494,11 +494,14 @@ def socketpair(family=None, type=SOCK_STREAM, proto=0):
 else:
 
     # Origin: https://gist.github.com/4325783, by Geert Jansen.  Public domain.
-    def socketpair(family=AF_INET, type=SOCK_STREAM, proto=0):
+    def socketpair(family=AF_UNIX, type=SOCK_STREAM, proto=0):
```
paulmon: This causes test failures in appveyor and Azure Pipelines.  Must be refactored

Lib/socket.py
```diff
@@ -494,11 +494,19 @@ def socketpair(family=None, type=SOCK_STREAM, proto=0):
...
+                family = AF_INET
         if family == AF_INET:
             host = _LOCALHOST
         elif family == AF_INET6:
             host = _LOCALHOST_V6
+        elif family == AF_UNIX:
+            from uuid import uuid4
+            host = os.path.join(os.environ['TEMP'],  str(uuid4()))
         else:
             raise ValueError("Only AF_INET and AF_INET6 socket address families "
```
zooba, line 511: Update this message?

Lib/socket.py
```diff
@@ -494,11 +494,19 @@ def socketpair(family=None, type=SOCK_STREAM, proto=0):
...
+                family = AF_UNIX
+            except NameError:
+                family = AF_INET
         if family == AF_INET:
             host = _LOCALHOST
         elif family == AF_INET6:
             host = _LOCALHOST_V6
+        elif family == AF_UNIX:
+            from uuid import uuid4
+            host = os.path.join(os.environ['TEMP'],  str(uuid4()))
```
zooba, line 509: I'm really not a fan of this - can we use unnamed sockets instead?

Lib/socket.py
```diff
@@ -529,6 +543,8 @@ def socketpair(family=AF_INET, type=SOCK_STREAM, proto=0):
                 raise
         finally:
             lsock.close()
+            if family == AF_UNIX:
+                os.unlink(host)
```
zooba, line 547: Yeah, I don't like this at all :)

Lib/test/test_asyncore.py
```diff
@@ -469,7 +469,12 @@ class BaseServer(asyncore.dispatcher):
     def __init__(self, family, addr, handler=BaseTestHandler):
         asyncore.dispatcher.__init__(self)
         self.create_socket(family)
-        self.set_reuse_addr()
+        if sys.platform == 'win32' and family == socket.AF_UNIX:
+            # calling set_reuse_addr() on Windows with family AF_UNIX results in:
+            # OSError: [WinError 10045] The attempted operation is not supported for the type of object referenced
```
zooba, line 474: `set_reuse_addr` seems to handle all `OSError` silently - why do we need to skip it here?

Lib/test/test_socket.py
```diff
@@ -5276,8 +5276,14 @@ def bind(self, sock, path):
                 raise
 
     def testUnbound(self):
-        # Issue #30205 (note getsockname() can return None on OS X)
-        self.assertIn(self.sock.getsockname(), ('', None))
+        if sys.platform == 'win32':
+            # Getting the name of unbound socket on Windows
+            # raises an exception
```
zooba, line 5281: Does this make more sense than returning `None`? Which is apparently a valid return value

Lib/test/test_socketserver.py
```diff
@@ -232,12 +233,14 @@ def test_ForkingUDPServer(self):
                             self.dgram_examine)
 
     @requires_unix_sockets
+    @unittest.skipIf(sys.platform == 'win32', "no datagram support")
```
zooba, line 236: I'm seeing a lot of these - is there any more specific way to detect datagram support? The socket module already has a number of conditional constants, so maybe one of those will help?

Lib/socket.py
asvetlov: Abstract sockets (starting from NULL in the name) is another options.
Web says that Windows supports it.

csabella:
@paulmon, please address the code reviews and resolve the merge conflict.  Thank you!

paulmon:
I don't currently have time to follow up on this PR or the immediate need for AF_UNIX in Windows

AmirHmZz:
Will python support AF_UNIX on windows ?

PR #14828, opened 2019-07-18
Indicate locations of parse_qs, parse_qsl, escape

simonw:
Since they have been removed from cgi it's useful to remind people where they
can be found instead.


simonw:
I believe this change is OK to have "skip issue" and "skip news" applied.

PR #14833, opened 2019-07-18
Adjust zip() docstring

sir-sigurd:
zip() can be called without any arguments

### DIFF
--- a/Python/bltinmodule.c
+++ b/Python/bltinmodule.c
@@ -2655,7 +2655,7 @@ static PyMethodDef zip_methods[] = {
 };
 
 PyDoc_STRVAR(zip_doc,
-"zip(iter1 [,iter2 [...]]) --> zip object\n\
+"zip(*iterables) --> zip object\n\
 \n\
 Return a zip object whose .__next__() method returns a tuple where\n\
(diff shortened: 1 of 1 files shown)


tirkarthi (approved):
LGTM. Docs `Doc/library/functions.rst` already use it. Thanks.

tirkarthi:
cc: @rhettinger

brettcannon (approved):

brettcannon:
Thanks!

tirkarthi:
Brett, can this be backported to 3.8 and 3.7? Even 2.7 has similar docstring that could be updated. https://github.com/python/cpython/blob/46c2eff5adca7dc309f077ec65faf95b95c47b43/Python/bltinmodule.c#L2619

Thanks

brettcannon:
@tirkarthi I'm a bit hesitant just because you never know who's parsing this stuff and it's a minor change that doesn't correct bad information, it just makes it easier to read. If another core dev wants to do the backport I won't object, though.

brettcannon:
@tirkarthi  Actually, I take that back as it does fix the `zip()` doc issue, so I will backport for 3.7 and 3.8.

tirkarthi:
Thanks Brett :)

PR #14835, opened 2019-07-18
bpo-13272: 2to3 string constants fixer

aldwinaldwin:
2to3 fixer for string constants.
Renames `lowercase`, `uppercase`, and `letters` with the prefix `ascii_`.

https://bugs.python.org/issue13272


iritkatriel:
Closed under bpo-45544.

PR #14836, opened 2019-07-18
bpo-37619: don't ignore wrapper descriptors for wrong class

jdemeyer:
https://bugs.python.org/issue37619


epicfaace (approved):

Yhg1s:
This is a rather subtle bug, and I can't tell if it was an oversight in logic or just a plain bug in the code... good job finding it :)

PR #14841, opened 2019-07-18
bpo-37609: Add device path support in ntpath splitdrive

nsiregar:
Add device path support in `ntpath.splitdrive`

https://bugs.python.org/issue37609

### DIFF
--- a/Lib/ntpath.py
+++ b/Lib/ntpath.py
@@ -171,7 +171,7 @@ def is_drive_path(path, colon):
 
 
 def is_extended_unc(path, colon):
-    return path[2] in ['?', '.'] and path[-2] != colon
+    return path[2] in ['?', '.'] and path[-2] != colon and path[4:7] == 'UNC'
 
 
 def get_separator(path):
--- a/Lib/ntpath.py
+++ b/Lib/ntpath.py
@@ -146,7 +146,9 @@ def splitdrive(p):
                 return p[:0], p
             if is_extended_unc(normp, colon):
                 start = normp.find(sep, index + 1)
-                index = normp.find(sep, start + 1)
+                drive_colon = normp.find(colon, index + 1)
+                if drive_colon == -1 or drive_colon > start:
+                    index = normp.find(sep, start + 1)
             index2 = normp.find(sep, index + 1)
             # a UNC path can't have two slashes in a row
             # (after the initial two)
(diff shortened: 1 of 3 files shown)


Misc/NEWS.d/next/Library/2019-07-18-14-39-59.bpo-37609.1ZYNbG.rst
```diff
@@ -0,0 +1 @@
+Add support for unc device path in :mod:`ntpath.splitdrive`
```
ZackerySpytz: Using the mod role here is not correct. Please see https://devguide.python.org/documenting/#id4.

nsiregar: Thanks, I've updated it to use `meth`.

ZackerySpytz: It should be `func`.

nsiregar: Thanks, I've updated it

zooba:
Sorry for not seeing this earlier, I have a few general suggestions:
* put a leading underscore on the new functions (or else add full documentation for them)
* use constant strings in the comparisons, rather than `sep*2`
* add some tests using backslashes as well, not just forward slashes

Lib/ntpath.py
```diff
@@ -159,11 +157,30 @@ def splitdrive(p):
             if index2 == -1:
                 index2 = len(p)
             return p[:index2], p[index2:]
-        if normp[1:2] == colon:
+        if is_drive_path(normp, colon):
             return p[:2], p[2:]
     return p[:0], p
 
 
+def is_unc_path(path, sep):
```
eryksun, line 165: These functions are clearly private (e.g. parameterized by `sep` and `colon`), so the function names should have a leading underscore.

nsiregar:
working on it

eryksun:
For an alternate and more complete implementation, see the splitdrive.py file that I attached to bpo-37609.

barneygale:
I managed to fix the issue elsewhere (in #91882), so I'm going to close this PR, but thank you very much for taking a look at it! :)

PR #14842, opened 2019-07-18
bpo-36833: Add tests for Datetime C API Macros

nanjekyejoannah:
I have added the missing tests for the Datetime C API Macros.

https://bugs.python.org/issue36833

### DIFF
--- a/Lib/test/datetimetester.py
+++ b/Lib/test/datetimetester.py
@@ -5963,7 +5963,7 @@ class CapiTest(unittest.TestCase):
     def test_get_datetime_object_macros(self):
         class DateTimeSubclass(datetime):
             pass
-    
+
         exp_dt = datetime(1993, 8, 26, 22, 12, 55, 99999)
         exp_dts = DateTimeSubclass(1993, 8, 26, 22, 12, 55, 99999)
 
--- a/Lib/test/datetimetester.py
+++ b/Lib/test/datetimetester.py
@@ -5920,24 +5920,24 @@ class CapiTest(unittest.TestCase):
         class TimeDeltaSubclass(timedelta):
             pass
 
-        exp_td = timedelta(1)
-        exp_tds = TimeDeltaSubclass(1)
+        exp_delta = timedelta(26, 55, 99999)
+        exp_delta_s = TimeDeltaSubclass(26, 55, 99999)
 
-        days, seconds, microseconds = _testcapi.get_fields_from_delta(exp_td)
-        days_s, seconds_s, microseconds_s = _testcapi.get_fields_from_delta(exp_tds)
+        days, seconds, microseconds = _testcapi.get_fields_from_delta(exp_delta)
+        days_s, seconds_s, microseconds_s = _testcapi.get_fields_from_delta(exp_delta_s)
 
         with self.subTest(testname="PyDateTime_DELTA_GET_DAYS"):
-            self.assertEqual(days, exp_dt.days)
-            self.assertEqual(days_s, exp_tds.days)
+            self.assertEqual(days, exp_delta.days)
+            self.assertEqual(days_s, exp_delta_s.days)
 
         with self.subTest(testname="PyDateTime_DELTA_GET_SECONDS"):
-            self.assertEqual(seconds, exp_td.seconds)
-            self.assertEqual(seconds_s, exp_tds.seconds)
+            self.assertEqual(seconds, exp_delta.seconds)
+            self.assertEqual(seconds_s, exp_delta_s.seconds)
 
         with self.subTest(testname="PyDateTime_DELTA_GET_MICROSECONDS"):
-            self.assertEqual(microseconds, exp_td.microseconds)
-            self.assertEqual(microseconds_s, exp_tds.microseconds)
-    
+            self.assertEqual(microseconds, exp_delta.microseconds)
+            self.assertEqual(microseconds_s, exp_delta_s.microseconds)
+
     def test_get_date_object_macros(self):
         class DateSubclass(date):
             pass
@@ -5980,7 +5980,7 @@ class CapiTest(unittest.TestCase):
 
         with self.subTest(testname="PyDateTime_DATE_GET_SECOND"):
             self.assertEqual(second, exp_dt.second)
-            self.assertEqual(minute_s, exp_dts.second)
+            self.assertEqual(second_s, exp_dts.second)
 
         with self.subTest(testname="PyDateTime_DATE_GET_MICROSECOND"):
             self.assertEqual(microsecond, exp_dt.microsecond)
@@ -5990,8 +5990,8 @@ class CapiTest(unittest.TestCase):
         class TimeSubclass(time):
             pass
 
-        exp_t = time(12, 30)
-        exp_ts = TimeSubclass(12, 30)
+        exp_t = time(12, 30, 20, 10)
+        exp_ts = TimeSubclass(12, 30, 20, 10)
 
         hour, minute, second, microsecond = _testcapi.get_fields_from_time(exp_t)
         hour_s, minute_s, second_s, microsecond_s = _testcapi.get_fields_from_time(exp_ts)
@@ -6006,7 +6006,7 @@ class CapiTest(unittest.TestCase):
 
         with self.subTest(testname="PyDateTime_TIME_GET_SECOND"):
             self.assertEqual(second, exp_t.second)
-            self.assertEqual(minute_s, exp_ts.second)
+            self.assertEqual(second_s, exp_ts.second)
 
         with self.subTest(testname="PyDateTime_TIME_GET_MICROSECOND"):
             self.assertEqual(microsecond, exp_t.microsecond)
--- a/Lib/test/datetimetester.py
+++ b/Lib/test/datetimetester.py
@@ -5916,6 +5916,102 @@ class CapiTest(unittest.TestCase):
 
                 self.assertEqual(dt1.astimezone(timezone.utc), dt_utc)
 
+    def test_get_delta_object_macros(self):
+        class TimeDeltaSubclass(timedelta):
+            pass
+
+        exp_td = timedelta(1)
+        exp_tds = TimeDeltaSubclass(1)
+
+        days, seconds, microseconds = _testcapi.get_fields_from_delta(exp_td)
+        days_s, seconds_s, microseconds_s = _testcapi.get_fields_from_delta(exp_tds)
+
+        with self.subTest(testname="PyDateTime_DELTA_GET_DAYS"):
+            self.assertEqual(days, exp_dt.days)
+            self.assertEqual(days_s, exp_tds.days)
+
+        with self.subTest(testname="PyDateTime_DELTA_GET_SECONDS"):
+            self.assertEqual(seconds, exp_td.seconds)
+            self.assertEqual(seconds_s, exp_tds.seconds)
+
+        with self.subTest(testname="PyDateTime_DELTA_GET_MICROSECONDS"):
+            self.assertEqual(microseconds, exp_td.microseconds)
+            self.assertEqual(microseconds_s, exp_tds.microseconds)
+    
+    def test_get_date_object_macros(self):
+        class DateSubclass(date):
+            pass
+
+        exp_dt = date(2000, 1, 2)
+        exp_dts = DateSubclass(2000, 1, 2)
+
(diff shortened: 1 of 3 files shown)


nanjekyejoannah:
cc @vstinner , @ericsnowcurrently

pganssle (changes requested):
I have not had time to give this an amazingly thorough review - overall I like the PR, but I think we can refactor the tests to avoid some repetition like so:

```python
def test_get_date_object_macros(self):
    class DateSubclass(date):
        pass

    for klass in [date, DateSubclass]:
        for args in [(2000, 1, 2), (2012, 2, 29)]:
            d = klass(*args)
            with self.subtest(cls=klass, date=args):
                year, month, day = _testcapi.get_fields_from_date(d)

                self.assertEqual(year, d.year)
                self.assertEqual(month, d.month)
                self.assertEqual(day, d.day)
```

I like the use of subtests in general, but I think if you're going to go that granular with the subtests it would also make sense to make a separate field retrieval method for each field (because the "get year" test seems about as likely to fail with an error as to fail by getting the wrong value), but I don't know if that's necessary.

I will look at the other parts of this when I have a bit more time.

nanjekyejoannah:
@pganssle I will wait for your final review to update the PR.

Modules/_testcapimodule.c
```diff
@@ -2587,6 +2587,117 @@ get_datetime_fromtimestamp(PyObject* self, PyObject *args)
     return rv;
 }
 
+static PyObject *
+get_fields_from_date(PyObject* self, PyObject *args)
```
pganssle:
```suggestion
get_fields_from_date(PyObject *self, PyObject *args)
```

Same change applies to all the other C functions - the `*` needs to be grouped with the variable name.

Modules/_testcapimodule.c
```diff
@@ -2587,6 +2587,117 @@ get_datetime_fromtimestamp(PyObject* self, PyObject *args)
...
+{
+    PyObject *obj = NULL;
+    PyObject *rv;
+    int year, month, day;
+
+    if (!PyArg_ParseTuple(args, "O", &obj)) {
+        return NULL;
+    }
+
```
pganssle: It's not showing up in the GH interface, but on my local checkout this line has trailing whitespace. Can you check and see if that's the case for you, and if so remove it?

pganssle (changes requested):
Other than the changes to the tests that I already mentioned and a few PEP 7 nits, this looks good to me, thanks Joannah!

By the way, in case you think I somehow have the ability to actually *remember* what PEP 7 says, let me correct that misapprehension right now :wink:. I created a [clang-format style](https://gist.github.com/pganssle/0e3a5f828b4d07d79447f6ced8e7e4db) that gets as close to PEP 7 as I could figure out how to approximate, and I just ran it on your changes.

At some point I will figure out how to integrate that into git so that I can auto-format my changes in `git`, but it seems finicky.

nanjekyejoannah:

Modules/_testcapimodule.c
nanjekyejoannah: I saw a trailing space. I tried to remove it. Please confirm.

Modules/_testcapimodule.c
```diff
@@ -2587,6 +2587,117 @@ get_datetime_fromtimestamp(PyObject* self, PyObject *args)
...
+    year = PyDateTime_GET_YEAR(obj);
```
vstinner, line 2595: PyDateTime_GET_YEAR is documented to return an int... but does it really return a *signed* integer? Macro:

```
typedef struct
{
    _PyTZINFO_HEAD
    unsigned char data[_PyDateTime_DATE_DATASIZE];
} PyDateTime_Date;

#define PyDateTime_GET_YEAR(o)     ((((PyDateTime_Date*)o)->data[0] << 8) | \
                     ((PyDateTime_Date*)o)->data[1])
```

I don't know how the compiler promotes integer types when using unsigned char << 8.

Same question for PyDateTime_GET_MONTH and PyDateTime_GET_DAY which seem to return *unsigned char* (so unsigned integer) in practice.

@pganssle: What do you think of modifying these 3 macros to explicitly cast to int? By the way, I'm trying to move away from macros and use static inline functions instead. Would it make sense for these 3 macros?

pganssle: No valid `datetime` object will return a negative number for any of these values, so the difference between signed and unsigned integers in this case seems a bit academic. Are you suggesting that these tests would need to change if the macro were explicitly casting to int? Otherwise I think this is a separate issue unrelated to this PR.

> By the way, I'm trying to move away from macros and use static inline functions instead. Would it make sense for these 3 macros?

I'm not entirely sure what the criteria are for this *migrating* a macro to a function like this, but I would absolutely like to at least *expose* a function that does this. It will help with PyPy and other language compatibilities. Somewhere deep in my to do list is to explicitly document the capsule-based API and recommend it as the preferred way to interact with the `datetime` module in extension modules. That seems like a natural place to put accessor functions (though I admit I don't know how well that will play with the "static" and "inline" portions of your suggestion), with the added benefit of no backwards compatibility issues.

vstinner: static inline doesn't bring anything to PyPy and other languages: it's just to better define the return type and avoid macros uglyness. See https://bugs.python.org/issue35059 and https://gcc.gnu.org/onlinedocs/cpp/Macro-Pitfalls.html for example. See also § Convert macros to static inline functions: https://vstinner.github.io/split-include-directory-python38.html

> No valid datetime object will return a negative number for any of these values, so the difference between signed and unsigned integers in this case seems a bit academic. Are you suggesting that these tests would need to change if the macro were explicitly casting to int? Otherwise I think this is a separate issue unrelated to this PR.

It can be done in a separated PR. I was surprised by the usage of PyLong_FromSsize_t(): I don't expect a year to need size_t type, the datetime module is limited to year 9999 which fits into an int.

Maybe replace PyLong_FromSsize_t() with PyLong_FromLong()? :-)

nanjekyejoannah: I guess changing from PyLong_FromSsize_t() to PyLong_FromLong() should only apply to the year. no? @vstinner

vstinner: You can use PyLong_FromLong() for all C variables of type "int". Which seems to be the case for all your variables, no?

nanjekyejoannah:

Modules/_testcapimodule.c
```diff
@@ -5076,6 +5187,10 @@ static PyMethodDef TestMethods[] = {
     {"get_delta_fromdsu",        get_delta_fromdsu,              METH_VARARGS},
     {"get_date_fromtimestamp",   get_date_fromtimestamp,         METH_VARARGS},
     {"get_datetime_fromtimestamp", get_datetime_fromtimestamp,   METH_VARARGS},
+    {"get_fields_from_date",       get_fields_from_date,         METH_VARARGS},
```
vstinner: I suggest to rename these functions to have a name closer to their macro. Example:

```suggestion
    {"PyDateTime_GET",       test_PyDateTime_GET,         METH_VARARGS},
```

* get_fields_from_date => PyDateTime_GET
* get_fields_from_datetime => PyDateTime_DATE_GET
* get_fields_from_time => PyDateTime_TIME_GET
* get_fields_from_delta => PyDateTime_DELTA_GET

Currently, I'm very confused by "from_date" which uses "PyDateTime_GET", whereas "from_datetime" uses "PyDateTime_DATE_GET".

Note: I know that there are existing "get_xxx" helper functions. I'm fine with leaving existing names unchanged. But I would prefer to not add futher confusing names. _testcapi is used by the whole stdlib, I prefer to use reuse names of the C API.

vstinner (changes requested):

Lib/test/datetimetester.py
```diff
@@ -5916,6 +5916,65 @@ def test_timezones_capi(self):
 
                 self.assertEqual(dt1.astimezone(timezone.utc), dt_utc)
 
+    def test_get_delta_object_macros(self):
```
vstinner: Would you mind to also rename the test methods? Like test_PyDateTime_DELTA_GET().

Modules/_testcapimodule.c
```diff
@@ -2587,6 +2587,117 @@ get_datetime_fromtimestamp(PyObject* self, PyObject *args)
...
 }
 
+static PyObject *
+test_PyDateTime_GET(PyObject *self, PyObject *args)
+{
+    PyObject *obj = NULL;
+    PyObject *rv;
+    int year, month, day;
+
+    if (!PyArg_ParseTuple(args, "O", &obj)) {
```
vstinner: Since the function takes exactly one argument of type PyObject: you can change the call convention from METH_VARARGS to METH_O.

(Same comment for the other added functions.)

Modules/_testcapimodule.c
```diff
@@ -2587,6 +2587,117 @@ get_datetime_fromtimestamp(PyObject* self, PyObject *args)
...
+    rv = PyTuple_New(3);
+    if (rv == NULL) {
+        return NULL;
+    }
+
+    PyTuple_SET_ITEM(rv, 0, PyLong_FromLong(year));
+    PyTuple_SET_ITEM(rv, 1, PyLong_FromLong(month));
+    PyTuple_SET_ITEM(rv, 2, PyLong_FromLong(day));
+
+    return rv;
```
vstinner: These 10 lines of code can be simplfiied using Py_BuildValue as a single line, Py_BuildValue handles errors for you.

```
return Py_BuildValue("iii", year, month, day);
```

(Same comment for the other added functions.)

Misc/NEWS.d/next/Tests/2019-07-18-14-52-58.bpo-36833.Zoe9ek.rst
```diff
@@ -0,0 +1 @@
+Added tests for the macros with no corresponding API module.
```
vstinner: I propose:

```
Added tests for PyDateTime_xxx_GET_xxx() macros of the C API of the :mod:`datetime` module.
```

vstinner (changes requested):

Modules/_testcapimodule.c
pganssle: I'm not sure I agree with this change. These are Python wrapper functions for the C API, and they don't do the same thing as anything in the C API - they construct a tuple of Python objects rather than retrieving the fields from the C structs. As such, I think it makes sense to use a Python naming convention.

I will concede, though, that it does make it clearer what these are intended to test just from reading the names.

> _testcapi is used by the whole stdlib, I prefer to use reuse names of the C API.

Is this true? Are we only supposed to be putting test support functions in here? If so, I think we may want to move the datetime C API test functions into a dedicated module, because I've been using this file as explicit wrappers for *specific tests*.

vstinner: Sorry, I wanted to write _testcapi is used to **test** the whole stdlib.

I prefer names starting with "PyDateTime_" to show that these functions are related to the datetime module. The relationship between "get_fields_from_delta" and "datetime" in less obvious.

> If so, I think we may want to move the datetime C API test functions into a dedicated module, because I've been using this file as explicit wrappers for specific tests.

_testcapi became really large: 6154 lines of C code. It tests a lot of unrelated things. I'm open to discuss creating a new separated C extension just to test the datetime module, but that should be done in a separated issue and separated PR.

nanjekyejoannah:

pganssle (approved):
This looks good to me.

@vstinner Any further changes from you?

Modules/_testcapimodule.c
```diff
@@ -2587,6 +2587,55 @@ get_datetime_fromtimestamp(PyObject* self, PyObject *args)
...
+static PyObject *
+test_PyDateTime_GET(PyObject *self, PyObject *obj)
+{
+    int year, month, day;
+
+    year = PyDateTime_GET_YEAR(obj);
+    month = PyDateTime_GET_MONTH(obj);
+    day = PyDateTime_GET_DAY(obj);
+
+    return Py_BuildValue("[lll]", year, month, day);
```
vstinner: [...] is to build a list, previously you built a tuple. Did you use a list on purpose? If not, please use (...) syntax for Py_BuildValue format.
https://docs.python.org/dev/c-api/arg.html#c.Py_BuildValue

vstinner (approved):
LGTM, good job @nanjekyejoannah! Thanks for enhancements, it's now much better than the first version of the PR :-)

PR #14845, opened 2019-07-18
bpo-36084: Add threading Native ID information to What's New documentation

jaketesler:
This PR includes a minor documentation change. It adds a short blurb to the "What's New in Python 3.8" doc about the new Threading module Native ID functionality (`threading.get_native_id()`, et al.)

If this PR for [`:master`](/python/cpython/tree/master/) is approved, I will create a backport PR for the [`:3.8`](/python/cpython/tree/3.8/) branch.

See #13463 ([bpo-36084](https://bugs.python.org/issue36084)) for more information.

cc @vstinner, @pitrou 

### DIFF
--- a/Doc/whatsnew/3.8.rst
+++ b/Doc/whatsnew/3.8.rst
@@ -770,8 +770,8 @@ threading
 
 * Add a new
   :func:`threading.get_native_id` function and a :data:`~threading.Thread.native_id`
-  attribute to the :class:`threading.Thread` class. These return a thread's native
-  integral identifier (``TID``) provided by the OS (kernel).
+  attribute to the :class:`threading.Thread` class. These return the native
+  integral Thread ID of the current thread assigned by the kernel.
   This feature is only available on certain platforms, see
   :func:`get_native_id <threading.get_native_id>` for more information.
   (Contributed by Jake Tesler in :issue:`36084`.)
--- a/Doc/whatsnew/3.8.rst
+++ b/Doc/whatsnew/3.8.rst
@@ -763,10 +763,18 @@ in a standardized and extensible format, and offers several other benefits.
 threading
 ---------
 
-Add a new :func:`threading.excepthook` function which handles uncaught
-:meth:`threading.Thread.run` exception. It can be overridden to control how
-uncaught :meth:`threading.Thread.run` exceptions are handled.
-(Contributed by Victor Stinner in :issue:`1230540`.)
+* Add a new :func:`threading.excepthook` function which handles uncaught
+  :meth:`threading.Thread.run` exception. It can be overridden to control how
+  uncaught :meth:`threading.Thread.run` exceptions are handled.
+  (Contributed by Victor Stinner in :issue:`1230540`.)
+
+* Add a new
+  :func:`threading.get_native_id` function and a :data:`~threading.Thread.native_id`
+  attribute to the :class:`threading.Thread` class. These return a thread's native
+  integral identifier (``TID``) provided by the OS (kernel).
+  This feature is only available on certain platforms, see
+  :func:`get_native_id <threading.get_native_id>` for more information.
+  (Contributed by Jake Tesler in :issue:`36084`.)
 
 
 tokenize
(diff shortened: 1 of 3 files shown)


Doc/whatsnew/3.8.rst
```diff
@@ -763,10 +763,18 @@ in a standardized and extensible format, and offers several other benefits.
...
-(Contributed by Victor Stinner in :issue:`1230540`.)
+* Add a new :func:`threading.excepthook` function which handles uncaught
+  :meth:`threading.Thread.run` exception. It can be overridden to control how
+  uncaught :meth:`threading.Thread.run` exceptions are handled.
+  (Contributed by Victor Stinner in :issue:`1230540`.)
+
+* Add a new
+  :func:`threading.get_native_id` function and a :data:`~threading.Thread.native_id`
+  attribute to the :class:`threading.Thread` class. These return a thread's native
+  integral identifier (``TID``) provided by the OS (kernel).
```
eryksun: This should be "integer", no? As an adjective, "integral" can qualify a number as an integer, but its primary meaning is something that's necessary or inherent to the whole . So when I first read "integral identifier", I though about what makes an identifier "integral" for half a second (since an identifier is an external label, i.e. a rose by any other name would smell as sweet) before it dawned on me that the intended meaning here is "integer identifier". Google has about 250,000 results for the latter, and only about 500 results for "integral identifier".

jaketesler: The intended word is `integral`, not `integer`. 

As in...
> The `pthread_getunique_np()` function retrieves the unique integral identifier that can be used to identify the thread in some context for application debugging or tracing support.
_Source: [IBM API](https://www.ibm.com/support/knowledgecenter/ssw_ibm_i_72/apis/users_23.htm)_

Or...

> `pthread_getthreadid_np` -- get the calling thread's	integral ID
_Source: [FreeBSD Man pages](https://www.freebsd.org/cgi/man.cgi?query=pthread_getthreadid_np)_

eryksun: This usage is unusual and awkward compared to the far more common (500 to 1) phrasing "integer identifier". Pointing me to man pages that use the same phrasing doesn't change the way I feel about it. But I didn't expect push back, and it's not worth arguing over, not to me. It's all good.

jaketesler: Sorry about any confusion – perhaps I was aiming for by-the-books accuracy over strict readability. Part of the initial reason for this wording was to distinguish between it and the Python thread `ident` attribute, which is a completely different, Python-specific identifier. For the previous PRs implementing this feature (#13463, et al.), we had been attempting to keep consistency with the broader definition of what the attribute really represents (and for this PR, attempting to remain consistent with the language in the Python documentation we'd already written). 

Specifically, regarding your comments, I'd venture to say that the `native ID` _is_ integral to a thread, insofar as a `PID` is integral to a system process – a handle for the kernel to facilitate scheduling. 

But in any case, while I agree that the language here might not be as immediately readable as `integer`, I'd prefer to keep this language, if nothing else to be explicitly clear about what the value represents. Hope this clears things up!

jaketesler:
I didn't expect the Spanish Inquisition

jaketesler:
^ I did that since @bedevere-bot didn't initially add any labels and I didn't want this small PR to get lost in the fray.

Doc/whatsnew/3.8.rst
vstinner: I would prefer that What's New and https://docs.python.org/dev/library/threading.html#threading.get_native_id use the same wording: either update get_native_id() to also use "return a thread's native integral identifier (``TID``) provided by the OS (kernel)" or update this doc for "Return the native integral Thread ID of the current thread assigned by the kernel".

Misc/NEWS.d/next/Documentation/2019-07-24-22-46-04.bpo-36084.2rmo4T.rst
```diff
@@ -0,0 +1 @@
+Add Threading Native ID info to What's New doc
```
vstinner: Please remove this NEWS entry, it's not needed. I added the "skip news" label.

jaketesler:
@vstinner Updated the 3.8 doc per your suggestion.

PR #14849, opened 2019-07-19
bpo-37610: improve Editors & IDE in using documentation

aldwinaldwin:
https://bugs.python.org/issue37610


aldwinaldwin:
Sorry, i made some mistake with previous PR ... will try to fix.

PR #14854, opened 2019-07-19
bpo-37624: random.choices has unexpected behavior with negative weights

aldwinaldwin:
https://bugs.python.org/issue37624


rhettinger:
Sorry, but I'm going to decline this patch.  It isn't worth adding an extra O(n) step.

PR #14867, opened 2019-07-19
bpo-34160: explain how to deal with attribute order in ElementTree

scoder:
https://bugs.python.org/issue34160

### DIFF
--- a/Doc/library/xml.etree.elementtree.rst
+++ b/Doc/library/xml.etree.elementtree.rst
@@ -938,6 +938,36 @@ Element Objects
      if element is None:
          print("element not found")
 
+   Prior to Python 3.8, the serialisation order of the attributes of Elements
+   was artificially made predictable by sorting the attributes by their name.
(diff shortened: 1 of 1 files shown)


pablogsal (approved):
LGTM

Doc/library/xml.etree.elementtree.rst
```diff
@@ -930,6 +938,36 @@ Element Objects
      if element is None:
          print("element not found")
 
+   Prior to Python 3.8, the serialisation order of the attributes of Elements
```
pablogsal: Maybe adding a reference to the  class? Feel free to ignore this suggestion :)

```suggestion
Prior to Python 3.8, the serialisation order of the attributes of :class:`Element` objects
```

scoder: Thanks for the suggestion, but I think this would further increase the ambiguity whether it's referring to the attributes of the Python object or the XML attributes of the XML element. I'll actually change it to "XML attributes" to make that clear.

PR #14868, opened 2019-07-20
bpo-36161: Use thread-safe function ttyname_r instead of unsafe one ttyname

chibby0ne:
Signed-off-by: Antonio Gutierrez <chibby0ne@gmail.com>

https://bugs.python.org/issue36161

### DIFF
--- a/Misc/NEWS.d/next/Library/2019-07-20-01-17-43.bpo-36161.Fzf-f9.rst
+++ b/Misc/NEWS.d/next/Library/2019-07-20-01-17-43.bpo-36161.Fzf-f9.rst
@@ -1,1 +1,1 @@
-Use ttyname_r instead of ttyname for thread safety
+In :mod:`posix`, use ``ttyname_r`` instead of ``ttyname`` for thread safety.
--- a/Misc/NEWS.d/next/Library/2019-07-20-01-17-43.bpo-36161.Fzf-f9.rst
+++ b/Misc/NEWS.d/next/Library/2019-07-20-01-17-43.bpo-36161.Fzf-f9.rst
@@ -1,0 +1,1 @@
+Use ttyname_r instead of ttyname for thread safety
--- a/Modules/posixmodule.c
+++ b/Modules/posixmodule.c
@@ -2774,13 +2774,24 @@ static PyObject *
 os_ttyname_impl(PyObject *module, int fd)
 /*[clinic end generated code: output=c424d2e9d1cd636a input=9ff5a58b08115c55]*/
 {
-    char *ret;
 
-    ret = ttyname(fd);
-    if (ret == NULL) {
+    long size = sysconf(_SC_TTY_NAME_MAX);
+    if (size == -1) {
         return posix_error();
     }
-    return PyUnicode_DecodeFSDefault(ret);
+    char *buffer = (char *)PyMem_RawMalloc(size);
+    if (buffer == NULL) {
+        return PyErr_NoMemory();
+    }
+    int ret = ttyname_r(fd, buffer, size);
+    if (ret != 0) {
+        PyMem_RawFree(buffer);
+        errno = ret;
+        return posix_error();
+    }
+    PyObject *res = PyUnicode_DecodeFSDefault(buffer);
+    PyMem_RawFree(buffer);
+    return res;
 }
 #endif
 


chibby0ne:
Hi all, I'm a new contributor and I've made a PR taking into account the discussion seen in the issue page.

For the `ttyname_r()`, I've used the buffer length of `MAXPATHLEN`, since I saw it used for several functions inside the same file, like: `os.readlink`, `os.getcw`, among others.

For the `crypt_r()` side, I've added a few unit tests one that raises the error and another that doesn't but could add more if needed.

This is how a raised error would look when passing an invalid salt:
```
>>> crypt.crypt('password', '__', raise_error=True)
  File "/tmp/cpython/Lib/crypt.py", line 73, in crypt
    return _crypt.crypt(word, salt, raise_error)
OSError: [Errno 22] Invalid argument
```

chibby0ne:
I don't understand the error in the MacOS Builder:

```
ERROR: test_invalid_salt_raises_error_false (test.test_crypt.CryptTestCase)
  File "/Users/vsts/agent/2.154.3/work/1/s/Lib/test/test_crypt.py", line 90, in test_invalid_salt_raises_error_false
    res = crypt.crypt('password', '__')
  File "/Users/vsts/agent/2.154.3/work/1/s/Lib/crypt.py", line 73, in crypt
    return _crypt.crypt(word, salt, raise_error)
UnicodeDecodeError: 'utf-8' codec can't decode byte 0xfd in position 3: invalid start byte
```

I don't understand since they are both unicode, and password is clearly utf-8 and the second argument are two `_` underscore chars. Also the 0xfd byte is `'ý'`.

If it were some codec issue wouldn't it fail in other OSes?

Likewise for the other added test.

mangrisano (approved):
Hi and thank you for this PR and for providing tests.

LGTM.

chibby0ne:
@mangrisano Thanks for the approval. Unfortunately the CI is failing in MacOS. If you can, would you give me any pointers as to the cause of the issue? I truly don't understand it :disappointed:

chibby0ne:
Hi @mangrisano, and any others. Just a friendly ping to mention that this is ready to be merged.

chibby0ne:
Hi all,
I need some help understanding why the MacOS test fails with that unicode error, specially since it passes for every other OS. I don't have access to a MacOS environment and have no way of testing this locally.
image

Modules/posixmodule.c
```diff
@@ -2712,13 +2712,14 @@ static PyObject *
 os_ttyname_impl(PyObject *module, int fd)
 /*[clinic end generated code: output=c424d2e9d1cd636a input=9ff5a58b08115c55]*/
 {
-    char *ret;
+    int ret;
+    char buffer[MAXPATHLEN];
```
benjaminp: seems like the right buffer size is actually `TTY_NAME_MAX`

Modules/posixmodule.c
```diff
@@ -2712,13 +2712,14 @@ static PyObject *
...
 
-    ret = ttyname(fd);
-    if (ret == NULL) {
+    ret = ttyname_r(fd, buffer, MAXPATHLEN);
+    if (ret != 0) {
         return posix_error();
```
benjaminp, line 2790: I think you need to set `errno = ret` because the C library isn't required to fill `errno` with `ret`.

Modules/posixmodule.c
```diff
@@ -2712,13 +2712,14 @@ static PyObject *
 os_ttyname_impl(PyObject *module, int fd)
 /*[clinic end generated code: output=c424d2e9d1cd636a input=9ff5a58b08115c55]*/
 {
-    char *ret;
+    int ret;
+    char buffer[MAXPATHLEN];
 
-    ret = ttyname(fd);
-    if (ret == NULL) {
+    ret = ttyname_r(fd, buffer, MAXPATHLEN);
```
benjaminp: Prefer `sizeof(buffer)`

benjaminp (changes requested):
I don't understand these `raise_error` changes. Those don't seem related to changing the underlying library call.

Modules/posixmodule.c
chibby0ne: Correct. This macro wasn't mentioned in the version of the man pages of my local dev machine (i.e: `man ttyname`) but after googling this constant, it *is indeed* mentioned in `man limits.h` which is imported by `Python.h` in the base code.
Good catch! Thanks for the recommendation!

Modules/posixmodule.c
chibby0ne: Indeed only `ttyname` sets it, with `ttyname_r` you have to do it yourself. Thanks!

Modules/posixmodule.c
chibby0ne: Done.

chibby0ne:
> I don't understand these `raise_error` changes. Those don't seem related to changing the underlying library call.

Well according to the discussion in: https://bugs.python.org/issue36161, due to the fact that we are now returning the error code set by `crypt_r` (if any), this is a breaking change, so in order to maintain backwards compatible behavior (where regarless of any error it would return `None`), we add this boolean keyword argument `raise_error.`
But ofc I'm open to any suggestions :smile: 

benjaminp:
> > I don't understand these `raise_error` changes. Those don't seem related to changing the underlying library call.
> 
> Well according to the discussion in: https://bugs.python.org/issue36161, due to the fact that we are now returning the error code set by `crypt_r` (if any), this is a breaking change, so in order to maintain backwards compatible behavior (where regarless of any error it would return `None`), we add this boolean keyword argument `raise_error.`

While I agree it's technically a breaking change, I think not raising errors from a failed cryptographic function is severely buggy behavior. "Errors should never pass silently."

> But ofc I'm open to any suggestions smile

I think errors should always be propagated from `crypt` or you should follow the suggestion of https://bugs.python.org/issue36161#msg351981 and simply drop the `crypt` changes.

chibby0ne:
> While I agree it's technically a breaking change, I think not raising errors from a failed cryptographic function is severely buggy behavior. "Errors should never pass silently."

Wise aphorism! I agree wholeheartedly. 

> I think errors should always be propagated from crypt or you should follow the suggestion of https://bugs.python.org/issue36161#msg351981 and simply drop the crypt changes.

Yeah it looks like this will be eventually deprecated and removed. I've decided to simply raise the error in accordingly, since it helps the user but also we are not over engineering future deprecated code.

chibby0ne:

benjaminp (changes requested):
okay, but now the tests are failng

Modules/posixmodule.c
```diff
@@ -2712,13 +2712,20 @@ static PyObject *
 os_ttyname_impl(PyObject *module, int fd)
 /*[clinic end generated code: output=c424d2e9d1cd636a input=9ff5a58b08115c55]*/
 {
-    char *ret;
 
-    ret = ttyname(fd);
-    if (ret == NULL) {
+#ifndef TTY_NAME_MAX    /* MacOS doesn't define this macro */
```
benjaminp: Huh, maybe it's better to use `sysconf(_SC_TTY_NAME_MAX)` for portability.

Lib/crypt.py
```diff
@@ -79,7 +79,12 @@ def _add_method(name, *args, rounds=None):
     method = _Method(name, *args)
     globals()['METHOD_' + name] = method
     salt = mksalt(method, rounds=rounds)
-    result = crypt('', salt)
+    result = None
+    try:
+        result = crypt('', salt)
+    except OSError:
```
benjaminp: Can you check that the `errno` of this error is `EINVAL`?

I think it would also be good to ensure that at least one method works.

Misc/NEWS.d/next/Library/2019-07-20-01-17-43.bpo-36161.Fzf-f9.rst
```diff
@@ -0,0 +1 @@
+Use thread-safe functions instead of unsafe ones (crypt, ttyname)
```
benjaminp: "Use ttyname_r instead of ttyname and crypt_r instead of crypt for thread safety."

chibby0ne: I forgot to update this, because it was autogenerated from the first commit. Thanks a lot for your review :wink:

Lib/crypt.py
chibby0ne: Good thinking! Done! Thanks a lot :+1:

Modules/posixmodule.c
chibby0ne: I wasn't sure about this because I saw the values were quite different between the ones used in OpenBSD and the ones used in Linux, but I agree is the more portable solution

chibby0ne:

Lib/crypt.py
```diff
@@ -79,7 +80,15 @@ def _add_method(name, *args, rounds=None):
     method = _Method(name, *args)
     globals()['METHOD_' + name] = method
     salt = mksalt(method, rounds=rounds)
-    result = crypt('', salt)
+    result = None
+    try:
+        result = crypt('', salt)
+    except OSError as e:
+        # Not all libc libraries support all encryption methods
```
benjaminp: This is a sentence, so it should have a period.

Lib/crypt.py
```diff
@@ -79,7 +80,15 @@ def _add_method(name, *args, rounds=None):
...
+        if e.errno == errno.EINVAL:
+            return False
+        # If it's a different exception we raise it
+        raise e
```
benjaminp: Just bare `raise` would be sufficient.

Lib/crypt.py
```diff
@@ -79,7 +80,15 @@ def _add_method(name, *args, rounds=None):
...
     salt = mksalt(method, rounds=rounds)
-    result = crypt('', salt)
+    result = None
+    try:
+        result = crypt('', salt)
+    except OSError as e:
+        # Not all libc libraries support all encryption methods
+        if e.errno == errno.EINVAL:
+            return False
+        # If it's a different exception we raise it
```
benjaminp: I don't think we need this comment.

Modules/posixmodule.c
```diff
@@ -2712,13 +2712,20 @@ static PyObject *
...
+#define TTY_NAME_MAX sysconf(_SC_TTY_NAME_MAX)
```
benjaminp: Let's completely abandon the preprocessor and always call `sysconf`. You'll need to check it for an (probably only theoretical) error. You'll also need to read into a dynamically-allocated buffer.

benjaminp (changes requested):
Please rebase or merge master on the next update; GitHub is claiming there's a conflict.

Lib/crypt.py
chibby0ne: Done!

Lib/crypt.py
chibby0ne: ditto

Modules/posixmodule.c
chibby0ne: Coolio. I used `PyMem_RawMalloc` and `PyMem_RawFree` for the buffer, as I saw it used in other parts of the file.

chibby0ne (commented):
> Please rebase or merge master on the next update; GitHub is claiming there's a conflict.

I rebased and also squashed all the commits.

chibby0ne:

Modules/posixmodule.c
```diff
@@ -2774,13 +2774,24 @@ static PyObject *
...
-    char *ret;
 
-    ret = ttyname(fd);
-    if (ret == NULL) {
+    long size = sysconf(_SC_TTY_NAME_MAX);
+    if (size == -1) {
         return posix_error();
     }
-    return PyUnicode_DecodeFSDefault(ret);
+    char *buffer = (char *) PyMem_RawMalloc(size);
```
benjaminp: Nit: no space between the cast and the function call.

Modules/posixmodule.c
```diff
@@ -2774,13 +2774,24 @@ static PyObject *
...
+    if (buffer == NULL) {
+        return PyErr_NoMemory();
+    }
+    int ret = ttyname_r(fd, buffer, size);
+    if (ret != 0) {
+        errno = ret;
+        return posix_error();
```
benjaminp, line 2790: This branch now leaks `buffer`.

Modules/posixmodule.c
```diff
@@ -2774,13 +2774,24 @@ static PyObject *
...
+    }
+    PyObject *res;
+    res = PyUnicode_DecodeFSDefault(buffer);
```
benjaminp: Fold these two lines into one:
```
PyObject *res = PyUnicode_DecodeFSDefault(buffer);
```

Misc/NEWS.d/next/Library/2019-07-20-01-17-43.bpo-36161.Fzf-f9.rst
```diff
@@ -0,0 +1 @@
+For thread safetyy, use ``ttyname_r(3)`` instead of ``ttyname(3)`` and ``crypt_r(3)`` instead of ``crypt(3)``.
```
benjaminp: This is not correct, though, since you're not using `crypt_r` any more. :P

(I should have seen that last time I looked at the message.)

benjaminp (changes requested):
There are really two logical changes in this PR: checking the error of `crypt` and using `ttyname_r`. It would make sense to split them into separate PRs I think.

Modules/posixmodule.c
chibby0ne: Done!

Modules/posixmodule.c
chibby0ne: Good catch! Thanks ;)

Modules/posixmodule.c
chibby0ne: Done!

Misc/NEWS.d/next/Library/2019-07-20-01-17-43.bpo-36161.Fzf-f9.rst
chibby0ne: In that case will only leave this PR for the `ttyname_r`, and then create another one for just the error checking of `crypt`, and will change the news message accordingly.

chibby0ne:

chibby0ne:
@benjaminp What do you think I should name the other PR? AFAIK the PR branch should match the name of the issue. Should I name `fix-issue-36161_part2`?

chibby0ne:
Hi @benjaminp, the error check for crypt/crypt_r was moved to: https://github.com/python/cpython/pull/16599

benjaminp:
`bazel test` [sets `TZ=UTC`](https://docs.bazel.build/versions/master/test-encyclopedia.html#initial-conditions), which probably accounts for the difference.

benjaminp:
Thank you for the PR.

PR #14870, opened 2019-07-20
bpo-37627: Initialize IDLE Custom Run dialog with previous entries

nsiregar:
Repeat the command line arguments most recently entered before so the user can edit them.

https://bugs.python.org/issue37627


terryjreedy:
Please be very careful with bpo numbers when submitting.  To fix, one must go the wrong issue and remove the misplaced link, then change the number in the title and 3 places in the comment.  (This should be in the devguide.)

terryjreedy (changes requested):
The tracker issue and this PR are about two different subjects: the Run menu and persisting user command line entries.  I will want these split, but can't now say what change to make.

So don't do anything immediately.

rhettinger:
Can we get this done before 3.8 goes final?  The user testing indicated that both sub-issues were problematic.  That said, I'm happy to have the feature at all.  It was a nice improvment for developing command-line scripts using IDLE.

terryjreedy:
I want to split the menu change, for which I have an alternative proposal, from the argument persistence, which I was going to merge by tomorrow.  I did not realize that 'delete file' meant 'delete the whole file' as opposed to the changes to the file, so I have to patch this PR or write another.

terryjreedy (approved):
The fact that assigning a list such as `['a', 'b=3']` to the StringVar textvariable of an Entry results in a display of `'a b=3'` is a bit of undocumented magic that special-cases lists and tuples.  I made cli_args always be a list and documented and tested the behavior we are depending on.

 I expect to merge this tomorrow.  I don't think the fact that the previous args entry reappears needs a doc change.  The runscript module needs some real automated tests, but I leave that for another issue.

Ngalim, welcome to Python development.  You might be our first contributor from Papua.  I plan to add you to our acknowledgments file.  Are you properly alphabetized by your second name, Siregar?

nsiregar:
Yeah, Siregar it is

terryjreedy:
@Mariatta I have re-added 'needs backport to 3.8' twice with no response at all after several minutes

Mariatta:
The last one seemed to work

PR #14872, opened 2019-07-21
Fix typos in docs, comments and test assert messages

minho42:

### DIFF
--- a/Doc/library/ctypes.rst
+++ b/Doc/library/ctypes.rst
@@ -1175,7 +1175,7 @@ Keep in mind that retrieving sub-objects from Structure, Unions, and Arrays
 doesn't *copy* the sub-object, instead it retrieves a wrapper object accessing
 the root-object's underlying buffer.
 
-Another example that may behave different from what one would expect is this::
+Another example that may behave differently from what one would expect is this::
 
    >>> s = c_char_p()
    >>> s.value = b"abc def ghi"
--- a/Lib/test/support/__init__.py
+++ b/Lib/test/support/__init__.py
@@ -2978,7 +2978,7 @@ def fd_count():
     if sys.platform.startswith(('linux', 'freebsd')):
         try:
             names = os.listdir("/proc/self/fd")
-            # Subtract one because listdir() opens internally a file
+            # Subtract one because listdir() internally opens a file
             # descriptor to list the content of the /proc/self/fd/ directory.
             return len(names) - 1
         except FileNotFoundError:
--- a/Lib/test/support/__init__.py
+++ b/Lib/test/support/__init__.py
@@ -2978,7 +2978,7 @@ def fd_count():
     if sys.platform.startswith(('linux', 'freebsd')):
         try:
             names = os.listdir("/proc/self/fd")
-            # Substract one because listdir() opens internally a file
+            # Subtract one because listdir() opens internally a file
             # descriptor to list the content of the /proc/self/fd/ directory.
             return len(names) - 1
         except FileNotFoundError:
--- a/Lib/test/test_dataclasses.py
+++ b/Lib/test/test_dataclasses.py
@@ -3043,11 +3043,11 @@ class TestMakeDataclass(unittest.TestCase):
     def test_non_identifier_field_names(self):
         for field in ['()', 'x,y', '*', '2@3', '', 'little johnny tables']:
             with self.subTest(field=field):
-                with self.assertRaisesRegex(TypeError, 'must be valid identifers'):
+                with self.assertRaisesRegex(TypeError, 'must be valid identifiers'):
                     make_dataclass('C', ['a', field])
-                with self.assertRaisesRegex(TypeError, 'must be valid identifers'):
+                with self.assertRaisesRegex(TypeError, 'must be valid identifiers'):
                     make_dataclass('C', [field])
-                with self.assertRaisesRegex(TypeError, 'must be valid identifers'):
+                with self.assertRaisesRegex(TypeError, 'must be valid identifiers'):
                     make_dataclass('C', [field, 'a'])
 
     def test_underscore_field_names(self):
--- a/Lib/test/test_dataclasses.py
+++ b/Lib/test/test_dataclasses.py
@@ -3043,7 +3043,7 @@ class TestMakeDataclass(unittest.TestCase):
     def test_non_identifier_field_names(self):
         for field in ['()', 'x,y', '*', '2@3', '', 'little johnny tables']:
             with self.subTest(field=field):
-                with self.assertRaisesRegex(TypeError, 'must be valid identifiers'):
+                with self.assertRaisesRegex(TypeError, 'identifiers must be valid'):
                     make_dataclass('C', ['a', field])
                 with self.assertRaisesRegex(TypeError, 'must be valid identifiers'):
                     make_dataclass('C', [field])
--- a/Lib/test/test_dataclasses.py
+++ b/Lib/test/test_dataclasses.py
@@ -3043,7 +3043,7 @@ class TestMakeDataclass(unittest.TestCase):
     def test_non_identifier_field_names(self):
         for field in ['()', 'x,y', '*', '2@3', '', 'little johnny tables']:
             with self.subTest(field=field):
-                with self.assertRaisesRegex(TypeError, 'identifiers must be valid'):
+                with self.assertRaisesRegex(TypeError, 'must be valid identifiers'):
                     make_dataclass('C', ['a', field])
                 with self.assertRaisesRegex(TypeError, 'must be valid identifiers'):
                     make_dataclass('C', [field])
--- a/Lib/test/test_importlib/test_main.py
+++ b/Lib/test/test_importlib/test_main.py
@@ -32,7 +32,7 @@ class BasicTests(fixtures.DistInfoPkg, unittest.TestCase):
 class ImportTests(fixtures.DistInfoPkg, unittest.TestCase):
     def test_import_nonexistent_module(self):
         # Ensure that the MetadataPathFinder does not crash an import of a
-        # non-existant module.
+        # non-existent module.
         with self.assertRaises(ImportError):
             importlib.import_module('does_not_exist')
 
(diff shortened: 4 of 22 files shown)


Doc/library/ctypes.rst
```diff
@@ -1175,7 +1175,7 @@ Keep in mind that retrieving sub-objects from Structure, Unions, and Arrays
 doesn't *copy* the sub-object, instead it retrieves a wrapper object accessing
 the root-object's underlying buffer.
 
-Another example that may behave different from what one would expect is this::
+Another example that may behave differently from what one would expect is this::
```
aeros, line 1178: The "is this" at the end of the sentence looks a bit unneeded.
```suggestion
Another example that may behave differently from what one would expect::
```

Lib/test/support/__init__.py
```diff
@@ -2978,7 +2978,7 @@ def fd_count():
     if sys.platform.startswith(('linux', 'freebsd')):
         try:
             names = os.listdir("/proc/self/fd")
-            # Substract one because listdir() opens internally a file
+            # Subtract one because listdir() opens internally a file
```
aeros:
```suggestion
            # Subtract one because listdir() internally opens a file
```

Lib/test/test_dataclasses.py
```diff
@@ -3043,11 +3043,11 @@ def test_keyword_field_names(self):
     def test_non_identifier_field_names(self):
         for field in ['()', 'x,y', '*', '2@3', '', 'little johnny tables']:
             with self.subTest(field=field):
-                with self.assertRaisesRegex(TypeError, 'must be valid identifers'):
+                with self.assertRaisesRegex(TypeError, 'must be valid identifiers'):
```
aeros, line 3046: Edit: Disregard this suggestion.

Lib/test/test_dataclasses.py
```diff
@@ -3043,11 +3043,11 @@ def test_keyword_field_names(self):
...
                     make_dataclass('C', ['a', field])
-                with self.assertRaisesRegex(TypeError, 'must be valid identifers'):
+                with self.assertRaisesRegex(TypeError, 'must be valid identifiers'):
```
aeros, line 3048: Edit: Disregard this suggestion.

Lib/test/test_dataclasses.py
```diff
@@ -3043,11 +3043,11 @@ def test_keyword_field_names(self):
...
                     make_dataclass('C', [field])
-                with self.assertRaisesRegex(TypeError, 'must be valid identifers'):
+                with self.assertRaisesRegex(TypeError, 'must be valid identifiers'):
```
aeros, line 3050: Edit: Disregard this suggestion.

Lib/test/test_importlib/test_main.py
```diff
@@ -32,7 +32,7 @@ def test_new_style_classes(self):
 class ImportTests(fixtures.DistInfoPkg, unittest.TestCase):
     def test_import_nonexistent_module(self):
         # Ensure that the MetadataPathFinder does not crash an import of a
-        # non-existant module.
+        # non-existent module.
```
aeros:
```suggestion
        # nonexistent module.
```

aeros (approved):
@minho42 Thanks for the contribution! I've seen quite a number of PRs which fix typos, but I've yet to see one involving 22 different files at once. Nicely done. 

I approve of all of the typo fixes, but I noticed a few which could use some minor improvements.

Doc/library/ctypes.rst
terryjreedy: This suggestion is about graceful wording, which leads to nit-picking arguments.  Let us let this be about uncontroversial spelling changes.

Lib/test/support/__init__.py
terryjreedy: @minho42 Even though I could edit the file and commit the same change, I cannot commit 'suggestions' to your PR.  Please see if you can.  I am suggesting accepting this one because it is simple and uncontroversial to me.

Lib/test/test_importlib/test_main.py
terryjreedy: Accept this, as it is a spelling correction correction.  I looked up 'non-existent versus nonexistent' and the hyphen used to be a British spelling, but is not now, and we use American spellings.

terryjreedy (commented):
I have verified that all but 1 of the corrections is correct and that that 1 is corrected in the comment.  That should be enough review of the corrections themselves.  I approve of this given that one correction.  I consider the one wording change I approved of to be optional.

'minho': I am willing to merge this, but I first need to know if you intend to fix up the merge conflicts I expect if we try to backport to 3.8, or 3.7.  I expect that at least one misspelling is new in master.

Doc/library/ctypes.rst
aeros: Yeah that's definitely fair, the phrasing just caught my attention. Feel free to disregard this suggestion.

minho42: I will disregard this suggestion then.

Lib/test/test_dataclasses.py
ericvsmith: I think `must be valid identifiers` is better. That wording says that the text is not a a valid identifier. `identifiers must be valid` says that it's an identifier, but not a "valid" one, which doesn't mean anything.

Lib/test/test_importlib/test_main.py
ericvsmith: Agreed that "nonexistent" is better.

Lib/test/test_dataclasses.py
aeros: Nevermind, this can be disregarded as well then.

minho42:
@terryjreedy
Are you able to merge this?
Does that mean you are going to fix the merge conflicts?

I'd like to resolve the merge conflicts myself but I won't be able to use my laptop for the next few days.

Lib/test/test_dataclasses.py
aeros: @ericvsmith Thanks for the feedback. After looking more closely at the context, the current wording is more correct than my suggestion. `must be valid identifiers` looked incorrect when I was initially reviewing it. There might be a way to improve it, but `identifiers must be valid` would be incorrect.

Plus, as @terryjreedy mentioned in one of my other suggestions, it's probably best for this to primarily focus on uncontroversial typos rather than focus on the wording.

minho42:
@aeros167
Thanks for the suggestions.

aeros167:
@minho42
>Are you able to merge this?
>Does that mean you are going to fix the merge conflicts?

Unless I'm misunderstanding what terry said, the merge conflicts only need to be resolved if he adds the backport labels so that these changes are included in the 3.8 and 3.7 docs. If it's only merged into 3.9 without a backport, the merge conflicts shouldn't be an issue. 

If all of the deployment tests are passing and there's no indication otherwise, this should be able to get auto-merged into 3.9 without issues.

aeros167:
@minho42 Undo the following commit:
https://github.com/python/cpython/pull/14872/commits/d91fb725cfca9f0cf81950b8269d70b17f7150df

minho42:
@aeros167
Then I'm happy to let this merged into 3.9 without backporting!

minho42:
@aeros167 
Changed back.

terryjreedy (approved):
Two doc build tests failed with what seems like a false positive
```
Warning, treated as error:
[distutils/examples:267] "`" found in "This is the description of the ``foobar`` package."
```
This makes no sense to me because this PR does not touch this doc and because I think the backticks are normal and not an error or even 'suspcious'.

I moved the idlelib fix to a separate PR, #14879, to make sure it gets immediately backported.  I keep help.py synchronized across active versions.  This change will trigger a test there and a retest here.

terryjreedy:
Thank you for the corrections.  I was correct that backport would need fixup.  I am done with this unless one of you wants to do them.

aeros167:
@terryjreedy I wouldn't mind trying to help work through the merge conflicts, but I don't think there's way for me to directly edit the PR.

terryjreedy:
Right, you prepare a new pr, much as the bot does, using the cherry_picker app on pypi, with the command line given above, except that you continue where the bot quit.    To install for recent python,
`python -m pip install cherry_picker`
There may be some doc on the site.  Or 
`python -m cherry_picker -h`
for help.  (I use `python - m` because `cherry_picker -h` may or may not work, depending on your install.)
It should be educational, and more requiring of attention to detail than really hard.  This should be a good pr to practice with.

You can use git cherry-pick directly, instead, but cherry_picker will do the routine parts, including pushing to github and making a PR.

minho42:
Thanks @aeros167 @terryjreedy

aeros167:
@terryjreedy Sounds good, I'll give it a try then. I'm interested in the new GitHub triager role, so it would be good to learn how to manually (with `cherry_picker`) reserve merge conflicts.  If it works out well, I'll also work on the manual backport next.

When opening the new PR, I should name it something using a similar format to the PR you did, but with 3.8 in brackets?:
`[3.8] Fix typos found by Min Ho Kim`?

Also, I'm assuming that I should specifically direct the PR at the branch `python:3.8` instead of the usual `python:master`. Apologies if any of this is fairly obvious, I'm not as familiar with the manual back porting process. I'll make a comment in here linking to it when it's ready.

aeros167:
I was able to resolve the conflicts and create the backport with `cherry_picker`. It looks like all of the tests are passing, let me know if there's any issues: https://github.com/python/cpython/pull/14900. Should I attempt to do to the same for 3.7 as well?

terryjreedy:
In response to `cherry_picker 96e12d5f4f3c5a20986566038ee763dff3c228a1 3.8`, cherry_picker checks out permanent branch 3.8, creates (checkout -b) a branch based on that (of whatever name, not important), and then applies the indicated commit diff.  If there are no merge conflicts, it proceeds.  Otherwise it stops and tells you to fix them.  It copies the original title and adds [3.8] (in this case).  When you tell it to continue, it commits with the original message and adds the boilerplate.  You saw what you got.

To do 3.7, change 3.8 to 3.7 and leave the commit-to-master hash as is.  The conventions here is that all backports are from the original commit to master.  A git expert (not me ;-) could fake it by manually applying the (in this case) reduced 3.8 diff, fixing any additional conflicts, and then adding the comment
```
Original PR: #14900 
(cherry picked from commit 96e12d5)
```
as long as the result was the same as if that were done.  

These occasional typo fixes are about the only PRs with such increasing conflicts.

aeros167:
Created backport for 3.7: GH-14901. There were significantly more conflicts with this one, and several of the sections that changes were made to did not exist in 3.7.

PR #14874, opened 2019-07-21
bpo-37616: Fix incorrect zip path for 3.10 prep

nsiregar:
[3.10 prep] fix incorrect zip path

https://bugs.python.org/issue37616

### DIFF
--- a/Misc/NEWS.d/next/Core and Builtins/2019-07-21-11-11-58.bpo-37616.0pEhuj.rst
+++ b/Misc/NEWS.d/next/Core and Builtins/2019-07-21-11-11-58.bpo-37616.0pEhuj.rst
@@ -1,0 +1,1 @@
+Update `getpath` detect zip path in :mod:`sys.path`
--- a/Misc/NEWS.d/next/Core and Builtins/2019-07-21-11-11-58.bpo-37616.0pEhuj.rst
+++ b/Misc/NEWS.d/next/Core and Builtins/2019-07-21-11-11-58.bpo-37616.0pEhuj.rst
@@ -1,1 +1,2 @@
-Update `getpath` detect zip path in :mod:`sys.path`
+Update ``Module/getpath.c`` read python zip path if python
+version size more than 2 digit.
--- a/Modules/getpath.c
+++ b/Modules/getpath.c
@@ -1284,7 +1284,7 @@ calculate_read_pyenv(PyCalculatePath *calculate)
 static PyStatus
 calculate_zip_path(PyCalculatePath *calculate)
 {
-    const wchar_t *lib_python = L"lib/python00.zip";
+    const wchar_t *lib_python = L"lib/python000.zip";
 
     if (calculate->prefix_found > 0) {
         /* Use the reduced prefix returned by Py_GetPrefix()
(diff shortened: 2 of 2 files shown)


Misc/NEWS.d/next/Core and Builtins/2019-07-21-11-11-58.bpo-37616.0pEhuj.rst
```diff
@@ -0,0 +1 @@
+Update `getpath` detect zip path in :mod:`sys.path`
```
ZackerySpytz: Please use proper sentences in news entries.

I believe using single backquotes around "getpath" is incorrect (as per https://devguide.python.org/documenting). Also, I think the full file path should be used (instead of just "getpath").

`sys.path` is not a module.

Modules/getpath.c
```diff
@@ -1016,15 +1016,21 @@ calculate_zip_path(PyCalculatePath *calculate, const wchar_t *prefix)
             return PATHLEN_ERR();
         }
     }
-    status = joinpath(calculate->zip_path, L"lib/python00.zip", zip_path_len);
+    status = joinpath(calculate->zip_path, L"lib/python000.zip", zip_path_len);
```
ZackerySpytz: I believe this change will cause backward compatibility issues.

Modules/getpath.c
```diff
@@ -1016,15 +1016,21 @@ calculate_zip_path(PyCalculatePath *calculate, const wchar_t *prefix)
...
 
-    /* Replace "00" with version */
+    /* Replace "000" with version */
     size_t bufsz = wcslen(calculate->zip_path);
-    calculate->zip_path[bufsz - 6] = VERSION[0];
-    calculate->zip_path[bufsz - 5] = VERSION[2];
+    calculate->zip_path[bufsz - 7] = VERSION[0];
+    calculate->zip_path[bufsz - 6] = VERSION[2];
+    if (sizeof(VERSION) == 5) {
+      calculate->zip_path[bufsz - 5] = VERSION[3];
```
ZackerySpytz: Please use four-space indentation.

Modules/getpath.c
```diff
@@ -1016,15 +1016,21 @@ calculate_zip_path(PyCalculatePath *calculate, const wchar_t *prefix)
...
+    }
+    else {
+      memmove(&calculate->zip_path[bufsz - 5], &calculate->zip_path[bufsz - 4], bufsz);
```
ZackerySpytz: Limit lines to 79 chars ([PEP 7](https://www.python.org/dev/peps/pep-0007/#code-lay-out)).

ZackerySpytz (commented):
I'm not sure if this fix is correct.

nsiregar:
Do I need to change the algorithm just to replace `00` in `python00.zip` with any concatenated version number?

Modules/getpath.c
ronaldoussoren: > I believe this change will cause backward compatibility issues.

Why do you think that? The change looks harmless to me, especially with the code below that replaces the three zeros by the version information. That code shortens the file name when the version information only requires two characters.

ZackerySpytz: > That code shortens the file name when the version information only requires two characters.

You're right. Disregard my comment.

csabella:
@nsiregar, please resolve the merge conflict.  Thank you!

nsiregar:
@csabella I've resolved the conflict

Misc/NEWS.d/next/Core and Builtins/2019-07-21-11-11-58.bpo-37616.0pEhuj.rst
```diff
@@ -0,0 +1,2 @@
+Update ``Module/getpath.c`` read python zip path if python
+version size more than 2 digit.
```
brettcannon, line 2:
```suggestion
Update ``Module/getpath.c`` to read the stdlib from a zip file when the minor version is more than a single digit.
```

brettcannon:
The `memmove` doesn't seem like the right solution here. Why not have two template file paths and then use the appropriate one based on the number of digits required? Otherwise you might as well use `malloc()` and build the string from scratch.

nsiregar:
working on it

ambv:
Since this wasn't merged in time, we went with https://github.com/python/cpython/commit/b4d4aef8433da2657c8d80207686124d15d32054

PR #14878, opened 2019-07-21
bpo-37642: Update max and min offset in datetime module

nsiregar:
Update max and min offset for `datetime.timezone` to closer 24:00

https://bugs.python.org/issue37642

### DIFF
--- a/Lib/datetime.py
+++ b/Lib/datetime.py
@@ -2269,7 +2269,7 @@ class timezone(tzinfo):
         raise TypeError("fromutc() argument must be a datetime instance"
                         " or None")
 
-    _maxoffset = timedelta(hours=23, minutes=59)
+    _maxoffset = timedelta(hours=24, microseconds=-1)
     _minoffset = -_maxoffset
 
     @staticmethod
@@ -2293,8 +2293,11 @@ class timezone(tzinfo):
         return f'UTC{sign}{hours:02d}:{minutes:02d}'
 
 timezone.utc = timezone._create(timedelta(0))
-timezone.min = timezone._create(timezone._minoffset)
-timezone.max = timezone._create(timezone._maxoffset)
+# bpo-37642: These attributes are rounded to the nearest minute for backwards
+# compatibility, even though the constructor will accept a wider range of
+# values. This may change in the future.
+timezone.min = timezone._create(-timedelta(hours=23, minutes=59))
+timezone.max = timezone._create(timedelta(hours=23, minutes=59))
 _EPOCH = datetime(1970, 1, 1, tzinfo=timezone.utc)
 
 # Some time zone algebra.  For a datetime x, let
--- a/Lib/test/datetimetester.py
+++ b/Lib/test/datetimetester.py
@@ -388,6 +388,31 @@ class TestTimeZone(unittest.TestCase):
         tz_copy = copy.deepcopy(tz)
         self.assertIs(tz_copy, tz)
 
+    def test_offset_boundaries(self):
+        # Test timedeltas close to the boundaries
+        time_deltas = [
+            timedelta(hours=23, minutes=59),
+            timedelta(hours=23, minutes=59, seconds=59),
+            timedelta(hours=23, minutes=59, seconds=59, microseconds=999999),
+        ]
+        time_deltas.extend([-delta for delta in time_deltas])
+
+        for delta in time_deltas:
+            with self.subTest(test_type='good', delta=delta):
+                timezone(delta)
+
+        # Test timedeltas on and outside the boundaries
+        bad_time_deltas = [
+            timedelta(hours=24),
+            timedelta(hours=24, microseconds=1),
+        ]
+        bad_time_deltas.extend([-delta for delta in bad_time_deltas])
+
+        for delta in bad_time_deltas:
+            with self.subTest(test_type='bad', delta=delta):
+                with self.assertRaises(ValueError):
+                    timezone(delta)
+
 
 #############################################################################
 # Base class for testing a particular aspect of timedelta, time, date and
--- a/Misc/ACKS
+++ b/Misc/ACKS
@@ -1881,3 +1881,4 @@ Aleksandr Balezin
 Robert Leenders
 Tim Hopper
 Dan Lidral-Porter
+Ngalim Siregar
--- a/Misc/NEWS.d/next/Library/2019-07-21-20-59-31.bpo-37642.L61Bvy.rst
+++ b/Misc/NEWS.d/next/Library/2019-07-21-20-59-31.bpo-37642.L61Bvy.rst
@@ -1,0 +1,3 @@
+Allowed the pure Python implementation of :class:`datetime.timezone` to represent
+sub-minute offsets close to minimum and maximum boundaries, specifically in the
+ranges (23:59, 24:00) and (-23:59, 24:00). Patch by Ngalim Siregar
--- a/Modules/_datetimemodule.c
+++ b/Modules/_datetimemodule.c
@@ -1099,7 +1099,9 @@ new_timezone(PyObject *offset, PyObject *name)
         Py_INCREF(PyDateTime_TimeZone_UTC);
         return PyDateTime_TimeZone_UTC;
     }
-    if ((GET_TD_DAYS(offset) == -1 && GET_TD_SECONDS(offset) == 0) ||
+    if ((GET_TD_DAYS(offset) == -1 &&
+            GET_TD_SECONDS(offset) == 0 &&
+            GET_TD_MICROSECONDS(offset) < 1) ||
         GET_TD_DAYS(offset) < -1 || GET_TD_DAYS(offset) >= 1) {
         PyErr_Format(PyExc_ValueError, "offset must be a timedelta"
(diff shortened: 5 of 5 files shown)


Lib/datetime.py
```diff
@@ -2269,7 +2269,7 @@ def fromutc(self, dt):
         raise TypeError("fromutc() argument must be a datetime instance"
                         " or None")
 
-    _maxoffset = timedelta(hours=23, minutes=59)
+    _maxoffset = timedelta(seconds=86339, microseconds=999999)
```
joernheissler: should be `86399`. Or better to use hours, minutes and seconds?

Lib/test/datetimetester.py
```diff
@@ -390,6 +390,14 @@ def test_deepcopy(self):
         tz_copy = copy.deepcopy(tz)
         self.assertIs(tz_copy, tz)
 
+    def test_timezone_offset_close_twenty_four(self):
```
joernheissler: Does that code actually test the python implementation? To do my local experiments, I had to comment out the lines that import the c module.

pganssle: Yeah, `datetimetester.py` is in named weirdly because the tests are actually *run* by `test_datetime.py`, which will run these tests using both the pure python `datetime` module and the C `datetime` module.

Lib/datetime.py
pganssle:
```suggestion
    _maxoffset = timedelta(hours=23, minutes=59, seconds=59, microseconds=999999)
```

I prefer the full version to make the intent clearer and to avoid further mistakes.

Lib/test/datetimetester.py
```diff
@@ -390,6 +390,14 @@ def test_deepcopy(self):
...
+        tz = timezone(timedelta(seconds=86399))
+        self.assertIsInstance(tz, timezone)
+        with self.assertRaises(ValueError):
+            timezone(timedelta(hours=24))
+            timezone(timedelta(seconds=86400))
+            timezone(timedelta(seconds=86399, microseconds=1000000))
```
pganssle: This normalizes to the same thing as the previous one.

Lib/test/datetimetester.py
pganssle: Can you make this a more generic, parametrized test (use `self.subTest`)? I believe we should test both the negative and positive bounds (e.g. `-23:59:59.999999` and `+23:59:59.999999`).

Also, please rename the test to something like `test_offset_boundaries`.

Misc/NEWS.d/next/Library/2019-07-21-20-59-31.bpo-37642.L61Bvy.rst
```diff
@@ -0,0 +1 @@
+Update :mod:`datetime.timezone` to get closer to 24:00
```
pganssle: I think it will be very unclear what this change is to someone reading it without having read the bug report. This specifically allows the *pure Python* implementation `datetime.timezone` to represent sub-minute offsets close to 24 hours.

I suggest wording like:

"Allowed the pure Python implementation of ``:class:`datetime.timezone` `` to represent sub-minute offsets close to minimum and maximum boundaries, specifically in the ranges (23:59, 24:00) and (-23:59, -24:00). Patch by Ngalim Siregar."

You may also want to add yourself to the [Misc/Acks](https://github.com/python/cpython/blob/master/Misc/ACKS) file as well.

pganssle (changes requested):

nsiregar:

nsiregar:
Looks like there is different implementation in max and min offset in `_datetimemodule.c`.

min offset in C is https://github.com/python/cpython/blob/35b87e6001bd991f625abe305951c77ddeb9a9c5/Modules/_datetimemodule.c#L6487 
max offset in C is https://github.com/python/cpython/blob/35b87e6001bd991f625abe305951c77ddeb9a9c5/Modules/_datetimemodule.c#L6496

Should I also update these?

joernheissler:
> Looks like there is different implementation in max and min offset in `_datetimemodule.c`.
> …
> Should I also update these?

I think you should.

A shorter way to write the new limits would be `timedelta(days=1, microseconds=-1)` but I'm not sure about readability.

nsiregar:
i was plan to use `new_delta(-1, 0, 1,1)` for negative offset, is it okay?

pganssle:
> Looks like there is different implementation in max and min offset in `_datetimemodule.c`.

I commented recently [on the bpo issue](https://bugs.python.org/issue37642#msg348298), but I think we should keep `timezone.min` and `timezone.max` the same for this PR, which would allow us to backport it to Python 3.7 and 3.8. We should probably have a separate discussion about the scope and convenience of `timezone.min` and `timezone.max`, because I suspect that changing those to a value microsecond offsets could cause a lot of needless churn.

Lib/datetime.py
```diff
@@ -2269,8 +2269,8 @@ def fromutc(self, dt):
         raise TypeError("fromutc() argument must be a datetime instance"
                         " or None")
 
-    _maxoffset = timedelta(hours=23, minutes=59)
-    _minoffset = -_maxoffset
+    _maxoffset = timedelta(hours=23, minutes=59, seconds=59, microseconds=999999)
+    _minoffset = - timedelta(hours=23, minutes=59, seconds=59, microseconds=999999)
```
pganssle: Why change the second line? By defining `_minoffset` in terms of `_maxoffset`, you only have to change `_maxoffset`.

Lib/datetime.py
```diff
@@ -2269,8 +2269,8 @@ def fromutc(self, dt):
         raise TypeError("fromutc() argument must be a datetime instance"
                         " or None")
 
-    _maxoffset = timedelta(hours=23, minutes=59)
-    _minoffset = -_maxoffset
+    _maxoffset = timedelta(hours=23, minutes=59, seconds=59, microseconds=999999)
```
pganssle:
```suggestion
    _maxoffset = timedelta(hours=23, minutes=59, seconds=59,
                           microseconds=999999)
```

Line is now too long. Either use this suggestion or @joernheissler's suggestion of `timedelta(hours=24, microseconds=-1)`. I think there are upsides and downsides to both ways of expressing this concept, so I'll leave it in your hands which you like better.

nsiregar:

pganssle (changes requested):
Ugh, I wrote this review yesterday and didn't realize I never submitted it.

Per [this comment](https://github.com/python/cpython/pull/14878#issuecomment-513805238), I think we need to keep `timezone.min` and `timezone.max` the same for now. I would say to move the changes for `timezone.min` and `timezone.max` into another branch, but I think we need a new BPO issue for it because I think that needs more discussion.

Lib/test/datetimetester.py
```diff
@@ -266,7 +266,8 @@ def test_repr(self):
             self.assertEqual(tz, eval(tzrep))
 
     def test_class_members(self):
-        limit = timedelta(hours=23, minutes=59)
+        limit = timedelta(hours=23, minutes=59,
```
pganssle: In addition to my other comments, I guess this comment got lost. Please leave this value intact as `timedelta(hours=23, minutes=59)` and leave `datetime.py` intact. You can do this by changing [the definition of `timezone.(min|max)`](https://github.com/python/cpython/blob/717cccfbac474de580263ae93a1c75f37c1033ed/Lib/datetime.py#L2297) like this:

```python
timezone.utc = timezone._create(timedelta(0))
# bpo-37642: These attributes are rounded to the nearest minute for backwards
# compatibility, even though the constructor will accept a wider range of
# values. This may change in the future.
timezone.min = timezone._create(-timedelta(hours=23, minutes=59))
timezone.max = timezone._create(timedelta(hours=23, minutes=59))
_EPOCH = datetime(1970, 1, 1, tzinfo=timezone.utc)
```

Note the comment. If you want you can put a similar comment in `_datetimemodule.c`.

Lib/test/datetimetester.py
```diff
@@ -279,7 +280,7 @@ def test_constructor(self):
             tz = timezone(subminute)
             self.assertNotEqual(tz.utcoffset(None) % timedelta(minutes=1), 0)
         # invalid offsets
-        for invalid in [timedelta(1, 1), timedelta(1)]:
+        for invalid in [timedelta(1, 1), timedelta(days=1, microseconds=1)]:
```
pganssle: Why does this need to be changed?

pganssle (changes requested):

Lib/test/datetimetester.py
nsiregar: Sorry, looks like i forgot to restore it

nsiregar:

pganssle (approved):
I pushed one change refactoring the tests a bit, I think this is ready to merge.

Thank you @nsiregar and @joernheissler for your contributions!

Misc/ACKS
```diff
@@ -1881,3 +1881,4 @@ Aleksandr Balezin
 Robert Leenders
 Tim Hopper
 Dan Lidral-Porter
+Ngalim Siregar
```
merwok, line 1884: Please note that we don’t append to this file, but keep it sorted alphabetically by author last name.

PR #14879, opened 2019-07-21
Fix typo found by Min ho Kim

terryjreedy:


terryjreedy:
[distutils/examples:267] "`" found in "This is the description of the ``foobar`` package."
https://travis-ci.org/python/cpython/builds/561701846?utm_source=github_status&utm_medium=notification
https://travis-ci.org/python/cpython/jobs/561701848

PR #14880, opened 2019-07-21
gh-42550: Add 'Expect: 100-Continue' support to httplib

tbartlett0:
Previously, `http.client` would always send content body immediately and ignore any 100 responses. This change makes `HTTPClient.request()` wait for a Continue response if the `Expect: 100-Continue` header is set, and adds a parameter to `HTTPClient.getresponse()` that will cause it to return 100 responses instead of eating them.

https://bugs.python.org/issue1346874

* Issue: gh-42550


tbartlett0:
This PR is intended to be backwards-compatible, but I have identified two potential edge cases:
- clients that currently set `Expect: 100-Continue` with a broken server that ignores the header: This will introduce a delay of _expect_timeout_ seconds to each request. However, this can be fixed by a) not sending the header or b) setting _expect_timeout_ to 0.
- anyone who redefines `HTTPClient.response_class`: this is likely to be a breaking change, as a new kwarg is used on `HTTPResponse.begin()`. (This seems like a fairly rare situation, as it's not exposed as an `__init__` param, but chances are that someone, somewhere has done it...)

emnoor-reef:
Just bumping. Is this PR on consideration for getting merged? Or, are there any other plans for #42550 ? I am encountering this problem at work.

IngridMorstrad:
Hi, @tbartlett0 
Is this PR still being worked on? Or can someone els e(me) take it over?

AA-Turner:
Tim would need to (re-)sign the CLA.

tbartlett0:
Hi, I haven't touched this since it was submitted, since there seemed to be limited interest and I ended up working around it in other ways. If someone else wants to take it on I have no problem with that (either using my code or rewriting their own, there might be a better way to achieve it)

IngridMorstrad:
I don't have permissions here, so created a new PR: https://github.com/python/cpython/pull/133276

PR #14887, opened 2019-07-21
bpo-37644: Fix susp-ignored.csv to not break CI

terryjreedy:
Remove obsolete line number.

https://bugs.python.org/issue37644


terryjreedy:
Ned Deily already merged a change.

PR #14896, opened 2019-07-21
bpo-36338: Reject hostname with [ at position > 0

jpic:
Before:

    >>> urlparse('http://good.com[malicious.com]/aoeu').hostname
    'malicious.com'

After:

    >>> urlparse('http://good.com[malicious.com]/aoeu')
    ValueError: Invalid IPv6 URL

https://bugs.python.org/issue36338

### DIFF
--- a/Lib/urllib/parse.py
+++ b/Lib/urllib/parse.py
@@ -436,7 +436,8 @@ def urlsplit(url, scheme='', allow_fragments=True):
             if url[:2] == '//':
                 netloc, url = _splitnetloc(url, 2)
                 if (('[' in netloc and ']' not in netloc) or
-                        (']' in netloc and '[' not in netloc)):
+                        (']' in netloc and '[' not in netloc) or
+                        ('[' in netloc and netloc.index('[') != 0)):
                     raise ValueError("Invalid IPv6 URL")
             if allow_fragments and '#' in url:
                 url, fragment = url.split('#', 1)
--- a/Lib/test/test_urlparse.py
+++ b/Lib/test/test_urlparse.py
@@ -531,6 +531,19 @@ class UrlParseTestCase(unittest.TestCase):
         self.assertEqual(p.hostname, b"fe80::822a:a8ff:fe49:470c%tESt")
         self.assertEqual(p.netloc, b'[FE80::822a:a8ff:fe49:470c%tESt]:1234')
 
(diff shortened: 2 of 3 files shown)


mangrisano (approved):
LGTM. Thank you for providing the test as well.

jpic:
Any time ! Will try to keep on to have always one one patch at the time, focusing on security issues at first ;)
Have fun !

CuriousLearner (approved):
Looks good to me! :taco:

jpic:
Thanks for the kind words, looking forward to review prior to starting on another ticket ;)

Lib/urllib/parse.py
```diff
@@ -436,7 +436,8 @@ def urlsplit(url, scheme='', allow_fragments=True):
             if url[:2] == '//':
                 netloc, url = _splitnetloc(url, 2)
                 if (('[' in netloc and ']' not in netloc) or
-                        (']' in netloc and '[' not in netloc)):
+                        (']' in netloc and '[' not in netloc) or
+                        ('[' in netloc and netloc.index('[') != 0)):
                     raise ValueError("Invalid IPv6 URL")
```
vstinner, line 441: _splitnetloc() is called 2 times and the same code to validate the IPv6 address is duplicated, whereas you only fix one place. IMHO it would be better to move the check into _splitnetloc().

vstinner (changes requested):
Additional checks are very incomplete. IMHO the urllib.parser is a weak implementation of RFC 2396 and RFC 2732.

For example, I don't think such URLs are valid according to the RFCs:

```
>>> urlparse('http://google.com::::80/')
ParseResult(scheme='http', netloc='google.com::::80', path='/', params='', query='', fragment='')
>>> urlparse('http://[::1]/')
ParseResult(scheme='http', netloc='[::1]', path='/', params='', query='', fragment='')
>>> urlparse('http://[[::1]]/')
ParseResult(scheme='http', netloc='[[::1]]', path='/', params='', query='', fragment='')
>>> urlparse('http://[::1][]/')
ParseResult(scheme='http', netloc='[::1][]', path='/', params='', query='', fragment='')
```

IMHO the code should be rewritten to better respect the RFCs.

vstinner:
I proposed a stricter change: PR #16780.

jpic:
Thanks for your comments @vstinner, apparently #16780 superseeds this one, nice approach BTW.

PR #14912, opened 2019-07-22
bpo-37534: Adding ability to add Standalone Document Declaration when generating XML documents

henryh9n:
### When using minidom module to generate XML documents the ability to add Standalone Document Declaration is added.

In case it's set to None, which is by default, the declaration is omitted, otherwise it is set to 'yes' when a True value is passed and 'no' when False is passed.
All the changes are made to generate a document in compliance with Extensible Markup Language (XML) 1.0 (Fifth Edition) W3C Recommendation (available here: [https://www.w3.org/TR/xml/#sec-prolog-dtd](https://www.w3.org/TR/xml/#sec-prolog-dtd)).
As minidom module is missing documentation for document generator, no changes are made for current functionality. Documentation may be added separately later.

Resolves: [37534](https://bugs.python.org/issue37534)

https://bugs.python.org/issue37534


ZackerySpytz:
Why are there no unit tests?

hharutyunyan:
> Why are there no unit tests?

Forgot to include those. :D See I'm all new in here.
Will add immediately.

hharutyunyan:
@ZackerySpytz I've added the unit tests and made a few other changes. Can you please pay me a favour to review those? 
Also the Azure Pipeline timed out and cancelled on the macOS PR Test without any failure and I guess there's nothing I can do.

zware:
@scoder, you're about the closest we have to an xml.minidom expert as far as I can tell; would you mind to review this?

Lib/test/test_minidom.py
```diff
@@ -1152,6 +1152,19 @@ def testEncodings(self):
...
+    def testStandalone(self):
+        doc = parseString('<foo>&#x20ac;</foo>')
+        self.assertEqual(doc.toxml(),
+                         '<?xml version="1.0" ?><foo>\u20ac</foo>')
+        self.assertEqual(doc.toxml(standalone=None),
+                         '<?xml version="1.0" ?><foo>\u20ac</foo>')
+        self.assertEqual(doc.toxml(standalone=True),
+             '<?xml version="1.0" standalone=\'yes\'?><foo>\u20ac</foo>')
+        self.assertEqual(doc.toxml(standalone=False),
+             '<?xml version="1.0" standalone=\'no\'?><foo>\u20ac</foo>')
```
scoder: I think both values should use the same quotes here. Since both the version and the encoding used double quotes, please also use double quotes for `standalone`.

Please also add a test that declares an encoding.

Lib/xml/dom/minidom.py
```diff
@@ -1787,12 +1788,21 @@ def importNode(self, node, deep):
...
+        # In case standalone declaration is set
+        if standalone is not None:
+            standalone = "standalone='{}'".format('yes' if standalone else 'no')
         else:
-            writer.write('<?xml version="1.0" encoding="%s"?>%s' % (
-                encoding, newl))
+            standalone = ''
+
+        writer.write(
+            '<?xml version="1.0" {encoding}{standalone}?>{newline}'.format(
```
scoder: Feel free to use f-strings.

Lib/xml/dom/minidom.py
```diff
@@ -1787,12 +1788,21 @@ def importNode(self, node, deep):
...
         return _clone_node(node, deep, self)
 
-    def writexml(self, writer, indent="", addindent="", newl="", encoding=None):
-        if encoding is None:
-            writer.write('<?xml version="1.0" ?>'+newl)
+    def writexml(self, writer, indent="", addindent="", newl="", encoding=None,
+                 standalone=None):
+        # In case standalone declaration is set
+        if standalone is not None:
+            standalone = "standalone='{}'".format('yes' if standalone else 'no')
```
scoder: I would put a space before `standalone=…` here.

Misc/ACKS
```diff
@@ -1883,3 +1883,4 @@ Robert Leenders
 Tim Hopper
 Dan Lidral-Porter
 Ngalim Siregar
+Henrik Harutyunyan
```
scoder: The intended name sorting is alphabetically by last name. Please try to adhere to that.

scoder (changes requested):
Thanks, looks mostly good. Some more tests will uncover the missing space in formatting.

Docstrings would be nice. I'm aware that they are currently missing, but writing them shouldn't be all that much work, and we have to start somewhere.

The news entry seems a bit verbose, but it's your choice if you want to change it.

Lib/test/test_minidom.py
henryh9n: Yup, missed the test case with an encoding. 
Added and the quote styles are made consistent either.

Lib/xml/dom/minidom.py
henryh9n: Right. Done as well.

Lib/xml/dom/minidom.py
henryh9n: Absolutely. Added.

hharutyunyan:

hharutyunyan:
> Docstrings would be nice. I'm aware that they are currently missing, but writing them shouldn't be all that much work, and we have to start somewhere.

The next thing I'm gonna do is to enhance the documentation for the module, so I'd rather leave it for now.

Lib/test/test_minidom.py
```diff
@@ -1152,6 +1152,22 @@ def testEncodings(self):
...
         doc.unlink()
 
+    def testStandalone(self):
+        doc = parseString('<foo>&#x20ac;</foo>')
+        self.assertEqual(doc.toxml(),
+                         '<?xml version="1.0" ?><foo>\u20ac</foo>')
+        self.assertEqual(doc.toxml(standalone=None),
+                         '<?xml version="1.0" ?><foo>\u20ac</foo>')
+        self.assertEqual(doc.toxml(standalone=True),
+            '<?xml version="1.0"  standalone="yes"?><foo>\u20ac</foo>')
```
scoder: There shouldn't be a double space before `standalone`. It's unfortunate that there is a space after the version even without further parameters, but then, maybe there should just be a space after each of the parameters?
In any case: we should keep the same serialisation as before for all cases without `standalone`, and make the ones with the new parameter consistent with what was there before.

henryh9n: @scoder the double space there looks very much distracting, one the other hand adding a space after each of the parameters (i.e. `encoding`) causes failures in some of the past tests.

How about doing something like this:
```python
if encoding and standalone:
    standalone = ' ' + standalone
```

hharutyunyan:
@scoder Are you reviewing this sometime soon? :)

scoder (approved):

scoder:
The implementation is good now, but the documentation needs updating, since the signatures are wrong now. See `xml.dom.minidon.rst`.

hharutyunyan:
> The implementation is good now, but the documentation needs updating, since the signatures are wrong now. See `xml.dom.minidon.rst`.

@scoder I've updated the docs. You can have a look and merge now.

Doc/library/xml.dom.minidom.rst
```diff
@@ -144,11 +145,18 @@ module documentation.  This section lists the differences between the API and
    For the :class:`Document` node, an additional keyword argument *encoding* can
    be used to specify the encoding field of the XML header.
 
+   Silimarly, explicitly stating the *standalone* argument causes the
+   standalone document declarations to be added to the prologue of the XML
+   document.
+   If the value is set to `True`, `standalone="yes"` is added,
+   otherwise the it is set to `"no"`.
```
scoder:
```suggestion
   otherwise it is set to `"no"`.
```

scoder:
Thanks!

PR #14915, opened 2019-07-23
Add missing space to warning message

The-Compiler:
This was introduced in GH-13409 when changing the message text.

cc @Carreau 

This should probably go to the 3.8 branch as well?

### DIFF
--- a/Lib/collections/__init__.py
+++ b/Lib/collections/__init__.py
@@ -47,7 +47,7 @@ def __getattr__(name):
         obj = getattr(_collections_abc, name)
         import warnings
         warnings.warn("Using or importing the ABCs from 'collections' instead "
-                      "of from 'collections.abc' is deprecated since Python 3.3,"
+                      "of from 'collections.abc' is deprecated since Python 3.3, "
                       "and in 3.9 it will stop working",
                       DeprecationWarning, stacklevel=2)
         globals()[name] = obj


the-knights-who-say-ni:

Carreau (approved):

Carreau:
Oops, thanks !

Lib/collections/__init__.py
```diff
@@ -47,7 +47,7 @@ def __getattr__(name):
         obj = getattr(_collections_abc, name)
         import warnings
         warnings.warn("Using or importing the ABCs from 'collections' instead "
-                      "of from 'collections.abc' is deprecated since Python 3.3,"
+                      "of from 'collections.abc' is deprecated since Python 3.3, "
                       "and in 3.9 it will stop working",
```
aeros, line 51: ~~This might not be necessary, but it looks like it's also missing a period at the end.~~

aeros (approved):

aeros167:
@The-Compiler Thanks for the contribution! Yes, this should be backported to 3.8 as well since the space is also missing on that branch.

The-Compiler:
FWIW I signed the CLA right after opening this PR, but looks like it hasn't been received yet?

Lib/collections/__init__.py
The-Compiler: FWIW looks like most warning messages (`grep -R -F -A2 'warnings.warn(' Lib`) don't have a trailing dot.

aeros167:
@The-Compiler CLA approval usually takes around 24 hours. Sometimes though you have to manually refresh the status on GitHub, use the link from @the-knights-who-say-ni: [check yourself](https://check-python-cla.herokuapp.com/).

Lib/collections/__init__.py
aeros: Oh my mistake then, feel free to disregard that suggestion.

The-Compiler:
I'm aware - it's been three days and the checker still says "The-Compiler has not signed CLA".

aeros167:
@The-Compiler The only other option that I'm aware of is attempting to send email to contributors@python.org

The-Compiler:
@aeros167 Thanks, that helped - turns out I somehow have two bugs.python.org accounts ("The-Compiler" from 2012, "The Compiler" from 2014) but never noticed, and they accidentally marked the wrong one.

Anything else I need to do at this point? I don't think this needs a news item or issue number FWIW.

aeros167:
@The-Compiler At this point, it's just waiting for a core dev review, they can add a skip news and skip issue label for you

rhettinger (approved):

PR #14919, opened 2019-07-23
bpo-29446: IDLE -- add explicit imports

terryjreedy:
Stop depending on tkinter import *.

https://bugs.python.org/issue29446

### DIFF
--- a/Lib/idlelib/editor.py
+++ b/Lib/idlelib/editor.py
@@ -2,7 +2,9 @@ import importlib.abc
 import importlib.util
 import os
 import platform
+import re
 import string
+import sys
 import tokenize
 import traceback
 import webbrowser


terryjreedy:
@taleinat @vstinner  This buildbot IDLE test failure is a false positive that coincidentally occurred when testing another, trivial, IDLE change.  What this report omits is that the test passed on the retest.  The test that failed is a year old, written by Tal.  (ENV CHANGED is from test_multiprocessing_spawn.)

Victor: Have you seen this failure before.  Also, I thought the point of retest was to ignore transient failures like this?
Tal: Another test involving a mouse and live widget failed a year ago.  bpo 31971.  Perhaps there is some similarity.

terryjreedy:
I forgot about seeing this before.  See issue 35771.

PR #14925, opened 2019-07-24
gh-74940: Allow fallback to UTF-8 encoding on systems with no locales installed.

gordonmessmer:
This change removes the alias of the 'C' locale to 'en_US'.  Because of this alias, it is currently impossible for an application to use setlocale() to specify a UTF-8 locale on a system that has no locales installed, but which supports the C.UTF-8 locale/encoding.

https://bugs.python.org/issue30755

* Issue: gh-74940


furkanonder (approved):

arhadthedev:
@vstinner @methane (as contributors into `Lib/locale.py`)

methane (approved):

PR #14927, opened 2019-07-24
bpo-37667: Only setup PGO tests when PGO is actually enabled.

gpshead:
We're skipping all but 40 of the tests when running the test suite now as of https://github.com/python/cpython/commit/2406672984e4c1b18629e615edad52928a72ffcc

OOPS :)

https://bugs.python.org/issue37667

Automerge-Triggered-By: @gpshead

### DIFF
--- a/Lib/test/libregrtest/main.py
+++ b/Lib/test/libregrtest/main.py
@@ -215,8 +215,9 @@ class Regrtest:
 
         removepy(self.tests)
 
-        # add default PGO tests if no tests are specified
-        setup_pgo_tests(self.ns)
+        if self.ns.pgo:
+            # add default PGO tests if no tests are specified
+            setup_pgo_tests(self.ns)
 
         stdtests = STDTESTS[:]
         nottests = NOTTESTS.copy()


methane (approved):

gpshead:
CI is happily running all 419 tests again with this PR. =)

and I have manually confirmed that ./python -m test.regrtest --pgo runs its select set of 40.

nascheme:
Thanks for fixing this.  I botched the refactoring that introduced setup_pgo_tests(). :cry:

gpshead:
No prob. I'm not sure if my fix is _ideal_, I was just surgically solving
the immediate problem.

>

nascheme:
I think it is the correct fix.  It is exactly what I had in my tree before I realized you already fixed it.  Initially I had the "if ns.pgo" test inside the setup_pgo_tests().  However, I decided that makes the call in main.py a bit confusing since it looks like something gets done in the non-PGO case.  So, I removed the test from setup_pgo_tests() but then I forget to add it to main.py.  So, your fix makes the code the way I intended it.

I should have run the test suite after the refactoring.  My testing consisted of running the build using "--enable-optimizations" and "--enable-optimzations PROFILE=TASK=...".  I forget to run the normal test suite.  No major harm done but kind of embarrassing. :grin:

PR #14933, opened 2019-07-24
bpo-37669: Make mock_open return per-file-name content

Anvil:
This change should allow people to have different contents available through mock_open handles.
Some new items in the file may be poorly named or documentation about them unclear. Feel free to tell me.

With this change I should be able to perform this kind of test with a mock_open.
```python
with open("file1") as file1:
    assert file1.read() == "data1"

with open("file2") as file2:
    assert file2.read() == "data2"
```

https://bugs.python.org/issue37669

### DIFF
--- a/Lib/unittest/mock.py
+++ b/Lib/unittest/mock.py
@@ -2682,7 +2682,63 @@ def _to_stream(read_data):
         return io.StringIO(read_data)
 
 
-def mock_open(mock=None, read_data=''):
+class _MockOpenHandleHandler:
+
+    def __init__(self, read_data):
+        self.read_data = read_data
+        self._state = [_to_stream(read_data), None]
+        self.handle = handle = MagicMock(spec=file_spec)
+
+        handle.__enter__.return_value = handle
+
+        handle.write.return_value = None
+        handle.read.return_value = None
+        handle.readline.return_value = None
+        handle.readlines.return_value = None
+
+        handle.read.side_effect = self._read_side_effect
+        self._state[1] = self._readline_side_effect()
+        handle.readline.side_effect = self._state[1]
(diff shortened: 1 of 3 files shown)


Anvil:
Some tests are broken because of the "multiple handles" thingy I'm introducing. The question is: should the tests be changed, or another function (`mock_open_many_files` be created) instead of an altered `mock_open` ?

Anvil:
Nevermind, I've found what I got wrong.

Lib/unittest/mock.py
```diff
@@ -2693,70 +2749,39 @@ def mock_open(mock=None, read_data=''):
...
-            while True:
-                yield handle.readline.return_value
-        for line in _state[0]:
-            yield line
-
-    def _next_side_effect():
-        if handle.readline.return_value is not None:
-            return handle.readline.return_value
-        return next(_state[0])
+    `data` is any value acceptable for `dict` constructor. The keys
```
ZackerySpytz:
```suggestion
    `data` is any value acceptable for the `dict` constructor. The keys
```

Lib/unittest/mock.py
```diff
@@ -2693,70 +2749,39 @@ def mock_open(mock=None, read_data=''):
...
+    represent file names. If a file name, found in `data`, is
+    mock-opened, and the value matching the file name in `data` will
```
ZackerySpytz: This sentence doesn't read right to me. Should the "and" be replaced with "then" here?

Anvil:
I'm sorry for being AWOL. "Life happened", for better or worse. 

@ZackerySpytz I've applied the changes you've suggested : yes, obviously my english was broken.

cjw296:
Closing this PR as discussed on https://bugs.python.org/issue37669

PR #14934, opened 2019-07-24
bpo-37664: ensurepip bundle update

pradyunsg:
https://bugs.python.org/issue37664


zooba:
@pradyunsg Can you add a NEWS entry for this (including the version numbers of the new packages)?

There are instructions under the "Details" link if you haven't done this for CPython before.

pradyunsg:
Had some mixup with the issue number but all should be good now. :)

PR #14938, opened 2019-07-24
Remove duplicate call to strip method

hansrajdas:
* Strip method was already used at starting of the loop, no need to call again
* Combined 2 if blocks as in both we need to continue


pablogsal:
Thanks, @hansrajdas for the PR!

PR #14941, opened 2019-07-24
bpo-37663: have venv activation scripts all consistently use __VENV_PROMPT__ for prompt customization

brettcannon:
The activation scripts generated by venv were inconsistent in how they changed the shell's prompt. Some used `__VENV_PROMPT__` exclusively, some used `__VENV_PROMPT__` if it was set although by default `__VENV_PROMPT__` is always set and the fallback matched the default, and one ignored `__VENV_PROMPT__` and used `__VENV_NAME__` instead (and even used a differing format to the default prompt). This change now has all activation scripts use `__VENV_PROMPT__` only and relies on the fact that venv sets that value by default.

The color of the customization is also now set in fish to the blue from the Python logo for as hex color support is built into that shell (much like PowerShell where the built-in green color is used).

https://bugs.python.org/issue37663


mental32 (approved):
Thanks for cc'ing :)

Changes look good to me!

I'm curious about the special casing of "aspen magic directories" would you know anything about these? the link seems to point to a broken site.

Also I noticed you made typo changes, can I assume that'd be alright to do? (in case I catch an issue that needs a patch and a tidy up)

brettcannon:
@mental32 I have no clue about them and since the URL is dead and the site seems to now be a news site I figured it wasn't worth keeping the code around.

And yes, you can always touch up typos and such in comments if you come across them while working on code. The only reason I touched up the punctuation for the whole fish file was it's small enough to not be horribly distracting in the PR.

brettcannon:
@vsajip any issues with this change?

vsajip (approved):
LGTM.

PR #14947, opened 2019-07-25
Swap 'if' branches so content matches to condition in importlib example

uranusjr:
Prior to this change the guard on an 'elif' used an assignment expression whose value was used in a later 'else' block, causing some confusion for people.

(Discussion on Twitter: https://twitter.com/brettsky/status/1153861041068994566.)

Automerge-Triggered-By: @brettcannon


tirkarthi:
cc: @brettcannon

pradyunsg (approved):
Me likey this more.

mental32 (approved):
:+1: This reads cleaner

brettcannon (approved):

PR #14950, opened 2019-07-25
bpo-24935: On all posix systems, not just Darwin, set LDSHARED (if not set) according to CC

xuhdev:
This patch is slightly different from https://bugs.python.org/issue24935
, except that we now handle LDSHARED according to CC on all posix
systems, not just Darwin or Linux.

https://bugs.python.org/issue24935

### DIFF
--- a/Lib/distutils/sysconfig.py
+++ b/Lib/distutils/sysconfig.py
@@ -193,10 +193,10 @@ def customize_compiler(compiler):
 
         if 'CC' in os.environ:
             newcc = os.environ['CC']
-            if (sys.platform == 'darwin'
+            if (os.name == 'posix'
                     and 'LDSHARED' not in os.environ
                     and ldshared.startswith(cc)):
-                # On OS X, if CC is overridden, use that as the default
+                # On POSIX systems, if CC is overridden, use that as the default
                 #       command for LDSHARED as well
                 ldshared = newcc + ldshared[len(cc):]
             cc = newcc


the-knights-who-say-ni:

Lib/distutils/sysconfig.py
```diff
@@ -193,7 +193,7 @@ def customize_compiler(compiler):
 
         if 'CC' in os.environ:
             newcc = os.environ['CC']
-            if (sys.platform == 'darwin'
+            if (os.name == 'posix'
                     and 'LDSHARED' not in os.environ
                     and ldshared.startswith(cc)):
                 # On OS X, if CC is overridden, use that as the default
```
ned-deily: If you change the test to include all posix platforms, this comment needs to be changed as well.

xuhdev: Nice catch! Updated!

ned-deily:
The only concern I have at the moment is if this change is appropriate for a cross-compiling environment.  @doko42, what do you think?

simark:
I stumbled on this problem when trying to build a project I'm working on (babeltrace) with `CC=clang`.  I put some clang-specific warning flags in `CFLAGS`.  My module gets compiled with clang as expected, with those warning flags.  However, distutils uses its default compiler (gcc) to do the linking _and_ passes it `$CFLAGS`.  Since `CFLAGS` contains clang-specific warning flags, it makes gcc error out.

Anything I can do to make this patch/issue progress?

FRidh:
Since https://github.com/NixOS/nixpkgs/pull/91178 we use this patch in Nixpkgs for cross-compilation.

iritkatriel:
Closing this PR following the closure of its b.p.o issue (https://bugs.python.org/issue24935)

PR #14951, opened 2019-07-25
bpo-37684: Improved list.extend example in datastructures docs

wimglenn:
The current example is not a good equivalent in regards to `list.extend`'s atomicity (or lack thereof).

Please add skip news label, thanks.

https://bugs.python.org/issue37684

### DIFF
--- a/Doc/tutorial/datastructures.rst
+++ b/Doc/tutorial/datastructures.rst
@@ -26,7 +26,7 @@ objects:
    :noindex:
 
    Extend the list by appending all the items from the iterable.  Equivalent to
-   ``a[len(a):] = iterable``.
+   ``for x in iterable: a.append(x)``.
 
 
 .. method:: list.insert(i, x)


Doc/tutorial/datastructures.rst
```diff
@@ -26,7 +26,7 @@ objects:
    :noindex:
 
    Extend the list by appending all the items from the iterable.  Equivalent to
-   ``a[len(a):] = iterable``.
+   ``for x in iterable: a.append(x)``.
```
aeros, line 29:
```suggestion
      >>> round_table = ['Arthur']
      >>> knights = ['Bedevere', 'Lancelot', 'Gallahad', 'Robin']
      >>> round_table.extend(knights)
      >>> round_table
      ['Arthur', 'Bedevere', 'Lancelot', 'Gallahad', 'Robin']
```
This would also require a change to the description as well, but that could be fixed simply by removing the "Equivalent to".

Alternatively, I would at least suggest adjusting your example to use two separate lines and indentation, to meet the standard styling conventions:
```python
for x in iterable: 
    a.append(x)
```

Doc/tutorial/datastructures.rst
```diff
@@ -26,7 +26,7 @@ objects:
    :noindex:
 
    Extend the list by appending all the items from the iterable.  Equivalent to
```
aeros, line 28:
```suggestion
    Extend the list by appending all the items from the iterable.
```

aeros (changes requested):
@wimglenn Thanks for the contribution! I agree that a `skip news` label would be appropriate for this PR.

I agree that the suggested changes would provide a significant improvement from the current code example, but since this is in the tutorial, I think it would be helpful to provide a more detailed example showing outputs from the interpreter. This is less commonly done in the rest of the docs, but the tutorial tends to prefer the usage of more detailed examples (often times with some fun theming).

The one line equivalent seems like it would be more suitable to the library docs, rather than the tutorial. A visual example with the interpreter would be significantly more informative to newer users of Python, which is the main target audience of the tutorial.

PR #14954, opened 2019-07-26
bpo-37585: Add clarification to dict.values()

aeros:
Based on the [python-dev discussion](https://mail.python.org/archives/list/python-dev@python.org/thread/R2MPDTTMJXAF54SICFSAWPPCCEWAJ7WF/#723CHZBH4ZBKQJOOPXFFX3HYSPDBPDPR) of the ``dict.values()`` method, the most agreed upon course of action seems to be mentioning the equality comparison behavior in the documentation rather than changing the existing one. This will help provide clarification on this behavior to users without requiring modifications to the equality comparison for dictionary views. 

https://bugs.python.org/issue37585

### DIFF
--- a/Doc/library/stdtypes.rst
+++ b/Doc/library/stdtypes.rst
@@ -4354,6 +4354,9 @@ pairs within braces, for example: ``{'jack': 4098, 'sjoerd': 4127}`` or ``{4098:
       Return a new view of the dictionary's values.  See the
       :ref:`documentation of view objects <dict-views>`.
 
+      Equality comparisons between the values view of two
+      dictionaries will always return ``False``.
+
    Dictionaries compare equal if and only if they have the same ``(key,
    value)`` pairs. Order comparisons ('<', '<=', '>=', '>') raise
    :exc:`TypeError`.
--- a/Doc/library/stdtypes.rst
+++ b/Doc/library/stdtypes.rst
@@ -4358,7 +4358,7 @@ pairs within braces, for example: ``{'jack': 4098, 'sjoerd': 4127}`` or ``{4098:
       will always return ``False``. This also applies when comparing
       ``dict.values()`` to itself::
 
-         >>> d = {'a' : 1}
+         >>> d = {'a': 1}
          >>> d.values() == d.values()
          False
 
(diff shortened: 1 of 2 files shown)


nanjekyejoannah (commented):
This statement "values view" doesn't sound clear. Did you mean "values' view" or something more clearer. Am not a native English speaker so you can elaborate If am wrong.

aeros167:
> This statement "values view" doesn't sound clear. Did you mean "values' view" or something more clearer. Am not a native English speaker so you can elaborate If am wrong.

By "values view" I was referring the dictionary view accessed through usage of ``dict.values()``. For this scenario, ``values'`` would be not be correct, as that would be stating that the values have ownership of the view. For example: "The students' calculators" is stating that there are multiple students which have ownership of multiple calculators.

Perhaps it would make that section more clear if I explicitly state that "values" is referring to "dict.values()" and change "of" to "from". Let me know if this would be any more clear:

Equality comparisons between the ``dict.values()`` view from two dictionaries will always return ``False``.

aeros167:
I was also thinking of adding another sentence to this section, to explicitly specify that this behavior applies when comparing a ``dict.values()`` to itself. For example:

```python3
>>> d = {'a': 1}
>>> d1.values() == d1.values()
False
```

nanjekyejoannah:
> Perhaps it would make that section more clear if I explicitly state that "values" is referring to "dict.values()" and rephrase the section a bit:

paraphrasing would work.

nanjekyejoannah:
> I was also thinking of adding another sentence to this section, to explicitly specify that this behavior applies even when comparing a `dict.values()` to itself. For example:
> 
> ```python
> >>> d = {'a': 1}
> >>> d1.values() == d1.values()
> False
> ```
> 
> This behavior is not very intuitive at first, read through the python-dev discussion that I linked for more information on why this happens.

This sounds okay to me. You may need to get another +1 from someone else though.  cc @brettcannon

aeros167:
Brett had suggested for the behavior to be documented (in the ML thread), so if he could review this PR that'd be great. (:

I'll add the clarification.

brettcannon (approved):

aeros167:
Thanks @brettcannon!

PR #14956, opened 2019-07-26
bpo-37409: fix relative import with no parent

benjimin:
`builtins.__import__` was missing a check to flag that if no dot was found in the `__name__` of a module that isn't explicitly a submodule of a package then a relative import is incorrect.

https://bugs.python.org/issue37409

### DIFF
--- a/Lib/test/test_builtin.py
+++ b/Lib/test/test_builtin.py
@@ -162,7 +162,7 @@ class BuiltinTest(unittest.TestCase):
         self.assertRaises(TypeError, __import__, 'sys', name='sys')
         # relative import with no parent package, issue37409
         self.assertRaises(ImportError, __import__, '',
-                          {'__package__': None, '__name__': '__main__'},
+                          {'__package__': None, '__spec__': None, '__name__': '__main__'},
                           {}, ('foo',), 1)
         # embedded null character
         self.assertRaises(ModuleNotFoundError, __import__, 'string\x00')
--- a/Lib/test/test_builtin.py
+++ b/Lib/test/test_builtin.py
@@ -163,7 +163,7 @@ class BuiltinTest(unittest.TestCase):
         # relative import with no parent package, issue37409
         self.assertRaises(ImportError, __import__, '',
                           {'__package__': None, '__spec__': None, '__name__': '__main__'},
-                          {}, ('foo',), 1)
+                          locals={}, fromlist=('foo',), level=1)
         # embedded null character
         self.assertRaises(ModuleNotFoundError, __import__, 'string\x00')
 
--- a/Lib/test/test_builtin.py
+++ b/Lib/test/test_builtin.py
@@ -160,7 +160,7 @@ class BuiltinTest(unittest.TestCase):
         self.assertRaises(TypeError, __import__, 1, 2, 3, 4)
         self.assertRaises(ValueError, __import__, '')
         self.assertRaises(TypeError, __import__, 'sys', name='sys')
-        # relative import with no parent package, issue37409
+        # Relative import outside of a package with no __package__ or __spec__ (bpo-37409).
         self.assertRaises(ImportError, __import__, '',
                           {'__package__': None, '__spec__': None, '__name__': '__main__'},
                           locals={}, fromlist=('foo',), level=1)
--- a/Lib/test/test_builtin.py
+++ b/Lib/test/test_builtin.py
@@ -160,6 +160,10 @@ class BuiltinTest(unittest.TestCase):
         self.assertRaises(TypeError, __import__, 1, 2, 3, 4)
         self.assertRaises(ValueError, __import__, '')
         self.assertRaises(TypeError, __import__, 'sys', name='sys')
+        # relative import with no parent package, issue37409
+        self.assertRaises(ImportError, __import__, '',
+                          {'__package__': None, '__name__': '__main__'},
+                          {}, ('foo',), 1)
         # embedded null character
         self.assertRaises(ModuleNotFoundError, __import__, 'string\x00')
 
--- a/Lib/test/test_import/__init__.py
+++ b/Lib/test/test_import/__init__.py
@@ -775,6 +775,11 @@ class RelativeImportTests(unittest.TestCase):
         ns = dict(__package__=object())
         self.assertRaises(TypeError, check_relative)
 
+    def test_import_shadowed_by_global(self):
+        # Regression test for https://bugs.python.org/issue37409
+        script_helper.assert_python_failure('-W', 'ignore', '-c',
+            "foo = 1; from . import foo")
+
     def test_absolute_import_without_future(self):
         # If explicit relative import syntax is used, then do not try
         # to perform an absolute import in the face of failure.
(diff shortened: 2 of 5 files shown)


benjimin:
This is a conservative (minimal) change, intended to also be appropriate for back-porting.

The bug was that (sibling-level) relative import statements would in some cases (when unable to identify any enclosing package) instead retrieve objects from the current module rather than raising the expected ImportError. (Only `builtins.__import__` was affected, not `importlib.__import__`)

Specifically, the problem concerned circumstances where `__package__` and `__spec__.parent` are uninformative and the current module is not itself a package (i.e. `__path__` is undefined, contrary to common *`__init__.py`*). In these circumstances either the package can be identified from `__name__` (after truncating the last .component), or no package can be found (e.g. top-level modules and, more commonly, scripts/modules run as `"__main__"`). This last case was failing to trigger (instead treating the module itself as the package).

Lib/test/test_builtin.py
```diff
@@ -160,6 +160,10 @@ def test_import(self):
         self.assertRaises(TypeError, __import__, 1, 2, 3, 4)
         self.assertRaises(ValueError, __import__, '')
         self.assertRaises(TypeError, __import__, 'sys', name='sys')
+        # relative import with no parent package, issue37409
```
brettcannon:
```suggestion
        # Relative import outside of a package with no __package__ or __spec__ (bpo-37409).
```

Lib/test/test_builtin.py
```diff
@@ -160,6 +160,10 @@ def test_import(self):
...
+        self.assertRaises(ImportError, __import__, '',
+                          {'__package__': None, '__name__': '__main__'},
```
brettcannon:
```suggestion
                          {'__package__': None, '__spec__': None, '__name__': '__main__'},
```

Lib/test/test_builtin.py
```diff
@@ -160,6 +160,10 @@ def test_import(self):
...
+                          {}, ('foo',), 1)
```
brettcannon:
```suggestion
                          locals={}, fromlist=('foo',), level=1)
```

Lib/test/test_import/__init__.py
```diff
@@ -775,6 +775,11 @@ def check_relative():
         ns = dict(__package__=object())
         self.assertRaises(TypeError, check_relative)
 
+    def test_import_shadowed_by_global(self):
+        # Regression test for https://bugs.python.org/issue37409
+        script_helper.assert_python_failure('-W', 'ignore', '-c',
```
brettcannon, line 780: Why execute Python as a subprocess for something that should raise an exception? Any reason you can't just follow the `test_absolute_import_without_future()` example below and do the import and catch the exception?

Lib/test/test_import/__init__.py
```diff
@@ -775,6 +775,11 @@ def check_relative():
         ns = dict(__package__=object())
         self.assertRaises(TypeError, check_relative)
 
+    def test_import_shadowed_by_global(self):
+        # Regression test for https://bugs.python.org/issue37409
```
brettcannon:
```suggestion
        # bpo-37409
```

Python/import.c
```diff
@@ -1648,13 +1648,15 @@ resolve_name(PyThreadState *tstate, PyObject *name, PyObject *globals, int level
                 goto error;
             }
```
brettcannon: Instead of the implicit failure by faking that the dot was found at 0, add a clause of the failure case and then make the latter guard just be an else:

```suggestion
            else if (dot == -1) {
                _PyErr_SetString(tstate, PyExc_ImportError,
                         "attempted relative import "
                         "with no known parent package");
                goto error;
            }
            else {
```
This more clearly communicates what you're trying to accomplish. Basically there's no need to be clever here as it isn't a performance win in a critical code path.

brettcannon (changes requested):
I have a suggestion on how to make the fix while taking a more explicit/simpler solution.

And please add a news entry for this.

Lib/test/test_import/__init__.py
benjimin: Because this `__init__.py` is expected to be loaded as a package, whereas the bugfix is about scripts, interactive sessions, etc. (So without starting a new interpreter, the relative import would be expected to always succeed, uninformatively.)

My other thought behind this test was to focus on the behaviour from a user perspective while being agnostic about all details of how `import` statements are implemented under the hood. This test would stay relevant even if the import internals were refactored (in contrast to the other test, that narrowly targets the specific `__import__` case).

Python/import.c
benjimin: Agree with your point. I've pushed another commit to tidy this up.

benjimin:

Python/import.c
```diff
@@ -1689,7 +1686,13 @@ resolve_name(PyThreadState *tstate, PyObject *name, PyObject *globals, int level
     Py_DECREF(base);
     return abs_name;
 
+  no_parent_error:
+
```
brettcannon:
```suggestion
```

Python/import.c
```diff
@@ -1689,7 +1686,13 @@ resolve_name(PyThreadState *tstate, PyObject *name, PyObject *globals, int level
...
+    _PyErr_SetString(tstate, PyExc_ImportError,
+                     "attempted relative import "
+                     "with no known parent package");
   error:
+
```
brettcannon, line 1693:
```suggestion
```

Python/import.c
```diff
@@ -1689,7 +1686,13 @@ resolve_name(PyThreadState *tstate, PyObject *name, PyObject *globals, int level
     Py_DECREF(base);
     return abs_name;
 
+  no_parent_error:
+
+    _PyErr_SetString(tstate, PyExc_ImportError,
+                     "attempted relative import "
+                     "with no known parent package");
```
brettcannon, line 1692:
```suggestion
                     "with no known parent package");

```

Misc/NEWS.d/next/Core and Builtins/2019-08-06-23-39-05.bpo-37409.1qwzn2.rst
```diff
@@ -0,0 +1 @@
+Ensure explicit relative imports from interactive sessions and scripts (having no parent package) always raise ImportError, rather than treating the current module as the package.
```
brettcannon:
```suggestion
Ensure explicit relative imports from interactive sessions and scripts (having no parent package) always raise ImportError, rather than treating the current module as the package.
Patch by Ben Lewis.
```

Lib/test/test_import/__init__.py
```diff
@@ -775,6 +775,11 @@ def check_relative():
         ns = dict(__package__=object())
         self.assertRaises(TypeError, check_relative)
 
+    def test_parentless_import_shadowed_by_global(self):
+        # bpo-37409
```
brettcannon:
```suggestion
        # Test as if this were done from the REPL where this error most commonly occurs (bpo-37409).
```

brettcannon (changes requested):
Some very minor tweaks and then this will be ready to be merged!

benjimin:
@brettcannon is this also appropriate for backporting to 3.7?

I hope this fix may help people encounter more informative error messages when they are first learning/experimenting with explicit relative imports (say, migrating code from py2).

encukou (approved):
This looks great!
@brettcannon, do you still want to re-review?

brettcannon (approved):

brettcannon:
#15925 is the 3.7 backport

brettcannon:
And all the backports are done! Thanks, @benjimin .

PR #14957, opened 2019-07-26
bpo-37538: Zipfile refactor

danifus:
This pull request aims to make ZipFile easier to subclass and extend.

The general goals of the refactor were to:
- Add hooks for extending the way zipfile works to enable a subclass to 
  add AES encryption without having to duplicate most of the zipfile
  module. This included adding hooks to:
    - Read and write new "extra" data records in the central file directory
      and local header records.
    - Provide a mechanism to substitute ZipInfo, ZipExtFile and ZipWriteFile
      classes used in a subclass of ZipFile to ease use of subclassed ZipInfo,
      ZipExtFile or ZipWriteFile. This avoids having to rewrite large parts of
      the zipfile module if we only want to change the behaviour of a small
      part of one of those classes.
- Contain all code that reads the header, contents and tail of a file in the
  archive to within ZipExtFile. Previously reading the header and some other
  things were done in the ZipFile class before handing the rest of the
  processing to ZipExtFile.
- Contain all code that writes the header, contents and tail of a file in the
  archive to within ZipWriteFile. Previously reading the header and some other
  things were done in the ZipFile class before handing the rest of the
  processing to ZipExtFile.
- Move generation of local file header and central directory record content to
  the ZipInfo class to be alongside the data that it is packing.
- Add comments to provide context from the zip spec. Replace explicit numbers
  to variables with explanatory names or adding comments.

This patch contains some bug fixes (seeking an encrypted file) that were made
possible by the refactor.

Happy for suggestions to improve this patch.

https://bugs.python.org/issue37538

### DIFF
--- a/Lib/test/test_zipfile.py
+++ b/Lib/test/test_zipfile.py
@@ -407,7 +407,7 @@ class AbstractTestsWithSourceFile:
             def write(self, data):
                 nonlocal count
                 if count is not None:
-                    if count == stop:
+                    if (count > stop):
                         raise OSError
                     count += 1
                 super().write(data)
@@ -424,11 +424,12 @@ class AbstractTestsWithSourceFile:
                     with zipfp.open('file2', 'w') as f:
                         f.write(b'data2')
                 except OSError:
-                    stop += 1
+                    pass
                 else:
                     break
                 finally:
                     count = None
+                stop += 1
             with zipfile.ZipFile(io.BytesIO(testfile.getvalue())) as zipfp:
                 self.assertEqual(zipfp.namelist(), ['file1'])
                 self.assertEqual(zipfp.read('file1'), b'data1')
--- a/Lib/test/test_zipfile.py
+++ b/Lib/test/test_zipfile.py
@@ -1289,7 +1289,8 @@ class OtherTests(unittest.TestCase):
         with zipfile.ZipFile(TESTFN2, 'w') as orig_zip:
             for data in 'abcdefghijklmnop':
                 zinfo = zipfile.ZipInfo(data)
-                zinfo.flag_bits |= 0x08  # Include an extended local header.
+                # Include an extended local header.
+                zinfo.flag_bits |= zipfile._MASK_USE_DATA_DESCRIPTOR
                 orig_zip.writestr(zinfo, data)
 
     def test_close(self):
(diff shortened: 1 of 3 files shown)


danifus:
Here is a gist which shows the refactor in action. It implements winzip's AES encryption and decryption based on these changes.

https://gist.github.com/danifus/73d258df243bbb386c1dd64c0888cddf

If this pull request is too large, I'm happy to split it up with some guidance as to what would be acceptable.

Also, the CI tests is having an issue with 'env changed' but I'm unable to reproduce it locally. Any tips for chasing those type of errors down?

Thanks, Dan

gnprice:
Hi @danifus, and thanks! This sounds like a useful direction.

You're right that this is a large diff. In general, that makes it challenging for someone to review it. I'd recommend you try to split it into a sequence of smaller changes, each of which on its own is a good change and can be understood by the reader. That can make for much less total work for someone to understand and review, even when the sequence adds up to the exact same total change in the end.

As a logistical matter, I'd suggest you then send the first change in the sequence as its own PR; once that's merged, send the next one. You might keep posting the whole sequence as a branch, and linking to it for context, but that'd let a reviewer focus the detailed code-reading on just the one change at a time. (Someone else might have another suggestion, though.)

For the substance of where to split it:

---
> Add hooks for extending the way zipfile works to enable a subclass to
> add AES encryption without having to duplicate most of the zipfile
> module. This included adding hooks to [...]

This sounds like a good separate change. It might come after some other changes (maybe all of them?), if you need some of your other refactors before this change makes sense.

---
> This patch contains some bug fixes (seeking an encrypted file) that were made
> possible by the refactor.

This sounds great, and I'd definitely recommend making it its own separate change -- that will help in crisply explaining what behavior you're changing. I'd make it the first PR in the sequence if you can (without making it a big one). If not, then I'd make it the second or third, and start with whichever refactors you need specifically for this.

---
```
# Zip Appnote: 4.4.4 general purpose bit flag: (2 bytes)
_MASK_ENCRYPTED = 1 << 0
_MASK_COMPRESS_OPTION_1 = 1 << 1
# ...
```

This looks like a very nice improvement (vs. `if self.flag_bits & 0x08:` etc.) -- and one that will be easy for someone to review, if you can separate out a PR that's just this change. (I.e. adding these constants and their comments, and replacing the `0x08` and so on with the named constants.) If it weren't for the bugfix, this is probably the one I'd start with.

---
There are probably more splits it'd be helpful to make, which you can probably think of as you look back through your code with these in mind.

You also don't have to split everything into polished self-contained changes at once -- it's enough to identify one change you can send as a first PR in the series. That will help people get started reviewing, and then you can split more changes out as you go.

danifus:
Thanks for the advice! I'll start splitting this patch into bite size pieces.

danifus:
Closing this big PR in favour of implementing something similar but in smaller PRs

PR #14958, opened 2019-07-26
bpo-37628: Fix IDLE config sample sizes

taleinat:
This fixes the issue where setting a large font size makes the config dialog grow, in some cases making it unusable by pushing the dialog buttons (e.g. "Ok", "Cancel") off-screen.

Tested on Windows 10 and Ubuntu 18.10.

https://bugs.python.org/issue37628

### DIFF
--- a/Lib/idlelib/configdialog.py
+++ b/Lib/idlelib/configdialog.py
@@ -33,6 +33,7 @@ from idlelib.codecontext import CodeContext
 from idlelib.parenmatch import ParenMatch
 from idlelib.format import FormatParagraph
 from idlelib.squeezer import Squeezer
+from idlelib.textview import ScrollableTextFrame
 
 changes = ConfigChanges()
 # Reload changed options in the following classes.
@@ -194,7 +195,7 @@ class ConfigDialog(Toplevel):
 
     def destroy(self):
         global font_sample_text
-        font_sample_text = self.fontpage.font_sample.get('1.0', 'end')
+        font_sample_text = self.fontpage.font_sample.text.get('1.0', 'end')
         self.grab_release()
         super().destroy()
 
@@ -556,8 +557,11 @@ class FontPage(Frame):
                 frame_font_param, variable=self.font_bold,
                 onvalue=1, offvalue=0, text='Bold')
         # frame_sample.
-        self.font_sample = Text(frame_sample, width=20, height=20)
-        self.font_sample.insert(END, font_sample_text)
+        # self.font_sample = Text(frame_sample, width=20, height=20)
+        # self.font_sample.insert(END, font_sample_text)
+        self.font_sample = ScrollableTextFrame(frame_sample)
+        self.font_sample.text.config(wrap='word', width=1, height=1)
+        self.font_sample.text.insert('1.0', font_sample_text)
         # frame_indent.
         indent_title = Label(
                 frame_indent, justify=LEFT,
@@ -568,8 +572,9 @@ class FontPage(Frame):
 
         # Grid and pack widgets:
         self.columnconfigure(1, weight=1)
+        self.rowconfigure(2, weight=1)
         frame_font.grid(row=0, column=0, padx=5, pady=5)
-        frame_sample.grid(row=0, column=1, rowspan=2, padx=5, pady=5,
+        frame_sample.grid(row=0, column=1, rowspan=3, padx=5, pady=5,
                           sticky='nsew')
         frame_indent.grid(row=1, column=0, padx=5, pady=5, sticky='ew')
         # frame_font.
@@ -657,7 +662,7 @@ class FontPage(Frame):
         font_name = self.font_name.get()
         font_weight = tkFont.BOLD if self.font_bold.get() else tkFont.NORMAL
         new_font = (font_name, self.font_size.get(), font_weight)
-        self.font_sample['font'] = new_font
+        self.font_sample.text['font'] = new_font
         self.highlight_sample['font'] = new_font
 
     def load_tab_cfg(self):
--- a/Lib/idlelib/configdialog.py
+++ b/Lib/idlelib/configdialog.py
@@ -557,8 +557,8 @@ class FontPage(Frame):
                 frame_font_param, variable=self.font_bold,
                 onvalue=1, offvalue=0, text='Bold')
         # frame_sample.
-        font_sample_scrollable_frame = ScrollableTextFrame(frame_sample)
-        self.font_sample = font_sample_scrollable_frame.text
+        font_sample__frame = ScrollableTextFrame(frame_sample)
+        self.font_sample = font_sample__frame.text
         self.font_sample.config(wrap=NONE, width=1, height=1)
         self.font_sample.insert(END, font_sample_text)
         # frame_indent.
@@ -586,7 +586,7 @@ class FontPage(Frame):
         self.sizelist.pack(side=LEFT, anchor=W)
         self.bold_toggle.pack(side=LEFT, anchor=W, padx=20)
         # frame_sample.
-        font_sample_scrollable_frame.pack(expand=TRUE, fill=BOTH)
+        font_sample__frame.pack(expand=TRUE, fill=BOTH)
         # frame_indent.
         indent_title.pack(side=TOP, anchor=W, padx=5)
         self.indent_scale.pack(side=TOP, padx=5, fill=X)
--- a/Lib/idlelib/configdialog.py
+++ b/Lib/idlelib/configdialog.py
@@ -195,7 +195,7 @@ class ConfigDialog(Toplevel):
 
     def destroy(self):
         global font_sample_text
-        font_sample_text = self.fontpage.font_sample.text.get('1.0', 'end')
+        font_sample_text = self.fontpage.font_sample.get('1.0', 'end')
         self.grab_release()
         super().destroy()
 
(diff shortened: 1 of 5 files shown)


Lib/idlelib/configdialog.py
```diff
@@ -556,7 +557,9 @@ def create_page_font_tab(self):
                 frame_font_param, variable=self.font_bold,
                 onvalue=1, offvalue=0, text='Bold')
         # frame_sample.
-        self.font_sample = Text(frame_sample, width=20, height=20)
+        font_sample_scrollable_frame = ScrollableTextFrame(frame_sample)
+        self.font_sample = font_sample_scrollable_frame.text
+        self.font_sample.config(wrap='word', width=1, height=1)
```
terryjreedy: I am turning wrapping off as it is really ugly to me and partially undermines the purpose of the sample.
It seems that for non-ascii chars, each char is a word, so that word wrap is the same as char wrap.

terryjreedy:
I am reviewing and making a few edits now.

Lib/idlelib/configdialog.py
```diff
@@ -556,7 +557,9 @@ def create_page_font_tab(self):
                 frame_font_param, variable=self.font_bold,
                 onvalue=1, offvalue=0, text='Bold')
         # frame_sample.
-        self.font_sample = Text(frame_sample, width=20, height=20)
+        font_sample_scrollable_frame = ScrollableTextFrame(frame_sample)
```
terryjreedy: `font_sample_scrollable_frame` is too much of a mouthful and mindful for me ;-).  Shortening to `font_sample_frame`, similar to (highlight)`sample_frame`, but different enough from `frame_sample`.

Lib/idlelib/textview.py
```diff
@@ -28,52 +29,104 @@ def place(self, **kwargs):
...
+                                           takefocus=False,
+                                           command=self.text.yview)
+        self.yscroll.grid(row=0, column=1, sticky=NS)
+        self.text['yscrollcommand'] = self.yscroll.set
+
+        # horizontal scrollbar
+        self.xscroll = None
+        self._set_wrapping(self.text.cget('wrap'))
+
+    def _config_callback(self, cnf):
```
terryjreedy: I renamed this `_handle_wrap`, which is what it does.

Lib/idlelib/textview.py
```diff
@@ -28,52 +29,104 @@ def place(self, **kwargs):
...
+        if 'wrap' in cnf:
+            self._set_wrapping(cnf['wrap'])
+
+    def _set_wrapping(self, wrap):
```
terryjreedy: I set this to _set_xscroll, which is what it does.  'wrap' itself is set elsewhere.

Lib/idlelib/textview.py
```diff
@@ -28,52 +29,104 @@ def place(self, **kwargs):
...
-        self['height'] = 700
+        super().__init__(master, **kwargs)
 
-        self.text = text = Text(self, wrap=wrap, highlightthickness=0)
-        color_config(text)
-        text.grid(row=0, column=0, sticky=N+S+E+W)
+        @add_config_callback(self._config_callback)
+        class TextWithConfigHook(Text):
+            pass
+        self.text = TextWithConfigHook(self)
```
terryjreedy: Cheryl and I intentional added the local name 'text'.  Removing it has no relation to the purpose of this PR.  Writing `text = self.text = Text...` would have been and would be clearer by putting the name used in the rest of the method first.

Lib/idlelib/textview.py
```diff
@@ -28,52 +29,104 @@ def place(self, **kwargs):
...
+            self.xscroll = None
 
 
 class ViewFrame(Frame):
     "Display TextFrame and Close button."
     def __init__(self, parent, text, wrap='word'):
+        """Create a frame for viewing text with a "Close" button.
+
+        parent - parent widget for this frame
+        text - text to display
```
terryjreedy: Using `text` for both the text string to display and the text widget it is inserted in is a mistake I will fix later.  The former is rare while the latter is everywhere, so I will change the parameter name, perhaps to 'textstring', perhaps to 'source_text', 'view_text, or 'display_text'.   (configdialog, for instance, uses 'font_sample_text'.)  Any preference?

Lib/idlelib/textview.py
```diff
@@ -28,52 +29,104 @@ def place(self, **kwargs):
...
-            xscroll.grid(row=1, column=0, sticky=E+W)
+        self.yscroll = AutoHiddenScrollbar(self, orient=VERTICAL,
+                                           takefocus=False,
+                                           command=self.text.yview)
+        self.yscroll.grid(row=0, column=1, sticky=NS)
+        self.text['yscrollcommand'] = self.yscroll.set
+
+        # horizontal scrollbar
+        self.xscroll = None
+        self._set_wrapping(self.text.cget('wrap'))
```
terryjreedy: Since 'wrap' defaults to None and 'xscroll' has just been set to None, I believe that this call does nothing.

Lib/idlelib/textview.py
```diff
@@ -28,52 +29,104 @@ def place(self, **kwargs):
...
+        The Text widget is accessible via the 'text' attribute.
+        """
         super().__init__(parent)
         self.parent = parent
         self.bind('<Return>', self.ok)
         self.bind('<Escape>', self.ok)
-        self.textframe = TextFrame(self, text, wrap=wrap)
+        self.textframe = ScrollableTextFrame(self, relief=SUNKEN, height=700)
+
+        self.text = self.textframe.text
```
terryjreedy: Once the parameter name is changed, I would prefer
```
text = self.text = Scrollable...
text.configure...
text.insert...
...
```
so I don't have to read the somewhat redundant and scarcely informative, in context, 'self.' over and over.

Lib/idlelib/textview.py
```diff
@@ -28,52 +29,104 @@ def place(self, **kwargs):
...
+        self.text.configure(wrap=wrap, highlightthickness=0)
+        self.text.insert('1.0', text)
+        color_config(self.text)
+        self.text['state'] = 'disabled'
```
terryjreedy: color_config just makes a text.configure call.  When I moved this above the color_config call, the textview htest showed the proper dark theme colors.  So I believe that this can be included in the configure call above.

Lib/idlelib/textview.py
```diff
@@ -28,52 +29,104 @@ def place(self, **kwargs):
...
-    "Display text with scrollbar."
+def add_config_callback(config_callback):
+    """Class decorator adding a configuration callback for Tk widgets"""
+    def decorator(cls):
+        class WidgetWithConfigHook(cls):
+            def __setitem__(self, key, value):
+                config_callback({key: value})
+                super().__setitem__(key, value)
 
-    def __init__(self, parent, rawtext, wrap='word'):
```
terryjreedy, line 34: I think I understand why you removed 'wrap' from the signature of this class, but I think it is a mistake.  First, the current code to add scrollbars worked fine and I am rather sure it would work, with whatever minor adjustment, for the purposes of this PR.  Second, removing the parameter requires the addition of a new feature comprising about 50 lines of non-trivial code that I not only believe is not needed for this issue, but will never by needed by IDLE.  If you have evidence otherwise, it should be presented in a separate issue on scrollable widgets (which I think is needed anyway).

To put it another way, your clever code might be useful if we were writing a library for 3rd party use, but we are not.  I can't think of any reason why IDLE would want to change the wrap setting after the initial setting.  We certainly do not need to do so for this issue.

Since Texts default to wrap='char', which we are unlikely to want, I think 'wrap' should be required and not have a default value.  (added) And we should have documented originally that the wrap attribute should not be changed once set.

terryjreedy (commented):
Thank you for the quick response.  It stops the size explosion issue and with a minor tweak each, the samples look as good as they can without letting the boxes get wider (up to some limit, and at best a future issue).  (In reviewing, I found a couple of other issues with the highlight sample that I want to fix tomorrow. along with the key bindings).

My main concern is the addition of two new features (the Text subclass and its decorator and associated code) that seem to me not unneeded here.  Until persuaded otherwise, I would like the code simplified, but given the time situation, I am willing to try that in a separate PR.

Lib/idlelib/configdialog.py
taleinat: Yes please do!

I was using that just for testing and forgot to set it back to `NONE` (oops!); good catch!

Lib/idlelib/textview.py
taleinat: It was previously 'rawtext', which I dislike.

FWIW I find the variable `text` for a text widget, rather than an actual string of text, confusing. I realize that ship has sailed, though.

I suggest 'contents' or 'text_string'.

Lib/idlelib/textview.py
taleinat: As you understood, I removed this to avoid a name clash with the 'text' input parameters. Since we're changing the input parameter's name, you're welcome to re-introduce the local 'text' variable for the Text widget.

Lib/idlelib/textview.py
taleinat: In my testing, 'wrap' sometimes did not default to 'none' (a.k.a. `tkinter.NONE`), but to 'char'. This is safe and correct.

Also, you've got the logic reversed: When 'wrap' is set to 'none', **thats** when we need the horizontal scrollbar.

Lib/idlelib/textview.py
taleinat: You're right, the added complexity is likely unwarranted here. I'll revert this change.

Lib/idlelib/textview.py
taleinat: Yes, but `color_config()` is re-used several times in the codebase. Are you worried about the overhead of calling `text.configure()` twice?

Lib/idlelib/textview.py
taleinat: Regardless, with the simplification as you've requested, this is no longer relevant.

taleinat:
@terryjreedy, I've made the changes you suggested, except for the comment about `color_config()`; see my reply to that.

taleinat:
FYI, I put up a cleaned version of the generic scrollable text frame as a [gist](https://gist.github.com/taleinat/985f1d47ea73b133b2faed77bd88d928).

Lib/idlelib/textview.py
terryjreedy: I was not clear.  text.insert requires state 'normal' (but not any particular wrap or highlightthickness).  I tested whether color_config also requires state 'normal'.  It seems not. One configure call *other* than the one in color_config, which *should* be left alone,
`text.configure(wrap=wrap, highlightthickness=0, state='disabled')`
should work if after the insert.  

This is not terribly important, but I like to know when the order of actions is necessary and when it is arbitrary.

terryjreedy (approved):
Great.  I will merge when CI passes after the minor tweaks.

PR #14959, opened 2019-07-26
bpo-17535: IDLE: increase line number horizontal padding by 2 pixels

taleinat:
As suggested on the issue tracker, though I found that adding 2 pixels looks much nicer than adding just 1.

https://bugs.python.org/issue17535

### DIFF
--- a/Lib/idlelib/sidebar.py
+++ b/Lib/idlelib/sidebar.py
@@ -51,7 +51,7 @@ class BaseSideBar:
 
         _padx, pady = get_widget_padding(self.text)
         self.sidebar_text = tk.Text(self.parent, width=1, wrap=tk.NONE,
-                                    padx=0, pady=pady,
+                                    padx=2, pady=pady,
                                     borderwidth=0, highlightthickness=0)
         self.sidebar_text.config(state=tk.DISABLED)
         self.text['yscrollcommand'] = self.redirect_yscroll_event


terryjreedy:
@aeros167 
Kyle, the change is a single char, replacing 0 with something.  I am undecided on whether I prefer 1 or 2.  For me, the text has 2 pixel of padding on the left.  padx=1 in the sidebar matches this on the right of the numbers; =2 exceeds it.  Do you have a stronger opinion?  To test, edit repository idlelib.sidebar in IDLE, change line 54, and run.  When the htest box appears, after unittests, hit the button.  Try changing the initial 'a' to 'M', in the first two lines, for a different visual effect.  Revert to 0 when done.

terryjreedy:
The Azure pipelines is a timeout failure of the Mac test.  This happens on and off.  Not related.

Lib/idlelib/sidebar.py
```diff
@@ -51,7 +51,7 @@ def __init__(self, editwin):
 
         _padx, pady = get_widget_padding(self.text)
         self.sidebar_text = tk.Text(self.parent, width=1, wrap=tk.NONE,
-                                    padx=0, pady=pady,
+                                    padx=2, pady=pady,
```
aeros, line 54:
```suggestion
                                    padx=4, pady=pady,
```

aeros (approved):
@taleinat Thanks for the PR, I definitely like the look horizontal padding of 2 better than before. However, I think 4 might be a slight further improvement. To compare the two, I did a side-by-side screenshot. The one on the left has a padding of 2 and the one on the right has a padding of 4:

image

Tested on Linux using a large font size of 29 with the Arial font style so that the gap is easier to compare between the two.

Since it's a bit subjective, it might be a bit difficult to "objectively" determine which one looks better, but I personally have a preference towards the one of the right with the horizontal padding of 4. Let me know what you think. I would definitely approve of the PR either way though (:

aeros167:
@terryjreedy If the mac tests are periodically timing out, would it be helpful to increase the timeout duration beyond the present 60 to something like 80? I'm not sure what that would involve, I don't have much experience with regards to modifying integration tests.

Also, I did the review without reading the comments to your instructions, my apologies. Would you like me to do another review following your instructions or is the one that I did adequate? Between 1 and 2, I would lean towards two, but I personally have preference of padx=4. However, as I mentioned in the review, this might be quite subjective. I would consider any increase to be an improvement though.

terryjreedy:
@aeros167 If there is not an issue for Mac tests timing out, and I would search first, there should be, with Victor Stinner and Steve Dower nosey.  I suspect that the hung tests will not finish with 20 more minutes.

You images lack letters in the text for comparison.  A total padding of 3 on the number side is already more than the total 2 on the text side.  It may be the the absolute padding in real units (mm) should depend somewhat on the absolute height, in which case the pixel padding would depend on the pixels/inch or cm.  At least in some places, numbers can be given with real units.

Right now, I am concerned with getting merges done for a3, scheduled Monday.  So I am going to merge this as is.  It is better than no change.  Anything more sophisticated should be a new issue, not this closed one.

aeros167:
@terryjreedy Oh okay, that makes sense. Personally, I think just having line numbers in general is a huge improvement, and adding this change to slightly adjust the margin provides a significant improvement. 

Also, with regards there not being letters, I was primarily referring to the gap between the line numbers and the dividing line in my feedback in the BPO issue, rather than the gap between the numbers and the letters. Apologies if I wasn't clear on that.

terryjreedy:
I understand.  I was saying that in my opinion, spacing should be looked at in context, with real text.  But anyway, this is for a followup.

PR #14965, opened 2019-07-26
bpo-37690: Simplify linking of shared libraries on the AIX OS

ericvw:
Have the approach of building shared libraries on the AIX operating
system be similar to that of a System V system.  The primary benefit of
this change is the elimination of custom AIX paths and reducing the
changes at `./configure` to affect just the `LDSHARED` environment
variable.

For background context, AIX sees shared libraries as fully linked and
resolved, where symbol references are resolved at link-time and cannot
be rebound at load-time.  System V resolves all global symbols by the
run-time linker.  Thus, conventional shared libraries in AIX cannot have
undefined symbols, while System V can.

However, AIX does allow for run-time linking in allowing symbols to be
undefined until load-time.

Therefore, this change affects how linking of shared libraries are
performed on AIX to behave similarly to that of System V.

Given that symbols are now going to be allowed to be undefined for AIX,
all the code paths for generating exported symbols and the related
wrapper scripts go away.

The real magic is in the `-G` flag for `LDSHARED`.  Effectively, `-G`
is equivalent to specifying the following:

* -berok: Suppress errors even if there are unresolved symbols
* -brtl: Enable run-time linking
* -bnortllib: Do not include a reference to the run-time linker
* -bnosymbolic: Assigns 'nosymbolic' attribute to most symbols (i.e.,
                can be rebound)
* -bnoautoexp: Prevent auto exportation of any symbols
* -bM:SRE: Set the module type to reusable (i.e., require a private copy
           of the data area for each process).

https://bugs.python.org/issue37690


configure
```diff
@@ -9357,8 +9348,7 @@ if test -z "$LDSHARED"
 then
 	case $ac_sys_system/$ac_sys_release in
 	AIX*)
-		BLDSHARED="Modules/ld_so_aix \$(CC) -bI:Modules/python.exp"
-		LDSHARED="\$(LIBPL)/ld_so_aix \$(CC) -bI:\$(LIBPL)/python.exp"
+		LDSHARED='$(CC) -G'
```
ericvw, line 9351: This line is one of the important changes.

configure.ac
```diff
@@ -2513,8 +2504,7 @@ if test -z "$LDSHARED"
 then
 	case $ac_sys_system/$ac_sys_release in
 	AIX*)
-		BLDSHARED="Modules/ld_so_aix \$(CC) -bI:Modules/python.exp"
-		LDSHARED="\$(LIBPL)/ld_so_aix \$(CC) -bI:\$(LIBPL)/python.exp"
+		LDSHARED='$(CC) -G'
```
ericvw, line 2507: This line is the other important changes.

ericvw (commented):
The two lines highlighted are the important ones which allow for the removal of other AIX related code or files entirely.

ericvw:
Abandoning this change due to feedback provided in https://bugs.python.org/issue37690.

PR #14967, opened 2019-07-26
bpo-37354: Make Powershell Activate.ps1 script static to allow for signing

d3r3kk:
Make Powershell Activation.ps1 script static and therefore sign-able. The script now determines the interpreter path based on where the Activate script is. The prompt is determined based on the directory name of the parent directory the script is located in, or can be overridden via an argument to the Activate script, or from within the `pyvenv.cfg` file. 

### Tested On:
| OS                     | Powershell Version | Powershell Build Version |
|------------------------|--------------------|--------------------------|
| Windows 8.1            | 5.1.14409.1018     | 10.0.14409.1018          |
| Windows 7 SP1          | 5.1.14409.1018     | 10.0.14409.1018          |
| Windows Server 2012R2  | 4.0                | 6.3.9600.19170           |
| Windows 10             | 5.1.18362.145      | 10.0.18362.145           |
| Debian 10.0 (Buster)   | 6.2.2 (Core)       | ?                        |

https://bugs.python.org/issue37354

### DIFF
--- a/Lib/venv/scripts/common/Activate.ps1
+++ b/Lib/venv/scripts/common/Activate.ps1
@@ -163,7 +163,7 @@ $pyvenvCfg = Get-PyVenvCfgOverrides -ConfigDir $VenvDir
 
 # Next, set the prompt from the command line, or the config file, or
 # just use the name of the virtual environment folder.
-if (-not $Promp) {
+if (-not $Prompt) {
     if ($pyvenvCfg -and $pyvenvCfg['prompt']) {
         $Prompt = $pyvenvCfg['prompt'];
     }
--- a/Lib/venv/scripts/common/Activate.ps1
+++ b/Lib/venv/scripts/common/Activate.ps1
@@ -1,56 +1,211 @@
-function Script:add-bin([string]$envPath) {
-    $binPath = Join-Path -Path $env:VIRTUAL_ENV -ChildPath '__VENV_BIN_NAME__'
-    return ($binPath, $envPath) -join [IO.Path]::PathSeparator
-}
+<#
+.Synopsis
+Activate a Python virtual environment for the current Powershell session.
 
+.Description
+Pushes the python executable for a virtual environment to the front of the
+$Env:PATH environment variable and sets the prompt to signify that you are
+in a Python virtual environment. Makes use of the command line switches as
+well as the `pyvenv.cfg` file values present in the virtual environment.
+
+.Parameter VenvDir
+Path to the directory that contains the virtual environment to activate. The
+default value for this is the parent of the directory that the Activate.ps1
+script is located within.
+
+.Parameter Prompt
+The prompt prefix to display when this virtual environment is activated. By
+default, this prompt is the name of the virtual environment folder (VenvDir)
+surrounded by parenthesis and followed by a single space (ie. '(.venv) ').
+
+.Example
+Activate.ps1
+Activates the Python virtual environment that contains the Activate.ps1 script.
+
+.Example
+Activate.ps1 -Verbose
+Activates the Python virtual environment that contains the Activate.ps1 script,
+and shows extra information about the activation as it executes.
+
+.Example
+Activate.ps1 -VenvDir C:\Users\MyUser\Common\.venv
+Activates the Python virtual environment located in the specified location.
+
+.Example
+Activate.ps1 -Prompt "MyPython"
+Activates the Python virtual environment that contains the Activate.ps1 script,
+and prefixes the current prompt with the specified string (surrounded in
+parenthesis) while the virtual environment is active.
+
+
+#>
+Param(
+    [Parameter(Mandatory = $false)]
+    [String]
+    $VenvDir,
+    [Parameter(Mandatory = $false)]
+    [String]
+    $Prompt
+)
+
+<# Function declarations --------------------------------------------------- #>
+
+<#
+.Synopsis
+Remove all shell session elements added by the Activate script, including the
+addition of the virtual environment's Python executable from the beginning of
+the PATH variable.
+
+.Parameter NonDestructive
+If present, do not remove this function from the global namespace for the
+session.
+
+#>
 function global:deactivate ([switch]$NonDestructive) {
     # Revert to original values
-    if (Test-Path function:_OLD_VIRTUAL_PROMPT) {
-        copy-item function:_OLD_VIRTUAL_PROMPT function:prompt
-        remove-item function:_OLD_VIRTUAL_PROMPT
+
+    # The prior prompt:
+    if (Test-Path -Path Function:_OLD_VIRTUAL_PROMPT) {
+        Copy-Item -Path Function:_OLD_VIRTUAL_PROMPT -Destination Function:prompt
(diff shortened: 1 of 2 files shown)


d3r3kk:
Note, also fixes bpo-35667.

Lib/venv/scripts/common/Activate.ps1
```diff
@@ -1,56 +1,229 @@
...
+
+# Next, set the prompt from the command line, or the config file, or
+# just use the name of the virtual environment folder.
+if ($Prompt) {
+    Write-Verbose "Prompt specified as argument, using '$Prompt'"
+} else {
+    Write-Verbose "Prompt not specified as argument to script, checking pyvenv.cfg value"
+    if ($pyvenvCfg -and $pyvenvCfg['prompt']) {
+        Write-Verbose "  Setting based on value in pyvenv.cfg='$($pyvenvCfg['prompt'])'"
+        $Prompt = $pyvenvCfg['prompt'];
```
zooba, line 187: This is the line that gave me the most nightmares when thinking about how to do this, because the prompt value in the config file is quoted and escaped (basically, it's `repr(prompt)`), so we need to do more processing here.

d3r3kk: Thanks for the review @zooba!

Agreed that the escaped character problem is a bit tricky in the context of Powershell... `\n` is not really available in Powershell, for instance, but could be replaced with the back-tick-quote `. Finding and replacing _all types_ of escaped characters (newlines, tabs, colors?) might be a bit of a deep dark hole (there be ANSI-escaped coloured dragons there)!

Would it be appropriate to add handling of escaped characters in a different issue/PR? I don't expect that the prior version of this Activate.ps1 script handled this problem particularly well either (a brief test with the latest 3.8 beta shows a few hiccups right away).
Of course, I would be most happy to tackle that problem next!

What sample input would show the shortcoming of this implementation, so that I can write some unit tests and fix it correctly?

Just from thinking about it briefly I can think up input that contains:

- newlines (`\n`, `\r\n`)
- tabs (`\t`)
- colours? (`\u001b[31m`, ... , `\u001b[0m`)

Ideally I'd love to show something that would have the class of input we need to be concerned about working with, and have it handled at least as well as the prior implementation handled it.

A specific scenario that is trivial to setup and play with, for example the newline escaped character `\n` in the prompt:

```pwsh
PS /home/d3r3kk/dev/venvtest> ../cpython/python -m venv --prompt "\nd3r3kk\n" .newline_prompt
PS /home/d3r3kk/dev/venvtest> cat ./.newline_prompt/pyvenv.cfg
home = /home/d3r3kk/dev/cpython
include-system-site-packages = false
version = 3.9.0
prompt = '\\nd3r3kk\\n'
PS /home/d3r3kk/dev/venvtest> ./.newline_prompt/bin/Activate.ps1
(\\nd3r3kk\\n) PS /home/d3r3kk/dev/venvtest> deactivate
PS /home/d3r3kk/dev/venvtest> exit
```
...and then in bash...

```bash
d3r3kk@deb:~/dev/venvtest$ . .newline_prompt/bin/activate
(
d3r3kk
) d3r3kk@deb:~/dev/venvtest$ deactivate
d3r3kk@deb:~/dev/venvtest$ 
```

...as we can see, the `\\n` being stored in the `pyvenv.cfg` is written to the prompt verbatim in Powershell with this change (and is also in the latest 3.8 beta release). 

_**Note**_: it _is_ possible in Powershell to replace some escaped character's `\\` with a ` and it will work. (But definitely not colours, those are weird special creatures!).

Lib/venv/scripts/common/Activate.ps1
```diff
@@ -1,56 +1,229 @@
...
+
+.Parameter VenvDir
+Path to the directory that contains the virtual environment to activate. The
+default value for this is the parent of the directory that the Activate.ps1
+script is located within.
+
+.Parameter Prompt
+The prompt prefix to display when this virtual environment is activated. By
+default, this prompt is the name of the virtual environment folder (VenvDir)
+surrounded by parenthesis and followed by a single space (ie. '(.venv) ').
```
vsajip: parenthesis -> parentheses

Lib/venv/scripts/common/Activate.ps1
```diff
@@ -1,56 +1,229 @@
...
+
+.Example
+Activate.ps1 -VenvDir C:\Users\MyUser\Common\.venv
+Activates the Python virtual environment located in the specified location.
+
+.Example
+Activate.ps1 -Prompt "MyPython"
+Activates the Python virtual environment that contains the Activate.ps1 script,
+and prefixes the current prompt with the specified string (surrounded in
+parenthesis) while the virtual environment is active.
```
vsajip: parenthesis -> parentheses

Lib/venv/scripts/common/Activate.ps1
```diff
@@ -1,56 +1,229 @@
...
+then it is considered a `key = value` line. The left hand string is the key,
+the right hand is the value.
+
+If the value starts with a `'` or a `"` then the first and last character is
+stripped from the value before being captured.
+
+.Parameter ConfigDir
+Path to the directory that contains the `pyvenv.cfg` file.
+#>
+function Get-PyVenvCfgOverrides(
```
vsajip: Why `Overrides` in the name? Isn't this just getting the contents of `pyvenv.cfg`?

Lib/venv/scripts/common/Activate.ps1
```diff
@@ -1,56 +1,229 @@
...
+                if ("'""".Contains($val.Substring(0,1))) {
+                    $val = $val.Substring(1, $val.Length - 2)
+                }
+
+                $pyvenvConfig[$keyval[0]] = $val
+                Write-Verbose "Adding Key: '$($keyval[0])'='$val'"
+            }
+        }
+    }
+    return $pyvenvConfig
```
vsajip, line 149: If the file doesn't exist, what exactly is returned? Presumably not an empty map?

vsajip (commented):
I'm a PowerShell tyro, so I can't comment on some of the finer points. The handling of escapes seems orthogonal to the making of the script to be entirely static, but I concur with the comment on the issue made by @zooba : "get it out in 3.8.0b4 and see how it fares".

Lib/venv/scripts/common/Activate.ps1
d3r3kk: Agreed, it's not getting `overrides` at all. Renaming.

Lib/venv/scripts/common/Activate.ps1
d3r3kk: Actually it would return `$null` the way I'd written it, which is pretty jarring now that I re-read it!

I've altered this to show that indeed, I return an empty map if the file doesn't exist.

_(You can see where this returned value is used we do a check to ensure it isn't `$null` anyways.)_

d3r3kk:
Thanks for the review @vsajip! (and for introducing me to the term _tyro_). I've updated the PR with your suggestions.

vsajip (approved):
Nothing more to add to my earlier review.

d3r3kk:
@zooba anything else for me to address?

PR #14982, opened 2019-07-27
bpo-37689: add Path.is_relative_to() method

shihai1991:
https://bugs.python.org/issue37689

### DIFF
--- a/Doc/library/pathlib.rst
+++ b/Doc/library/pathlib.rst
@@ -539,6 +539,18 @@ Pure paths provide the following methods and properties:
       ValueError: '/etc/passwd' does not start with '/usr'
 
 
+.. method:: PurePath.is_relative_to(*other)
+
+    Compute the boolean result that this path is relative to the path
+    represented by *other* or not.
+
+      >>> p = PurePosixPath('/etc/passwd')
+      >>> p.is_relative_to('/etc')
+      True
+      >>> p.is_relative_to('/usr')
+      False
+
+
 .. method:: PurePath.with_name(name)
 
    Return a new path with the :attr:`name` changed.  If the original path
--- a/Doc/library/pathlib.rst
+++ b/Doc/library/pathlib.rst
@@ -541,14 +541,16 @@ Pure paths provide the following methods and properties:
 
 .. method:: PurePath.is_relative_to(*other)
 
-    Compute the boolean result that this path is relative to the path
-    represented by *other* or not.
+   Return the boolean result that this path is relative to *other* path.
 
-      >>> p = PurePosixPath('/etc/passwd')
+      >>> from pathlib import PurePath
+      >>> p = PurePath('/etc/passwd')
       >>> p.is_relative_to('/etc')
       True
       >>> p.is_relative_to('/usr')
       False
+
+   .. versionadded:: 3.9
 
 
 .. method:: PurePath.with_name(name)
--- a/Doc/library/pathlib.rst
+++ b/Doc/library/pathlib.rst
@@ -462,6 +462,20 @@ Pure paths provide the following methods and properties:
       True
 
 
+.. method:: PurePath.is_relative_to(*other)
+
+   Return whether or not this path is relative to the *other* path.
+
+      >>> from pathlib import PurePath
+      >>> p = PurePath('/etc/passwd')
+      >>> p.is_relative_to('/etc')
+      True
+      >>> p.is_relative_to('/usr')
+      False
+
+   .. versionadded:: 3.9
+
+
 .. method:: PurePath.is_reserved()
 
    With :class:`PureWindowsPath`, return ``True`` if the path is considered
@@ -537,20 +551,6 @@ Pure paths provide the following methods and properties:
         File "pathlib.py", line 694, in relative_to
           .format(str(self), str(formatted)))
       ValueError: '/etc/passwd' does not start with '/usr'
-
-
-.. method:: PurePath.is_relative_to(*other)
-
-   Return the boolean result that this path is relative to *other* path.
-
-      >>> from pathlib import PurePath
-      >>> p = PurePath('/etc/passwd')
-      >>> p.is_relative_to('/etc')
-      True
-      >>> p.is_relative_to('/usr')
-      False
-
-   .. versionadded:: 3.9
 
 
 .. method:: PurePath.with_name(name)
--- a/Doc/library/pathlib.rst
+++ b/Doc/library/pathlib.rst
@@ -466,7 +466,6 @@ Pure paths provide the following methods and properties:
 
    Return whether or not this path is relative to the *other* path.
 
-      >>> from pathlib import PurePath
       >>> p = PurePath('/etc/passwd')
       >>> p.is_relative_to('/etc')
       True
--- a/Doc/library/pathlib.rst
+++ b/Doc/library/pathlib.rst
@@ -273,7 +273,7 @@ Methods and properties
 
 .. testsetup::
 
-   from pathlib import PurePosixPath, PureWindowsPath
+   from pathlib import PurePath, PurePosixPath, PureWindowsPath
 
 Pure paths provide the following methods and properties:
 
--- a/Lib/pathlib.py
+++ b/Lib/pathlib.py
@@ -886,6 +886,15 @@ class PurePath(object):
         return self._from_parsed_parts('', root if n == 1 else '',
                                        abs_parts[n:])
 
+    def is_relative_to(self, *other):
+        """Return True if the path is relative to another path or False.
+        """
+        try:
+            self.relative_to(*other)
+            return True
+        except ValueError:
+            return False
+
     @property
     def parts(self):
         """An object providing sequence-like access to the
--- a/Lib/test/test_pathlib.py
+++ b/Lib/test/test_pathlib.py
@@ -619,12 +619,17 @@ class _BasePurePathTest(object):
         self.assertRaises(ValueError, p.relative_to, '')
         self.assertRaises(ValueError, p.relative_to, P('a'))
 
-    def test_is_relative_to(self):
+    def test_is_relative_to_common(self):
         P = self.cls
         p = P('a/b')
         self.assertRaises(TypeError, p.is_relative_to)
         self.assertRaises(TypeError, p.is_relative_to, b'a')
         self.assertTrue(p.is_relative_to(P()))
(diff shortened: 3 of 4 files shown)


Doc/library/pathlib.rst
```diff
@@ -539,6 +539,18 @@ Pure paths provide the following methods and properties:
       ValueError: '/etc/passwd' does not start with '/usr'
 
 
+.. method:: PurePath.is_relative_to(*other)
```
tirkarthi: Please add a `versionadded` tag.

Doc/library/pathlib.rst
```diff
@@ -539,6 +539,18 @@ Pure paths provide the following methods and properties:
...
+
+    Compute the boolean result that this path is relative to the path
+    represented by *other* or not.
+
+      >>> p = PurePosixPath('/etc/passwd')
```
aeros: I would suggest changing `PurePosixPath` to `PurePath`, `PurePath` will use `PurePosixPath` or `PureWindowsPath` based on the environment (for windows compatibility), and adding the import statement so that users can directly run the interpreter example.
```suggestion
      >>> from pathlib import PurePath
      >>> p = PurePath('/etc/passwd')
```

Misc/NEWS.d/next/Library/2019-07-27-18-00-43.bpo-37689.glEmZi.rst
```diff
@@ -0,0 +1 @@
+Add an is_relative_to() in PurePath class to compute a path is relative to the path or not.
```
aeros: Add rst inline markup and adjust the phrasing:
```suggestion
Add :meth:`is_relative_to` in :class:`PurePath` to determine whether or not one path is relative to another.
```

Doc/library/pathlib.rst
```diff
@@ -539,6 +539,18 @@ Pure paths provide the following methods and properties:
       ValueError: '/etc/passwd' does not start with '/usr'
 
 
+.. method:: PurePath.is_relative_to(*other)
+
+    Compute the boolean result that this path is relative to the path
```
aeros: Slight phrasing adjustment:
```suggestion
    Return the boolean result that this path is relative to *other* path.
```

Doc/library/pathlib.rst
```diff
@@ -539,6 +539,18 @@ Pure paths provide the following methods and properties:
...
+    represented by *other* or not.
```
aeros: In combination with the above recommendation
```suggestion
```

aeros (approved):
@shihai1991 Thanks for the contribution! I definitely think that this function would be quite helpful.

I did some testing in the interpreter after installing python through your branch:
image
OS: Arch Linux 5.2.3

Looks good to me! I just had a few minor suggestions:

Doc/library/pathlib.rst
shihai1991: thanks for review, done :)

Doc/library/pathlib.rst
shihai1991: thanks for review, done:)

Doc/library/pathlib.rst
shihai1991: you are right, show the PurePath demo would be better to user.

Misc/NEWS.d/next/Library/2019-07-27-18-00-43.bpo-37689.glEmZi.rst
shihai1991: done

shihai1991:
@pitrou Hi, Antoine. pls review this PR, if you have free time, thanks :)

aeros (approved):

aeros167:
@shihai1991 Btw, after you've made any changes based on suggestions you can use "Resolve conversation" to show that each one has been addressed (or directly commit the suggestion to the PR). Thanks for making the recommended changes (:

Doc/library/pathlib.rst
pitrou: Can you keep the list of methods in alphabetical order?

Doc/library/pathlib.rst
```diff
@@ -539,6 +539,20 @@ Pure paths provide the following methods and properties:
       ValueError: '/etc/passwd' does not start with '/usr'
 
 
+.. method:: PurePath.is_relative_to(*other)
+
+   Return the boolean result that this path is relative to *other* path.
```
pitrou: How about `Return whether this path is relative to the *other* path`?

Lib/test/test_pathlib.py
```diff
@@ -619,6 +619,21 @@ def test_relative_to_common(self):
         self.assertRaises(ValueError, p.relative_to, '')
         self.assertRaises(ValueError, p.relative_to, P('a'))
 
+    def test_is_relative_to(self):
```
pitrou: I would like to see more comprehensive testing. You can simply reuse the various cases from the `relative_to` tests.

pitrou (changes requested):
Thanks for doing this. Please find some comments below.

Doc/library/pathlib.rst
aeros: I agree, this would be an improvement over my earlier suggestion. Since it's modifying a verb, "Return" in this case, I think it might be slightly more correct to explicitly specify the "or not":

``Return whether or not this path is relative to the *other* path``.

shihai1991:
> @shihai1991 Btw, after you've made any changes based on suggestions you can use "Resolve conversation" to show that each one has been addressed (or directly commit the suggestion to the PR). Thanks for making the recommended changes (:

np, looks I need change my habit(because i am a gerrit user before:)

Doc/library/pathlib.rst
shihai1991: of course.

Doc/library/pathlib.rst
shihai1991: ops， i am not a native english speaker, so i have no idea about "whether(or not)" :(

aeros: @shihai1991 That's okay, the distinction is often a pain, even for those of us who are native speakers (:

Lib/test/test_pathlib.py
shihai1991: add all test_relative_to and test_relative_to_common test cases:)

shihai1991:

Doc/library/pathlib.rst
```diff
@@ -462,6 +462,20 @@ Pure paths provide the following methods and properties:
       True
 
 
+.. method:: PurePath.is_relative_to(*other)
```
CuriousLearner, line 465: I think to name this parameter as `other_path` would be more explicit.

Lib/pathlib.py
```diff
@@ -886,6 +886,15 @@ def relative_to(self, *other):
         return self._from_parsed_parts('', root if n == 1 else '',
                                        abs_parts[n:])
 
+    def is_relative_to(self, *other):
```
CuriousLearner, line 889: I'd suggest renaming this to `*other_path`

CuriousLearner (commented):
Overall, this looks good to me. I just have a minor suggestion on renaming the parameter as mentioned inline.

Doc/library/pathlib.rst
shihai1991: Some other functions use `*other` too, such as: PurePath.relative_to(*other), PurePath.joinpath(*other). So I think keep consistency is ok. If most people like use `other_path`, i would update it.

shihai1991: Some other functions use `*other` too, such as: PurePath.relative_to(*other), PurePath.joinpath(*other). So I think keep consistency is ok. If most people like use `other_path`, i would update it.

Lib/pathlib.py
shihai1991: ditto

Doc/library/pathlib.rst
CuriousLearner: Yes, I just checked `PurePath.relative_to` & `PurePath.joinpath`. In that case, it seems we're good on this part.

CuriousLearner (approved):
Since other methods already have `other` parameter, I think we're good to go on this.

LGTM 🌮

Doc/library/pathlib.rst
```diff
@@ -462,6 +462,20 @@ Pure paths provide the following methods and properties:
...
+
+   Return whether or not this path is relative to the *other* path.
+
+      >>> from pathlib import PurePath
```
pitrou: Nit: can you remove the import? Other snippets don't show the import.

pitrou (changes requested):
Thanks for the update! Just one additional comment and I think we're good to go.

Doc/library/pathlib.rst
aeros: What are the general guidelines for imports in interpreter examples? In the `pathlib` documentation, does anything from the `pathlib` module not require an import?

pitrou: I don't know if there are official guidelines (you may ask on the ML). Personally, I think imports in interpreter snippets are most of the time a distraction, especially when there are tens of snippets on a single documentation page.

Generally speaking, I expect the user to be able to make out that the `PurePath` class listed in the documentation for the `pathlib` module needs the corresponding import before being used.

aeros: >Generally speaking, I expect the user to be able to make out that the PurePath class listed in the documentation for the pathlib module needs the corresponding import before being used.

@pitrou Thanks for the clarification, I think that's perfectly reasonable and agree that an excessive number of imports can be distracting. Unless someone else disagrees with your rationale, I don't think that a new ML thread is needed.

I think that the explicit imports are primarily appropriate when importing modules that are external to the documentation file. This can be see in one of the existing examples already in ``pathlib.rst``, when ``os`` is imported:
```
>>> import os
>>> p = PurePath('/etc')
>>> os.fspath(p)
'/etc'
```

pitrou: > I think that the explicit imports are primarily appropriate when importing modules that are external to the documentation file.

Agreed.

shihai1991: guys, thanks :)

shihai1991:

CuriousLearner (approved):

pitrou (approved):

pitrou:
Thank you @shihai1991 !

PR #14983, opened 2019-07-27
bpo-37692: Improve highlight config sample

terryjreedy:
Use an example shell interaction in the sample and better labels for shell elements.

https://bugs.python.org/issue37692

### DIFF
--- a/Lib/idlelib/configdialog.py
+++ b/Lib/idlelib/configdialog.py
@@ -867,8 +867,8 @@ class HighPage(Frame):
             ('  breakpoint("line")', 'break'), ('\n\n', 'normal'),
             ('>>>', 'console'), (' 3.14**2\n', 'normal'),
             ('9.8596', 'stdout'), ('\n', 'normal'),
-            ('>>>', 'console'), (' a ', 'normal'),
-            ('b', 'error'), ('\n', 'normal'),
+            ('>>>', 'console'), (' py ', 'normal'),
+            ('t', 'error'), ('hon\n', 'normal'),
             ('SyntaxError', 'stderr'), ('\n', 'normal'))
         for string, tag in string_tags:
             text.insert(END, string, tag)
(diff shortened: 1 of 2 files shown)


terryjreedy:
I intend to merge this tomorrow (Sunday).  Any comments before that?

I am pretty happy with the simulated shell session.  I tried putting the cursor blob where the blank line above the prompt is, but that blank is.  I will try putting it up under code context, as it has no particular assoication with Shell.

As for the label changes.  My special concern is with beginners.

 'Normal Text' is mainly used for code unless one is editing a text file, so 'Normal Code of Text'. 

'Shell Normal Text', is technically correct (and confusing), since it is used for normal output from Shell itself (but not, now, for Restart lines).   But what is that?  The prompt.  And '[DEBUG ONOFF]', which beginners will likely not have seen.  So 'Shell Prompt' seems better.

'Shell Error Text' is wrong in that it is mainly (exclusively?) used to mark syntax error locations in *both* shell and editor.  So just 'Error Text' is better.

'Shell Stdout Text' is *not* the stdout of Shell.  It is used for text the users send to stdout, usually by print, usually by default, usually in a separate process which gets displayed in Shell.  Beginners might not even be familiar with 'stdout' (and 'stderr')  'Shell User Output' seems better even if not the best possible.  In any case, the output that clearly results from user input should make its meaning clear.  

The case with 'Shell Stderr Text' is similar.

taleinat:
@terryjreedy, I trust your judgement, and this looks alright.

My only minor note is that typing `python` in the shell isn't common; perhaps use a more realistic example such as `print`?

terryjreedy:
@Mariatta I routinely backport to 3.8 and 3.7 and the 3.8 backport routinely fails on the first try, and succeeds on the second.  There clearly is some problem with doing two backports in the same invocation.  Sometimes the failure is almost immediate (the last two times before this), sometime delayed after the timeout.

terryjreedy:
Tal, obviously real code with a simulated accidental space syntax error looks better.

PR #14985, opened 2019-07-28
bpo-36044: Avoid warnings in Windows PGO build and add lzma, bz2 and sqlite coverage

zooba:
https://bugs.python.org/issue36044


gpshead:
I doubt test_bz2 and test_lzma and test_sqlite add much given the important thing is compiling the bz2 and lzma and sqlite3 libraries themselves with PGO (which isn't part of our build).  but these tests in total only take ~11 seconds so it doesn't hurt.

gpshead (approved):

PR #14987, opened 2019-07-28
Remove trailing .0 from version changed note

ncoghlan:
The convention is to omit the trailing `.0` from version changed notes.

(Found while looking for documented changes and additions in minor
releases)


ncoghlan:
I found a few like this for 3.6.0 as well, but that's long enough ago now that it doesn't seem worth fixing.

PR #14988, opened 2019-07-28
gh-47169: Use the symbolic errno name in OSError error messages

ZackerySpytz:
For example:
```
>>> open('')
FileNotFoundError: [Errno ENOENT] No such file or directory: ''
```
`ENOENT` is shown instead of the integer value.

https://bugs.python.org/issue2920

* Issue: gh-47169

### DIFF
--- a/Lib/test/test_exceptions.py
+++ b/Lib/test/test_exceptions.py
@@ -1285,6 +1285,13 @@ class ExceptionTests(unittest.TestCase):
                 next(i)
                 next(i)
 
+    def test_OSError_errno_error_message(self):
+        # The symbolic errno name should be shown in the error message of
+        # OSError and its subclasses.
+        with self.assertRaisesRegex(OSError,
+                f'Errno {errno.errorcode[errno.ENOENT]}'):
+            open('non-existent')
+
 
 class ImportErrorTests(unittest.TestCase):
 
--- a/Lib/test/test_exceptions.py
+++ b/Lib/test/test_exceptions.py
@@ -292,8 +292,7 @@ class ExceptionTests(unittest.TestCase):
             w = OSError(9, 'foo', 'bar')
             self.assertEqual(w.errno, 9)
             self.assertEqual(w.winerror, None)
-            self.assertEqual(str(w),
-                f"[Errno {errno.errorcode[errno.EBADF]}] foo: 'bar'")
+            self.assertEqual(str(w), "[Errno EBADF] foo: 'bar'")
             # ERROR_PATH_NOT_FOUND (win error 3) becomes ENOENT (2)
             w = OSError(0, 'foo', 'bar', 3)
             self.assertEqual(w.errno, 2)
@@ -1289,8 +1288,7 @@ class ExceptionTests(unittest.TestCase):
     def test_OSError_errno_error_message(self):
         # The symbolic errno name should be shown in the error message of
         # OSError and its subclasses.
-        with self.assertRaisesRegex(OSError,
-                f'Errno {errno.errorcode[errno.ENOENT]}'):
+        with self.assertRaisesRegex(OSError, 'Errno ENOENT'):
             open('non-existent')
 
 
--- a/Lib/test/test_exceptions.py
+++ b/Lib/test/test_exceptions.py
@@ -292,7 +292,8 @@ class ExceptionTests(unittest.TestCase):
             w = OSError(9, 'foo', 'bar')
             self.assertEqual(w.errno, 9)
             self.assertEqual(w.winerror, None)
-            self.assertEqual(str(w), "[Errno 9] foo: 'bar'")
+            self.assertEqual(str(w),
+                f"[Errno {errno.errorcode[errno.EBADF]}] foo: 'bar'")
             # ERROR_PATH_NOT_FOUND (win error 3) becomes ENOENT (2)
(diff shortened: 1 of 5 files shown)


Lib/test/test_exceptions.py
```diff
@@ -292,7 +292,8 @@ def test_WindowsError(self):
             w = OSError(9, 'foo', 'bar')
             self.assertEqual(w.errno, 9)
             self.assertEqual(w.winerror, None)
-            self.assertEqual(str(w), "[Errno 9] foo: 'bar'")
+            self.assertEqual(str(w),
+                f"[Errno {errno.errorcode[errno.EBADF]}] foo: 'bar'")
```
eryksun: Why didn't you just use `"[Errno EBADF] foo: 'bar'"` instead of getting "EBADF" from the mapping?

ZackerySpytz: Hmm...

Objects/exceptions.c
```diff
@@ -1115,23 +1115,51 @@ OSError_str(PyOSErrorObject *self)
                                     self->winerror ? self->winerror: Py_None,
                                     self->strerror ? self->strerror: Py_None);
 #endif
+    PyObject *str = self->myerrno;
```
gnprice: This variable name is confusing, because `myerrno` isn't a string.

Objects/exceptions.c
```diff
@@ -1115,23 +1115,51 @@ OSError_str(PyOSErrorObject *self)
...
+    static PyObject *errorcodedict;
+    // Try to obtain the symbolic errno name (e.g., ENOENT).
+    if (str != NULL) {
+        if (errorcodedict == NULL) {
+            PyObject *errnomod = PyImport_ImportModule("errno");
+            if (errnomod == NULL) {
+                PyErr_Clear();
+                goto skip;
```
gnprice: I think this logic would be clearer by avoiding `goto`.

Probably a good way to do that is to pull this whole block into its own function.

gnprice (commented):
Thanks -- this will be a useful feature!

The patch calls `PyImport_ImportModule("errno")`, the equivalent of `import errno`, to get the table of names. In that bpo discussion, several people were concerned about doing an import here. It'd be good to say a few words about how this patch addresses those concerns.

Objects/exceptions.c
eryksun: Using `goto` is fine, but the label should describe the target, such as `format_errno`. Just `skip` isn't meaningful. 

There are some generic labels used for well-known targets such as  `cleanup`, `error`, `fail`, `success`, `done`, and `exit`. I wouldn't include `skip` in that list. It would be helpful if PEP 7 discussed the use of `goto` and recommended label names.

gnprice: Hmm. I think most likely such a discussion would say not to use `goto` here. :slightly_smiling_face:

There's one very useful way of using `goto` in C which is done widely in the codebase: namely, to clean something up or tie up a loose end just before returning from a function. It serves much like a substitute for `finally:`.

I think in fact all the common label names you mention are examples of the names one might use in that use case. :wink:

Other cases where `goto` is worth it are quite rare, and I don't think this is one of them.

eryksun: Point taken. Jumping to a non-final label is rare in CPython's original C code. (One place I've seen it is in `tok_get` in Parser/tokenizer.c.) Jumping to the label in this case didn't bother me because it's a single step forward, as opposed to spaghetti that twists my brain into a knot. That said, adding a function to get the error name would improve readability.

gnprice: That is in fact exactly the example that came to my mind too. :-) It does contribute to making that logic rather tricky to understand; but of course the inner loop of the tokenizer is quite a hot path, where saving a few cycles potentially justifies some virtuoso loop-optimization tricks. When I looked hard at that code a couple of years ago, I remember thinking that it was one of the rare cases where a non-final goto makes sense.

Even there, who knows; computers have changed a lot since Guido wrote that code, in 1990 or so. (In the early history, the loop looks pretty much like it does now.) It's possible that something more straightforward might be preferable today.

ZackerySpytz: I will use `goto` in `OSError_str()` unless a core dev says not to.

gnprice:
Hmm, GitHub's UI now hides this comment [above](https://github.com/python/cpython/pull/14988#discussion_r308594545), so I'll make it again here in the main thread:
```
            if (errnomod == NULL) {
                PyErr_Clear();
                goto formaterrno;
```

I think this logic would be clearer by avoiding `goto`.

Probably a good way to do that is to pull this whole block into its own function.

(In discussion, I added:)

There's one very useful way of using `goto` in C which is done widely in the codebase: namely, to clean something up or tie up a loose end just before returning from a function. It serves much like a substitute for `finally:`.

Other cases where `goto` is worth it are quite rare, and I don't think this is one of them.

---

@ZackerySpytz , I see your comment suggesting you'd like to hear from a core developer on this question; wanted to pull this out just so the thread isn't lost in the GitHub mists. :wink:

MaxwellDupre (approved):
For changed test the results seen:
Ran 54 tests in 1.967s
OK (skipped=1)
Ran 54 tests in 1.967s
OK (skipped=1)
Ran 52 tests in 25.364s
OK (skipped=4)
Ran 20 tests in 1.104s
OK
Looks ok.

PR #14998, opened 2019-07-29
bpo-36002: fail to find -llvm-profdata

weibullguy:
The llvm-profdata is added using the AC_PATH_TARGET_TOOL in configure.ac.  Per the autoconf documentation for the AC_PATH_TARGET_TOOL, "If the tool cannot be found with a prefix, and if the build and target types are equal, then it is also searched for without a prefix."  However, only the AC_CANONICAL_HOST macro is called in the configure.ac script which means target will be set to '' while host and build are set to the values produced by config.guess.  Thus, when configure attempts to find $target_alias-llvm-profdata, it will search for -llvm-profdata (note leading dash), which will not be found on probably any system.

However, the macro AC_PATH_TOOL will search for the un-prefixed tool (llvm-profdata in this case) if a prefixed version is not found on the path.  This pull request replaces the AC_PATH_TARGET_TOOL macro with the AC_PATH_TOOL macro in configure.ac in two places (for llvm-profdata and llvm-ar).

With this change, Python built and tested satisfactorily using gcc v5.5.0 and clang v3.8.0, both with the --target switch set and without.

https://bugs.python.org/issue36002

### DIFF
--- a/Misc/NEWS.d/next/Build/2019-09-13-14-12-36.bpo-36002.Bcl4oe.rst
+++ b/Misc/NEWS.d/next/Build/2019-09-13-14-12-36.bpo-36002.Bcl4oe.rst
@@ -1,0 +1,2 @@
+Locate ``llvm-profdata`` and ``llvm-ar`` binaries using ``AC_PATH_TOOL``
+rather than ``AC_PATH_TARGET_TOOL``.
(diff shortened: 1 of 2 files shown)


the-knights-who-say-ni:

benjaminp:
Is there a reason we shouldn't fix this problem by calling `AC_CANONICAL_TARGET`?

benjaminp:
I'm accepting this on the basis that `llvm-ar` and `llmv-profdata` should be host tools not target tools.

PR #14999, opened 2019-07-29
bpo-37702: Fix SSL certificate-store-handles leak

neonene:
In Windows, CertCloseStore(hCollectionStore,[any option]) closes only hCollectionStore.
To avoid out-of-memory, CertCloseStore() shuld be called to each hSystemStore
which was added to hCollectionStore.

I think CERT_CLOSE_STORE_FORCE_FLAG is rare option and not much needed.
MSDN says CertCloseStore(handle, 0) is ideal and works fine for me.

https://bugs.python.org/issue37702

### DIFF
--- a/Misc/NEWS.d/next/Windows/2019-07-29-16-49-31.bpo-37702.Lj2f5e.rst
+++ b/Misc/NEWS.d/next/Windows/2019-07-29-16-49-31.bpo-37702.Lj2f5e.rst
@@ -1,0 +1,1 @@
+Fix memory leak in ssl certification.
--- a/Misc/NEWS.d/next/Windows/2019-07-29-16-49-31.bpo-37702.Lj2f5e.rst
+++ b/Misc/NEWS.d/next/Windows/2019-07-29-16-49-31.bpo-37702.Lj2f5e.rst
@@ -1,1 +1,2 @@
-Fix memory leak in ssl certification.
+Fix memory leak on Windows in creating an SSLContext object or
+running urllib.request.urlopen('https://...').
--- a/Modules/_ssl.c
+++ b/Modules/_ssl.c
@@ -5544,7 +5544,9 @@ parseKeyUsage(PCCERT_CONTEXT pCertCtx, DWORD flags)
 }
 
 static HCERTSTORE
-ssl_collect_certificates(const char *store_name)
+ssl_collect_certificates(const char *store_name,
+                         DWORD *system_stores, size_t system_stores_count,
+                         HCERTSTORE *system_store_handles)
 {
 /* this function collects the system certificate stores listed in
  * system_stores into a collection certificate store for being
@@ -5553,14 +5555,6 @@ ssl_collect_certificates(const char *store_name)
  */
 
     HCERTSTORE hCollectionStore = NULL, hSystemStore = NULL;
-    static DWORD system_stores[] = {
-        CERT_SYSTEM_STORE_LOCAL_MACHINE,
-        CERT_SYSTEM_STORE_LOCAL_MACHINE_ENTERPRISE,
-        CERT_SYSTEM_STORE_LOCAL_MACHINE_GROUP_POLICY,
-        CERT_SYSTEM_STORE_CURRENT_USER,
-        CERT_SYSTEM_STORE_CURRENT_USER_GROUP_POLICY,
-        CERT_SYSTEM_STORE_SERVICES,
-        CERT_SYSTEM_STORE_USERS};
     size_t i, storesAdded;
     BOOL result;
 
@@ -5570,12 +5564,14 @@ ssl_collect_certificates(const char *store_name)
         return NULL;
     }
     storesAdded = 0;
-    for (i = 0; i < sizeof(system_stores) / sizeof(DWORD); i++) {
+    for (i = 0; i < system_stores_count; i++) {
+        system_store_handles[i] = NULL;
         hSystemStore = CertOpenStore(CERT_STORE_PROV_SYSTEM_A, 0,
                                      (HCRYPTPROV)NULL,
                                      CERT_STORE_READONLY_FLAG |
                                      system_stores[i], store_name);
         if (hSystemStore) {
+            system_store_handles[i] = hSystemStore;
             result = CertAddStoreToCollection(hCollectionStore, hSystemStore,
                                      CERT_PHYSICAL_STORE_ADD_ENABLE_FLAG, 0);
             if (result) {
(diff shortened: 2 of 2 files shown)


the-knights-who-say-ni:

zooba:
The theory seems fine. I haven't checked the implementation - can't say when I'll get to it, but I'll try and do it before the next release.

Modules/_ssl.c
```diff
@@ -5767,10 +5801,20 @@ _ssl_enum_crls_impl(PyObject *module, const char *store_name)
...
 
-    /* CERT_CLOSE_STORE_FORCE_FLAG forces freeing of memory for all contexts
-       associated with the store, in this case our collection store and the
-       associated system stores. */
-    if (!CertCloseStore(hCollectionStore, CERT_CLOSE_STORE_FORCE_FLAG)) {
+    /* When a collection store and its sibling stores are closed
+       with CertCloseStore using CERT_CLOSE_STORE_FORCE_FLAG,
+       the collection store must be closed before its sibling stores.
+      (https://docs.microsoft.com/en-us/windows/win32/api/wincrypt/nf-wincrypt-certaddstoretocollection) */
+    success = CertCloseStore(hCollectionStore, CERT_CLOSE_STORE_FORCE_FLAG);
```
epicfaace: You could remove the `BOOL success;` declaration before and do this?:
```suggestion
    BOOL success = CertCloseStore(hCollectionStore, CERT_CLOSE_STORE_FORCE_FLAG);
```

Misc/NEWS.d/next/Windows/2019-07-29-16-49-31.bpo-37702.Lj2f5e.rst
```diff
@@ -0,0 +1 @@
+Fix memory leak in ssl certification.
```
epicfaace: It might be good to specify which methods this applies to, and that it is only on Windows.

neonene: I'd rather not describe c-implementation level information.

Modules/_ssl.c
```diff
@@ -5767,10 +5799,20 @@ _ssl_enum_crls_impl(PyObject *module, const char *store_name)
     Py_XDECREF(enc);
     Py_XDECREF(tup);
 
-    /* CERT_CLOSE_STORE_FORCE_FLAG forces freeing of memory for all contexts
-       associated with the store, in this case our collection store and the
-       associated system stores. */
-    if (!CertCloseStore(hCollectionStore, CERT_CLOSE_STORE_FORCE_FLAG)) {
+    /* When a collection store and its sibling stores are closed
```
epicfaace: Do you think it would be good to keep in the definition of CERT_CLOSE_STORE_FORCE_FLAG ("forces freeing of memory for all contexts associated with the store")?

neonene: Thanks for Bool position.
For now, both 0 and CERT_CLOSE_STORE_FORCE_FLAG  work fine for my pc.
I'd like to keep changes to a minimum and leave it to other ssl experts' judgement whether the flag should revert to 0 or not.

epicfaace: Oh, I didn't mean reverting the flag to 0, I meant, did you want to keep the part in the comment that says "CERT_CLOSE_STORE_FORCE_FLAG forces freeing of memory for all contexts associated with the store"?

neonene: MS says,
CERT_CLOSE_STORE_FORCE_FLAG:
Forces the freeing of memory for all contexts associated with the store.
This flag can be safely used only when the store is opened in a function
and neither the store handle nor any of its contexts are passed to
any called functions. For details, see ...  

I think, keeping short sentences caused and will cause a misunderstanding that
this rarely used flag is always safe and releases all certificate-store-handles.
Having said that, I'm not in favor of long description in the source file.

neonene:
MS  doesn't recommend using CERT_CLOSE_STORE_FORCE_FLAG.
https://blogs.msdn.microsoft.com/winsdk/2010/01/29/passing-the-flag-cert_close_store_force_flag-to-certclosestore-may-cause-your-application-to-crash/

And I don't want to leave confusing comments on the flag that is no longer necessary to close all handles.
So, I decided to change the flag back to 0.

neonene:
I'm done. Sorry for a dirty PR.

Modules/_ssl.c
```diff
@@ -5626,12 +5624,25 @@ _ssl_enum_certificates_impl(PyObject *module, const char *store_name)
     PCCERT_CONTEXT pCertCtx = NULL;
     PyObject *keyusage = NULL, *cert = NULL, *enc = NULL, *tup = NULL;
     PyObject *result = NULL;
+    static DWORD system_stores[] = {
```
tiran: I don't like code duplication. Please make the system store list a global static

Modules/_ssl.c
```diff
@@ -5626,12 +5624,25 @@ _ssl_enum_certificates_impl(PyObject *module, const char *store_name)
...
+        CERT_SYSTEM_STORE_LOCAL_MACHINE,
+        CERT_SYSTEM_STORE_LOCAL_MACHINE_ENTERPRISE,
+        CERT_SYSTEM_STORE_LOCAL_MACHINE_GROUP_POLICY,
+        CERT_SYSTEM_STORE_CURRENT_USER,
+        CERT_SYSTEM_STORE_CURRENT_USER_GROUP_POLICY,
+        CERT_SYSTEM_STORE_SERVICES,
+        CERT_SYSTEM_STORE_USERS};
+    size_t i, system_stores_count = sizeof(system_stores) / sizeof(DWORD);
```
tiran: This style is confusing and hard to read. Split ``size_t i`` and ``size_t system_stores_count`` into two lines.

Modules/_ssl.c
```diff
@@ -5686,15 +5697,21 @@ _ssl_enum_certificates_impl(PyObject *module, const char *store_name)
     Py_XDECREF(keyusage);
     Py_XDECREF(tup);
 
-    /* CERT_CLOSE_STORE_FORCE_FLAG forces freeing of memory for all contexts
-       associated with the store, in this case our collection store and the
-       associated system stores. */
-    if (!CertCloseStore(hCollectionStore, CERT_CLOSE_STORE_FORCE_FLAG)) {
+    /* CertCloseStore() with any argument can't close handles together.
+       In case of using CERT_CLOSE_STORE_FORCE_FLAG (not recommended),
```
tiran: The comment is out of date. You are referencing ``CERT_CLOSE_STORE_FORCE_FLAG`` but no longer using ``CERT_CLOSE_STORE_FORCE_FLAG`` as flag.

tiran (changes requested):

neonene:
Please check.
I'm relieved my flag change is not rejected.

neonene:

neonene:
Oh, I should have enclosed #define in brackets.

neonene:
By any chance, Is this PR for Windows needs further review from zooba?
Thanks to tiran and epicfaace, I think it have become readable.

neonene:
I raised another PR, which fix store leaks rather than handle leaks.
https://github.com/python/cpython/pull/15632

zooba:
I'm closing this as it's unnecessary with the other fix (I verified there is no handle leak or memory leak with `urlopen` anymore). Thanks for your efforts! They're appreciated.

PR #15008, opened 2019-07-29
Improve wording.

jugmac00:
"Arguments may be integers... " could be misunderstand as they also could be strings.

New wording makes it clear, that arguments have to be integers.

### DIFF
--- a/Doc/library/datetime.rst
+++ b/Doc/library/datetime.rst
@@ -401,7 +401,7 @@ systems.
 
 .. class:: date(year, month, day)
 
-   All arguments are required.  Arguments may be integers, in the following
+   All arguments are required.  Arguments have to be integers in the following
    ranges:
 
    * ``MINYEAR <= year <= MAXYEAR``


the-knights-who-say-ni:

ned-deily:
@jugmac00, thanks for your suggested change and contributing to Python.  Some comments on your proposed change:

1. I suggest using "must be" rather than "have to be": it's more compact and is used much more often in the documentation;
2. There are two other almost identical usages elsewhere on the datetime page that should also be changed: for datetime and time.  (All three were changed a long time ago in 5c10664043448f1d27d7ce7eb67c2afe3f54b018).
3. Our normal process for making changes to code and documentation is to first make the PR against the master branch.  Once the PR is reviewed and accepted, the core developer handling the PR will decide whether to backport to existing release branches like 3.7.  The [devguide](https://devguide.python.org/docquality/) goes into more detail.

Since the change in the PR is simple to recreate but changing a PR from one branch to another is error-prone, I am going to close this PR.  Please resubmit it with the suggested changes as a new PR against the master branch (and reference this PR in the description so you don't have to copy everything).  Thanks!

jugmac00:
@ned-deily Thank you very much for taking the time to give me that kind of detailed feedback!

PR #15009, opened 2019-07-29
bpo-37707: Exclude expensive unit tests from PGO task

nascheme:
Mark some individual tests to skip when --pgo is used.  The tests
marked increase the PGO task time significantly and likely don't
help improve optimization of the final executable.

Marked tests:

    pickletester
      check_frame_opcodes
      test_framing_many_objects

    test_bz2
      testCompress4G
      testDecompress4G

    test_itertools
      test_long_chain_of_empty_iterables

    test_lzma
      test_compressor_bigmem
      test_decompressor_bigmem

    test_statistics
      test_inv_cdf

https://bugs.python.org/issue37707

### DIFF
--- a/Lib/test/libregrtest/main.py
+++ b/Lib/test/libregrtest/main.py
@@ -643,6 +643,7 @@ class Regrtest:
             input("Press any key to continue...")
 
         support.PGO = self.ns.pgo
+        support.PGO_EXTENDED = self.ns.pgo_extended
 
         setup_tests(self.ns)
 
--- a/Lib/test/pickletester.py
+++ b/Lib/test/pickletester.py
@@ -2281,6 +2281,7 @@ class AbstractPickleTests(unittest.TestCase):
     FRAME_SIZE_MIN = 4
     FRAME_SIZE_TARGET = 64 * 1024
 
+    @support.skip_if_pgo_task
     def check_frame_opcodes(self, pickled):
         """
         Check the arguments of FRAME opcodes in a protocol 4+ pickle.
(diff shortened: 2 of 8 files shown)


gpshead (approved):
Nice!  Such a pgo skip decorator approach is exactly what I was thinking of recommending when you brought this up. :)

zooba:
I haven't looked through exactly what these tests do, but I'd say it's worth evaluating ("guessing") whether they represent a common scenario or not. Even if they run for a long time, if it's something that's going to run a lot in user code then I'd rather keep it in.

99% of realistic Python runtime ("evaluated"/"guessed") is success cases, but a lot of the quick tests in the suite are going to exercise setup/teardown and failure paths. Let's not over-focus on those for the sake of a quick PGO run.

nascheme:
I intended that --pgo will give you an optimized build without too much extra build time.  Before PR #14702, the PGO builds took so long that in practice it seems that many did not bother to use them.  If you don't care at all about long build time, then --pgo-extended is the option for that.  If we make expensive tests run in the --pgo case, I don't see the point of --pgo-extended.

Something I was thinking about previously, --pgo-extended should not not skip the tests marked with @skip_if_pgo_task.  That way, people who want the most extensive code coverage can still use --pgo-extended to get the old behavior.

nascheme:
Here is a comparison of test run time before and after my skip decorator.

test | before [s] | after [s]                                                           
-----|--------|------                                                           
test_pickle | 4.86 | 1.79                                                         
test_bz2 | 1.98 | 0.51                                                            
test_itertools | 1.51 | 1.12                                                      
test_lzma | 2.65 | 0.19                                                           
test_statistics | 1.84 | 0.57

pablogsal:
> I haven't looked through exactly what these tests do, but I'd say it's worth evaluating ("guessing") whether they represent a common scenario or not.

I did two PGO builds with and `pyperformance` shows no significant difference (all differences were <1%). I am using core isolation and affinity for them so I trust the stability of the result.

PR #15017, opened 2019-07-30
bpo-37268: Add deprecation notice and a DeprecationWarning for the parser module

pablogsal:
Deprecate the parser module and add a deprecation warning triggered on import and a warning block in the documentation.

https://bugs.python.org/issue37268

### DIFF
--- a/Lib/test/test_parser.py
+++ b/Lib/test/test_parser.py
@@ -6,6 +6,7 @@ import operator
 import struct
 from test import support
 from test.support.script_helper import assert_python_failure
+from test.support.script_helper import assert_python_ok
 
 #
 #  First, we test that we can generate trees from valid source fragments,
@@ -986,6 +987,14 @@ class OtherParserCase(unittest.TestCase):
         # See bug #12264
         with self.assertRaises(TypeError):
             parser.expr("a", "b")
+
+
+class TestDeprecation(unittest.TestCase):
+    def test_deprecation_message(self):
+        code = "def f():\n  import parser\n\nf()"
+        rc, out, err = assert_python_ok('-c', code)
+        self.assertIn(b'<string>:2: DeprecationWarning', err)
+
 
 if __name__ == "__main__":
     unittest.main()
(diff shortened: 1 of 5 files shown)


pablogsal:
Let's deprecate this once and for all! :crossed_swords:

Modules/parsermodule.c
```diff
@@ -1158,6 +1158,12 @@ PyInit_parser(void)
 {
     PyObject *module, *copyreg;
 
+    if (PyErr_WarnEx(PyExc_DeprecationWarning,
+            "The parser module is deprecated and will be removed "
+            "in future versions of Python", 7) != 0) {
```
pablogsal, line 1163: 7 is the magic number to skip all `importlib._bootstrap_external` (tested in `test_deprecation_message`). :upside_down_face:

Lib/test/test_parser.py
```diff
@@ -987,5 +988,13 @@ def test_two_args_to_expr(self):
         with self.assertRaises(TypeError):
             parser.expr("a", "b")
 
+
+class TestDeprecation(unittest.TestCase):
+    def test_deprecation_message(self):
+        code = "def f():\n  import parser\n\nf()"
+        rc, out, err = assert_python_ok('-c', code)
+        self.assertIn(b'<string>:2: DeprecationWarning', err)
```
vstinner, line 996: Can you use support.check_warnings() with support.import_fresh_module() to check the warning? I don't think so since we cannot reload a C extension twice.

vstinner (commented):
You should document the deprecation around https://docs.python.org/dev/whatsnew/3.9.html#deprecated

pablogsal:
> You should document the deprecation around https://docs.python.org/dev/whatsnew/3.9.html#deprecated

Done!

Lib/test/test_parser.py
pablogsal: Yeah, I don't think it works because C extension module initialization code can only be run once.

vstinner (approved):
LGTM.

PR #15018, opened 2019-07-30
Fix typos in comments, docs and test names

minho42:


Lib/test/test_email/test_email.py
```diff
@@ -1026,14 +1026,14 @@ def test_multiple_splittable_leading_char_followed_by_overlong_unsplitable(self)
 , ,
  but_the_second_part_does_not_fit_within_maxlinelen_and_thus_should_be_on_a_line_all_by_itself""")
 
-    def test_trailing_splitable_on_overlong_unsplitable(self):
+    def test_trailing_splitable_on_overlong_unsplittable(self):
```
terryjreedy: The first 'splitable' needs a double t also.

Lib/test/test_email/test_email.py
```diff
@@ -1026,14 +1026,14 @@ def test_multiple_splittable_leading_char_followed_by_overlong_unsplitable(self)
...
         eq = self.ndiffAssertEqual
         h = Header('this_part_does_not_fit_within_maxlinelen_and_thus_should_'
             'be_on_a_line_all_by_itself;')
         eq(h.encode(), "this_part_does_not_fit_within_maxlinelen_and_thus_should_"
             "be_on_a_line_all_by_itself;")
 
-    def test_trailing_splitable_on_overlong_unsplitable_with_leading_splitable(self):
+    def test_trailing_splitable_on_overlong_unsplittable_with_leading_splitable(self):
```
terryjreedy:
```suggestion
    def test_trailing_splittable_on_overlong_unsplittable_with_leading_splittable(self):
```

Lib/test/test_email/test_email.py
terryjreedy:
```suggestion
    def test_trailing_splittable_on_overlong_unsplittable(self):
```

Objects/object.c
```diff
@@ -2072,7 +2072,7 @@ _PyTrash_thread_deposit_object(PyObject *op)
     tstate->trash_delete_later = op;
 }
 
-/* Dealloccate all the objects in the _PyTrash_delete_later list.  Called when
+/* Dealloccte all the objects in the _PyTrash_delete_later list.  Called when
```
terryjreedy:
```suggestion
/* Deallocate all the objects in the _PyTrash_delete_later list.  Called when
```

terryjreedy:
Travis failure had something to do with checksum and clinic.

Error in file "./Modules/posixmodule.c" on line 14543:
Checksum mismatch!
Expected: b3ae8afd275ea5cd
Computed: 1e001c855e011720
Suggested fix: remove all generated code including the end marker,
or use the '-f' option.
Makefile:573: recipe for target 'clinic' failed
make: *** [clinic] Error 255

My guess is that it has to do with correction of `identifiedy` to `identified` on line 2820 of posixmodule.c.h and a need to regenerate something after the fix.  Someone else who knows more will have to deal with this.
--------------------------------------

The Appveyor build succeeded, the tests failed with
FAIL: test_study2 (idlelib.idle_test.test_pyparse.PyParseTest) (string="'''This is a multiline continuation docstring.\n\n")
  File "C:\projects\cpython\lib\idlelib\idle_test\test_pyparse.py", line 252, in test_study2
    eq(p.stmt_end, test.end)
AssertionError: 48 != 49
The correction of `continutation` to `continuation` on line 209 shorted the string from 49 to 48, so on the next line which has ` 0, 49, "'", None, ((0, 0), (0, 1), (49, 0))),`, 49 has to be changed to 48 twice.

The same comment applies to all 4 Pipelines failures.  I will make this change and trigger a retest.

encukou:
> need to regenerate something after the fix

`Tools/clinic/clinic.py -f ./Modules/posixmodule.c` should regenerate that.

Lib/test/test_email/test_email.py
minho42: Thanks.

terryjreedy:
On my Windows machine, I ran the command Petr Viktorin suggested to fix the Travis failure (with `python` prepended) and committed and pushed the result.  There was some fiddy note about line ending and regenerating git index, but the diff looks right in relation to the failure message.  We'll see.

terryjreedy:
@minho42 @aeros167 There is again the question of backports.  Kyle, are you willing to try it again, or should I do the idlelib changes separately.  (I read elsewhere that you would be doing other stuff for awhile.)

To do backports, it would have been better to have run the spell checking on 3.8 and 3.7 branches also and used the results to submit 3 separate PRs, one for master only, once for master + 3.8, and one for all 3.

aeros167:
@terryjreedy I'm currently out of town on vacation without access to any development utilities (otherwise I'd end up spending the whole vacation reviewing PRs, haha). I'm still responding to any comments directly addressed to me and emails, but I won't be able to submit anything new.

I can look into backporting the changes in about a couple of weeks when I'm back at home, but you could probably do the idle fixes separately in the meantime. Thanks for mentioning me.

The other stuff I've been working on has mostly involved the devguide, specifically with writing sections for the new Python triage team (on github, not exclusively bpo as it was before) and making changes to the experts index. 

I don't mind helping with this as well once I'm back home, I'll be on break for a few weeks between semesters until the 25th so I'll have plenty of time for Python contributions (:

terryjreedy:
Idlelib typos in #15029.  Will delete them here.

terryjreedy:
@aeros167 3.8.0b3 released yesterday.  b4 is due 8/26, 3.7.5 late Sept.  I will merge this when CI passes and you can do whatever whenever.  Take a break and enjoy yourself in the meanwhile.

minho42:
Thanks.

PR #15020, opened 2019-07-30
bpo-37631: Append $(EXTRA_CFLAGS) to $(CFLAGS_NODIST) as well

hroncok:
Previously, the $(EXTRA_CFLAGS) were only appended to $(CFLAGS),
hence any conflicting $(CFLAGS_NODIST) would override them,
as $(CFLAGS_NODIST) is appended to $(CFLAGS).

For example `CFLAGS_NODIST="-O2" EXTRA_CFLAGS="-Og"` would result in optimized
build instead of the debug one.

https://bugs.python.org/issue37631


mangrisano (approved):
LGTM.

PR #6384, opened 2018-04-05
bpo-15817: Expose command documentation to gdb help

smontanaro:
https://bugs.python.org/issue15817

PR #15021, opened 2019-07-30
bpo-15817: gdbinit: Document commands after defining them

The-Compiler:
The [gdb manual](https://sourceware.org/gdb/current/onlinedocs/gdb/Define.html) says the following for "document":

>  The command *commandname* must already be defined.

And indeed when trying to use the gdbinit file with gdb 8.3, I get:

> .../cpython/Misc/gdbinit:17: Error in sourced command file:
> Undefined command: "pyo".  Try "help".

Fix this by moving all documentation blocks after the define blocks.

This was introduced in GH-6384 (cc @smontanaro / @abalkin).

(Note that the diff algorithm gets a bit confused by those changes - try running `git diff --histogram` for a more understandable output. Also, note I took the freedom to re-use the issue number of the original implementation, hope that's okay.)

https://bugs.python.org/issue15817

### DIFF
--- a/Misc/gdbinit
+++ b/Misc/gdbinit
@@ -14,30 +14,27 @@
 # with embedded macros that you may find superior to what is in here.
 # See Tools/gdb/libpython.py and http://bugs.python.org/issue8032.
 
-document pyo
-  Prints a representation of the object to stderr, along with the
-  number of reference counts it currently has and the hex address the
-  object is allocated at.  The argument must be a PyObject*
-end
 define pyo
     # side effect of calling _PyObject_Dump is to dump the object's
     # info - assigning just prevents gdb from printing the
     # NULL return value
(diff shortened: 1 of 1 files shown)


ammaraskar (approved):
Can recreate this issue on GDB 7.7.1 and this patch fixes the problem. I'm guessing the patch was tested on an old version of gdb that didn't require the command to exist.

Long term, I question the utility of maintaining the `gdbinit` file considering it was broken for a year and no one complained.

zware:
I trust @ammaraskar's review of this; @smontonaro/@abalkin, I'll merge this tomorrow if you don't object before then.

zware:
Thank you for the patch, @The-Compiler, and thanks for the review @ammaraskar!

PR #15044, opened 2019-07-31
bpo-32178: Fix IndexError trying to parse 'To' header starting with ':'.

maxking:
This should fix the IndexError trying to retrieve `DisplayName.display_name` and `DisplayName.value` when the `value` is basically an empty string.

https://bugs.python.org/issue32178

### DIFF
--- a/Lib/email/_header_value_parser.py
+++ b/Lib/email/_header_value_parser.py
@@ -561,6 +561,8 @@ class DisplayName(Phrase):
     @property
     def display_name(self):
         res = TokenList(self)
+        if len(res) == 0:
+            return res.value
         if res[0].token_type == 'cfws':
             res.pop(0)
         else:
@@ -582,7 +584,7 @@ class DisplayName(Phrase):
             for x in self:
                 if x.token_type == 'quoted-string':
                     quote = True
-        if quote:
+        if self and quote:
             pre = post = ''
             if self[0].token_type=='cfws' or self[0][0].token_type=='cfws':
                 pre = ' '
--- a/Lib/email/_header_value_parser.py
+++ b/Lib/email/_header_value_parser.py
@@ -584,7 +584,7 @@ class DisplayName(Phrase):
             for x in self:
                 if x.token_type == 'quoted-string':
                     quote = True
-        if self and quote:
+        if len(self) != 0 and quote:
             pre = post = ''
             if self[0].token_type=='cfws' or self[0][0].token_type=='cfws':
                 pre = ' '
--- a/Lib/test/test_email/test__header_value_parser.py
+++ b/Lib/test/test_email/test__header_value_parser.py
@@ -1700,6 +1700,12 @@ class TestParser(TestParserMixin, TestEmailBase):
         self.assertEqual(display_name[3].comments, ['with trailing comment'])
         self.assertEqual(display_name.display_name, 'simple phrase.')
 
+    def test_get_display_name_bpo_32178(self):
+        display_name = self._test_get_x(
+            parser.get_display_name,
+            ':Foo ', '', '', [errors.InvalidHeaderDefect], ':Foo ')
+        self.assertEqual(display_name.value, '')
+
     # get_name_addr
 
     def test_get_name_addr_angle_addr_only(self):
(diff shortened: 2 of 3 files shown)


kushaldas:
I will review this.

kushaldas (approved):
Works as suggested. Tested on Debian Stretch. Thank you @maxking :rainbow:

Lib/email/_header_value_parser.py
```diff
@@ -582,7 +584,7 @@ def value(self):
             for x in self:
                 if x.token_type == 'quoted-string':
                     quote = True
-        if quote:
+        if self and quote:
```
matrixise: in which case, `self` would be `None|False|0`?

maxking: I think it should be `len(self) != 0` for being more explicit, like in the method above, I'll make the changes.

This would be when the `TokenList` is empty with no tokens.

Lib/test/test_email/test__header_value_parser.py
```diff
@@ -1700,6 +1700,12 @@ def test_get_display_name_ending_with_obsolete(self):
         self.assertEqual(display_name[3].comments, ['with trailing comment'])
         self.assertEqual(display_name.display_name, 'simple phrase.')
 
+    def test_get_display_name_bpo_32178(self):
```
CuriousLearner: I don't think we generally add a suffix in the test case name referring to the issue. Is there a reason for adding this in?

maxking: There is really no hard and fast rule for or against it IIUC, but I couldn't come up with a better name to differentiate this test case :)

CuriousLearner: Well, I guess we can name this: 

```suggestion
    def test_get_display_name_for_invalid_address_fields(self):
```

CuriousLearner: What do you think about this?

maxking: Thanks, I updated the name and added a comment to link to bpo instead.

Misc/NEWS.d/next/Library/2019-07-30-22-41-05.bpo-32178.X-IFLe.rst
```diff
@@ -0,0 +1 @@
+Fix IndexError trying to parse invalid address fields starting with ``:``.
```
CuriousLearner:
```suggestion
Fix IndexError in :mod:`email` package when trying to parse invalid address fields starting with ``:``.
```

CuriousLearner (commented):
Overall looks good. Just one more minor suggestion. Please find my suggestion inline.

Lib/test/test_email/test__header_value_parser.py
```diff
@@ -1700,6 +1700,14 @@ def test_get_display_name_ending_with_obsolete(self):
         self.assertEqual(display_name[3].comments, ['with trailing comment'])
         self.assertEqual(display_name.display_name, 'simple phrase.')
 
+    def test_get_display_name_for_invalid_address_field(self):
+        # bpo-32178: Test that address fields starting with `:` don't cause
```
CuriousLearner, line 1704: Since every test essentially tests something, it is redundant to mention it in a comment starting with `check if` or `test that` and the likes. I'd suggest reframing this to:

```suggestion
        # bpo-32178: Address fields starting with `:` shouldn't cause an
```

CuriousLearner (commented):
You may re-run the Travis tests and they should pass. Just one more nit-pick.

Overall LGTM 🌮  💯

Lib/test/test_email/test__header_value_parser.py
maxking: Leading comments in test cases usually tries to answer "What does this test do?" which naturally tend to end up being "Test if" or "test that" and I disagree a few redundant extra words in the comments really matter so much. DRY principle doesn't usually apply to comments (in some cases, I'd argue reverse DRY applies to comments, but I digress :-).

I am going to skip this suggestion (and another rebuild) this time, but I'll keep this in mind next time, thanks!

PR #15050, opened 2019-07-31
bpo-37729: gc: write stats at once

methane:
gc used several PySys_WriteStderr() calls to write stats.
This caused stats mixed up when stderr is shared by multiple
threads or processes.

```
gc: collecting generation 0...
gc: objects in each generation: 782 0 7759
gc: objects in permanent generation: 0gc: collecting generation 0...

gc: objects in each generation: 782 0 7764
gc: objects in permanent generation: 0gc: done, 781 unreachable, 0 uncollectable, 0.0001s elapsed

gc: done, 781 unreachable, 0 uncollectable, 0.0000s elapsed
```

https://bugs.python.org/issue37729

### DIFF
--- a/Modules/gcmodule.c
+++ b/Modules/gcmodule.c
@@ -962,11 +962,11 @@ clear_freelists(void)
     (void)PyContext_ClearFreeList();
 }
 
-// This function is called at start of collect() if DEBUG_STATS is enabled.
+// Show stats for objects in each gennerations.
 static void
-show_stats(struct _gc_runtime_state *state, int generation)
+show_stats_each_generations(struct _gc_runtime_state *state)
 {
-    char buf[512];
+    char buf[100];
     size_t pos = 0;
 
     for (int i = 0; i < NUM_GENERATIONS && pos < sizeof(buf); i++) {
@@ -975,10 +975,9 @@ show_stats(struct _gc_runtime_state *state, int generation)
     }
 
     PySys_WriteStderr(
-        "gc: collecting generation %d...\n"
         "gc: objects in each generation:%s\n"
         "gc: objects in permanent generation: %zd\n",
-        generation, buf, gc_list_size(&state->permanent_generation.head));
+        buf, gc_list_size(&state->permanent_generation.head));
 }
 
 /* This is the main function.  Read this to understand how the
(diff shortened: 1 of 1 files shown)


Modules/gcmodule.c
```diff
@@ -1103,16 +1113,16 @@ collect(struct _gc_runtime_state *state, int generation,
...
-        if (m == 0 && n == 0)
-            PySys_WriteStderr("gc: done");
-        else
-            PySys_FormatStderr(
-                "gc: done, %zd unreachable, %zd uncollectable",
-                n+m, n);
-        PySys_WriteStderr(", %.4fs elapsed\n",
-                          _PyTime_AsSecondsDouble(t2 - t1));
+        if (m == 0 && n == 0) {
+            PySys_WriteStderr("gc: done, %.4fs elapsed\n", d);
```
methane: Is this special case worth?

Modules/gcmodule.c
```diff
@@ -962,6 +962,24 @@ clear_freelists(void)
...
+{
+    char buf[100];
+    size_t pos = 0;
+
+    for (int i = 0; i < NUM_GENERATIONS && pos < sizeof(buf); i++) {
+        pos += PyOS_snprintf(buf+pos, sizeof(buf)-pos,
+                             " %zd", gc_list_size(GEN_HEAD(state, i)));
+    }
+
+    PySys_WriteStderr(
```
methane: We need to use `PySys_FormatStderr`  if `%zd` format is not portable yet.
I sent mail about it to Python-dev.

Modules/gcmodule.c
```diff
@@ -962,6 +962,24 @@ clear_freelists(void)
...
+// Show stats for objects in each gennerations.
+static void
+show_stats_each_generations(struct _gc_runtime_state *state)
+{
+    char buf[100];
+    size_t pos = 0;
+
+    for (int i = 0; i < NUM_GENERATIONS && pos < sizeof(buf); i++) {
+        pos += PyOS_snprintf(buf+pos, sizeof(buf)-pos,
+                             " %zd", gc_list_size(GEN_HEAD(state, i)));
```
methane: ditto.  Can we use `%zd` here?

PR #15051, opened 2019-07-31
Fix punctuation in `os.execvpe` docstring.

hramezani:
Fix punctuation in `os.execvpe` docstring.

### DIFF
--- a/Lib/os.py
+++ b/Lib/os.py
@@ -568,7 +568,7 @@ def execvpe(file, args, env):
     """execvpe(file, args, env)
 
     Execute the executable file (which is searched for along $PATH)
-    with argument list args and environment env , replacing the
+    with argument list args and environment env, replacing the
     current process.
     args may be a list or tuple of strings. """
     _execvpe(file, args, env)


the-knights-who-say-ni:

mangrisano (approved):
Thank you.

Lib/os.py
```diff
@@ -568,7 +568,7 @@ def execvpe(file, args, env):
     """execvpe(file, args, env)
 
     Execute the executable file (which is searched for along $PATH)
-    with argument list args and environment env , replacing the
+    with argument list args and environment env, replacing the
     current process.
     args may be a list or tuple of strings. """
```
hansrajdas, line 573: nit: Please remove extra space from end of doc string also.

hramezani: We have this extra space in all function docstring. for example:
https://github.com/python/cpython/blob/24fe46081be3d1c01b3d21cb39bc3492ab4485a3/Lib/os.py#L532
https://github.com/python/cpython/blob/24fe46081be3d1c01b3d21cb39bc3492ab4485a3/Lib/os.py#L555

hansrajdas: Strange, don't know this is intentional or what. In that case, we should take this trailing space removal in a separate PR. Thank you!

hansrajdas (approved):

PR #15052, opened 2019-07-31
bpo-37726: prefer argparse over getopt

mental32:
[bpo-37726](https://bugs.python.org/issue37726): prefer argparse over getopt in tutorial

The tutorial section on "command line arguments" recommended using getopt for argument parsing and referenced argparse as the fancy alternative.

The topic was raised in the python-ideas by Andrew and Guido created the issue and tagged it with the new "newcomer friendly" tag.

This is my first doc change so I'd greatly appreciate a review and especially criticism, if I did something wrong point it out so I can avoid repeating it.

https://bugs.python.org/issue37726

### DIFF
--- a/Doc/tutorial/stdlib.rst
+++ b/Doc/tutorial/stdlib.rst
@@ -86,7 +86,7 @@ Take, for example, the below snippet of code::
    >>> greeting = ["Hi", "Hello", "Greetings! its very nice to meet you"][args.verbose % 3]
    >>> print(f'{greeting}, {args.name}')
    >>> if not args.verbose:
-   >>>     print('Try running this again with multiple `-v` flags!')
+   >>>     print('Try running this again with multiple "-v" flags!')
 
 .. _tut-stderr:
 
--- a/Doc/tutorial/stdlib.rst
+++ b/Doc/tutorial/stdlib.rst
@@ -78,29 +78,15 @@ It should always be preferred over directly processing ``sys.argv`` manually.
 Take, for example, the below snippet of code::
 
    >>> import argparse
-   >>> parser = argparse.ArgumentParser(description='An argparse example.')
-   >>> parser.add_argument('name', help='The name of someone to greet.')
-   >>> args = parser.parse_args()
-   >>> print(f'Hello, {args.name}!')
-
-The three lines of code needed for creating the command line argument formula
-are simple to read and understand and are easy to extend in terms of your
-applications functionality.
-
-What if you wanted to provide a default value?
-
-::
-
    >>> from getpass import getuser
+   >>> parser = argparse.ArgumentParser(description='An argparse example.')
    >>> parser.add_argument('name', nargs='?', default=getuser(), help='The name of someone to greet.')
-
-How about if you wanted to control how verbose you are with your output?
-
-::
-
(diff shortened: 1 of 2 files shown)


Doc/tutorial/stdlib.rst
```diff
@@ -72,10 +72,35 @@ three`` at the command line::
    >>> print(sys.argv)
    ['demo.py', 'one', 'two', 'three']
 
-The :mod:`getopt` module processes *sys.argv* using the conventions of the Unix
-:func:`getopt` function.  More powerful and flexible command line processing is
-provided by the :mod:`argparse` module.
+The :mod:`argparse` module provides a mechanism to process command line arguments,
```
isidentical: I think sentence should end with `.`

gvanrossum (approved):
Thanks! I fixed the punctuation and otherwise I'm happy with this. I'll wait 24 hours for other core devs to jump in; if there's no more feedback I'll merge it.

isidentical (approved):

Doc/tutorial/stdlib.rst
```diff
@@ -72,10 +72,35 @@ three`` at the command line::
...
+
+   >>> import argparse
+   >>> parser = argparse.ArgumentParser(description='An argparse example.')
+   >>> parser.add_argument('name', help='The name of someone to greet.')
+   >>> args = parser.parse_args()
+   >>> print(f'Hello, {args.name}!')
+
+The three lines of code needed for creating the command line argument formula
+are simple to read and understand and are easy to extend in terms of your
+applications functionality.
```
ericvsmith: I think you should just let the example stand, instead of asserting that it's simple to read and extend.

Doc/tutorial/stdlib.rst
```diff
@@ -72,10 +72,35 @@ three`` at the command line::
...
+   >>> from getpass import getuser
+   >>> parser.add_argument('name', nargs='?', default=getuser(), help='The name of someone to greet.')
+
+How about if you wanted to control how verbose you are with your output?
+
+::
+
+  >>> parser.add_argument('--verbose', '-v', action='count')
+  >>> greeting = ["Hi", "Hello", "Greetings! its very nice to meet you"][args.verbose % 3]
+  >>> print(f'{greeting}, {args.name}')
```
ericvsmith: I think you should just include all of these examples in with the first example, and only have a single example, without any additional text. That seems to be how it's done in the rest of the document.

ericvsmith (commented):
Thanks for your contribution!
See my comments about keeping with the same style as the rest of the document.

gvanrossum:
@mental32 Do you have time to make the changes suggested by Eric? The seem a good idea to me.

mental32:
@gvanrossum Indeed I do, changes will be pushed in the next twenty minutes :)

Doc/tutorial/stdlib.rst
mental32: I was attempting to make obvious my point, but if the examples can stand on their own and have the same effect then I'll gladly omit the statement :+1:

matrixise (approved):
Thank you for your contribution, I think we can merge this PR.

gvanrossum:
Thank you @mental32. Your contributions makes a difference!

PR #9959, opened 2018-10-18
bpo-28053: Complete and fix custom reducers in multiprocessing.

pablogsal:
This PR tries to complete and fix the implementation of the custom reducer classes in `multiprocessing`. 

**Important**

I have marked the PR as `DO-NOT-MERGE` because I have still several doubts about the previous implemented API, regarding the `AbstractReducer` base class and the methods that the user needs to implement and how the rest of the library interacts with `multiprocessing.reducer`. For example:

1. I am not sure `multiprocessing.reducer.dumps` and `multiprocessing.reducer.register` are needed outside the `ForklingPickler` class and how that interacts with the ABC.

2. I am not sure the `AbstractReducer` is implemented completely (there is no abstract methods marked).

This PR is a **draft** implementation of the complete API, tests and documentation so we can discuss how to implement these correctly in a better way. 

https://bugs.python.org/issue28053

PR #15058, opened 2019-07-31
gh-72240: Allow custom reducer when using multiprocessing

pierreglaser:
This PR is a follow up on #9959. It implements the suggestions I made here https://github.com/python/cpython/pull/9959#issuecomment-513305975.

I'm making it a new PR to get the CI working now and gather separate feedback.

- [x] pass the current `context`  to `Process` objects
- [x] split `AbstractPickler` into `AbstractPickler` and `AbstractUnpickler`
- [x] clean everything up (still WIP)

https://bugs.python.org/issue28053

* Issue: gh-72240

### DIFF
--- a/Doc/library/multiprocessing.rst
+++ b/Doc/library/multiprocessing.rst
@@ -1245,7 +1245,7 @@ version 2 to be able to communicate with a Python 2.x programs.::
            return super().dumps(obj, protocol=pickle_protocol)
 
    class PickleProtocol2Reducer(AbstractReducer):
-       def get_pickler_class(self):
+       def get_pickler(self):
            return ForkingPicklerProtocol2
 
    multiprocessing.set_reducer(PickleProtocol2Reducer)
--- a/Doc/library/multiprocessing.rst
+++ b/Doc/library/multiprocessing.rst
@@ -1216,18 +1216,18 @@ the reduction mechanism:
    Abstract base class that can be implemented in order to replace the standard
    reduction mechanism used in multiprocessing
 
-   .. function:: get_pickler_class():
+   .. method:: get_pickler_class():
 
-      This method must return an instance of a subclass of :class:`pickler.Pickler`
-      to be used by the multiprocessing reducer mechanism.
+      This method must return an subclass of :class:`pickler.Pickler` to be used by
+      the multiprocessing reducer mechanism.
 
 .. currentmodule:: multiprocessing
 
 .. method:: set_reducer(reduction)
 
-   Sets a reduction class to be used for serialization and deserialization by the module
-   primitive internals. **reduction** must be an instance of a subclass of
-   :class:`multiprocessing.reduction.AbstractReducer`.
+   Sets a reduction instance to be used for serialization and deserialization
+   by the module primitive internals. **reduction** must be an instance of a
+   subclass of :class:`multiprocessing.reduction.AbstractReducer`.
 
 .. method:: get_reducer()
 
@@ -1241,14 +1241,14 @@ version 2 to be able to communicate with a Python 2.x programs.::
 
    class ForkingPicklerProtocol2(ForkingPickler):
        @classmethod
-       def dumps(cls, obj, pickle_protocol=2):
-           return super().dumps(obj, protocol=pickle_protocol)
+       def dumps(cls, obj, protocol=2):
+           return super().dumps(obj, protocol=protocol)
 
    class PickleProtocol2Reducer(AbstractReducer):
        def get_pickler_class(self):
            return ForkingPicklerProtocol2
 
-   multiprocessing.set_reducer(PickleProtocol2Reducer)
+   multiprocessing.set_reducer(PickleProtocol2Reducer())
 
 Notice that using :meth:`multiprocessing.set_reducer` changes the reducer globally. If
 changing this setting globally is undesirable you could call :meth:`context.set_reducer`,
--- a/Doc/library/multiprocessing.rst
+++ b/Doc/library/multiprocessing.rst
@@ -1216,7 +1216,7 @@ the reduction mechanism:
    Abstract base class that can be implemented in order to replace the standard
    reduction mechanism used in multiprocessing
 
-   .. function:: get_pickler():
+   .. function:: get_pickler_class():
 
       This method must return an instance of a subclass of :class:`pickler.Pickler`
       to be used by the multiprocessing reducer mechanism.
@@ -1245,7 +1245,7 @@ version 2 to be able to communicate with a Python 2.x programs.::
            return super().dumps(obj, protocol=pickle_protocol)
 
    class PickleProtocol2Reducer(AbstractReducer):
-       def get_pickler(self):
+       def get_pickler_class(self):
            return ForkingPicklerProtocol2
 
    multiprocessing.set_reducer(PickleProtocol2Reducer)
--- a/Doc/library/multiprocessing.rst
+++ b/Doc/library/multiprocessing.rst
@@ -1212,29 +1212,48 @@ the reduction mechanism:
 .. currentmodule:: multiprocessing.reduction
 
 .. class:: AbstractPickler()
-   Abstract base class that can be implemented in order to override specidifc
-   methods of the reduction machinery used by multiprocessing.
+
+   Base class that can be implemented in order to override
+   serialization methods of the reduction machinery used by multiprocessing.
 
    .. method:: dump(obj)
+
       Write a pickled representation of obj to the open file.
 
-      Defaults to :meth:`pickle.Pickle.dump`
+      Defaults to :meth:`pickle.Pickler.dump`
 
-   .. classmethod:: loads(bytes_object, *, fix_imports=True, encoding="ASCII", errors="strict")
-      Read a pickled object hierarchy from a bytes object and return the reconstituted
-      object hierarchy specified therein.
+.. class:: AbstractUnpickler(bytes_object, *, fix_imports=True, encoding="ASCII", errors="strict")
 
-      Defaults to :func:`pickle.loads`
+   Base class that can be implemented in order to override
+   multiprocessing's default unserialization mechanism.
+
+   .. method:: load()
+
+      Read a pickled object hierarchy from the open file and return the
+      reconstituted object hierarchy specified therein.
+
+      Defaults to :meth:`pickle.Unpickler.load`
 
 .. class:: AbstractReducer()
 
-   Abstract base class that can be implemented in order to replace the standard
-   reduction mechanism used in multiprocessing
+   Base class providing access to custom ``Pickler`` and ``Unpickler``
+   classes to be used by ``multiprocessing`` when serializing objects.
 
    .. method:: get_pickler_class():
 
-      This method must return an subclass of :class:`pickler.Pickler` to be used by
-      the multiprocessing reducer mechanism.
+      This method must return an subclass of :class:`AbstractPickler` to be used
+      by ``multiprocessing`` when serializing objects.
+
+   .. method:: get_unpickler_class():
+
+      This method must return an subclass of :class:`AbstractUnpickler` to be used
+      by ``multiprocessing`` when unserializing objects.
+
+
+Note that both methods of the :class:`AbstractReducer` are optional. If
+:meth:`get_pickler_class` (resp.  :meth:`get_unpickler_class`) is not
+implemented, multiprocessing will fallback to the :class:`pickle.Pickler`
+(resp. :class:`pickle.Unpickler`) class to serialize objects.
 
 .. currentmodule:: multiprocessing
 
@@ -1242,7 +1261,7 @@ the reduction mechanism:
 
    Sets a reduction instance to be used for serialization and deserialization
    by the module primitive internals. **reduction** must be an instance of a
-   subclass of :class:`multiprocessing.reduction.AbstractReducer`.
+   subclass of :class:`AbstractReducer`.
 
 .. method:: get_reducer()
 
--- a/Doc/library/multiprocessing.rst
+++ b/Doc/library/multiprocessing.rst
@@ -1211,6 +1211,21 @@ the reduction mechanism:
 
 .. currentmodule:: multiprocessing.reduction
 
+.. class:: AbstractPickler()
+   Abstract base class that can be implemented in order to override specidifc
+   methods of the reduction machinery used by multiprocessing.
+
+   .. method:: dump(obj)
+      Write a pickled representation of obj to the open file.
+
+      Defaults to :meth:`pickle.Pickle.dump`
+
+   .. classmethod:: loads(bytes_object, *, fix_imports=True, encoding="ASCII", errors="strict")
+      Read a pickled object hierarchy from a bytes object and return the reconstituted
+      object hierarchy specified therein.
+
+      Defaults to :func:`pickle.loads`
+
 .. class:: AbstractReducer()
 
    Abstract base class that can be implemented in order to replace the standard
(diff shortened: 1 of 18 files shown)


pierreglaser:
You will notice that for the `Process` class, I chose to pass the `reducer` (which is the only bit of information we need from the context for now) to the Process class at construction instead of the `Context` 
```python
    def process(self, *args, **kwargs):
        return self._Process(*args, reducer=self.get_reducer(), **kwargs)
```
This comes from the fact that the`Process` is actually relying on the default context to be properly defined:
https://github.com/python/cpython/blob/7397cda99795a4a8d96193d710105e77a07b7411/Lib/multiprocessing/context.py#L220-L224

For the process class, passing the `context` object a `Process` bind to would require changing it to something like:
```python
class Process(process.BaseProcess):
    _start_method = None
    @staticmethod
    def _Popen(process_obj):
          if self._ctx is not None:
              return self._ctx.get_context()._Process._Popen(process_obj)
          return _default_context.get_context()._Process._Popen(process_obj)
```
with `_Process` being a `Process` class specified (or not, in this case the default is `Process`) by the user. You can see the infinite recursion here as if  `_Process` is not specified or set to `Process`, `self._ctx.get_context()._Process._Popen` is `Process._Popen`... To prevent this situation we would need to add a bunch of guards, and this ends up being not so trivial.

So implementation-wise, it is easier to simply pass the `reducer` in `BaseContext.process`, and not the whole context. Also, one noticeable consequence is that the reducer in `Process` is static, as it is not called using something like `self._ctx.get_reducer()` as it is done elsewhere.

But I am opened to suggestions.

Lib/multiprocessing/context.py
```diff
@@ -46,6 +51,18 @@ def cpu_count(self):
...
             return num
 
+    @property
+    def Process(self):
+        if self._reducer is None:
+            # Ensure backward compatibility by returning a class when no
+            # custom reducer was specified
+            return self._Process
+        else:
+            return self.process
```
pppery:
```suggestion
            return self._process
```
See below

Lib/multiprocessing/context.py
```diff
@@ -46,6 +51,18 @@ def cpu_count(self):
...
+
+    def process(self, *args, **kwargs):
```
pppery:
```suggestion
    def _process(self, *args, **kwargs):
```
This attribute should not be public.

Lib/multiprocessing/context.py
```diff
@@ -200,15 +227,34 @@ def get_start_method(self, allow_none=False):
...
+
+    def set_reducer(self, reduction):
+        if reduction is original_reducer:
+            self._reducer = None
+            return
 
-    @reducer.setter
-    def reducer(self, reduction):
-        globals()['reduction'] = reduction
+        if not isinstance(reduction, original_reducer.AbstractReducer):
```
pppery, line 241: The checks below aren't strict enough; they pass for a `.register()`ed subclass of `Abstract<Reducer|Pickler|Unpickler>`, but the code won't actually work with such a class due to depending on undocumented APIs.

Lib/multiprocessing/forkserver.py
```diff
@@ -10,8 +10,9 @@
 
 from . import connection
 from . import process
-from .context import reduction
 from . import resource_tracker
+from . import context
+from .context import reduction
```
pppery: Unused import

Lib/multiprocessing/context.py
pierreglaser: I do not understand what you mean by "Undocumented API". This PR comes with documentation for the `AbstractReducer`, `AbstractPickler`, and `AbstractUnpickler`.

pierreglaser:
@pitrou what could I do to ease the review process of this PR?

Lib/multiprocessing/context.py
pppery: What I mean is that the documentation suggests that the below code should work:
```python
class DummyReducer:
    def get_pickler_class():
        return pickle.Pickler
    def get_unpickler_class():
        return pickle.Unpickler
AbstractReducer.register(DummyReducer)
multiprocessing.set_reducer(DummyReducer())
```
However, in practice, it appears that running that code (and then doing something that uses reduction) will produce a (rather cryptic) error message in some cases. Either the documentation should be clear that virtual subclasses don't work and the code should raise an error when a virtual subclass is passed, or the code should be updated to handle virtual subclasses.

pitrou:
@pierreglaser The first necessary condition is for me to find some time... sorry :-/

Lib/multiprocessing/context.py
pierreglaser: > What I mean is that the documentation suggests that the below code should work

Not really: the `get_pickler_class` API documentation specifies that `get_pickler_class` must return an`AbstractPickler` subclass, but your `DummyReducer` (which itself, should be an `AbstractReducer` subclass) simply returns a `Pickler`, which is not an `AbstractPickler` subclass. Or maybe I misunderstood your message?

pierreglaser:
@pitrou no worries :)

Lib/multiprocessing/context.py
```diff
@@ -1,10 +1,14 @@
 import os
+import pickle
+import _pickle
```
pppery: Niether of these imports appear to be used.

Lib/multiprocessing/context.py
pppery: You didn't misunderstand anything, my example was accidentally bogus, and I misunderstood the code.  The actual bug I was trying to point out appears not to exist with reducers, but instead with picklers.
```python
class DummyReducer(AbstractReducer):
    def get_pickler_class():
        return DummyPickler
class DummyPickler(pickle.Pickler):
    pass
AbstractPickler.register(DummyPickler)
multiprocessing.set_reducer(DummyReducer())
```
This seemingly harmless code appears to have the effect of breaking the passing of all of the custom types that `multiprocessing` registers, in an entirely non-intuitive way, which I think should either be fixed or documented.

Lib/multiprocessing/context.py
```diff
@@ -316,13 +362,13 @@ def _check_available(self):
     class SpawnProcess(process.BaseProcess):
         _start_method = 'spawn'
         @staticmethod
-        def _Popen(process_obj):
+        def _Popen(process_obj,):
```
pppery: This change appears to have nothing to do with the patch

Lib/multiprocessing/reduction.py
```diff
@@ -248,34 +273,61 @@ def _rebuild_socket(df, family, type, proto):
...
 
-class AbstractReducer(metaclass=ABCMeta):
+class AbstractReducer(metaclass=abc.ABCMeta):
     '''Abstract base class for use in implementing a Reduction class
     suitable for use in replacing the standard reduction mechanism
     used in multiprocessing.'''
-    ForkingPickler = ForkingPickler
-    register = register
-    dump = dump
     send_handle = send_handle
```
pppery, line 280: None of the attributes defined between here and `__init__` appear to be used.

Lib/multiprocessing/reduction.py
```diff
@@ -248,34 +273,61 @@ def _rebuild_socket(df, family, type, proto):
...
 
     def __init__(self, *args):
         register(type(_C().f), _reduce_method)
         register(type(list.append), _reduce_method_descriptor)
         register(type(int.__add__), _reduce_method_descriptor)
         register(functools.partial, _reduce_partial)
         register(socket.socket, _reduce_socket)
+
+    @property
+    def HAVE_SEND_HANDLE(self):
```
pppery, line 306: This property does not appear to be used.

Lib/multiprocessing/reduction.py
```diff
@@ -248,34 +273,61 @@ def _rebuild_socket(df, family, type, proto):
...
+                 hasattr(socket, 'SCM_RIGHTS') and
+                 hasattr(socket.socket, 'sendmsg')))
+
+    def get_pickler_class(self):
+        return AbstractPickler
+
+    def get_unpickler_class(self):
+        return AbstractUnpickler
+
+    def register(self, type_, reduce_):
```
pppery, line 319: This method (a) doesn't appear to be used and (b) doesn't necessarily work, because nothing requires the `get_pickler_class` to define a `register` method.

pierreglaser:
(sorry, failed rebasing attempt)

Lib/multiprocessing/reduction.py
pierreglaser: This is work taken from @pablogsal previous commits (see #9959). I assume there was a good reason for this, but he'll probably know better than me.

Lib/multiprocessing/reduction.py
pierreglaser: Same comment as below.

Lib/multiprocessing/reduction.py
pierreglaser: Regarding (b), this is stated in the documentation that `get_pickler_class` should return a subclass of an `AbstractPickler`, that itself (as a subclass of `ForkingPickler`), implements a `register` method. Regarding (a), same comment as above (this @pablogsal's work), but I'll think about whether this function can be removed.

pierreglaser:
@pablogsal friendly ping: what are your thoughts on this?

Misc/NEWS.d/next/Library/2018-12-09-19-20-12.bpo-28053.JiOeSY.rst
```diff
@@ -0,0 +1 @@
+Fix the implementation of custom reducers for in `multiprocessing`.
```
terryjreedy, line 1: 'for in' Either a word is missing or there is an extra word.

terryjreedy:
> (sorry, failed rebasing attempt)

Rebasing is, as evidenced by repeated failures like this, error prone.  Besides the obvious noise, removing the spurious review requests does not unsubscribe people.

If you were just trying to update your branch to your current repository/fork master (once synchronized), rebasing is also not needed.  After checking out the branch, `git pull origin/master` will, if there are no conflicts, pull, merge, and commit updates to the branch with a standard commit message.

Misc/NEWS.d/next/Library/2018-12-09-19-20-12.bpo-28053.JiOeSY.rst
aeros: To me, it looks like "for" is the extra word. I'd personally go with:

```suggestion
Fix the implementation of custom reducers in :mod:`multiprocessing`.
```

(Also adds the sphinx role `:mod:` to provide an inline link to the module. It's optional for `Misc/NEWS.d`, but we may as well include it with the typo fix. This makes it easier to move over to the "What's New" doc later.)

MaxwellDupre (approved):
Run multiprocessing tests:
Ran 369 tests in 215.720s
OK (skipped=35)
Ran 39 tests in 22.709s
OK
Ran 369 tests in 85.400s
OK (skipped=39)
Ran 369 tests in 99.032s
OK (skipped=32)
Look ok to me.

Doc/library/multiprocessing.rst
```diff
@@ -1193,6 +1193,103 @@ For example:
...
+   Sets a reduction instance to be used for serialization and deserialization
+   by the module primitive internals. **reduction** must be an instance of a
+   subclass of :class:`AbstractReducer`.
+
+.. method:: get_reducer()
+
+   Gets the current reduction class in use by the module's primitive internals.
+
+For example, substituting the reducer class to use the :mod:`pickle` protocol
+version 2 to be able to communicate with a Python 2.x programs.::
```
gpshead, line 1271: Please change documentation and examples not to reference Python 2 as that long EOL at this point. Just say it in the abstract "communicate with processes using older Python versions" and offer some explanation as to how this is even possible by linking to the relevant APIs in multiprocessing that enable people to create such unexpected monsters. (see https://github.com/python/cpython/issues/67592#issuecomment-1093674359)

**Most** `multiprocessing` users view it as a module with a single parent process controlling identical by definition Python version child processes that it launches and controls.

Doc/library/multiprocessing.rst
```diff
@@ -1193,6 +1193,103 @@ For example:
...
+:meth:`get_pickler_class` (resp.  :meth:`get_unpickler_class`) is not
+implemented, multiprocessing will fallback to the :class:`pickle.Pickler`
+(resp. :class:`pickle.Unpickler`) class to serialize objects.
+
+.. currentmodule:: multiprocessing
+
+.. method:: set_reducer(reduction)
+
+   Sets a reduction instance to be used for serialization and deserialization
+   by the module primitive internals. **reduction** must be an instance of a
```
ezio-melotti, line 1263:
```suggestion
   by the module primitive internals. *reduction* must be an instance of a
```
Arguments should be marked in italic: https://devguide.python.org/documenting/#rest-inline-markup

Doc/library/multiprocessing.rst
```diff
@@ -1193,6 +1193,103 @@ For example:
...
+.. currentmodule:: multiprocessing.reduction
+
+.. class:: AbstractPickler(file, protocol)
+
+   Base class that can be implemented in order to override
+   serialization methods of the reduction machinery used by multiprocessing.
+
+   .. method:: dump(obj)
+
+      Write a pickled representation of obj to the open file.
```
ezio-melotti, line 1221:
```suggestion
      Write a pickled representation of *obj* to the open file.
```

Doc/library/multiprocessing.rst
```diff
@@ -1193,6 +1193,103 @@ For example:
...
+      reconstituted object hierarchy specified therein.
+
+      Defaults to :meth:`pickle.Unpickler.load`
+
+.. class:: AbstractReducer()
+
+   Base class providing access to custom ``Pickler`` and ``Unpickler``
+   classes to be used by ``multiprocessing`` when serializing objects.
+
+   .. method:: get_pickler_class():
```
ezio-melotti, line 1242:
```suggestion
   .. method:: get_pickler_class()
```

Doc/library/multiprocessing.rst
```diff
@@ -1193,6 +1193,103 @@ For example:
...
+
+      This method must return an subclass of :class:`AbstractPickler` to be used
+      by ``multiprocessing`` when serializing objects.
+
+   .. method:: get_unpickler_class():
```
ezio-melotti, line 1247:
```suggestion
   .. method:: get_unpickler_class()
```

Misc/NEWS.d/next/Library/2018-12-09-19-20-12.bpo-28053.JiOeSY.rst
ezio-melotti: Also note that `` `multiprocessing` `` will cause `make check` to fail: it should either be `` :mod:`multiprocessing` `` (as suggested by @aeros)  or ` ``multiprocessing`` `.

Doc/library/multiprocessing.rst
```diff
@@ -1193,6 +1193,103 @@ For example:
...
     impossible to be sure where the message boundaries lie.
 
+Custom Reduction
+~~~~~~~~~~~~~~~~
+
+.. currentmodule:: multiprocessing
+
+Several primitives of the :mod:`multiprocessing` module such as
+:class:`multiprocessing.Queue`, :class:`multiprocessing.connection.Listener` or
+:class:`multiprocessing.connection.Server` need to serialize and deserialize Python
```
ezio-melotti, line 1203: Aren't either the `currentmodule` directive or the `multiprocessing.*`s in the roles redundant in this part?

PR #15068, opened 2019-08-01
bpo-37737: mmap module, track anonymous paging case on macOS

devnexen:
https://bugs.python.org/issue37737

### DIFF
--- a/Misc/NEWS.d/next/macOS/2019-08-01-13-52-52.bpo-37737.1HE5BK.rst
+++ b/Misc/NEWS.d/next/macOS/2019-08-01-13-52-52.bpo-37737.1HE5BK.rst
@@ -1,0 +1,1 @@
+on macOS, anonymous paging cases can be tracked via tooling as vmmap.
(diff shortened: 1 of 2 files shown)


devnexen:
Allows output as :
image

ronaldoussoren:
What is the use case for this? 

And why one memory tag with a specific value?

devnexen:
If it is you wish the python mmap module mapping standing out to monitor it more accurately.
Otherwise tag ID is allowed up to 255, guaranteed from 240 to be freely available at least.

ronaldoussoren:
I still don't understand what the use case is, why is it useful to track usage of mmap.mmap in this way? 

I'd expect that tracking Python's memory allocator would be more useful to track memory usage patterns and debug production issues.

devnexen:
It is a more native mean ... when it occurs you monitor the whole system and allows to distinguish python from the rest.

ronaldoussoren:
I still don't think this is a useful feature. I'm therefore closing this PR (and the related issue).

PR #15075, opened 2019-08-02
bpo-18049: Reintroduce platform guard in recursion limit test

ronaldoussoren:
A previous PR fixed and issue with recursion in threads on
macOS and completely removed the platform guard. Removing this
guard broke tests on AIX, therefore reintroducing a guard that
only runs this test on macOS.

### DIFF
--- a/Lib/test/test_threading.py
+++ b/Lib/test/test_threading.py
@@ -1057,6 +1057,7 @@ class ThreadingExceptionTests(BaseTestCase):
         lock = threading.Lock()
         self.assertRaises(RuntimeError, lock.release)
 
+    @unittest.skipUnless(sys.platform == 'darwin', 'test macos problem')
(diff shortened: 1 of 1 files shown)


ronaldoussoren:
This should fix a problem on the AIX buildbots introduced by PR 14748.

ronaldoussoren:
@vstinner: this PR skips the failing test on all platforms except macOS, as it did before PR 14748.  An alternative is to skip the test only on AIX, which is the platform where this test fails.

vstinner:
> skip the test only on AIX, which is the platform where this test fails.

That's a good idea.

vstinner:
Oh. I see that the test is now fixed on AIX. That's even better :-) https://bugs.python.org/issue18049#msg348945

PR #15077, opened 2019-08-02
bpo-37742: Return the root logger when logging.getLogger('root') is c…

vsajip:
…alled.

https://bugs.python.org/issue37742


maggyero:
Nice.

PR #15081, opened 2019-08-02
bpo-18049: Define THREAD_STACK_SIZE for AIX to pass default recursion limit test

aixtools:


aixtools:
Still needs some adjustment - test_queue now fails with default memory size...

ronaldoussoren (approved):
The patch looks good to me.

Thank you for the patch.

eamanu (approved):

ned-deily:
(Original 3.8 backport (GH-15089) failed due to remote server problem; retrying backport in GH-15572.)

PR #15083, opened 2019-08-02
bpo-36311: Fixes decoding multibyte characters around chunk boundaries and improves decoding performance

zooba:
https://bugs.python.org/issue36311

### DIFF
--- a/Objects/unicodeobject.c
+++ b/Objects/unicodeobject.c
@@ -7422,8 +7422,11 @@ decode_code_page_stateful(int code_page,
     do
     {
 #ifdef NEED_RETRY
-        if (size > INT_MAX) {
-            chunk_size = INT_MAX;
+        /* INT_MAX is the theoretical largest chunk, but INT_MAX / 4
+           performs better and avoids partial characters overrunning the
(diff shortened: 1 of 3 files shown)


zooba:
I also fixed the nearly-relevant bigmem test and added one that covers this case. Both pass - the new one would fail with WindowsError 87 without the unicodeobject fix.

Objects/unicodeobject.c
```diff
@@ -7422,8 +7422,11 @@ decode_code_page_stateful(int code_page,
     do
     {
 #ifdef NEED_RETRY
-        if (size > INT_MAX) {
-            chunk_size = INT_MAX;
+        /* INT_MAX is the theoretical largest chunk, but INT_MAX / 4
+           performs better and avoids partial characters overrunning the
+           length limit in MultiByteToWideChar on Windows */
+        if (size > INT_MAX / 4) {
+            chunk_size = INT_MAX / 4;
```
ZackerySpytz: It seems that `INT_MAX / 4` is done in quite a few places. IMHO, it would be nice to use a constant.

PR #15092, opened 2019-08-03
bpo-37751: Fix normalizestring() with hyphens and spaces converted to underscores

fusioncid:
* Fix normalizestring() with hyphens and spaces converted to underscores
* Reuse _Py_normalize_encoding() in normalizestring()

https://bugs.python.org/issue37751


the-knights-who-say-ni:

shihai1991 (approved):
All the check have been passed, so LGTM.

vstinner:
Would it be possible to reuse _Py_normalize_encoding() in codecs.c normalizestring()?

qigangxu:
> Would it be possible to reuse _Py_normalize_encoding() in codecs.c normalizestring()?

I went through two functions and found that they do have similar functions. At the same time, I tried to modify the code and the test cases passed.

The code can be modified just like 'check_force_ascii‘  calls ‘_Py_normalize_encoding‘ . And I think it's better to call _Py_normalize_encoding as an external function to other modules.

There is also a process problem. I am not quite sure whether to submit another issue to discuss this modification, or can I directly modify it in this issue?

qigangxu:
> Would it be possible to reuse _Py_normalize_encoding() in codecs.c normalizestring()?

I will do more test for  reusing  _Py_normalize_encoding() in codecs.c normalizestring().  Thank you for your helpful suggestions。

vstinner:
You can create a new PR, this one can be closed when the new one is merged.

qigangxu:
> You can create a new PR, this one can be closed when the new one is merged.

Let me try it.

Python/codecs.c
```diff
@@ -49,36 +49,36 @@ int PyCodec_Register(PyObject *search_function)
...
         return PyErr_NoMemory();
-    for (i = 0; i < len; i++) {
-        char ch = string[i];
-        if (ch == ' ')
-            ch = '-';
-        else
-            ch = Py_TOLOWER(Py_CHARMASK(ch));
-        p[i] = ch;
+
+    if (!_Py_normalize_encoding(string, encoding, len + 1))
```
vstinner, line 73: I suggest to either ignore the return value, or use an assertion. The function cannot fail with len+1. It only fails if the second string argument is too short, which cannot happen here.

Python/codecs.c
```diff
@@ -49,36 +49,36 @@ int PyCodec_Register(PyObject *search_function)
     return -1;
 }
 
+extern int _Py_normalize_encoding(const char *, char *, size_t);
+
 /* Convert a string to a normalized Python string: all characters are
-   converted to lower case, spaces are replaced with underscores. */
+   converted to lower case, spaces and hyphens are replaced with underscores. */
```
vstinner, line 55: Can you please specify that the string is decoded from UTF-8?

Python/codecs.c
```diff
@@ -49,36 +49,36 @@ int PyCodec_Register(PyObject *search_function)
...
-        char ch = string[i];
-        if (ch == ' ')
-            ch = '-';
-        else
-            ch = Py_TOLOWER(Py_CHARMASK(ch));
-        p[i] = ch;
+
+    if (!_Py_normalize_encoding(string, encoding, len + 1))
+    {
+        Py_FatalError("normalizestring(): encoding failed");
```
vstinner: Or use a regular exception:

```suggestion
        PyErr_SetString(PyExc_RuntimeError, "_Py_normalize_encoding() failed");
```

vstinner (changes requested):

vstinner:
Can you please try to add a NEWS entry using the blurb tool? (install it using: python3 -m pip install --user blurb)

Python/codecs.c
fusioncid: Using a regular exception here is better.

Python/codecs.c
fusioncid: done

qigangxu:
> Can you please try to add a NEWS entry using the blurb tool? (install it using: python3 -m pip install --user blurb)

Thank you for your careful guidance. I need to take a moment to familiarize myself with the blurb tool.

qigangxu:

Misc/NEWS.d/next/Core and Builtins/2019-08-20-04-36-37.bpo-37751.CSFzUd.rst
```diff
@@ -0,0 +1 @@
+Fix ``normalizestring`` with hyphens and spaces converted to underscores and reuse ``_Py_normalize_encoding()`` in ``normalizestring``.
```
vstinner: normalizestring() is as private function, not visible to users. I propose to mention codecs.lookup() instead. I propose:

```
Fix :func:`codecs.lookup` to normalize the encoding name
the same way than :func:`encodings.normalize_encoding`,
except that :func:`codecs.lookup` also converts the name
to lower case.
```

My text is based on the _Py_normalize_encoding() comment:

```
/* Normalize an encoding name: similar to encodings.normalize_encoding(), but
   also convert to lowercase. Return 1 on success, or 0 on error (encoding is
   longer than lower_len-1). */
```

I compared _Py_normalize_encoding() and encodings.normalize_encoding(): the only difference is that _Py_normalize_encoding() converts also to lower case.

Subtle detail: _Py_normalize_encoding() only converts A-Z letters to a-z: it's different than str.lower() which implements the Unicode standard and considers the whole Unicode Character Set, not only ASCII. But I don't think that it's worth it to mention it here.

fusioncid: Thank you for your patience and help. I have revised this issue.

vstinner (approved):
LGTM. Thanks for the update.

PR #15094, opened 2019-08-03
bpo-1047397: prevent repr/getattr exceptions in cgitb

RMJ10:
Robin Becker's patch, slightly tweaked (for rotting) and tidied.  Basically it wraps calls to getattr and repr while inspecting the code for which the traceback is being generated to prevent them from causing more exceptions.

https://bugs.python.org/issue1047397

### DIFF
--- a/Lib/cgitb.py
+++ b/Lib/cgitb.py
@@ -43,6 +43,8 @@ Content-Type: text/html
 </table> </table> </table> </table> </table> </font> </font> </font>'''
 
 __UNDEF__ = []                          # a special sentinel object
+__GETATTR_FAILED__ = '<em>&lt;ungettable attribute&gt;</em>'
+__HTML_REPR_FAILED__ = '<em>&lt;unreprable value&gt;</em>'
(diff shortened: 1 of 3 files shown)


Lib/cgitb.py
```diff
@@ -74,9 +76,32 @@ def lookup(name, frame, locals):
...
+                or args[0] is __GETATTR_FAILED__
+                ):
+            return args[0]
+        try:
+            return self.func(*args)
+        except:
+            return self.failval
+
+def _pydoc_repr(value, sub='html'):
+    return getattr(pydoc, 'html').repr(value)
```
ethanfurman: Should `'html'` be `sub` instead?

ethanfurman (commented):
We need a news entry (see the `blurb` tool).

Also, some tests would be good .

Lib/cgitb.py
RMJ10: Plausibly.  I'll test it and see.

PR #15103, opened 2019-08-04
bpo-37706: fix sidebar drag tests

taleinat:
The test failures were caused by the sidebar code getting screen coordinates from the main editor window's text widget rather than the sidebar's text widget. Empirically, these are not in sync quite as expected when running tests.

This PR also fixes an actual bug (i.e. not only tests): Drag-selecting a range of lines by click-dragging on the line numbers, and then clicking a line number outside of the selected range, expands the selection range rather than selecting only the corresponding line. This is very problematic to test automatically, so the htest has been expanded to cover this too.

https://bugs.python.org/issue37706

### DIFF
--- a/Lib/idlelib/idle_test/htest.py
+++ b/Lib/idlelib/idle_test/htest.py
@@ -210,13 +210,20 @@ _linenumbers_drag_scrolling_spec = {
     'file': 'sidebar',
     'kwds': {},
     'msg': textwrap.dedent("""\
-        Click on the line numbers and drag down below the edge of the
+        1. Click on the line numbers and drag down below the edge of the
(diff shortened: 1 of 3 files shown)


taleinat:
I think this doesn't need a NEWS entry since it hasn't yet been included in any (non-beta) release.

terryjreedy (approved):
Given the clear description of how to reproduce the bug, I verified its presence in .0b3 on Win10 both with htest and live editor, and it absence after the patch.

I noticed editwin.getlineno before, but did not think to question it either.

Factoring out simulated drag into a helper function is a good idea.  If it is ever needed for another test, it should go into idle_test/mock_tk.py.  More so for simulated mouse/key click, which already had duplicated code in multiple files.

PR #15109, opened 2019-08-04
bpo-37449: Move ensurepip off of pkgutil and to importlib.resources

nanjekyejoannah:
Move ensurepip off of pkgutil and to importlib.resources.

https://bugs.python.org/issue37449

Automerge-Triggered-By: @ericsnowcurrently

### DIFF
--- a/Lib/ensurepip/__init__.py
+++ b/Lib/ensurepip/__init__.py
@@ -1,8 +1,10 @@
 import os
 import os.path
-import pkgutil
 import sys
 import tempfile
+from . import _bundled
+
+from importlib.resources import read_binary
 
 
 __all__ = ["version", "bootstrap"]
@@ -96,9 +98,9 @@ def _bootstrap(*, root=None, upgrade=False, user=False,
         additional_paths = []
         for project, version in _PROJECTS:
             wheel_name = "{}-{}-py2.py3-none-any.whl".format(project, version)
-            whl = pkgutil.get_data(
-                "ensurepip",
-                "_bundled/{}".format(wheel_name),
+            whl = read_binary(
+                _bundled,
+                wheel_name,
             )
             with open(os.path.join(tmpdir, wheel_name), "wb") as fp:
                 fp.write(whl)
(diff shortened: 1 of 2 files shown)


nanjekyejoannah:
cc @brettcannon @warsaw .

Lib/ensurepip/__init__.py
```diff
@@ -1,8 +1,10 @@
 import os
 import os.path
-import pkgutil
 import sys
 import tempfile
+from . import _bundled
+
+from importlib.resources import read_binary
```
brettcannon: Import to the module, not the object (both to follow the style of the module and because it avoids circular import issues):
```suggestion
from importlib import resources
```

Lib/ensurepip/__init__.py
```diff
@@ -1,8 +1,10 @@
 import os
 import os.path
-import pkgutil
 import sys
 import tempfile
+from . import _bundled
```
brettcannon, line 7: [PEP 8 says](https://www.python.org/dev/peps/pep-0008/#imports) to sort imports so that stdlib imports com first, then a blank line, then package-local modules.

brettcannon (changes requested):
I don't see a change in here to add an `__init__.py` file to `Lib/ensurepip/_bundled` to make it be a concrete package instead of a namespace package as `importlib.resources` won't work with the latter. Did you happen to manually test this and have it succeed? Or am I missing something?

Lib/ensurepip/__init__.py
nanjekyejoannah: IIRC, I was not able to do the import without adding the __init__.py to make it  a package.

nanjekyejoannah:

Lib/ensurepip/__init__.py
```diff
@@ -1,9 +1,12 @@
 import os
 import os.path
-import pkgutil
 import sys
 import tempfile
 
+from importlib import resources
```
brettcannon, line 5: This should get moved up with the other stdlib modules above.

brettcannon (changes requested):
One very minor PEP 8 thing, then this looks ready to go!

nanjekyejoannah:

ericsnowcurrently (approved):
LGTM

PR #15125, opened 2019-08-05
bpo-37758: Extend unicodedata checksum tests to cover all of Unicode.

gnprice:
Unicode has grown since Python first gained support for it,
when Unicode itself was still rather new.

This pair of test cases was added in commit 6a20ee7de back in 2000,
and they haven't needed to change much since then.  But do change
them to look beyond the Basic Multilingual Plane (range(0x10000))
and cover all 17 planes of Unicode's final form.

This adds about 5 seconds to the test suite's runtime.  Mark the
tests as CPU-using accordingly.

https://bugs.python.org/issue37758

### DIFF
--- a/Lib/test/test_unicodedata.py
+++ b/Lib/test/test_unicodedata.py
@@ -11,17 +11,18 @@ from http.client import HTTPException
 import sys
 import unicodedata
 import unittest
-from test.support import open_urlresource, script_helper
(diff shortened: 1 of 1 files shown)


benjaminp:
This seems okay, though it now has conflicts with master.

gnprice:
Thanks for the review -- rebased!

PR #15127, opened 2019-08-05
bpo-37759:  Significant edits to Whatsnew 3.8

rhettinger:
Work in progress.  Please add suggestions here as you find them.

https://bugs.python.org/issue37759

### DIFF
--- a/Doc/whatsnew/3.8.rst
+++ b/Doc/whatsnew/3.8.rst
@@ -454,7 +454,7 @@ Other Language Changes
   inverse for other negative exponents.  For example, to compute the
   `modular multiplicative inverse
   <https://en.wikipedia.org/wiki/Modular_multiplicative_inverse>`_ of 38
-  in modulo 137, write::
+  modulo 137, write::
 
     >>> pow(38, -1, 137)
     119
--- a/Doc/whatsnew/3.8.rst
+++ b/Doc/whatsnew/3.8.rst
@@ -115,16 +115,7 @@ the expression body::
     if (clean_name := normalize('NFC', name)) in allowed_names]
 
 Try to limit use of the walrus operator to clean cases that reduce
-complexity and improve readability.  When combined with other features
-like f-strings or conditional expressions, it is easy to go over-board
-with code that is challenging to decipher::
-
-  # Cryptic assignment expression embedded in a conditional expression
-  binomial_coefficient = [(c := c * (n - k + 1) // k if k else 1) for k in range(n+1)]
-
-  # Visually confusing assignment expression embedded in an f-string
-  for angle in range(360):
-    print(f'{angle=}\N{degree sign} {(theta:=radians(angle))=:.3f} {sin(theta)=:.3f}')
+complexity and improve readability.
 
 See :pep:`572` for a full description.
 
--- a/Doc/whatsnew/3.8.rst
+++ b/Doc/whatsnew/3.8.rst
@@ -144,7 +144,7 @@ In the following example, parameters *a* and *b* are positional-only,
 while *c* or *d* can be positional or keyword, and *e* or *f* are
 required to be keywords::
 
-  def f(a, b, \, c, d, *, e, f):
+  def f(a, b, /, c, d, *, e, f):
       print(a, b, c, d, e, f)
 
 The following is a valid call::
--- a/Doc/whatsnew/3.8.rst
+++ b/Doc/whatsnew/3.8.rst
@@ -55,6 +55,11 @@ notable items not yet covered are:
 * :pep:`578` - Runtime audit hooks for potentially sensitive operations
 * ``python -m asyncio`` runs a natively async REPL
 
+.. testsetup::
+
+   from datetime import date
+   from math import cos, radians
+
 
 Summary -- Release highlights
 =============================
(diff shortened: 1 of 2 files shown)


Doc/whatsnew/3.8.rst
```diff
@@ -92,20 +129,69 @@ See :pep:`572` for a full description.
...
+function parameters must be specified positionally and cannot be used as
+keyword arguments.  This is the same notation shown by ``help()`` for C
+functions annotated with Larry Hastings' `Argument Clinic
+<https://docs.python.org/3/howto/clinic.html>`_ tool.
+
+In the following example, parameters *a* and *b* are positional-only,
+while *c* or *d* can be positional or keyword, and *e* or *f* are
+required to be keywords::
+
+  def f(a, b, \, c, d, *, e, f):
```
willingc:
```suggestion
  def f(a, b, /, c, d, *, e, f):
```

willingc: @rhettinger Looks like there is one slash going the wrong direction :wink:

rhettinger: Thanks.  Should be fixed now.

Doc/whatsnew/3.8.rst
```diff
@@ -76,12 +79,52 @@ New Features
...
+
+Another motivating use case arises in list comprehensions where
+a value computed in a filtering condition is also needed in
+the expression body::
+
+   [clean_name.title() for name in names
+    if (clean_name := normalize('NFC', name)) in allowed_names]
+
+Try to limit use of the walrus operator to clean cases that reduce
+complexity and improve readability.  When combined with other features
```
gvanrossum: I would stop at "readability." I don't think specific examples of what's *not* okay help the readers of this document.

rhettinger: Will do.

At some point, we should put in some PEP 8 guidance.  I've been using the beta with customers and students for the last several weeks.  These examples were typical.

ericvsmith (approved):
This looks great to me. Thanks for your work on this, Raymond!

Doc/whatsnew/3.8.rst
```diff
@@ -343,10 +447,30 @@ Other Language Changes
...
-  the modulus. It then computes a modular inverse to the base when the exponent
-  is ``-1``, and a suitable power of that inverse for other negative exponents.
+* For integers, the three-argument form of the :func:`pow` function now
+  permits the exponent to be negative in the case where the base is
+  relatively prime to the modulus. It then computes a modular inverse to
+  the base when the exponent is ``-1``, and a suitable power of that
+  inverse for other negative exponents.  For example, to compute the
+  `modular multiplicative inverse
+  <https://en.wikipedia.org/wiki/Modular_multiplicative_inverse>`_ of 38
+  in modulo 137, write::
```
jdemeyer: "in modulo 137" should be "modulo 37".

rhettinger: Dropped the "in".  Kept the 137.

Doc/whatsnew/3.8.rst
```diff
@@ -92,20 +127,69 @@ See :pep:`572` for a full description.
...
+
+Another use case is to preclude keyword arguments when the parameter
+name is not helpful.  For example, the builtin :func:`len` function has
+the signature ``len(obj, /)``.  This precludes awkward calls such as::
 
-Now ``pow(2, 10)`` and ``pow(2, 10, 17)`` are valid calls, but
-``pow(x=2, y=10)`` and ``pow(2, 10, z=17)`` are invalid.
+  len(obj='hello')  # The "obj" keyword argument impairs readability
+
+A further benefit of marking a parameter as positional-only is that it
```
pablogsal, line 167: I find this section very well structured and informative! This rewrite really makes a difference :) What do you think about adding to this paragraph also the case of going from one argument to `*args` without breaking client code. For example going from

```python
def add_to_queue(item):
   ...
```
to
```python
def add_to_queue(*items):
   ...
```

tirkarthi:
With respect to unittest module. It now has the ability to collect async functions where before 3.8 tests have to be a normal function that could have an async definition inside and uses `asyncio.run` or `loop` to run it. This was not mentioned in whatsnew. Maybe a good addition given it improves testing asyncio code. Relevant commit : 4dd3e3f9bbd320f0dd556688e04db0a6b55a7b52

```python
import asyncio
import unittest

from unittest.async_case import IsolatedAsyncioTestCase

class FooTest(IsolatedAsyncioTestCase):

    async def asyncSetUp(self):
        print("Async Setup")

    async def test_sleep(self):
        await asyncio.sleep(1)
        self.assertEqual(1, 1)

    async def asyncTearDown(self):
        print("Async Teardown")

if __name__ == "__main__":
    unittest.main()
```

```
./python.exe /tmp/foo.py
Async Setup
Async Teardown
.
Ran 1 test in 1.028s

OK
```

Thanks for improving the changelog examples.

willingc (approved):

rhettinger:
tirkarthi , thanks for the suggestions.  I'll add the unittest example to the next round of updates (I'm making multiple PRs to minimize merge conflicts).

PR #15128, opened 2019-08-05
bpo-37760: Avoid cluttering work tree with downloaded Unicode files.

gnprice:

### DIFF
--- a/Tools/unicode/makeunicodedata.py
+++ b/Tools/unicode/makeunicodedata.py
@@ -886,15 +886,19 @@ def merge_old_version(version, new, old):
                         normalization_changes))
 
 
+DATA_DIR = os.path.join('Tools', 'unicode', 'data')
+
 def open_data(template, version):
-    local = template % ('-'+version,)
+    local = os.path.join(DATA_DIR, template % ('-'+version,))
     if not os.path.exists(local):
         import urllib.request
         if version == '3.2.0':
             # irregular url structure
-            url = 'http://www.unicode.org/Public/3.2-Update/' + local
+            url = ('http://www.unicode.org/Public/3.2-Update/'+template) % ('-'+version,)
         else:
             url = ('http://www.unicode.org/Public/%s/ucd/'+template) % (version, '')
+        if not os.path.exists(DATA_DIR):
+            os.mkdir(DATA_DIR)
         urllib.request.urlretrieve(url, filename=local)
     if local.endswith('.txt'):
         return open(local, encoding='utf-8')
--- a/Tools/unicode/makeunicodedata.py
+++ b/Tools/unicode/makeunicodedata.py
@@ -897,8 +897,7 @@ def open_data(template, version):
             url = ('http://www.unicode.org/Public/3.2-Update/'+template) % ('-'+version,)
         else:
             url = ('http://www.unicode.org/Public/%s/ucd/'+template) % (version, '')
-        if not os.path.exists(DATA_DIR):
-            os.mkdir(DATA_DIR)
+        os.makedirs(DATA_DIR, exist_ok=True)
         urllib.request.urlretrieve(url, filename=local)
     if local.endswith('.txt'):
         return open(local, encoding='utf-8')
--- a/.gitignore
+++ b/.gitignore
@@ -74,6 +74,7 @@ PCbuild/arm32/
 PCbuild/arm64/
 PCbuild/obj/
 PCbuild/win32/
+Tools/unicode/data/
 .purify
 __pycache__
 autom4te.cache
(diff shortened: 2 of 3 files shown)


gnprice:
@benjaminp , might I persuade you to take a look at this patch? The Git history suggests you might run this script as much as anybody, so you may have a direct interest in making that more convenient :slightly_smiling_face: 

Then #15171 makes the resulting diffs a bit nicer, too.

Tools/unicode/makeunicodedata.py
```diff
@@ -886,15 +886,19 @@ class Difference(Exception):pass
...
     if not os.path.exists(local):
         import urllib.request
         if version == '3.2.0':
             # irregular url structure
-            url = 'http://www.unicode.org/Public/3.2-Update/' + local
+            url = ('http://www.unicode.org/Public/3.2-Update/'+template) % ('-'+version,)
         else:
             url = ('http://www.unicode.org/Public/%s/ucd/'+template) % (version, '')
+        if not os.path.exists(DATA_DIR):
+            os.mkdir(DATA_DIR)
```
benjaminp: `os.makedirs` has a nice `exist_ok` parameter now.

benjaminp (commented):
I actually don't mind the `git status` clutter because then I remember to clean the data files out when I'm done. If there are partial downloads, `git clean -f` will put you back in a working state.

This script does trigger my distaste for build tools that write to the source tree. Of course, doing that is its _raison d'être_, so I can hardly ask that it stop doing that completely.

Tools/unicode/makeunicodedata.py
gnprice: Oh neat! I have written little helpers with that behavior more times than I would like. :-) Update pushed.

gnprice:
Thanks for the review (and the merges!)

> I actually don't mind the `git status` clutter because then I remember to clean the data files out when I'm done. If there are partial downloads, `git clean -f` will put you back in a working state.

Ah, fair enough. I guess that means the main situation where the clutter is annoying might be when doing development work on this script... as I've been doing recently. :wink: That means rerunning the script a lot, and it's much faster when it doesn't have to download the files fresh. So I want to keep them around, and I'd like that not to become clutter.

To keep good support for your use case too, where you want the data to disappear between the occasional times you run the script, perhaps I'll add it to `make clean`? That way it gets handled the same as the many other random build products.


> This script does trigger my distaste for build tools that write to the source tree. Of course, doing that is its _raison d'être_, so I can hardly ask that it stop doing that completely.

Yeah, I'd love for the whole build to go into some tidy set-aside place. I'd be glad to adjust this to follow such a thing, or to go to an existing somewhat-better place if you can suggest one.

benjaminp:
> Thanks for the review (and the merges!)
> 
> > I actually don't mind the `git status` clutter because then I remember to clean the data files out when I'm done. If there are partial downloads, `git clean -f` will put you back in a working state.
> 
> Ah, fair enough. I guess that means the main situation where the clutter is annoying might be when doing development work on this script... as I've been doing recently. wink That means rerunning the script a lot, and it's much faster when it doesn't have to download the files fresh. So I want to keep them around, and I'd like that not to become clutter.
> 
> To keep good support for your use case too, where you want the data to disappear between the occasional times you run the script, perhaps I'll add it to `make clean`? That way it gets handled the same as the many other random build products.

The problem with the `make clean` change is that it breaks when doing out-of-tree builds.

I don't feel strongly about this, so I could merge the original PR without the `Makefile` change.

> 
> > This script does trigger my distaste for build tools that write to the source tree. Of course, doing that is its _raison d'être_, so I can hardly ask that it stop doing that completely.
> 
> Yeah, I'd love for the whole build to go into some tidy set-aside place. I'd be glad to adjust this to follow such a thing, or to go to an existing somewhat-better place if you can suggest one.

The only other thing I can think of is adding a script flag to control the cache location.

gnprice:
> The problem with the `make clean` change is that it breaks when doing out-of-tree builds.

Ah, out-of-tree builds -- that does sound nice :slightly_smiling_face: and I've been meaning to go try those.

> I don't feel strongly about this, so I could merge the original PR without the `Makefile` change.

Cool -- I pushed a revert.


> The only other thing I can think of is adding a script flag to control the cache location.

Sure, that'd be reasonable. Alternatively, perhaps it could notice if you've set up an out-of-tree build (that's done at `configure` time, right?) and place its output automatically in an appropriate spot in the build tree? ... Or maybe the right way to implement that is a script flag like you described, and then a `make` rule that passes that flag.

I'll probably have a clearer sense of the right design there later, after I've played with out-of-tree builds a bit.

benjaminp:
> > The only other thing I can think of is adding a script flag to control the cache location.
> 
> Sure, that'd be reasonable. Alternatively, perhaps it could notice if you've set up an out-of-tree build (that's done at `configure` time, right?) and place its output automatically in an appropriate spot in the build tree? ... Or maybe the right way to implement that is a script flag like you described, and then a `make` rule that passes that flag.

Yeah, I think everything else that knows the builddir finds it out by parsing the Makefile (!) or being invoked by it.

gnprice:
> by parsing the Makefile (!)

Yikes! I can imagine situations where that's locally the least-bad option -- but yeah, here I'd opt for having the Makefile invoke it.

gnprice:
(Thanks for the review and the merge!)

PR #15131, opened 2019-08-05
bpo-37757: Disallow PEP 572 cases that expose implementation details

ncoghlan:
- drop TargetScopeError in favour of raising SyntaxError directly
  as per the updated PEP 572
- comprehension iteration variables are explicitly local, but
  named expression targets in comprehensions are nonlocal or
  global. Raise SyntaxError as specified in PEP 572
- named expression targets in the outermost iterable of a
  comprehension have an ambiguous target scope. Avoid resolving
  that question now by raising SyntaxError. PEP 572
  originally required this only for cases where the bound name
  conflicts with the iteration variable in the comprehension,
  but CPython can't easily restrict the exception to that case
  (as it doesn't know the target variable names when visiting
  the outermost iterator expression)

https://bugs.python.org/issue37757

### DIFF
--- a/Lib/test/test_named_expressions.py
+++ b/Lib/test/test_named_expressions.py
@@ -104,7 +104,7 @@ class NamedExpressionInvalidTest(unittest.TestCase):
         with self.assertRaisesRegex(SyntaxError, "invalid syntax"):
             exec(code, {}, {})
 
-    def test_named_expression_invalid_18(self):
+    def test_named_expression_invalid_in_class_body(self):
         code = """class Foo():
             [(42, 1 + ((( j := i )))) for i in range(5)]
         """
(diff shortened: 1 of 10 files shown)


ncoghlan:
There's a test design issue that comes up with this PR: in CPython, all the different comprehension types share several common code paths, so these tests only test list comprehensions, and rely on the common code paths to extend that to the rules also being applied correctly for set, dict, and generator comprehensions.

From a *language* testing perspective though, perhaps we should be testing all 4 cases, as other implementations may not have the same level of code sharing, and hence have bugs specific to one of the other variants that don't exist for list comprehensions?

ncoghlan:
@emilyemorehouse The PEP update to remove TargetScopeError and clarify the rationale for treating these cases as syntax errors has been accepted and merged, so this is ready for review now.

There are some additional cases that I need to make sure the test suite is covering: situations where we *reference* a global or nonlocal in a comprehension before we rebind it, like:

    [j := i for i in random.choices(range(10)) if j < i]

ncoghlan:
I don't believe edit notifications get sent via email, so adding an explicit note to say that I've added the extra test cases I mentioned in my last comment, and edited that comment accordingly.

So unless I've missed something, I believe this should be good to go now.

ncoghlan:
Note that I'm deliberately not bumping the bytecode magic number here - this means that you could technically compile affected code under 3.8b3 or earlier, and 3.8b4 and later will still run it. I think that's fine, and don't see a reason to compel a rebuild of all bytecode just to enforce new syntax errors.

ncoghlan:
My current plan is to merge this on Sunday August 25th, to make the 3.8b4 cut-off. If folks have comments before then, I can take them into account for the beta, otherwise, I'm still interested in post-merge reviews, but I'll handle addressing them as follow-up changes before rc1.

Lib/test/test_named_expressions.py
```diff
@@ -421,6 +442,34 @@ def spam():
...
+            ("Rebind global", f"x = 1; result = {rebinding}"),
+            ("Rebind nonlocal", f"result, x = (lambda x=1: ({rebinding}, x))()"),
+            ("Filter global", f"x = 1; result = {filter_ref}"),
+            ("Filter nonlocal", f"result, x = (lambda x=1: ({filter_ref}, x))()"),
+            ("Body global", f"x = 1; result = {body_ref}"),
+            ("Body nonlocal", f"result, x = (lambda x=1: ({body_ref}, x))()"),
+            ("Nested global", f"x = 1; result = {nested_ref}"),
+            ("Nested nonlocal", f"result, x = (lambda x=1: ({nested_ref}, x))()"),
+        ]
+        msg = "assignment expression cannot be used in a comprehension iterable expression"
```
emilyemorehouse: `msg` isn't used (probably a copy-paste relic since this test checks for successful use cases)

emilyemorehouse (approved):
Very small nit, otherwise LGTM. Thanks, Nick!

Lib/test/test_named_expressions.py
ncoghlan: Removed! (and yep, it was indeed a copy-paste relic)

ncoghlan:
@emilyemorehouse Thanks for the review!

ncoghlan:
#15491 is the deconflicted backport - removing the Windows DLL export symbol for TargetScopeError failed in the implicit backport, as the two branches have different version numbers in the symbols.

pablogsal:
This commit introduced a reference leak that is being tracked in https://bugs.python.org/issue37954

PR #15134, opened 2019-08-05
bpo-37587: optimize json.loads for large string

methane:
Use a tighter scope temporary variable to help register allocation.


pablogsal:
@methane Should we backport this to 3.8? The other PR was indeed backported

methane:
I'm not sure.  This is very small (1~3% in a specific case) or optimization.
Not so important.

PR #15136, opened 2019-08-05
bpo-37763: Teach setup.py to pick up "-isystem <dir>" from $CPPFLAGS

jherland:
When (cross-)compiling, setup.py will pick up header search dirs from
$CPPFLAGS and lib search dirs from $LDFLAGS. However, the code only
looks at the -I options in $CPPFLAGS to find header search dirs.

Using the -isystem option (instead of -I) to pass header search dirs is
also a valid option, but this is currently unsupported by setup.py.

This patch adds handling of -isystem options, so that these are picked
up the same way that -I options currently are.

https://bugs.python.org/issue37763


erlend-aasland:
Thanks for you PR. `setup.py` is no longer part of the code base.

PR #15138, opened 2019-08-05
bpo-37765: Add keywords to IDLE tab completions

terryjreedy:
Keywords are present in the main module tab completion lists generated by rlcompleter, which is used by REPLs on *nix.  Add all keywords to to IDLE's main module name list except those  already added from builtins (True, False, and None) .  This list may also be used by Show Completions on the Edit menu, and its hot key.

Rewrite Completions doc.

https://bugs.python.org/issue37765


tirkarthi (approved):
LGTM. Thanks @terryjreedy .

Lib/idlelib/autocomplete.py
```diff
@@ -175,6 +176,7 @@ def fetch_completions(self, what, mode):
                     namespace = {**__main__.__builtins__.__dict__,
                                  **__main__.__dict__}
                     bigl = eval("dir()", namespace)
+                    bigl.extend(keyword.kwlist)
```
csabella: This looks good, but `True` is on the list twice, so it looks like `bigl` needs to be dedupped.

csabella (changes requested):
Make sure list has no duplicates.

Lib/idlelib/autocomplete.py
terryjreedy: Ditto for None and False.  Thanks for catching this.  Revision excludes None and True by length and False by equality and includes a new test of each.

terryjreedy:
Cheryl, please review again, including the rewritten Completions section.

Doc/library/idle.rst
```diff
@@ -469,52 +469,47 @@ are restricted to four spaces due to Tcl/Tk limitations.
...
+being completed and the item highlighted in the box can be changed by
+adding and deleting characters, with Up, Down, Page Up, Page Down, Home
+and End keys, and by a single click within the box.  Keys <Escape>,
+<Enter>, and double <Tab> and clicks outside the box close the box.
+A double click within the box selects and closes.
+
+An attribute completion box will automatically open, if any completions
+are available, if one types '.' and waits for a predefined
+delay. The delay defaults to 2 seconds and can be set in the settings
+dialog.  If one types os.sep or os.altsep in a string and waits,
```
csabella:
```suggestion
dialog.  If one types an :data:`os.sep` or :data:`os.altsep` in a string and waits,
```

Doc/library/idle.rst
```diff
@@ -469,52 +469,47 @@ are restricted to four spaces due to Tcl/Tk limitations.
...
-an Editor window which are not via ``__main__`` and :data:`sys.modules` will
-not be found.  Run the module once with your imports to correct this situation.
-Note that IDLE itself places quite a few modules in sys.modules, so
-much can be found by default, e.g. the re module.
-
-If you don't like the ACW popping up unbidden, simply make the delay
-longer or disable the extension.
+When available, completions are supplied for filenames and for
+attributes of modules, classes, and functions.  This is usually done
+by displaying a completion box listing existing names.  The name
```
csabella: Lots of 'ing' words here, which may be more difficult for non-English speakers.  Perhaps:
```suggestion
by displaying a completion box that lists existing names.  The name
```

Doc/library/idle.rst
```diff
@@ -469,52 +469,47 @@ are restricted to four spaces due to Tcl/Tk limitations.
...
+being completed and the item highlighted in the box can be changed by
+adding and deleting characters, with Up, Down, Page Up, Page Down, Home
+and End keys, and by a single click within the box.  Keys <Escape>,
```
csabella: Here you mention keys without enclosing them in brackets, but in the next line you enclose the keys in brackets.  You also use the `:kbd:` directive below. Was there a reason for using different styles?

Doc/library/idle.rst
```diff
@@ -469,52 +469,47 @@ are restricted to four spaces due to Tcl/Tk limitations.
...
-Completions are currently limited to those in the namespaces. Names in
-an Editor window which are not via ``__main__`` and :data:`sys.modules` will
-not be found.  Run the module once with your imports to correct this situation.
-Note that IDLE itself places quite a few modules in sys.modules, so
-much can be found by default, e.g. the re module.
-
-If you don't like the ACW popping up unbidden, simply make the delay
-longer or disable the extension.
+When available, completions are supplied for filenames and for
+attributes of modules, classes, and functions.  This is usually done
```
csabella: Why 'usually'?

Doc/library/idle.rst
```diff
@@ -469,52 +469,47 @@ are restricted to four spaces due to Tcl/Tk limitations.
...
+by displaying a completion box listing existing names.  The name
+being completed and the item highlighted in the box can be changed by
+adding and deleting characters, with Up, Down, Page Up, Page Down, Home
+and End keys, and by a single click within the box.  Keys <Escape>,
+<Enter>, and double <Tab> and clicks outside the box close the box.
```
csabella: I think 'close the box' is the point of this sentence, but it gets lost.  It might be clearer to start with that.  Or, alternatively, to reformat this paragraph somehow as a navigation section with bullets or subsections.  Something like 

```
Open the completion list:
Get new completions:
Select a completion:
Close the box:
```

csabella:
The functionality works as designed.  I agree with Tal that having some missing keywords (like `pass`) is a little awkward.

Doc/library/idle.rst
terryjreedy: Because if there is only one matching option, tab completion replaces without a box, as explained later.
When revising, I added the exception.

Doc/library/idle.rst
terryjreedy: I was consistent within sentences ;-).  Habit and history.  The directive translates to `<kbd class="kbd docutils literal notranslate">Backspace</kbd>`.  In my Firefox, this displays a  slightly bold sans serif.  The contrast with normal text depends on whether normal text has serifs or not, and with styling turned on, it does not.  In help.py, the `kbd` tag is not translated, so no contrast.   We should revise that to 'em' (currently italics) or a new tk 'bd' tag.

Doc/library/idle.rst
terryjreedy: I revised to "Close the box with ...", parallel to previous sentence.  As to structure, we have a 3 dimensions: Open method (wait, hot key, tab), namespace (main, attribute, files), and stage (open select, close), with some cells blank (wait never opens the main namespace) and some entries identical.  The existing order made little sense to me.  I started with select and close because these are common to all open methods and lists.  I made this more explicit.  I follow with the 3 invocation methods and module level completions.

In any case, there are more edits all through the doc and a will wait a day for more comments before merging.

Doc/library/idle.rst
terryjreedy: I rewrote the sentence.

Lib/idlelib/autocomplete.py
```diff
@@ -171,10 +172,13 @@ def fetch_completions(self, what, mode):
...
         else:
             if mode == ATTRS:
-                if what == "":
+                if what == "":  # Main module names.
                     namespace = {**__main__.__builtins__.__dict__,
                                  **__main__.__dict__}
                     bigl = eval("dir()", namespace)
+                    kwds = (s for s in keyword.kwlist
+                            if s not in {'True', 'False', 'None'})
+                    bigl.extend(kwds)
```
csabella, line 181: As you said on [issue 27609](https://bugs.python.org/issue27609), this method needs to be refactored, so the looks good for adding the keywords and anything else can be handled on the refactor.

csabella (approved):

Lib/idlelib/autocomplete.py
terryjreedy: The refactoring of fetch_completions is issue 37766.  I attached a refactoring of the computation of bigl.  I did not use it here because the computation of smalll for imported modules containing __all__ uses 'namespace' and changing the smalll code is out of scope here.  (Otherwise, I would have.)  Keywords cannot complete module.?.

PR #15144, opened 2019-08-06
 bpo-37774: use Py_LIKELY/Py_UNLIKELY for vectorcall 

jdemeyer:
CC @markshannon @methane 

https://bugs.python.org/issue37774

### DIFF
--- a/Include/cpython/abstract.h
+++ b/Include/cpython/abstract.h
@@ -64,7 +64,7 @@ _PyVectorcall_Function(PyObject *callable)
 {
     assert(callable != NULL);
     PyTypeObject *tp = Py_TYPE(callable);
-    if (!PyType_HasFeature(tp, _Py_TPFLAGS_HAVE_VECTORCALL)) {
+    if (!Py_LIKELY(PyType_HasFeature(tp, _Py_TPFLAGS_HAVE_VECTORCALL))) {
         return NULL;
     }
     assert(PyCallable_Check(callable));
--- a/Include/pymacro.h
+++ b/Include/pymacro.h
@@ -103,8 +103,10 @@
 #define Py_UNREACHABLE() abort()
 
 
-/* If we're using GCC, use __builtin_expect() */
-#if defined(__GNUC__) && (__GNUC__ > 2)
+/* If we are using GCC, use __builtin_expect(). There is no need to check
+ * the GCC version since old versions that don't have __builtin_expect()
+ * can't compile CPython anyway */
+#if defined(__GNUC__)
 #  define Py_UNLIKELY(value) __builtin_expect(!!(value), 0)
 #  define Py_LIKELY(value) __builtin_expect(!!(value), 1)
 #else
--- a/Include/pymacro.h
+++ b/Include/pymacro.h
@@ -102,4 +102,15 @@
 
 #define Py_UNREACHABLE() abort()
 
+
+/* If we're using GCC, use __builtin_expect() */
+#if defined(__GNUC__) && (__GNUC__ > 2)
+#  define Py_UNLIKELY(value) __builtin_expect(!!(value), 0)
+#  define Py_LIKELY(value) __builtin_expect(!!(value), 1)
+#else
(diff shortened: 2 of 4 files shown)


Include/pymacro.h
```diff
@@ -102,4 +102,15 @@
 
 #define Py_UNREACHABLE() abort()
 
+
+/* If we're using GCC, use __builtin_expect() */
+#if defined(__GNUC__) && (__GNUC__ > 2)
```
methane: I think `__GNUC__ > 2` is redundant.

jdemeyer: Redundant because Python won't compile with GCC 2.x anyway?

methane: Because:

* I don't know GCC 2.x supports all of C99 features we use in PEP 7.
* I don't know how to use GCC 2 to compile Python in recent days.
* I don't know who try to compile Python with GCC 2.

methane: I found this page: https://gcc.gnu.org/c99status.html
We use some GCC>=3.0 features already.

* designated initializers
* new block scopes for selection and iteration statements
* stdbool.h (GCC 2.95 had `<stdbool.h>`, but based on an early draft with major differences from C99 semantics.)

methane (approved):

Include/cpython/abstract.h
```diff
@@ -64,7 +64,7 @@ _PyVectorcall_Function(PyObject *callable)
 {
     assert(callable != NULL);
     PyTypeObject *tp = Py_TYPE(callable);
-    if (!PyType_HasFeature(tp, _Py_TPFLAGS_HAVE_VECTORCALL)) {
+    if (!Py_LIKELY(PyType_HasFeature(tp, _Py_TPFLAGS_HAVE_VECTORCALL))) {
```
MojoVampire, line 67: Shouldn't the entire test be inside the macro? It's possible it's fine either way, but every time I've seen __builtin_expect used (and every preexisting use of the LIKELY/UNLIKELY macros), the whole test was put inside the macro, so this would be something like: if (Py_UNLIKELY(!PyType_HasFeature(tp, _Py_TPFLAGS_HAVE_VECTORCALL))) {

jdemeyer: I don't think it matters. Either way would be OK. I found it easier to understand as "`_Py_TPFLAGS_HAVE_VECTORCALL` is likely" instead of "not having `_Py_TPFLAGS_HAVE_VECTORCALL` is unlikely".

pablogsal (approved):
I am not super fan of exposing these macros, by I agree with @methane in that they are battle tested. Do we have some benchmarks numbers for the effect of this (with PGO, as PGO also activates branch-prediction annotations)?

methane:
I don't think these macros are useful for PGO build.  In case of obmalloc.c, it only affects non-PGO build.

jdemeyer:
Rebased to latest master, fixed conflicts in `obmalloc.c`

@markshannon: what's your opinion on this? It achieves the same speedup as #14735 but in a more structural way.

matrixise:
@markshannon do you have time for a review of this PR? Thank you

matrixise (approved):

csabella:
This has three approvals - should it be merged once the conflicts are resolved?

markshannon:
As @methane has said, given PGO does this for us, what's the point?

Having said that. I'd be OK with https://github.com/python/cpython/pull/15144/files#diff-7f17c8d8448b7b6f90549035d2147a9fR4978 being added, as performance is deliberately being sacrificed when `sys.settrace` is used.

As for the others, probably not.

markshannon (commented):
I see no benefit to this.
I'd rather leave this sort of tweaking to tools as they can adapt as the surrounding code changes.

markshannon:
@jdemeyer do you mind if I close this?

markshannon:
This is rather out of date now, so I'm closing it.

kumaraditya303:
Closing as there are conflicts and the benefit is unclear.

PR #15145, opened 2019-08-06
Fix typos mostly in comments, docs and test names

minho42:
Found some more typos in various files.


terryjreedy (changes requested):
1. Remove duplicate changes.  Many of the changes for idlelib were already done in #15018, copied over to #15029, a week ago.  I suspect the same is true of some of the other changes.  You should have branched off of a freshly updated master.  Perhaps you reused your old branch instead of deleting and starting fresh.  For a PR like this, one should refresh with `git merge upsteam/master`  in the branch just before creating a PR.  Anyway, refresh your branch now on your machine and fix the conflicts.  Then push again.

2. Put the new idlelib changes in a separate PR so that they can be backported.

minho42:
@terryjreedy 
I will make a new PR. Thanks.

PR #15146, opened 2019-08-06
bpo-37734: Fix use of registry values to launch Python from Microsoft Store app

zooba:
https://bugs.python.org/issue37734


PC/layout/main.py
```diff
@@ -153,19 +153,25 @@ def in_build(f, dest="", new_name=None):
                 yield "libs/" + n + ".lib", lib
 
     if ns.include_appxmanifest:
+        yield from in_build("python_uwp.exe", new_name="python{}".format(VER_DOT))
+        yield from in_build("pythonw_uwp.exe", new_name="pythonw{}".format(VER_DOT))
+        # For backwards compatibility, but we don't reference these ourselves
```
brettcannon:
```suggestion
        # For backwards compatibility, but we don't reference these ourselves.
```

PC/layout/main.py
```diff
@@ -153,19 +153,25 @@ def in_build(f, dest="", new_name=None):
...
         yield from in_build("python_uwp.exe", new_name="python")
         yield from in_build("pythonw_uwp.exe", new_name="pythonw")
     else:
+        yield from in_build("python.exe", new_name="python{}".format(VER_DOT))
+        yield from in_build("pythonw.exe", new_name="pythonw{}".format(VER_DOT))
+        # For backwards compatibility, but we don't reference these ourselves
```
brettcannon:
```suggestion
        # For backwards compatibility, but we don't reference these ourselves.
```

PR #15150, opened 2019-08-06
bpo-37778: Fixes the icons used for file associations to the Microsoft Store package

zooba:
https://bugs.python.org/issue37778


eamanu (approved):
LGTM

PR #1511, opened 2017-05-09
bpo-21261: IDLE: Add autocomplete when completing dictionary keys

louisom:

PR #15169, opened 2019-08-07
bpo-21261: IDLE - add completion of dict sting keys

taleinat:
This is a new implementation, created from scratch. I intentionally didn't base this on PR GH-1511 after reviewing it thoroughly.

Note that completion is intentionally limited to keys of type str and bytes only.

Additional changes made in this PR:
* The Escape key now closes the completion window. [already done separately]
* Completing a string or bytes dict key adds a closing square bracket.
* Some of the auto-completion tests have been changed a bit to run without a GUI.
* Significant refactoring and cleanup of some existing tests, with better use of mocking.
* The fake Text widget used for tests has had its index handling greatly improved.
* Added a very functional fake Listbox widget class.

https://bugs.python.org/issue21261


Lib/idlelib/autocomplete.py
```diff
@@ -90,6 +92,126 @@ def _delayed_open_completions(self, args):
...
+        escapes.update({
+            7: '\\a',
+            8: '\\b',
+            9: '\\t',
+            10: '\\n',
+            11: '\\v',
+            12: '\\f',
+            13: '\\r',
+        })
+        high_byte_escape_re = re.compile(r'[\x80-\xff]')
```
aeros, line 160: Based on the [travis error log](https://travis-ci.org/python/cpython/jobs/569053155), it looks like this line is likely the culprit (missing empty line before function):
```suggestion
        high_byte_escape_re = re.compile(r'[\x80-\xff]')
        
```

Misc/NEWS.d/next/IDLE/2019-08-07-23-27-54.bpo-21261.sqNQCD.rst
```diff
@@ -0,0 +1 @@
+IDLE's shell auto-completion completes dict keys of type str and bytes.
```
aeros: The main suggestion here is adjusting "auto-completion completes" -> "auto-completion works for" and "type" -> "types". 

The inline markup for str and bytes is an extra addition, sphinx converts this into a link to their stdtypes sections. Respectively: [str](https://docs.python.org/3.9/library/stdtypes.html#str) and [bytes](https://docs.python.org/3.9/library/stdtypes.html#bytes). For an example of this being used in the docs, see [common sequence operators](https://docs.python.org/3.9/library/stdtypes.html#common-sequence-operations), in the section for the first note. If you don't want the link, it could be instead replaced with ``str`` and ``bytes``.

```suggestion
IDLE's shell auto-completion works for dict keys of types :class:`str` and :class:`bytes`.
```

aeros (approved):
Tested using the IDLE shell from the PR branch. OS: Arch Linux 5.2.3. 

Test for str
Before tab:
image
After tab:
image

Test for bytes
Before tab:
image
After tab:
image


Looks like it's working as intended, nicely done! I also tested to ensure that ``excessively_long_key`` and didn't auto-complete when using ``d[b'e`` and the same for ``d['t`` -> ``d['thanks taleinat'``. I intentionally used a space instead of an underscore for the second one (even if it goes against naming convention) to ensure there weren't any issues there. LGTM.

Lib/idlelib/autocomplete.py
```diff
@@ -90,6 +92,126 @@ def _delayed_open_completions(self, args):
...
+        def high_byte_sub(char):
+            """escaping for bytes values above 127"""
+            return f'\\x{ord(char):x}'
+
+        # Split the completions into lists of str and bytes.
+        str_comps = (c for c in comp_list if type(c) == str)
+        bytes_comps = (c for c in comp_list if type(c) == bytes)
+
+        # Create reprs for the str objects.
+        str_repr_template = f"{prefix}{quote_type}{{}}{quote_type}"
```
aeros: Minor QoL suggestion:

In any scenario I can imagine for the dict key auto-complete, the user would end up typing `]` immediately after using `tab`. As a result, it might be beneficial to add it to the end of the auto-completion.

The best place to add this would probably be in the string template:
```suggestion
        str_repr_template = f"{prefix}{quote_type}{{}}{quote_type}]"
```

aeros: Tested this suggestion locally, no issues as far as I can tell.

Before tab:
image
After tab:
image

Lib/idlelib/autocomplete.py
```diff
@@ -90,6 +92,126 @@ def _delayed_open_completions(self, args):
         if self.text.index("insert") == self._delayed_completion_index:
             self.open_completions(args)
 
+    _dict_str_key_prefix_re = re.compile(r'\s*(br|rb|fr|rf|b|f|r|u|)', re.I)
+
+    @classmethod
+    def is_completing_dict_key(cls, hp, curline):
```
aeros: Another minor suggestion. While reading over the changes, the purpose of the *hp* argument in `is_completing_dict_key()` was not clear until I saw line 228. The variable name `hp` isn't as much of an issue on 228 where it's defined, but I think it would make the purpose of the argument in `is_completing_dict_key()` much more clear if it was instead *parser* or *hparser*:

```suggestion
    def is_completing_dict_key(cls, hparser, curline):
```
With this change, anyone reading over the function definition could much more easily understand the purpose of the argument without having to understand the exact implementation details. 

I prefer *hparser*, to avoid any potential confusion with argparse, since *parser* is frequently used as a variable name when creating an instance of the ArgumentParser class.

taleinat: A fine suggestion, I've just done this.

Lib/idlelib/autocomplete.py
taleinat: This is a great idea and I've just implemented it, albeit using a more robust and complete approach.

Misc/NEWS.d/next/IDLE/2019-08-07-23-27-54.bpo-21261.sqNQCD.rst
taleinat: Great suggestions, thanks!

taleinat:
@aeros167, I've addressed all of the issues you brought up, plus your suggestions. I've also completed the implementation and added a ton of tests. If you have the time to give this another go, that would be great!

taleinat:
@terryjreedy, in order to have many of the tests not require a GUI, I've expanded the existing `mock_tk` module quite a bit, most notably adding a rather full-featured Listbox class and expanding the index handling of the Text widget to be comprehensive. Please take a look at this!

I'm a bit worried about the long-term maintenance of this, though Tkinter changes little enough that it may be negligible.

If you prefer that I roll back these additions and just have the relevant tests require having a working GUI environment, I'm fine with that too.

aeros167:
Thanks @taleinat, I'll run through the tests again later tonight and report back with the results.

Lib/idlelib/autocomplete.py
aeros: Awesome! I definitely agree with the usage of a variable instead of appending a constant, that allows far more extensibility. I hadn't thought of using anything besides `]`, but that could definitely be useful.

Edit: I also would not have though of adding this to the `keypress_event()` function instead of just appending as a string to the output. Good idea.

Lib/idlelib/autocomplete_w.py
```diff
@@ -153,6 +173,39 @@ def _selection_changed(self):
...
+        # the closing quote(s), then the quote hasn't been closed.
+        i = len(start) - (len(quote) - 1)
+        while (
+                i >= quote_index + len(quote) + 2 and
+                start[i - 2:i] == '\\\\'
+        ):
+            i -= 2
+        return start[i - 1] != '\\'
+
     def show_window(self, comp_lists, index, complete, mode, userWantsWin):
```
aeros: Could we potentially change the parameter `userWantsWin` -> `user_wants_win` or `wants_win`? This already existed prior to this change, but it looks a bit odd to have one parameter within the function use the typical underscore naming convention and another using camelCase.

```suggestion
     def show_window(self, comp_lists, index, complete, mode, wants_win):
```

aeros (approved):
Tested the dict key auto-completion again after the most recent changes:

Before tab:
image
After tab:
image
Looks like it's worked as intended. Nicely done.

Let me know if there's any more in-depth testing you'd like for me to do.

terryjreedy:
Since Kyle's suggestions seemed sensible to good, I left it to you to respond.  Now that you have, I plan to review tomorrow. 

I have been wanting to add some of the scattered tk mocks used or usable in more than one test file to mock_tk, so that they can be polished and possibly tested.  Some mocks should probably be used more.  Long ago, I tested mock text by running the test I wrote it for with both with real Text and mock Text (on Windows).  Expanding mock Text indexing has been on my todo list since.

Lib/idlelib/autocomplete_w.py
terryjreedy: Yes, we have be gradually PEP8-ifying idlelib as modules are otherwise reviewed and edited *and tests added*.  And shorter is fine here.

taleinat: This argument is actually no longer used for anything as far as I can tell. I double-check and then remove it.

taleinat:
> I have been wanting to add some of the scattered tk mocks used or usable in more than one test file to mock_tk, so that they can be polished and possibly tested. Some mocks should probably be used more.

Ah, then I think you'll like what I've done there! :)

Lib/idlelib/autocomplete_w.py
taleinat: Removed in latest commit.

Lib/idlelib/autocomplete_w.py
```diff
@@ -24,6 +29,13 @@
 WINCONFIG_SEQUENCE = "<Configure>"
 DOUBLECLICK_SEQUENCE = "<B1-Double-ButtonRelease>"
 
+
+_quote_re = re.compile(r"""["']""")
+def _find_first_quote(s):
```
aeros:
```suggestion
def _find_first_quote(line):
```

Edit: Oops, this should be line, not quote.

Lib/idlelib/autocomplete_w.py
```diff
@@ -24,6 +29,13 @@
...
+    m = _quote_re.search(s)
```
aeros:
```suggestion
    m = _quote_re.search(line)
```

aeros (commented):
This is a similar recommendation to the one in first PR review I had made, changing a single letter parameter name to something slightly more descriptive. It might not be needed in this case since it's a brief underscore function, but I figured it wouldn't hurt to ask. I don't think `m` as a variable name is an issue, since it's used frequently enough for regex matches and not a parameter.

Edit: The variable suggestion should be `line`, not quote.

Lib/idlelib/autocomplete_w.py
taleinat: Thanks for the suggestion! I've done this, and otherwise improved the function e.g. adding a doc-string.

aeros: Thanks! Doc-strings are always a good idea to include.

terryjreedy (changes requested):
Initial use test:
`>>> globals()[`
Wow.  The popup seems so natural I had to remind myself that it is new.  I definitely want this in 3.8.0b4, due  Monday, 2019-08-26.

But... given the current completion design, calling a function for an auto popup is a bug.  Dict completion should act the same as attribute completion (and calltips).
`>>> globals().`  # No attribute completion popup unless explicitly request (force, control-space).
`>>> g = globals(); g.`  # Attribute completion pops up, starting with 'clear'.

Bug 2.  If I close the box (Esc, immediate Enter (see below), click outside the box) and enter a 2nd '[', the 1st '[' is ignored and a box pop up.  Same if close and enter 3rd '[', and so on
`>>> g[[`  # Completion box for g pops up.  But `get_entity('g[')` (or for 'g[[', etc) should fail.

I would have preferred at least 2 separate PRs, the first being for changing existing tests and improving mock_tk.  But separate PRs would add calendar time, and I realize that you probably did not know what you wanted to add before working on new tests.  I will, however, review those separately.  It helps a lot that you make multiple commits and don't squash them.

Though a separate issue, closing with Escape is great, and part of making all popups close with Escape.  It opens the opportunity to reassess how completion boxes open and possibly no longer using Enter to close without completing.  But let's leave that for another issue.  I prefer no more feature creep on this PR ;-).  This addition should be mentioned on the blurb.

idle.rst (and help.html with it) need revision.  I can do that.

taleinat:
Good catch, Terry! I've fixed the coding error that caused both of those bugs.

terryjreedy:
Use test 2: Bugs 1 and 2 are fixed, but there are 2 more plus a question.
1. `globals()[` no longer auto pops a box; but does with ^space. 
2. After `g=globals()`, `g[[` no longer opens a box;   neither does ^space. 

Bug 3. The keys completion box should only have the dict keys, as with attributes, but not the global list.

"Escape key closes the completion window" does not say what happens to the editor line.  Currently, it is almost* the same as Return.  If an item was selected, it stays.  I was expecting it to cancel any change to the text, as if the completion window never happened.  What did you intend?  (This is an example of why we need to start with better specs.)

* If the text has a prefix of exactly one possible completion, Return complete it, Escape does not.
Return adds ']', Escape does not,

Bug4. Using Show Completions ^space for attributes as a reference, the same for dict keys (^space) has bugs.  Given `g.|c|o|p|y|`, where | is a possible cursor position, ^space brings up the attribute completion box.  Given something like g["__builtins_"], with the cursor anywhere after '[' and before the closing " raises 

Exception in Tkinter callback
  File "F:\dev\3x\lib\tkinter\__init__.py", line 1883, in __call__
    return self.func(*args)
  File "F:\dev\3x\lib\idlelib\multicall.py", line 176, in handler
    r = l[i](event)
  File "F:\dev\3x\lib\idlelib\autocomplete.py", line 62, in force_open_completions_event
    self.open_completions(FORCE)
  File "F:\dev\3x\lib\idlelib\autocomplete.py", line 277, in open_completions
    comp_what_and_start = self._is_completing_dict_key(hp)
  File "F:\dev\3x\lib\idlelib\autocomplete.py", line 139, in _is_completing_dict_key
    hp.set_index("insert-%dc" % (len(hp.rawtext) - opening_bracket_idx))
  File "F:\dev\3x\lib\idlelib\hyperparser.py", line 89, in set_index
    raise ValueError("Index %s precedes the analyzed statement"
ValueError: Index insert-16c precedes the analyzed statement

(16 is just after the closing char).  This is definitely wrong.  I would like the keys box back, as with attributes.
If the  cursor is after the closing ", we get global name completions.  If immediately after the closing quote, the keys box would be better, analogous to when the cursor immediately follows the last attribute char.

If I remove ']', then 16 becomes 15 and ^space does nothing when the cursor is just before the closing quote.

Lib/idlelib/autocomplete.py
```diff
@@ -203,6 +396,14 @@ def fetch_completions(self, what, mode):
                 except OSError:
                     return [], []
 
+            elif mode == DICTKEYS:
+                try:
+                    entity = self.get_entity(what)
+                    keys = entity.keys() if isinstance(entity, dict) else []
+                except:
+                    return [], []
+                smalll = bigl = [k for k in keys if type(k) in (bytes, str)]
```
aeros, line 447: > Bug 3. The keys completion box should only have the dict keys, as with attributes, but not the global list.

@terryjreedy Could we fix this by adding another conditional filter to the list comprehension?

```suggestion
                smalll = bigl = [k for k in keys if type(k) in (bytes, str) and k not in globals()]
```

taleinat:
> Bug 3. The keys completion box should only have the dict keys, as with attributes, but not the global list.

This is intentional: When completing dict keys, it is still useful to be able to complete variable names from the global namespace, since one might want to use them as the key. Completing such variable names is currently possible before this change, and I think disabling this would be a UX regression.

Having said that, we can separately consider dict key completions with or without an opening quote. When completion is invoked after an opening quote, it is clear that a string completion is desired, and perhaps then only string/bytes completions should be offered.

Lib/idlelib/autocomplete.py
taleinat: @aeros167, this was done intentionally (see details in separate comment.)

Lib/idlelib/autocomplete.py
```diff
@@ -90,14 +93,177 @@ def _delayed_open_completions(self, args):
...
+        if prefix_match is not None:
+            prefix = prefix_match.group(1)
+            post_prefix = comp_start[prefix_match.end():]
+
+            is_raw = 'r' in prefix.lower()
+            if post_prefix:
+                quote_type_candidate = post_prefix[0]
+                if quote_type_candidate not in "'\"":
+                    pass  # use the default quote type
+                elif post_prefix[1:2] == quote_type_candidate:
```
aeros: > When completion is invoked after an opening quote, it is clear that a string completion is desired, and perhaps then only string/bytes completions should be offered.

@taleinat Could this be done by filtering ``comp_list`` after ``quote_type_candidate`` is determined to be either ``'`` or ``"``?
```suggestion
                comp_list = [c for c in comp_list if c not in globals()]
                elif post_prefix[1:2] == quote_type_candidate:
```
Or is there a better location to do this?

Lib/idlelib/autocomplete.py
terryjreedy: I don't see from the snippet given how the 230+ globals are there.  But they should never be in keys.

Lib/idlelib/autocomplete.py
```diff
@@ -106,7 +272,17 @@ def open_completions(self, args):
         hp = HyperParser(self.editwin, "insert")
         curline = self.text.get("insert linestart", "insert")
         i = j = len(curline)
-        if hp.is_in_string() and (not mode or mode==FILES):
+
+        if mode in (None, DICTKEYS):
```
terryjreedy, line 318: It looks like this suite should be in the block for `elif mode in (None, DICTKEYS):` (revised from  `elif mode == DICTKEYS:` ).

Lib/idlelib/autocomplete.py
```diff
@@ -137,18 +313,35 @@ def open_completions(self, args):
...
             return None
         comp_lists = self.fetch_completions(comp_what, mode)
+        if mode == DICTKEYS:
+            assert comp_lists[0] == comp_lists[1]
+            reprs = self._dict_key_reprs(comp_start, comp_lists[0])
+
+            if hp.is_in_code():
+                # Add variables in the global namespace to the comp lists.
+                comp_lists = (reprs, list(reprs))
+                extra_comp_lists = self.fetch_completions("", ATTRS)
```
terryjreedy: This should be deleted.

terryjreedy (changes requested):
A bare beginning of code review.  It is more complicated than I hoped.  I see the problem with strings (bytes) just being strings, while string (bytes) literals have all the complications.  Even without that, strings can be indefinitely long, so we have to truncate.  A later issue could be to allow a wider box if needed.
Replacing newlines in multiline strings with '\n' seems like a good idea also.

terryjreedy:
Bug 5: keys with astral chars raise.  Traceback goes to parent console, if there is one.  No box.
```
>>> a = 'ab\U00011111ba'
>>> b = {a:5}
>>> b
{'ab\U00011111ba': 5}
>>> b[
```
Raises, in parent console,
```
Exception in Tkinter callback
  File "F:\dev\3x\lib\tkinter\__init__.py", line 1883, in __call__
    return self.func(*args)
  File "F:\dev\3x\lib\tkinter\__init__.py", line 804, in callit
    func(*args)
  File "F:\dev\3x\lib\idlelib\autocomplete.py", line 94, in _delayed_open_completions
    self.open_completions(args)
  File "F:\dev\3x\lib\idlelib\autocomplete.py", line 281, in open_completions
    self._remove_autocomplete_window()
  File "F:\dev\3x\lib\idlelib\autocomplete.py", line 57, in _remove_autocomplete_window
    self.autocompletewindow.hide_window()
  File "F:\dev\3x\lib\idlelib\autocomplete_w.py", line 541, in hide_window
    self.widget.event_delete(HIDE_VIRTUAL_EVENT_NAME, seq)
  File "F:\dev\3x\lib\idlelib\multicall.py", line 392, in event_delete
    triplets.remove(triplet)
ValueError: list.remove(x): x not in list
```
After ^space,
```
  File "F:\dev\3x\lib\tkinter\__init__.py", line 1883, in __call__
    return self.func(*args)
  File "F:\dev\3x\lib\idlelib\multicall.py", line 176, in handler
    r = l[i](event)
  File "F:\dev\3x\lib\idlelib\autocomplete.py", line 62, in force_open_completions_event
    self.open_completions(FORCE)
```
and same as before.

terryjreedy:
Bug 6, related, maybe same, to 4.  `globals('ab\U00011111ba') = 5` is a SyntaxError.  Retrieve and edit by inserting ')[' after '('.  Raises 
```
  File "F:\dev\3x\lib\idlelib\autocomplete.py", line 94, in _delayed_open_completions
    self.open_completions(args)
  File "F:\dev\3x\lib\idlelib\autocomplete.py", line 277, in open_completions
    comp_what_and_start = self._is_completing_dict_key(hp)
  File "F:\dev\3x\lib\idlelib\autocomplete.py", line 139, in _is_completing_dict_key
    hp.set_index("insert-%dc" % (len(hp.rawtext) - opening_bracket_idx))
  File "F:\dev\3x\lib\idlelib\hyperparser.py", line 89, in set_index
    raise ValueError("Index %s precedes the analyzed statement"
ValueError: Index insert-22c precedes the analyzed statement
```

terryjreedy:
Bug 5, continued.
```
>>> a  # Tab pops up box at 'abs'.
>>> globals()['ab\U00011111ba'] = 5  # Put astral name into globa scope.
>>> a  # Tab does nothing, ^space raises same "ValueError: list.remove(x): x not in list"
>>> del globals()['ab\U00011111ba']  # Completions not restored.
```

Lib/idlelib/autocomplete.py
aeros: >I don't see from the snippet given how the 230+ globals are there. But they should never be in keys.

@terryjreedy Hmm, this suggestion might be in the wrong location then. This PR was the first time I've read through idlelib/autocomplete.py. Based on my understanding, this was list of potential completions for dict keys, so filtering out any of the keys which were located in ``globals()`` would address bug 3.

@taleinat I made another suggestion in a comment below which performs filtering from ``globals()`` only when the completion is started after an opening quote.

taleinat:
> 2. After `g=globals()`, `g[[` no longer opens a box; neither does ^space.

What would you expect to be shown after `g[[` and ^space? The usual global variable names completion list?

taleinat:
> Bug4. [...] Given something like g["__builtins_"], with the cursor anywhere after '[' and before the closing " raises

Fixed.

Lib/idlelib/autocomplete.py
taleinat: I'm not sure what you mean... I put this block at the top intentionally. Note that it is followed by `if` rather than `elif`, so that if the block doesn't decide that dict key completion should take place, the other completion modes are checked.

Lib/idlelib/autocomplete.py
taleinat: Please see my comment above re. intentionally including global variables in the lists.

taleinat:
@terryjreedy, the astral (non-BMP) character-related issues should all be resolved now, as well as all other previous bugs, except for "Bug 3". I'm waiting on your response re. keeping global variables in the list for e.g. `d[`, but not for e.g. `d["`.

terryjreedy:
My comment beginning '2.'  was that the bug of g[[ ^space opening dict key list was no longer present.  But without a screenshot of the ephemeral box, I can no longer swear that it was, as opposed to the global namelist.  If it was opening the latter, sorry for the mistake.

The Completions doc says "‘Show Completions’ will force open a completions window".  I take this to mean it should always open something, the default being the global namelist if none of the special cases applies, even if the result will not be legal in the resulting state of the text. If you were referring to this in one of your statements that I did not understand, I again apologize.  

Currently, ^space after '[' and '[[' and 'd[[' opens the global namespace, and it should continue to do so.  In all three cases, the immediately preceding '[' can only be the beginning of a list display or comprehension, not a subscription, and any object is legal as a list item.  The fact that the list resulting from [...] will not be a legal key in the 3rd case is not the business of completions to detect or worry about.

Lib/idlelib/autocomplete.py
terryjreedy: I did notice, but it is not yet obvious why the preliminary check in needed if it can be overriden.  But leave it for now.

Lib/idlelib/autocomplete.py
terryjreedy: I will try to explain better in a post to the issue why I am currently opposed to this.

Lib/idlelib/autocomplete.py
```diff
@@ -183,10 +183,14 @@ def control_char_escape(match):
         def high_byte_sub(match):
             """escaping for bytes values above 127"""
             return f'\\x{ord(match.group()):x}'
+        non_bmp_re = re.compile(r'[\U00010000-\U0010ffff]')
```
terryjreedy: As suggested by my email, this and the function or something like them, and probably the code that calls it, is needed many other places.  But I don't know yet where it should go.  Thanks for the start.

Lib/idlelib/autocomplete.py
```diff
@@ -183,10 +183,14 @@ def control_char_escape(match):
         def high_byte_sub(match):
```
terryjreedy, line 161: I think non-ascii BMP chars should be left alone, as they currently are for tab-completion. 
```
>>> あいうえ = 3
>>> a<tab>
```
displays a box with `あいうえ` sorted at the end. `\\x3042\\x3044\\x3046\\x3048` would be useless to most people using hiragana.  I think we should optimize for people who sensibly limit string keys as well names to printable unicode.

taleinat: The "high bytes" substitution is only used for bytes literals, and it is required for those literals to be valid. According to [the docs](https://docs.python.org/3/reference/lexical_analysis.html#string-and-bytes-literals):

>  They may only contain ASCII characters; bytes with a numeric value of 128 or greater must be expressed with escapes.

taleinat:
@terryjreedy

> Currently, ^space after '[' and '[[' and 'd[[' opens the global namespace, and it should continue to do so.

I agree; I'll look into it.

terryjreedy:
Bug 7: (noticed on first review, maybe forgot to add, still present and a great nuisance especially for review) When box opens 'above', it is half line+ too low, and mostly covers the text being completed.

Lib/idlelib/autocomplete.py
terryjreedy: Got it.

taleinat:
> Bug 7: When box opens 'above', it is half line+ too low, and mostly covers the text being completed.

This is unrelated and reproduces with current master as well.

See bpo-37849.

terryjreedy:
I believe the change to What's New 3.8 will prevent the auto backport to 3.7, where the file does not exist. I have usually changed What's New in separate issues, currently bpos 33821 and 33821.  (That is partly because What's New has been an after thought.)  Since the change to the earliest version (now WN 3.7) will backport all the way, just the change to WN 3.8 needs to be pulled to a separate PR, also attached to 21261, that is only backported to 3.8.




  After 3.8.0, when What's New 3.9 is added, I will add an issue for that and 3 separate PRs will be needed to avoid manual backport.

taleinat:
> I believe the change to What's New 3.8 will prevent the auto backport to 3.7, where the file does not exist.

I'm happy to execute that manual backport. It's simpler for me than having separate PRs.

Lib/idlelib/autocomplete.py
```diff
@@ -4,29 +4,32 @@
 pop up a list of candidates.
 """
 import __main__
+import itertools
 import os
+import re
 import string
 import sys
 
 # Two types of completions; defined here for autocomplete_w import below.
```
decaz: *Three* types (or modes?) of completions; ...

taleinat: Yes, thanks!
```suggestion
# Three types of completions; defined here for autocomplete_w import below.
```

taleinat:
I've fixed completions after opening brackets when there are no available dict key completions, e.g. `[`, `globals()[[`, `{}[` and `{1: 1}[`. I've also ensure things work properly after typing an opening bracket (`[`) in a string.

At the same time, I've removed all cases where dict keys were mixed with global variables in the completion list, as requested by @terryjreedy.

@terryjreedy, that addresses all outstanding issues with this PR.

taleinat:

taleinat:
I've merged in the updated `master` branch and removed the outdated "What's New" entries.

This is still waiting for @terryjreedy's review.

taleinat:
I must say that I'm especially sad that this has not made it into IDLE. :(

terryjreedy:
I did not proper cognize until today that mixing in the tab list had been undone.  Moving on...

I want to handle mock-tk stuff separately. I believe that your patch would supersede those of issue 18504.  I am reserving comments and change request for that or a new issue.  I will make the new PR if you want.  Nothing should be removed from this patch until obsoleted by merging the new PR.

As for dict string key completion:  I am surprised (and not happy) that a seemingly simple and marginally useful addition should nearly double the code in autocomplete.py.  So I will be looking to see if anything can be shrunk.

I redid all the experiments discussed in my bug list with string keys and a very short delay   All the lists look correct.  The main issue I encountered was navigating the list by typing partial entries.  I discussed this as part of post on the issue about reconsidering inclusion of bytes keys.

terryjreedy:
Changed title to reflect that string key completion works in editor (once code is run, as with imports and definitions) and upon request.

taleinat:
> As for dict string key completion: I am surprised (and not happy) that a seemingly simple and marginally useful addition should nearly double the code in autocomplete.py. So I will be looking to see if anything can be shrunk.

Indeed, I'm not convinced that including this feature is worth the extra code and effort. Removing the support for bytes would help somewhat, but the support for string keys would still be complex.

I'd be fine if we decide not to implement this feature: If it means that the work in making this PR helped us make the decision, then it was worth it!

taleinat:
Closing in favor of a newer version of this PR, GH-26039.

PR #15170, opened 2019-08-07
bpo-37772: fix zipfile.Path.iterdir() outputs

shireenrao:
https://bugs.python.org/issue37772


ericvsmith (dismissed):
Could you create a test for this? Preferably something that fails without your fix, but succeeds with it?

shireenrao:
Should both the tests be created in the same branch where I have this fix? Any thoughts or recommendations on how to have the old and new versions of _add_implied_dirs? Can I mock the old version in my test file?

ericvsmith:
I mean must have one test. That test would fail when run against the code in master, but would succeed when run against your branch. So the net result is just a test checked in to your branch, and the test would succeed. But by knowing it fails when run against master, you know it's actually testing your fix. I hope that makes sense.

jaraco:
@shireenrao: The way I accomplish this test-driven-development approach with git is to create a different branch from master (or wherever your fix branched from master), write the test there. It should fail because you don't have your fix yet. Then, merge the test into the branch with the fix. Then, the tests should pass. Push this to the pull request and you should be good to go.

shireenrao:
@ericvsmith - I added tests for the two scenarios explained in the original issue at https://bugs.python.org/issue37772. For my tests to work, I also had to add changes to add_dirs  method in test_zipfile as this method had the same bug I fixed in zipfile.py. The tests succeed in my bugfix branch and fails in the master branch. I am concerned about the exception which gets thrown in the middle of a generator when running the assert statement. How do I gracefully fail this test?

shireenrao:
My Apologies Everybody! I think I messed up. I merged the latest changes from master into my bugfix branch and pushed it along with my new tests.

jaraco:
I don’t think you messed up. If the diff looks right then you’re fine, and everything will get squashed on merge... so don’t worry about the noise from the merged commits.

shireenrao:
@jaraco - thank you for letting me know I didn’t mess up! Seeing all the new reviewers added after my git push got me worried.

jaraco:
Looking more closely, it does look as if something has gone wrong and this PR now seems to be including the diffs from other commits. I'm going to see if I can cherry-pick the commits to reconstruct the intended diff.

jaraco:
It looks like I was able to fix the diff by merging master back into the feature branch (cedac04).

shireenrao:
Thank you! So what’s best practice? Once I have a feature branch, I should not merge origin?

Lib/test/test_zipfile.py
```diff
@@ -2403,10 +2403,19 @@ def add_dirs(zipfile):
     any directories implied by the presence of children.
     """
     names = zipfile.namelist()
+    subdirs = set([
```
jaraco: Remove the `[` and `]` from these expressions. They're superfluous and cause the sequence to be materialized as a list before constructing the set.

Lib/test/test_zipfile.py
```diff
@@ -2403,10 +2403,19 @@ def add_dirs(zipfile):
...
+        name for name in map(posixpath.dirname, names)
+        if name and name + "/" not in names
+    ])
+    missingdirs = set([
+        str(p) for sd in subdirs
```
jaraco: Here and on line 2407, I suggest putting the `for` on a separate line, especially when there are multiple `for` statements, as that more clearly signals the different loops involved.

Lib/test/test_zipfile.py
jaraco: This whole function (`add_dirs`) looks suspiciously-similar to the code in `zipfile.Path._add_implied_dirs`. When these functions were simple, it maybe made sense to have some duplication across them, but if they're going to grow all of this new behavior, it probably makes sense to extract the common logic and re-use it.

jaraco: On the other hand, it might be important to keep these separate so as to avoid the tests and implementation being over-matched. Really what we need is a way to "determine implied dirs" from a set of names, have a test for that functionality, and then re-use that function.

Lib/test/test_zipfile.py
```diff
@@ -2431,6 +2440,46 @@ def build_abcde_files():
     return zf
 
 
+def build_abcdef_files():
```
jaraco: I wasn't happy with the name `abcde` and now I'm even less happy to see that concept proliferate. If possible, I'd like to create a single example that represents the full breadth of supported files. Or if that's not possible, these fixtures should probably have more meaningful names.

Lib/test/test_zipfile.py
```diff
@@ -2442,6 +2491,18 @@ def zipfile_abcde(self):
         with self.subTest():
             yield add_dirs(build_abcde_files())
 
+    def zipfile_abcdef(self):
```
jaraco: These tests should describe what it is they're attempting to test (or have a docstring to clarify).

Lib/zipfile.py
```diff
@@ -8,6 +8,7 @@
 import importlib.util
 import io
 import os
+import pathlib
```
jaraco: I think including pathlib might be problemmatic, mainly because of the added dependency on the backport. I encountered that [same concern here](https://github.com/jaraco/zipp/pull/10#issuecomment-511134529) and [here](https://github.com/jaraco/zipp/pull/11).

Lib/zipfile.py
```diff
@@ -2228,11 +2229,18 @@ def joinpath(self, add):
...
     @staticmethod
     def _add_implied_dirs(names):
-        return names + [
+        subdirs = set([
             name + "/"
             for name in map(posixpath.dirname, names)
             if name and name + "/" not in names
-        ]
+        ])
+        missing_dirs = set()
```
jaraco: I prefer the pattern used in the test file, where the set was constructed in a single expression instead of an init/for/append pattern.

Lib/zipfile.py
```diff
@@ -2228,11 +2229,18 @@ def joinpath(self, add):
...
+        for sd in list(subdirs):
+            for p in pathlib.PurePath(sd).parents:
+                if str(p) not in {".", "/"} and str(p) + "/" not in subdirs:
+                    missing_dirs.add(str(p) + "/")
+        subdirs.update(missing_dirs)
```
jaraco: Since subdirs is already a set, it's probably not necessary for missing_dirs to be a set (just an iterable).

Lib/zipfile.py
```diff
@@ -2228,11 +2229,18 @@ def joinpath(self, add):
 
     @staticmethod
     def _add_implied_dirs(names):
-        return names + [
+        subdirs = set([
```
jaraco: This technique is likely to change the order of the names, which seems undesirable to me. Probably better would be to use [unique_everseen](https://docs.python.org/3/library/itertools.html#itertools-recipes) to eliminate duplicates without changing order.

jaraco (changes requested):
Thanks for continuing to push this along. It's looking good.

I've gone through and made some modest suggestions that I'd like to see before accepting this.

There's another concern here and that's the backport. Whatever work is done here needs also to be ported to [the backport](/jaraco/zipp).

To the extent that you're comfortable, please proceed with the suggestions.

jaraco:
> Thank you! So what’s best practice? Once I have a feature branch, I should not merge origin?

Honestly, I can't tell where things went wrong. From the codebase that I checked out, it looked to me like you did everything correctly and it was GitHub that was confused, perhaps because of the multiple merges. You should feel free to merge with origin/master (and often you have to do that or rebase against origin/master to resolve merge conflicts).

jaraco:
If you're able to remove those other reviewers from this PR, please do so (I cannot).

ericvsmith:
My only design concern in this whole thing is importing pathlib. Currently importing zipfile won't require pathlib to be imported, but now it will be. Maybe it's not worth worrying about, since zipfile isn't needed for interpreter startup. Thoughts?

shireenrao:
I will try and see if I can use  zipfile.Path.parent instead. It was just easier to get all parents at once with pathlib.

ericvsmith:
I'm not completely opposed to pathlib, but I think it will raise some eyebrows. How much code would you have to copy out of pathlib? 

In general, I like using pathlib (we added it for a reason!). I just want to see what our options are here.

It looks like there are three options here:

1. Use pathlib.

Lib/test/test_zipfile.py
shireenrao: I feel like the function (``add_dirs``) is not needed as _add_implied_dirs does exactly the same thing. ``add_dirs`` is also only used in the tests. I think we should just remove add_dirs altogether. What do you think?

shireenrao: I have fixed this in my branch and haven't pushed it yet. I will push it with all the rest of the changes.

Lib/test/test_zipfile.py
shireenrao: done.

Lib/test/test_zipfile.py
shireenrao: I can change function (``build_abcde_files``) to be ``build_files`` and make it a generator where all three types of files are returned one by one. The problem I will face with doing that is that all tests are dependent on the type of zipfile and its structure, so one test will not work for all file types. Do you have any thoughts or suggestions on this?

Lib/zipfile.py
shireenrao: done

Lib/zipfile.py
shireenrao: Thanks for showing unique_everseen. As I understand, this is a new function I will have to add to zipfile.py, correct? This will also guarantee order.

shireenrao:
@ericvsmith - what are the other two options?

ericvsmith:
I pressed "send" too early. I've edited the comment to finish my thought.

Lib/zipfile.py
shireenrao: I have updated my PR with a bunch of changes based on your review including using unique_everseen. I am still thinking about the naming issue you brought up for testing the 3 files.

Lib/zipfile.py
shireenrao: I am working on putting together the 3 options on solving this as suggested by @ericvsmith below.

shireenrao:
Using zipfile.Path.parent will be an issue because when we are trying to figure out the directories in ``_add_implied_dirs`` we are dealing with naive files and paths, where as ``parent`` is a property on the Path object. From what I see the parent can be used after these naive files and paths become a Path object in iterdir() after line 2214. Am I correct?

shireenrao:
Thank you @ericvsmith 🙏 for removing the other reviewers. How did you do this? I didn’t see any option for me to do it.

Lib/test/test_zipfile.py
shireenrao: Another thought is to change the method ``zipfile_abcde`` to add be called zipfile_type with an added parameter which would pass in the string for file name. Based on the filename we can then yield the file?

ericvsmith:
@shireenrao: there's a gear icon to the right of the Reviewers label (above the list of reviewers). Click on it, get a dropdown, uncheck the ones I don't want. But I swear I didn't see this earlier when I looked, and I might also have more permissions than you.

Lib/test/test_zipfile.py
jaraco: `add_dirs` is used in the tests to create two variants of a zip file, one with implicit dirs and one with explicit dirs. `zipfile.Path` is designed to work with both, so it's important that the tests exercise both variants.

Lib/test/test_zipfile.py
jaraco: I want to move away from that naming altogether. I chose `abcde` because it was not important what it was named (it was the only thing). Now that we have new variants (although I'm not sure we need them), we need more meaningful names that describe the purpose of the zipfiles that the function is generating.

jaraco:
> It looks like there are three options here:
> 
>     1. Use pathlib.
>     2. Copy code from pathlib.
>     3. Use zipfile.Path.parent.

I think there's another option, and that's to write a standalone function like `parents()` that uses the lower-level posixpath primitives to extract all parent dirs from a path.

shireenrao:
@jaraco  - As you mentioned, I have added a standalone function called ``_parents`` and removed the use of pathlib.

Lib/zipfile.py
shireenrao: I have removed pathlib in my latest update.

ericvsmith:
> > It looks like there are three options here:
> > ```
> > 1. Use pathlib.
> > 2. Copy code from pathlib.
> > 3. Use zipfile.Path.parent.
> > ```
> 
> I think there's another option, and that's to write a standalone function like `parents()` that uses the lower-level posixpath primitives to extract all parent dirs from a path.

I think that's my option 2.

shireenrao:
I implemented ``_parents`` without copying code from pathlib and just used ``posixpath``. Should I being investigating what it will take to copy pathlib code here?

ericvsmith:
> I implemented `_parents` without copying code from pathlib and just used `posixpath`. Should I begin investigating what it will take to copy pathlib code here?

@shireenrao : No, my suggestion was just an implementation detail. There's no need to literally copy the code, as long as what you have works.

Lib/test/test_zipfile.py
shireenrao: @jaraco - based on my latest update where I added a standalone ``_parents`` method, Should I add tests for ``_add_implied_dirs``?

jaraco: I'm not sure I fully understand the question, but I don't believe the code needs tests for internal behavior as long as the user-facing behavior is exercised.

jaraco (approved):

shireenrao:
@jaraco - Thank you for your help.

jaraco:
That looks like a false positive to me; ignoring.

PR #15171, opened 2019-08-08
bpo-37760: Mark all generated Unicode data headers as generated.

gnprice:
This causes them to be collapsed by default in diffs shown on GitHub.

Automerge-Triggered-By: @benjaminp


benjaminp (approved):

benjaminp:
(Added the non-3.8 tags by mistake.)

PR #15173, opened 2019-08-08
bpo-25810: Clarify eval() docs, it does not keywords (GH-15173)

smokephil:
[bpo-25810](https://bugs.python.org/issue25810): Clarify eval() docs, it does not keywords (GH-15173)

https://bugs.python.org/issue25810

### DIFF
--- a/Doc/library/functions.rst
+++ b/Doc/library/functions.rst
@@ -454,7 +454,7 @@ are always available.  They are listed here in alphabetical order.
               n += 1
 
 
-.. function:: eval(expression, globals=None, locals=None)
+.. function:: eval(expression[, globals[, locals]])
 
    The arguments are a string and optional globals and locals.  If provided,
    *globals* must be a dictionary.  If provided, *locals* can be any mapping


the-knights-who-say-ni:

Doc/library/functions.rst
```diff
@@ -454,7 +454,7 @@ are always available.  They are listed here in alphabetical order.
               n += 1
 
 
-.. function:: eval(expression, globals=None, locals=None)
+.. function:: eval(expression[, globals[, locals]])
```
aeros, line 457:
```suggestion
.. function:: eval(expression, globals=None, locals=None, /)
```
This clearly communications to the users that *expression*, *globals*, and *locals* are positional only parameters.

aeros (changes requested):
Welcome and thanks for the contribution!

Based on the comment from R. David Murray:
>I think we are moving toward using (and documenting in the main docs) the / notation.  Especially since it shows up in the inspect module as well as the docstring help.

as well as the rest of the discussion on the issue, I think it would be more appropriate to use the / notation:

Doc/library/functions.rst
smokephil: I think it's easier to see with the square bracket notation that these are positional only arguments. It is also consistent with the exec documentation.

aeros: As mentioned in the issue discussion, the square bracket notation was previously used, and the `/` notation was specifically created to reduce the ambiguity. Without it, it's not entirely clear to users that they should not use keyword arguments for the *globals* and *locals* parameters. IMO, exec should be updated as well, unless a consensus was reached that differs from what was stated on the issue page.

We'll have wait on feedback from the core devs. In the meantime, I would recommend looking over [PEP 570](https://www.python.org/dev/peps/pep-0570/) if you're not already familiar with it.

gpshead: This PR is fine.  We use the traditional `[, []]` syntax all over the place in this file.  We do not use `, /)` anywhere in this .rst file.

`exec` is already documented using []s instead of wrongly implying it accepts named arguments.

aeros: @gpshead:
> This PR is fine. We use the traditional [, []] syntax all over the place in this file. We do not use , /) anywhere in this .rst file.

Is there any intention to move towards the usage of the PEP 570 syntax within this page at some point? I'm good with this PR using the old syntax if you're okay with it (since it's still an improvement from before either way), but my impression was that it was our intention to gradually move towards using it across the documentation. 

> exec is already documented using []s instead of wrongly implying it accepts named arguments.

Isn't the main purpose of the PEP 570 syntax to have a clear and consistent way to explicitly differentiate positional only parameters from named ones? ``param1=None, /, param2=None`` states that ``param1`` is an optional parameter that can't be passed a named argument and ``param2`` is an optional parameter that accepts positional or named arguments. That format is more clear than the brackets, which can get convoluted with the layers of nesting rather easily.

PR #15175, opened 2019-08-08
bpo-37788: fix reference leak caused by threading._shutdown_locks

akruis:
If a program creates non-daemon threads but does not join them, the global set `threading._shutdown_locks` accumulates the shutdown locks of the threads. 

This pull request adds a test case for non-daemon threads without join. The test must not leak references, if called with `python -m test -R: test_threading`.

This pull request also fixes the reference leak. The `tstate->on_delete`-hook now discards per thread shutdown locks from threading._shutdown_locks.

### DIFF
--- a/Modules/_threadmodule.c
+++ b/Modules/_threadmodule.c
@@ -1195,12 +1195,22 @@ This function is meant for internal and specialized purposes only.\n\
 In most applications `threading.enumerate()` should be used instead.");
 
 static void
-release_sentinel(void *wr_raw)
+release_sentinel(void *data)
 {
-    PyObject *wr = _PyObject_CAST(wr_raw);
+    PyObject *wr = ((PyObject **)data)[0];
     /* Tricky: this function is called when the current thread state
        is being deleted.  Therefore, only simple C code can safely
        execute here. */
+    lockobject * shutdown_locks_lock = (lockobject *) ((PyObject **)data)[1];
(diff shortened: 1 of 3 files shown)


vstinner:
I proposed a different approach: PR #15228 adds a `__del__()` method.

csabella:
@vstinner, should this be closed as the other PR has been merged for this issue?  Thanks.

jussike:
Is someone ( @akruis ?) working with this issue? It's needed because PR #15228 (by @vstinner)  has been reverted.

Modules/_threadmodule.c
```diff
@@ -1242,13 +1276,28 @@ thread__set_sentinel(PyObject *self, PyObject *Py_UNUSED(ignored))
...
+        Py_DECREF(wr);
+        return NULL;
+    }
+    data[0] = wr;
+    Py_INCREF(shutdown_locks_lock);
+    data[1] = shutdown_locks_lock;
+    Py_INCREF(shutdown_locks);
+    data[2] = shutdown_locks;
+
+    tstate->on_delete_data = (void *) data;
```
vadmium, line 1294: _On_delete_data_ is no longer just a weak reference to a lock, so the comment in Include/cpython/pystate.h is out of date.

pitrou:
PR #26103 is an arguably simpler fix for this (it doesn't complicate the C side).

pitrou:
PR #26103 is merged, closing this.

PR #15177, opened 2019-08-08
bpo-37779 : Add information about the overriding behavior of ConfigParser.read

sblondon:
This PR adds documentation about the ability to override values when `ConfigParser.read()`, `.read_string()`, `.read_dict()` are called several times from the same instance of `ConfigParser`.

https://bugs.python.org/issue37779

### DIFF
--- a/Doc/library/configparser.rst
+++ b/Doc/library/configparser.rst
@@ -136,6 +136,26 @@ sections [1]_.  Note also that keys in sections are
 case-insensitive and stored in lowercase [1]_.
 
 
+If we need to read several configurations, each one having more priority than the previous one, we can use the same :class:`ConfigParser` instance to override previous defined data and keep not redefined data.
+
+.. doctest::
+
+   >>> config = configparser.ConfigParser()
+   >>> config.read('example.ini')
+   ['example.ini']
+   >>> config['topsecret.server.com']['Port']
+   '50022'
+   >>> config.read_string("[topsecret.server.com]\nPort=48484")
+   >>> config['topsecret.server.com']['Port']
+   '48484'
+   >>> config.read_dict({"topsecret.server.com": {"Port": 21212}})
+   >>> config['topsecret.server.com']['Port']
+   '21212'
+   >>> config['topsecret.server.com']['ForwardX11']
+   'no'
+
+This behaviour is equivalent to a :meth:`ConfigParser.read` call with several files passed to ``filenames`` parameter.
+
 Supported Datatypes
 -------------------
 
--- a/Doc/library/configparser.rst
+++ b/Doc/library/configparser.rst
@@ -142,18 +142,18 @@ configuration while the previously existing keys are retained.
 
 .. doctest::
 
-   >>> config = configparser.ConfigParser()
-   >>> config.read('example.ini')
+   >>> another_config = configparser.ConfigParser()
+   >>> another_config.read('example.ini')
    ['example.ini']
-   >>> config['topsecret.server.com']['Port']
+   >>> another_config['topsecret.server.com']['Port']
    '50022'
-   >>> config.read_string("[topsecret.server.com]\nPort=48484")
-   >>> config['topsecret.server.com']['Port']
+   >>> another_config.read_string("[topsecret.server.com]\nPort=48484")
+   >>> another_config['topsecret.server.com']['Port']
    '48484'
-   >>> config.read_dict({"topsecret.server.com": {"Port": 21212}})
-   >>> config['topsecret.server.com']['Port']
+   >>> another_config.read_dict({"topsecret.server.com": {"Port": 21212}})
+   >>> another_config['topsecret.server.com']['Port']
    '21212'
-   >>> config['topsecret.server.com']['ForwardX11']
+   >>> another_config['topsecret.server.com']['ForwardX11']
    'no'
 
 This behaviour is equivalent to a :meth:`ConfigParser.read` call with several
--- a/Doc/library/configparser.rst
+++ b/Doc/library/configparser.rst
@@ -156,7 +156,7 @@ used to override previously defined properties and retain existing ones.
    'no'
 
 This behaviour is equivalent to a :meth:`ConfigParser.read` call with several
-files passed to ``filenames`` parameter.
+files passed to *filenames* parameter.
 
 
 Supported Datatypes
--- a/Doc/library/configparser.rst
+++ b/Doc/library/configparser.rst
@@ -136,7 +136,9 @@ sections [1]_.  Note also that keys in sections are
 case-insensitive and stored in lowercase [1]_.
 
 
-If we need to read several configurations, each one having more priority than the previous one, we can use the same :class:`ConfigParser` instance to override previous defined data and keep not redefined data.
+   If it is necessary to read several configurations, with each one having more 
+   priority than the previous one, an instance of :class:`ConfigParser` can be 
+   used to override previously defined properties and retain existing ones.
 
 .. doctest::
 
--- a/Doc/library/configparser.rst
+++ b/Doc/library/configparser.rst
@@ -135,9 +135,10 @@ involves the ``DEFAULT`` section which provides default values for all other
 sections [1]_.  Note also that keys in sections are
 case-insensitive and stored in lowercase [1]_.
 
-If it is necessary to read several configurations, with each one having more
-priority than the previous one, an instance of :class:`ConfigParser` can be
-used to override previously defined properties and retain existing ones.
+It is possible to read several configurations into a single
+:class:`ConfigParser`, where the most recently added configuration has the
+highest priority. Any conflicting keys are taken from the more recent
+configuration while the previously existing keys are retained.
 
 .. doctest::
 
--- a/Doc/library/configparser.rst
+++ b/Doc/library/configparser.rst
@@ -155,8 +155,9 @@ used to override previously defined properties and retain existing ones.
    >>> config['topsecret.server.com']['ForwardX11']
    'no'
 
+This behaviour is equivalent to a :meth:`ConfigParser.read` call with several
+files passed to ``filenames`` parameter.
 
-This behaviour is equivalent to a :meth:`ConfigParser.read` call with several files passed to ``filenames`` parameter.
 
 Supported Datatypes
 -------------------
--- a/Doc/library/configparser.rst
+++ b/Doc/library/configparser.rst
@@ -135,10 +135,9 @@ involves the ``DEFAULT`` section which provides default values for all other
 sections [1]_.  Note also that keys in sections are
 case-insensitive and stored in lowercase [1]_.
 
-
-   If it is necessary to read several configurations, with each one having more 
-   priority than the previous one, an instance of :class:`ConfigParser` can be 
-   used to override previously defined properties and retain existing ones.
+If it is necessary to read several configurations, with each one having more
+priority than the previous one, an instance of :class:`ConfigParser` can be
+used to override previously defined properties and retain existing ones.
 
 .. doctest::
 
@@ -155,6 +154,7 @@ case-insensitive and stored in lowercase [1]_.
    '21212'
    >>> config['topsecret.server.com']['ForwardX11']
    'no'
+
 
 This behaviour is equivalent to a :meth:`ConfigParser.read` call with several files passed to ``filenames`` parameter.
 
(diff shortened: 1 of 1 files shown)


sblondon:
I can add an entry into `NEWS` but I wonder if it's really useful.
What do you think about it?

Doc/library/configparser.rst
```diff
@@ -136,6 +136,26 @@ sections [1]_.  Note also that keys in sections are
 case-insensitive and stored in lowercase [1]_.
```
nanjekyejoannah, line 137: nit: single line here.

Doc/library/configparser.rst
```diff
@@ -136,6 +136,26 @@ sections [1]_.  Note also that keys in sections are
...
+   >>> config['topsecret.server.com']['Port']
+   '48484'
+   >>> config.read_dict({"topsecret.server.com": {"Port": 21212}})
+   >>> config['topsecret.server.com']['Port']
+   '21212'
+   >>> config['topsecret.server.com']['ForwardX11']
+   'no'
+
+This behaviour is equivalent to a :meth:`ConfigParser.read` call with several files passed to ``filenames`` parameter.
+
```
nanjekyejoannah, line 162: nit: Add one more line before the new section.

Doc/library/configparser.rst
```diff
@@ -136,6 +136,26 @@ sections [1]_.  Note also that keys in sections are
 case-insensitive and stored in lowercase [1]_.
 
 
+If we need to read several configurations, each one having more priority than the previous one, we can use the same :class:`ConfigParser` instance to override previous defined data and keep not redefined data.
```
nanjekyejoannah: Am not a native English speaker but this "keep not redefined data" phrase is confusing at least to me. Not sure how you can improve it.

nanjekyejoannah: I would just say  "use" instead of "we can use".

nanjekyejoannah (commented):
@sblondon, am not a native English speaker I just left some nit formatting comments. Someone else can review the rest of the documentation.

nanjekyejoannah:
cc @csabella @pganssle

Doc/library/configparser.rst
aeros: Thanks for the contribution @sblondon and suggestion @nanjekyejoannah.

A simple grammatical improvement would be to change "keep not redefined data" to "not keep redefined data". However, I would suggest a few other changes to this part as well. This retains the same meaning, but significantly improves the phrasing:
```suggestion
If it is necessary to read several configurations, with each one having more 
priority than the previous one, an instance of :class:`ConfigParser` can be 
used to override previously defined properties and retain existing ones.
```
reST and Sphinx specifications (this applies to the other sections as well):
1) The maximum number of characters per line is 80 characters. 

Significant phrasing changes:
1) In technical writing, it is preferable to avoid using second person pronouns, such as "we" when possible.
2) "data" -> "properties". This term is more of an accurate descriptor than "data". An alternative term would be "attributes".
3) "keep not redefined data" -> "retain existing ones". This removes redundancy, "ones" refers to the previously mentioned properties, which implies "retain existing properties". 

Edit: Removed part about 3 spaces, see my comment below.

sblondon: I reused aeros167's version so do you think this change is needed?
The new version is `...an instance of :class:ConfigParser can be
used to override...`.

Doc/library/configparser.rst
sblondon: This line is not modified by this PR. So I wonder if it's a good place to fix it.

What do you think about it?

sblondon (commented):
@aeros167 Thanks for the rewrited paragraph, I merged it. I removed the 3 spaces before each line because the paragraph should not be indented.

The whole 'Quick Start' section use 'we', I wonder if the section should be rewritten to remove them. If it's the case, perhaps another bug report+PR is a better choice? Or reusing this one?

Doc/library/configparser.rst
sblondon: Fixed

Doc/library/configparser.rst
nanjekyejoannah: What I know is there should be a single line before the next paragraph in a section. not two.

Doc/library/configparser.rst
nanjekyejoannah: No problem.

Doc/library/configparser.rst
sblondon: I think it's ok now.

aeros167:
>The whole 'Quick Start' section use 'we', I wonder if the section should be rewritten to remove them. If it's the case, perhaps another bug report+PR is a better choice?

Yes, it should probably be in a separate PR, since it''s involving a different section. Preferably, it would be done multiple PRs (single issue), instead of all at once. Smaller PRs tend to be significantly easier to review.

However, I think this issue should be marked as "newcomer friendly" (reserved mostly for first time contributors), I think it would serve as a great introduction into contributing to Python. We'll have to see what the core devs think.

aeros (changes requested):
@sblondon Thanks for making the changes to the phrasing.

It looks like Travis CI is failing due to the doctests starting from line 240 to 248, specifically these lines (244 and 245):
```   
>>> topsecret.getboolean('BatchMode', fallback=True)
True
```
The doctests determined that the output should be `False` instead of `True` after running through the same process in the interpreter. I'm not certain if this is the correct behavior, but either way, the differing output is causing the CI to fail from the warnings.

Since this section is unrelated to this PR and in an unaltered section, I think it should be ignored for this PR and addressed in a new issue. 

One more minor suggestion, the conventional format for defining parameters is to italicize them, rather than using code blocks:

Edit: Removed 3 spaces before the line part, see my comment below.

Doc/library/configparser.rst
```diff
@@ -135,6 +135,29 @@ involves the ``DEFAULT`` section which provides default values for all other
...
+   >>> config['topsecret.server.com']['Port']
+   '48484'
+   >>> config.read_dict({"topsecret.server.com": {"Port": 21212}})
+   >>> config['topsecret.server.com']['Port']
+   '21212'
+   >>> config['topsecret.server.com']['ForwardX11']
+   'no'
+
+This behaviour is equivalent to a :meth:`ConfigParser.read` call with several
+files passed to ``filenames`` parameter.
```
aeros:
```suggestion
files passed to *filenames* parameter.
```

sblondon:
> @sblondon Thanks for making the changes to the phrasing.
> 
> > I removed the 3 spaces before each line because the paragraph should not be indented.
> 
> The three space indentation before each line is a requirement for reST, so any .rst file must do so. See the [devguide](https://devguide.python.org/documenting/?highlight=All%20reST%20files%20use%20an%20indentation%20of%203%20spaces#use-of-whitespace) for more information.

Thanks for the link @aeros167 . However, I don't understand the sentence `All reST files use an indentation of 3 spaces; no tabs are allowed.` like you.
To me, it means 'if indentation is needed, use 3 spaces', not 'every line must be indented by 3 spaces'.

I think that because every .rst file I watch does not have 3 spaces indentations for the main text. Examples in cpython documentation:
 - https://raw.githubusercontent.com/python/cpython/master/Doc/about.rst
 - https://raw.githubusercontent.com/python/cpython/master/Doc/library/math.rst

Example in the devguide: 
 - https://raw.githubusercontent.com/python/devguide/master/documenting.rst

What do you think about it?

aeros167:
>Thanks for the link @aeros167 . However, I don't understand the sentence All reST files use an indentation of 3 spaces; no tabs are allowed. like you.
To me, it means 'if indentation is needed, use 3 spaces', not 'every line must be indented by 3 spaces'.

Upon further inspection of the area being modified, you're correct. For some reason, I had thought that this area was surrounded by an [reST directive](https://devguide.python.org/documenting/#directives). It looks like there's only the initial `codeblock` and then `doctests` for the interpreter, with no directive surrounding that paragraph I suggested the changes to. 

Thanks again for the PR and responsiveness to the feedback. Apologies if I caused any confusion, I'll edit my comments accordingly.

sblondon:
@aeros167 Thank you for the review ! :)

(By the way, beware of burn out if you try to do too much volonteer work.)

aeros167:
> Thank you for the review ! :)
(By the way, beware of burn out if you try to do too much volonteer work.)

No problem, and yes I definitely need to make sure that I pace myself. This is the first open source project that I've found myself incredibly motivated to help with, since Python was my first serious programming language 5+ years ago. 

It can be tempting some days to just endlessly sink my free time into it, but burnout is definitely a concern. @vstinner mentioned this [in a talk](https://www.youtube.com/watch?v=O3UmUj9h_Eo&t=1173s) earlier this year at PyCon 2019.

Doc/library/configparser.rst
```diff
@@ -135,6 +135,29 @@ involves the ``DEFAULT`` section which provides default values for all other
...
+   >>> config['topsecret.server.com']['Port']
+   '48484'
+   >>> config.read_dict({"topsecret.server.com": {"Port": 21212}})
+   >>> config['topsecret.server.com']['Port']
+   '21212'
+   >>> config['topsecret.server.com']['ForwardX11']
+   'no'
+
+This behaviour is equivalent to a :meth:`ConfigParser.read` call with several
+files passed to *filenames* parameter.
```
pganssle:
```suggestion
files passed to the *filenames* parameter.
```

Doc/library/configparser.rst
```diff
@@ -135,6 +135,29 @@ involves the ``DEFAULT`` section which provides default values for all other
 sections [1]_.  Note also that keys in sections are
 case-insensitive and stored in lowercase [1]_.
 
+If it is necessary to read several configurations, with each one having more
+priority than the previous one, an instance of :class:`ConfigParser` can be
+used to override previously defined properties and retain existing ones.
```
pganssle: One problem I see with this is that the first few clauses of this sentence seem to be introducing a tutorial, i.e. "If you want to do X, you can achieve it with Y", but then it turns out that you are actually *describing the existing behavior* of `ConfigParser` rather than providing a recipe or something.

I would recommend rewriting this as an introduction to the properties of a `ConfigParser`, like so:

> It is also possible to read several configurations into a single ```:class:`ConfigParser` ``` object, in which case the most recent configuration has priority over the previous ones; conflicting keys will be taken from the new configuration and existing keys will be retained.

That's mostly an example, I'm not super happy with that particular wording...

pganssle (changes requested):
@ambv You are listed as a section author for this section - do you have any thoughts on the advisability of adding this section? If not, do you know who the right person would be?

Doc/library/configparser.rst
aeros: @pganssle:
> That's mostly an example, I'm not super happy with that particular wording...

What do you think of this wording?

> It is possible to read several configurations into a single :class:`ConfigParser`, where the most recently added configuration has the highest priority. Any conflicting keys are taken from the more recent configuration while the previously existing keys are retained.

pganssle: :shipit: That looks great!

sblondon: @pganssle I pushed @aeros167 's version to the repository.

sblondon: I have made the requested changes; please review again

sblondon:
I don't think a `news` entry is necessary so it's good to me.

aeros (approved):

ambv:
Closing and re-opening to re-trigger CI checks which were spuriously failing before.

ambv:
Looks like the failure is real: the documentation doctest fails with:
```
File "library/configparser.rst", line 246, in default
Failed example:
    topsecret.getboolean('BatchMode', fallback=True)
Expected:
    False
Got:
    True

...

 Makefile:127: recipe for target 'doctest' failed
make: *** [doctest] Error 1
make: Leaving directory '/home/runner/work/cpython/cpython/Doc'
```

@sblondon, please confirm you can successfully build the documentation locally by running this in the main directory of your `cpython` checkout:
```
make -C Doc/ PYTHON=../python doctest suspicious html
```

sblondon:
@ambv I renamed the `config` object created by the patch so now the first instance is not overwritten and the failing test uses it like when the patch wasn't there. Is this workaround enough?

I checked the fix on my computer with:
```python
>>> import doctest
>>> doctest.testfile("Doc/library/configparser.rst")
TestResults(failed=0, attempted=88)
```

sblondon:
Thanks everyone!

PR #15178, opened 2019-08-08
bpo-37793:  Replace wording "container object" with "iterable" in argparse

shireenrao:
https://bugs.python.org/issue37793


nanjekyejoannah:
I think consesus hasnt been reached yet on whether to just use the phrase iterable. It looks like @ericvsmith commented on bpo that he is still researching too.

ericvsmith:
I made an error when looking through the code, and "iterable" is not correct. So I'm going to close this. Sorry, @shireenrao for wasting your time on this. I promise there will be other opportunities.

PR #15179, opened 2019-08-08
bpo-37794: MSVCCompiler: Replace /Ox with /O2

Kojoley:
The /O2 is a superset of /Ox with additional /GF and /Gy switches which
enables strings and functions deduplication and almost always are
favorable optimizations without downsides.

https://docs.microsoft.com/en-us/cpp/build/reference/ox-full-optimization

https://bugs.python.org/issue37794

### DIFF
--- a/Misc/NEWS.d/next/Library/2019-08-08-20-04-58.bpo-37794.mTM2Kt.rst
+++ b/Misc/NEWS.d/next/Library/2019-08-08-20-04-58.bpo-37794.mTM2Kt.rst
@@ -1,0 +1,1 @@
+:class:`distutils.MSVCCompiler` now uses ``/O2`` switch instead of ``/Ox``.
--- a/Lib/distutils/_msvccompiler.py
+++ b/Lib/distutils/_msvccompiler.py
@@ -278,7 +278,7 @@ class MSVCCompiler(CCompiler) :
         # use /MT[d] to build statically, then switch from libucrt[d].lib to ucrt[d].lib
         # later to dynamically link to ucrtbase but not vcruntime.
         self.compile_options = [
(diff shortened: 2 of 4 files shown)


Misc/NEWS.d/next/Library/2019-08-08-20-04-58.bpo-37794.mTM2Kt.rst
```diff
@@ -0,0 +1 @@
+:class:`distutils.MSVCCompiler` now uses ``/O2`` switch instead of ``/Ox``.
```
epicfaace, line 1:
```suggestion
:class:`distutils.MSVCCompiler` now uses the ``/O2`` optimization instead of ``/Ox``.
```

epicfaace (commented):
Sounds good.

Do we also need to change this line? https://github.com/python/cpython/blob/5affd23e6f42125998724787025080a24839266e/Tools/freeze/winmakemakefile.py#L68

Kojoley:
> Sounds good.
> 
> Do we also need to change this line?
> 
> https://github.com/python/cpython/blob/5affd23e6f42125998724787025080a24839266e/Tools/freeze/winmakemakefile.py#L68

I know nothing about that code. It does not belong to `distutils`, so it is out of scope of my intentions too. If the only code it affects is a wrapper code that the file contains - it is probably not worth bothering.

EvgenKo423:
It's been a year, any progress on this? @Kojoley, you've got a feedback on bug tracker. Could you please finish with this?

Kojoley:
No, I am sorry, there is too much time consuming bureaucracy for a simple thing.

iritkatriel:
Closing this PR following the closure of its b.p.o. issue (https://bugs.python.org/issue37794)

PR #15181, opened 2019-08-08
bpo-36511 fix failures in Windows Python arm32 buildbot tests

paulmon:
@zooba, @zware 

win_utils.py:  If an exception is thrown during WindowsLoadTracker creation, which is expected on Windows ARM32, then when WindowsLoadTracker.close( ) is called then self.p is not defined.  This change ensures that self.p is always initialized.

test_compile.py: test_stack_overflow throws a MemoryException randomly on Windows arm32.  I believe this test is running out of memory because the arm32 test hardware only has 1 GB of RAM.  What is the best way to fix this?  Is handling the MemoryException ok? or does that cover up a real problem?

remoteDeploy.bat:  the files in the Parser directory are needed to prevent warnings when running test_asdl_parser.

Please let me know if I should split this into multiple PRs or create new issue(s).

Thanks,
Paul

### DIFF
--- a/Lib/test/libregrtest/win_utils.py
+++ b/Lib/test/libregrtest/win_utils.py
@@ -25,6 +25,7 @@ class WindowsLoadTracker():
 
     def __init__(self):
         self.load = 0.0
+        self.p = None
         self.start()
 
     def start(self):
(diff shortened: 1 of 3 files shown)


zooba:
The MemoryError might be hiding something else, unless it's because of the string generation. In which case, I'd separate that out from the `compile` call.

paulmon:
The string generation didn't cause a MemoryError.
Undoing suppressing the exception.

PR #15183, opened 2019-08-08
bpo-37268: test_parser fails when run with -Werror

ZackerySpytz:
Use warnings.filterwarnings() when importing the deprecated parser
module.

@pablogsal 

https://bugs.python.org/issue37268


pablogsal (approved):
LGTM

Thanks @ZackerySpytz for fixing this!

PR #15192, opened 2019-08-09
bpo-37802: Slightly improve perfomance of PyLong_FromSize_t()

sir-sigurd:
https://bugs.python.org/issue37802

### DIFF
--- a/Misc/NEWS.d/next/Core and Builtins/2019-08-09-18-28-57.bpo-37802.pKxcAW.rst
+++ b/Misc/NEWS.d/next/Core and Builtins/2019-08-09-18-28-57.bpo-37802.pKxcAW.rst
@@ -1,0 +1,3 @@
+Slightly improve performance of :c:func:`PyLong_FromUnsignedLong`,
+:c:func:`PyLong_FromUnsignedLongLong` and :c:func:`PyLong_FromSize_t`.
+Patch by Sergey Fedoseev.
--- a/Objects/longobject.c
+++ b/Objects/longobject.c
@@ -43,6 +43,7 @@ PyObject *_PyLong_One = NULL;
 static PyLongObject small_ints[NSMALLNEGINTS + NSMALLPOSINTS];
 
 #define IS_SMALL_INT(ival) (-NSMALLNEGINTS <= (ival) && (ival) < NSMALLPOSINTS)
+#define IS_SMALL_UINT(ival) ((ival) < NSMALLPOSINTS)
 
 #ifdef COUNT_ALLOCS
 Py_ssize_t _Py_quick_int_allocs, _Py_quick_neg_int_allocs;
@@ -78,6 +79,7 @@ maybe_small_long(PyLongObject *v)
 }
 #else
 #define IS_SMALL_INT(ival) 0
+#define IS_SMALL_UINT(ival) 0
 #define get_small_int(ival) (Py_UNREACHABLE(), NULL)
 #define maybe_small_long(val) (val)
 #endif
@@ -378,32 +380,52 @@ PyLong_FromLong(long ival)
     return (PyObject *)v;
 }
 
+#define PYLONG_FROM_UINT(INT_TYPE, ival) \
+    do { \
+        if (IS_SMALL_UINT(ival)) { \
+            return get_small_int((ival)); \
+        } \
+        /* Count the number of Python digits. */ \
+        Py_ssize_t ndigits = 0; \
+        INT_TYPE t = (ival); \
+        while (t) { \
+            ++ndigits; \
+            t >>= PyLong_SHIFT; \
+        } \
+        PyLongObject *v = _PyLong_New(ndigits); \
+        if (v == NULL) { \
+            return NULL; \
+        } \
+        digit *p = v->ob_digit; \
+        while ((ival)) { \
+            *p++ = (digit)((ival) & PyLong_MASK); \
+            (ival) >>= PyLong_SHIFT; \
+        } \
+        return (PyObject *)v; \
+    } while(0)
+
 /* Create a new int object from a C unsigned long int */
 
 PyObject *
 PyLong_FromUnsignedLong(unsigned long ival)
 {
-    PyLongObject *v;
-    unsigned long t;
-    int ndigits = 0;
+    PYLONG_FROM_UINT(unsigned long, ival);
+}
 
-    if (ival < PyLong_BASE)
-        return PyLong_FromLong(ival);
-    /* Count the number of Python digits. */
-    t = ival;
-    while (t) {
-        ++ndigits;
-        t >>= PyLong_SHIFT;
-    }
-    v = _PyLong_New(ndigits);
-    if (v != NULL) {
-        digit *p = v->ob_digit;
-        while (ival) {
-            *p++ = (digit)(ival & PyLong_MASK);
-            ival >>= PyLong_SHIFT;
-        }
-    }
-    return (PyObject *)v;
+/* Create a new int object from a C unsigned long long int. */
+
+PyObject *
+PyLong_FromUnsignedLongLong(unsigned long long ival)
+{
+    PYLONG_FROM_UINT(unsigned long long, ival);
+}
+
+/* Create a new int object from a C size_t. */
+
+PyObject *
+PyLong_FromSize_t(size_t ival)
+{
+    PYLONG_FROM_UINT(size_t, ival);
 }
 
 /* Create a new int object from a C double */
@@ -1186,34 +1208,6 @@ PyLong_FromLongLong(long long ival)
     return (PyObject *)v;
 }
 
-/* Create a new int object from a C unsigned long long int. */
-
-PyObject *
-PyLong_FromUnsignedLongLong(unsigned long long ival)
-{
-    PyLongObject *v;
-    unsigned long long t;
-    int ndigits = 0;
-
-    if (ival < PyLong_BASE)
-        return PyLong_FromLong((long)ival);
-    /* Count the number of Python digits. */
-    t = ival;
-    while (t) {
-        ++ndigits;
-        t >>= PyLong_SHIFT;
-    }
-    v = _PyLong_New(ndigits);
-    if (v != NULL) {
-        digit *p = v->ob_digit;
-        while (ival) {
-            *p++ = (digit)(ival & PyLong_MASK);
-            ival >>= PyLong_SHIFT;
-        }
-    }
-    return (PyObject *)v;
-}
-
 /* Create a new int object from a C Py_ssize_t. */
 
 PyObject *
@@ -1252,35 +1246,6 @@ PyLong_FromSsize_t(Py_ssize_t ival)
         while (t) {
             *p++ = (digit)(t & PyLong_MASK);
             t >>= PyLong_SHIFT;
-        }
-    }
-    return (PyObject *)v;
-}
-
-/* Create a new int object from a C size_t. */
-
-PyObject *
-PyLong_FromSize_t(size_t ival)
-{
-    PyLongObject *v;
-    size_t t;
-    int ndigits = 0;
-
-    if (ival < PyLong_BASE)
-        return PyLong_FromLong((long)ival);
-    /* Count the number of Python digits. */
-    t = ival;
-    while (t) {
-        ++ndigits;
-        t >>= PyLong_SHIFT;
-    }
-    v = _PyLong_New(ndigits);
-    if (v != NULL) {
-        digit *p = v->ob_digit;
-        Py_SIZE(v) = ndigits;
-        while (ival) {
-            *p++ = (digit)(ival & PyLong_MASK);
-            ival >>= PyLong_SHIFT;
         }
     }
     return (PyObject *)v;


Objects/longobject.c
```diff
@@ -64,6 +64,10 @@ get_small_int(sdigit ival)
     do if (-NSMALLNEGINTS <= ival && ival < NSMALLPOSINTS) { \
         return get_small_int((sdigit)ival); \
     } while(0)
+#define CHECK_SMALL_UINT(ival) \
+    do if (ival < NSMALLPOSINTS) { \
```
mdickinson: Can you fix the comparison of signed with unsigned here? It may also be good to add a test for the case where `ival == (size_t)(-1)` (for example), just to make sure that the code does the right thing.

Misc/NEWS.d/next/Core and Builtins/2019-08-09-18-28-57.bpo-37802.pKxcAW.rst
```diff
@@ -0,0 +1,2 @@
+Slightly improve perfomance of PyLong_FromSize_t(). Patch by Sergey
```
aeros: Thanks for the PR.

I would recommend adding inline markup to "PyLong_FromSize_t". After Sphinx processes the news entry, it allows users to click on the function to be directed to the relevant documentation.
```suggestion
Slightly improve perfomance of :c:func:`PyLong_FromSize_t`. Patch by Sergey
```

Objects/longobject.c
sir-sigurd: > Can you fix the comparison of signed with unsigned here?

I'm not sure what should be fixed here, could you elaborate on that?

aeros: I'm not overly experienced with the c-api, so take my interpretation with a grain of salt.

In this case, `ival` is unsigned and `NSMALLPOSINTS` is signed. @mdickinson is suggesting that the two should not be directly compared. I think this offset could be compensated with:

`ival - NSMALLNEGINTS`

Correct me if I'm mistaken.

Edit: `ival` -> unsigned, `NSMALLPOSINTS` -> signed

sir-sigurd: `ival` is unsigned and `NSMALLPOSINTS` is signed, but as far as I understand this doesn't produce any problems since `NSMALLPOSINTS` is non-negative.

aeros: @sir-sigurd You're right, it should be the other way around. Not sure why I initally though `ival` was signed. In that case, the offset could be compensated by subtracting `NSMALLNEGINTS` rather than adding.

mdickinson: Indeed `NSMALLPOSINTS` is signed; `ival` is unsigned. My recollection is that some compilers will complain about comparisons between unsigned and signed types - in this case it's fine, but the C rules can give surprising results in other cases, so one often needs a cast to suppress a compiler warning.

But I may be misremembering: maybe compilers are smart enough to figure out that this isn't a case that needs a warning, given that `NSMALLPOSINTS` is a known-at-compile-time positive constant.

mdickinson (approved):
LGTM

Objects/longobject.c
gnprice: I think compilers are smart enough not to complain about this, at least when the signed-but-positive value is a literal. Which is good, because there'd be a lot of noise if they did!

A super handy tool for experimenting with this kind of question is Compiler Explorer. Here's a test program:
https://godbolt.org/z/ZYt9pI

I've enabled every possible warning on clang, the relevant warning on GCC, and (I think) every possible warning on MSVC.

None of them complain when it's a literal in the right range, which is our case here.

For entertainment and wider context:
* All complain when it's an unknown value.
* All complain when it's 0 (so the comparison is always false).
* When it's a `const` value (rather than a literal) that is in the right range, clang accepts it but GCC and MSVC warn. You might change the constant's type, or add a cast.
* When it's a literal that is too big so the comparison is always true, clang and GCC warn but MSVC doesn't.

gnprice:
BTW, I think the behavior of `CHECK_SMALL_INT` with its implicit `return` (and similarly the new `CHECK_SMALL_UINT`) is pretty surprising when reading the code. I wrote up a patch the other day that replaces it with an explicit `return`; I'll take this as a prompt to go make an issue for it and send a PR.

That's certainly not a reason not to merge this, though. I'll happily rebase my changes on top of this once it's in.

Objects/longobject.c
aeros: Thanks for the detailed explanations, @mdickinson and @gnprice! I've recently been trying to help out with more PR reviews (since that seems to a bottleneck for CPython), but my experience is primarily with Python and writing documentation, not with C. I have some idea of the fundamentals, but I'm still learning.

I was misunderstanding the suggestion initially, and thought that the issue was that since `ival` is unsigned, it needed to be adjusted in value in order to accurately compare it against the constant `NSMALLPOSINTS`. So effectively, we would be converting `ival` into a signed integer just for the comparison. I suppose the correct way to do this would be adding a cast (if it were necessary, but it looks like it isn't in this case since `ival` is a literal).

>When it's a const value (rather than a literal) that is in the right range, clang accepts it but GCC and MSVC warn. You might change the constant's type, or add a cast.

So if it were instead an unsigned constant we were comparing with `NSMALLPOSINTS`, let's call it `UCONST` for the sake of the example, would we have to add a cast to `UCONST`?

```c
do if ((int)UCONST < NSMALLPOSINTS) { \
```

aeros: Based on the above conclusions, would it be helpful to at least add a comment, mentioning why the cast is not needed in this case?

```suggestion
    do if (ival < NSMALLPOSINTS) { /* since ival is a literal, a cast is not needed */ \ 
```

gnprice: No, I think best not to. The comment would be more work for the reader than simply adding the cast would :wink: 

The code does exactly the simple thing that it looks like it does. If it did something tricky, or if it might look like it did one thing but it actually did another, that would be a good occasion for a comment.

aeros: @gnprice Ah okay, fair enough. I mostly just thought of adding it because there seemed to be some degree of nuance involved here. Also, it may help readers understand why there isn't a cast here. But that's a good point, it might be adding to much additional work for the reader.

gnprice:
> BTW, I think the behavior of `CHECK_SMALL_INT` with its implicit `return` (and similarly the new `CHECK_SMALL_UINT`) is pretty surprising when reading the code. I wrote up a patch the other day that replaces it with an explicit `return`; I'll take this as a prompt to go make an issue for it and send a PR.
> 
> That's certainly not a reason not to merge this, though. I'll happily rebase my changes on top of this once it's in.

(Posted; see https://bugs.python.org/issue37812 and #15203 .)

Objects/longobject.c
gnprice: > Thanks for the detailed explanations, @mdickinson and @gnprice! I've recently been trying to help out with more PR reviews (since that seems to a bottleneck for CPython), but my experience is primarily with Python and writing documentation, not with C. I have some idea of the fundamentals, but I'm still learning.

Great! I agree that seems to be an important bottleneck.


> So if it were instead an unsigned constant we were comparing with `NSMALLPOSINTS`, let's call it `UCONST` for the sake of the example, would we have to add a cast to `UCONST`?
> 
> ```c
> do if ((int)UCONST < NSMALLPOSINTS) { \
> ```

Pretty much! Just to make the setup explicit -- the case I'm referring to is that there's something like
```
static const unsigned int UCONST = 42;
```
somewhere, perhaps at global scope. Which also means it's not a macro, so we'd typically spell it lowercase:
```
static const unsigned int some_value = 42;
```

If you follow the https://godbolt.org/z/ZYt9pI link above, the example I'm referring to there looks like
```
static const int32_t maxval = 257;

static int
check(uint32_t ival)
{
    return /* ... */ (ival < maxval) /* ... */;
}
```

Then a reasonable response to the warning could be to add that cast, just as you showed, which in this example would make
```
    return /* ... */ (ival < (uint32_t)maxval) /* ... */;
```

Or it might instead be to change the declared type of that const variable:
```
static const uint32_t maxval = 257;
```
Which one to do would depend on which one we felt better matched the way we wanted to think about this constant, and how else it was used.

The third possible response to the warning would be... to notice that there was a bug, and fix the code! That possibility is why the warning exists, after all. :wink: In this example, perhaps that would look like noticing that `ival` really should be signed, and that in fact it comes from someplace where we'd inappropriately cast it from unsigned to signed -- so that if the original unsigned value is large, we'd treat it as negative when we really did want to treat it as large.

aeros: @gnprice Thank you very much for the detailed explanation! That compiler explorer tool looks incredibly useful for exploring errors (and c in general). The visual highlighting where it shows the c line and it's x86 equivalent is also pretty nice.

Objects/longobject.c
```diff
@@ -1270,8 +1276,9 @@ PyLong_FromSize_t(size_t ival)
     size_t t;
     int ndigits = 0;
 
-    if (ival < PyLong_BASE)
-        return PyLong_FromLong((long)ival);
```
gnprice, line 1270: How about making this just call `PyLong_FromUnsignedLong` instead?

... Ah, it looks like `PyLong_FromUnsignedLong` does the same thing. Well: how about applying this optimization to `PyLong_FromUnsignedLong` instead of `PyLong_FromSize_t`, and making this just call `PyLong_FromUnsignedLong` instead? :wink: 

The usual practice across the codebase is to use either `PyLong_FromLong` or `PyLong_FromUnsignedLong` whenever they'll do, so there are a lot more calls to `PyLong_FromUnsignedLong` than to `PyLong_FromSize_t`. Which means the optimization will apply much more broadly if it's present there.

Objects/longobject.c
```diff
@@ -47,6 +47,11 @@ is_small_int(long long ival)
 {
     return -NSMALLNEGINTS <= ival && ival < NSMALLPOSINTS;
 }
+static inline int
+is_small_uint(uintmax_t ival)
```
gnprice: Ah, good thought re `uintmax_t`!

It'd be a nice small fix to have `is_small_int` take an `intmax_t`, too. I had forgotten that type existed when I wrote it.

gnprice (commented):
This looks good to me... except it's not ambitious enough! :slightly_smiling_face: Details below.

(Happily the more ambitious version is only a line or two more.)

Objects/longobject.c
aeros: > Which means the optimization will apply much more broadly if it's present there.

I'm curious as to how much of an optimization this would lead to for the performance of ``PyLong_FromUnsignedLong()``. There are approximately 81 calls to the function with the C-API, (according to ``git grep -E "PyLong_FromUnsignedLong\(.+\)" -- "*.c" | wc -l``) so even a minor optimization could make a significant difference. 

Also, if anyone was curious, there are approximately 16 calls to ``PyLong_FromSize_t()``.

sir-sigurd: Probably I'll refactor PyLong_From*() functions like in #15457 later, but I don't want to make any other changes in this PR because it blocks me from further work, specifically from replacing `PyLong_FromSsize_t()` with `PyLong_FromSsize_t()` in the places where it makes sense.

sir-sigurd:
@mdickinson anything should be done to get this merged?

Objects/longobject.c
gnprice: Well, a version of this PR that applies to the ...UnsignedLong function is literally one line bigger than this version is. (Apply the same diff to that function instead of this one; and in this one, just s/Long/UnsignedLong/.) And it makes the change more logically coherent. So I'm not sure that making that revision will make it any slower to get the PR merged; it could make it faster.

That other PR #15457 looks much more complex.

> replacing PyLong_FromSsize_t() with PyLong_FromSsize_t()

I think you meant something else here. ;-) I'm curious exactly what.

sir-sigurd: I meant replacing `PyLong_FromSsize_t()` with `PyLong_FromSize_t()`.

sir-sigurd: > Apply the same diff to that function instead of this one; and in this one, just s/Long/UnsignedLong/

`if (ival < PyLong_BASE)` makes `PyLong_FromSsize_t()` slower than it is in this PR.

gnprice: > `if (ival < PyLong_BASE)` makes `PyLong_FromSsize_t()` slower than it is in this PR.

Hmm, I didn't mean to suggest doing anything at all to `PyLong_FromSsize_t`; so I think there must have been a misunderstanding. I'll try again, with more explicitness. (I wrote my previous message from a phone, which made typing harder. Back at a nice keyboard now.)

The current version of the PR has this diff:
```
-    if (ival < PyLong_BASE)
-        return PyLong_FromLong((long)ival);
+    if (is_small_uint(ival)) {
+        return get_small_int(ival);
+    }
```
inside the implementation of `PyLong_FromSize_t`.

I think it would be better to (a) apply essentially the exact same diff:
```
-    if (ival < PyLong_BASE)
-        return PyLong_FromLong(ival);
+    if (is_small_uint(ival)) {
+        return get_small_int(ival);
+    }
```
but inside the implementation of `PyLong_UnsignedLong`. Then (b) inside the implementation of this function `PyLong_FromSize_t`, apply instead this diff:
```
     if (ival < PyLong_BASE)
-        return PyLong_FromLong((long)ival);
+        return PyLong_FromUnsignedLong((unsigned long)ival);
```

This will make the optimization apply more broadly. It will make the logic of the change better fit with how we generally use the `PyLong_From*` functions.

And the cost is one additional line changed in the diff, and zero additional complexity in the resulting code.

sir-sigurd: > Hmm, I didn't mean to suggest doing anything at all to PyLong_FromSsize_t

I mixed up them again, actually I meant `PyLong_FromSize_t()`.

gnprice: > I mixed up them again, actually I meant `PyLong_FromSize_t()`.

Got it.

I suggested pointing at `PyLong_FromUnsignedLong` because your original explanation (in the issue thread) for why this was an optimization was that `PyLong_FromLong` needs to handle the sign, and `PyLong_FromUnsignedLong` would suffice to avoid that.

Here's a variation: take that same 2-line diff that replaces `if (ival < PyLong_BASE)` and `PyLong_FromLong` with `is_small_int` and `get_small_int`... and apply it to *both* of `PyLong_FromUnsignedLong` and `PyLong_FromSize_t`.

That will again make the optimization apply more broadly; and it'll keep these two functions closely parallel, as one would naturally expect them to be.

The cost this time is two additional lines changed in the diff (three, counting the close-brace). And it actually reduces complexity in the resulting code, because it doesn't leave readers having to wonder why the constructors taking `unsigned long` on the one hand and `size_t` on the other -- which are even the same type on common platforms -- have different logic.

sir-sigurd:
I decided to deduplicate `PyLong_FromUnsigned*` functions in this PR, to apply this optimization to all of them at once.

Objects/longobject.c
```diff
@@ -410,6 +414,30 @@ PyLong_FromUnsignedLong(unsigned long ival)
     return (PyObject *)v;
 }
 
+/* Same as above, but for unsigned long. */
```
gnprice: Nit: better to avoid "Same as above". Partly because the "above" is already a few dozen lines away and so potentially a bit unclear... and more so because this kind of reference from a distance is highly vulnerable to getting broken by future changes to the code :slightly_smiling_face: .

If the other thing's description were long, a good solution would be to refer to it by name: "Like long_from_uint, but ...".

In this case the other description is only a line long in itself, so it's simplest just to copy the common part: "Create a new int object from a C unsigned long".

(Maybe even skip the word "C"? I think that really makes sense only for `PyLong_FromLong`, where there's a Python type that (used to, and in some ways still) has the same name as the C type.)

Objects/longobject.c
```diff
@@ -47,6 +47,11 @@ is_small_int(long long ival)
 {
     return -NSMALLNEGINTS <= ival && ival < NSMALLPOSINTS;
 }
+static inline int
```
gnprice: Nit: put a blank line between functions.

gnprice (approved):
Neat! This simplifies the code *and* it's faster (based on the measurements you posted in the issue thread) -- that's a nice combination. :smile: 

I'd be glad to see the signed versions get a similar treatment. (As a separate PR.) Might not deliver a performance win at the same level, since your hypothesis is that the speedup here is from cutting out the sign logic...

but it should be performance-neutral at worst, and it's plausible it could be an improvement by generating less code so it's friendlier to the cache. Also could speed things up just as a natural consequence of deduplicating the code, by making a single place to apply all the optimizations we have in one function or another, plus any new ones you think of while staring at it.

(And even if it does turn out only performance-neutral: I think opinions will vary, but for code deduplication at this scale I'd certainly be in favor.)

Objects/longobject.c
sir-sigurd: Might make sense to simply remove these comments, because it's quite obvious that they do the same thing as `long_from_uint()`.

gnprice: That's true, but then you have to go look up what `long_from_uint` does. :-)

These functions are part of the API exposed outside this module, so in general they should have a doc-comment to explain what their intended meaning is. In this case the meaning is pretty simple, and *almost* self-explanatory from the name... but that just means that a very short doc-comment is enough. Which leaves little to be gained from skipping it.

sir-sigurd: Non-existing comments never lie and don't require updates :slightly_smiling_face: .

gnprice:
Wise words!

Yet I think for doc-comments the risk is worth it. If the comment simply
describes the interface of the function, i.e. the things a caller needs to
know about it... then if that ever needs an update, there will be much
bigger burdens than updating the comment. Especially for functions in the
public API like these.

But that is why I would like the comments to describe the interface
directly (like they do in master), rather than have a dependency on
something more fluid like the arrangement of this code. :-)

sir-sigurd:
The current approach using inline function have the same problems as here (https://bugs.python.org/issue38015). I'll rework it when #15718 will be merged.

gnprice:
> The current approach using inline function have the same problems as here (https://bugs.python.org/issue38015). I'll rework it when #15718 will be merged.

Hmm, you measured this as being a significant speedup in your microbenchmark, right? That hardly seems like a problem :wink: even if there's a way to make the microbenchmark go even faster still.

I like this PR as it is :slightly_smiling_face: and I'd be glad to see it merged. It actually simplifies the code at the same time as optimizing it, so there's no need to weigh one value against another.

As I commented over on #15718, turning `get_small_int` into a macro makes the code inside it quite a bit more complex to read. For this one... is the idea that you would write `long_from_uint` as a macro? That's a 20-line function, so that sounds probably worse :-/. I think that'd be much too big a cost to pay in the code for the payoff of this microbenchmark speedup.

If that is your plan, then it's also basically additive upon the change in this PR. So I think it'd still be good to merge this PR, even if a subsequent PR goes on to turn the function into a macro.

sir-sigurd:
> Hmm, you measured this as being a significant speedup in your microbenchmark, right? That hardly seems like a problem wink even if there's a way to make the microbenchmark go even faster still.

It induces the same problems as your is_small_int() function on 32-bit platforms.

Objects/longobject.c
aeros: @sir-sigurd:
> Non-existing comments never lie and don't require updates :slightly_smiling_face:.

Between the choice of keeping the comment as is or removing it, I would support removing it more. But, I think it would be useful to readers to include @gnprice's suggestion (removing the "C" part as well):

```suggestion
/* Create a new int object from an unsigned long. */
```

I think it's fairly common for developers to look at the implementation for a single function in isolation rather than reading the entire file at once, so it's not always optimal to rely on comments in earlier sections for context. To some degree, each function should aim to explain itself.

gnprice:
> It induces the same problems as your is_small_int() function on 32-bit platforms.

Right but that "problem" is that a few extra instructions are emitted, right? And so the microbenchmark is a few percent slower than it would be.

If the microbenchmark is nevertheless faster than before, then this is still an optimization.

sir-sigurd:
> > It induces the same problems as your is_small_int() function on 32-bit platforms.
> 
> Right but that "problem" is that a few extra instructions are emitted, right? And so the microbenchmark is a few percent slower than it would be.
> 
> If the microbenchmark is nevertheless faster than before, then this is still an optimization.

https://bugs.python.org/msg351052

gnprice:
> I leave it to you to benchmark it on 32-bit platform.

Hmm -- do you mean you haven't tried it on a 32-bit platform? Because you said that it had a problem there, I assumed that meant you'd seen something empirically.

If you have, I think it would be helpful to say what results you got.

sir-sigurd:
I mean I checked how it compiles on godbolt.org and that was sufficient to me to not benchmark it.

gnprice:
> I mean I checked how it compiles on godbolt.org and that was sufficient to me to not benchmark it.

Cool, got it.

Well -- as I said upthread, if this PR were to end up making something as big as `long_from_uint` into a macro, I think that would be much more of a cost than the micro-optimization is worth.

If you can find a way to get all the desired speedups without that kind of complexity in the code, that'd be great.

As I see it, I would also be glad to see a PR version merged that did something like
* a nice simplification to the code (like your current version here)
* ~10% speedup in microbenchmark on x86_64 (like you quoted for this version on the issue thread)
* ~5% slowdown in microbenchmark on x86_32 (like you found for `get_small_ints` as function vs. macro)

because that's good for the source code and no worse than a wash on performance.

(Note also that anyone who's working to squeeze out the last drop of performance in running their code is much more likely to be using a 64-bit platform already.)

aeros167:
@gnprice:
> Note also that anyone who's working to squeeze out the last drop of performance in running their code is much more likely to be using a 64-bit platform already

Agreed, I don't think anyone using a 32-bit platform is going to be overly concerned about a 5% drop in performance. A more significant performance loss may be an issue, but not when it's that small. IMO, an equivalent or larger increase in the performance for 64-bit outweighs an equal or smaller loss for 32-bit.

sir-sigurd:
> If you can find a way to get all the desired speedups without that kind of complexity in the code, that'd be great.

I agree that'd be great to write some generic function, but C doesn't allow to write type generic functions and using inline function here induces these unwanted type casts that makes emitted code less efficient in unpredictable way. Such inline function creates a false feeling that it works like type generic function (as it was with `is_small_int()`), but it doesn't.

There are platforms besides x86 and I don't want to spend time assessing how inline function degrades performance on each of these platforms, and you?

sir-sigurd:
Here's inline function vs macro demo: https://godbolt.org/z/K5tbF2.
On x86-32:
* inline function version: 53 instructions
* macro version: 38 instructions

vstinner:
> On x86-32

Do you mean x86 (32 bits) or x32 (64-bit integers, but use 32 bit pointers)?

sir-sigurd:
@vstinner x86 (32 bits).

gnprice:
> > ~5% slowdown in microbenchmark on x86_32 (like you found for get_small_ints as function vs. macro)
> 
> Where did you get this ~5%? Did you run benchmark?

(That's from here:
https://bugs.python.org/issue38015#msg351255

Just as you quoted me saying: it's what you found for `get_small_ints` as a function vs. as a macro.)

gnprice:
> There are platforms besides x86 and I don't want to spend time assessing how inline function degrades performance on each of these platforms, and you?

Consider this the other way around: I don't think it's a good use of time to investigate what small unnecessary bits of work each compiler on each platform emits, or going to great lengths in our code to coax them into doing a bit better.

(Remember that there's absolutely nothing in the language that means a compiler *can't* emit the exact same code in the static-function case as you're seeing in the macro case -- it's purely a matter of how smart the compiler gets. See e.g. this demo: https://github.com/python/cpython/pull/15216#issuecomment-525464723 )

That can be well worth it when the payoff is large -- when there's an opportunity for a significant win on ordinary Python code. (Or as a good proxy for that: a large win on code that shows up in profiles of ordinary Python code.) But many-line macros and other code complications have a real cost, and the payoff has to be worth the cost.

Compilers will emit less-than-optimal code. That's a fact of life, no matter what we do to our code. We have to keep going anyway and write the code in a way that helps it make sense to each other as humans.

When the compiler emits code that's much slower than it could be, in a place where that significantly matters, that's one thing... but if we insisted on squeezing the last cycle out of every function, we'd never get anything else done.

sir-sigurd:
> > > ~5% slowdown in microbenchmark on x86_32 (like you found for get_small_ints as function vs. macro)
> > 
> > 
> > Where did you get this ~5%? Did you run benchmark?
> 
> (That's from here:
> https://bugs.python.org/issue38015#msg351255
> 
> Just as you quoted me saying: it's what you found for `get_small_ints` as a function vs. as a macro.)

This 5% of difference is the cost of 2 extra instructions on x86_64, and if you check [demo](https://godbolt.org/z/K5tbF2), you can see that in this case there is more substantial difference (53 vs 38 instructions and some of them are in loop).

sir-sigurd:
> But many-line macros and other code complications have a real cost, and the payoff has to be worth the cost.

This code is trivial, making it a macro doesn't make it less trivial. It makes it less "beautiful", but this code was modified [once](https://github.com/python/cpython/commit/86aa269646fa73bbcbc26f45ed854359d04c1fde) in the last 10 years, so I doubt there are many people interested in reading it. 

> (Remember that there's absolutely nothing in the language that means a compiler can't emit the exact same code in the static-function case as you're seeing in the macro case -- it's purely a matter of how smart the compiler gets. See e.g. this demo: #15216 (comment) )

My PR provides speed-up be eliminating these lines:

https://github.com/python/cpython/blob/c59295a1ca304f37ca136dd7efca9e560db27d28/Objects/longobject.c#L390-L391

I guess that the person who wrote these lines many years ago was thinking the same way as you, years have passed, but compilers are still not so smart.

Objects/longobject.c
```diff
@@ -378,32 +380,51 @@ PyLong_FromLong(long ival)
...
+        } \
+        /* Count the number of Python digits. */ \
+        Py_ssize_t ndigits = 0; \
+        INT_TYPE t = (ival); \
+        while (t) { \
+            ++ndigits; \
+            t >>= PyLong_SHIFT; \
+        } \
+        PyLongObject *v = _PyLong_New(ndigits); \
+        if (v != NULL) { \
```
vstinner: Would you mind to write instead:

```
if (v == NULL) {
    return NULL;
}
```

I dislike long `if (... != NULL) { ... } return ...;` blocks which add one indentation level.

vstinner (approved):
LGTM, but I have one last request.

gpshead (approved):

PR #15197, opened 2019-08-10
Fix docs for assert_called and assert_called_once

ismail-s:
The code that this part of the docs documents is: https://github.com/python/cpython/blob/7d44e7a4563072d0fad00427b76b94cad61c38ae/Lib/unittest/mock.py#L838-L846


tirkarthi (approved):
LGTM. Thanks for catching this. cc: @cjw296 @mariocj89

cjw296 (approved):

tirkarthi:
@cjw296 Could this be backported to 3.8 and 3.7 since this is a doc fix? @miss-islington will take care of it as labels are added. Thanks for the merge :)

PR #15198, opened 2019-08-10
There is no pop(0)...

paulhoule:

### DIFF
--- a/Doc/library/collections.rst
+++ b/Doc/library/collections.rst
@@ -429,7 +429,7 @@ or subtracting from an empty counter.
 
     Though :class:`list` objects support similar operations, they are optimized for
     fast fixed-length operations and incur O(n) memory movement costs for
-    ``pop(0)`` and ``insert(0, v)`` operations which change both the size and
+    ``remove(0)`` and ``insert(0, v)`` operations which change both the size and
     position of the underlying data representation.
 
 


the-knights-who-say-ni:

sir-sigurd:
```
In [17]: l = [1, 2]

In [18]: l.pop(0)
Out[18]: 1

In [19]: l
Out[19]: [2]
```

Doc/library/collections.rst
```diff
@@ -429,7 +429,7 @@ or subtracting from an empty counter.
 
     Though :class:`list` objects support similar operations, they are optimized for
     fast fixed-length operations and incur O(n) memory movement costs for
-    ``pop(0)`` and ``insert(0, v)`` operations which change both the size and
+    ``remove(0)`` and ``insert(0, v)`` operations which change both the size and
     position of the underlying data representation.
```
Mariatta, line 433: It this correct that pop() doesn't take any parameter.
But remove(0) also doesn't seem right. Remove accepts the `value` to be removed, not index to be removed. Perhaps a better replacement is `remove(v)`.

I'll wait for @rhettinger's feedback on this.

aeros (commented):
Thanks for the PR.

I would recommend adjusting the title to something a bit more descriptive of the documentation being updated, the current title is a bit ambiguous. Something along the lines of "Docs: Fix collections.deque" or "Fix pop(0) documentation in collections.deque". This makes it far easier for relevant reviewers and experts to know what the PR is addressing without having to open it first.

Doc/library/collections.rst
sir-sigurd: This paragraph is about `list`. `list.pop()` takes `index` parameter.

sir-sigurd (changes requested):
PR is invalid.

Doc/library/collections.rst
aeros: @sir-sigurd Ah, it looks that the PR author might not have realized the context. To avoid any further confusion, we could make it a little bit more explicit:
```
 ``list.pop(0)`` and ``list.insert(0, v)`` operations which change both the size and
```

willingc: Hi @paulhoule. Thanks for taking the time to submit a PR.

This paragraph of the documentation is comparing the deque's memory management behavior to a list's behavior when making a change in the contents of the data collection (specifically, if acting on the first element of the collection). 

Compared to a list, the advantage of a deque is memory efficiency. The deque avoids moving memory around by adding or removing data from typically the front or end of the collection. As a doubly linked list, adding or removing an item from a deque does not move any of the other elements in memory. Instead pointers to the prior or next element in the deque are updated.

For a **list**, `pop()` is equivalent to `pop(-1)` to remove the last item in a list. `pop(0)` of a list would remove the first item in the list. Unlike a deque, removing an item from the first position on the list could cause movement of items stored in memory.

[Reference to source code](https://github.com/python/cpython/blob/762f93ff2efd6b7ef0177cad57939c0ab2002eac/Modules/_collectionsmodule.c#L34)

The original paragraph is trying to illustrate that adding a new element (`insert(0, v)`) or deleting an element (`pop(0)`) to the beginning of the list (index 0) can impact where elements are stored in memory. This differs from a deque since adding an item to the beginning of the collection or removing from the beginning (`popleft()`) does not move or shift the other elements in memory. 

`remove(0)` removes the value equal to the passed parameter in this case 0. `remove(0)` is a different concept since the documentation is illustrating a point about what happens when acting on the first element of a list or deque.

To summarize, the documentation is technically accurate as written. If you would like to update the PR to clarify the concept more, please feel free to update your PR. Alternatively, please feel free to close the PR. Thanks!

rhettinger:
Sorry, this PR is incorrect.  LIsts do have `pop(0)` and `insert(0, value)` operations that are `O(n)`.  Providing a fast `O(1)` alternative was in fact the principal motivation for adding *deque()*.

```
>>> s = dir(list)
>>> s.insert(0, 'new')
>>> s.pop(0)
'new'
```

PR #15200, opened 2019-08-10
bpo-14465: Add an indent() function to xml.etree.ElementTree to pretty-print XML trees

scoder:
https://bugs.python.org/issue14465


scoder:
Merging, since there were no objections in the last two weeks.

PR #15201, opened 2019-08-10
bpo-37810: Fix difflib `?` hint in diff output when dealing with tabs

asottile:
https://bugs.python.org/issue37810

### DIFF
--- a/Lib/difflib.py
+++ b/Lib/difflib.py
@@ -733,20 +733,15 @@ def get_close_matches(word, possibilities, n=3, cutoff=0.6):
     # Strip scores for the best n matches
     return [x for score, x in result]
 
-def _count_leading(line, ch):
-    """
-    Return number of `ch` characters at the start of `line`.
 
-    Example:
+def _keep_original_ws(s, tag_s):
+    """Replace whitespace with the original whitespace characters in `s`"""
+    return ''.join(
+        c if tag_c == " " and c.isspace() else tag_c
+        for c, tag_c in zip(s, tag_s)
+    )
 
-    >>> _count_leading('   abc', ' ')
-    3
-    """
 
-    i, n = 0, len(line)
-    while i < n and line[i] == ch:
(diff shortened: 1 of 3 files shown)


Lib/difflib.py
```diff
@@ -1047,22 +1032,28 @@ def _qformat(self, aline, bline, atags, btags):
...
+            def _decide(c, tag_c):
+                if not tag_c.isspace():
+                    return tag_c
+                # preserve tab indentation in original string
+                elif c.isspace():
+                    return c
+                else:
+                    return ' '
+
+            return ''.join(_decide(c, tag_c) for c, tag_c in zip(s, tag_s))
```
tim-one: The only kind of whitespace in a tag string is a blank, so this can be replaced with the simpler (and, to my eyes, clearer) clause
```python
    c if tag_c == " " and c.isspace() else tag_c
```
Then there's no need for a nested `_decide()` function (or the expense of calling it many times), and the body of `_keep_original_ws()` can be one statement:
```python
    # Replace blanks in tag_s with the corresponding whitespace
    # characters (if any) in `s`.  The aim is to improve visual
    # output when `s` contains tabs.
    return ''.join(c if tag_c == " " and c.isspace() else tag_c
                   for c, tag_c in zip(s, tag_s))
```
It would also be good then to move this private function to module level, (as the old `_count_leading()` was), so we don't incur the expense of creating a new function each time `_qformat()` is called.

tim-one (changes requested):
This is a fine idea, much better than the original hack!  As noted, I'd like to see some changes.

Lib/difflib.py
asottile: heh, I had that originally but thought it hard to read -- I can do that though :+1:

asottile:
I didn't expect the Spanish Inquisition!

tim-one (approved):

PR #15202, opened 2019-08-10
bpo-37811: FreeBSD, OSX: fix poll(2) usage in sockets module

akhramov:
FreeBSD implementation of poll(2) restricts the timeout argument to be
either zero, or positive, or equal to INFTIM (-1).

Unless otherwise overridden, socket timeout defaults to -1. This value
is then converted to milliseconds (-1000) and used as argument to the
poll syscall. poll returns EINVAL (22), and the connection fails.

This bug was discovered during the EINTR handling testing, and the
reproduction code can be found in
https://bugs.python.org/issue23618 (see connect_eintr.py,
attached). On GNU/Linux, the example runs as expected.

This change is trivial:
If the supplied timeout value is negative, truncate it to -1.

https://bugs.python.org/issue37811

### DIFF
--- a/Modules/socketmodule.c
+++ b/Modules/socketmodule.c
@@ -789,6 +789,17 @@ internal_select(PySocketSockObject *s, int writing, _PyTime_t interval,
     ms = _PyTime_AsMilliseconds(interval, _PyTime_ROUND_CEILING);
     assert(ms <= INT_MAX);
 
+    /* On some OSes, typically BSD-based ones, the timeout parameter of the
+       poll() syscall, when negative, must be exactly INFTIM, where defined,
+       or -1. See issue 37811. */
+    if (ms < 0) {
+#ifdef INFTIM
+        ms = INFTIM;
+#else
+        ms = -1;
+#endif
+    }
+
     Py_BEGIN_ALLOW_THREADS;
     n = poll(&pollfd, 1, (int)ms);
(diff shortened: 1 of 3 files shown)


the-knights-who-say-ni:

akhramov:
@vstinner I'm sorry if you're the not the person I am supposed to ping, but can you take a glance at this PR?

Modules/socketmodule.c
```diff
@@ -789,6 +789,8 @@ internal_select(PySocketSockObject *s, int writing, _PyTime_t interval,
     ms = _PyTime_AsMilliseconds(interval, _PyTime_ROUND_CEILING);
     assert(ms <= INT_MAX);
 
+    ms = ms < 0 ? -1 : ms;
```
vstinner: Please add a comment explaining why you do that.

I would prefer a regular if to make it more explicit that the value is only modified it's negative. Like:

```
if (ms < -1) {
    /* FreeBSD implementation of poll(2) restricts timeout argument to be either zero, or positive, or equal to INFTIM (-1). */
    ms = -1;
}
```

akhramov: Sure thing, see the updates. I borrowed the wording from #4033 -- the author was tackling exactly the same `poll(2)` problem.

Also I have created a news entry. I am not sure whether there's a necessity, though.

PR #15203, opened 2019-08-10
bpo-37812: Make the small-int alloc optimization unconditional.

gnprice:
This optimization was added in 1993 (commit 842d2ccdc), and we've
been running with it ever since.  I think at this point it's fair
to say that it's no longer experimental.  That lets us simplify
the code by removing a number of `#ifdef`s.

In particular this will help us simplify the usage of CHECK_SMALL_INT.

### DIFF
--- a/Misc/NEWS.d/next/Core and Builtins/2019-08-10-14-15-48.bpo-37812.jkh0Wp.rst
+++ b/Misc/NEWS.d/next/Core and Builtins/2019-08-10-14-15-48.bpo-37812.jkh0Wp.rst
@@ -1,0 +1,6 @@
+The "small int" optimization is now unconditionally enabled.  This
+optimization was introduced in Python 1.0.1, and means that :class:`int`
+values from -5 to 256 are pre-initialized rather than constructed on demand.
+The option was always enabled unless certain never-documented preprocessor
+macros were defined to 0 at build time; now it is always enabled,
+simplifying the logic.  Patch by Greg Price.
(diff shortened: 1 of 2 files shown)


Misc/NEWS.d/next/Core and Builtins/2019-08-10-14-15-48.bpo-37812.jkh0Wp.rst
```diff
@@ -0,0 +1,6 @@
+The "small int" optimization is now unconditionally enabled.  This
+optimization was introduced in Python 1.0.1, and means that :class:`int`
+values from -5 to 256 are pre-initialized rather than constructed on demand.
```
aeros: I think it is worthwhile to mention that this range is inclusive. Otherwise, users might assume that the maximum is 255 instead of 256, or that the minimum is -4 instead of -5. I've seen this range most commonly notated with `[-5, 257)`, but I think adding "(inclusive)" after it is probably more appropriate and easy to understand for a news entry.
```suggestion
values from -5 to 256 (inclusive) are pre-initialized rather than constructed on demand.
```

aeros (commented):
Thanks for the PR.

I had a minor suggestion for the news entry:

Misc/NEWS.d/next/Core and Builtins/2019-08-10-14-15-48.bpo-37812.jkh0Wp.rst
gnprice: Sure, thanks. I don't want to make a big deal of the exact numbers because it's not something it makes sense for people to depend on -- but it's a natural question for a reader to be curious about. I made it "-5 through 256", which I think is unambiguous while staying unobtrusive.

PR #15205, opened 2019-08-11
gh-81989: Deprecate passing a single argument to super()

ZackerySpytz:
https://bugs.python.org/issue37808

* Issue: gh-81989

### DIFF
--- a/Doc/library/functions.rst
+++ b/Doc/library/functions.rst
@@ -1640,6 +1640,9 @@ are always available.  They are listed here in alphabetical order.
    :func:`super`, see `guide to using super()
    <https://rhettinger.wordpress.com/2011/05/26/super-considered-super/>`_.
 
+   .. deprecated:: 3.9
+      Calling :func:`super` with one argument.
+
 
 .. _func-tuple:
 .. function:: tuple([iterable])
--- a/Objects/typeobject.c
+++ b/Objects/typeobject.c
@@ -7822,6 +7822,12 @@ super_init(PyObject *self, PyObject *args, PyObject *kwds)
     if (!PyArg_ParseTuple(args, "|O!O:super", &PyType_Type, &type, &obj))
         return -1;
 
+    if (PyTuple_GET_SIZE(args) == 1 &&
+        PyErr_WarnEx(PyExc_DeprecationWarning,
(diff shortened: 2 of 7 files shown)


Doc/library/functions.rst
```diff
@@ -1640,6 +1640,9 @@ are always available.  They are listed here in alphabetical order.
    :func:`super`, see `guide to using super()
    <https://rhettinger.wordpress.com/2011/05/26/super-considered-super/>`_.
 
+   .. deprecated:: 3.9
```
epicfaace, line 1643: Should we also mention that it's deprecated earlier on in the file where it says "If the second argument is omitted, the super object returned is unbound."?

Objects/typeobject.c
```diff
@@ -7822,6 +7822,12 @@ super_init(PyObject *self, PyObject *args, PyObject *kwds)
     if (!PyArg_ParseTuple(args, "|O!O:super", &PyType_Type, &type, &obj))
         return -1;
 
+    if (PyTuple_GET_SIZE(args) == 1 &&
+        PyErr_WarnEx(PyExc_DeprecationWarning,
+                     "Passing 1 argument to super() is deprecated", 1) < 0) {
```
epicfaace, line 7827: "Passing a single argument" might sound better / be more consistent with the rest of the wording, for the deprecation message, too.

Ri-yaB:
Subject: Feedback on Deprecating Unbound super()

Hi,

I’ve reviewed the proposal to deprecate unbound super() methods. Here are some thoughts:

1. Deprecation Warning Message: The current message, “Passing 1 argument to super() is deprecated,” could be clearer. How about:
"Using a single argument with super() is deprecated and will be removed in a future Python release. Please provide both a class and an instance to super() instead."

2. Documentation Updates: It might help to include an example in the documentation showing the recommended alternative to unbound super() for users who rely on this pattern.

3. Backward Compatibility: Some legacy code might still use unbound super(). Is there a plan to provide a transition guide or helper utility to identify such cases?

4. Testing Edge Cases: Have the changes been tested against scenarios involving inheritance chains, especially with multiple inheritance? Ensuring these cases work seamlessly post-deprecation is crucial.

I’m happy to assist further if needed. Let me know!

Best regards,
Riya

PR #1503, opened 2017-05-08
bpo-26543: fix `imaplib.noop()` failing when debug > 3

ankostis:
+ Add a failing TC for the problem.

https://bugs.python.org/issue26543

PR #15206, opened 2019-08-11
bpo-26543: Fix IMAP4.noop when debug mode is enabled

CuriousLearner:
This supersedes #1503 

I've tried to debug this issue and have the following observations:

- In case if `dump_ur` is directly called, then with this new patch, we're gracefully able to handle the bytes data and the new test cases pass.
- In case when `simple_command` is directly for `noop`, the `self.get_response` call inside  `_get_tagged_response` raises an error: 

```
socket error: unterminated line: b'* CAPABILITY IMAP4rev1 AUTH\n'
```

It is worth noting that this test case fails even if the debug level is decremented from the original of 3 as described in the bug.

I'm not sure what to do in such a situation to gracefully handle this. 

In case I just return the `typ` and `data` as in https://github.com/python/cpython/compare/master...CuriousLearner:[bpo-26543](https://bugs.python.org/issue26543)?expand=1#diff-ed3421b57957b3ffd24b79e746a88db4R1023

The test will pass, but I'm not sure what is the best approach to take here. Can you please provide a pointer?

https://bugs.python.org/issue26543


CuriousLearner:
Hi @vstinner 

Can you take a look at this and questions I posted above?

Thanks!

Lib/imaplib.py
```diff
@@ -1020,6 +1020,8 @@ def _command_complete(self, name, tag):
             self._check_bye()
         try:
             typ, data = self._get_tagged_response(tag, expect_bye=logout)
+            if name == "NOOP":
+                return typ, data
```
maxking: I don't think this is needed, if there are no untagged response from the server, otherwise, the command error is expected.

Lib/test/test_imaplib.py
```diff
@@ -879,6 +879,14 @@ def test_with_statement_logout(self):
                 self.assertIsNone(server.logged)
             self.assertIsNone(server.logged)
 
+    @reap_threads
+    def test_noop_with_debug(self):
```
maxking: For this test to work, there needs to be some untagged response from the server. 

A simpler test case IMO, which can reproduce the issue is to have a simple untagged_response_dict passed directly to `_dump_ur` and make sure that it is correctly able to handle bytes data.

maxking (changes requested):
I think the fix for `_dump_ur` should work, but we need a test case to make sure that it does indeed handle bytes data. 

We don't necessarily need to boot up a server, we could just pass the expected response data dict to the method and see that it prints correctly (we can capture stderr, so that it doesn't pollute test output).

Lib/imaplib.py
```diff
@@ -1226,13 +1228,10 @@ def _mesg(self, s, secs=None):
...
-            # Dump untagged responses (in `dict').
-            l = dict.items()
-            if not l: return
-            t = '\n\t\t'
-            l = map(lambda x:'%s: "%s"' % (x[0], x[1][0] and '" "'.join(x[1]) or ''), l)
-            self._mesg('untagged responses dump:%s%s' % (t, t.join(l)))
+        def _dump_ur(self, untagged_resp_dict):
+            if untagged_resp_dict:
+                self._mesg('untagged responses dump:' +
+                    ''.join('\n\t\t%s: %r' % (x[0], x[1]) for x in untagged_resp_dict.items()))
```
vstinner: I propose:

```
def _dump_ur(self, untagged_resp_dict): 
  if not untagged_resp_dict: 
    return
  items = ('%s: %r' % (key, value)
           for key, value in untagged_resp_dict.items())
  self._mesg('untagged responses dump:' + '\n\t\t'.join(items))
```

Lib/test/test_imaplib.py
vstinner: Yeah, I concur: test directly _dump_ur(). But since it's a private method, please decorate the test with @support.cpython_only since _dump_ur() is an implementation detail.

Misc/NEWS.d/next/Library/2019-08-11-16-28-03.bpo-26543.X-TJZO.rst
```diff
@@ -0,0 +1 @@
+Fix crash on :meth:`IMAP4.noop()` when bytes data is passed.
```
vstinner: For me, a "crash" means that the Python process is killed immediately, like a SIGSEGV. Here it's "just" a regular exception. Moreover, you should mention that the issue only affects debug mode. I suggest:

```
Fix :meth:`IMAP4.noop()` when debug mode is enabled (ex: ``imaplib.Debug = 3``).
```

csabella:
@CuriousLearner, please address the review comments from Victor.  Thanks!

rotuna:
I've implemented all requested changes in another PR (#19693)

CuriousLearner:
Ok, let me know if you want to take it forward, else I'll finish it now. I've been busy past few months.

csabella:
@rotuna, please respond to @CuriousLearner's question and the review on your PR if you would like to finish this PR.  My opinion is that @CuriousLearner should finish it on this PR if he's available to work on it now, but I'll let the two of you decide.

rotuna:
Hi, sorry I didn't get an email about this and I've been caught up in other things. If @CuriousLearner wants to finish this that'd be perfectly fine with me. If he's not free I can try to finish it up next week.

CuriousLearner:
No problem. I'll just finish it up. Thanks for your work :)

CuriousLearner:
The Travis shows error here:

```
  File "/home/travis/build/python/cpython/Lib/test/test_imaplib.py", line 583, in <module>

    class ThreadedNetworkedTests(unittest.TestCase):

  File "/home/travis/build/python/cpython/Lib/test/test_imaplib.py", line 936, in ThreadedNetworkedTests

    @reap_threads

NameError: name 'reap_threads' is not defined
```

We haven't touch that portion at all and `reap_threads` is already imported in the file. Not sure what is the issue here. @vstinner

vstinner:
You should use `@threading_helper.reap_threads`.

CuriousLearner:
@vstinner Thanks, issue fixed. Can you please take a look now?

vstinner:
Merged, thanks for the fix.

PR #15207, opened 2019-08-11
gh-82002: IDLE completion namespace fix

taleinat:
* Issue: gh-82002


iritkatriel (dismissed):
This has merge conflicts now.

terryjreedy:
Merge conflict in autocomplete.py: `__init__` parameter `tags` added for #25708, this adds `namespace`.  Put latter after former because `tags` passed positionally in editor.  Should make both or all 3 keyword-only in separate patch and fix call.

Merge conflict in run.py: changed code made conditional on not testing by #24302, which added tests for respecting excepthook.

PR #15209, opened 2019-08-11
Fix typos mostly in comments, docs and test names

minho42:
Found some more typos in various files.

### DIFF
--- a/Doc/library/importlib.rst
+++ b/Doc/library/importlib.rst
@@ -1379,7 +1379,7 @@ an :term:`importer`.
    bytecode file. An empty string represents no optimization, so
    ``/foo/bar/baz.py`` with an *optimization* of ``''`` will result in a
    bytecode path of ``/foo/bar/__pycache__/baz.cpython-32.pyc``. ``None`` causes
-   the interpter's optimization level to be used. Any other value's string
+   the interpreter's optimization level to be used. Any other value's string
    representation being used, so ``/foo/bar/baz.py`` with an *optimization* of
    ``2`` will lead to the bytecode path of
    ``/foo/bar/__pycache__/baz.cpython-32.opt-2.pyc``. The string representation
--- a/Doc/library/importlib.rst
+++ b/Doc/library/importlib.rst
@@ -1380,7 +1380,7 @@ an :term:`importer`.
    ``/foo/bar/baz.py`` with an *optimization* of ``''`` will result in a
    bytecode path of ``/foo/bar/__pycache__/baz.cpython-32.pyc``. ``None`` causes
    the interpreter's optimization level to be used. Any other value's string
-   representation being used, so ``/foo/bar/baz.py`` with an *optimization* of
+   representation is used, so ``/foo/bar/baz.py`` with an *optimization* of
    ``2`` will lead to the bytecode path of
    ``/foo/bar/__pycache__/baz.cpython-32.opt-2.pyc``. The string representation
    of *optimization* can only be alphanumeric, else :exc:`ValueError` is raised.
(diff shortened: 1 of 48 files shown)


Doc/library/importlib.rst
```diff
@@ -1379,7 +1379,7 @@ an :term:`importer`.
    bytecode file. An empty string represents no optimization, so
    ``/foo/bar/baz.py`` with an *optimization* of ``''`` will result in a
    bytecode path of ``/foo/bar/__pycache__/baz.cpython-32.pyc``. ``None`` causes
-   the interpter's optimization level to be used. Any other value's string
+   the interpreter's optimization level to be used. Any other value's string
    representation being used, so ``/foo/bar/baz.py`` with an *optimization* of
```
epicfaace:
```suggestion
   representation is used, so ``/foo/bar/baz.py`` with an *optimization* of
```

Lib/test/test_winreg.py
```diff
@@ -229,7 +229,7 @@ def test_connect_registry_to_local_machine_works(self):
         h.Close()
         self.assertEqual(h.handle, 0)
 
-    def test_inexistant_remote_registry(self):
+    def test_inexistent_remote_registry(self):
```
epicfaace:
```suggestion
    def test_nonexistent_remote_registry(self):
```

Lib/unittest/test/testmock/testpatch.py
```diff
@@ -1651,7 +1651,7 @@ def test_patch_imports_lazily(self):
             p1.stop()
         self.assertEqual(squizz.squozz, 3)
 
-    def test_patch_propogrates_exc_on_exit(self):
+    def test_patch_propagrates_exc_on_exit(self):
```
epicfaace:
```suggestion
    def test_patch_propagates_exc_on_exit(self):
```

epicfaace (changes requested):

Doc/library/importlib.rst
minho42: Thanks for the suggestions @epicfaace,
"Any other value’s string representation being used, so..."
This line is not very clear to me. I think there should be "If" at the beginning as well.
Anyway, I would like to leave this as is if possible, and get the typo bits merged first.

terryjreedy: Accept the suggestion as 'being' is effectively a typo.  With that change, the sentence is coherent, even if not optimal, whereas it it not now.

terryjreedy:
I am going to make and backport the idlelib changes, revert here, and then review.

Modules/expat/loadlibrary.c
```diff
@@ -88,7 +88,7 @@ HMODULE _Expat_LoadLibrary(LPCTSTR filename)
 
   /* Attempt to find LoadLibraryEx() which is only available on Windows 2000
      and above */
-  pLoadLibraryEx = (LOADLIBRARYEX_FN) GetProcAddress(hKernel32, LOADLIBARYEX);
+  pLoadLibraryEx = (LOADLIBRARYEX_FN) GetProcAddress(hKernel32, LOADLIBRARYEX);
```
terryjreedy, line 91: I verified with recursive grep that these are the only four instances of LOADLIBARYEX in any .c file or any .py file in lib or tools.  Note that the spelling is LOADLIBRARYEX_FN (with 2nd R) on line 51.

terryjreedy (approved):
All changes look correct and non-controversial.  I checked that the only code change (LOADLIBARYEX to LOADLIBRARYEX) was complete and consistent.

terryjreedy:
I expected the 3.8 backport to fail; good thing I tried it anyway.
If you or anyone makes a 3.7 backport PR, I think the LOADLIBARYEX fix should be skipped in case people have used it in their code.  They should not be surprised until they try 3.8.

minho42:
Thank you @terryjreedy

PR #15212, opened 2019-08-11
bpo-37819: Add Fraction.as_integer_ratio()

rhettinger:
https://bugs.python.org/issue37819

### DIFF
--- a/Doc/library/fractions.rst
+++ b/Doc/library/fractions.rst
@@ -96,8 +96,8 @@ another rational number, or from a string.
 
    .. method:: as_integer_ratio()
 
-      Return a tuple of integers, whose ratio is exactly equal
-      to the original Fraction and with a positive denominator.
+      Return a tuple of two integers, whose ratio is equal
+      to the original fraction and with a positive denominator.
 
       .. versionadded:: 3.8
 
--- a/Doc/library/fractions.rst
+++ b/Doc/library/fractions.rst
@@ -97,7 +97,7 @@ another rational number, or from a string.
    .. method:: as_integer_ratio()
 
       Return a tuple of two integers, whose ratio is equal
-      to the original fraction and with a positive denominator.
+      to the Fraction and with a positive denominator.
 
       .. versionadded:: 3.8
 
--- a/Doc/library/fractions.rst
+++ b/Doc/library/fractions.rst
@@ -94,6 +94,13 @@ another rational number, or from a string.
       Denominator of the Fraction in lowest term.
 
 
+   .. method:: as_integer_ratio()
+
+      Return a tuple of integers, whose ratio is exactly equal
+      to the original Fraction and with a positive denominator.
+
+      .. versionadded:: 3.8
+
    .. method:: from_float(flt)
 
       This class method constructs a :class:`Fraction` representing the exact
--- a/Lib/fractions.py
+++ b/Lib/fractions.py
@@ -219,8 +219,8 @@ class Fraction(numbers.Rational):
     def as_integer_ratio(self):
         """Return the integer ratio as a tuple.
 
-        Return a tuple of two integers, whose ratio is equal to the original
-        fraction and with a positive denominator.
+        Return a tuple of two integers, whose ratio is equal to the
+        Fraction and with a positive denominator.
         """
         return (self._numerator, self._denominator)
 
--- a/Lib/fractions.py
+++ b/Lib/fractions.py
@@ -217,10 +217,10 @@ class Fraction(numbers.Rational):
         return cls(*dec.as_integer_ratio())
 
     def as_integer_ratio(self):
-        """Return integer ratio.
+        """Return the integer ratio as a tuple.
 
-        Return a pair of integers, whose ratio is exactly equal to the original
-        Fraction and with a positive denominator.
+        Return a tuple of two integers, whose ratio is equal to the original
+        fraction and with a positive denominator.
         """
         return (self._numerator, self._denominator)
 
--- a/Lib/fractions.py
+++ b/Lib/fractions.py
@@ -216,6 +216,14 @@ class Fraction(numbers.Rational):
                 (cls.__name__, dec, type(dec).__name__))
         return cls(*dec.as_integer_ratio())
 
+    def as_integer_ratio(self):
+        """Return integer ratio.
+
+        Return a pair of integers, whose ratio is exactly equal to the original
+        Fraction and with a positive denominator.
+        """
+        return (self._numerator, self._denominator)
+
     def limit_denominator(self, max_denominator=1000000):
         """Closest Fraction to self with denominator at most max_denominator.
 
--- a/Lib/test/test_fractions.py
+++ b/Lib/test/test_fractions.py
@@ -302,6 +302,12 @@ class FractionTest(unittest.TestCase):
             ValueError, "cannot convert NaN to integer ratio",
             F.from_decimal, Decimal("snan"))
 
+    def test_as_integer_ratio(self):
+        self.assertEqual(F(4, 6).as_integer_ratio(), (2, 3))
+        self.assertEqual(F(-4, 6).as_integer_ratio(), (-2, 3))
+        self.assertEqual(F(4, -6).as_integer_ratio(), (-2, 3))
+        self.assertEqual(F(0, 6).as_integer_ratio(), (0, 1))
+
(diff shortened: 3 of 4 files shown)


Doc/library/fractions.rst
```diff
@@ -94,6 +94,13 @@ another rational number, or from a string.
       Denominator of the Fraction in lowest term.
 
 
+   .. method:: as_integer_ratio()
+
+      Return a tuple of integers, whose ratio is exactly equal
```
gvanrossum: Drop "exactly"

Lib/fractions.py
```diff
@@ -216,6 +216,14 @@ def from_decimal(cls, dec):
                 (cls.__name__, dec, type(dec).__name__))
         return cls(*dec.as_integer_ratio())
 
+    def as_integer_ratio(self):
+        """Return integer ratio.
+
+        Return a pair of integers, whose ratio is exactly equal to the original
```
gvanrossum: Ditto.

gvanrossum (commented):
I would change "exactly equal" to just "equal" -- there's nothing to be gained by adding "exactly".

Misc/NEWS.d/next/Library/2019-08-11-10-34-19.bpo-37819.LVJls-.rst
```diff
@@ -0,0 +1,2 @@
+Add Fraction.as_integer_ratio() to match the corresponding methods in bool,
```
aeros, line 1: Would it be helpful to provide a link to the method using inline markup? This would allow readers to easily find the documentation for the new method.

Also regarding this line, as far as I'm aware, there's not a bool.as_integer_ratio() method. To confirm this, I used git grep to search the documentation:

image

As expected, it returned the methods for decimal, int, and float but nothing for bool.

```suggestion
Add :meth:`Fraction.as_integer_ratio` to match the corresponding methods in
```

Lib/fractions.py
aeros: Would it be potentially helpful to readers to specify in the doc-string that the integer pair is returned as a tuple? This is not uncommonly done, but I think it's worthwhile to specifically mention this:
```suggestion
        Return a tuple containing two integers, whose ratio is exactly equal to the original
```

Lib/test/test_fractions.py
```diff
@@ -302,6 +302,12 @@ def testFromDecimal(self):
             ValueError, "cannot convert NaN to integer ratio",
             F.from_decimal, Decimal("snan"))
 
+    def test_as_integer_ratio(self):
```
aeros, line 305: I did a comparison of these tests to the existing ones, and noticed that `test_decimal.py` contains coverage for attempting to convert inf (`inf = 1e1000`), -inf, nan (`nan = inf - inf`), and an invalid string to an integer ratio:

```python
    def test_as_integer_ratio(self):
        Decimal = self.decimal.Decimal

        # exceptional cases
        self.assertRaises(OverflowError,
                          Decimal.as_integer_ratio, Decimal('inf'))
        self.assertRaises(OverflowError,
                          Decimal.as_integer_ratio, Decimal('-inf'))
        self.assertRaises(ValueError,
                          Decimal.as_integer_ratio, Decimal('-nan'))
        self.assertRaises(ValueError,
                          Decimal.as_integer_ratio, Decimal('snan123'))
```
For the sake of consistency, should we include that here as well? I'm not certain that it's necessary, but I figured it was worth mentioning.

aeros (commented):
Thanks for the PR @rhettinger. 

I had some minor suggestions:

Doc/library/fractions.rst
aeros: @gvanrossum To some degree, I think that mentioning "exactly" might help readers to understand the advantage of using an integer ratio when they need more precision. However, there might be a better way to mention this.

aeros167:
I didn't include this in my review since it's not directly related to the PR, but I noticed some minor inconsistencies in the tests between the different versions of `*.as_integer_ratio`. For example, the test function in float was called `test_floatasratio`, and it also seemed to be missing the test that decimal uses to ensure a ValueError is raised when attempting to use an invalid string:

```python
       self.assertRaises(ValueError,
                          Decimal.as_integer_ratio, Decimal('snan123'))
```

Also, as far as I can tell, there's no test coverage at all for the int version of the method. Should I open a new issue for this?

Doc/library/fractions.rst
rhettinger: Thanks. I'll drop this "exactly".  The wording was take from the docstring for *float.as_integer_ratio()* but in makes a lot more sense.

Lib/fractions.py
rhettinger: Will do.

Misc/NEWS.d/next/Library/2019-08-11-10-34-19.bpo-37819.LVJls-.rst
rhettinger: We don't have to provide cross-links from Misc/NEWS.  It is okay for entries there to be plain text.

Lib/fractions.py
rhettinger: Will do.

Lib/test/test_fractions.py
rhettinger: No need.  Fractions don't have that concept.

Misc/NEWS.d/next/Library/2019-08-11-10-34-19.bpo-37819.LVJls-.rst
aeros: @rhettinger Oh yeah it's okay either way, but since the Misc/NEWS entries support reST and Sphinx I think it's helpful to include links to functions, methods, or classes that were modified in functionality or documentation. Usually the news entries are very brief, so if a curious reader wants to get more information they can simply use the link.

To those of us who are familiar with the structuring of the documentation, it's usually easy to find the relevant sections. But, some less familiar or newer readers might have trouble doing so, especially if they use an external search engine (which frequently will lead them to the 2.7 docs or an incorrect page).

It's more a convenience/QoL addition for the readers than anything, and I don't think it hurts the readability or takes much effort from our end. I might start a thread over on python-dev to see what the others think. Not at all for this PR specifically, but just in general.

Edit: ~~Just realized that I can't start threads on the Python-Dev list. I'll open a new topic on Discuss instead, let me know if you'd like a link to the thread.~~ Never mind, I'm able to create a new thread on the ML. I'll do so later tonight, in the meantime here's the discuss post: https://discuss.python.org/t/should-news-entries-contain-documentation-links/2127

aeros: Also, what was the conclusion on `bool` being in the news entry? From my understanding, there is not a `bool.as_integer_ratio()` method.

jdemeyer:
Should this `as_integer_ratio` instead be moved to the `numbers.Rational` base class? The implementation already makes sense on that level, so it seems more natural to do it that way.

Misc/NEWS.d/next/Library/2019-08-11-10-34-19.bpo-37819.LVJls-.rst
methane: @aeros167 Have you validated your understanding?

From my understanding, bool inherits from int.  If int has `as_integer_ratio()`, bool should have it too.

```
$ ./python
Python 3.9.0a0 (heads/master:e95ac20103, Jul 24 2019, 21:48:01)
[GCC 8.3.0] on linux
>>> i = 42
>>> i.as_integer_ratio()
(42, 1)
>>> True.as_integer_ratio()
(1, 1)
```

aeros: @methane Thanks for the explanation!

At first this didn't make sense to me, but then I remembered that:

```python3
>>> int(True)
1
>>> int(False)
0
>>> True == 1
True
>>> False == 0
True
```

So it makes sense that the boolean values would also have the int methods, since they are effectively specialized ints. I was just confused at first since I could not find an explicit mention of ``bool.is_integer_ratio()`` in the docs, but I suppose it doesn't need to be mentioned since as you said, bool inherits from int.

PR #15216, opened 2019-08-11
bpo-37812: Expand confusing CHECK_SMALL_INT so return is explicit.

gnprice:
It's not obvious when looking at a call site of this macro that it
affects the control flow.  Move the `return` into the callers so
it's explicit there, and factor out the fiddly bit as IS_SMALL_INT.

As a bonus, a couple of other spots can also use IS_SMALL_INT with a
benefit to clarity.

### DIFF
--- a/Objects/longobject.c
+++ b/Objects/longobject.c
@@ -41,6 +41,9 @@ PyObject *_PyLong_One = NULL;
    -NSMALLNEGINTS (inclusive) to NSMALLPOSINTS (not inclusive).
 */
 static PyLongObject small_ints[NSMALLNEGINTS + NSMALLPOSINTS];
+
+#define IS_SMALL_INT(ival) (-NSMALLNEGINTS <= (ival) && (ival) < NSMALLPOSINTS)
+
 #ifdef COUNT_ALLOCS
 Py_ssize_t _Py_quick_int_allocs, _Py_quick_neg_int_allocs;
 #endif
@@ -49,7 +52,7 @@ static PyObject *
 get_small_int(sdigit ival)
 {
     PyObject *v;
-    assert(-NSMALLNEGINTS <= ival && ival < NSMALLPOSINTS);
+    assert(IS_SMALL_INT(ival));
     v = (PyObject *)&small_ints[ival + NSMALLNEGINTS];
     Py_INCREF(v);
 #ifdef COUNT_ALLOCS
@@ -60,17 +63,13 @@ get_small_int(sdigit ival)
 #endif
     return v;
 }
-#define CHECK_SMALL_INT(ival) \
-    do if (-NSMALLNEGINTS <= ival && ival < NSMALLPOSINTS) { \
-        return get_small_int((sdigit)ival); \
-    } while(0)
 
 static PyLongObject *
 maybe_small_long(PyLongObject *v)
 {
     if (v && Py_ABS(Py_SIZE(v)) <= 1) {
         sdigit ival = MEDIUM_VALUE(v);
-        if (-NSMALLNEGINTS <= ival && ival < NSMALLPOSINTS) {
+        if (IS_SMALL_INT(ival)) {
             Py_DECREF(v);
             return (PyLongObject *)get_small_int(ival);
         }
@@ -78,7 +77,8 @@ maybe_small_long(PyLongObject *v)
     return v;
 }
 #else
-#define CHECK_SMALL_INT(ival)
+#define IS_SMALL_INT(ival) 0
+#define get_small_int(ival) (assert(0), NULL)
 #define maybe_small_long(val) (val)
 #endif
 
@@ -293,7 +293,9 @@ _PyLong_Copy(PyLongObject *src)
         i = -(i);
     if (i < 2) {
         sdigit ival = MEDIUM_VALUE(src);
-        CHECK_SMALL_INT(ival);
+        if (IS_SMALL_INT(ival)) {
+            return get_small_int(ival);
+        }
     }
     result = _PyLong_New(i);
     if (result != NULL) {
@@ -315,7 +317,9 @@ PyLong_FromLong(long ival)
     int ndigits = 0;
     int sign;
 
-    CHECK_SMALL_INT(ival);
+    if (IS_SMALL_INT(ival)) {
+        return get_small_int((sdigit)ival);
+    }
 
     if (ival < 0) {
         /* negate: can't write this as abs_ival = -ival since that
@@ -1146,7 +1150,10 @@ PyLong_FromLongLong(long long ival)
     int ndigits = 0;
     int negative = 0;
 
-    CHECK_SMALL_INT(ival);
+    if (IS_SMALL_INT(ival)) {
+        return get_small_int((sdigit)ival);
+    }
+
     if (ival < 0) {
         /* avoid signed overflow on negation;  see comments
            in PyLong_FromLong above. */
@@ -1218,7 +1225,10 @@ PyLong_FromSsize_t(Py_ssize_t ival)
     int ndigits = 0;
     int negative = 0;
 
-    CHECK_SMALL_INT(ival);
+    if (IS_SMALL_INT(ival)) {
+        return get_small_int((sdigit)ival);
+    }
+
     if (ival < 0) {
         /* avoid signed overflow when ival = SIZE_T_MIN */
         abs_ival = (size_t)(-1-ival)+1;
--- a/Objects/longobject.c
+++ b/Objects/longobject.c
@@ -42,7 +42,11 @@ PyObject *_PyLong_One = NULL;
 */
 static PyLongObject small_ints[NSMALLNEGINTS + NSMALLPOSINTS];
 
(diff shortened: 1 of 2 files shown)


Objects/longobject.c
```diff
@@ -41,6 +41,9 @@ PyObject *_PyLong_One = NULL;
    -NSMALLNEGINTS (inclusive) to NSMALLPOSINTS (not inclusive).
 */
 static PyLongObject small_ints[NSMALLNEGINTS + NSMALLPOSINTS];
+
+#define IS_SMALL_INT(ival) (-NSMALLNEGINTS <= (ival) && (ival) < NSMALLPOSINTS)
```
benjaminp: Why not make this a function?

gnprice: I actually have a patch for that too :slightly_smiling_face: : https://github.com/gnprice/cpython/commit/cfc91e6ca

I left that out of this commit in order to change the fewest possible aspects at once, and figured I'd send it as a separate followup provided this one is merged. But I'd also be happy to add it to this PR, if there's agreement that it's desirable; I think it'd fit well as part of the same change.

(Note the "before" side of that commit looks slightly different from the "after" of this one, as there are a couple of other commits in between on that development branch. But it'd be no trouble for me to rebase it to directly on top of this one in order to add to this PR.)

benjaminp: Okay, if making a function is incoming, this PR lgtm.

I'll let Raymond weight in.

rhettinger: Go ahead and make it a static inlined function.

rhettinger (changes requested):

Objects/longobject.c
rhettinger: This is a somewhat minor change, but let's add a blurb NEWS entry in case anyone cases about these details.

gnprice: Sounds good! Thanks for the review.

gnprice:
Thanks for the reviews!

animalize:
Using inlined function generates unnecessary machine code than using macro.
I built a 32-bit Release version with VC2017, below is the starting code of `PyLong_FromLong` function.

The `ival` was first converted to a `long long` type, this is unnecessary .

use inlined function:
```asm
_PyLong_FromLong PROC					; COMDAT

; 317  : {

	push	ebp
	mov	ebp, esp
	and	esp, -8					; fffffff8H
	push	ecx
	mov	ecx, DWORD PTR _ival$[ebp]

; 318  :     PyLongObject *v;
; 319  :     unsigned long abs_ival;
; 320  :     unsigned long t;  /* unsigned so >> doesn't propagate sign bit */
; 321  :     int ndigits = 0;
; 322  :     int sign;
; 323  : 
; 324  :     if (is_small_int(ival)) {

	mov	eax, ecx
	push	ebx
	push	esi
	cdq
	push	edi
	xor	edi, edi

; 48   :     return -NSMALLNEGINTS <= ival && ival < NSMALLPOSINTS;

	add	eax, 5
	adc	edx, edi
	test	edx, edx
	ja	SHORT $LN6@PyLong_Fro
	jb	SHORT $LN32@PyLong_Fro
	cmp	eax, 261				; 00000105H
	ja	SHORT $LN6@PyLong_Fro
$LN32@PyLong_Fro:

; 60   :     v = (PyObject *)&small_ints[ival + NSMALLNEGINTS];

	movsx	eax, cx
	shl	eax, 4
	add	eax, OFFSET _small_ints+80
; File d:\dev\cpython\include\object.h

; 459  :     op->ob_refcnt++;

	inc	DWORD PTR [eax]
; File d:\dev\cpython\objects\longobject.c

; 383  : }

	pop	edi
	pop	esi
	pop	ebx
	mov	esp, ebp
	pop	ebp
	ret	0
```

use macro:
```asm
_PyLong_FromLong PROC					; COMDAT

; 318  : {

	push	ebp
	mov	ebp, esp
	mov	eax, DWORD PTR _ival$[ebp]
	push	edi

; 319  :     PyLongObject *v;
; 320  :     unsigned long abs_ival;
; 321  :     unsigned long t;  /* unsigned so >> doesn't propagate sign bit */
; 322  :     int ndigits = 0;

	xor	edi, edi

; 323  :     int sign;
; 324  : 
; 325  :     if (IS_SMALL_INT(ival)) {

	cmp	eax, -5					; fffffffbH
	jl	SHORT $LN6@PyLong_Fro
	cmp	eax, 257				; 00000101H
	jge	SHORT $LN6@PyLong_Fro

; 60   :     v = (PyObject *)&small_ints[ival + NSMALLNEGINTS];

	cwde
	shl	eax, 4
	add	eax, OFFSET _small_ints+80
	pop	edi
; File d:\dev\cpython\include\object.h

; 459  :     op->ob_refcnt++;

	inc	DWORD PTR [eax]
; File d:\dev\cpython\objects\longobject.c

; 384  : }

	pop	ebp
	ret	0
```

animalize:
If build a 32-bit version with gcc 9.2, it also generates unnecessary machine code (convert to `long long`) as above:
https://godbolt.org/z/qna2eG

64-bit Release version (compiled with VC2017) also converts `ival` to `long long` type. (see below)
Looks like this is a general behavior, maybe we can't assume the compiler will optimize the code.

inlined fuction:
```asm
PyLong_FromLong PROC					; COMDAT

; 317  : {

$LN155:
	push	rdi
	sub	rsp, 32					; 00000020H

; 318  :     PyLongObject *v;
; 319  :     unsigned long abs_ival;
; 320  :     unsigned long t;  /* unsigned so >> doesn't propagate sign bit */
; 321  :     int ndigits = 0;
; 322  :     int sign;
; 323  : 
; 324  :     if (is_small_int(ival)) {

	movsxd	rax, ecx
	xor	edi, edi

; 48   :     return -NSMALLNEGINTS <= ival && ival < NSMALLPOSINTS;

	lea	rdx, QWORD PTR [rax+5]
	cmp	rdx, 261				; 00000105H
	ja	SHORT $LN6@PyLong_Fro

; 60   :     v = (PyObject *)&small_ints[ival + NSMALLNEGINTS];

	shl	rax, 5
	lea	rcx, OFFSET FLAT:small_ints+160
	add	rax, rcx
; File d:\dev\cpython\include\object.h

; 459  :     op->ob_refcnt++;

	inc	QWORD PTR [rax]
; File d:\dev\cpython\objects\longobject.c

; 383  : }

	add	rsp, 32					; 00000020H
	pop	rdi
	ret	0
```
macro:
```asm
PyLong_FromLong PROC					; COMDAT

; 317  : {

$LN150:
	push	rdi
	sub	rsp, 32					; 00000020H

; 318  :     PyLongObject *v;
; 319  :     unsigned long abs_ival;
; 320  :     unsigned long t;  /* unsigned so >> doesn't propagate sign bit */
; 321  :     int ndigits = 0;

	xor	edi, edi

; 322  :     int sign;
; 323  : 
; 324  :     if (IS_SMALL_INT(ival)) {

	lea	eax, DWORD PTR [rcx+5]
	cmp	eax, 261				; 00000105H
	ja	SHORT $LN6@PyLong_Fro

; 60   :     v = (PyObject *)&small_ints[ival + NSMALLNEGINTS];

	movsxd	rax, ecx
	lea	rcx, OFFSET FLAT:small_ints+160
	shl	rax, 5
	add	rax, rcx
; File d:\dev\cpython\include\object.h

; 459  :     op->ob_refcnt++;

	inc	QWORD PTR [rax]
; File d:\dev\cpython\objects\longobject.c

; 383  : }

	add	rsp, 32					; 00000020H
	pop	rdi
	ret	0
```

animalize:
If build a 32-bit version with gcc 9.2, it also generates unnecessary machine code (convert to `long long`) as above:
https://godbolt.org/z/qna2eG

Same with ARM gcc 8.2: https://godbolt.org/z/qycDzR

`long long` appears in the argument of inlined function, maybe we can't ignore it.

animalize:
> VC2017 compiling for x86-64 produces essentially identical code for the macro version and the static-inline-function version

In this case, the `long` -> `long long` converting is still there, but doesn't use too much machine code:
https://godbolt.org/z/-78POZ

> GCC notices they're identical and makes the macro version just jump immediately to the function version 

I know why, on 64-bit Windows, `long` is 4-byte. But on 64-bit GCC, `long` is 8-byte, so `long` and `long long` are the same thing.
https://stackoverflow.com/questions/384502/what-is-the-bit-size-of-long-on-64-bit-windows

animalize:
> No, compilers have a lot more freedom than that. The code just needs to compute the right answer. The types in the source are part of defining what the right answer is, but there doesn't need to be anything at runtime that physically looks like them (in a case like this where they aren't part of any external interface.)

If one day a `int128_t` type and  `PyLong_FromInt128` function are used in the code, converting to `long long` generate wrong result.
So a certain type `long long` is not a good idea.

I have no opinion on the use of inlined function or macro here.
There are no wrong results now, just a little performance loss.

animalize:
It seems that all the major compilers have generated unnecessary code.
It's an unnecessary waste of performance, although very tiny. I created [issue38015](https://bugs.python.org/issue38015) for this.

PR #15221, opened 2019-08-12
bpo-37830: handle specific cases on FOR_ITER

isidentical:
handle specific cases on FOR_ITER, more details at issue

### DIFF
--- a/Lib/test/test_grammar.py
+++ b/Lib/test/test_grammar.py
@@ -969,6 +969,17 @@ class GrammarTests(unittest.TestCase):
             break
         self.assertEqual(count, 1)
 
+
+        def _wrap():
+            for number in range(2):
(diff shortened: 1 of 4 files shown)


isidentical:
Travis fails because of some irrelevant gdb dependicy thing
```
The following packages have unmet dependencies:
 gdb : Depends: libbabeltrace-ctf1 (>= 1.2.1) but it is not installable
       Depends: libbabeltrace1 (>= 1.2.1) but it is not installable
E: Unable to correct problems, you have held broken packages.
apt-get.diagnostics
apt-get install failed
The command "sudo -E apt-get -yq --no-install-suggests --no-install-recommends $(travis_apt_get_options) install gdb xvfb" failed and exited with 100 during .
```

PR #15226, opened 2019-08-12
[3.7] bpo-37642: Update max and min offset in datetime module (GH-14878)

pganssle:
This fixes an inconsistency between the Python and C implementations of
the datetime module. The pure python version of the code was not
accepting offsets greater than 23:59 but less than 24:00. This is an
accidental legacy of the original implementation, which was put in place
before tzinfo allowed sub-minute time zone offsets.

GH-14878

(cherry picked from commit 92c7e30adf5c81a54d6e5e555a6bdfaa60157a0d)

https://bugs.python.org/issue37642

### DIFF
--- a/Lib/datetime.py
+++ b/Lib/datetime.py
@@ -2226,7 +2226,7 @@ class timezone(tzinfo):
         raise TypeError("fromutc() argument must be a datetime instance"
                         " or None")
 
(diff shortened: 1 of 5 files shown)


pganssle:
CC @nsiregar

PR #15231, opened 2019-08-12
bpo-37834: Normalise handling of reparse points on Windows

zooba:
https://bugs.python.org/issue37834

### DIFF
--- a/Modules/_stat.c
+++ b/Modules/_stat.c
@@ -589,6 +589,13 @@ PyInit__stat(void)
     if (PyModule_AddIntMacro(m, FILE_ATTRIBUTE_SYSTEM)) return NULL;
     if (PyModule_AddIntMacro(m, FILE_ATTRIBUTE_TEMPORARY)) return NULL;
     if (PyModule_AddIntMacro(m, FILE_ATTRIBUTE_VIRTUAL)) return NULL;
+
+    if (PyModule_AddObject(m, "IO_REPARSE_TAG_SYMLINK",
+        PyLong_FromUnsignedLong(IO_REPARSE_TAG_SYMLINK))) return NULL;
+    if (PyModule_AddObject(m, "IO_REPARSE_TAG_MOUNT_POINT",
+        PyLong_FromUnsignedLong(IO_REPARSE_TAG_MOUNT_POINT))) return NULL;
+    if (PyModule_AddObject(m, "IO_REPARSE_TAG_APPEXECLINK",
+        PyLong_FromUnsignedLong(IO_REPARSE_TAG_APPEXECLINK))) return NULL;
 #endif
 
     return m;
--- a/Modules/posixmodule.c
+++ b/Modules/posixmodule.c
@@ -1625,6 +1625,7 @@ win32_wchdir(LPCWSTR path)
 */
 #define HAVE_STAT_NSEC 1
 #define HAVE_STRUCT_STAT_ST_FILE_ATTRIBUTES 1
+#define HAVE_STRUCT_STAT_ST_REPARSE_TAG 1
 
 static void
 find_data_to_file_info(WIN32_FIND_DATAW *pFileData,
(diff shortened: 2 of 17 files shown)


Modules/posixmodule.c
```diff
@@ -1695,97 +1696,83 @@ static int
...
-                if (!get_target_path(hFile2, &target_path)) {
-                    CloseHandle(hFile2);
-                    return -1;
-                }
+        /* Get the reparse tag if we will need it */
+        if (info.dwFileAttributes & FILE_ATTRIBUTE_REPARSE_POINT &&
+            !GetFileInformationByHandleEx(hFile, FileAttributeTagInfo,
+                                          &tagInfo, sizeof(tagInfo))) {
+            tagInfo.ReparseTag = 0;
+        }
```
eryksun, line 1790: `follow_symlinks` should only apply to actual symlinks. If `traverse` is false, then `GetFileInformationByHandleEx` : `FileAttributeTagInfo` should be checked before the (relatively expensive) `GetFileInformationByHandle` call, and only there because there's no point in checking it if `traverse` is true.  If it succeeds and `tagInfo.FileAttributes` contains `FILE_ATTRIBUTE_REPARSE_POINT` and the tag is not `IO_REPARSE_TAG_SYMLINK`, we have to close the handle and retstart with `win32_xstat_impl(path, result, TRUE)`.

eryksun: Well, not quite. Done like this it would also have to check here for `IO_REPARSE_TAG_APPEXECLINK`. It would be better to avoid hard coding that.  We need a value for `traverse` that indicates the call is handling `ERROR_CANT_ACCESS_FILE` and so shouldn't try again regardless of the tag. Either that or add another parameter for that case.

Modules/posixmodule.c
```diff
@@ -1695,97 +1696,83 @@ static int
...
-                   the file without the reparse handling flag set. */
-                hFile2 = CreateFileW(
-                           path, FILE_READ_ATTRIBUTES, FILE_SHARE_READ,
-                           NULL, OPEN_EXISTING,
-                           FILE_ATTRIBUTE_NORMAL|FILE_FLAG_BACKUP_SEMANTICS,
-                           NULL);
-                if (hFile2 == INVALID_HANDLE_VALUE)
-                    return -1;
-
-                if (!get_target_path(hFile2, &target_path)) {
```
eryksun, line 1765: `get_target_path` can be removed since it's no longer called.

Modules/posixmodule.c
zooba: We never retry because of the tag here - it is only used to set ST_IFLNK (which is then used to raise an error if we were supposed to follow links, which is the same as the previous behaviour). In `fileutils.c` I'm using the `IsReparseTagNameSurrogate(reparse_tag)` macro to set this flag now, so anything properly registered (which does not include `APPEXECLINK`) will appear as a link.

The `flags |= FILE_FLAG_OPEN_REPARSE_POINT` line higher up is what determines whether we are traversing links or not, and then we'll follow anything that the OS knows how to follow. If we're not traversing, the result is either success or an error - never a retry - so the value for `traverse` that indicates a retry is `FALSE`.

Python/fileutils.c
```diff
@@ -878,7 +878,8 @@ _Py_attribute_data_to_stat(BY_HANDLE_FILE_INFORMATION *info, ULONG reparse_tag,
     FILE_TIME_to_time_t_nsec(&info->ftLastAccessTime, &result->st_atime, &result->st_atime_nsec);
     result->st_nlink = info->nNumberOfLinks;
     result->st_ino = (((uint64_t)info->nFileIndexHigh) << 32) + info->nFileIndexLow;
-    if (reparse_tag == IO_REPARSE_TAG_SYMLINK) {
+    result->st_reparse_tag = reparse_tag;
+    if (IsReparseTagNameSurrogate(reparse_tag)) {
```
eryksun: A name-surrogate reparse point targets another path on the system. In some aspects this overlaps with how we sometimes want to handle any type of 'link' in the broadest sense. For example, in `shutil.rmtree` I would want to simply delete/remove all name surrogates and never follow them. CMD does this for its implementation of `rmdir /s`. 

I would be in favor of adding the ability to open all reparse points to `os.stat`, as a new Windows-only parameter, and adding a function such as `ntpath.issurrogate` and an `is_surrogate` method on the `os.scandir` `DirEntry`. This is useful to know.

But it does not follow and should not follow that `stat` should rubber stamp all name surrogates as `S_IFLNK`. This file type has a particular meaning in POSIX and by extension with how `IO_REPARSE_TAG_SYMLINK` is handled in Windows (e.g. for [`CopyFileExW`](https://docs.microsoft.com/en-us/windows/win32/api/winbase/nf-winbase-copyfileexw) `COPY_FILE_COPY_SYMLINK`) and related technologies such as [DFS Replication](https://docs.microsoft.com/en-us/windows-server/storage/dfs-replication/dfsr-faq): "[t]he only reparse points replicated by DFS Replication are those that use the IO_REPARSE_TAG_SYMLINK tag; however, DFS Replication does not guarantee that the target of a symlink is also replicated".

Most types of reparse points are not name surrogates, and  in this case there should be no question that we should always try to reparse them if we've opened them out of necessity as part of implementing `follow_symlinks=False`. If it fails with `ERROR_CANT_ACCESS_FILE`, then we can just open the reparse point since it's obviously meaningless. Typically reparse points are for cases such as HSM ([hierarchical storage management](https://en.wikipedia.org/wiki/Hierarchical_storage_management)) and SIS ([single-instance storage](https://en.wikipedia.org/wiki/Single-instance_storage)), which should be invisible to the end user. There are dozens of third-party HSM tags assigned in ntifs.h, but I don't know how many of them are being actively used. With HSM, a system might archive files on a large but relatively slow auxiliary volume, and leave a reparse point behind on the primary volume. When the file is accessed, a file-system filter driver fetches the target and replaces the reparse point with the actual file before completing the request. 

In contrast, a name surrogate simply reparses to another path in the system, which in the case of symlink could even be to a remote device (e.g. the MUP device that does UNC provider routing). Thus far the following name surrogates are defined in ntifs.h:

    Microsoft name surrogates

    IO_REPARSE_TAG_MOUNT_POINT                  0xA0000003
    IO_REPARSE_TAG_SYMLINK                      0xA000000C
    IO_REPARSE_TAG_IIS_CACHE                    0xA0000010
    IO_REPARSE_TAG_GLOBAL_REPARSE               0xA0000019
    IO_REPARSE_TAG_LX_SYMLINK                   0xA000001D
    IO_REPARSE_TAG_WCI_TOMBSTONE                0xA000001F
    IO_REPARSE_TAG_PROJFS_TOMBSTONE             0xA0000022

    Non-Microsoft name surrogates

    IO_REPARSE_TAG_SOLUTIONSOFT                 0x2000000D
    IO_REPARSE_TAG_OSR_SAMPLE                   0x20000017
    IO_REPARSE_TAG_QI_TECH_HSM                  0x2000002F
    IO_REPARSE_TAG_MAXISCALE_HSM                0x20000035
    IO_REPARSE_TAG_ALERTBOOT                    0x2000004C
    IO_REPARSE_TAG_NVIDIA_UNIONFS               0x20000054

A "tombstone" is used in [projected file systems](https://docs.microsoft.com/en-us/windows/win32/projfs/projected-file-system) such as VFS for Git to represent a locally deleted file. When enumerating a projected directory, the presence of a tombstone masks the existence of the original file. To the end user, the system behaves as if that file doesn't exist. This is nothing like a symlink, and `follow_symlinks` should have no bearing on this behavior. That breaks the semantics of VFS.

Mount points also should not be conflated with symlinks because they are directories and because the constraints of their creation and their behavior are different. Mount points act a lot like directory hard links. They're implemented as soft links, but the I/O manager goes out of its way to make them act like hard links; the docs tell us that, other than targeting directories and functioning across volumes, ["junctions operate identically to hard links"](https://docs.microsoft.com/en-us/windows/win32/fileio/hard-links-and-junctions). Relative symlinks are reparsed across them exactly so. In contrast, a directory symlink is a soft link that simply resolves to the target path, so relative symlinks are reparsed along the target path.  For example, given "C:/Temp1/Temp2/foo_link" targets "..\\foo", and given "C:/Junction" and "C:/Symlink" both target "\\??\\C:\\Temp1\\Temp2", then "C:/Junction/foo_link" resolves to "C:/foo" and "C:/Symlink/foo_link" resolves to "C:/Temp1/foo".

Also, again because it's like a directory hard link, a mount point for a directory is always local to a machine and always evaluated on the machine. If I'm accessing a mount point "//server/share/spam/windows_junction" that targets "\\??\\C:\\Windows", then it's the Windows directory on "//server/share". OTOH, a soft link is evaluated by the user. If there's a directory symlink "//server/share/spam/windows_symlink" that targets "\\??\\C:\\Windows", then it's the Windows directory on my own system, and this is allowed only if my system's R2L (remote to local) policy is enabled (check `fsutil behavior query symlinkevaluation`).

Calling a mount point a symlink is like mixing up "left" and "right". We have them as separate entities not simply because of some historical accident that NT 5 implemented junctions a few years before symlinks arrived in NT 6. Yet I'm constantly coming across received wisdom that a junction is just some kind of limited, out-moded symlink. This is a bad meme.

Modules/posixmodule.c
eryksun: `traverse` is from `follow_symlinks` and only applies to symlinks. It does not apply to other types of reparse points. If we've opened the reparse point to test for a symlink, we *must* reopen for all other types.

Python/fileutils.c
eryksun: >For example, given "C:/Temp1/Temp2/foo_link" targets "..\foo", and given "C:/Junction" and "C:/Symlink" both target "\??\C:\Temp1\Temp2", then "C:/Junction/foo_link" resolves to "C:/foo" and "C:/Symlink/foo_link" resolves to "C:/Temp1/foo".

To demonstrate the cross-platform implications, here's a version of the symlink/mountpoint scenario in Linux that uses a FUSE bindfs mount point.  "/mnt/temp1/temp2/foo_link" is a relative symlink that targets "../foo".

    /mnt# ln -s /mnt/temp1/temp2 symlink
    /mnt# mkdir mountpoint
    /mnt# bindfs /mnt/temp1/temp2 mountpoint

    /mnt# python3 -c 'import os; print(os.path.ismount("symlink"))'
    False
    /mnt# python3 -c 'import os; print(os.path.islink("symlink"))'
    True
    /mnt# python3 -c 'import os; print(os.readlink("symlink"))'
    /mnt/temp1/temp2
    
    /mnt# python3 -c 'import os; print(os.path.ismount("mountpoint"))'
    True
    /mnt# python3 -c 'import os; print(os.path.islink("mountpoint"))'
    False

    /mnt# python3 -c 'import os; print(os.readlink("mountpoint"))'
      File "<string>", line 1, in <module>
    OSError: [Errno 22] Invalid argument: 'mountpoint'

    /mnt# cat temp1/foo
    temp1 foo
    /mnt# cat ./foo
    mnt foo

    /mnt# readlink temp1/temp2/foo_link
    ../foo

    /mnt# cat symlink/foo_link
    temp1 foo
    /mnt# cat mountpoint/foo_link
    mnt foo

Modules/_stat.c
```diff
@@ -589,6 +589,13 @@ PyInit__stat(void)
     if (PyModule_AddIntMacro(m, FILE_ATTRIBUTE_SYSTEM)) return NULL;
     if (PyModule_AddIntMacro(m, FILE_ATTRIBUTE_TEMPORARY)) return NULL;
     if (PyModule_AddIntMacro(m, FILE_ATTRIBUTE_VIRTUAL)) return NULL;
+
+    if (PyModule_AddObject(m, "IO_REPARSE_TAG_SYMLINK",
+        PyLong_FromUnsignedLong(IO_REPARSE_TAG_SYMLINK))) return NULL;
+    if (PyModule_AddObject(m, "IO_REPARSE_TAG_MOUNT_POINT",
+        PyLong_FromUnsignedLong(IO_REPARSE_TAG_MOUNT_POINT))) return NULL;
+    if (PyModule_AddObject(m, "IO_REPARSE_TAG_APPEXECLINK",
+        PyLong_FromUnsignedLong(IO_REPARSE_TAG_APPEXECLINK))) return NULL;
```
eryksun, line 598: `IO_REPARSE_TAG_UNHANDLED` (0x80000020), going by its name, is always an unhandled tag. `IO_REPARSE_TAG_LX_SYMLINK` (0xA000001D) should be an unhandled link tag, since it's used for symlinks between an NT file system and a WSL file system (e.g. "/mnt/c/etc" -> "/etc"). 

I looked over Microsoft's tags for a *handled* tag that's not a link, but nothing stands out. I thought that OneDrive used them (i.e. `IO_REPARSE_TAG_CLOUD*`), but that doesn't appear to be the case anymore. I also thought the new Projected File System support would use them (i.e. `IO_REPARSE_TAG_PROJFS`). I built the [RegFS sample](https://github.com/Microsoft/Windows-classic-samples/tree/master/Samples/ProjectedFileSystem) to try to confirm this, and I was surprised to find that none of the virtual directories and  files (keys and values) was a reparse point.

However, both OneDrive and ProjFS do make extensive use of the following new file attributes:

```
FILE_ATTRIBUTE_PINNED  0x00080000
FILE_ATTRIBUTE_UNPINNED  0x00100000
FILE_ATTRIBUTE_RECALL_ON_OPEN  0x00040000
FILE_ATTRIBUTE_RECALL_ON_DATA_ACCESS  0x00400000
```

PR #15232, opened 2019-08-12
bpo-16637: libpython: interpret string as integer literal

marc1006:
This fixes the exception `ValueError: invalid literal for int() with base 10`
if `str(gdbval)` returns a hexadecimal value (e.g. '0xa0').

See https://docs.python.org/3/library/functions.html#int for more information.

https://bugs.python.org/issue16637

### DIFF
--- a/Tools/gdb/libpython.py
+++ b/Tools/gdb/libpython.py
@@ -1392,7 +1392,7 @@ class wrapperobject(PyObjectPtr):
 
 
 def int_from_int(gdbval):
-    return int(str(gdbval))
+    return int(gdbval)
 
 
 def stringify(val):


the-knights-who-say-ni:

Tools/gdb/libpython.py
```diff
@@ -1392,7 +1392,7 @@ def write_repr(self, out, visited):
 
 
 def int_from_int(gdbval):
-    return int(str(gdbval))
+    return int(str(gdbval), 0)
```
benjaminp: Can one not simply call `int` on integer-type GDB values?

marc1006: You mean something like

```suggestion
    return int(gdbval.cast(gdb.lookup_type('long')))
```
?

benjaminp: Yes, but do you have to do the `cast`?

marc1006: I think so - but to be sure I have to verify it. Will update my branch if it works.

marc1006: I've updated my branch. Seems a simple `int` call is enough - thanks (I assume this only holds if the gdbvalue has the right type).

PR #15234, opened 2019-08-12
Minor documentation fixes on library/enum

awecx:
This PR fixes 3 minor issues in library/enum doc:

- l.386: extra "were" that make the sentence incorrect.
- l.732: I believe this is a bad example.
- l.1056 typo: we're referring here to _list_ and _str_ as objects, not callables.

### DIFF
--- a/Doc/library/enum.rst
+++ b/Doc/library/enum.rst
@@ -383,8 +383,8 @@ enumeration, with the exception of special methods (:meth:`__str__`,
 variable names listed in :attr:`_ignore_`.
 
 Note:  if your enumeration defines :meth:`__new__` and/or :meth:`__init__` then
-whatever value(s) were given to the enum member will be passed into those
-methods.  See `Planet`_ for an example.
+whatever value(s) given to the enum member will be passed into those methods.
+See `Planet`_ for an example.
 
 
 Restricted Enum subclassing
@@ -730,8 +730,7 @@ Some rules:
 2. While :class:`Enum` can have members of any type, once you mix in an
    additional type, all the members must have values of that type, e.g.
    :class:`int` above.  This restriction does not apply to mix-ins which only
-   add methods and don't specify another data type such as :class:`int` or
-   :class:`str`.
+   add methods and don't specify another data type.
 3. When another data type is mixed in, the :attr:`value` attribute is *not the
    same* as the enum member itself, although it is equivalent and will compare
    equal.
@@ -1054,7 +1053,7 @@ Supported ``_sunder_`` names
 
 - ``_missing_`` -- a lookup function used when a value is not found; may be
   overridden
-- ``_ignore_`` -- a list of names, either as a :func:`list` or a :func:`str`,
+- ``_ignore_`` -- a list of names, either as a :class:`list` or a :class:`str`,
   that will not be transformed into members, and will be removed from the final
   class
 - ``_order_`` -- used in Python 2/3 code to ensure member order is consistent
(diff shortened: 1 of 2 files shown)


tirkarthi:
cc: @ethanfurman

Doc/library/enum.rst
```diff
@@ -383,8 +383,8 @@ enumeration, with the exception of special methods (:meth:`__str__`,
 variable names listed in :attr:`_ignore_`.
 
 Note:  if your enumeration defines :meth:`__new__` and/or :meth:`__init__` then
-whatever value(s) were given to the enum member will be passed into those
-methods.  See `Planet`_ for an example.
+whatever value(s) given to the enum member will be passed into those methods.
```
ethanfurman: "were" is necessary in the original sentence.  However, if you also change "whatever" to "any" (while still removing "were") I think we'll have a better sentence.

Doc/library/enum.rst
```diff
@@ -730,8 +730,7 @@ Some rules:
 2. While :class:`Enum` can have members of any type, once you mix in an
    additional type, all the members must have values of that type, e.g.
    :class:`int` above.  This restriction does not apply to mix-ins which only
-   add methods and don't specify another data type such as :class:`int` or
-   :class:`str`.
+   add methods and don't specify another data type.
```
ethanfurman: Instead of removing the `int` and `str` clause, make it a "for example" clause; i.e.:
```
... another data type; for example :class:`int` or :class:`str`.
```

Doc/library/enum.rst
```diff
@@ -1054,7 +1053,7 @@ Supported ``_sunder_`` names
 
 - ``_missing_`` -- a lookup function used when a value is not found; may be
   overridden
-- ``_ignore_`` -- a list of names, either as a :func:`list` or a :func:`str`,
+- ``_ignore_`` -- a list of names, either as a :class:`list` or a :class:`str`,
```
ethanfurman, line 1056: Good catch, thank you!

ethanfurman (changes requested):

Doc/library/enum.rst
awecx: Actually I believe there is a mistake here.

If I understand correctly, once you mix a type that specifies a data type (_int_, _float_ or whatever), all members must be of that type. There is no such restriction for mixins without constructor that only add "pure methods", which is NOT the case for _int_ and _str_. See https://pastebin.com/qeb4npZw for example.

And the doc seems to say that _int_ and _str_ only add methods without defining types, which, I believe, is incorrect ; hence I've removed the example.

Doc/library/enum.rst
awecx: Agreed, will change.

Doc/library/enum.rst
ethanfurman: Ah, I see it now.  The extra (supposed to be) clarifying example could actually be confusing instead.

Okay, keep this change as-is.

awecx:
As discussed, I've updated the sentence "any value(s) given to the enum member will be passed into those methods" and left the removal of the example in the mixin section.

ethanfurman (approved):

ethanfurman:
@awecx, Is your name in the Misc/ACKS file?  If you want it to be, please add that commit as well.

awecx:
> @awecx, Is your name in the Misc/ACKS file? If you want it to be, please add that commit as well.

No, it isn't.
I've just added myself with the last commit.

Doc/library/enum.rst
ethanfurman: Let's remove the word `data` -- so `... specify another type`.

ethanfurman (changes requested):
Apologies, one more quick change.

awecx:
I've applied your last suggestion.

ethanfurman (approved):

ethanfurman:
@awecx - Thank you for your patience and perseverance!

awecx:
> @awecx - Thank you for your patience and perseverance!

Thank YOU for advice and review of this (my first!) PR. I really appreciated.

PR #15237, opened 2019-08-12
Remove versioned executables from non-APPX packages

zooba:


zooba:
These should never have been in here anyway.

PR #15242, opened 2019-08-13
gh-59215: unittest: _top_level_dir is incorrectly persisted

ZackerySpytz:
Co-Authored-By: Igor Sobreira <igor@igorsobreira.com>

https://bugs.python.org/issue15010

* Issue: gh-59215


Lib/unittest/test/test_discovery.py
```diff
@@ -406,10 +406,34 @@ def _find_tests(start_dir, pattern, namespace=None):
...
+        original_isfile = os.path.isfile
+        original_isdir = os.path.isdir
+        original_sys_path = sys.path[:]
+        def restore():
+            os.path.isfile = original_isfile
+            os.path.isdir = original_isdir
+            sys.path[:] = original_sys_path
+        self.addCleanup(restore)
+
+        os.path.isfile = lambda path: True
```
tirkarthi: Is `unittest.mock.patch` a better option here instead of `addCleanup`?

```python
def test_discover_should_not_persist_top_level_dir_between_calls(self):
    with unittest.mock.patch('os.path.isfile', return_value=True):
        with unittest.mock.patch('os.path.isdir', return_value=True):
            loader = unittest.TestLoader()
            loader.suiteClass = str
            dir = '/foo/bar'
            top_level_dir = '/foo'

            loader.discover(dir, top_level_dir=top_level_dir)
            self.assertEqual(loader._top_level_dir, None)

            loader._top_level_dir = dir2 = '/previous/dir'
            loader.discover(dir, top_level_dir=top_level_dir)
            self.assertEqual(loader._top_level_dir, dir2)
```

encukou: Nowadays that can be `self.enterContext`. But, as is, the PR is consistent with other nearby tests; if this needs changing it would be best to change all of them (in a separate PR).

encukou (commented):
IMO, this is the right solution, even if `_top_level_dir` isn't perfect.
However, as evidenced by the changed test, this is a change in behaviour. IMO, it needs at least a `versionchanged` note in the docs, saying that `top_level_dir` is only stored for the duration of the `discover` call.

I've sent proposed docs changes as a PR to this PR, here: https://github.com/ZackerySpytz/cpython/pull/39/files [now merged here]
If an `unittest` expert has time to look over them, that would be great :)

PR #15244, opened 2019-08-13
bpo-21109: Add SafeTarFile

epicfaace:
Working off of the patch in [bpo-21109](https://bugs.python.org/issue21109); adding the tests so that all tests that apply to FileTest also apply to SafeFileTest. It's a bit tricky in this case because of multiple inheritance of the test classes.

https://bugs.python.org/issue21109


encukou:
Superseded by PEP-706. A safe(r) option is now the default.
Thank you for this attempt!

PR #15208, opened 2019-08-11
bpo-37814: Document the empty tuple type annotation syntax

keysmashes:
https://bugs.python.org/issue37814:

> The empty tuple syntax in type annotations, `Tuple[()]`, is not obvious from the examples given in the documentation (I naively expected `Tuple[]` to work); it has been documented in PEP 484 and in mypy, but not in the documentation for the typing module.

https://bugs.python.org/issue37814

Automerge-Triggered-By: @gvanrossum

PR #15262, opened 2019-08-13
[3.7] bpo-37814: Document the empty tuple type annotation syntax (GH-15208)

keysmashes:
https://bugs.python.org/issue37814:

> The empty tuple syntax in type annotations, `Tuple[()]`, is not obvious from the examples given in the documentation (I naively expected `Tuple[]` to work); it has been documented in PEP 484 and in mypy, but not in the documentation for the typing module.

https://bugs.python.org/issue37814
(cherry picked from commit 8a784af750fa82c8355903309e5089eb2b60c16b)

Co-authored-by: Josh Holland <anowlcalledjosh@gmail.com>

https://bugs.python.org/issue37814

### DIFF
--- a/Doc/library/typing.rst
+++ b/Doc/library/typing.rst
@@ -1056,7 +1056,8 @@ The module defines the following classes, functions and decorators:
 .. data:: Tuple
 
    Tuple type; ``Tuple[X, Y]`` is the type of a tuple of two items
-   with the first item of type X and the second of type Y.
+   with the first item of type X and the second of type Y. The type of
+   the empty tuple can be written as ``Tuple[()]``.
 
    Example: ``Tuple[T1, T2]`` is a tuple of two elements corresponding
    to type variables T1 and T2.  ``Tuple[int, float, str]`` is a tuple


gvanrossum (approved):

anowlcalledjosh:
Maybe worth noting that I ran into some problem with cherry_picker while submitting this:

```
Error: Run state cherry-picker.state=BACKPORT_LOOP_START in Git config is not known.
Perhaps it has been set by a newer version of cherry-picker. Try upgrading.
Valid states are: BACKPORT_PAUSED, UNSET. If this looks suspicious, raise an issue at https://github.com/python/core-workflow/issues/new.
As the last resort you can reset the runtime state stored in Git config using the following command: `git config --local --remove-section cherry-picker`
```

but I recloned the repository and it seemed to work fine on the second try.

gvanrossum:
Thanks! Hopefully it will auto-merge once the tests pass.

gvanrossum:
@mariatta, @brettcannon -- Any idea what could be the matter with cherry_picker?
https://github.com/python/cpython/pull/15262#issuecomment-520972508

Mariatta:
> Error: Run state cherry-picker.state=BACKPORT_LOOP_START in Git config is not known.

Perhaps you were in the middle of other backport process. It should be cleared if you run `cherry_picker --abort`.

Other issues about cherry_picker can be filed in https://github.com/python/cherry-picker

PR #15264, opened 2019-08-14
bpo-29535: make datetime hashes non-deterministic

epicfaace:
This PR does the following:
- Make the hash of datetime.datetime (with a timezone) non-deterministic in the Python datetime lib.
- Make the hash of datetime.datetime (with a timezone) non-deterministic in the C datetime lib.
- Make the hash of datetime.datetime (with a timezone) non-deterministic in the C datetime lib.
- Make the hash of datetime.time (with a timezone such that either _1) time - timezone < 0 [underflow] or 2) time - timezone > max_time [overflow]_ non-deterministic in the Python datetime lib.
- Make sure datetime tests in test_hash.py cover both the C datetime lib and the Python datetime lib (previously they covered only the C datetime lib), and cover the cases above.

https://bugs.python.org/issue29535

### DIFF
--- a/Lib/datetime.py
+++ b/Lib/datetime.py
@@ -2115,7 +2115,12 @@ class datetime(date):
             if tzoff is None:
                 self._hashcode = hash(t._getstate()[0])
             else:
-                self._hashcode = hash((t - tzoff)._getstate()[0])
+                try:
+                    t_with_offset = t - tzoff
+                except OverflowError:
+                    # t - tzoff is less than 0, which can't be represented as a datetime, so we use t + tzoff instead.
+                    t_with_offset = t + tzoff
+                self._hashcode = hash(t_with_offset._getstate()[0])
         return self._hashcode
 
     # Pickle support.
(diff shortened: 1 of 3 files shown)


Lib/datetime.py
```diff
@@ -1379,10 +1379,11 @@ def __hash__(self):
...
                 assert not m % timedelta(minutes=1), "whole minute"
                 m //= timedelta(minutes=1)
-                if 0 <= h < 24:
-                    self._hashcode = hash(time(h, m, self.second, self.microsecond))
-                else:
-                    self._hashcode = hash((h, m, self.second, self.microsecond))
+                if h < 0:
+                    h = -1 * h
+                if h >= 24:
+                    h = h - 24
```
brandtbucher, line 1385: These four lines seem a tad heavy (and odd). Would a simple `h %= 24` work here?

epicfaace:
Closing this as per discussion in bpo-29535.

PR #15265, opened 2019-08-14
bpo-37760: Convert from length-18 lists to a dataclass, in makeunicodedata.

gnprice:
Now the fields have names!  Much easier to keep straight as a
reader than 18 distinct indexes into a list.

Runs about 10-15% slower: from 10.8s to 12.3s, on my laptop.
Fortunately that's perfectly fine for this maintenance script.

### DIFF
--- a/Tools/unicode/makeunicodedata.py
+++ b/Tools/unicode/makeunicodedata.py
@@ -31,7 +31,6 @@ import os
 import sys
 import zipfile
 
-from dataclasses import dataclass
 from functools import partial
 from textwrap import dedent
 from typing import Iterator, List, Optional, Set, Tuple
@@ -947,7 +946,7 @@ class UcdFile:
                 yield char, rest
 
 
-@dataclass
+@dataclasses.dataclass
 class UcdRecord:
     # 15 fields from UnicodeData.txt .  See:
     #   https://www.unicode.org/reports/tr44/#UnicodeData.txt
--- a/Tools/unicode/makeunicodedata.py
+++ b/Tools/unicode/makeunicodedata.py
@@ -26,13 +26,15 @@
 # written by Fredrik Lundh (fredrik@pythonware.com)
 #
 
+import dataclasses
 import os
 import sys
 import zipfile
 
+from dataclasses import dataclass
 from functools import partial
 from textwrap import dedent
-from typing import Iterator, List, Tuple
+from typing import Iterator, List, Optional, Set, Tuple
 
 SCRIPT = sys.argv[0]
 VERSION = "3.3"
@@ -148,12 +150,12 @@ def makeunicodedata(unicode, trace):
         record = unicode.table[char]
         if record:
             # extract database properties
-            category = CATEGORY_NAMES.index(record[2])
-            combining = int(record[3])
-            bidirectional = BIDIRECTIONAL_NAMES.index(record[4])
-            mirrored = record[9] == "Y"
-            eastasianwidth = EASTASIANWIDTH_NAMES.index(record[15])
-            normalizationquickcheck = record[17]
+            category = CATEGORY_NAMES.index(record.general_category)
+            combining = int(record.canonical_combining_class)
+            bidirectional = BIDIRECTIONAL_NAMES.index(record.bidi_class)
+            mirrored = record.bidi_mirrored == "Y"
+            eastasianwidth = EASTASIANWIDTH_NAMES.index(record.east_asian_width)
+            normalizationquickcheck = record.quick_check
             item = (
                 category, combining, bidirectional, mirrored, eastasianwidth,
                 normalizationquickcheck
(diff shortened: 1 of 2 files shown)


Tools/unicode/makeunicodedata.py
```diff
@@ -979,19 +1012,17 @@ def __init__(self, version, cjk_check=True):
...
             s = table[i]
             if s:
-                if s[1][-6:] == "First>":
-                    s[1] = ""
-                    field = s
-                elif s[1][-5:] == "Last>":
-                    if s[1].startswith("<CJK Ideograph"):
+                if s.name[-6:] == "First>":
+                    s.name = ""
+                    field = dataclasses.astuple(s)[:15]
```
benjaminp, line 1016: Is the sole reason for this tuple conversion line 1025? Can you use the `replace` method of dataclasses there? The hardcoded [:15] somewhat defeats the nice benefit of naming fields you've achieved here.

(Also, I wonder if the first two loops of this function can be merged but that's orthogonal.)

Tools/unicode/makeunicodedata.py
```diff
@@ -26,13 +26,15 @@
 # written by Fredrik Lundh (fredrik@pythonware.com)
 #
 
+import dataclasses
 import os
 import sys
 import zipfile
 
+from dataclasses import dataclass
```
benjaminp: Can we just qualify with `dataclass.` everywhere? (seems this is already mostly happening aside from the decorator application)

Tools/unicode/makeunicodedata.py
```diff
@@ -949,25 +947,60 @@ def expanded(self) -> Iterator[Tuple[int, List[str]]]:
...
 
 # load a unicode-data file from disk
 
 class UnicodeData:
-    # Record structure:
-    # [ID, name, category, combining, bidi, decomp,  (6)
-    #  decimal, digit, numeric, bidi-mirrored, Unicode-1-name, (11)
-    #  ISO-comment, uppercase, lowercase, titlecase, ea-width, (16)
-    #  derived-props] (17)
+    # table: Dict[int, UcdRecord]  # key is codepoint
```
benjaminp: The right type is actually `List[Optional[UcdRecord]]`, right?

benjaminp (changes requested):
re commit message: Wasn't each record actually a heterogeneous list?

Tools/unicode/makeunicodedata.py
gnprice: Sure, WFM.

Tools/unicode/makeunicodedata.py
gnprice: ... So it is! Good catch, thanks.

Certainly this illustrates the danger in type annotations that aren't checked. :wink:  At least it was commented, so if it'd gotten committed this way the reader would have had *some* warning to take it skeptically.

Tools/unicode/makeunicodedata.py
gnprice: > Is the sole reason for this tuple conversion line 1025? Can you use the `replace` method of dataclasses there? The hardcoded [:15] somewhat defeats the nice benefit of naming fields you've achieved here.

Yeah. I think I originally did this with `replace` and that was too slow. Let's see if that reproduces...

Oh right, another bit is that we need a distinct `set` for the `binary_properties` attribute of each record, so the code ends up not quite as clean as you might hope anyway. In this version `from_row` is the place that knows that fact, and we use it in both these loops.

OK, the result is it makes the whole script about 20% slower: on my (fast) desktop, it's 8.3s in this version and 10.0s with `.replace`. Here's the patch:
```
             if s:
                 if s.name[-6:] == "First>":
                     s.name = ""
-                    field = dataclasses.astuple(s)[:15]
+                    field = s
                 elif s.name[-5:] == "Last>":
                     if s.name.startswith("<CJK Ideograph"):
-                        cjk_ranges_found.append((field[0],
+                        cjk_ranges_found.append((field.codepoint,
                                                  s.codepoint))
                     s.name = ""
                     field = None
             elif field:
-                table[i] = from_row(('%X' % i,) + field[1:])
+                table[i] = dataclasses.replace(field,
+                    codepoint='%X' % (i,), binary_properties=set())
```

Certainly cleaner code that way, but the 20% speed hit (~2 extra seconds) is noticeably painful when doing development on this script and consequently running it all the time to try changes.

This tuple is quite local -- it doesn't escape this loop of 20 lines or so -- so I'd prefer to accept the short stretch of lower-level fiddling in exchange for the speedup when working on the rest of the script.

gnprice (commented):
Thanks @benjaminp for the review!

Just pushed an update with all these changes except to the `astuple` bit; details below.

> Wasn't each record actually a heterogeneous list?

Hmm true! I guess logically I think of that as a "tuple", but it's certainly not a Python tuple. Updated the NEWS entry, and I'll edit the PR description/title too.

Tools/unicode/makeunicodedata.py
benjaminp: Okay, I suppose `[:15]` is unlikely to break too in the future given the stability of `UnicodeData.txt`.

gnprice: Agreed :)

PR #15266, opened 2019-08-14
bpo-37798: Add C fastpath for statistics.NormalDist.inv_cdf()

corona10:


Lib/statistics.py
```diff
@@ -111,7 +111,7 @@
 from math import hypot, sqrt, fabs, exp, erf, tau, log, fsum
 from operator import itemgetter
 from collections import Counter
-
+from _statistics import _normal_dist_inv_cdf
```
tirkarthi: Just curious how does this work in environments where C accelerator might not be present? Is it better to have the Python definition in `statistics.py` as it is and then try to import `_statistics` to override `_normal_dist_inv_cdf` if available?

corona10: @tirkarthi Great idea, ~by the way, do you know what I should do for windows platform when importing C module? I have only MacBook so I can't check it works well on windows.~ solved

cc @rhettinger @vstinner

corona10: @tirkarthi 
IMHO, the current C implementation does not have a specific platform-dependent code.
So I think that it will work most of the platform which can compile `C`Python.
What do you think?

tirkarthi: Modules like datetime have C implementation and a pure Python implementation. I have seen PyPy using pure Python implementation of datetime. Mine was just a concern that do we need to make `statistics.py` dependent on C accelerator always for this helper. My concern might be wrong or I am misunderstanding other implementations shipping C code. I will wait for a core dev input on this.

rhettinger:
The usual way we do this is to define the function in pure python.  Then, somewhere below that definition, try to import the C accelerator function in its place.  For examples, see line 458 in *Lib/collections/\_\_init\_\_.py* for *_count_elements()* or line 579 in *Lib/heapq.py*

corona10:
@rhettinger  

Thanks a lot.
I 've updated the PR as you suggested.
I both test when using pure python version and C version.
And it looks same on both versions.
Please take a look.

Lib/statistics.py
```diff
@@ -111,7 +111,6 @@
 from math import hypot, sqrt, fabs, exp, erf, tau, log, fsum
 from operator import itemgetter
 from collections import Counter
-
```
rhettinger: Please add back the spurious whitespace change.

corona10: Resolved!

Modules/_statisticsmodule.c
```diff
@@ -0,0 +1,125 @@
...
+    {NULL, NULL, 0, NULL}
+};
+
+
+static PyObject *
+normal_dist_inv_cdf(PyObject *self, PyObject *args)
+{
+    double p, mu, sigma, q;
+    double num, den, r, x;
+    if (!PyArg_ParseTuple(args, "d|d|d", &p, &mu, &sigma))
```
rhettinger: Try using the ArgumentClinic on this.  It will likely make the function call even faster.

Here's my timings on the current version.

```
--baseline--
$ ./python.exe -m timeit -r11 -s 'from statistics import NormalDist as ND, quantiles as q' -s 'Z=ND()' 'q(Z, n=100)'
2000 loops, best of 11: 111 usec per loop

-- patched --
$ ./python.exe -m timeit -r11 -s 'from statistics import NormalDist as ND, quantiles as q' -s 'Z=ND()' 'q(Z, n=100)'
5000 loops, best of 11: 44 usec per loop
```

rhettinger (changes requested):

Modules/_statisticsmodule.c
corona10: As you suggested ArgumentClinic make better performance than I wrote.

Thanks!
````
5000 loops, best of 11: 40.6 usec per loop
````

**Comment by corona10** [comment]

**Comment by corona10** [comment]
@rhettinger Can you please take a look?

File: Lib/statistics.py [hunk 4]
```diff
@@ -695,6 +695,80 @@ def _ss(data, c=None):
     return (T, total)
 
 
+def _normal_dist_inv_cdf(p, mu, sigma):
```
rhettinger: Minor nit:  Let's move this downward, just below the `## Normal Distribution #####################################################` section divider.

Modules/_statisticsmodule.c
```diff
@@ -0,0 +1,120 @@
...
+
+
+static PyMethodDef speedups_methods[] = {
+    _STATISTICS__NORMAL_DIST_INV_CDF_METHODDEF
+    {NULL, NULL, 0, NULL}
+};
+
+/*[clinic input]
+_statistics._normal_dist_inv_cdf -> double
+   p: 'd'
```
rhettinger: I haven't seen the `'d'` form before.  Is there a reason this isn't `double` like I'm seeing elsewhere?

Modules/_statisticsmodule.c
```diff
@@ -0,0 +1,120 @@
...
+static double
+_statistics__normal_dist_inv_cdf_impl(PyObject *module, double p, double mu,
+                                      double sigma)
+/*[clinic end generated code: output=02fd19ddaab36602 input=74c459adf13062f9]*/
+{
+    double q, num, den, r, x;
+    q = p - 0.5;
+    if(fabs(q) <= 0.425) {
+        r = 0.180625 - q * q;
+        // Hash sum: 55.8831928806149014439
```
rhettinger: I presume this is all correct, but will you double check the hash sums still add up.

Modules/_statisticsmodule.c
```diff
@@ -0,0 +1,120 @@
...
+    _STATISTICS__NORMAL_DIST_INV_CDF_METHODDEF
+    {NULL, NULL, 0, NULL}
+};
+
+/*[clinic input]
+_statistics._normal_dist_inv_cdf -> double
+   p: 'd'
+   mu: 'd'
+   sigma: 'd'
+[clinic start generated code]*/
```
rhettinger, line 25: Let's add a trailing `/` to mark all of this arguments as positional only.  There's no point for keyword arguments in an internal helper function.

rhettinger (changes requested):

rhettinger:
Thanks for doing this.  I think we're getting close :-)

corona10:
Thanks for the review, I will take look at!

Modules/_statisticsmodule.c
corona10: I've read the paper http://csg.sph.umich.edu/abecasis/gas_power_calculator/algorithm-as-241-the-percentage-points-of-the-normal-distribution.pdf. And it looks like numbers are correct.

corona10:
@rhettinger I've updated the PR!

rhettinger (approved):

rhettinger:
Nice work.  Thanks for doing this :-)

PR #15267, opened 2019-08-14
bpo-37849: IDLE: fix completion window positioning above line

taleinat:
This also adds a mechanism to avoid multiple, recursive invocations of the "\<Configure\>" event handler, required due to the added `.update()` call.

https://bugs.python.org/issue37849

### DIFF
--- a/Lib/idlelib/NEWS.txt
+++ b/Lib/idlelib/NEWS.txt
@@ -3,6 +3,11 @@ Released on 2019-10-20?
 ======================================
 
 
+bpo-37849: Fix completions list appearing too high or low when shown
+above the current line.
+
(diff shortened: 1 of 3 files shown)


terryjreedy:
New news items go on top, with the fixed length divider line and *2* blank lines above and a blank line and the previous item below.  As long as all are backported, all will backport without conflict in either the upper or lower 3-line context.  In this case, there was a previous item to be added also.

For the first news addition *after* a final release, which requires a new header at least on that version, the lower context will be different on different versions.  So I do the addition on separate PRs (master and manual backports as needed) attached to the current next-version NEWS issue (now bpo 34162 for 3.8).  I also use that for corrections and batch additions.

PR #15274, opened 2019-08-14
bpo-38094: Remove unneeded assignment in PyBytes_Concat()

sir-sigurd:
It looks unneeded since its introduction in 161d695fb0455ce52530d4f43a9eac4c738f64bb

https://bugs.python.org/issue38094

### DIFF
--- a/Objects/bytesobject.c
+++ b/Objects/bytesobject.c
@@ -2938,7 +2938,6 @@ PyBytes_Concat(PyObject **pv, PyObject *w)
         Py_ssize_t oldsize;
         Py_buffer wb;
 
-        wb.len = -1;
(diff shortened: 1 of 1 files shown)


epicfaace (commented):
Isn't it used here: https://github.com/python/cpython/blob/6d3905b19908eabade87676cbeaf092b53355493/Objects/bytesobject.c#L2949

sir-sigurd:
When `wb.len` is used it's set by [`PyObject_GetBuffer()`](https://docs.python.org/3/c-api/buffer.html#c.PyObject_GetBuffer).

gpshead (approved):

gpshead:
closing and reopening to re-trigger bedevere

PR #15275, opened 2019-08-14
bpo-34311: Add locale.localize

cedk:
This allow to format Decimal without losing precision by using the new
format syntax and than localize the result.

https://bugs.python.org/issue34311


Misc/NEWS.d/next/Library/2019-08-14-13-19-50.bpo-33731.9esS0d.rst
```diff
@@ -0,0 +1,2 @@
+Provide a locale.localize() function which convert a normalized number
```
epicfaace:
```suggestion
Provide a locale.localize() function, which converts a normalized number
```

Lib/test/test_locale.py
```diff
@@ -594,5 +594,30 @@ def test_atoi(self):
...
+
+    def test_localize(self):
+        self._test_localize('50000.00', '50000.00')
+
+
+class TestfrFRLocalize(FrFRCookedTest, BaseLocalizeTest):
+
+    def test_localize(self):
+        self._test_localize('50000.00', '50000,00')
+        self._test_localize('50000.00', '50 000,00', grouping=True)
```
epicfaace, line 658: Can you add a test in which you use locale.localize to format a Decimal (a Decimal which, if passed through locale.format_string, would lose precision)?

Lib/locale.py
```diff
@@ -185,8 +185,13 @@ def _format(percent, value, grouping=False, monetary=False, *additional):
         formatted = percent % ((value,) + additional)
     else:
         formatted = percent % value
+    if percent[-1] in 'eEfFgGdiu':
+        formatted = _localize(formatted, grouping, monetary)
+    return formatted
+
+def _localize(formatted, grouping=False, monetary=False):
```
epicfaace, line 193: Can you add a comment explaining what _localize does?

Lib/locale.py
```diff
@@ -185,8 +185,14 @@ def _format(percent, value, grouping=False, monetary=False, *additional):
...
+    # Transform formatted as locale number according to the locale settings
```
epicfaace: Can you enclose the comment around triple double quotes so it becomes a docstring (https://www.python.org/dev/peps/pep-0257/)?

Lib/locale.py
```diff
@@ -321,6 +327,11 @@ def delocalize(string):
         string = string.replace(dd, '.')
     return string
 
+def localize(string, grouping=False, monetary=False):
+    "Parses a string as locale number according to the locale settings"
```
epicfaace: Also make this a docstring

epicfaace (changes requested):

Lib/locale.py
cedk: It is kind of pointless to make docstring for private method.

Lib/locale.py
cedk: I do not understand, it is already a docstring.

epicfaace: I believe docstrings are supposed to be triple-double-quoted. See https://www.python.org/dev/peps/pep-0257/#one-line-docstrings

Lib/locale.py
epicfaace: Yeah, you're right, it doesn't matter here whether we have a docstring.

Lib/locale.py
cedk: Well, I follow style of this file.

epicfaace: Seems like the majority of the functions in this file have triple-double-quoted docstrings; the single double quotes in `delocalize`, `atof`, `atob` seem to be a mistake.

Lib/test/test_locale.py
```diff
@@ -594,5 +595,32 @@ def test_atoi(self):
...
+
+
```
matrixise, line 660: There is a missing test with `self._test_localize('abcdefg', '5000.0')`
Thanks

matrixise (changes requested):

cedk:
ping

cedk:
ping

cedk:
ping

cedk:
I really would like to see this landed in Python so I could apply a backport to fix https://bugs.tryton.org/issue8574

Lib/locale.py
```diff
@@ -321,6 +327,11 @@ def delocalize(string):
         string = string.replace(dd, '.')
     return string
 
+def localize(string, grouping=False, monetary=False):
+    """Parses a string as locale number according to the locale settings."""
+    float(string)
```
bharel: What's the float(string) for?

cedk: Ensure that it is a string that represent a valid number.

aeros: @cedk 

> Ensure that it is a string that represent a valid number.

Hmm, I'm not sure that I agree with this means of validating the strings. Primarily because it will result in a non-obvious error message for users. For example, `localize('')` would result in:

```
ValueError: could not convert string to float: ''
```

I suspect that this message would make zero sense to users unless they looked at the implementation of the function, which should not be a requirement for diagnosing error messages.

IMO, you should at the very least catch the `ValueError` and then re-raise it with a more informative error message. For example:

```suggestion
    try:
        float(string)
    except ValueError:
        raise ValueError(f"{string} is not a valid number")
```

Edit: It may also be worth considering that there might be some valid floats that you wouldn't consider to be valid for this function, such as ``float('NAN')``, ``float('INF')``, and ``float('-INF')``. For more thorough validation, it could be worthwhile to use regex instead.

cedk: Indeed I added this for https://github.com/python/cpython/pull/15275#pullrequestreview-287111359. But as delocalize does not perform any check and passing a non valid number will not fail, I think it is better to relax the check.

aeros: That review link seems to be no longer working. Who was it that originally suggested that change? Particularly if it was from @matrixise (or other core dev), I would highly recommended further discussing the removal of that change with them.

cedk:
ping @matrixise

cedk:
@matrixise could you review the changes you requested, thanks.

aeros:
@cedk If you wouldn't mind, could you avoid force-pushing the PR branch after each commit in the future? It makes it rather difficult to follow the commit history of the PR, before and after the review feedback. We advise against doing so in the [devguide](https://devguide.python.org/pullrequest/#submitting):

> Your pull request may involve several commits as a result of addressing code review comments. Please keep the commit history in the pull request intact by not squashing, amending, or anything that would require a force push to GitHub. A detailed commit history allows reviewers to view the diff of one commit to another so they can easily verify whether their comments have been addressed. The commits will be squashed when the pull request is merged.

cedk:
@aeros sorry, I would no more do that. Indeed I'm used to mercurial for which such workflow is well supported thanks to evolve.

cedk:
@matrixise could you point me to the changes you are requesting?

matrixise:
@cedk I am sincerely sorry if I was absent, I will look at your PR in the next two weeks.

matrixise:
Hi @cedk ,

As said @aeros , you don't need to force the push, just push with your commits (we can see all the commits and discuss some points), and when we merge the PR, we will use the **squash and merge** feature. It's pretty useful.

I continue the review of your PR.

Thank you

matrixise (approved):

matrixise:
@cedk just one thing, could you update it with the last master? Thank you so much,

cedk:
> @cedk just one thing, could you update it with the last master? Thank you so much,

But there is no conflict.

matrixise:
Thank you so much for your PR, and I am really sorry for this very bid delay between your proposal and the merge. Thank you again for your patience. Take care

PR #15285, opened 2019-08-14
fix link to time function from time_ns doc

merwok:
Because mod, func, class, etc all share one namespace, :func:time creates a link to the time module doc page rather than the time.time function.

Automerge-Triggered-By: @merwok

### DIFF
--- a/Doc/library/time.rst
+++ b/Doc/library/time.rst
@@ -608,7 +608,7 @@ Functions
 
 .. function:: time_ns() -> int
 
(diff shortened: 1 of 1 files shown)


pganssle (approved):
Good catch. I didn't know about the `~time.time` syntax.

merwok:
I don’t have a link handy, but for any interested reader: tilde is a way to refer to something with a full path (module.object or module.class.method), but only render the last bit as link text.

PR #15287, opened 2019-08-14
bpo-9949: Enable symlink traversal for ntpath.realpath

zooba:
https://bugs.python.org/issue9949

### DIFF
--- a/Lib/ntpath.py
+++ b/Lib/ntpath.py
@@ -545,9 +545,9 @@ else:
         return path
 
     def _getfinalpathname_nonstrict(path):
-        # Allow file (2) or directory (3) not found, or symlinks that
-        # cannot be followed (1921)
-        allowed_winerror = 2, 3, 1921
+        # Allow file (2) or directory (3) not found, invalid syntax (123),
+        # and symlinks that cannot be followed (1921)
+        allowed_winerror = 2, 3, 123, 1921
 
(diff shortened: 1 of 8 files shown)


zooba:
This is ready to merge once CI passes. I'll do it and then rebase #15231 before merging that one.

Lib/ntpath.py
```diff
@@ -519,8 +519,94 @@ def abspath(path):
...
+    realpath = abspath
+else:
+    def _readlink_deep(path, seen=None):
+        if seen is None:
+            seen = set()
+
+        while normcase(path) not in seen:
+            seen.add(normcase(path))
+            try:
+                path = _nt_readlink(path)
```
eryksun, line 535: If you're allowing reading junction targets, you must  fail a remote path here if `path` is a junction, per `os.lstat`. A junction target in a UNC path is meaningless to us. A junction has to target local devices on a system, i.e. its local DOS drive and volume GUID names. At best these aren't defined for us, and at worst they map to an unrelated drive on our side. If for some reason junction "spam" fails to resolve on the server, we cannot go any further. We leave it as "spam" in the UNC path, and we're done.

eryksun: If `path` is a relative symlink, it must be evaluated relative to its parent directory, not out current directory.

PR #15293, opened 2019-08-14
Fix unintialized pointer warning in Objects/obmalloc.c

hansrajdas:
Steps to reproduce, check compilation logs:
```
make clean;make
```

warning-2019-08-15 01-15-06

### DIFF
--- a/Objects/obmalloc.c
+++ b/Objects/obmalloc.c
@@ -1643,6 +1643,7 @@ _PyObject_Malloc(void *ctx, size_t nbytes)
 {
     void* ptr;
     if (LIKELY(pymalloc_alloc(ctx, &ptr, nbytes))) {
+        memset(ptr, 0, nbytes);
         return ptr;
     }
 


vstinner (changes requested):
This change is useless and would have a big negative impact on performances.

If you want initialized memory, use PyObject_Calloc(), not PyObject_Malloc(). PyObject_Malloc() returns uninitialized memory by design.

I understand that the compiler warnings says that ptr *value* **may** be initialized. GCC is wrong: it is always initialized. What is your GCC version? What is your OS? Please open a bug report.

vstinner:
If you want the compiler bug to be fixed, please open a bug report. But this change is wrong, so I closed the PR.

hansrajdas:
Thanks @vstinner for your valuable inputs. I am using **Ubuntu 18.04.2 LTS** and GCC version **7.4.0-1ubuntu1~18.04.1**. I will open a bug for this compiler warning.

As `PyObject_Malloc` is designed to return uninitialized memory, we can assign ptr to NULL(`void* ptr = NULL`) while declaration to fix this warning, will this be a better approach?

PR #15295, opened 2019-08-14
bpo-37645: simplify __str__ of function objects

jdemeyer:
CC @vstinner 

https://bugs.python.org/issue37645

### DIFF
--- a/Lib/test/test_asyncio/test_events.py
+++ b/Lib/test/test_asyncio/test_events.py
@@ -2155,11 +2155,10 @@ class HandleTests(test_utils.TestCase):
             filename, lineno = test_utils.get_function_source(method)
             h = asyncio.Handle(cb, (), self.loop)
 
-            cb_regex = r'<function HandleTests.test_handle_repr .*>'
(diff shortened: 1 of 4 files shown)


epicfaace (commented):
Would repr(f) when f takes arguments also look like `f()`? In that case, it might be misleading (it would look like f takes no arguments)

jdemeyer:
I'm not changing `repr(f)` at all, only `str(f)`. But yes, `str(f)` will always be `f()` regardless of the signature. Note that neither `repr(f)` nor error messages involving `f` write the signature, so I don't plan to add the signature either.
```
>>> def f(a, b): pass
>>> f
<function f at 0x7f9c6e259160>
>>> f(**1)
TypeError: f() argument after ** must be a mapping, not int
```

epicfaace:
Sorry, I meant `str(f)`. My concern was just that printing out `f()` may cause the misleading impression that `f` takes no arguments, while that is not the case for `<function f at 0x7f9c6e259160>`.

PR #15299, opened 2019-08-15
bpo-11671: add header validation from http.client to wsgiref.headers.Headers

epicfaace:
Adds header validation to disallow invalid characters in header names and values. This also fixes https://bugs.python.org/issue28778

I based this off the initial patch in [bpo-11671](https://bugs.python.org/issue11671), with a few changes:
- only check whether headers are valid upon class initialization or calling add_header
- use regex's from http.client, so that it disallows spaces in header names too

I know that http.client's regex (added in [bpo-22928](https://bugs.python.org/issue22928), a112a8ae47813f75aa8ad27ee8c42a7c2e937d13) is not RFC compliant, but it seemed better to at least make both libraries have the same behavior with regards to header validation.

https://bugs.python.org/issue11671

### DIFF
--- a/Lib/test/test_wsgiref.py
+++ b/Lib/test/test_wsgiref.py
@@ -525,14 +525,16 @@ class HeaderTests(TestCase):
             '\r\n'
         )
 
-    def testDisallowNewlines(self):
+    def testValidateHeaders(self):
         h = Headers([])
-        self.assertRaises(AssertionError, h.add_header, 'foo', 'bar\rbaz: bat')
-        self.assertRaises(AssertionError, h.add_header, 'foo', 'bar\nbaz: bat')
-        self.assertRaises(AssertionError, h.add_header, 'foo:\rbar', 'baz')
-        self.assertRaises(AssertionError, h.add_header, 'foo:\nbar', 'baz')
-        self.assertRaises(AssertionError, h.add_header, 'foo', 'bar', baz='bat\rqux: spam')
-        self.assertRaises(AssertionError, h.add_header, 'foo', 'bar', baz='bat\nqux: spam')
+        self.assertRaises(ValueError, h.add_header, 'foo', 'bar\rbaz: bat')
+        self.assertRaises(ValueError, h.add_header, 'foo', 'bar\nbaz: bat')
+        self.assertRaises(ValueError, h.add_header, 'foo: bar', 'baz')
+        self.assertRaises(ValueError, h.add_header, 'foo:\rbar', 'baz')
+        self.assertRaises(ValueError, h.add_header, 'foo:\nbar', 'baz')
+        self.assertRaises(ValueError, h.add_header, 'foo', 'bar\nbaz: bat')
+        self.assertRaises(ValueError, h.add_header, 'foo', 'bar', baz='bat\rqux: spam')
+        self.assertRaises(ValueError, h.add_header, 'foo', 'bar', baz='bat\nqux: spam')
 
 class ErrorHandler(BaseCGIHandler):
     """Simple handler subclass for testing BaseHandler"""
--- a/Lib/test/test_wsgiref.py
+++ b/Lib/test/test_wsgiref.py
@@ -541,20 +541,22 @@ class HeaderTests(TestCase):
         ('ValidName', 'Invalid\nValue'),
         ('ValidName', 'InvalidValue\r\n'),
         ('ValidName', 'InvalidValue\r'),
-        ('ValidName', 'InvalidValue\n')
+        ('ValidName', 'InvalidValue\n'),
+        (b'InvalidName', 'ValidValue', AssertionError),
+        ('ValidName', b'InvalidValue', AssertionError)
     )
 
     def test_add_header_validation(self):
         h = Headers([])
-        for name, value in self.validation_cases:
-            with self.subTest((name, value)):
-                with self.assertRaisesRegex(ValueError, 'Invalid header'):
+        for name, value, *exception in self.validation_cases:
+            with self.subTest((name, value, exception)):
+                with self.assertRaises(exception[0]) if len(exception) else self.assertRaisesRegex(ValueError, 'Invalid header'):
                     h.add_header(name, value)
 
     def test_initialize_validation(self):
-        for name, value in self.validation_cases:
-            with self.subTest((name, value)):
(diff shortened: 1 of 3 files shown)


Lib/wsgiref/headers.py
```diff
@@ -32,6 +34,11 @@ def __init__(self, headers=None):
         headers = headers if headers is not None else []
         if type(headers) is not list:
             raise TypeError("Headers must be a list of name/value tuples")
+        for header, value in headers:
+            if not _is_legal_header_name(header.encode('ascii')):
+                raise ValueError('Invalid header name %r' % (header,))
+            if _is_illegal_header_value(value.encode('ascii')):
+                raise ValueError('Invalid header value %r' % (value,))
```
vstinner: I would prefer to initialize self.headers to an empty list and then call "for header, value in headers: self.add_header(header, value)". Then remove self._convert_string_type(k) calls below.

Lib/test/test_wsgiref.py
```diff
@@ -525,6 +525,38 @@ def testExtras(self):
...
+    )
+
+    def test_add_header_validation(self):
+        h = Headers([])
+        for name, value in self.validation_cases:
+            with self.subTest((name, value)):
+                with self.assertRaisesRegex(ValueError, 'Invalid header'):
+                    h.add_header(name, value)
+
+    def test_initialize_validation(self):
```
vstinner, line 556: Would it be possible to add a test for header name/value of invalid type when calling Header()? Like a bytes string.

vstinner (changes requested):

Lib/test/test_wsgiref.py
epicfaace: Done

Lib/wsgiref/headers.py
epicfaace: Added this. Why would we want to remove self._convert_string_type(k) calls though?

epicfaace:

Lib/test/test_wsgiref.py
```diff
@@ -525,6 +525,40 @@ def testExtras(self):
...
+        ('Invalid:Name', 'ValidValue'),
+        (':InvalidName', 'ValidValue'),
+        ('ValidName', 'Invalid\r\nValue'),
+        ('ValidName', 'Invalid\rValue'),
+        ('ValidName', 'Invalid\nValue'),
+        ('ValidName', 'InvalidValue\r\n'),
+        ('ValidName', 'InvalidValue\r'),
+        ('ValidName', 'InvalidValue\n'),
+        (b'InvalidName', 'ValidValue', AssertionError),
+        ('ValidName', b'InvalidValue', AssertionError)
```
vstinner, line 546: I would prefer to get a TypeError rather than an AssertionError.

Lib/test/test_wsgiref.py
```diff
@@ -525,6 +525,40 @@ def testExtras(self):
...
+    )
+
+    def test_add_header_validation(self):
```
vstinner, line 549: I suggest to merge the two tests a single test, to factorize code.

I suggest to remove the b'InvalidName' and b'InvalidValue' from validation_cases. Maybe create a second list, maybe put it directly in the test method to avoid `with self.assertRaises(exception[0]) if len(exception) else self.assertRaisesRegex(ValueError, 'Invalid header'): ` complexity.

Lib/test/test_wsgiref.py
```diff
@@ -525,6 +525,40 @@ def testExtras(self):
             '\r\n'
         )
 
+    validation_cases = (
```
vstinner, line 528: Please add a comment mentioning bpo-11671.

Rather than an hardcoded list, why not building such list using a few invalid characters. For example for header names:

```
invalid_names = []
invalid_name_chars = ['\r', '\n', '\r\n', ':']
for invalid_chars in invalid_name_chars:
  invalid_names.append(f'{invalid_chars}InvalidName')
  invalid_names.append(f'Invalid{invalid_chars}Name')
  invalid_names.append(f'InvalidName{invalid_chars}')
invalid_names.append(' InvalidName')
```

Are whitespaces (" ") and colon (":") only invalid at the beginning of the name?

Lib/wsgiref/headers.py
```diff
@@ -181,4 +183,10 @@ def add_header(self, _name, _value, **_params):
             else:
                 v = self._convert_string_type(v)
                 parts.append(_formatparam(k.replace('_', '-'), v))
-        self._headers.append((self._convert_string_type(_name), "; ".join(parts)))
+        header = self._convert_string_type(_name)
```
vstinner, line 186: Can you rename the variable to header_name (or maybe name, I  didn't check if it's already used)?

Lib/wsgiref/headers.py
```diff
@@ -181,4 +183,10 @@ def add_header(self, _name, _value, **_params):
...
+        value = "; ".join(parts)
+        if not _is_legal_header_name(header.encode('ascii')):
+            raise ValueError('Invalid header name %r' % (header,))
+        if _is_illegal_header_value(value.encode('ascii')):
```
vstinner, line 190: http/client.py encodes the value to Latin1, not ASCII

Lib/test/test_wsgiref.py
```diff
@@ -525,6 +525,40 @@ def testExtras(self):
...
+        (':InvalidName', 'ValidValue'),
+        ('ValidName', 'Invalid\r\nValue'),
+        ('ValidName', 'Invalid\rValue'),
+        ('ValidName', 'Invalid\nValue'),
+        ('ValidName', 'InvalidValue\r\n'),
+        ('ValidName', 'InvalidValue\r'),
+        ('ValidName', 'InvalidValue\n'),
+        (b'InvalidName', 'ValidValue', AssertionError),
+        ('ValidName', b'InvalidValue', AssertionError)
+    )
```
vstinner, line 547: Are there already tests for header names which cannot be encoded to ASCII? And tests for values not encodable to Latin1?

vstinner (changes requested):

csabella:
@epicfaace, please address the code review comments.  Thank you!

csabella:
@epicfaace ping

picnixz:
Closing because superseded by https://github.com/python/cpython/pull/143917

PR #15300, opened 2019-08-15
bpo-37864: Correct and deduplicate "isprintable" docs; add test.

gnprice:
We had the definition of what makes a character "printable"
documented in three places, giving two different definitions.

The definition in the comment on `_PyUnicode_IsPrintable` was
inverted; correct that.

With that correction, the two definitions turn out to be equivalent --
but to confirm that, you have to go look up, or happen to know, that
those are the only five "Other" categories and only three "Separator"
categories in the Unicode character database.  That makes it hard for
the reader to tell whether they really are the same, or if there's
some subtle difference in the intended semantics.

Fix that by cutting the C API docs' and the C comment's copies of
the subtle details, in favor of referring to the Python-level docs.
That ensures it's explicit that these are all meant to agree, and
also lets us concentrate improvements to the wording in one place.

Speaking of which, borrow some ideas from the C comment, along with
other tweaks, to hopefully add a bit more clarity to that one
newly-centralized copy in the docs.

Also add a thorough test that the implementation agrees with
this definition.

https://bugs.python.org/issue37864

### DIFF
--- a/Doc/c-api/unicode.rst
+++ b/Doc/c-api/unicode.rst
@@ -305,13 +305,8 @@ the Python configuration.
 
 .. c:function:: int Py_UNICODE_ISPRINTABLE(Py_UNICODE ch)
 
-   Return ``1`` or ``0`` depending on whether *ch* is a printable character.
-   Nonprintable characters are those characters defined in the Unicode character
-   database as "Other" or "Separator", excepting the ASCII space (0x20) which is
-   considered printable.  (Note that printable characters in this context are
-   those which should not be escaped when :func:`repr` is invoked on a string.
-   It has no bearing on the handling of strings written to :data:`sys.stdout` or
-   :data:`sys.stderr`.)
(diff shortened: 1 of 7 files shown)


terryjreedy:
"Generated files not up to date
 M Objects/clinic/unicodeobject.c.h
 M Objects/unicodeobject.c"

I believe you have to regenerate these files and include the new versions in the branch and the new diffs in this  PR.  If 'how' is not in the devguide, ask on core-mentorship list.

gnprice:
@terryjreedy Ah, thanks for spotting that!

Indeed, `clinic` needed to be run after I'd edited a docstring. Fixed now.

gnprice:
(I also just added a NEWS entry, which should turn that red X into a happy green check-mark.)

encukou:
Merged in #130118, with another regen-all.

Sorry for the wait!

PR #44, opened 2017-02-12
[backport to 2.7] Change documentation's `Show Source` link to GitHub

Mariatta:
Change documentation's `Show Source` link to GitHub (2.7 branch)
Render issue role as bpo-NNNN

PR #15301, opened 2019-08-15
bpo-36502: Update link to UAX #44, the Unicode doc on the UCD.

gnprice:
The link we have points to the version from Unicode 6.0.0, dated 2010.
There have been numerous updates to it since then:
  https://www.unicode.org/reports/tr44/#Modifications

Change the link to one that points to the current version.

https://bugs.python.org/issue36502

### DIFF
--- a/Doc/library/unicodedata.rst
+++ b/Doc/library/unicodedata.rst
@@ -22,7 +22,7 @@ this database is compiled from the `UCD version 12.1.0
 
 The module uses the same names and symbols as defined by Unicode
 Standard Annex #44, `"Unicode Character Database"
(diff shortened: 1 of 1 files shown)


gnprice:
(Attached to this particular issue because it's where I added a link to this docs page from another docs page, in a context where if the reader wants to really know details then it's likely they'll in turn follow this link to the UCD. I guess I could also have left off an issue number, probably.)

benjaminp (approved):

gnprice:
> use https

Ah, good catch :-) -- I hadn't noticed that it was an `http:` link.

Thanks for the review and merge!

PR #15302, opened 2019-08-15
bpo-37758: Cut always-constant conditionals on sys.maxunicode.

gnprice:
Since PEP 393 in Python 3.3, this value is always 0x10ffff, the
maximum codepoint in Unicode; there's no longer such a thing as a
UCS-2 build of Python, which couldn't properly represent some
characters.

There are a couple of spots left where we still condition on the value
of this constant.  Take them out.

https://bugs.python.org/issue37758


asottile (approved):
looks good

PR #15309, opened 2019-08-15
bpo-37732 : Fix unintialized pointer warning in Objects/obmalloc.c

hansrajdas:
* Initialize ptr with NULL pointer while declaring

warning-2019-08-15 01-15-06

https://bugs.python.org/issue37732

### DIFF
--- a/Objects/obmalloc.c
+++ b/Objects/obmalloc.c
@@ -1641,7 +1641,7 @@ pymalloc_alloc(void *ctx, void **ptr_p, size_t nbytes)
 static void *
 _PyObject_Malloc(void *ctx, size_t nbytes)
 {
-    void* ptr;
+    void* ptr = NULL;
     if (LIKELY(pymalloc_alloc(ctx, &ptr, nbytes))) {
         return ptr;
     }


nanjekyejoannah (commented):
I had already opened a bpo issue here for this : https://bugs.python.org/issue37732 for someone to investigate. Please link to it in this PR title. 
i.e

bpo-37732 : Fix unintialized pointer warning in Objects/obmalloc.c

hansrajdas:
> I had already opened a bpo issue here for this : https://bugs.python.org/issue37732 for someone to investigate. Please link to it in this PR title.
> i.e
> 
> bpo-37732 : Fix unintialized pointer warning in Objects/obmalloc.c

Thanks @nanjekyejoannah, I will link this PR with opened bpo issue. And also move this PR from draft to open state.

methane:
Why no warning in `_PyObject_Calloc`?

hansrajdas:
> Why no warning in `_PyObject_Calloc`?

This is because in `_PyObject_Calloc` memory is initialized using `memset` but in case of  `_PyObject_Malloc` memory is just allocated and returned(no initialization is done). If we remove memset statement from `_PyObject_Calloc` we will get the same pointer uninitialized warning for `_PyObject_Calloc` also. I have verified this.

This is another PR https://github.com/python/cpython/pull/15293 where Victor suggested that `_PyObject_Malloc` was intentionally designed to return uninitialized memory so we can't use memset here to fix this warning so I have just initialized ptr with `NULL` pointer to fix this warning. Please guide me if we can do anythig better to fix this warning.

qigangxu:
It seems that warning was introduced by bpo-37543: optimize pymalloc (#14674 

Why does this warning appear after using 'LIKELY'?

* https://github.com/python/cpython/pull/14674
* https://github.com/python/cpython/commit/fb26504d14a08fcd61bb92bb989b6d2b12188535#diff-c078b6b7dd8dc2125fb4d48f7020b796

qigangxu:
And maybe we can resue _PyObject_Malloc in _PyObject_Calloc.

vstinner:
I dislike this change since it adds an **useless** initialization to NULL, ptr is already always initialized: GCC is just wrong. I wrote PR #15333 to fix the warning without adding an useless initialization.

_PyObject_Malloc() function is performance critical so I would prefer to not make it slower ;-)

vstinner:
Close: I merged my PR #15333 instead.

hansrajdas:
Thanks @vstinner for your guidance, It was good learning for me following your PR https://github.com/python/cpython/pull/15333 :)

PR #15315, opened 2019-08-16
bpo-37878: Make PyThreadState_DeleteCurrent() Internal

nanjekyejoannah:
Make `PyThreadState_DeleteCurrent()` Internal.

https://bugs.python.org/issue37878


nanjekyejoannah:
cc  @ericsnowcurrently

Doc/c-api/init.rst
```diff
@@ -1023,6 +1023,12 @@ All of the following functions must be called after :c:func:`Py_Initialize`.
    The thread state must have been reset with a previous call to
    :c:func:`PyThreadState_Clear`.
 
+.. c:function:: void PyThreadState_DeleteCurrent()
+
+    Destroy the current thread state and releases the global interpreter. The global
```
willingc:
```suggestion
    Destroy the current thread state and release the global interpreter lock.
```

Doc/c-api/init.rst
```diff
@@ -1023,6 +1023,12 @@ All of the following functions must be called after :c:func:`Py_Initialize`.
...
+    interpreter lock need not be held. Like :c:func:`PyThreadState_Delete`, the global
```
willingc:
```suggestion
    Like :c:func:`PyThreadState_Delete`, the global
```

Doc/c-api/init.rst
```diff
@@ -1023,6 +1023,12 @@ All of the following functions must be called after :c:func:`Py_Initialize`.
...
+    interpreter lock need not be held. The thread state must have been reset with a previous
```
willingc:
```suggestion
    interpreter lock need not be held. When using :c:func:`PyThreadState_DeleteCurrent`, the thread state must have been reset with a previous
```

willingc (commented):
Thanks @nanjekyejoannah for the PR. I've made some suggested wording changes for reader clarity.

nanjekyejoannah:
@willingc  Changes committed.

willingc:
Let's give @ericsnowcurrently a chance to review too. Thanks. Please ping me if this is still open later this week and I will merge.

willingc (dismissed):

vstinner (changes requested):
I'm not sure that it's a good idea to document this function: https://bugs.python.org/issue37878#msg349954

Modules/_threadmodule.c
```diff
@@ -1025,7 +1025,7 @@ t_bootstrap(void *boot_raw)
     PyMem_DEL(boot_raw);
     tstate->interp->num_threads--;
     PyThreadState_Clear(tstate);
-    PyThreadState_DeleteCurrent();
+    _PyThreadState_DeleteCurrent(&_PyRuntime);
```
vstinner: Please add "_PyRuntimeState *runtime = &_PyRuntime;" at the beginning of the function, and replace &_PyRuntime with runtime.

Or maybe even better: add the field to "struct bootstate" and fill the field in thread_PyThread_start_new_thread(). It would make t_bootstrap() more "state-less" (don't access directly global variables).

vstinner (changes requested):

vstinner (commented):
Since it is a public function, even if it's not documented, please document the removal at:
https://docs.python.org/dev/whatsnew/3.9.html#removed

You can mention that the removed function was not documented.

For example, in Python 3.8, I wrote:

"The PyByteArray_Init() and PyByteArray_Fini() functions have been removed. They did nothing since Python 2.7.4 and Python 3.2.0, were excluded from the limited API (stable ABI), and were not documented. (Contributed by Victor Stinner in bpo-35713.)"

nanjekyejoannah:

vstinner (approved):
LGTM.

vstinner:
Thanks @nanjekyejoannah! It now looks better: remove the function rather than documenting it :-) I prefer to have a smaller C API!

tacaswell:
This breaks pybind 11.  Opened an issue with them (https://github.com/pybind/pybind11/issues/1932).  Where this should be fixed is a bit over my head at the moment.

wjakob:
See https://bugs.python.org/issue38266.

henryiii:
I'm confused by the attached issue; it looks like the decision was to simply revert the change with the function moved, and not add the underscore; and maybe even make it slightly easier to use. Now that 3.9b1 is out, PyBind11-based libraries cannot build, and the only fix would be to start using a now internal function in CPython. SciPy is an example of a PyBind11-using library.

vstinner:
PyThreadState_DeleteCurrent() is declared by the public C API, but it's now excluded from the limited C API. Does pybind11 uses the limited C API? What is your issue? Please comment https://bugs.python.org/issue37878 instead of this merged PR.

henryiii:
I'm sorry, I'll add a bugs.python.org account or see if I had one in the past if I need to comment there. I realize, from looking through the code, that an underscore was not added, the function was moved intact (which was not clear from the PR), exactly as it was supposed to be in the issue. The only remaining problem is that the release notes here are wrong: https://docs.python.org/3.9/whatsnew/3.9.html#removed (which is a much better problem).

I'd love it if PyBind11 could use the limited API, but it can't.

Thank you, @vstinner !

vstinner:
> The only remaining problem is that the release notes here are wrong: https://docs.python.org/3.9/whatsnew/3.9.html#removed (which is a much better problem).

I created https://github.com/python/cpython/pull/20489 to update the What's New in Python 3.9 document. Would you mind to review my PR?

PR #15318, opened 2019-08-17
Fix an invalid assertEqual() call in test_descr.py

ZackerySpytz:
It seems that `self.fail()` was intended, not `self.assertEqual()`.


tirkarthi (approved):
LGTM. The commit 479a7e7af669cea2eb93e02d87abe8d9b8a50ae2 changes `raise TestFailed(reason)` statements to use `self.fail` and the check is done correctly `test_classmethods`. Thanks.

pablogsal (approved):

pablogsal:
Thanks, @ZackerySpytz for the PR!

PR #15323, opened 2019-08-17
bpo-37879: Suppress subtype_dealloc decref when base type is a C heap type

eduardo-elizondo:
The instance destructor for a type is responsible for preparing
an instance for deallocation by decrementing the reference counts
of its referents.

If an instance belongs to a heap type, the type object of an instance
has its reference count decremented while for static types, which
are permanently allocated, the type object is unaffected by the
instance destructor.

Previously, the default instance destructor searched the class
hierarchy for an inherited instance destructor and, if present,
would invoke it.

Then, if the instance type is a heap type, it would decrement the
reference count of that heap type.  However, this could result in the
premature destruction of a type because the inherited instance
destructor should have already decremented the reference count
of the type object.

This change avoids the premature destruction of the type object
by suppressing the decrement of its reference count when an
inherited, non-default instance destructor has been invoked.

Finally, an assertion on the Py_SIZE of a type was deleted.  Heap
types have a non zero size, making this into an incorrect assertion.

https://bugs.python.org/issue37879


Objects/typeobject.c
```diff
@@ -1157,11 +1157,9 @@ subtype_dealloc(PyObject *self)
     /* Test whether the type has GC exactly once */
 
     if (!PyType_IS_GC(type)) {
-        /* It's really rare to find a dynamic type that doesn't have
-           GC; it can only happen when deriving from 'object' and not
-           adding any slots or instance variables.  This allows
-           certain simplifications: there's no need to call
-           clear_slots(), or DECREF the dict, or clear weakrefs. */
+        /* A non GC dyanmic type allows certain simplifications:
```
kevinjacobs-progenity: typo in dynamic

Objects/typeobject.c
```diff
@@ -1189,8 +1186,10 @@ subtype_dealloc(PyObject *self)
         assert(basedealloc);
         basedealloc(self);
 
-        /* Can't reference self beyond this point */
-        Py_DECREF(type);
+       /* Only decref if the base type is not already a heap allocated type.
+          Otherwise, basedealloc should have decref'd it already */
+        if (type->tp_flags & Py_TPFLAGS_HEAPTYPE && !(base->tp_flags & Py_TPFLAGS_HEAPTYPE))
+                                                Py_DECREF(type);
```
kevinjacobs-progenity: odd looking indent, at least via github

Objects/typeobject.c
eduardo-elizondo: Thanks, fixed!

Objects/typeobject.c
eduardo-elizondo: You are right that was definitely non-intentional. I think my auto rebase messed this up. Anyways, fixed!

Objects/typeobject.c
```diff
@@ -1189,8 +1186,10 @@ subtype_dealloc(PyObject *self)
         assert(basedealloc);
         basedealloc(self);
 
-        /* Can't reference self beyond this point */
-        Py_DECREF(type);
+       /* Only decref if the base type is not already a heap allocated type.
+          Otherwise, basedealloc should have decref'd it already */
+        if (type->tp_flags & Py_TPFLAGS_HEAPTYPE && !(base->tp_flags & Py_TPFLAGS_HEAPTYPE))
```
jdemeyer, line 1191: Why the first condition `type->tp_flags & Py_TPFLAGS_HEAPTYPE`? There is already an assertion at the beginning of `subtype_dealloc`.

jdemeyer (changes requested):
Apart from the one comment, the changes to `typeobject.c` look good. I haven't yet looked at the testcases.

Objects/typeobject.c
eduardo-elizondo: The type can be changed by `tp_del`. If you look closely at the implementation, the type is extracted again. So the type that you are looking at this spot might not be the same type that you started off with.

jdemeyer: > The type can be changed by `tp_del`

I see. Reading the code for `__class__` assignment, this would only be possible for subtypes of `module`. Do you know if there is a testcase for this?

eduardo-elizondo: It's possible for both subtypes of module and heap types. The specific check done by `object_set_class` in `typeobject.c` is: https://github.com/python/cpython/blob/master/Objects/typeobject.c#L4049-L4057

And yes, I've added two tests for this. One for the `!PyType_IS_GC` case (which is the one you have here). And for the `PyType_IS_GC` case down below.

Both tests have been added to `test_capi.py` and are named `test_c_subclass_of_heap_ctype_with_del_modifying_dunder_class_only_decrefs_once` and `test_subclass_of_heap_gc_ctype_with_del_modifying_dunder_class_only_decrefs_once` respectively.

Modules/_testcapimodule.c
```diff
@@ -6012,6 +6012,180 @@ static struct PyModuleDef _testcapimodule = {
...
+    }
+    PyTypeObject *ctypesubclass = (PyTypeObject *)PyObject_GetAttrString(m, "HeapCTypeSubclass");
+    if (ctypesubclass == NULL) {
+        return -1;
+    }
+
+    PyTypeObject *base = (PyTypeObject *)PyType_GetSlot(ctypesubclass, Py_tp_base);
+    Py_DECREF(ctypesubclass);
+    initproc base_init = PyType_GetSlot(base, Py_tp_init);
+    if (base_init(self, args, kwargs) < 0) {
```
encukou: Is this setup to get the superclass necessary?
It should be possible to call `heapctype_init(self, args, kwargs)` directly.

Modules/_testcapimodule.c
```diff
@@ -6012,6 +6012,180 @@ static struct PyModuleDef _testcapimodule = {
...
+heapctypesubclasswithfinalizer_finalize(PyObject *self)
+{
+    PyObject *error_type, *error_value, *error_traceback, *m, *oldtype, *newtype;
+
+    /* Save the current exception, if any. */
+    PyErr_Fetch(&error_type, &error_value, &error_traceback);
+
+    m = PyState_FindModule(&_testcapimodule);
+    if (m == NULL) {
+        return;
```
encukou: This leaks error_type, error_value, error_traceback.

Modules/_testcapimodule.c
eduardo-elizondo: It is indeed. I actually had your suggestion initially but ultimately ended up changing it to this approach. My rationale was: To show that `HeapCTypeSubclass` is indeed a C extension subclass and that you can call the `tp_init` slot by pulling its `tp_base` out. To me, it seemed closer to what a `__init__` would do by calling `super().__init__`.

That being said, I don't have any strong opinions on either. I'll keep this as is but feel free to add one more comment to change to to call `heapctype_init` and I will change it this time!

Modules/_testcapimodule.c
eduardo-elizondo: You are right. Let me fix this.

Edit: Fixed

eduardo-elizondo:
@encukou Comments have been addressed. Also, hope the core sprint is going well (if you are there) :)

encukou:
Yup, the sprint is going well!

I have added some commits directly to this branch, rather than tell you about the nitpicks I have and let you write the code.
Could you take a look at them? Feel free to remove/change any of them (or tell me to do so) if that doesn't work for you.

Modules/_testcapimodule.c
encukou: There are some subtle issues in doing dynamic `super()`-like calls from C code, but using `PyState_FindModule`/`PyObject_GetAttrString` is especially fragile.
If the elaborate code manages to set `base_init` to anything except `heapctype_init`, it's a bug. It's not the bug this is supposed to be testing; and it's a bug that's likely to segfault.

I'm actually trying to provide a better way to do this without `PyState_FindModule` ([PEP 573](https://www.python.org/dev/peps/pep-0573/)), but so far, using `heapctype_init` is, IMO, best.

eduardo-elizondo (commented):
LGTM

eduardo-elizondo (commented):
Both changes and the reasoning looks good to me

eduardo-elizondo (commented):
Nice catch!

eduardo-elizondo (commented):
Looks good

eduardo-elizondo:
@encukou I went through all of you changes and they all look good! I also rebased this to master to get rid of the merge conflicts so this change should be good to go!

Given that we are in the topic of heap allocated types, I've got posixmodule to use heap allocated types only in https://github.com/python/cpython/pull/15892 in case you'd like to take a look 😄

PR #15328, opened 2019-08-18
bpo-37836: document as_integer_ratio() in the data model

jdemeyer:
https://bugs.python.org/issue37836

### DIFF
--- a/Doc/reference/datamodel.rst
+++ b/Doc/reference/datamodel.rst
@@ -2399,6 +2399,18 @@ left undefined.
    and :func:`complex` fall back to :meth:`__index__`.
 
 
+.. method:: object.as_integer_ratio(self)
+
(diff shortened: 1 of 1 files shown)


rhettinger:
You should get Guido to opine on this.  I think you're making this more official and pervasive and prescriptive than he intended.

There is a difference between making a helpful utility available versus trying to make everyone change their code to use it.

gvanrossum (approved):

jdemeyer:
I made this a draft PR since the doc says
```
This is used for conversion to :class:`fractions.Fraction`
```
but this is not yet true at the moment. So it depends on either #15327 (by me) or #15329 (by @serhiy-storchaka).

AA-Turner:
The linked issue was closed (https://github.com/python/cpython/issues/82017#issuecomment-1093831647) -- should this PR remain?

A

PR #15330, opened 2019-08-19
bpo-37885: venv: Don't produce unbound variable warning on deactivate

danabr:
Before, running deactivate from a bash shell configured to treat undefined variables as errors (`set -u`) would produce a warning:

``` 
$ python3 -m venv test
$ source test/bin/activate
(test) $ deactivate
-bash: $1: unbound variable
```

https://bugs.python.org/issue37885

### DIFF
--- a/Lib/test/test_venv.py
+++ b/Lib/test/test_venv.py
@@ -9,6 +9,7 @@ import ensurepip
 import os
 import os.path
 import re
+import shutil
 import struct
 import subprocess
 import sys
@@ -359,6 +360,25 @@ class BasicTest(BaseTest):
             'print(pool.apply_async("Python".lower).get(3)); '
             'pool.terminate()'])
         self.assertEqual(out.strip(), "python".encode())
+
+    @unittest.skipIf(os.name == 'nt', 'not relevant on Windows')
+    def test_deactivate_with_strict_bash_opts(self):
+        bash = shutil.which("bash")
+        if bash is None:
+            self.skipTest("bash required for this test")
+        rmtree(self.env_dir)
+        builder = venv.EnvBuilder(clear=True)
+        builder.create(self.env_dir)
+        activate = os.path.join(self.env_dir, self.bindir, "activate")
+        test_script = os.path.join(self.env_dir, "test_strict.sh")
+        with open(test_script, "w") as f:
+            f.write("set -euo pipefail\n"
+                    f"source {activate}\n"
+                    "deactivate\n")
+        out, err = check_output([bash, test_script])
+        self.assertEqual(out, "".encode())
+        self.assertEqual(err, "".encode())
+
 
 @requireVenvCreate
 class EnsurePipTest(BaseTest):
--- a/Lib/test/test_venv.py
+++ b/Lib/test/test_venv.py
@@ -324,7 +324,7 @@ class BasicTest(BaseTest):
             'import sys; print(sys.executable)'])
         self.assertEqual(out.strip(), envpy.encode())
 
-    @unittest.skipUnless(os.name == 'nt', 'only relevant on Windows')
+    @unittest.skipUnless(sys.platform == 'win32', 'only relevant on Windows')
     def test_unicode_in_batch_file(self):
         """
         Test handling of Unicode paths
--- a/Lib/test/test_venv.py
+++ b/Lib/test/test_venv.py
@@ -324,7 +324,7 @@ class BasicTest(BaseTest):
             'import sys; print(sys.executable)'])
         self.assertEqual(out.strip(), envpy.encode())
 
-    @unittest.skipUnless(sys.platform == 'win32', 'only relevant on Windows')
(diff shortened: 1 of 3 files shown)


the-knights-who-say-ni:

epicfaace (approved):
I tested this and it works, thanks!

Do you think it might be possible to add a test to https://github.com/python/cpython/blob/c4cacc8c5eab50db8da3140353596f38a01115ca/Lib/test/test_venv.py? If it's difficult, then probably not necessary to do so.

danabr:
@epicfaace : I wrote at test, but am not sure if it is good enough to include. Could you have a look here: https://github.com/python/cpython/commit/a242378e9cab20b95b28d9c0d3498268b993dd2d

Yhg1s:
The test looks fine, except it should probably use shutil.which to find bash (and I have some style comments), so please include it in this PR.

Lib/test/test_venv.py
```diff
@@ -360,6 +361,26 @@ def test_multiprocessing(self):
...
+        bash = shutil.which("bash")
+        if bash is None:
+            self.skipTest("bash required for this test")
+        rmtree(self.env_dir)
+        builder = venv.EnvBuilder(clear=True)
+        builder.create(self.env_dir)
+        activate = os.path.join(self.env_dir, self.bindir, 'activate')
+        test_script = os.path.join(self.env_dir, 'test_strict.sh')
+        with open(test_script, 'w') as f:
+            f.write("""
```
Yhg1s: Using a multiline string like this is a little jarring (because it breaks indentation). Either move the test script to a global variable, or build the string out of single-line, indented strings (like a few of the other tests do).

Lib/test/test_venv.py
```diff
@@ -360,6 +361,26 @@ def test_multiprocessing(self):
...
+set -euo pipefail
+source {}
+deactivate
+            """.format(activate))
```
Yhg1s: Use an f-string (we won't be backporting it to 3.5 anyway.)

Lib/test/test_venv.py
```diff
@@ -360,6 +361,26 @@ def test_multiprocessing(self):
...
+        out, err = check_output([bash, test_script])
+        self.assertEqual(out, "".encode())
```
Yhg1s, line 379: I was going to complain about the inconsistent use of quotes... but it turns out this is actually somewhat consistent with the rest of the file.

Lib/test/test_venv.py
```diff
@@ -360,6 +361,26 @@ def test_multiprocessing(self):
...
+        builder.create(self.env_dir)
+        activate = os.path.join(self.env_dir, self.bindir, 'activate')
+        test_script = os.path.join(self.env_dir, 'test_strict.sh')
+        with open(test_script, 'w') as f:
+            f.write("""
+set -euo pipefail
+source {}
+deactivate
+            """.format(activate))
+        out, err = check_output([bash, test_script])
```
Yhg1s, line 378: Turns out this fails on Windows on the Azure pipeline, because 'bash' is available there, despite being windows. To debug it I think this'll have to cach CalledProcessError and print the stderr.

gpshead: Alternatively you could just `@unittest.skipUnless(sys.platform == "win32")` that test or a better test.support equivalent API?

I _assume_ venv activation stuff isn't intended to work on bash on windows?  @zware as a random one to confirm that.

zware: I don't think there's really any reason for it not to.  The `activate` script is installed on Windows, and `source test_env/Scripts/activate` seems to work on my Windows 8.1 buildbot which has `bash` available at `C:\Program Files\Git\bin\bash.exe` (which is on PATH for the builder).

zware: I would guess that this failure on Azure is due to `bash` not actually being `bash`, but a redirection to "here's how to install bash".  @zooba might have a way to detect that?

gpshead: It gets pretty esoteric to check that specific of an external environment behavior reliably.  you basically need to test the output and exit code of `subprocess.run([bash_path, "-c", "echo 'this is actually bash' ; exit 99"])` or similar to see if your found bash acts like bash.  is it worth it?

zooba: It's trying to run Git Bash, but there's no indication in the logged output what the error message is, so who knows?

I'm okay with skipping Bash activation on Windows.

Lib/test/test_venv.py
danabr: Thanks. Fixed.

Lib/test/test_venv.py
danabr: Switched to f-string.

Lib/test/test_venv.py
danabr: I went with skipping the test on Windows. I assume you would use activate.bat there anyways?

PR #15338, opened 2019-08-19
Revert "bpo-37788: Fix a reference leak if a thread is not joined"

vstinner:
Reverts python/cpython#15228


pablogsal (approved):

pablogsal:
Buildbot failures at https://bugs.python.org/issue37889

PR #15346, opened 2019-08-20
Remove a dead comment from ossaudiodev.c

brettcannon:


native-api:
While you are at it, could you remove "all rights reserved", too, since that's untrue?

brettcannon:
@native-api I am purposefully doing the non-controversial, IANAL change. What you're asking for I'm not personally willing to do without legal clearance.

PR #15347, opened 2019-08-20
Update asyncio.ensure_future() documentation

akindofyoga:
Added back mention that ensure_future actually scheduled obj. This documentation just mentions what ensure_future returns, so I did not realize that ensure_future also schedules obj.

### DIFF
--- a/Doc/library/asyncio-future.rst
+++ b/Doc/library/asyncio-future.rst
@@ -37,7 +37,7 @@ Future Functions
    * a :class:`Task` object wrapping *obj*, if *obj* is a
      coroutine (:func:`iscoroutine` is used for the test);
      in this case the coroutine will be scheduled by
-    ``ensure_future()``.
+     ``ensure_future()``.
 
    * a :class:`Task` object that would await on *obj*, if *obj* is an
      awaitable (:func:`inspect.isawaitable` is used for the test.)
--- a/Doc/library/asyncio-future.rst
+++ b/Doc/library/asyncio-future.rst
@@ -28,8 +28,8 @@ Future Functions
 
 .. function:: ensure_future(obj, \*, loop=None)
 
-   Schedule the execution of a :ref:`coroutine object <coroutine>`: wrap it in
-   a future. Return a :class:`Task` object.
+   Schedule the execution of a :term:`coroutine object <coroutine>` and wrap it
+   in a future. Return a :class:`Task` object.
 
    Return:
 
--- a/Doc/library/asyncio-future.rst
+++ b/Doc/library/asyncio-future.rst
@@ -28,9 +28,6 @@ Future Functions
 
 .. function:: ensure_future(obj, \*, loop=None)
 
-   Schedule the execution of a :term:`coroutine object <coroutine>` and wrap it
-   in a future. Return a :class:`Task` object.
-
    Return:
 
    * *obj* argument as is, if *obj* is a :class:`Future`,
@@ -38,7 +35,9 @@ Future Functions
      is used for the test.)
 
    * a :class:`Task` object wrapping *obj*, if *obj* is a
-     coroutine (:func:`iscoroutine` is used for the test.)
+     coroutine (:func:`iscoroutine` is used for the test);
+     in this case the coroutine will be scheduled by
+    ``ensure_future()``.
 
    * a :class:`Task` object that would await on *obj*, if *obj* is an
      awaitable (:func:`inspect.isawaitable` is used for the test.)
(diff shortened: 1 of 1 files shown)


the-knights-who-say-ni:

Doc/library/asyncio-future.rst
```diff
@@ -28,6 +28,9 @@ Future Functions
 
 .. function:: ensure_future(obj, \*, loop=None)
 
+   Schedule the execution of a :ref:`coroutine object <coroutine>`: wrap it in
```
aeros: The use of the ``:term:`` Sphinx role links to Python's glossary directly. ``:ref:`` would not be incorrect in this case, but ``:term:`` is more specific. The glossary would be probably be the best place to direct users to if they are unfamiliar with coroutines.

Additionally, the extra ``:`` character after the Sphinx role isn't needed. 
```suggestion
   Schedule the execution of a :term:`coroutine object <coroutine>` and wrap it in
```

After adding this change, it may be necessary to split this into an additional line. [reST specifies](https://devguide.python.org/documenting/#use-of-whitespace) that the maximum line width should be 80 characters, this change would place the total for this line at 82.

aeros (changes requested):
Thanks for the PR @rogerthat94 and welcome!

I have a few recommendations for you:

1) Create an account on https://bugs.python.org/ and sign the CLA. Usually this takes around 24 hours and sometimes the status must be updated manually by [checking the status](https://check-python-cla.herokuapp.com/) for GitHub.

2) Adjust the PR title to ``Update asyncio.ensure_future() documentation``. The title you are using is definitely along the right lines, but it will make it easier for core developers knowledgeable with that specific function to find this PR.

Let me know if you have any questions and I'll do my best to answer them. (:

Doc/library/asyncio-future.rst
akindofyoga: Thanks! I have signed the CLA, updated the title, and made the changes that you requested.

1st1 (changes requested):
I don't understand why this change is needed.  The description you added isn't correct -- the function is quite complex and can do multiple things based on the type of its argument. The documentation already describes this all in full in a succinct form.

1st1:
@rogerthat94 If you want we can do this the other way:

```
   * a :class:`Task` object wrapping *obj*, if *obj* is a
     coroutine (:func:`iscoroutine` is used for the test);
     in this case the coroutine will be scheduled by
    ``ensure_future()``.
     
```

rogerthat94:
The suggestion in your second comment looks fine to me.

The current documentation does not mention that asyncio.ensure_future() will schedule a coroutine. It just says what asyncio.ensure_future() returns. I think highlighting the fact that asyncio.ensure_future() can schedule a coroutine in addition to returning something is important.

rogerthat94:

asvetlov:
Is there a thing like *scheduled coroutine* at all?
There is scheduled *callback*.
A task starts running just after creation, so there is no created-but-not-started task.

That's how asyncio works, nothing special for `ensure_future()`: is you have a task the object is either running or finished (successfully or by exception).

Scheduled coroutine looks incorrect.

Doc/library/asyncio-future.rst
```diff
@@ -35,7 +35,9 @@ Future Functions
      is used for the test.)
 
    * a :class:`Task` object wrapping *obj*, if *obj* is a
-     coroutine (:func:`iscoroutine` is used for the test.)
+     coroutine (:func:`iscoroutine` is used for the test);
+     in this case the coroutine will be scheduled by
+    ``ensure_future()``.
```
1st1:
```suggestion
     ``ensure_future()``.
```

rogerthat94:
The old documentation for [ensure_future](https://docs.python.org/3.5/library/asyncio-task.html#asyncio.ensure_future) says "Schedule the execution of a coroutine object: wrap it in a future. Return a Task object."

1st1:
> There is scheduled callback.

Well... A Task schedules a callback to advance a coroutine and it controls it after that with more callbacks. I say this is a technicality, but updated documentation **does read a bit better**.

asvetlov:
Ok, up to you.
I refer myself on your judgment in documentation wording :)

1st1:
I mean I say "schedule coroutine" all the time :) Maybe we do need a better terminology, but as it currently stands I guess this one is fine. :)

1st1 (approved):

1st1:
Thank you, @rogerthat94!

aeros167:
Thanks for the PR @rogerthat94 and thanks for the review @1st1 and @asvetlov!

aeros167:
@1st1 
> Maybe we do need a better terminology, but as it currently stands I guess this one is fine. :)

If we do come up with a better word for it, be sure to add it to the glossary! This will help to ensure that readers have a clear understanding of what is being referred to. There's several async terms on there already.

PR #15353, opened 2019-08-21
Unmark files as executable that can't actually be executed.

gnprice:
There are plenty of legitimate scripts in the tree that begin with a
`#!`, but also a few that seem to be marked executable by mistake.

Found them with this command -- it gets executable files known to Git,
filters to the ones that don't start with a `#!`, and then unmarks
them as executable:

    $ git ls-files --stage \
      | perl -lane 'print $F[3] if (!/^100644/)' \
      | while read f; do
          head -c2 "$f" | grep -qxF '#!' \
          || chmod a-x "$f"; \
        done

Looking at the list by hand confirms that we didn't sweep up any
files that should have the executable bit after all.  In particular

 * The `.psd` files are images from Photoshop.

 * The `.bat` files sure look like things that can be run.
   But we have lots of other `.bat` files, and they don't have
   this bit set, so it must not be needed for them.

Automerge-Triggered-By: @benjaminp

PR #15354, opened 2019-08-21
Mark files as executable that are meant as scripts.

gnprice:
This is the converse of GH-15353 -- in addition to plenty of
scripts in the tree that are marked with the executable bit
(and so can be directly executed), there are a few that have
a leading `#!` which could let them be executed, but it doesn't
do anything because they don't have the executable bit set.

Here's a command which finds such files and marks them.  The
first line finds files in the tree with a `#!` line *anywhere*;
the next-to-last step checks that the *first* line is actually of
that form.  In between we filter out files that already have the
bit set, and some files that are meant as fragments to be
consumed by one or another kind of preprocessor.

    $ git grep -l '^#!' \
      | grep -vxFf <( \
          git ls-files --stage \
          | perl -lane 'print $F[3] if (!/^100644/)' \
        ) \
      | grep -ve '\.in$' -e '^Doc/includes/' \
      | while read f; do
          head -c2 "$f" | grep -qxF '#!' \
          && chmod a+x "$f"; \
        done


epicfaace (commented):
Do the `.wpr` files need to be executable?

gnprice:
> Do the `.wpr` files need to be executable?

Good question!

I took a look at those when preparing this, because I wasn't sure what they were or why they apparently had `#!` lines.

Here's one of them, which is representative:
```
$ cat Misc/python-wing5.wpr
#!wing
#!version=5.0
##################################################################
# Wing IDE project file                                          #
##################################################################
[project attributes]
proj.directory-list = [{'dirloc': loc('..'),
                        'excludes': [u'.hg',
                                     u'Lib/unittest/__pycache__',
                                     u'Lib/unittest/test/__pycache__',
                                     u'Lib/__pycache__',
                                     u'build',
                                     u'Doc/build'],
                        'filter': '*',
                        'include_hidden': False,
                        'recursive': True,
                        'watch_for_changes': True}]
proj.file-type = 'shared'
```

There sure is a `#!wing` line at the top, and it seems unlikely it's an accident. If it's not simply an accident, then the one effect of choosing to write the file that way is that simply `Misc/python-wing5.wpr` will have the same effect as `wing Misc/python-wing5.wpr`. (For example, that might open the given project in the Wing IDE.) But if the file isn't executable, then that effect is defeated. So, if the `#!` isn't simply an accident, then these files are indeed intended to be executable.

I don't think it's worth spending a lot of time thinking about these particular three files, though -- so if people don't find that convincing, then I'm happy to take that out and proceed with the other 8 files in this PR. Those other 8 files are all Python scripts.

gnprice:
(Also, the last time any of them was edited was 2013, in e28bb1505... and it looks like Wing itself [moved on to](https://wingware.com/news) version 6 in 2016, and version 7 this year, while the latest of these is for version 5. So it seems unlikely that anyone is actually using them today. In light of that I think probably the best thing to do with those three files `Misc/python-wing[345].wpr` is to delete them; they'll always be in the Git history for reference if needed in the future.

But until/unless we decide to do that I think it makes sense to mark them executable for the reasons above. And in any event, as I said, I'm most interested in the other files in this PR.)

epicfaace (approved):
Ok, that makes sense. It seems good to make the .wpr files executable, then.

gnprice:
Thanks @epicfaace for the review, and @Yhg1s for the merge!

PR #15358, opened 2019-08-21
bpo-37851: faulthandler allocates its stack on demand

vstinner:
The faulthandler module no longer allocates its alternative stack at
Python startup. Now the stack is only allocated at the first
faulthandler usage.

faulthandler no longer ignores memory allocation failure when
allocation the stack. sigaltstack() failure now raises an OSError
exception, rather than being ignored.

Moreover, the alternative stack is no longer used if sigaction() is
not available. In practice, sigaltstack() should only be available
when sigaction() is avaialble, so this change should have no effect
in practice.

faulthandler.dump_traceback_later() internal locks are now only
allocated at the first dump_traceback_later() call, rather than
always being allocated at Python startup.

https://bugs.python.org/issue37851

### DIFF
--- a/Misc/NEWS.d/next/Library/2019-08-21-13-43-04.bpo-37851.mIIfD_.rst
+++ b/Misc/NEWS.d/next/Library/2019-08-21-13-43-04.bpo-37851.mIIfD_.rst
@@ -1,0 +1,3 @@
+The :mod:`faulthandler` module no longer allocates its alternative stack at
+Python startup. Now the stack is only allocated at the first faulthandler
+usage.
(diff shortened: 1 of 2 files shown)


vstinner:
Travis CI job failed with a network failure:

"E: Failed to fetch http://archive.ubuntu.com/ubuntu/pool/main/b/babeltrace/libbabeltrace1_1.3.2-1_amd64.deb  Could not connect to apt.cache.travis-ci.com:80 (34.96.81.152), connection timed out"

I restarted the job.

tacaswell:
This causes python to segfault with

```python
import faulthandler
faulthandler.cancel_dump_traceback_later()
```

Also opening an issue on bpo.

<details>
used the following to bisect this:

```sh
make clean
make -j 8

./python -c "import faulthandler; faulthandler.cancel_dump_traceback_later()"
if [ $? -eq 139 ]; then
    echo "It crashed!"
    exit 1
fi


```

</details>

tacaswell:
xref: https://bugs.python.org/issue37933

PR #15366, opened 2019-08-21
Revert mode change that loses information in directory listings

skrah:


rhettinger (commented):
Thanks for noticing this.

PR #15369, opened 2019-08-21
bpo-9949: Call normpath first in realpath

zooba:
https://bugs.python.org/issue9949


zooba:
Symlink loops are evil...

PR #15380, opened 2019-08-22
[3.8] bpo-37482: Fix email address name with encoded words and specia…

bsiem:
…l chars (GH-14561)

Special characters in email address header display names are normally
put within double quotes. However, encoded words (=?charset?x?...?=) are
not allowed withing double quotes. When the header contains a word with
special characters and another word that must be encoded, the first one
must also be encoded.

In the next example, the display name in the From header is quoted and
therefore the comma is allowed; in the To header, the comma is not
within quotes and not encoded, which is not allowed and therefore
rejected by some mail servers.

From: "Foo Bar, France" <foo@example.com>
To: Foo Bar, =?utf-8?q?Espa=C3=B1a?= <foo@example.com>

https://bugs.python.org/issue37482
(cherry picked from commit df0c21ff46c5c37b6913828ef8c7651f523432f8)

Co-authored-by: bsiem <52461103+bsiem@users.noreply.github.com>

https://bugs.python.org/issue37482

### DIFF
--- a/Lib/email/_header_value_parser.py
+++ b/Lib/email/_header_value_parser.py
@@ -2722,6 +2722,9 @@ def _refold_parse_tree(parse_tree, *, policy):
             wrap_as_ew_blocked -= 1
             continue
         tstr = str(part)
+        if part.token_type == 'ptext' and set(tstr) & SPECIALS:
+            # Encode if tstr contains special characters.
+            want_encoding = True
         try:
             tstr.encode(encoding)
             charset = encoding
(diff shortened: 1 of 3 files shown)


epicfaace (approved):

ned-deily:
It looks like this backport had to be restarted and did not get merged like the 3.7 backport did.  With the 3.8.0b4 cutoff approaching, I'm going to manually push merge it for 3.8.

PR #15381, opened 2019-08-22
bpo-14050: Note that not all data can be sorted

rhettinger:
https://bugs.python.org/issue14050

### DIFF
--- a/Doc/tutorial/datastructures.rst
+++ b/Doc/tutorial/datastructures.rst
@@ -125,6 +125,12 @@ only modify the list have no return value printed -- they return the default
 ``None``. [1]_  This is a design principle for all mutable data structures in
 Python.
 
+Another thing you might notice is that not all data can be sorted.  For
+instance, ``[None, 'hello', 10]`` doesn't sort because integers can't be
+compared to strings and *None* can't be compared to other types.  Also,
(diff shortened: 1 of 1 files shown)


raulcd (commented):
This note on the documentation tutorial makes a lot of sense to me but shouldn't we say that it would fail when trying to sort?
When I read that this suggested me it wouldn't sort it but not that it would fail when trying to do so:
```
>>> a.sort()
TypeError: '<' not supported between instances of 'int' and 'NoneType'
```
Maybe changing this:
```
For instance, ``[None, 'hello', 10]`` doesn't sort because integers can't be compared to strings and *None* can't be compared to other types.
```
for
```
For instance, ``[None, 'hello', 10]`` will fail when trying to sort because integers can't be compared to strings and *None* can't be compared to other types.
```
Not sure if in that part of the tutorial is too early to say `will raise a TypeError Exception`.
Nonetheless I think is a pretty good addition to the tutorial.

rhettinger:
Thanks for the suggestion.  To keep the section affirmatively worded, I avoided talking about "failure".  And yes, it is too early in the text to go in to specific exceptions.  I just want to introduce the concept that not all orderings are defined.

raulcd (approved):

PR #15383, opened 2019-08-22
bpo-37913: document that __length_hint__ can return NotImplemented

jdemeyer:
CC @aeros167 

https://bugs.python.org/issue37913

### DIFF
--- a/Doc/reference/datamodel.rst
+++ b/Doc/reference/datamodel.rst
@@ -2144,8 +2144,10 @@ through the container; for mappings, :meth:`__iter__` should be the same as
 
    Called to implement :func:`operator.length_hint`. Should return an estimated
    length for the object (which may be greater or less than the actual length).
-   The length must be an integer ``>=`` 0. This method is purely an
-   optimization and is never required for correctness.
+   The length must be an integer ``>=`` 0. The return value may also be
+   *NotImplemented*, which is treated the same as if the ``__length_hint__``
+   method didn't exist at all. This method is purely an optimization and is
+   never required for correctness.
 
    .. versionadded:: 3.4
 


Doc/reference/datamodel.rst
```diff
@@ -2144,8 +2144,10 @@ through the container; for mappings, :meth:`__iter__` should be the same as
 
    Called to implement :func:`operator.length_hint`. Should return an estimated
    length for the object (which may be greater or less than the actual length).
-   The length must be an integer ``>=`` 0. This method is purely an
-   optimization and is never required for correctness.
+   The length must be an integer ``>=`` 0. The return value may also be
+   *NotImplemented*, which is treated the same as if the ``__length_hint__``
```
aeros167:
```suggestion
   :const:`NotImplemented`, which is treated the same as if the ``__length_hint__``
```
Let's add a Sphinx role so that users can click on the constant to be directed towards its documentation. This change will roughly convert *NotImplemented* -> [NotImplemented](https://docs.python.org/3/library/constants.html#NotImplemented). It might be helpful for users who aren't familiar with it.

This of course doesn't need be done for ``__length_hint__`` since this is within its dedicated section.

aeros (approved):
Thanks for the PR @jdemeyer!

I approve of the section added, I just have a minor formatting suggestion:

zooba:
@jdemeyer Could you apply the suggested change so we can merge? Or else select the "Allow edits from maintainers" checkbox on the right hand side and let us know so we can finish it off.

jdemeyer:
> Or else select the "Allow edits from maintainers" checkbox on the right hand side

It was already checked :-)

aeros167:
@zooba 
> Or else select the "Allow edits from maintainers" checkbox on the right hand side and let us know so we can finish it off.

As far as I'm aware, that option is supposed to allow anyone with write access to the repository to directly commit suggestions to the PR branch. But, it doesn't always work correctly in my experience. The GitHub interface may not be recognizing that I made an actual suggestion, since I chose "Approve" when I added my review instead of "Request Changes" or "Comment".

I'll try to do it again, but instead select "Request Changes" this time.

Doc/reference/datamodel.rst
aeros, line 2148:
```suggestion
   :const:`NotImplemented`, which is treated the same as if the ``__length_hint__``
```

aeros (changes requested):
Deleted my previous suggestion and changed the review option to "Request changes". That might allow the suggestion to be directly added to the PR branch.

zooba:
No, the original submitter has to select the checkbox on the right to allow maintainers to push changes to their branch. We think it changed from default-checked to default-unchecked at some point.

I'll just merge this and then apply your suggestion myself.

zooba:
Oh, just spotted Jeroen's comment in amongst all the rest of the boxes here (still not a fan of this UI...). Weird that it wasn't working, but it's done now so 🤷‍♂️

PR #15384, opened 2019-08-22
Define _PyTime_MIN/MAX with macros specific to the actual type of _PyTime_t

sir-sigurd:

### DIFF
--- a/Include/pytime.h
+++ b/Include/pytime.h
@@ -17,8 +17,8 @@ extern "C" {
    store a duration, and so indirectly a date (related to another date, like
    UNIX epoch). */
 typedef int64_t _PyTime_t;
-#define _PyTime_MIN PY_LLONG_MIN
-#define _PyTime_MAX PY_LLONG_MAX
+#define _PyTime_MIN INT64_MIN
+#define _PyTime_MAX INT64_MAX
 
 typedef enum {
     /* Round towards minus infinity (-inf).


aeros (commented):
Thanks for the PR @sir-sigurd.

This looks like a minor enough adjustment to warrant a ``skip issue`` and ``skip news``, particularly since it's modifying internal constants. I'll notify the datetime experts (although this might not be needed, since they were automatically added as reviewers).

/cc @abalkin @pganssle

vstinner:
> The principle of Chesterton's Fence applies here, so I think we need to figure out whether this was actually a mistake or a deliberate choice. Looks like @vstinner added these macros in this commit. Maybe he can comment as to why he chose PY_LLONG_MIN and PY_LLONG_MAX over INT64_MIN and INT64_MAX here.

There is no secret: it's a bug :-)

I don't recall the detail, but I recall that I wasn't sure about if using int64_t would cause any portability issue. There were very old discussions about using <stdint.h> and why PY_LONG_LONG was needed, etc.

I chose to attempt to use int64_t and see who is going to complain. Years later: nobody complains, so it seems like the (Python) World is ready for int64_t :-D

Maybe the first iteration of my change used "typedef long _PyTime_t;" and then I changed my mind.

vstinner:
If you are curious about _PyTime API, I wrote two articles about its history :-)

* https://vstinner.github.io/pytime.html
* https://vstinner.github.io/python37-perf-counter-nanoseconds.html

aeros167:
@vstinner Thanks for the detailed explanation and context!

> There were very old discussions about using <stdint.h> and why PY_LONG_LONG was needed, etc.

Is there still any use case for ``Py_LONG_LONG`` types at this point?

vstinner:
> Are there still any use cases for Py_LONG_LONG types at this point (within the Python C-API)?

long long is now always avaialble and PY_LONG_LONG is always defined as "long long". The macro is no more used in the Python code base, except in dict_get_version() of _testcapimodule.c. Maybe this function can be updated.

sir-sigurd:
@aeros167 there is already https://github.com/python/cpython/pull/15386.

PR #15385, opened 2019-08-22
bpo-27961: Replace PY_LLONG_MAX, PY_LLONG_MIN and PY_ULLONG_MAX with standard macros

sir-sigurd:

These ones
https://github.com/python/cpython/blob/c4106af38bbcb180725fe0d9478e6a11f7a5e7b9/Include/pytime.h#L20-L21
will be handled in #15384.


vstinner:
Please open a new issue on bugs.python.org for this change.

On Windows, PC/pyconfig.h does not define PY_LLONG_MIN as LLONG_MIN but _I64_MIN:

```
/* 64 bit ints are usually spelt __int64 unless compiler has overridden */
#ifndef PY_LONG_LONG
#       define PY_LONG_LONG __int64
#       define PY_LLONG_MAX _I64_MAX
#       define PY_LLONG_MIN _I64_MIN
#       define PY_ULLONG_MAX _UI64_MAX
#endif
```

Which Visual Studio versions support LLONG_MIN? I never recall what is the minimum Visual Studio version to build Python on Windows. My notes say "VS 2017" for Python 3.8 and newer: https://pythondev.readthedocs.io/windows.html#python-and-visual-studio-version-matrix

sir-sigurd:
> Which Visual Studio versions support LLONG_MIN?

I don't know, but it's used on Win here:
https://github.com/python/cpython/blob/c4106af38bbcb180725fe0d9478e6a11f7a5e7b9/Modules/_io/_iomodule.h#L90-L97

since https://github.com/python/cpython/commit/47ff0734b817d293ae84603efad2f3f508d85f78.

vstinner:
Ok. Visual Studio seems to be fine. Again, please open a new issue for this change.

vstinner:
Thanks @sir-sigurd !

PR #15388, opened 2019-08-22
bpo-27961: Remove leftovers from the times when long long wasn't required

sir-sigurd:


jdemeyer (approved):
Documentation-only change, looks good to me.

pablogsal:
Thanks, @sir-sigurd for the Patch!

PR #15390, opened 2019-08-22
bpo-37915: Fix comparison between tzinfo objects and timezone objects

pablogsal:
https://bugs.python.org/issue37915

### DIFF
--- a/Lib/test/datetimetester.py
+++ b/Lib/test/datetimetester.py
@@ -413,6 +413,9 @@ class TestTimeZone(unittest.TestCase):
                 with self.assertRaises(ValueError):
                     timezone(delta)
 
+    def test_comparison_with_tzinfo(self):
+        self.assertNotEqual(timezone.utc, tzinfo())
+        self.assertNotEqual(timezone(timedelta(hours=1)), tzinfo())
 
 #############################################################################
 # Base class for testing a particular aspect of timedelta, time, date and
--- a/Lib/test/datetimetester.py
+++ b/Lib/test/datetimetester.py
@@ -414,6 +414,8 @@ class TestTimeZone(unittest.TestCase):
                     timezone(delta)
 
     def test_comparison_with_tzinfo(self):
+        # Constructing tzinfo objects directly should not be done by users
+        # and serves only to check the bug described in bpo-37915
         self.assertNotEqual(timezone.utc, tzinfo())
         self.assertNotEqual(timezone(timedelta(hours=1)), tzinfo())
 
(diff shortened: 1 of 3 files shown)


Lib/test/datetimetester.py
```diff
@@ -413,6 +413,9 @@ def test_offset_boundaries(self):
                 with self.assertRaises(ValueError):
                     timezone(delta)
 
+    def test_comparison_with_tzinfo(self):
+        self.assertNotEqual(timezone.utc, tzinfo())
```
pganssle, line 419: I was going to say harumph about constructing a `tzinfo` object, but this is apparently done elsewhere in the tests, with this comment:

```
        # In order to allow subclasses to get pickled, the C implementation
        # wasn't able to get away with having __init__ raise
        # NotImplementedError
```

In any case, I think maybe a comment referencing `bpo-37915` would be a good idea.

Misc/NEWS.d/next/Library/2019-08-22-16-13-27.bpo-37915.xyoZI5.rst
```diff
@@ -0,0 +1,3 @@
+Fix a segmentation fault that appear when comparing instances of
```
pganssle:
```suggestion
Fix a segmentation fault that appeared when comparing instances of
```

Modules/_datetimemodule.c
```diff
@@ -32,6 +32,8 @@
 #define PyTZInfo_Check(op) PyObject_TypeCheck(op, &PyDateTime_TZInfoType)
 #define PyTZInfo_CheckExact(op) (Py_TYPE(op) == &PyDateTime_TZInfoType)
 
+#define PyTimezone_Check(op) PyObject_TypeCheck(op, &PyDateTime_TimeZoneType)
```
pganssle, line 35: I believe that not having a `PyTimezone_Check` was a deliberate omission by @abalkin and that `timezone` is not intended to be subclassed.

I guess this isn't public anyway, but considering that it's not, I think there's no need to define both `PyTimezone_Check` and `PyTimezone_CheckExact`, just define whichever one you need. I'm thinking `PyTimezone_CheckExact` is most in keeping with the "`timezone` shouldn't be subclassed" philosophy (we can always add subclass support later, after all, while taking it away would be a PITA).

pganssle (changes requested):

Modules/_datetimemodule.c
pablogsal: ~`PyTZInfo_Check` was used elsewhere and I prefer that this PR does not touch those functions. I can inline the macro if you prefer~

Sorry, I totally misread the comment. I have removed the other macro.

pablogsal:

tirkarthi:
3.7 also has this change with https://github.com/python/cpython/commit/6ed20e54e4c110e9adcfb70aba85310625e3edb4 and would require a backport. Thanks Pablo for the fix.

serhiy-storchaka (approved):

PR #5944, opened 2018-03-01
bpo-33018: Improve issubclass() error checking and message.

jab:
If you try something like `issubclass('not a class', str)`, you get a helpful error message that immediately clues you in on what you did wrong:

```python
>>> issubclass('not a class', str)
TypeError: issubclass() arg 1 must be a class
```

("AHA! I meant *isinstance* there. Thanks, friendly error message!")

But if you try this with some ABC, the error message is much less friendly!
```python
>>> from some_library import SomeAbc
>>> issubclass('not a class', SomeAbc)
Traceback (most recent call last):
  File "<stdin>", line 1, in <module>
  File "/usr/local/Cellar/python3/3.6.4_2/Frameworks/Python.framework/Versions/3.6/lib/python3.6/abc.py", line 230, in __subclasscheck__
    cls._abc_negative_cache.add(subclass)
  File "/usr/local/Cellar/python3/3.6.4_2/Frameworks/Python.framework/Versions/3.6/lib/python3.6/_weakrefset.py", line 84, in add
    self.data.add(ref(item, self._remove))
TypeError: cannot create weak reference to 'str' object
```

("WTF just went wrong?" Several more minutes of head-scratching ensues. Maybe a less experienced Python programmer who hits this hasn't seen weakrefs before and gets overwhelmed, maybe needlessly proceeding down a deep rabbit hole before realizing no knowledge of weakrefs was required to understand what they did wrong.)

Or how about this example:

```python
>>> from collections import Reversible
>>> issubclass([1, 2, 3], Reversible)
Traceback (most recent call last):
  File "<stdin>", line 1, in <module>
  File "/usr/local/Cellar/python3/3.6.4_2/Frameworks/Python.framework/Versions/3.6/lib/python3.6/abc.py", line 207, in __subclasscheck__
    ok = cls.__subclasshook__(subclass)
  File "/usr/local/Cellar/python3/3.6.4_2/Frameworks/Python.framework/Versions/3.6/lib/python3.6/_collections_abc.py", line 305, in __subclasshook__
    return _check_methods(C, "__reversed__", "__iter__")
  File "/usr/local/Cellar/python3/3.6.4_2/Frameworks/Python.framework/Versions/3.6/lib/python3.6/_collections_abc.py", line 73, in _check_methods
    mro = C.__mro__
AttributeError: 'list' object has no attribute '__mro__'
```

Here you don't even get the same *type* of error (`AttributeError` rather than `TypeError`), which seems unintentionally inconsistent.

[...]

PR #15394, opened 2019-08-22
Less harsh implementation for issubclass()

raamana:
Apologies if this does not fully conform to the contrib guidelines (please close it if so), but I think this is a trivial change, and likely does not need any discussion on bugs.python.org.

When trying to write tests for one of my libraries, I wanted to ensure a certain input param is a proper subclass, and to ensure it fails for invalid input, I ran tests with not-class inputs like a string etc. This has been discussed in #5944 and #6188, where the decision was raise an error requiring a class-type input. I think this could be made less *harsh* by simply returning False, while issuing a warning.

I'm not a cpython expert and not sure warnings module is available in that environment, I put it in there to give an idea of the ideal behaviour of issubclass I'd expect. Apologies if this does not fully conform to the contrib guidelines -please close it if so. Thank you for the consideration.

### DIFF
--- a/Lib/_py_abc.py
+++ b/Lib/_py_abc.py
@@ -1,5 +1,5 @@
 from _weakrefset import WeakSet
-
+from warnings import warn
 
 def get_cache_token():
     """Returns the current ABC cache token.
@@ -108,7 +108,8 @@ class ABCMeta(type):
     def __subclasscheck__(cls, subclass):
         """Override for issubclass(subclass, cls)."""
         if not isinstance(subclass, type):
-            raise TypeError('issubclass() arg 1 must be a class')
+            warn('issubclass() arg 1 is not a class. Returning False.')
+            return False
         # Check cache
         if subclass in cls._abc_cache:
             return True


the-knights-who-say-ni:

tirkarthi:
This could result in confusion where it's not actually a subclass returning False and also an invalid call returning False. Warnings may not be visible in all environments and configurations and probably ignored. It's better to have an error and if you want to ignore the error then you can catch the `TypeError` and silence as needed. This would require a discussion in https://bugs.python.org or python-dev mailing list since this is a backwards incompatible change if you would like to proceed further.

Thanks

raamana:
The confusion isn’t harmful and does not lead to any incorrect result as any input that’s not a class also is not actually a subclass of the second arg. Throwing an error here or adding try/catch code downstream is unnecessarily harsh. 

I understand the backwards incompatibility issue, but perhaps have a deprecation cycle? Please go ahead pass it along to bugs.python.org. Happy to help in minor ways, I won’t be able to lead the debate or dev.

jdemeyer:
> does not need any discussion on bugs.python.org.

It's a functional change, it certainly needs an issue.

matrixise:
Hi all and @raamana 

Thank you for your contribution but I am going to close this PR, the CLA is not signed by @raamana. @raamana once you will have signed the CLA and this one will be in https://bugs.python.org you are free to re-open your PR and we will process it.

Also, because there is a change in the algo, as @jdemeyer as suggested, I propose to you to open an issue in BPO (bugs.python.org) and discuss that issue.

Thank you so much,

PR #15396, opened 2019-08-22
bpo-28556: Add a regression test to typing

ilevkivskyi:
This adds a regression test for the issue found in the Python 2 backport, see https://github.com/python/typing/issues/656

https://bugs.python.org/issue28556


gvanrossum (approved):
Thanks!

PR #15397, opened 2019-08-22
bpo-30550: Clarify JSON orderering guarantees

rhettinger:
https://bugs.python.org/issue30550


lebigot:
Thanks! This perfectly addresses my post about ordering guarantees.

PR #15400, opened 2019-08-22
bpo-37953: Fix ForwardRef equality checks

plokmijnuhby:
Ideally if we stick a ForwardRef in a dictionary we would like to reliably be able to get it out again.

### DIFF
--- a/Lib/test/test_typing.py
+++ b/Lib/test/test_typing.py
@@ -2336,6 +2336,65 @@ class ForwardRefTests(BaseTestCase):
         self.assertEqual(fr, typing.ForwardRef('int'))
         self.assertNotEqual(List['int'], List[int])
 
+    def test_forward_equality_gth(self):
+        c1 = typing.ForwardRef('C')
+        c1_gth = typing.ForwardRef('C')
+        c2 = typing.ForwardRef('C')
+        c2_gth = typing.ForwardRef('C')
+
+        class C:
+            pass
+        def foo(a: c1_gth, b: c2_gth):
+            pass
+
+        self.assertEqual(get_type_hints(foo, globals(), locals()), {'a': C, 'b': C})
+        self.assertEqual(c1, c2)
+        self.assertEqual(c1, c1_gth)
+        self.assertEqual(c1_gth, c2_gth)
+        self.assertEqual(List[c1], List[c1_gth])
+        self.assertNotEqual(List[c1], List[C])
+        self.assertNotEqual(List[c1_gth], List[C])
+        self.assertEquals(Union[c1, c1_gth], Union[c1])
+        self.assertEquals(Union[c1, c1_gth, int], Union[c1, int])
+
+    def test_forward_equality_hash(self):
+        c1 = typing.ForwardRef('int')
+        c1_gth = typing.ForwardRef('int')
+        c2 = typing.ForwardRef('int')
+        c2_gth = typing.ForwardRef('int')
+
+        def foo(a: c1_gth, b: c2_gth):
+            pass
+        get_type_hints(foo, globals(), locals())
+
+        self.assertEqual(hash(c1), hash(c2))
+        self.assertEqual(hash(c1_gth), hash(c2_gth))
+        self.assertEqual(hash(c1), hash(c1_gth))
+
+    def test_forward_equality_namespace(self):
+        class A:
+            pass
+        def namespace1():
+            a = typing.ForwardRef('A')
+            def fun(x: a):
+                pass
+            get_type_hints(fun, globals(), locals())
+            return a
+
+        def namespace2():
+            a = typing.ForwardRef('A')
+
+            class A:
+                pass
+            def fun(x: a):
+                pass
+
+            get_type_hints(fun, globals(), locals())
+            return a
+
+        self.assertEqual(namespace1(), namespace1())
+        self.assertNotEqual(namespace1(), namespace2())
+
     def test_forward_repr(self):
         self.assertEqual(repr(List['int']), "typing.List[ForwardRef('int')]")
 
@@ -2354,6 +2413,64 @@ class ForwardRefTests(BaseTestCase):
 
         self.assertEqual(get_type_hints(foo, globals(), locals()),
                          {'a': Tuple[T]})
+
+    def test_forward_recursion_actually(self):
+        def namespace1():
+            a = typing.ForwardRef('A')
+            A = a
+            def fun(x: a): pass
+
+            ret = get_type_hints(fun, globals(), locals())
+            return a
+
+        def namespace2():
+            a = typing.ForwardRef('A')
+            A = a
+            # class A: pass
+            def fun(x: a): pass
+
+            ret = get_type_hints(fun, globals(), locals())
+            return a
+
+        def cmp(o1, o2):
+            return o1 == o2
+
+        r1 = namespace1()
+        r2 = namespace2()
+        assert r1 is not r2
+        self.assertRaises(RecursionError, cmp, r1, r2)
+
+    def test_union_forward_recursion(self):
+        ValueList = List['Value']
+        Value = Union[str, ValueList]
+
+        class C:
+            foo: List[Value]
+        class D:
+            foo: Union[Value, ValueList]
+        class E:
+            foo: Union[List[Value], ValueList]
+        class F:
+            foo: Union[Value, List[Value], ValueList]
+
+        self.assertEqual(get_type_hints(C, globals(), locals()), get_type_hints(C, globals(), locals()))
+        self.assertEqual(get_type_hints(C, globals(), locals()),
+                         {'foo': List[Union[str, List[Union[str, List['Value']]]]]})
+        self.assertEqual(get_type_hints(D, globals(), locals()),
+                         {'foo': Union[str, List[Union[str, List['Value']]]]})
+        self.assertEqual(get_type_hints(E, globals(), locals()),
+                         {'foo': Union[
+                             List[Union[str, List[Union[str, List['Value']]]]],
+                             List[Union[str, List['Value']]]
+                         ]
+                          })
+        self.assertEqual(get_type_hints(F, globals(), locals()),
+                         {'foo': Union[
+                             str,
+                             List[Union[str, List['Value']]],
+                             List[Union[str, List[Union[str, List['Value']]]]]
+                         ]
+                          })
 
     def test_callable_forward(self):
 
@@ -2734,17 +2851,6 @@ class GetTypeHintTests(BaseTestCase):
                          {'z': ClassVar[CSub], 'y': int, 'b': int,
                           'x': ClassVar[Optional[B]]})
         self.assertEqual(gth(G), {'lst': ClassVar[List[T]]})
-        
-    def test_get_type_hints_retains_forward_equality(self):
-        fr = typing.ForwardRef('int')
-        hsh = hash(fr)
-        
-        def foo(a: fr):
-            pass
-        
-        gth(foo)
-        self.assertEqual(fr, typing.ForwardRef('int'))
-        self.assertEqual(hash(fr), hsh)
 
 
 class GetUtilitiesTestCase(TestCase):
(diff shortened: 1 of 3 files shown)


gvanrossum (commented):
This looks reasonable, but let's make sure @ilevkivskyi agrees.

Can you add a test (preferably one that shows what's not working without the fix)?

It's fairly unorthodox to have a PR without a bug in the tracker -- can you create a bug describing the problem and link this to it? Also please add a news item.

Lib/typing.py
```diff
@@ -524,11 +524,10 @@ def _evaluate(self, globalns, localns):
     def __eq__(self, other):
         if not isinstance(other, ForwardRef):
             return NotImplemented
-        return (self.__forward_arg__ == other.__forward_arg__ and
-                self.__forward_value__ == other.__forward_value__)
+        return self.__forward_arg__ == other.__forward_arg__
 
     def __hash__(self):
-        return hash((self.__forward_arg__, self.__forward_value__))
+        return hash((self.__forward_arg__,))
```
brandtbucher: Is this tuple still needed here?

gvanrossum: Good eye -- it is not.

ilevkivskyi:
TBH I don't understand what problem is being solved. This definitely requires tests.

gvanrossum:
The problem seems clear enough to me: The `__hash__()` and `__eq__()` operators compute a value that depends on `__forward_value__`, but it's possible to use a `ForwardRef` as a dict key before the latter field is assigned (it is initialized to `None`). If then (through an alias) its `_evaluate()` is called, which assigns a non-None value to `__forward_value__`, the hash and equality *change*, which violates the rules for hashable objects and confuses the dict lookup operation so the lookup will fail.

I'm not clear however on *why* someone would use a `ForwardRef` as a dict key. Maybe @plokmijnuhby can explain their use case?

ilevkivskyi:
OK, I see how the current status is wrong. On the other hand, just removing the value for evaluated forward references will make two references referring to different classes with the same name equal.

I think that the first one is more important, so I am fine with just removing the `__forward_value__` from `__hash__()` and `__eq__()`.

Lib/typing.py
```diff
@@ -486,7 +486,7 @@ def open_helper(file: str, mode: MODE) -> str:
     """)
 
 
-class ForwardRef(_Final, _root=True):
+class ForwardRef(_Final, _Immutable, _root=True):
```
ilevkivskyi: Why is this needed? This will preserve _identity_ (not just value) after copying.

ilevkivskyi (commented):
Mostly looks good, just one question.

Please also add a `NEWS` item.

hongweipeng:
I try a test in this PR:
```
from typing import ForwardRef, get_type_hints

def namespace1():
    a = ForwardRef('A')

    class A: pass
    def fun(x: a): pass

    get_type_hints(fun, globals(), locals())
    return hash(a)

def namespace2():
    a = ForwardRef('A')

    class A: pass
    def fun(x: a): pass

    get_type_hints(fun, globals(), locals())
    return hash(a)

print(namespace1() == namespace2())
```
Output `True` in this PR and output `False` in master branch. 

I'm afraid of removing the `__forward_value__ ` can break a legitimate code.

brandtbucher:
> I'm afraid of removing the `__forward_value__ ` can break a legitimate code.

@hongweipeng, do you know of any code that relies on this behavior? This seems like a pretty clear bugfix to me... `True` seems correct here!

hongweipeng:
@brandtbucher Yeah, I want to use `__eq__` to remove duplicate item instead of using `__hash__`.

Now I think that `__eq__` is only used for `ForwardRef`, while others keep the original `__hash__` way.

ilevkivskyi:
> Now I think that `__eq__` is only used for ForwardRef, while others keep the original `__hash__` way.

TBH I don't really like making `__hash__()` and `__eq__()` inconsistent. There are however precedents, for example `hash('abc') == hash(b'abc')`is `True` while `'abc' == b'abc'` is `False`. So a possible thing I could imagine is keeping only `__forward_arg__`, but also take into account `__forward_value__` in `__eq__()` if it is not `None` for `self` and `other`.

IOW I am fine with
> I think combining the `__eq__` from #15628 with the `__hash__` from this PR makes the most sense.

hongweipeng:
Removing `__forward_value__` can solve the recursion problem incidentally [issue37806](https://bugs.python.org/issue37806).

ilevkivskyi:
> Removing `__forward_value__` can solve the recursion problem incidentally.

Do you mean removing it from `__hash__` or from both places?

Generally, I would propose to keep only one PR open, because currently the discussion is scattered between two PRs. @plokmijnuhby Are you interested in continuing to work on this PR?

hongweipeng:
@ilevkivskyi Yes, we can remove `__forward_value__` in `__hash__` and keep it in `__eq__`. In this case, an actually recursive object will also throw an `RecursionError` normally.

plokmijnuhby:
@ilevkivskyi I'm happy to keep working on it, but I'm not quite sure how where we want to take this. It seems like there are still a few ideas bouncing around here. I mean, in theory we could remove the caching thing entirely, and just perform a lookup of the class name in the function's __globals__ whenever we need to evaluate a forward reference.

aeros167:
I'm not entirely certain if this is something we'd consider to be a behavioral changing enhancement or if the current behavior is an actual bug. For now, I'll categorize it as enhancement but feel free to change it.

Lib/typing.py
```diff
@@ -524,11 +524,10 @@ def _evaluate(self, globalns, localns):
     def __eq__(self, other):
         if not isinstance(other, ForwardRef):
             return NotImplemented
-        return (self.__forward_arg__ == other.__forward_arg__ and
-                self.__forward_value__ == other.__forward_value__)
+        return self.__forward_arg__ == other.__forward_arg__
```
brandtbucher, line 530: So it seems best to keep the check for forward value, contingent upon them both existing:

```suggestion
        if self.__forward_evaluated__ and other.__forward_evaluated__:
            return (self.__forward_arg__ == other.__forward_arg__ and
                    self.__forward_value__ == other.__forward_value__)
        return self.__forward_arg__ == other.__forward_arg__
```

I think that this, combined with the new `__hash__`, is probably the most useful and least surprising implementation.

Lib/typing.py
brandtbucher:
```suggestion
        return hash(self.__forward_arg__)
```

brandtbucher (changes requested):

plokmijnuhby:
Okay, here's my thoughts on forward references. @hongweipeng's code example doesn't make much sense to me because users are unlikely to evaluate references right after creating them, i.e. in the same calling context, but much more likely to have someone else's code evaluate them a little later, in a completely different calling context. The trouble is that we have no good way of determining the correct context after we've left it.

There are several possible solutions to this:
- WONTFIX the issue and ignore it.
- Detect the calling context the moment a forward reference is created, through `sys._getframe` hacks. Not a brilliant idea, particularly since using the form `def foo(a: 'Bar')` to use strings as forward references means we'd also have to do some some sort of monkey patch to `str`, so let's avoid that one.
- Changes to the parser itself. This would probably slow down the whole python system.
- A change to how we think about forward references. Forward references could be considered to have no meaning unless they have an obvious context available, eg if globals and locals were passed to get_type_hints. This would mean that a forward reference is basically just a string that can be executed in a fancy way, so it makes sense to mark it as immutable. This is the approach I was leaning towards.

Lib/typing.py
plokmijnuhby: Well, it's not strictly needed, but there's no reason that two references to the same thing should refer to different objects.

ilevkivskyi: > Well, it's not strictly needed

Then don't do this. This can break things in unexpected ways. Also this "thing" has internal state that does change.

ilevkivskyi:
So we now have two competing almost identical PRs: this one and https://github.com/python/cpython/pull/15650. This one was first, but the other has more tests. I think the best way is to polish this one, so I am going to close the alternative one.

Misc/NEWS.d/next/Library/2019-09-06-17-40-34.bpo-37953.db5FQq.rst
```diff
@@ -0,0 +1,3 @@
+Forward references in the typing module now have different hash and eq methods. References will now compare equal properly if only one of them has been evaluated and the other hasn't. Hash values are now consistent across the lifetime of a forward reference.
```
ilevkivskyi: Use `` `__hash__` `` and `` `__eq__` `` instead of just `hash` and `eq`. Also this news item is too long, couple short sentences (or even one) should be enough.

Lib/test/test_typing.py
```diff
@@ -2734,6 +2734,17 @@ def test_get_type_hints_ClassVar(self):
...
+    def test_get_type_hints_retains_forward_equality(self):
+        fr = typing.ForwardRef('int')
+        hsh = hash(fr)
+        
+        def foo(a: fr):
+            pass
+        
+        gth(foo)
+        self.assertEqual(fr, typing.ForwardRef('int'))
+        self.assertEqual(hash(fr), hsh)
```
ilevkivskyi: Please add few more tests, e.g. checking also equality and hashes between references that got evaluated to different things etc.

Misc/NEWS.d/next/Library/2019-09-06-17-40-34.bpo-37953.db5FQq.rst
plokmijnuhby: I changed it, is that better?

Lib/test/test_typing.py
```diff
@@ -2355,6 +2414,64 @@ def foo(a: Tuple['T']):
...
+
+            ret = get_type_hints(fun, globals(), locals())
+            return a
+
+        def cmp(o1, o2):
+            return o1 == o2
+
+        r1 = namespace1()
+        r2 = namespace2()
+        assert r1 is not r2
```
hongweipeng: It would be better to replace `assert r1 is not r2` with `self.assertIsNot(r1, r2)`.

Lib/test/test_typing.py
```diff
@@ -2355,6 +2414,64 @@ def foo(a: Tuple['T']):
...
+            A = a
+            def fun(x: a): pass
+
+            ret = get_type_hints(fun, globals(), locals())
+            return a
+
+        def namespace2():
+            a = typing.ForwardRef('A')
+            A = a
+            # class A: pass
```
hongweipeng: Remove this line `# class A: pass`.

Lib/test/test_typing.py
aeros: Agreed, bare ``assert`` expression shouldn't be used in the tests.

Misc/NEWS.d/next/Library/2019-09-06-17-40-34.bpo-37953.db5FQq.rst
```diff
@@ -0,0 +1 @@
+Forward references in the typing module now have improved __hash__ and __eq__ methods.
```
aeros: Could we slightly modify this news entry to be more specific to the actual class the methods are being added to? Also fixes the formatting to use codeblocks for ``__hash__`` and ``__eq__`` as well as adding a Sphinx role to ``ForwardReferences`` and ``typing``. This allows readers to click on them to be linked to the corresponding documentation. 

```suggestion
In :mod:`typing`, improved the ``__hash__`` and ``__eq__`` methods for
:class:`ForwardReferences`.
```

Note that this change also requires this to be split into a second line, which I've included in the suggestion. This is because it goes above the 80 character width limit for reST.

aeros167:
Also, thanks for the PR and welcome @plokmijnuhby. (:

ilevkivskyi (approved):
Thanks for updates! LGTM now.

ilevkivskyi:
Thanks everyone for help!

@plokmijnuhby To make things consistent, you might also want to add this fix to the backport at https://github.com/python/typing.

rfrowe:
Are there plans to backport this fix to previous versions of cpython aside from 3.8? I can confirm the issue exists in 3.7.6

PR #15405, opened 2019-08-22
test_gdb: use unittest test discovery

vstinner:
Replace test.support.run_unittest() with unittest.main().

Remove also an unused variable.


vstinner:
cc @jdemeyer

PR #15406, opened 2019-08-22
bpo-12634: Clarify an awkward section of the tutorial

rhettinger:
https://bugs.python.org/issue12634

### DIFF
--- a/Doc/tutorial/classes.rst
+++ b/Doc/tutorial/classes.rst
@@ -475,12 +475,20 @@ Random Remarks
 
 .. These should perhaps be placed more carefully...
 
-Data attributes override method attributes with the same name; to avoid
-accidental name conflicts, which may cause hard-to-find bugs in large programs,
-it is wise to use some kind of convention that minimizes the chance of
-conflicts.  Possible conventions include capitalizing method names, prefixing
-data attribute names with a small unique string (perhaps just an underscore), or
-using verbs for methods and nouns for data attributes.
+If the same attribute name occurs in both an instance and in a class,
+then attribute lookup prioritizes the instance::
+
+    >>> class Warehouse:
+            purpose = 'storage'
+            region = 'west'
+
+    >>> w1 = Warehouse()
+    >>> print(w1.purpose, w1.region)
+    storage west
+    >>> w2 = Warehouse()
+    >>> w2.region = 'east'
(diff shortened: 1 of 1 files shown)


eamanu (approved):
I think is a good improvement.

Doc/tutorial/classes.rst
```diff
@@ -475,12 +475,20 @@ Random Remarks
 
 .. These should perhaps be placed more carefully...
 
-Data attributes override method attributes with the same name; to avoid
-accidental name conflicts, which may cause hard-to-find bugs in large programs,
-it is wise to use some kind of convention that minimizes the chance of
-conflicts.  Possible conventions include capitalizing method names, prefixing
-data attribute names with a small unique string (perhaps just an underscore), or
-using verbs for methods and nouns for data attributes.
```
terryjreedy, line 483: Wow.  At first I was confused.  If one defines module, class, or instance attributes with the same name, the second wins.  Then the first two advice items are wrong, the third correct, but misses the realistic case of noun data attribute duplication on class and instance.  Your example covers that nicely. Please apply and backport.

terryjreedy (approved):

Doc/tutorial/classes.rst
```diff
@@ -475,12 +475,20 @@ Random Remarks
...
+If the same attribute name occurs in both an instance and in a class,
+then attribute lookup prioritizes the instance::
```
terryjreedy, line 479: Addendum: The exception is with implicit lookup of (reserved) dunder names.
```
class C:
    def __len__(self): return 0

c = C()
c.__len__ = False
print(c.__len__, len(c))
# False, 0
```
Do you want to mention this, or qualify with 'direct attribute lookup'?  This is something that occasionally trips up newbies when they, for instance, add (in this instance) a .__len__ function to an instance and find that indirect access via len() still gets the class method.

terryjreedy (changes requested):

Doc/tutorial/classes.rst
rhettinger: I think that level of detail is more suited for the language reference.  In a tutorial, it just invites confusion.  Special handling of dunder methods, data descriptors, and non-data descriptors is an advanced topic.

terryjreedy: Agreed.

PR #15407, opened 2019-08-23
bpo-30826: Improve control flow examples

rhettinger:
https://bugs.python.org/issue30826

### DIFF
--- a/Doc/tutorial/controlflow.rst
+++ b/Doc/tutorial/controlflow.rst
@@ -66,20 +66,20 @@ they appear in the sequence.  For example (no pun intended):
    window 6
    defenestrate 12
 
-If you need to modify the sequence you are iterating over while inside the loop
-(for example to duplicate selected items), it is recommended that you first
-make a copy.  Iterating over a sequence does not implicitly make a copy.  The
-slice notation makes this especially convenient::
+Code that modifies a collection while iterating over that same collection can
+be tricky to get right.  Instead, it is usually more straight-forward to loop
+over a copy of the collection or to create a new collection::
 
-   >>> for w in words[:]:  # Loop over a slice copy of the entire list.
-   ...     if len(w) > 6:
-   ...         words.insert(0, w)
-   ...
-   >>> words
-   ['defenestrate', 'cat', 'window', 'defenestrate']
+    # Strategy:  Iterate over a copy
+    for user, status in users.copy().items():
+        if status == 'inactive':
+            del users[user]
 
-With ``for w in words:``, the example would attempt to create an infinite list,
-inserting ``defenestrate`` over and over again.
+    # Strategy:  Create a new collection
+    active_users = {}
+    for user, status in users.items():
+        if status == 'active':
+            active_users[user] = status
 
 
 .. _tut-range:


Doc/tutorial/controlflow.rst
```diff
@@ -66,20 +66,20 @@ they appear in the sequence.  For example (no pun intended):
...
-   ...     if len(w) > 6:
-   ...         words.insert(0, w)
-   ...
-   >>> words
-   ['defenestrate', 'cat', 'window', 'defenestrate']
-
-With ``for w in words:``, the example would attempt to create an infinite list,
-inserting ``defenestrate`` over and over again.
+Code that modifies a collection while iterating over that same collection can
+be tricky to get right.  Instead, it is usually more straight-forward to loop
```
terryjreedy, line 70: Consider adding, after 'right.', "For instance, one may accidentally create an infinite list."

Doc/tutorial/controlflow.rst
```diff
@@ -66,20 +66,20 @@ they appear in the sequence.  For example (no pun intended):
...
+over a copy of the collection or to create a new collection::
```
terryjreedy, line 71: The three lines are clearer and more correct in giving both strategies.

Doc/tutorial/controlflow.rst
```diff
@@ -66,20 +66,20 @@ they appear in the sequence.  For example (no pun intended):
...
+
+    # Strategy:  Iterate over a copy
+    for user, status in users.copy().items():
```
terryjreedy, line 74: The problem I have with a dict example is that modifying a dict while iterating is not 'tricky' (for the user), but rather is so tricky (for the implementation) that it is flat out prohibited.

Doc/tutorial/controlflow.rst
```diff
@@ -66,20 +66,20 @@ they appear in the sequence.  For example (no pun intended):
...
+        if status == 'inactive':
+            del users[user]
+
+    # Strategy:  Create a new collection
+    active_users = {}
+    for user, status in users.items():
+        if status == 'active':
+            active_users[user] = status
```
terryjreedy, line 82: This is correct yet not idiomatic.
`active_users = {user:status for user, status in users.items if status == 'active'}`
For the previous example,
`long_words = [w for w in words if len(w) > 6]`
I presume comprehensions have not yet been discussed, but if they are later, I would add a forward reference.  Perhaps "See xxx for how to do this more compactly."

terryjreedy (commented):
An improvement worth merging as is, with suggestions for maybe making it even better.

Doc/tutorial/controlflow.rst
rhettinger: * Comprehensions haven't been discussed yet. 
* Also, they are a distraction from the key point, "don't mutate while iterating"
* An unrolled loop is still considered valid, correct Python.  Comprehensions are not required.
* I prefer to keep this example as-is.  It is the simplest example of filtering an existing structure into a new structure rather than deleting entries.
* IMO, forward references are distracting and make the tutorial harder to read.
* At the point in the tutorial, we can't drown a person.  This section is the *first* introduction to a for-loop.

Doc/tutorial/controlflow.rst
rhettinger: * The dict example is especially important because it teaches the copy() technique.  Otherwise, all a user knows is that they've got an error.  This shows what to do about it.

* FWIW, dicts do allow value mutations while iterating.

* https://imgur.com/gallery/eEsd4V1

Doc/tutorial/controlflow.rst
rhettinger: You're happy with the proposed wording, correct?

Doc/tutorial/controlflow.rst
rhettinger: The devguide suggests we not go down that route.  Generally, we keep the text affirmatively worded and not full of warnings and worries.  Otherwise, the implied message is that Python is fragile and bad things can happen to you if you make a mistake.

Doc/tutorial/controlflow.rst
terryjreedy: Yes.  There is no checkbox for "(*) Approve either as is as a definite and acceptable improvement over the status quo or after possible further improvements based on my comments -- you decide."

In this case, you considered each suggestion and and gave a short coherent reason why not, one which I can consider when writing or reviewing other doc patches.

I appreciate you tackling some of these lingering doc issues which have seemingly been blocked by the quest for perfection even when improvement is clearly possible.

PR #15408, opened 2019-08-23
bpo-14112: Add link for users to explore shallowness in greater depth

rhettinger:
<pun intended>

https://bugs.python.org/issue14112

### DIFF
--- a/Doc/tutorial/introduction.rst
+++ b/Doc/tutorial/introduction.rst
@@ -394,7 +394,9 @@ indexed and sliced::
    [9, 16, 25]
 
 All slice operations return a new list containing the requested elements.  This
-means that the following slice returns a new (shallow) copy of the list::
+means that the following slice returns a `shallow copy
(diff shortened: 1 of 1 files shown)


terryjreedy (changes requested):
I am strongly opposed to adding this link.  The site requests that one 'sign in' with either Facebook or Google.  We should not promote them and their omni-surveillance.  The beginning of the text is a mutilated copy of parts of our copy module doc, https://docs.python.org/3/library/copy.html, by someone who I doubt is a native English speaker.  I could not stomach reading all of it.  A reference to the original would be better.  The inferior copy, but not the original, says, wrongly, " In Python, there are two ways to create copies :"  How confusing in the context of giving a third for lists.  (and list.copy is a fourth).

If you think the diagrams are helpful to beginners, we should add one or more to our doc.

rhettinger:
Hmm, I didn't see the sign in request because I'm already signed in.

PR #15415, opened 2019-08-23
bpo-37926: Fix PySys_SetArgvEx(0, NULL, 0)

vstinner:
empty_argv is no longer static in Python 3.8, but it was declared in
a temporary scope, whereas argv kept a reference to it. Define
empty_argv in PySys_SetArgvEx() body, to ensure that it remains valid
for the whole lifetime of the PySys_SetArgvEx() call.

Previously, empty_argv memory (allocated on the stack) was reused by
make_sys_argv() code which is inlined when using gcc -O3.

https://bugs.python.org/issue37926

### DIFF
--- a/Misc/NEWS.d/next/C API/2019-08-23-11-35-55.bpo-37926.hnI5IQ.rst
+++ b/Misc/NEWS.d/next/C API/2019-08-23-11-35-55.bpo-37926.hnI5IQ.rst
@@ -1,0 +1,1 @@
+Fix a crash in ``PySys_SetArgvEx(0, NULL, 0)``.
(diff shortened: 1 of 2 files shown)


pablogsal:
Does this need backport to 3.7? The issue says 3.7 and 3.8

vstinner:
I'm working on a backport to 3.8.

vstinner:
> Does this need backport to 3.7? The issue says 3.7 and 3.8

Python 3.7 is not affected: it uses "static wchar_t *empty_argv[1] = {L""};".

vstinner:
Moreover, the issue says "This works in 3.7 but no longer does in 3.8" :-)

pablogsal:
> Moreover, the issue says "This works in 3.7 but no longer does in 3.8" :-)

Sorry, I just need more coffee (Rode "between Python 3.7 and 3.8" and registered 3.7 and 3.8)

PR #15424, opened 2019-08-23
bpo-23878: Remove an unneeded fseek() call in _Py_FindEnvConfigValue()

ZackerySpytz:
https://bugs.python.org/issue23878


Python/pathconfig.c
```diff
@@ -741,7 +741,9 @@ _Py_FindEnvConfigValue(FILE *env_file, const wchar_t *key,
     char buffer[MAXPATHLEN * 2 + 1];  /* allow extra for key, '=', etc. */
     buffer[Py_ARRAY_LENGTH(buffer)-1] = '\0';
```
vstinner, line 743: Please update the function comment: explain that it returns 1 and copy it into value if found, or return 0 on error or if not found. Something like that.

I'm not sure why key is a parameter: its value is always "home". Maybe write a following PR to remove the key parameter? (need to update getpath.c and getpathp.c)

vstinner: Note: I'm not sure why fseek(0) is called: we just opened the file in read mode, not in append mode, the file cursor should already be 0.

ZackerySpytz: Thank you very much for your review, @vstinner. I have removed the `fseek()` call -- it indeed seems to be unneeded.
I will create a pull request to remove the key parameter.

vstinner:
I don't think that it's worth it to backport this change to Python 3.7 or 3.8.

PR #15439, opened 2019-08-23
bpo-37937: Mention frame.f_trace in sys.settrace docs

cool-RR:
After I released `PySnooper`, @alexmojaki added the awesome optimization of using `f_trace` instead of `sys.settrace`. He needed to explain to me how it works, because I couldn't find any documentation of `f_trace` online. So I figured I'd add it to Python's documentation.

https://bugs.python.org/issue37937

### DIFF
--- a/Doc/library/sys.rst
+++ b/Doc/library/sys.rst
@@ -1306,8 +1306,9 @@ always available.
    trace function. This is also required for activating the trace function on
    the current frame, which :func:`settrace` doesn't do. If you do this, you
    must also call :func:`settrace` with the same tracing function for it to
-   work. For more information on code and frame objects, refer to
-   :ref:`types`.
+   affect called functions.
+   
+   For more information on code and frame objects, refer to :ref:`types`.
 
    .. audit-event:: sys.settrace "" sys.settrace
 
--- a/Doc/library/sys.rst
+++ b/Doc/library/sys.rst
@@ -1300,7 +1300,14 @@ always available.
    Note that as an exception is propagated down the chain of callers, an
    ``'exception'`` event is generated at each level.
 
-   For more information on code and frame objects, refer to :ref:`types`.
+   For more fine-grained usage, it's possible to set a trace function by
+   assigning ``frame.f_trace = tracefunc`` explicitly, rather than relying on
+   it being set indirectly via the return value from an already installed
+   trace function. This is also required for activating the trace function on
+   the current frame, which :func:`settrace` doesn't do. If you do this, you
+   must also call :func:`settrace` with the same tracing function for it to
+   work. For more information on code and frame objects, refer to
+   :ref:`types`.
 
    .. audit-event:: sys.settrace "" sys.settrace
 
--- a/Doc/library/sys.rst
+++ b/Doc/library/sys.rst
@@ -1310,7 +1310,7 @@ always available.
    but it doesn't need to be the same tracing function (e.g. it could be a
    low overhead tracing function that simply returns ``None`` to disable
    itself immediately on each frame).
-   
+
    For more information on code and frame objects, refer to :ref:`types`.
 
    .. audit-event:: sys.settrace "" sys.settrace
--- a/Doc/library/sys.rst
+++ b/Doc/library/sys.rst
@@ -1304,9 +1304,12 @@ always available.
    assigning ``frame.f_trace = tracefunc`` explicitly, rather than relying on
    it being set indirectly via the return value from an already installed
    trace function. This is also required for activating the trace function on
-   the current frame, which :func:`settrace` doesn't do. If you do this, you
-   must also call :func:`settrace` with the same tracing function for it to
-   affect called functions.
+   the current frame, which :func:`settrace` doesn't do. Note that in order
+   for this to work, a global tracing function must have been installed
+   with :func:`settrace` in order to enable the runtime tracing machinery,
+   but it doesn't need to be the same tracing function (e.g. it could be a
+   low overhead tracing function that simply returns ``None`` to disable
+   itself immediately on each frame).
    
    For more information on code and frame objects, refer to :ref:`types`.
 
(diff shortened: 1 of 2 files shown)


the-knights-who-say-ni:

Doc/library/sys.rst
```diff
@@ -1300,6 +1300,9 @@ always available.
    Note that as an exception is propagated down the chain of callers, an
    ``'exception'`` event is generated at each level.
 
+   For more fine-grained usage, it's possible to set a trace function for
+   each frame individually rather than all code, by setting
```
brandtbucher: It looks like this line has a trailing space (which is breaking CI). Do you mind removing it?

```suggestion
   each frame individually rather than all code, by setting
```

cool-RR: Fixed.

cool-RR:
@brandtbucher Do you think this PR qualifies for the "skip issue" and "skip news" labels? If it's no on the latter I can add a news entry, let me know.

alexmojaki:
I think there's still some confusion. `f_trace` is not being used as an optimisation in PySnooper. It's there for correctness, not performance. Specifically it makes `with snoop():` actually trace the frame that it's in, since `sys.settrace()` can't. Using it as an optimisation instead of  just `sys.settrace()` would be very difficult because you generally want to start tracing frames from the beginning of their execution but it's hard to obtain a frame anywhere but from within it. Are you using this in a project of your own like that?

cool-RR:
Point taken, changed the text to remove the reference to optimization.

alexmojaki:
I think it's important to mention (though I'm not sure of the best way to word it) that if you implement a debugger and you provide some function to 'start tracing *here*' like a breakpoint (e.g. `pdb.set_trace` or `pysnooper.Tracer.__enter__`), then you need to set `f_trace` on the frame where that function was called. `sys.settrace()` will never do that on its own because it only starts working when you enter a new scope. That is the thing that confused and bothered me when I was implementing the functionality in both PySnooper and [another project of mine](https://github.com/alexmojaki/heartrate), and I had to look at pdb's source to figure out what was going on.

alexmojaki:
Here's a suggestion:

> Note when implementing a function which acts like a breakpoint and starts tracing at the point where it was called (e.g. `pdb.set_trace()`): `sys.settrace()` only starts tracing when it first enters a new scope. To start tracing immediately from your debugger's own function, obtain the frame where the function was called and set `frame.f_trace = tracefunc` in addition to `sys.settrace(tracefunc)`.

brandtbucher:
@cool-RR Given that this isn’t a trivial doc fix, I’d still include a basic NEWS entry... it’s probably fine to skip the issue here, though. @aeros167?

cool-RR:
Added to NEWS, created a bpo ticket. Let me know whether anything else is missing.

brandtbucher (approved):

terryjreedy:
With help from 2 non-coredev reviews, this looks ready for a coredev to review, and if the addition is correct,  likely merge.  This is too far from my areas of expertise to know.  If the addition applies to 3.8 and 3.7, add backport labels.  I suggest that one of you peruse the experts list and select a couple to request.

cool-RR:
I'd like to add the backport label for 3.7 and 3.8, but I'm not seeing a way to do that. Can anyone help?

cool-RR:
@benjaminp @pitrou I'm tagging you guys because I saw you touched `f_trace` in the past. Would any of you care to review this PR?

cool-RR:
I guess @benjaminp  and @pitrou  are busy. Maybe @ncoghlan or @serhiy-storchaka would like to review this?

cool-RR:
Paging @abalkin  , maybe you'd like to review this?

zooba:
The text itself is fine, but I'm not totally sure whether `f_trace` is intended to be writable here.

I'm pretty sure this belongs to @ncoghlan

cool-RR:
bdb writes to it, and at least one commercial debugger does too, so yes.

ncoghlan:
Belongs to only in the sense of "I think I'm the one that touched it last" ;)

The `f_trace` frame attribute is certainly intended to be read/write, it's just that the conventional way of setting it is indirectly via the return value from a globally installed trace function.

cool-RR:
Thanks for chiming in, Nick. Are you saying that because of your statement above, you disagree with this addition to the docs?

Doc/library/sys.rst
```diff
@@ -1300,7 +1300,12 @@ always available.
    Note that as an exception is propagated down the chain of callers, an
    ``'exception'`` event is generated at each level.
 
-   For more information on code and frame objects, refer to :ref:`types`.
+   For more fine-grained usage, it's possible to set a trace function for each
```
ncoghlan: I think the only thing missing here is some indication that ``frame.f_trace`` is used to store the return value from the global trace function when the frame is entered, and on each trace function invocation after that:
```suggestion
   For more fine-grained usage, it's possible to set a trace function by assigning ``frame.f_trace = tracefunc`` explicitly, rather than relying on it being set indirectly via the return value from an already installed trace function.
```

Without that, folks might be surprised by `frame.f_trace` mysteriously changing out from under them.

ncoghlan (changes requested):
This mostly looks excellent, thanks - just one point that I think could be confusing if we don't adjust the wording slightly.

Doc/library/sys.rst
ncoghlan: (Note that that suggestion is intended to replace the whole first sentence)

cool-RR:

ncoghlan (approved):
Thanks! There was one further amendment needed (related to exactly what's required to enable the C level trace hook that calls frame.f_trace), which I went ahead and made after double-checking the exact requirements in the interactive REPL.

cool-RR:
Wheee 😊

And thanks to @alexmojaki for the guidance.

PR #15442, opened 2019-08-24
bpo-19119: Remove invalid test and rename a misnamed test

rhettinger:
https://bugs.python.org/issue19119

### DIFF
--- a/Lib/test/test_heapq.py
+++ b/Lib/test/test_heapq.py
@@ -280,11 +280,6 @@ class LenOnly:
     def __len__(self):
         return 10
 
-class GetOnly:
-    "Dummy sequence class defining __getitem__ but not __len__."
-    def __getitem__(self, ndx):
-        return 10
-
 class CmpErr:
     "Dummy element that always raises an error during comparison"
     def __eq__(self, other):
--- a/Lib/test/test_heapq.py
+++ b/Lib/test/test_heapq.py
@@ -396,15 +396,7 @@ class TestErrorHandling:
         for f in (self.module.nlargest, self.module.nsmallest):
             self.assertRaises(TypeError, f, 2, LenOnly())
 
-    def test_get_only(self):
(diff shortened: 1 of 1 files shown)


aeros (commented):
@rhettinger 

Did you determine that the suggested changes in the [patch by roippi](https://bugs.python.org/file36683/fix_heapq_tests.patch) to the ``GetOnly`` class and first ``test_get_only`` method were invalid? From what I can tell, their intention was to modify the ``GetOnly`` class, modify the first ``test_get_only`` method, and rename the second one, whereas this PR is removing the ``GetOnly`` class, removing the first ``test_get_only`` method, and renaming the second one. 

This is what ``GetOnly`` would look like after the changes suggested in the patch:

```python3
class GetOnly:
    "Dummy sequence class defining __getitem__ but not __len__."
    def __getitem__(self, ndx):
        if ndx > 10:
            raise IndexError
        return 10
```

This is what the first ``test_get_only`` would look like after the changes suggested in the patch:

```python3
def test_get_only(self):
    for f in (self.module.heapify, self.module.heappop):
        self.assertRaises((TypeError, AttributeError), f, GetOnly())
    for f in (self.module.heappush, self.module.heapreplace):
        self.assertRaises((TypeError, AttributeError), f, GetOnly(), 10)
    for f in (self.module.nlargest, self.module.nsmallest):
        self.assertEqual(f(2, GetOnly()), [10,10])
```

I'm not arguing against the removal of the ``GetOnly`` class and the first ``test_get_only`` method, I'm just trying to understand the reasoning for doing so.

rhettinger:
The issue with the proposed test revision is that it doesn't really test anything we care about regarding heaps.

The original test was just a cut-and-paste error from another test source.  It doesn't actually help with testing heaps.

Thanks for taking a look at this.

aeros167:
> Thanks for taking a look at this.

No problem, thanks for the clarification. (:

PR #15446, opened 2019-08-24
[3.5] bpo-37461: Fix infinite loop in parsing of specially crafted email headers (GH-14794)

maxking:
* [bpo-37461](https://bugs.python.org/issue37461): Fix infinite loop in parsing of specially crafted email headers.

Some crafted email header would cause the get_parameter method to run in an
infinite loop causing a DoS attack surface when parsing those headers. This
patch fixes that by making sure the DQUOTE character is handled to prevent
going into an infinite loop.
(cherry picked from commit a4a994bd3e619cbaff97610a1cee8ffa87c672f5)

Co-authored-by: Abhilash Raj <maxking@users.noreply.github.com>

### DIFF
--- a/Misc/NEWS.d/next/Security/2019-07-16-08-11-00.bpo-37461.1Ahz7O.rst
+++ b/Misc/NEWS.d/next/Security/2019-07-16-08-11-00.bpo-37461.1Ahz7O.rst
@@ -1,0 +1,2 @@
+Fix an inifite loop when parsing specially crafted email headers. Patch by
+Abhilash Raj.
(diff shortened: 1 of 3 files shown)


Misc/NEWS.d/next/Security/2019-07-16-08-11-00.bpo-37461.1Ahz7O.rst
```diff
@@ -0,0 +1,2 @@
+Fix an inifite loop when parsing specially crafted email headers. Patch by
```
epicfaace:
```suggestion
Fix an infinite loop when parsing specially crafted email headers. Patch by
```

epicfaace (changes requested):

larryhastings:
Thanks for the 3.5 backport love!

PR #15448, opened 2019-08-24
bpo-37812: Expand CHECK_BINOP to make an (almost) explicit return.

gnprice:
OK, `Py_RETURN_NOTIMPLEMENTED` isn't quite an explicit `return`
statement either... but it does have the word "return" in there.
That makes it a lot easier to notice when scanning through the code,
and once noticed, it makes it unambiguous that what it's going to do
is return.

### DIFF
--- a/Misc/NEWS.d/next/Core and Builtins/2019-08-23-22-52-22.bpo-37812.nV83ZY.rst
+++ b/Misc/NEWS.d/next/Core and Builtins/2019-08-23-22-52-22.bpo-37812.nV83ZY.rst
@@ -1,0 +1,3 @@
+The ``CHECK_BINOP`` macro inside :file:`Object/longobject.c` has been
+replaced with an explicit :c:macro:`Py_RETURN_NOTIMPLEMENTED` at each call
+site, inside an ordinary ``if``.
--- a/Objects/longobject.c
+++ b/Objects/longobject.c
@@ -1601,13 +1601,6 @@ _PyLong_Size_t_Converter(PyObject *obj, void *ptr)
     return 1;
 }
 
-
-#define CHECK_BINOP(v,w)                                \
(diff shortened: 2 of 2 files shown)


Misc/NEWS.d/next/Core and Builtins/2019-08-23-22-52-22.bpo-37812.nV83ZY.rst
```diff
@@ -0,0 +1,3 @@
+The ``CHECK_BINOP`` macro inside :file:`Object/longobject.c` has been
+replaced with an explicit :c:macro:`Py_RETURN_NOTIMPLEMENTED` at each call
+site, inside an ordinary ``if``.
```
aeros, line 3: I think the Misc/NEWS entry could be a bit more succinct:

```
The ``CHECK_BINOP`` macro in :file:`Object/longobject.c` has been replaced by 
using explicit :c:macro:`Py_RETURN_NOTIMPLEMENTED`s.
```

For more details, readers can look at the changes made to ``longobject.c``.

aeros (approved):
Thanks for the PR @gnprice.

Overall, I agree with the changes, I definitely prefer the more explicit pseudo returns. As far as I can tell, this does not modify the logic (unlike the proposed PR for ``CHECK_SMALL_INT``)

/cc @rhettinger 

I just have a minor suggestion for the news entry:

animalize:
If change the check order:
```diff
-    CHECK_BINOP(a, b);
+    CHECK_BINOP(b, a);
```
Will it be a little bit faster when the check fails?
I think when it fails, usually the second operand is wrong.

rhettinger:
See comments on the tracker.  I'd like to not churn correct code for minor aesthetic reasons.

PR #15449, opened 2019-08-24
Minor changes in Doc/faq/library.

awecx:
A few improvments on _faq/library_ doc page : 

- l. 162 : it should be "function" not "functions" otherwise the "behaviours" only refer to "class", which, I believe, was not intended. Thanks @christopheNan for spotting this.
- l. 298 : IMO concurrent.futures introduced in 3.2 should not be considered "new" in 3.9 anymore.
- l. 780 : update reference to deprecated asyncore module.

### DIFF
--- a/Doc/faq/library.rst
+++ b/Doc/faq/library.rst
@@ -832,7 +832,7 @@ There are also many other specialized generators in this module, such as:
 
 Some higher-level functions operate on sequences directly, such as:
 
-* ``choice(S)`` chooses random element from a given sequence.
+* ``choice(S)`` chooses a random element from a given sequence.
 * ``shuffle(L)`` shuffles a list in-place, i.e. permutes it randomly.
 
 There's also a ``Random`` class you can instantiate to create independent
--- a/Doc/faq/library.rst
+++ b/Doc/faq/library.rst
@@ -765,20 +765,20 @@ The :mod:`select` module is commonly used to help with asynchronous I/O on
 sockets.
 
 To prevent the TCP connect from blocking, you can set the socket to non-blocking
-mode.  Then when you do the ``connect()``, you will either connect immediately
+mode.  Then when you do the :meth:`socket.connect`, you will either connect immediately
 (unlikely) or get an exception that contains the error number as ``.errno``.
 ``errno.EINPROGRESS`` indicates that the connection is in progress, but hasn't
 finished yet.  Different OSes will return different values, so you're going to
 have to check what's returned on your system.
 
-You can use the ``connect_ex()`` method to avoid creating an exception.  It will
-just return the errno value.  To poll, you can call ``connect_ex()`` again later
+You can use the :meth:`socket.connect_ex` method to avoid creating an exception.  It will
+just return the errno value.  To poll, you can call :meth:`socket.connect_ex` again later
 -- ``0`` or ``errno.EISCONN`` indicate that you're connected -- or you can pass this
 socket to ``select()`` to check if it's writable.
 
 .. note::
-   The :mod:`asyncio` module presents a framework-like approach to the problem
-   of writing non-blocking networking code.
+   The :mod:`asyncio` module provides a general purpose concurrency 
+   library, which can be used for writing non-blocking networking code.
    The third-party `Twisted <https://twistedmatrix.com/trac/>`_ library is
    a popular and feature-rich alternative.
 
@@ -832,7 +832,7 @@ There are also many other specialized generators in this module, such as:
 
 Some higher-level functions operate on sequences directly, such as:
 
-* ``choice(S)`` chooses a random element from a given sequence.
+* ``choice(S)`` chooses random element from a given sequence.
 * ``shuffle(L)`` shuffles a list in-place, i.e. permutes it randomly.
 
 There's also a ``Random`` class you can instantiate to create independent
--- a/Doc/faq/library.rst
+++ b/Doc/faq/library.rst
@@ -160,8 +160,8 @@ The "global main logic" of your program may be as simple as ::
 at the bottom of the main module of your program.
 
 Once your program is organized as a tractable collection of function and class
-behaviours you should write test functions that exercise the behaviours.  A test
-suite that automates a sequence of tests can be associated with each module.
+behaviours, you should write test functions that exercise the behaviours.  A
+test suite that automates a sequence of tests can be associated with each module.
 This sounds like a lot of work, but since Python is so terse and flexible it's
 surprisingly easy.  You can make coding much more pleasant and fun by writing
 your test functions in parallel with the "production code", since this makes it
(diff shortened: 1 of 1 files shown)


Doc/faq/library.rst
```diff
@@ -125,7 +125,7 @@ argument list.  It is called as ::
 
    handler(signum, frame)
 
-so it should be declared with two arguments::
+so it should be declared with two parameters::
```
aeros, line 128: Agreed, a function is *called* with arguments and *declared* with parameters.

Doc/faq/library.rst
```diff
@@ -295,7 +295,7 @@ queue as there are threads.
 How do I parcel out work among a bunch of worker threads?
 ---------------------------------------------------------
 
-The easiest way is to use the new :mod:`concurrent.futures` module,
+The easiest way is to use the :mod:`concurrent.futures` module,
```
aeros, line 298: Agreed, ``concurrent.futures`` is no longer new.

Doc/faq/library.rst
```diff
@@ -774,10 +774,10 @@ have to check what's returned on your system.
 You can use the ``connect_ex()`` method to avoid creating an exception.  It will
 just return the errno value.  To poll, you can call ``connect_ex()`` again later
 -- ``0`` or ``errno.EISCONN`` indicate that you're connected -- or you can pass this
-socket to select to check if it's writable.
+socket to ``select()`` to check if it's writable.
```
aeros: If we're going to reference the method itself, a link should be provided to the [corresponding documentation](https://docs.python.org/3.5/library/select.html#select.select) with Sphinx roles:
```suggestion
socket to :meth:`select.select` to check if it's writable.
```

Doc/faq/library.rst
```diff
@@ -774,10 +774,10 @@ have to check what's returned on your system.
...
 
 .. note::
-   The :mod:`asyncore` module presents a framework-like approach to the problem
+   The :mod:`asyncio` module presents a framework-like approach to the problem
```
aeros: Asyncore is a separate module from asyncio, see https://docs.python.org/3.5/library/asyncore.html

Doc/faq/library.rst
```diff
@@ -832,8 +832,8 @@ There are also many other specialized generators in this module, such as:
 
 Some higher-level functions operate on sequences directly, such as:
 
-* ``choice(S)`` chooses random element from a given sequence
-* ``shuffle(L)`` shuffles a list in-place, i.e. permutes it randomly
+* ``choice(S)`` chooses random element from a given sequence.
```
aeros:
```suggestion
* ``choice(S)`` chooses a random element from a given sequence.
```

Doc/faq/library.rst
```diff
@@ -159,7 +159,7 @@ The "global main logic" of your program may be as simple as ::
 
 at the bottom of the main module of your program.
 
-Once your program is organized as a tractable collection of functions and class
+Once your program is organized as a tractable collection of function and class
```
aeros, line 162: I think this section is okay as it is. I can see where you were going with this though, you might have assumed that the "and" was implying: "... collection of functions behaviors and class behaviors" (which would be grammatically incorrect). But, in this case the two are entirely separate. The "behaviors" part is only connected to "class". The two items are:

1) Collection of functions
2) Collection of class behaviors

aeros (changes requested):
Thanks for the PR @awecx!

I some some feedback regarding the changes made:

Doc/faq/library.rst
awecx: Of course, but since "connect_ex()" and "errno.EISCONN" were only highlighted in the previous sentence, I felt I would do the same for "select()". But I agree maybe we should provide a link to those as well.

Doc/faq/library.rst
awecx: Of course but it is now deprecated and all all its functionnality has been moved to asyncio, hence the proposed update.

Doc/faq/library.rst
awecx: :+1:

Doc/faq/library.rst
awecx: In that case, I don't understand why the doc states that "you should write test functions that exercise the behaviours". If "behaviours" only applies to "class" and not to "function", then the doc would say that functions not embedded into classes should not be tested. Am I missing something here ?

Doc/faq/library.rst
aeros: > Of course but it is now deprecated and all all its functionnality has been moved to asyncio, hence the proposed update.

My point was that the two are not exactly one-to-one. 

I agree that this section should be updated to recommend asyncio, but the surrounding text should be adjusted a bit as well. Their purposes are similar, but the existing text is more aimed at asyncore. Asyncio is a general purpose concurrency library, whereas asyncore was more of a framework for a specific purpose.

```
The :mod:`asyncio` module provides a general purpose single-threaded and
concurrent asynchronous library, which can be used for writing non-blocking
network code.
 ```

Edit: Slightly improved section update suggestion.

File: Doc/faq/library.rst (hunk 3 above)
Comment (aeros) [reply Doc/faq/library.rst]: > Of course, but since "connect_ex()" and "errno.EISCONN" were only highlighted in the previous sentence, I felt I would do the same for "select()".

There's quite a lot of the existing documentation which simply hasn't been updated to use the Sphinx roles, since they're a fairly new addition in comparison. Generally speaking, they are only intentionally not used when the link wouldn't provide readers any additional information, such as when an object is being referred to within it's own definition.

File: Doc/faq/library.rst (hunk 6 above)
Comment (aeros) [reply Doc/faq/library.rst:162]: > If "behaviours" only applies to "class" and not to "function", then the doc would say that functions not embedded into classes should not be tested.

Hmm, good point. I might've been looking at this correction too much in isolation without fully considering the surrounding context. While reading over that sentence a couple of times, I also realized it's missing a comma:

```
Once your program is organized as a tractable collection of function and class
behaviours, you should write test functions that exercise the behaviours.
```

Also an interesting tidbit (this is definitely subjective though, so I wouldn't suggest changing it for this PR), but we actually alternate between the American English and British English versions of "behaviors/behaviours" within the documentation:

```bash
$ git grep -E "behaviors" -- "*.rst" | wc -l
26
$ git grep -E "behaviours" -- "*.rst" | wc -l
8
```

Doc/faq/library.rst
awecx: I see. I have made a new commit where "connect" and "connect_ex" have the link, but not "errno.EISCONN", nor "EISINPROGRESS" as their meaning is already detailed in this doc.

Doc/faq/library.rst
awecx: OK great, I've committed your proposal.

Doc/faq/library.rst
awecx: Ah nice catch (never occured to me there were two versions of this word). I've added the comma.

Doc/faq/library.rst
aeros: > I see. I have made a new commit where "connect" and "connect_ex" have the link, but not "errno.EISCONN", nor "EISINPROGRESS" as their meaning is already detailed in this doc.

Just for clarification: Terms on the same doc page can be linked to if they're not defined in the same section (or in a very nearby paragraph), but that's fine.

aeros (approved):

matrixise:
Thank you for your contribution.

PR #15450, opened 2019-08-24
bpo-32118:  Simplify docs for sequence comparison

rhettinger:
This section had become a mess, focusing on a pathological case and not clearly specifying the essential behavior: "sequences bypass equality checks for identical objects".

NaN behavior is already covered elsewhere.

https://bugs.python.org/issue32118


terryjreedy (approved):
Nice.  It add something useful about singletons.  The condensed replacement is better than what I vaguely suggested.  I agree that the second sentence gives the essential what -- what people need to know about what Python does to write successful code.

I considered whether 'internal invariants' should be expanded.  The main one is that an object put into a sequence should be seen as in the sequence.  But anything about that really belongs in discussion of 'in' and `'__contains__'`.  I decided that the 3rd sentence is sufficient justification as is for why Python makes the stated assumption.

PR #15451, opened 2019-08-24
bpo-37936: Avoid ignoring files that we actually do track.

gnprice:
There were about 14 files that are actually in the repo but that are
covered by the rules in .gitignore.

Git itself takes no notice of what .gitignore says about files that
it's already tracking... but the discrepancy can be confusing to a
human that adds a new file unexpectedly covered by these rules, as
well as to non-Git software that looks at .gitignore but doesn't
implement this wrinkle in its semantics.  (E.g., `rg`.)

Several of these are from rules that apply more broadly than
intended: for example, `Makefile` applies to `Doc/Makefile` and
`Tools/freeze/test/Makefile`, whereas `/Makefile` means only the
`Makefile` at the repo's root.

And the `Modules/Setup` rule simply wasn't updated after 961d54c5c.

https://bugs.python.org/issue37936

Automerge-Triggered-By: @zware

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -27,12 +27,12 @@ Include/pydtrace_probes.h
 Lib/distutils/command/*.pdb
 Lib/lib2to3/*.pickle
 Lib/test/data/*
-Makefile
+!Lib/test/data/README
+/Makefile
 Makefile.pre
 Misc/python.pc
 Misc/python-embed.pc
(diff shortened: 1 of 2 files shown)


zware (approved):
LGTM after a rebase.

zware:
Thanks for the patch!

PR #15452, opened 2019-08-24
bpo-37929: avoid Squeezer-related config dialog crashes

taleinat:
These crashes were caused by keeping around a reference to the Squeezer instance and calling it's `load_font()` upon config changes, which sometimes happened even if the shell window no longer existed.

This change completely removes that update mechanism, instead having the editor window properly update its width attribute, which can then be used by Squeezer.

Note that the editor window's width is already used in a couple of other places, where it can currently be the incorrect value if the font, font size and/or window width are changed. With this PR, the entire codebase can correctly assume that `EditorWindow.width` is the updated width of the text widget in characters.

https://bugs.python.org/issue37929

### DIFF
--- a/Lib/idlelib/editor.py
+++ b/Lib/idlelib/editor.py
@@ -10,6 +10,7 @@ import traceback
 import webbrowser
 
 from tkinter import *
+from tkinter.font import Font
 from tkinter.ttk import Scrollbar
 import tkinter.simpledialog as tkSimpleDialog
 import tkinter.messagebox as tkMessageBox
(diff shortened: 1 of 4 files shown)


terryjreedy (approved):
Looks good to me. I manually tested code-context and squeezer for some possible regressions.
Please merge Sunday if you think ready, so in b4.

taleinat:
> Looks good to me. I manually tested code-context and squeezer for some possible regressions.
> Please merge Sunday if you think ready, so in b4.

Thanks for the quick review @terryjreedy! Merged.

PR #15453, opened 2019-08-24
bpo-37798: Update test_statistics.py

corona10:
Update test_statistics.py to verify that the statistics module works well for
both C and Python implementations


corona10:
@rhettinger Please take a look!

rhettinger (approved):

PR #15455, opened 2019-08-24
bpo-37908: Add an example of ArgumentParser.exit()

shihai1991:
https://bugs.python.org/issue37908

### DIFF
--- a/Doc/library/argparse.rst
+++ b/Doc/library/argparse.rst
@@ -2007,7 +2007,9 @@ Exiting methods
 
     import sys
 
+
     class ErrorCatchingArgumentParser(argparse.ArgumentParser):
+
         def exit(self, status=0, message=None)
             if status:
                 print(f'Exiting because of an error: {message}')
--- a/Doc/library/argparse.rst
+++ b/Doc/library/argparse.rst
@@ -2006,10 +2006,11 @@ Exiting methods
    this method to handle these steps differently::
 
     import sys
+
     class ErrorCatchingArgumentParser(argparse.ArgumentParser):
         def exit(self, status=0, message=None)
             if status:
-                print('Exiting because of an error:', message)
+                print(f'Exiting because of an error: {message}')
             sys.exit(status)
 
 .. method:: ArgumentParser.error(message)
--- a/Doc/library/argparse.rst
+++ b/Doc/library/argparse.rst
@@ -2005,12 +2005,12 @@ Exiting methods
    and, if given, it prints a *message* before that. The user can override
    this method to catching errors manually::
 
-   import sys
-   class ErrorCatchingArgumentParser(argparse.ArgumentParser):
-       def exit(self, status=0, message=None)
-           if status:
-               print('Catching an error!')
-           sys.exit(status)
+    import sys
+    class ErrorCatchingArgumentParser(argparse.ArgumentParser):
+        def exit(self, status=0, message=None)
+            if status:
+                print('Catching an error!')
+            sys.exit(status)
 
 .. method:: ArgumentParser.error(message)
 
--- a/Doc/library/argparse.rst
+++ b/Doc/library/argparse.rst
@@ -2007,10 +2007,13 @@ Exiting methods
 
     import sys
 
+    class AbnormalError(Exception):
+        pass
+
     class ErrorCatchingArgumentParser(argparse.ArgumentParser):
-        def exit(self, status=0, message=None)
+        def exit(self, status=0, message=None):
             if status:
-                print(f'Exiting because of an error: {message}')
+                raise AbnormalError(f'Raising an abnormal error: {message}')
             sys.exit(status)
 
 .. method:: ArgumentParser.error(message)
--- a/Doc/library/argparse.rst
+++ b/Doc/library/argparse.rst
@@ -2005,16 +2005,11 @@ Exiting methods
    and, if given, it prints a *message* before that. The user can override
    this method to handle these steps differently::
 
-    import sys
-
-    class AbnormalError(Exception):
-        pass
-
     class ErrorCatchingArgumentParser(argparse.ArgumentParser):
         def exit(self, status=0, message=None):
             if status:
-                raise AbnormalError(f'Raising an abnormal error: {message}')
-            sys.exit(status)
+                raise Exception(f'Exiting because of an error: {message}')
+            exit(status)
 
 .. method:: ArgumentParser.error(message)
 
--- a/Doc/library/argparse.rst
+++ b/Doc/library/argparse.rst
@@ -2007,9 +2007,7 @@ Exiting methods
 
     import sys
 
-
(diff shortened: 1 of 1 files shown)


Doc/library/argparse.rst
```diff
@@ -2002,7 +2002,15 @@ Exiting methods
 .. method:: ArgumentParser.exit(status=0, message=None)
 
    This method terminates the program, exiting with the specified *status*
-   and, if given, it prints a *message* before that.
+   and, if given, it prints a *message* before that. The user can override
+   this method to catching errors manually::
```
brandtbucher: They're not actually _catching_ the errors though... So maybe something like:

```suggestion
   this method to handle these steps differently::
```

Doc/library/argparse.rst
```diff
@@ -2002,7 +2002,15 @@ Exiting methods
...
+
+    import sys
+    class ErrorCatchingArgumentParser(argparse.ArgumentParser):
+        def exit(self, status=0, message=None)
+            if status:
+                print('Catching an error!')
```
brandtbucher: I would use `message` here somewhere, as an example. I would also avoid the word "catch", as above:

```suggestion
                print('Exiting because of an error:', message)
```

brandtbucher (changes requested):

Doc/library/argparse.rst
shihai1991: done

Doc/library/argparse.rst
shihai1991: Hm, let think about it. Actually, I want tell the user that we can judge the status and raise Error manually. So your are right, my desc is unclear.

shihai1991:
cc @rhettinger

Doc/library/argparse.rst
```diff
@@ -2002,7 +2002,15 @@ Exiting methods
 .. method:: ArgumentParser.exit(status=0, message=None)
 
    This method terminates the program, exiting with the specified *status*
-   and, if given, it prints a *message* before that.
+   and, if given, it prints a *message* before that. The user can override
+   this method to handle these steps differently::
+
+    import sys
+    class ErrorCatchingArgumentParser(argparse.ArgumentParser):
```
matrixise, line 2008: Could you reformat your code and use the f-string for your print()?

Thank you

matrixise (changes requested):

Doc/library/argparse.rst
shihai1991: Hi, stéphane, your mean Reformating my code by `>>>`?
If user use python's version is lower than 3.6, looks like they can not use f-string
REF: https://www.python.org/dev/peps/pep-0498/

matrixise: but your example is for Python 3.8 and 3.9 and they are f-string compliant ;-)

For the reformating, add spaces between some lines etc...

shihai1991: Hm, it is a new feature, so looks no relation with 3.6. Your request looks good to here.
Ok, i change it soon, thanks ;)

About format, i would check it soon.

matrixise: Other point, when you check the code of ArgumentParser.exit(), the condition for the print of the message is different, do you want that behaviour?

```python
def exit(self, status=0, message=None):
    if message:
        self._print_message(message, _sys.stderr)
    _sys.exit(status)
```

shihai1991: I checked the `exit` function again, v3.6 have this function too in https://github.com/python/cpython/blob/v3.6.0/Lib/argparse.py#L2373.
If some python user(python version < 3.6) would see this example it would be confused. Of course, if it is not merged in v3.6, it's fine. If we maintain 3.6, it looks merge in 3.6 would be better to our user. I am not clear about our backport policies.

matrixise: 3.6 is a stable release but security only, we only fix the security issues. 
In this case, we won't backport to 3.7 and 3.8 because they are stable. 3.9 is the only one to be impacted by your PR.

shihai1991: > Other point, when you check the code of ArgumentParser.exit(), the condition for the print of the message is different, do you want that behaviour?
> 
> ```python
> def exit(self, status=0, message=None):
>     if message:
>         self._print_message(message, _sys.stderr)
>     _sys.exit(status)
> ```

IMHO, we don't show this behavior is fine.

shihai1991: > 3.6 is a stable release but security only, we only fix the security issues.
> In this case, we won't backport to 3.7 and 3.8 because they are stable. 3.9 is the only one to be impacted by your PR.

Good to know, thanks.

Doc/library/argparse.rst
```diff
@@ -2006,10 +2006,11 @@ Exiting methods
    this method to handle these steps differently::
 
     import sys
+
```
shihai1991: two lines of space?

matrixise: just one line, 
```python
import sys

class ErrorCatchingArgumentParser(argparse.ArgumentParser):
    pass
```

shihai1991: I check some other examples in rst, single line is fine.

JulienPalard: Yet the rst indentation is wrong (single space)

Doc/library/argparse.rst
JulienPalard: Hi @shihai1991!
The only place where I see it could be usefull to overload the `exit` method is when we don't want to actually exit (unit tests, ...), but just report an error.

So I'd prefer the example to demo this instead, like:

```python
         def exit(self, status=0, message=None)
             raise SomeExeption(status, message)
```
(My example won't work as `SomeException` does not exist, please write a better example by, for example, implementing the exception).

shihai1991: @JulienPalard Yep. You are right. some user case is that we user input some wrong params and they hope they don't exit the argparse process. Hm, let me try it.

Doc/library/argparse.rst
shihai1991: @JulienPalard And i think it is confusing some users too :(.
I found some code examples in rst using `>>>` and the others don't using `>>>` like this example. it is confusing too(i don't have time to look the details.)

JulienPalard: Blocks with '>>>' are here to demostrate an interactive session (when the responses are important), like:
```pycon
>>> 1.2 - 1.0
0.19999999999999996
```

The ones with no '>>>' are to demonstrate code (implementation of classes, functions, ...) when there is typically no result, like:

```
def foo(x, y):
    return 5 * x - 3 * y
```

matrixise (approved):

Doc/library/argparse.rst
shihai1991: Oh, thanks, Julien Palard. First time to know the difference.

PR #15465, opened 2019-08-24
bpo-14112: Allow beginners to explore shallowness in greater depth ;-)

rhettinger:
https://bugs.python.org/issue14112


terryjreedy:
Good.

PR #9855, opened 2018-10-13
gh-79157: IDLE: first updated working version of a search bar

taleinat:
Based on [my old SearchBar IDLE extension](https://pypi.org/project/SearchBar/).

This is still missing tests and much required polish. This is here to make giving it a try easy, so that we may discuss whether we'd like to include this or something like it in IDLE.

https://bugs.python.org/issue34976

* Issue: gh-79157

PR #15474, opened 2019-08-24
bpo-37903: IDLE: Shell sidebar with prompts

taleinat:
This is a working implementation of a sidebar for IDLE's shell, removing the prompts from the main text widget (except for special prompts such as the one used by the debugger.) It works very well in my testing.

__Please give this a try and let me know what you think!__

Currently known to be missing:
1. handling of left mouse button clicks/drags on the sidebar

Notes on the implementation:
* I mostly made each change in a separate commit to make reviewing easier.
* This does *not* use the same base-class as the editor line-numbers sidebar, because it uses a canvas widget rather than a text widget. This is needed due to the shell soft-wrapping lines, causing lines to have varied height. Also, Squeezer inserts buttons into the shell's text window, and those lines have a different height as well.
* This includes quite a bit of work to get the "stdin" tag applied consistently to user input, which was rather broken but not relied on for anything. (Actually, this fixes the "recall" functionality in some edge cases.)
* Removing prompts from the text widget made knowing where each statement started impossible, but the sidebar needs to know this. This PR adds the "console" tag to the newline before each statement to overcome this.
* Removing prompts has other side effects which needed to be addressed. For example, successive statements without output now create a single long range of input tagged "stdin", which required some changes to the `recall()` method.

https://bugs.python.org/issue37903

### DIFF
--- a/Lib/idlelib/colorizer.py
+++ b/Lib/idlelib/colorizer.py
@@ -120,12 +120,10 @@ class ColorDelegator(Delegator):
             "BUILTIN": idleConf.GetHighlight(theme, "builtin"),
             "STRING": idleConf.GetHighlight(theme, "string"),
             "DEFINITION": idleConf.GetHighlight(theme, "definition"),
-            "SYNC": {'background':None,'foreground':None},
-            "TODO": {'background':None,'foreground':None},
+            "SYNC": {'background': None, 'foreground': None},
+            "TODO": {'background': None, 'foreground': None},
             "ERROR": idleConf.GetHighlight(theme, "error"),
-            # The following is used by ReplaceDialog:
-            "hit": idleConf.GetHighlight(theme, "hit"),
-            }
+        }
 
         if DEBUG: print('tagdefs',self.tagdefs)
 
--- a/Lib/idlelib/colorizer.py
+++ b/Lib/idlelib/colorizer.py
@@ -123,6 +123,11 @@ class ColorDelegator(Delegator):
             "SYNC": {'background': None, 'foreground': None},
             "TODO": {'background': None, 'foreground': None},
             "ERROR": idleConf.GetHighlight(theme, "error"),
+            # "hit" is used by ReplaceDialog to mark matches. It shouldn't be changed by Colorizer, but
+            # that currently isn't technically possible. This should be moved elsewhere in the future
+            # when fixing the "hit" tag's visibility, or when the replace dialog is replaced with a
+            # non-modal alternative.
+            "hit": idleConf.GetHighlight(theme, "hit"),
         }
 
         if DEBUG: print('tagdefs',self.tagdefs)
--- a/Lib/idlelib/editor.py
+++ b/Lib/idlelib/editor.py
@@ -1322,8 +1322,6 @@ class EditorWindow(object):
         # Debug prompt is multilined....
         ncharsdeleted = 0
         while 1:
-            if chars == self.prompt_last_line:  # '' unless PyShell
-                break
             chars = chars[:-1]
             ncharsdeleted = ncharsdeleted + 1
             have = len(chars.expandtabs(tabwidth))
@@ -1402,7 +1400,7 @@ class EditorWindow(object):
 
             # Strip whitespace before insert point unless it's in the prompt.
             i = 0
-            while line and line[-1] in " \t" and line != self.prompt_last_line:
+            while line and line[-1] in " \t":
                 line = line[:-1]
                 i += 1
             if i:
--- a/Lib/idlelib/editor.py
+++ b/Lib/idlelib/editor.py
@@ -782,9 +782,7 @@ class EditorWindow(object):
             self.color = self.ColorDelegator()
         # can add more colorizers here...
         if self.color:
-            self.per.removefilter(self.undo)
-            self.per.insertfilter(self.color)
-            self.per.insertfilter(self.undo)
+            self.per.insertfilterafter(filter=self.color, after=self.undo)
 
     def _rmcolorizer(self):
         if not self.color:
--- a/Lib/idlelib/editor.py
+++ b/Lib/idlelib/editor.py
@@ -807,10 +807,6 @@ class EditorWindow(object):
         """
         EditorWindow.color_config(self.text)
 
-        tag_colors = self.get_tag_colors()
-        for tag, tag_colors_config in tag_colors.items():
-            self.text.tag_configure(tag, **tag_colors_config)
-
         self.ResetColorizer()
 
         if self.code_context is not None:
@@ -818,13 +814,6 @@ class EditorWindow(object):
 
         if self.line_numbers is not None:
             self.line_numbers.update_colors()
-
-    def get_tag_colors(self):
-        theme = idleConf.CurrentTheme()
-        return {
-            # The following is used by ReplaceDialog:
-            "hit": idleConf.GetHighlight(theme, "hit"),
-        }
 
     IDENTCHARS = string.ascii_letters + string.digits + "_"
 
--- a/Lib/idlelib/editor.py
+++ b/Lib/idlelib/editor.py
@@ -68,6 +68,7 @@ class EditorWindow(object):
 
     allow_code_context = True
     allow_line_numbers = True
+    user_input_insert_tags = None
 
     def __init__(self, flist=None, filename=None, key=None, root=None):
         # Delay import: runscript imports pyshell imports EditorWindow.
@@ -1331,7 +1332,7 @@ class EditorWindow(object):
         text.undo_block_start()
         text.delete("insert-%dc" % ncharsdeleted, "insert")
         if have < want:
-            text.insert("insert", ' ' * (want - have))
+            text.insert("insert", ' ' * (want - have), self.user_input_insert_tags)
         text.undo_block_stop()
         return "break"
 
@@ -1364,7 +1365,7 @@ class EditorWindow(object):
                     effective = len(prefix.expandtabs(self.tabwidth))
                     n = self.indentwidth
                     pad = ' ' * (n - effective % n)
-                text.insert("insert", pad)
+                text.insert("insert", pad, self.user_input_insert_tags)
             text.see("insert")
             return "break"
         finally:
@@ -1395,7 +1396,7 @@ class EditorWindow(object):
             if i == n:
                 # The cursor is in or at leading indentation in a continuation
                 # line; just inject an empty line at the start.
-                text.insert("insert linestart", '\n')
+                text.insert("insert linestart", '\n', self.user_input_insert_tags)
                 return "break"
             indent = line[:i]
 
@@ -1412,7 +1413,7 @@ class EditorWindow(object):
                 text.delete("insert")
 
             # Insert new line.
-            text.insert("insert", '\n')
+            text.insert("insert", '\n', self.user_input_insert_tags)
 
             # Adjust indentation for continuations and block open/close.
             # First need to find the last statement.
@@ -1448,7 +1449,7 @@ class EditorWindow(object):
                 elif c == pyparse.C_STRING_NEXT_LINES:
                     # Inside a string which started before this line;
                     # just mimic the current indent.
-                    text.insert("insert", indent)
+                    text.insert("insert", indent, self.user_input_insert_tags)
                 elif c == pyparse.C_BRACKET:
                     # Line up with the first (if any) element of the
                     # last open bracket structure; else indent one
@@ -1462,7 +1463,7 @@ class EditorWindow(object):
                     # beyond leftmost =; else to beyond first chunk of
                     # non-whitespace on initial line.
                     if y.get_num_lines_in_stmt() > 1:
-                        text.insert("insert", indent)
+                        text.insert("insert", indent, self.user_input_insert_tags)
                     else:
                         self.reindent_to(y.compute_backslash_indent())
                 else:
@@ -1473,7 +1474,7 @@ class EditorWindow(object):
             # indentation of initial line of closest preceding
             # interesting statement.
             indent = y.get_base_indent_string()
-            text.insert("insert", indent)
+            text.insert("insert", indent, self.user_input_insert_tags)
             if y.is_block_opener():
                 self.smart_indent_event(event)
             elif indent and y.is_block_closer():
@@ -1520,7 +1521,7 @@ class EditorWindow(object):
         if text.compare("insert linestart", "!=", "insert"):
             text.delete("insert linestart", "insert")
         if column:
-            text.insert("insert", self._make_blanks(column))
+            text.insert("insert", self._make_blanks(column), self.user_input_insert_tags)
         text.undo_block_stop()
 
     # Guess indentwidth from text content.
--- a/Lib/idlelib/editor.py
+++ b/Lib/idlelib/editor.py
@@ -266,7 +266,7 @@ class EditorWindow(object):
         io.set_filename_change_hook(self.filename_change_hook)
         self.good_load = False
         self.set_indentation_params(False)
-        self.color = None # initialized below in self.ResetColorizer
+        self.color = None # initialized below in self.update_colors()
         self.code_context = None # optionally initialized later below
         self.line_numbers = None # optionally initialized later below
         if filename:
@@ -279,7 +279,7 @@ class EditorWindow(object):
                 io.set_filename(filename)
                 self.good_load = True
 
-        self.ResetColorizer()
+        self.update_colors()
         self.saved_change_hook()
         self.update_recent_files_list()
         self.load_extensions()
@@ -792,17 +792,39 @@ class EditorWindow(object):
         self.color = None
 
     def ResetColorizer(self):
-        "Update the color theme"
-        # Called from self.filename_change_hook and from configdialog.py
+        """Reset the syntax colorizer.
+
+        For example, this is called after filename changes, since the file
+        type can affect the highlighting.
+        """
         self._rmcolorizer()
         self._addcolorizer()
+
+    def update_colors(self):
+        """Update the color theme.
+
+        For example, this is called after highlighting config changes.
+        """
         EditorWindow.color_config(self.text)
+
+        tag_colors = self.get_tag_colors()
+        for tag, tag_colors_config in tag_colors.items():
+            self.text.tag_configure(tag, **tag_colors_config)
+
+        self.ResetColorizer()
 
         if self.code_context is not None:
             self.code_context.update_highlight_colors()
 
         if self.line_numbers is not None:
             self.line_numbers.update_colors()
+
+    def get_tag_colors(self):
+        theme = idleConf.CurrentTheme()
+        return {
+            # The following is used by ReplaceDialog:
+            "hit": idleConf.GetHighlight(theme, "hit"),
+        }
 
     IDENTCHARS = string.ascii_letters + string.digits + "_"
 
--- a/Lib/idlelib/editor.py
+++ b/Lib/idlelib/editor.py
@@ -60,7 +60,6 @@ class EditorWindow(object):
     from idlelib.sidebar import LineNumbers
     from idlelib.format import FormatParagraph, FormatRegion, Indents, Rstrip
     from idlelib.parenmatch import ParenMatch
-    from idlelib.squeezer import Squeezer
     from idlelib.zoomheight import ZoomHeight
 
     filesystemencoding = sys.getfilesystemencoding()  # for file names
--- a/Lib/idlelib/idle_test/test_sidebar.py
+++ b/Lib/idlelib/idle_test/test_sidebar.py
@@ -1,4 +1,4 @@
-"""Test sidebar, coverage 96%"""
+"""Test sidebar, coverage 93%"""
 from textwrap import dedent
 import sys
 
--- a/Lib/idlelib/idle_test/test_sidebar.py
+++ b/Lib/idlelib/idle_test/test_sidebar.py
@@ -1,4 +1,7 @@
 """Test sidebar, coverage 93%"""
+from textwrap import dedent
+from time import sleep
+
 import idlelib.sidebar
 from itertools import chain
 import unittest
@@ -7,7 +10,12 @@ from test.support import requires
 import tkinter as tk
 
 from idlelib.delegator import Delegator
+from idlelib.editor import fixwordbreaks
+from idlelib import macosx
 from idlelib.percolator import Percolator
+import idlelib.pyshell
+from idlelib.pyshell import fix_x11_paste, PyShell, PyShellFileList
+from idlelib.run import fix_scaling
 
 
 class Dummy_editwin:
@@ -353,7 +361,7 @@ class LineNumbersTest(unittest.TestCase):
         ln.hide_sidebar()
 
         self.highlight_cfg = test_colors
-        # Nothing breaks with inactive code context.
+        # Nothing breaks with inactive line numbers.
         ln.update_colors()
 
         # Show line numbers, previous colors change is immediately effective.
@@ -368,6 +376,148 @@ class LineNumbersTest(unittest.TestCase):
         self.highlight_cfg = orig_colors
         ln.update_colors()
         assert_colors_are_equal(orig_colors)
+
+
+class TestShellSidebar(unittest.TestCase):
+    root: tk.Tk = None
+    shell: PyShell = None
+
+    @classmethod
+    def setUpClass(cls):
+        requires('gui')
+
+        idlelib.pyshell.use_subprocess = True
+
+        cls.root = root = tk.Tk()
+
+        fix_scaling(root)
+        fixwordbreaks(root)
+        fix_x11_paste(root)
+
+        cls.flist = flist = PyShellFileList(root)
+        macosx.setupApp(root, flist)
+        root.update_idletasks()
+
+    @classmethod
+    def tearDownClass(cls):
+        cls.root.update_idletasks()
+        cls.root.destroy()
+
+    def setUp(self):
+        self.shell = self.flist.open_shell()
+        self.root.update_idletasks()
+
+    def tearDown(self):
+        self.shell.close()
+
+    def get_sidebar_lines(self):
+        canvas = self.shell.shell_sidebar.canvas
+        sleep(0.1)
+        self.root.update()
+        texts = list(canvas.find(tk.ALL))
+        texts.sort(key=lambda text: canvas.bbox(text)[1])
+        return [canvas.itemcget(text, 'text') for text in texts]
+
+    def assertSidebarLinesEndWith(self, expected_lines):
+        self.assertEqual(
+            self.get_sidebar_lines()[-len(expected_lines):],
+            expected_lines,
+        )
+
+    def getShellLineYCoords(self):
+        text = self.shell.text
+        y_coords = []
+        index = text.index("@0,0")
+        while True:
+            lineinfo = text.dlineinfo(index)
+            if lineinfo is None:
+                break
+            y_coords.append(lineinfo[1])
+            index = text.index(f"{index} +1line")
+        return y_coords
+
+    def getSidebarLineYCoords(self):
+        canvas = self.shell.shell_sidebar.canvas
+        texts = list(canvas.find(tk.ALL))
+        texts.sort(key=lambda text: canvas.bbox(text)[1])
+        return [canvas.bbox(text)[1] for text in texts]
+
+    def assertSidebarLinesSynced(self):
+        self.assertEqual(
+            self.getSidebarLineYCoords(),
+            self.getShellLineYCoords(),
+        )
+
+    def doInput(self, input):
+        root = self.root
+        shell = self.shell
+        text = shell.text
+        for line_index, line in enumerate(input.split('\n')):
+            if line_index > 0:
+                text.event_generate('<Key-Return>')
+                text.event_generate('<KeyRelease-Return>')
+                sleep(0.1)
+                root.update()
+            for char in line:
+                text.event_generate(char)
+
+    def testInitialState(self):
+        sidebar_lines = self.get_sidebar_lines()
+        self.assertEqual(
+            sidebar_lines,
+            ['   '] * (len(sidebar_lines) - 1) + ['>>>'],
+        )
+
+    def testSingleEmptyInput(self):
+        self.doInput('\n')
+        self.assertSidebarLinesEndWith(['>>>', '>>>'])
+
+    def testSingleLineCommand(self):
+        self.doInput('1\n')
+        self.assertSidebarLinesEndWith(['>>>', '   ', '>>>'])
+
+    def testMultiLineCommand(self):
+        # note: block statements are not indented because IDLE auto-indents
+        self.doInput(dedent('''\
+            if True:
+            print(1)
+
+            '''))
+        self.assertSidebarLinesEndWith([
+            '>>>',
+            '...',
+            '...',
+            '...',
+            '   ',
+            '>>>',
+        ])
+
+    def testSingleLongLineWraps(self):
+        self.doInput('1' * 200 + '\n')
+        self.assertSidebarLinesEndWith(['>>>', '   ', '>>>'])
+        self.assertSidebarLinesSynced()
+
+    def testSqueezeSingleLineCommand(self):
+        root = self.root
+        shell = self.shell
+        text = shell.text
+
+        self.doInput('1\n')
+        self.assertSidebarLinesEndWith(['>>>', '   ', '>>>'])
+
+        line = int(shell.text.index('insert -1line').split('.')[0])
+        text.mark_set('insert', f"{line}.0")
+        text.event_generate('<<squeeze-current-text>>')
+        sleep(0.1)
+        root.update()
+        self.assertSidebarLinesEndWith(['>>>', '   ', '>>>'])
+        self.assertSidebarLinesSynced()
+
+        shell.squeezer.expandingbuttons[0].expand()
+        sleep(0.1)
+        root.update()
+        self.assertSidebarLinesEndWith(['>>>', '   ', '>>>'])
+        self.assertSidebarLinesSynced()
 
 
 if __name__ == '__main__':
--- a/Lib/idlelib/idle_test/test_sidebar.py
+++ b/Lib/idlelib/idle_test/test_sidebar.py
@@ -1,4 +1,4 @@
-"""Test sidebar, coverage 93%"""
+"""Test sidebar, coverage 96%"""
 from textwrap import dedent
 import sys
 
--- a/Lib/idlelib/idle_test/test_sidebar.py
+++ b/Lib/idlelib/idle_test/test_sidebar.py
@@ -400,15 +400,32 @@ class TestShellSidebar(unittest.TestCase):
 
     @classmethod
     def tearDownClass(cls):
+        if cls.shell is not None:
+            cls.shell.executing = False
+            cls.shell.close()
+            cls.shell = None
+        cls.flist = None
         cls.root.update_idletasks()
         cls.root.destroy()
+        cls.root = None
+
+    @classmethod
+    def init_shell(cls):
+        cls.shell = cls.flist.open_shell()
+        cls.root.update()
+        cls.n_preface_lines = int(cls.shell.text.index('end-1c').split('.')[0]) - 1
+
+    @classmethod
+    def reset_shell(cls):
+        cls.shell.per.bottom.delete(f'{cls.n_preface_lines+1}.0', 'end-1c')
+        cls.shell.shell_sidebar.update_sidebar()
+        cls.root.update()
 
     def setUp(self):
-        self.shell = self.flist.open_shell()
-        self.root.update_idletasks()
-
-    def tearDown(self):
-        self.shell.close()
+        if self.shell is None:
+            self.init_shell()
+        else:
+            self.reset_shell()
 
     def get_sidebar_lines(self):
         canvas = self.shell.shell_sidebar.canvas
@@ -449,17 +466,16 @@ class TestShellSidebar(unittest.TestCase):
         )
 
     def doInput(self, input):
-        root = self.root
         shell = self.shell
         text = shell.text
         for line_index, line in enumerate(input.split('\n')):
             if line_index > 0:
                 text.event_generate('<Key-Return>')
                 text.event_generate('<KeyRelease-Return>')
-                sleep(0.1)
-                root.update()
             for char in line:
                 text.event_generate(char)
+        sleep(0.1)
+        self.root.update()
 
     def testInitialState(self):
         sidebar_lines = self.get_sidebar_lines()
@@ -467,6 +483,7 @@ class TestShellSidebar(unittest.TestCase):
             sidebar_lines,
             ['   '] * (len(sidebar_lines) - 1) + ['>>>'],
         )
+        self.assertSidebarLinesSynced()
 
     def testSingleEmptyInput(self):
         self.doInput('\n')
--- a/Lib/idlelib/idle_test/test_sidebar.py
+++ b/Lib/idlelib/idle_test/test_sidebar.py
@@ -378,7 +378,7 @@ class LineNumbersTest(unittest.TestCase):
         assert_colors_are_equal(orig_colors)
 
 
-class TestShellSidebar(unittest.TestCase):
+class ShellSidebarTest(unittest.TestCase):
     root: tk.Tk = None
     shell: PyShell = None
 
--- a/Lib/idlelib/idle_test/test_sidebar.py
+++ b/Lib/idlelib/idle_test/test_sidebar.py
@@ -162,7 +162,7 @@ class LineNumbersTest(unittest.TestCase):
         self.assert_sidebar_n_lines(3)
         self.assert_state_disabled()
 
-        # Note: deleting up to "2.end" doesn't delete the final newline.
+        # Deleting up to "2.end" doesn't delete the final newline.
         self.text.delete('2.0', '2.end')
         self.assert_text_equals('fbarfoo\n\n\n')
         self.assert_sidebar_n_lines(3)
@@ -173,7 +173,7 @@ class LineNumbersTest(unittest.TestCase):
         self.assert_sidebar_n_lines(1)
         self.assert_state_disabled()
 
-        # Note: Text widgets always keep a single '\n' character at the end.
+        # Text widgets always keep a single '\n' character at the end.
         self.text.delete('1.0', 'end')
         self.assert_text_equals('\n')
         self.assert_sidebar_n_lines(1)
@@ -242,7 +242,7 @@ class LineNumbersTest(unittest.TestCase):
         self.assert_sidebar_n_lines(4)
         self.assertEqual(get_width(), 1)
 
-        # Note: Text widgets always keep a single '\n' character at the end.
+        # Text widgets always keep a single '\n' character at the end.
         self.text.delete('1.0', 'end -1c')
         self.assert_sidebar_n_lines(1)
         self.assertEqual(get_width(), 1)
@@ -495,7 +495,7 @@ class TestShellSidebar(unittest.TestCase):
         self.assert_sidebar_lines_end_with(['>>>', '   ', '>>>'])
 
     def test_multi_line_command(self):
-        # note: block statements are not indented because IDLE auto-indents
+        # Block statements are not indented because IDLE auto-indents.
         self.do_input(dedent('''\
             if True:
             print(1)
--- a/Lib/idlelib/idle_test/test_sidebar.py
+++ b/Lib/idlelib/idle_test/test_sidebar.py
@@ -571,6 +571,64 @@ class ShellSidebarTest(unittest.TestCase):
         self.assert_sidebar_lines_end_with(['>>>', '   ', '>>>'])
         self.assert_sidebar_lines_synced()
 
+    def test_font(self):
+        sidebar = self.shell.shell_sidebar
+
+        test_font = 'TkTextFont'
+
+        def mock_idleconf_GetFont(root, configType, section):
+            return test_font
+        GetFont_patcher = unittest.mock.patch.object(
+            idlelib.sidebar.idleConf, 'GetFont', mock_idleconf_GetFont)
+        GetFont_patcher.start()
+        def cleanup():
+            GetFont_patcher.stop()
+            sidebar.update_font()
+        self.addCleanup(cleanup)
+
+        def get_sidebar_font():
+            canvas = sidebar.canvas
+            texts = list(canvas.find(tk.ALL))
+            fonts = {canvas.itemcget(text, 'font') for text in texts}
+            self.assertEqual(len(fonts), 1)
+            return next(iter(fonts))
+
+        self.assertNotEqual(get_sidebar_font(), test_font)
+        sidebar.update_font()
+        self.assertEqual(get_sidebar_font(), test_font)
+
+    def test_highlight_colors(self):
+        sidebar = self.shell.shell_sidebar
+
+        test_colors = {"background": '#abcdef', "foreground": '#123456'}
+
+        orig_idleConf_GetHighlight = idlelib.sidebar.idleConf.GetHighlight
+        def mock_idleconf_GetHighlight(theme, element):
+            if element in ['linenumber', 'console']:
+                return test_colors
+            return orig_idleConf_GetHighlight(theme, element)
+        GetHighlight_patcher = unittest.mock.patch.object(
+            idlelib.sidebar.idleConf, 'GetHighlight',
+            mock_idleconf_GetHighlight)
+        GetHighlight_patcher.start()
+        def cleanup():
+            GetHighlight_patcher.stop()
+            sidebar.update_colors()
+        self.addCleanup(cleanup)
+
+        def get_sidebar_colors():
+            canvas = sidebar.canvas
+            texts = list(canvas.find(tk.ALL))
+            fgs = {canvas.itemcget(text, 'fill') for text in texts}
+            self.assertEqual(len(fgs), 1)
+            fg = next(iter(fgs))
+            bg = canvas.cget('background')
+            return {"background": bg, "foreground": fg}
+
+        self.assertNotEqual(get_sidebar_colors(), test_colors)
+        sidebar.update_colors()
+        self.assertEqual(get_sidebar_colors(), test_colors)
+
 
 if __name__ == '__main__':
     unittest.main(verbosity=2)
--- a/Lib/idlelib/idle_test/test_sidebar.py
+++ b/Lib/idlelib/idle_test/test_sidebar.py
@@ -2,7 +2,6 @@
 from textwrap import dedent
 from time import sleep
 
-import idlelib.sidebar
 from itertools import chain
 import unittest
 import unittest.mock
@@ -16,6 +15,8 @@ from idlelib.percolator import Percolator
 import idlelib.pyshell
 from idlelib.pyshell import fix_x11_paste, PyShell, PyShellFileList
 from idlelib.run import fix_scaling
+import idlelib.sidebar
+from idlelib.sidebar import get_lineno
 
 
 class Dummy_editwin:
@@ -378,6 +379,12 @@ class LineNumbersTest(unittest.TestCase):
         assert_colors_are_equal(orig_colors)
 
 
+def test_coroutine(test_method):
+    def new_method(self):
+        return self.run_test_coroutine(test_method(self))
+    return new_method
+
+
 class ShellSidebarTest(unittest.TestCase):
     root: tk.Tk = None
     shell: PyShell = None
@@ -413,8 +420,9 @@ class ShellSidebarTest(unittest.TestCase):
     @classmethod
     def init_shell(cls):
         cls.shell = cls.flist.open_shell()
+        cls.shell.pollinterval = 10
         cls.root.update()
-        cls.n_preface_lines = int(cls.shell.text.index('end-1c').split('.')[0]) - 1
+        cls.n_preface_lines = get_lineno(cls.shell.text, 'end-1c') - 1
 
     @classmethod
     def reset_shell(cls):
@@ -430,13 +438,12 @@ class ShellSidebarTest(unittest.TestCase):
 
     def get_sidebar_lines(self):
         canvas = self.shell.shell_sidebar.canvas
-        sleep(0.1)
-        self.root.update()
         texts = list(canvas.find(tk.ALL))
         texts.sort(key=lambda text: canvas.bbox(text)[1])
         return [canvas.itemcget(text, 'text') for text in texts]
 
     def assert_sidebar_lines_end_with(self, expected_lines):
+        self.shell.shell_sidebar.update_sidebar()
         self.assertEqual(
             self.get_sidebar_lines()[-len(expected_lines):],
             expected_lines,
@@ -475,8 +482,28 @@ class ShellSidebarTest(unittest.TestCase):
                 text.event_generate('<KeyRelease-Return>')
             for char in line:
                 text.event_generate(char)
-        sleep(0.1)
-        self.root.update()
+
+    def run_test_coroutine(self, coroutine):
+        root = self.root
+        # Exceptions raised by self.assert...() need to be raised outside of
+        # the after callback in order for the test harness to capture them.
+        exception = None
+        def after_callback():
+            nonlocal exception
+            try:
+                interval_multiplier = next(coroutine) or 1
+            except StopIteration:
+                root.quit()
+            except Exception as exc:
+                exception = exc
+                root.quit()
+            else:
+                root.after(100 * interval_multiplier, after_callback)
+        root.after(0, after_callback)
+        root.mainloop()
+
+        if exception:
+            raise exception
 
     def test_initial_state(self):
         sidebar_lines = self.get_sidebar_lines()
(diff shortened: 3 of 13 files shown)


aeros (commented):
Thanks for the PR @taleinat. I'm definitely onboard with the concept of this change, a dedicated sidebar prompt for the shell would be a nice addition.

From an initial review of the GUI, it looks like the font of the characters in the sidebar aren't scaling dynamically when the font is adjusted in the shell:

Opening the shell with a font size of 22 Arial (expected):
image

Opening the shell while the font size is set to 8 Arial and then adjusting it to 22 within the shell interface (Not expected):
image

taleinat:
Thanks for giving this a whirl, @aeros167!

> From an initial review of the GUI, it looks like the font of the characters in the sidebar aren't scaling dynamically when the font is adjusted in the shell

Good catch! Fixed.

terryjreedy:
1. I do not intend to rush this into b4, but do hope to get it into 3.7.5rc1 and 3.8.0.rc1.

2. Kyle, thank you for reviewing.  On of the things I worked on todays was Tal's other recent patch, a bugfix for a regression, which I wanted and which will be in b4.

3. About history traversal: I have not looked at the mechanism in years.  I believe I was thinking that history could traverse '>>>' labels in the sidebar.  Not all stdin text is code, as  the same tag is used for `input('prompt')` responses, which is why the latter are colorized as code.  (Unsolved bug.)  I know I intend to use the labels for a future 'code input' only save. Does tk coalesce successive blocks tagged the same?  Does an untagged '\n' separate such?  (Looking back, I see you are tagging '\n' as an invisible prompt.  Clever idea.

taleinat:
> 3. Does tk coalesce successive blocks tagged the same?

Yes.

> Does an untagged '\n' separate such?

Yes, but those are currently also tagged "stdin", and changing that would be fiddly.

> (Looking back, I see you are tagging '\n' as an invisible prompt. Clever idea.

Indeed. Note that these are tagged both "console" and "stdin".

aeros167:
> Thanks for giving this a whirl, @aeros167!

> Kyle, thank you for reviewing.

No problem. Feel free to @mention me or add me to the reviewers for any IDLE-related testing. GUI design isn't my main area of experience, but as a regular user of it I'm glad to help out with testing new changes. (:

taleinat:
> > Does an untagged '\n' separate such?
> 
> Yes, but final '\n' chars of input blocks are currently also tagged "stdin", and changing that would be fiddly.

On second thought, removing the "stdin" tag from those newline might be relatively simple.

I seem to have things working rather well as they are right now, though. I can take a look at making such a change if you like, @terryjreedy, if we decide to move forward with this and everything else looks okay.

terryjreedy:
I like the look of this and it seems to run smoothly. 

 1. The most essential improvement is to use the configured space indent both for tab and smart indents.  This should partly be a matter of removing stuff that overrides the editor settings.  I am not sure if smart indent still needs to know when code is in Shell.  Cheryl and I worked on it in the last year or so to fix bugs.

terryjreedy:
2. Much less important: fix pre-existing glitch with multiline statements.  It looks worse now.
a) Backspace to remove smart indent before hitting return.
```
>>> if 1:
...         1
... 
    1
>>> 
```
I believe I would prefer removal of the final ... before the statement ending the blank line. However, it presence matches the console, though there is no option there.  Not essential.  (Note that I had to replace \t with spaces to look correct.

b) Hit return to end statement without backspacing.
```
>>> if 2:
...         2
... 
...         # End of still present smart indent, which was originally a tab.
    2
>>> 
```
A return is inserted both before and after the smart indent.  Instead, I think the smart indent should be removed, as if never inserted.  In the Windows console, at least, return after a (manual) indent does *not* terminate a statement.  I believe that this is why a newline is also inserted before the indent, so people do not have to hit return a third time to end the statement.  

This console behavior allows one to put visually blank lines in multiline statements, but I doubt people do this much.  I have never seen a complaint not being able to do this in IDLE, and I think it a bit confusing.
```
>>> if 4:
...   5
...  
...   if 5: 7
...
5
7
>>> if 5:
...   8
...
... if 9: 10
  File "<stdin>", line 4
    if 9: 10
    ^
SyntaxError: invalid syntax
```

terryjreedy:
3. Nice sometime: Removed the unused prompt before restart lines, and any statement marker added.  Then remaining prompts other than the last indicate entered and executed statements. 
```
    4
>>> 
    ================ RESTART: Shell ======================
>>> 
```
We can bikeshed over whether to leave a blank line after experience both ways.

4. I think there are some glitches with coloring (backgrounds) but I need to make a theme with stronger diffrences to be sure of what is going on.

terryjreedy:
1b (important).  Add a way to save Shell so that statements (or rather non-statements) are clearly marked.  Being able to get executable saves is another reason for moving prompts away from code.

At this point, I need to read the code so I have some idea of the difficulty of various possible changes.

aeros167:
@terryjreedy:
> This console behavior allows one to put visually blank lines in multiline statements, but I doubt people do this much. I have never seen a complaint not being able to do this in IDLE, and I think it a bit confusing.

I would agree. Whitespace and formatting concerns are generally not particularly important when using the shell.

taleinat:
@terryjreedy, perhaps I'm missing something, but all of the points you've just raised (1, 2, 3, and 1b) seem like further enhancements you'd like to make that could benefit from this change. As they are not directly related to this PR, perhaps they would be better discussed elsewhere?

Regarding the background glitches, I'll try to take a look.

terryjreedy:
2 & 3 were recorded here as I reviewed, but should be separate patches (and recorded on the main shell revamp issue).  Sidebars clicks can also be a future issue.

As to 1, we are both missing something.  You apparent think of this as a standalone patch, whereas I was only thinking of it as part of getting rid of tab indents and matching editor space indents.  That said, I really like having a proper even margin for multiline statements, even with the tabs still there.

My concern is that I expect some people will object to just this change.  Others might like it because the addition of '... ' secondary prompts makes Shell look more like the standard REPL.  Others who have never used the latter will not know.

As for 1b, Shell saves, the prompt removal is both a minus as far as finding code chunks but a plus as far as running code chunks.  I would prefer to include something that is a complete plus.

For 3.8.0rc1, scheduled Monday 9-30, this should be a moot point as both this and a couple of followups should be merged.   For 3.7.5rc1, scheduled Monday 09-16, I am not so sure.  Should this be released by itself, or should we consider not backporting to 3.7 until after.

@aeros167 Do you use IDLE regularly?  Would you want to see this in 3.7.5 even any followup patches?

In the meanwhile, I will review this as feature-complete.

Lib/idlelib/colorizer.py
```diff
@@ -120,12 +120,10 @@ def LoadTagDefs(self):
...
             "DEFINITION": idleConf.GetHighlight(theme, "definition"),
-            "SYNC": {'background':None,'foreground':None},
-            "TODO": {'background':None,'foreground':None},
+            "SYNC": {'background': None, 'foreground': None},
+            "TODO": {'background': None, 'foreground': None},
             "ERROR": idleConf.GetHighlight(theme, "error"),
-            # The following is used by ReplaceDialog:
-            "hit": idleConf.GetHighlight(theme, "hit"),
-            }
+        }
```
terryjreedy, line 131: Unlike tab indents, these changes, in particular the movement of 'hit' retrieval don't seem even indirectly related to the sidebar.  I can't see the net changes as an improvement.

Lib/idlelib/editor.py
```diff
@@ -790,18 +788,40 @@ def _rmcolorizer(self):
...
+
+        self.ResetColorizer()
+
         if self.code_context is not None:
             self.code_context.update_highlight_colors()
 
         if self.line_numbers is not None:
             self.line_numbers.update_colors()
 
+    def get_tag_colors(self):
```
terryjreedy: This confuses me. The name is misleading, as it should be 'get_hit_colors'.  It is a one-line function used only twice that for some reason singles out only one of the theme elements.  It is not a method (we need to discuss this point), and if it is be extracted, it should be a function defined in config after idleConf.

aeros167:
@terryjreedy:
> My concern is that I expect some people will object to just this change. Others might like it because the addition of '... ' secondary prompts makes Shell look more like the standard REPL. Others who have never used the latter will not know.

Perhaps to get an idea of the popularity of the change, we could open a poll on https://discuss.python.org/c/users. The detailed feedback is more valuable, but the polls tend to draw in more participants. Probably because they require a very minimal time investment. 

However, I would think that most people who have used the IDLE Shell have also used the standard REPL at some point. This may be anecdotal, but I've not interacted with anyone who has used the IDLE Shell and not the REPL. From my experience, it tends to be the other way around. 

My interactions with other Python users in person has been primarily through college, which may slightly skew the demographics. But if anything, I would think that newer users of Python would be the ones more likely to have exclusively used the IDLE shell since the REPL has been around for longer.

> @aeros167 Do you use IDLE regularly? Would you want to see this in 3.7.5 even any followup patches?

I would personally be interested in seeing this change in 3.7.5. I typically tend to alternate between using the REPL for the most simplistic usages of Python (checking the length of a line, scientific calculator, etc), IDLE for more involved tasks or scripts, and the vscode python extension for projects. I always thought it was a little bit odd that the IDLE shell didn't have the ellipsis prompts, but never thought to open an issue for it.

If I were to personally prioritize the followup patches, I would say the second one you mentioned would be the most significant to me:

> Much less important: fix pre-existing glitch with multiline statements. It looks worse now.

This is just my personal preference though. Even if they're less impactful at the end of the day, the visual bugs (even minor ones) have a tendency to stand out to me. Especially for visual bugs that would occur more commonly, such as this one.

Lib/idlelib/percolator.py
```diff
@@ -38,6 +38,21 @@ def insertfilter(self, filter):
         filter.setdelegate(self.top)
         self.top = filter
 
+    def insertfilterafter(self, filter, after):
+        assert isinstance(filter, Delegator)
```
terryjreedy, line 42: We do not routinely use assert.  Why here?  The lack of docstring and tests?  Bad design such that bad calls do not otherwise fail?  If filter is not a Delegator, would not filter.setdelegate fail?  If so, no check is needed here.

The advice I have seen is that assert, which can be optimized away, should only be used to check internal logic.  I guess the question is where the boundary of internal for IDLE.  If we do use assert, we should make sure it will be executed during testing.

Lib/idlelib/percolator.py
```diff
@@ -38,6 +38,21 @@ def insertfilter(self, filter):
...
+        assert isinstance(after, Delegator)
+        assert filter.delegate is None
+
+        f = self.top
+        f.resetcache()
+        while f is not after:
```
terryjreedy, line 48: Assert strikes me as an odd (confusing) way to end a loop.  More normal:
```suggestion
        while f is not after and f is not self.bottom:
```
followed after the loop by
        if f is self.bottom: raise ValueError("after must be in the delegator chain")

Lib/idlelib/editor.py
```diff
@@ -790,18 +788,40 @@ def _rmcolorizer(self):
...
         self._addcolorizer()
+
+    def update_colors(self):
+        """Update the color theme.
+
+        For example, this is called after highlighting config changes.
+        """
         EditorWindow.color_config(self.text)
 
+        tag_colors = self.get_tag_colors()
```
terryjreedy: If editor needs the 'hit' colors, why not add it to self.color_config?

Lib/idlelib/pyshell.py
```diff
@@ -380,6 +366,27 @@ def delete(self, index1, index2=None):
             pass
         UndoDelegator.delete(self, index1, index2)
 
+    def undo_event(self, event):
```
terryjreedy, line 371: What would be the manual test of this change?

Lib/idlelib/pyshell.py
```diff
@@ -334,34 +335,19 @@ def open_shell(self, event=None):
...
-    def __init__(self):
-        ColorDelegator.__init__(self)
-        self.LoadTagDefs()
-
     def recolorize_main(self):
         self.tag_remove("TODO", "1.0", "iomark")
         self.tag_add("SYNC", "1.0", "iomark")
         ColorDelegator.recolorize_main(self)
 
-    def LoadTagDefs(self):
```
terryjreedy, line 348: If I understand correctly, you moved the non-code colors from this ColorDelegator subclass to the Shell editor subclass because the colorizer does not use them and need not know about them.

Lib/idlelib/pyshell.py
```diff
@@ -929,6 +937,33 @@ def __init__(self, flist=None):
...
+    def get_tag_colors(self):
+        tag_colors = super().get_tag_colors()
+
+        theme = idleConf.CurrentTheme()
+        tag_colors.update({
+            "stdin": {'background': None, 'foreground': None},
+            "stdout": idleConf.GetHighlight(theme, "stdout"),
+            "stderr": idleConf.GetHighlight(theme, "stderr"),
+            "console": idleConf.GetHighlight(theme, "console"),
+        })
```
terryjreedy: It would be clearer to just add a line at the end:
`            "hit": idleConf.GetHightlight(theme, "hit"),

Lib/idlelib/pyshell.py
```diff
@@ -1163,13 +1198,30 @@ def enter_callback(self, event):
         # the current line, less a leading prompt, less leading or
         # trailing whitespace
         if self.text.compare("insert", "<", "iomark linestart"):
-            # Check if there's a relevant stdin range -- if so, use it
+            # Check if there's a relevant stdin range -- if so, use it.
```
terryjreedy, line 1211: I skipped this block for now.

Lib/idlelib/pyshell.py
```diff
@@ -1222,15 +1274,16 @@ def recall(self, s, event):
             if prefix.rstrip().endswith(':'):
                 self.newline_and_indent_event(event)
                 prefix = self.text.get("insert linestart", "insert")
-            self.text.insert("insert", lines[0].strip())
+            self.text.insert("insert", lines[0].strip(), "stdin")
```
terryjreedy: I believe the addition of 'stdin' tags has something to do with the background change I mentioned in point 3.  But I am not concerned about it now.

Lib/idlelib/pyshell.py
```diff
@@ -1273,7 +1330,14 @@ def restart_shell(self, event=None):
 
     def showprompt(self):
         self.resetoutput()
-        self.console.write(self.prompt)
+
+        prompt = self.prompt
+        if self.sys_ps1 and prompt.endswith(self.sys_ps1):
+            prompt = prompt[:-len(self.sys_ps1)]
```
terryjreedy, line 1337: This is pretty useless.  PyShell is initialized with
```
        self.sys_ps1 = sys.ps1 if hasattr(sys, 'ps1') else '>>> '
        self.prompt_last_line = self.sys_ps1.split('\n')[-1]
        self.prompt = self.sys_ps1  # Changes when debug active
```
The only way I know to either reset or delete sys.ps1 and start IDLE in the same process is to do so in python code and then `import idlelib.idle`.  Under normal circumstances, self.sys_ps1 exists, is not '', and and after the following code, equals self.prompt and self.prompt_last_line.

The new conditional will trigger once and clear the prompt to '' (or remove it from 'DEBUG ON\n'.   The conditional will never trigger again.  After it is cleared, the only possible 'prompt' within the text box is "DEBUG ON\n" .  (I don't think it should be a repeated prompt -- another issue.)  So we should remove sys_ps1.  Until we remove the debug prompt, we should just initialize self.prompt it to ''.

self.prompt_last_line is more problematical.  It should be set to '' or better deleted, but it is used (only) in EditorWindow to indicate "isinstance(self, PyShell)" and affect backspace and HOME.  (The assumption is that it is never '' in Shell, but it will be if sys.ps1 ends with '\n'.)  The uses of prompt_last_line need to be reviewed to remove uses of the string as a string and replace it with a boolean where the indicator is still needed.  (I started on this.)

Lib/idlelib/sidebar.py
```diff
@@ -1,19 +1,33 @@
...
 
+def get_lineno(text, index):
+    """Utility to get the line number of an index in a Tk text widget."""
+    return int(float(text.index(index)))
+
+
 def get_end_linenumber(text):
     """Utility to get the last line's number in a Tk text widget."""
-    return int(float(text.index('end-1c')))
+    return get_lineno(text, 'end-1c')
```
terryjreedy, line 21: The docstrings violate pep8 and only 1 function is needed.
```
def get_lineno(text, idex='end-1c'):
    "Return line number in text of idex, which default to the last line"
    return int(float(text.index(idex)))

File: Lib/idlelib/sidebar.py [hunk 13]
```diff
@@ -40,6 +54,15 @@ def get_widget_padding(widget):
     return padx, pady
 
 
+@contextlib.contextmanager
```
terryjreedy, line 57: This or something like it would seem useful in pyshell too.  Yet another spinoff.

Lib/idlelib/sidebar.py
```diff
@@ -40,6 +54,15 @@ def get_widget_padding(widget):
...
+    text.configure(state=tk.NORMAL)
+    try:
+        yield
+    finally:
+        text.configure(state=tk.DISABLED)
+
+
 class BaseSideBar:
     """
     The base class for extensions which require a sidebar.
```
terryjreedy: 'extensions is not the right word anymore.

I am stopping here for now.

terryjreedy (changes requested):

Lib/idlelib/editor.py
aeros: @terryjreedy:
>  It is not a method (we need to discuss this point), and if it is be extracted, it should be a function defined in config after idleConf.

I'm not sure that I understand what you mean here. Do you mean that it shouldn't be a method? 

From my understanding, in the current implementation @taleinat is using, ``get_tag_colors`` would definitely be considered a method, as it contains a ``self`` parameter and is being used within the ``EditorWindow`` class.

I don't mean to nitpick at all here, I was just a bit confused and wanted to make sure that I'm not misunderstanding what you meant.

taleinat: Note that PyShell extends this method.

This is intended as a way for sub-classes to add tag coloring definitions, and PyShell uses it to define colorings for the "console", "stdout", "stderr" and "stdin" tags.

In the existing codebase (before this PR), these (and "hit") were shoved into ColorDelegator, which had the unfortunate side-effect of ColorDelegator clearing these tags whenever it re-colorized a block of code. This is a bug, resulting from mixing tags used only for syntax highlighting (which can be safely cleared and set again later) with other tags.

This PR needs this bug fixed to keep these tags in place, especially the "console" and "stdin" tags which it uses to recognize prompts and user inputs.

Lib/idlelib/colorizer.py
taleinat: See my other comment, regarding `get_tag_colors()`, explaining this.

terryjreedy:
@aeros167  Thank you for the response.  If you like the missing dots back, I suspect others will too.

 I don't want opinions based on imagination.  Best would be that a person use the new version for at least 15 minutes, with a variety of statements.  Minimum would be looking at a picture of a session with at least a few statements entered.  If you feel like it, it might be helpful if you could upload a screenshot of duplicate before (installed) and after (patched repository) sessions side by side.  I may ask Ned Deily, the 3.7 release manager, for an opinion, or permission to add a patch after the release of the candidate.

Lib/idlelib/percolator.py
taleinat: > We do not routinely use assert. Why here? ...

I simply mirrored the existing use in `Percolator.insertfilter()`, to which this is very similar.

Lib/idlelib/sidebar.py
aeros: @terryjreedy:
> 'extensions is not the right word anymore.

Would the correct term here be "derived class"? That seems to be the term used in https://docs.python.org/3.9/tutorial/classes.html, but there may a better word to use that would be more specific to this context. 

If "extensions" is no longer the correct term to use, the documentation for IDLE should probably be updated accordingly https://docs.python.org/3.9/library/idle.html?highlight=extensions. Unless the documentation is referring to something different from what the code comment is referring to.

Also, minor grammar nit: "that" is more correct than "which" in this context. 

```suggestion
     The base class for derived classes that require a sidebar.
```

Lib/idlelib/editor.py
taleinat: See my other comment, regarding `get_tag_colors()`, explaining this.

terryjreedy:
@taleinat "Removing prompts from the text widget made knowing where each statement started impossible, but the sidebar needs to know this. This PR adds the "console" tag to the newline before each statement to overcome this."  A couple of question based on not having read the shell
sidebar (SSB) code yet.

Q1: Does SSB keep track of where text is (apparently) frozen, so it only rescans new stuff, and not start at the beginning each time?
Q2: If the console tag were removed from a newline and SSB were told to back up to that untagged newline, would it remove the '>>>' it previously inserted?  This will also come up if text, such as async output or warnings, needs to be inserted before an entry line.

Lib/idlelib/percolator.py
taleinat: Note that this mirrors the very similar piece of logic in `Percolator.removefilter()`.

The assert isn't ending the loop. The code is kept simpler by assuming that `after` is in the percolator; this simply validates that. My understanding is that this is done in hope that this will catch bugs early enough in the development process.

I agree, more generally, that Percolator should be made safer and better tested, or removed entirely in favor of refactoring to a less confusing design. However, given its current state, I believe that this change is in line with the "status quo".

Lib/idlelib/pyshell.py
taleinat: See my other comment, regarding `get_tag_colors()`, explaining this.

In a sentence, this is indeed proper separation of concerns, but more importantly this fixes a nasty bug.

Lib/idlelib/pyshell.py
taleinat: In the shell, write some code, delete some of it, then undo the deletion. The "stdin" tag should be present on the restored input.

terryjreedy:
About history recall: a glitch new to me of skipping 1,2 char name expressions remains after the patch.
```
>>> a=1
>>> b
  File "<pyshell#14>", line 1, in <module>
    b
NameError: name 'b' is not defined
>>> a+b
  File "<pyshell#15>", line 1, in <module>
    a+b
NameError: name 'b' is not defined
>>> bbb
  File "<pyshell#16>", line 1, in <module>
    bbb
NameError: name 'bbb' is not defined
>>> bb
  File "<pyshell#17>", line 1, in <module>
    bb
NameError: name 'bb' is not defined
>>>
```
Hit history-prev (Alt-P for me) several times and 'bb' is skipped, 'bbb' is shown, 'a+b' is shown, b is skipped, and 'a=1' is shown.

Lib/idlelib/pyshell.py
taleinat: You mean overriding the parent class's method rather than extending it? Isn't that a recipe for future bugs in later changes to `EditorWindow` or other classes in the hierarchy?

Lib/idlelib/pyshell.py
taleinat: Indeed, this is needed to support custom prompts, as currently used only by the debugger (and possibly by extensions).

I was aiming to change as little as possible. The places where `prompt_last_line` is used are rather delicate, and changing them could requiring making other related changes... @terryjreedy, are you sure you want this as part of this PR?

Lib/idlelib/sidebar.py
taleinat: I guess it's a matter of style, but IMO having `get_lineno()` return the last line number by default isn't an intuitive default behavior. It feels like favoring conciseness over clarity, which I usually try to avoid.

Lib/idlelib/sidebar.py
taleinat: Indeed!

taleinat:
Thanks for reviewing this, @terryjreedy! I've added comments clarifying the reasoning behind the changes that you commented on.

Please note that I'm not trying to be defensive, just explaining my thinking. I'm happy to make any changes you request, @terryjreedy!

taleinat:
> Q1: Does SSB keep track of where text is (apparently) frozen, so it only rescans new stuff, and not start at the beginning each time?

No. The major reason for this is soft-wrapping of lines, which can change when changing the font size, window width, etc. Also, outputs before the iomark can be squeezed/unsqueezed, so they are not totally "frozen".

Note that the sidebar only scans the currently visible lines, not the entire text widget contents.

Also note that getting the display info for lines in the text widget not currently displayed (outside the currently visible area) isn't possible with the current method (`.dlineinfo()`), and may be impossible (I haven't checked), so avoiding re-calculations for the "frozen" part may be difficult.

To summarize, the currently implementation is straightfoward, robust, and in my testing performs rather well. It might be worth checking on machines with slower CPUs, though.

taleinat:
> Q2: If the console tag were removed from a newline and SSB were told to back up to that untagged newline, would it remove the '>>>' it previously inserted? This will also come up if text, such as async output or warnings, needs to be inserted before an entry line.

IIRC such outputs are inserted before the final newline, therefore preserving the position of the "console" tag. Regardless, even if the tag were removed and then added back, the sidebar will "just work" as long as it receives a request to update itself, which the current mechanism should ensure.

Lib/idlelib/sidebar.py
aeros: @terryjreedy:
> The docstrings violate pep8

As far as I can tell, this would not be in violation of [PEP-257's section on one line docstrings](https://www.python.org/dev/peps/pep-0257/#one-line-docstrings) though. PEP-257 is main authority on docstrings. If anything, it indicates that triple quotes are preferable over single quotes:

> Triple quotes are used even though the string fits on one line. This makes it easy to later expand it.

Lib/idlelib/colorizer.py
terryjreedy: I completely agree that colorizer should not delete tags that are needed but not restored by colorizer.  But I don't think that this ever happens with 'hit' as it is only used while modal dialogs are open.  So no fix is needed for 'hit' now.

The irony of this discussion about the 'hit' (found) tag is that, on Windows at least, it is effectively unused by ReplaceDialog since hits are first tagged with the select tag so that the hit tag is ignored.  (At least on Windows, debugger uses it instead of the selection tag as Windows does not allow selections in inactive windows.  )  I think it should be used and visible in find, replace, and grep, but this is obviously another issue. 

As long as 'hit' is not used by pyshell, it has nothing to do with this issue and partially fixing it, but not really  fixing it, is a distraction.  So let's delete the 'hit' changes.  If we really fix 'hit', we might treat it separately, perhaps in searchbase, much as 'breakpoint' is.  The most we might do here is add a TODO comment here about moving it someday so that colorizer does not delete it, should that become possible.

I should have suggested two weeks ago that the minimal coloring fixes that *are*needed for Shell be split off into a separate PR to be discussed and merged first.  Too late now, but it is a somewhat separate discussion.

Lib/idlelib/editor.py
terryjreedy: A better answer, which I had forgotten, is that color_config sets text color options,  and only those, and 'hit' has nothing to do with them.  It has nothing to do with this issue.

Lib/idlelib/editor.py
terryjreedy: @aeros167 The self parameter is not used, which means it could be removed.  It could be retained as a staticmethod but I don't believe in that (another discussion for another time.

terryjreedy: If we needed to do anything about 'hit' (which we don't), I would prefer to put one combined line where the two calls are.  PyShell does not use this method for the PyShell tags.  The super call in the PyShell code depends on a broken inheritance structure that I intend to change.

Lib/idlelib/percolator.py
terryjreedy: See next response.

Lib/idlelib/percolator.py
terryjreedy: It is hard to review diffs to code that needs an overhaul.  So leave what you did until then.

Lib/idlelib/pyshell.py
terryjreedy: The 'stdin' is a null tag that currently has no effect.  So its presence is invisible.  This patch appears to change the stdin background so that it is not longer the default text background which is the normal text background.  But I do not yet see how this happens.  This is the 'background glitch' I mentioned.

Lib/idlelib/pyshell.py
terryjreedy: The parent class of PyShell should not have Editor-only specializations that have to be undone by subclasses.  Depending of this broken structure is also a recipe for future bugs ;-).  Anyway, this is a moot point if we leave 'hit' alone.

Lib/idlelib/pyshell.py
terryjreedy: Setting self.prompt directly to '' is less change that setting it to '>>>' and then correcting it to ''.  Either way, the debugger will add and delete its notice.

prompt_last_line is a separate issue.  Giving it a now incorrect value is what strikes me as dangerous.  Maybe it never matters, maybe it will in certain situations.

Lib/idlelib/sidebar.py
terryjreedy: PEP-257 says that docstrings should begin with a command verb, usually 'Return', occasionally something else.

`"Return x"` is common in the stdlib.  `"""Return x"""` takes 4 extra spaces, which might be needed, and to me is too heavy for a single line.

`get_end_linenumber(text)` is only 5 chars shorter than `get_lineno(text, 'end-1c')`.  Hardly enough to justify a new function called 3 times.  How about adding `END =  'end-1c'` and making the calls  `get_lineno(text, END)`?

Lib/idlelib/sidebar.py
terryjreedy: See `idlelib/config_extensions.def` for the meaning of 'extensions'.  Many IDLE features were once implemented as extensions but no longer.

terryjreedy: Docstrings should not start with quotes-only on the first line.  Delete both and 
```suggestion
    "A base class for sidebars using Text."
```

terryjreedy:
I agree that this is enough of a change to commit and that further behavioral changes can be left for other PRs.  With Guido's endorsement of the visible change, I want to get this merged fairly soon. I will re-read comments and responses and separate follow-up suggestions from whatever minimal internal changes I would like for this merge or to at least discuss further.

I also want to at least work on a manual test protocol.  Tal, do you happen to have or can you make a list of test examples that you used for development?  At least a few can be extracted from the discussion.  Example: enter 3 1-line statements.  Cycle through them with <<previous>> (Alt-P).  Are they separately recall and in the proper order.  (Yes.)

taleinat:
>> Q1: Does SSB keep track of where text is (apparently) frozen, so it only rescans new stuff, and not start at the beginning each time?

> ... To summarize, the currently implementation is straightfoward, robust, and in my testing performs rather well. It might be worth checking on machines with slower CPUs, though.

To follow this up, I do think it may be worthwhile to optimize for the cases of scrolling and text editing. The performance could certainly be improved for these cases via caching and only updating the required parts of the sidebar rather than re-rendering it completely. I'll try to start with scrolling, which should be affected the most by this change and where latency would be most apparent.

taleinat:
> Tal, do you happen to have or can you make a list of test examples that you used for development?

Yes, I'll try to work that up ASAP.

taleinat:
> Tal, do you happen to have or can you make a list of test examples that you used for development?

Here's a (not necessarily complete) list of things I've been manually testing in the shell while developing this:

1. Enter an empty line.
2. Enter a valid statement: single-/multi-line.
3. Enter an invalid statement: single-/multi-line.
4. Recall a previous statement: single-/multi-line.
5. Undo & redo: typing / recall; single-/multi-line. Pay special attention to input/output text coloring.
6. Enter a long line of code, triggering wrapping, including wrapping into more than two lines.
7. Add and remove text at the beginning, middle and end of a wrapped line.
8. Increase & decrease width of shell window, inducing various soft wrapping of lines.
9. Scrolling: mouse wheel / scrollbar (drag / click) / moving cursor via keyboard.
10. Squeezing/un-squeezing output blocks.
11. Restarting the shell.

cben:
I'd add bulk editing: 
select single/multiple/all lines in a multi-line input -> indent/dedent/comment/uncomment.

taleinat:
> I'd add bulk editing:
select single/multiple/all lines in a multi-line input -> indent/dedent/comment/uncomment

Thanks @cben, those are indeed good to check with regard to line wrapping.

E-Paine:
I have just tested this patch and wanted to point out a few issues I found / queries I have (TBH I haven't looked all the way through this conversation so don't know if these have already been reported - apologies if so):

- Any highlight changes do not apply immediately - Similar to what @aeros reported with the font size change, the highlight/theme changes are not applied until IDLE restart (e.g. IDLE starting in classic theme and switched to dark).
- Shell prompt is still a highlight option - I think it would be nice to still use this colour in the sidebar, though if you do not want to implement this then the option should be removed.
- Sidebar does not update for multiple-line history recall - The sidebar does not show the continuation lines when recalling a multi-line statement through history (an example is as follows).

  ```
  >>> abcd \
  ...      efgh
      SyntaxError: invalid syntax
  <alt-p>
  >>> abcd \
           efgh
  ```

  Pressing return at this point, and the sidebar updates itself to the correct state.

- Unable to select lines by clicking on the sidebar - This is less of a bug and more just a feature proposal. To be consistent with the line numbers in the editor, I think clicking/dragging on the sidebar should select the relevant shell lines. IMO, this should not just apply for lines where there is content in the sidebar, but all lines (to select stdout, for example).

Sorry, I know I have probably just created you a lot of work, but I think this patch could be a very nice addition :+1:. I will keep testing and let you know if I encounter any further issues.

taleinat:
@E-Paine, no need to apologize, your review and comments are great and very helpful!! You have my thanks :pray:

I hope to look into addressing the issues you've identified, and your suggestion for line selection similar to the line numbers, in the next few days.

E-Paine:
> I hope to look into ... your suggestion for line selection

The reason I removed this in the edit was because I took Terry's initial comment on bpo to mean that he did not want this: 
"I thing the initial implementation should not response to clicks"

Also, just adding a few more issues to the pile :wink:

- Edit > redo doesn't have continuation lines - If a redo contains a line break, a continuation prompt is not shown (until there is an update in the number of shell lines).

- Edit > replace doesn't have continuation lines - If a replace is done on the current command to add a line break, the sidebar does not update to show continuation lines (similar behaviour to above).

- Printing a line break causes a continuation line to show - If the first character of a print is a line break, a continuation line is added:

  ```
  >>> print("\n")
  ...
  
  >>> print("\nabc")
  ...
      abc
  >>>
  ```

- An input gets a continuation prompt - I haven't explored the cause, but I suspect it is related to the problem above. If the user requests input, after the return key (for the input) has been pressed a continuation line shows (this also applies for scripts where there is no other stdout before):

  ```
  >>> input("Prompt: ")
  ... Prompt: abc
      'abc'
  >>>
  ```

E-Paine:
Also, please could you merge/rebase the master into your branch so we can see if anything has changed in IDLE that may cause issues (I have done this on my local clone but would appreciate if you could do it upstream). It also means I am not reporting to you long-since fixed problems (by chance, while testing your patch, I stumbled upon [issue 35263](https://bugs.python.org/issue35263)).

E-Paine:
Wow, that was fixed quick! Got a couple more, though :-(

- Smart backspace denies deletion of prompt (type `>>> ` then try backspace)
- Sidebar doesn't update when squeezer action performed (squeezing or expanding - unless a scroll action is triggered by the change in number of lines)

taleinat:
@E-Paine, those were really great catches!

I've addressed them all except for the one about `input()` which is especially tricky.

> Also, please could you merge/rebase the master into your branch

Done. I rebased, though in hindsight I should have followed my own advice to always merge branches with open PRs. :/

I'll also look into the couple of new things you've brought up. But likely not today...

taleinat:
> * Smart backspace denies deletion of prompt (type >>> then try backspace)
> * Sidebar doesn't update when squeezer action performed (squeezing or expanding - unless a scroll action is triggered by the change in number of lines)

Both fixed!

taleinat:
> An input gets a continuation prompt

After my recent fixes, after `input()` (with or without a prompt) the sidebar now only shows '...' until you finishing supplying input. Once you hit Enter, the '...' for that line is removed.

This is far from perfect, but getting it "just right" would require more refactoring than I'd like to add to this already large PR.

taleinat:
@terryjreedy, I think at this point this is ready for another thorough review.

@E-Paine, if you could give this another whirl and see if you find any other issues, that would be very helpful!

E-Paine:
Firstly, well done: you have got responsivity of the sidebar very good on slower machines (I undervolted to 0.7GHz, was compiling in the background and the sidebar still kept up very well)

Now for the bugs!

- `SyntaxError`s cause a continuation line - This applies for both normal syntax errors and EOL errors:

  ```
  >>> print(")
  ...
      SyntaxError: EOL while scanning string literal
  >>> abcd efgh
  ... SyntaxError: invalid syntax
  >>>
  ```

- Continuation lines while command executing - While a command is executing, a continuation line shows (easily reproduced using `time.sleep`). If return is pressed during this time, the first continuation line is removed but others show instead.

- Sidebar doesn't respond to mousewheel - If the mouse pointer is over the sidebar when the mousewheel is scrolled, the action is ignored.

I will keep testing but will only let you know if I find further issues.

E-Paine:
IMO, we should also set the sidebar cursor to be the same as the main shell (xterm rather than arrow). To avoid issues with the xpadding between the sidebar and main shell, we would need to set the cursor for the text_frame. This would also set the attribute for its children (at least on my system) so we would need to revert the cursor on the scrollbar (but not bother setting the cursor attribute for the sidebar canvas). I believe this is achieved by changing lines 123 & 124 of editor from:

```
self.text_frame = text_frame = Frame(top)
self.vbar = vbar = Scrollbar(text_frame, name='vbar')
```

to:

```
self.text_frame = text_frame = Frame(top, cursor='xterm')
self.vbar = vbar = Scrollbar(text_frame, name='vbar', cursor='arrow')
```

taleinat:
> IMO, we should also set the sidebar cursor to be the same as the main shell (xterm rather than arrow).

I'm not sure I agree with this. [The I-beam pointer](https://en.wikipedia.org/wiki/Cursor_(user_interface)#I-beam_pointer) indicates that text can be edited and/or selected, which is not the case for the sidebar.

In the future we do hope to make dragging with the mouse on the sidebar select lines in the shell, and at that point I'd reconsider this.

taleinat:
I've added some tests!

Each test method creates a real IDLE shell window, simulates user interaction by generating Tk events, and then inspects the state of the shell (mostly the sidebar).

However, to do this reliably I've had to make the tests call `sleep()` and `.update()` quite often. The few tests I've added so far already take 5.5 seconds to run. I may be able to reduce the time with this approach, but probably not by orders of magnitude.

@terryjreedy, what do you think about this approach to testing this?

taleinat:
> However, to do this reliably I've had to make the tests call sleep() and .update() quite often. The few tests I've added so far already take 5.5 seconds to run. I may be able to reduce the time with this approach, but probably not by orders of magnitude.

Following up, most of those 5.5 seconds are actually the overhead from creating and destroying a complete shell window repeatedly. So the way to optimize this would be to write few, long tests, or to re-use the shell window.

taleinat:
I've managed to reduce the test run time by ~40% by reusing the shell window and avoiding a few `sleep()` calls.

E-Paine:
Hi Tal,

I *think* we need to update the code coverage stat at the top of the sidebar test file (I got 82% running the following - using the venv install method found on the [devguide](https://devguide.python.org/coverage/#install-coverage)):

```
./python -m venv ../cpython-venv
../cpython-venv/bin/pip install coverage
../cpython-venv/bin/python -m coverage run -L -m idlelib.idle_test.test_sidebar
../cpython-venv/bin/python -m coverage report | grep -E "Name|---|Lib/idlelib/sidebar.py"
```

taleinat:
>  think we need to update the code coverage stat

Yes, we'll update that at the end. I'm not done adding tests yet, but am waiting for feedback from @terryjreedy on the general approach before implementing more tests.

taleinat:
Ping, @terry? If possible, please only take a *quick* look at the new TestShellSidebar tests in `test_sidebar.py`, so I know whether you approve of the general approach. @csabella, your opinion would also be greatly appreciated!

E-Paine:
Please note that the current tests can occasionally randomly fail (I presume my computer has a slight increase in background tasks or something). Here is a sample output of a random failure when running `./python -m unittest -v idlelib.idle_test.test_sidebar` (I also note seeing `testSingleLineCommand` having failed in a similar way):

```
test_click_selection (idlelib.idle_test.test_sidebar.LineNumbersTest) ... ok
test_delete (idlelib.idle_test.test_sidebar.LineNumbersTest) ... ok
test_drag_selection_down (idlelib.idle_test.test_sidebar.LineNumbersTest) ... ok
test_drag_selection_up (idlelib.idle_test.test_sidebar.LineNumbersTest) ... ok
test_font (idlelib.idle_test.test_sidebar.LineNumbersTest) ... ok
test_highlight_colors (idlelib.idle_test.test_sidebar.LineNumbersTest) ... ok
test_init_empty (idlelib.idle_test.test_sidebar.LineNumbersTest) ... ok
test_init_not_empty (idlelib.idle_test.test_sidebar.LineNumbersTest) ... ok
test_insert (idlelib.idle_test.test_sidebar.LineNumbersTest) ... ok
test_scroll (idlelib.idle_test.test_sidebar.LineNumbersTest) ... ok
test_sidebar_text_width (idlelib.idle_test.test_sidebar.LineNumbersTest)
Test that linenumber text widget is always at the minimum ... ok
test_toggle_linenumbering (idlelib.idle_test.test_sidebar.LineNumbersTest) ... ok
testInitialState (idlelib.idle_test.test_sidebar.TestShellSidebar) ... ok
testMultiLineCommand (idlelib.idle_test.test_sidebar.TestShellSidebar) ... ok
testSingleEmptyInput (idlelib.idle_test.test_sidebar.TestShellSidebar) ... ok
testSingleLineCommand (idlelib.idle_test.test_sidebar.TestShellSidebar) ... ok
testSingleLongLineWraps (idlelib.idle_test.test_sidebar.TestShellSidebar) ... FAIL
testSqueezeSingleLineCommand (idlelib.idle_test.test_sidebar.TestShellSidebar) ... FAIL

FAIL: testSingleLongLineWraps (idlelib.idle_test.test_sidebar.TestShellSidebar)
  File "/home/elisha/Documents/Python/cp0/cpython/Lib/idlelib/idle_test/test_sidebar.py", line 514, in testSingleLongLineWraps
    self.assertSidebarLinesEndWith(['>>>', '   ', '>>>'])
  File "/home/elisha/Documents/Python/cp0/cpython/Lib/idlelib/idle_test/test_sidebar.py", line 439, in assertSidebarLinesEndWith
    self.assertEqual(
AssertionError: Lists differ: ['   ', '   ', '>>>'] != ['>>>', '   ', '>>>']

First differing element 0:
'   '
'>>>'

- ['   ', '   ', '>>>']
?   ^^^

+ ['>>>', '   ', '>>>']
?   ^^^


FAIL: testSqueezeSingleLineCommand (idlelib.idle_test.test_sidebar.TestShellSidebar)
  File "/home/elisha/Documents/Python/cp0/cpython/Lib/idlelib/idle_test/test_sidebar.py", line 523, in testSqueezeSingleLineCommand
    self.assertSidebarLinesEndWith(['>>>', '   ', '>>>'])
  File "/home/elisha/Documents/Python/cp0/cpython/Lib/idlelib/idle_test/test_sidebar.py", line 439, in assertSidebarLinesEndWith
    self.assertEqual(
AssertionError: Lists differ: ['   ', '   ', '>>>'] != ['>>>', '   ', '>>>']

First differing element 0:
'   '
'>>>'

- ['   ', '   ', '>>>']
?   ^^^

+ ['>>>', '   ', '>>>']
?   ^^^


Ran 18 tests in 2.333s

FAILED (failures=2)
```

Lib/idlelib/idle_test/test_sidebar.py
```diff
@@ -370,5 +378,164 @@ def assert_colors_are_equal(colors):
...
+    shell: PyShell = None
+
+    @classmethod
+    def setUpClass(cls):
+        requires('gui')
+
+        idlelib.pyshell.use_subprocess = True
+
+        cls.root = root = tk.Tk()
+
```
terryjreedy, line 400:
```suggestion
        root.withdraw()
```
Tests still pass without extra window.

Lib/idlelib/idle_test/test_sidebar.py
```diff
@@ -370,5 +378,164 @@ def assert_colors_are_equal(colors):
         assert_colors_are_equal(orig_colors)
 
 
+class TestShellSidebar(unittest.TestCase):
```
terryjreedy: 'IDLE Convention', Test at end:
```suggestion
class ShellSidebarTest(unittest.TestCase):
```

Lib/idlelib/idle_test/test_sidebar.py
```diff
@@ -370,5 +378,164 @@ def assert_colors_are_equal(colors):
...
+        texts.sort(key=lambda text: canvas.bbox(text)[1])
+        return [canvas.itemcget(text, 'text') for text in texts]
+
+    def assertSidebarLinesEndWith(self, expected_lines):
+        self.assertEqual(
+            self.get_sidebar_lines()[-len(expected_lines):],
+            expected_lines,
+        )
+
+    def getShellLineYCoords(self):
```
terryjreedy: Except for assert methods, please stick with PEP8 and current IDLE convention for function names.
```suggestion
    def get_shellline_ycoords(self):
```
Etcetera

Lib/idlelib/idle_test/test_sidebar.py
```diff
@@ -370,5 +378,164 @@ def assert_colors_are_equal(colors):
...
+    def testSingleEmptyInput(self):
+        self.doInput('\n')
+        self.assertSidebarLinesEndWith(['>>>', '>>>'])
+
+    def testSingleLineCommand(self):
+        self.doInput('1\n')
+        self.assertSidebarLinesEndWith(['>>>', '   ', '>>>'])
+
+    def testMultiLineCommand(self):
+        # note: block statements are not indented because IDLE auto-indents
```
terryjreedy:
```suggestion
        # Block statements are not indented here because IDLE auto-indents.
```
PEP8 comments.

Lib/idlelib/idle_test/test_sidebar.py
```diff
@@ -1,4 +1,7 @@
 """Test sidebar, coverage 93%"""
```
terryjreedy, line 1:
```suggestion
"""Test sidebar, coverage 84%"""
```
On my machine it was 91% today before patch.

terryjreedy (changes requested):
Of course I would like a faster test.  Some sleeps may be unneeded.  Some are not always enough.  Frustrating. though we don't know if failures will occur on buildbots.  

By running test_sidebar 20 times while running the Python test suite in parallel, thus using all cores, I got one failure of testSqueezeSingleLineCommand.

I would am not sure what you mean by 'general approach'.  What would be a different approach?

Comments other than test_sidebar are leftover from some time past and possibly repeats.

Lib/idlelib/colorizer.py
taleinat: Since this PR requires the removal of the shell tags from the colorizer's `tag_defs`, and the refactoring that entails, removing "hit" which obviously shouldn't be there was a simple and obvious fix. I'll revert it with a comment if you prefer.

(I will mention that this will affect my PR replacing the search/replace dialogs with a toolbar: GH-9855).

Lib/idlelib/idle_test/test_sidebar.py
taleinat: Good catch!

Lib/idlelib/idle_test/test_sidebar.py
taleinat: Bah, sorry, silly mistake!

Lib/idlelib/idle_test/test_sidebar.py
taleinat: I hope to increase coverage before this is merged by adding more tests. So let's not make this change quite yet.

Lib/idlelib/idle_test/test_sidebar.py
taleinat: Done.

Lib/idlelib/editor.py
taleinat: I've removed the `get_tag_colors()` method.

Lib/idlelib/sidebar.py
taleinat: Good point.

Lib/idlelib/sidebar.py
taleinat: This module currently consistently uses triple-quotes even for single-line doc-strings.

I'd be happy to change that in a follow-up PR, but I think it's better to avoid making more widespread changes in this (already large) PR, and I wouldn't want to break the current consistency.

taleinat:
> Of course I would like a faster test. Some sleeps may be unneeded. Some are not always enough. Frustrating. though we don't know if failures will occur on buildbots.

I can have the buildbots run the tests on this branch to check this.

> By running test_sidebar 20 times while running the Python test suite in parallel, thus using all cores, I got one failure of testSqueezeSingleLineCommand.

Yes, that's what I'm afraid of: these tests being unreliable. These need to be 100% robust.

> I would am not sure what you mean by 'general approach'. What would be a different approach?

I'm trying something else now, will update soon.

taleinat:
> > I would am not sure what you mean by 'general approach'. What would be a different approach?
> 
> I'm trying something else now, will update soon.

I've just pushed a commit which actually runs a Tk mainloop for each test method. It uses coroutines to define test cases with several steps, with a `yield` in between allowing for the mainloop to do its processing.

The latest commit also sets the PyShell's `pollinterval` attribute to 10 (milliseconds) rather than default 50. This controls how often the subprocess RPC is polled for responses, and allows for things to move faster during the test.

I'm going to try this on several of my machines to check how robust it really is.

taleinat:
@E-Paine, if you could try running the tests on your machines to check for sporadic failures, that would be helpful :)

taleinat:

E-Paine:
Just did and found it to pass in all repeats. Also, did you get a chance to look at the issues I raised in a [comment further up](https://github.com/python/cpython/pull/15474#issuecomment-694770293)?

taleinat:
> Just did and found it to pass in all repeats.

Thanks, great news!

> Also, did you get a chance to look at the issues I raised in a [comment further up](https://github.com/python/cpython/pull/15474#issuecomment-694770293)?

Yes, I've taken a quick look. The first two seem to be related but I haven't gotten to the bottom of the issue yet. Regarding the scroll wheel, I haven't implemented any event handling on the sidebar yet, though I agree that not responding to scrolling is a glaring omission. Since this appears to be nearing readiness for merging, I'll try to get to that soon.

taleinat:
> Regarding the scroll wheel, I haven't implemented any event handling on the sidebar yet,

Actually, I did have code to support scrolling there, but it doesn't work on platforms which used mouse buttons 4 and 5 to signify scrolling up/down. I've added the missing bit to support those as well.

taleinat:
Also note that there is currently much code duplication between the line numbers and shell sidebar. I've a got a refactoring of them ready, but I'd like to do that in a separate PR.

taleinat:
> ```
> ./python -m venv ../cpython-venv
> ../cpython-venv/bin/pip install coverage
> ../cpython-venv/bin/python -m coverage run -L -m idlelib.idle_test.test_sidebar
> ../cpython-venv/bin/python -m coverage report | grep -E "Name|---|Lib/idlelib/sidebar.py"
> ```

For IDLE modules, to generate coverage data you should instead run e.g. `... -m idlelib.sidebar`, which will run both the unit tests for the module and our manual "htests".

taleinat:
With the latest addition to the tests, I get 89% coverage (including htests). About half of the remaining untested lines are dead lines that will be removed as part of the aforementioned refactoring I'll do in a separate PR.

Lib/idlelib/idle_test/test_sidebar.py
terryjreedy: Fine.  We can leave this unresolved as a reminder.

Lib/idlelib/sidebar.py
terryjreedy: Leave the triple quote if you prefer and space allows.  I am more concerned about an unneeded quarter-liner.

Lib/idlelib/squeezer.py
```diff
@@ -285,7 +287,7 @@ def count_lines(self, s):
         """
         return count_lines_with_wrapping(s, self.editwin.width)
 
-    def squeeze_current_text_event(self, event):
+    def squeeze_current_text(self):
         """squeeze-current-text event handler
```
terryjreedy: Start with verb, end line with period.
```suggestion
        """Handle <<squeeze-current-text>> pseudoevent.
```
Brackets could be left off.

Lib/idlelib/colorizer.py
terryjreedy: I left a comment on your toolbar issue.  With the comment added and 'hit' left, we can resolve this for now.

taleinat:
I've removed some dead code (`ShellSidebar.hide_sidebar()`) and added tests for mouse scrolling, keyboard interrupt, history, recall, undo and redo.

This brings coverage (with the htest) up to 96%.

Lib/idlelib/squeezer.py
taleinat: Done.

Lib/idlelib/idle_test/test_sidebar.py
taleinat: I've updated this now that I'm done removing dead code and adding tests (now at 96%).

terryjreedy:
Some general comments.

1. This PR is getting overly long and unwieldy.  Some comment threads seems to have been duplicated, so that responding on one copy is not reflected on another.  Close and open a new PR if you want.  In any case, please refresh with `git merge upstream/master`.

2. On my Windows system, I can detect no difference between spacing of soft- and hard-wrapped lines.  This does not really matter though because of squeeze boxes.

3. Themes, tags, and colorizer: Leaving the cursor foreground color aside, theme elements correspond to tags configured with the element's colors.  Elements and tags can be divided into syntactical and others, which I call 'structural'.  There are also non-theme structure tags that do not affect color.   Colorizer itself only colors syntax and for its own purposes only needs to know about syntax tags and its own structure tags.  The module could use a few more docstrings and comments (another issue).

In the current implementation, Colorizer.tagdefs includes "ERROR", which it never sets, because other modules depend on a) it being defined here (this could be fixed by defining "ERROR" in editor, along with "sel"), and more importantly, b) it being removed by colorizer when one hits a key after compile fails.  Changing this would be harder and maybe not worthwhile.

Colorizer.tagdefs should not define --  and sometimes remove -- structure tags managed by other modules.  These include the PyShell tags.  Search 'hit' should be in this category, but I'd rather leave verifying that its management is complete, without colorizer, to another time.  (If there had been a separate colorizer tagdefs fix issue/pr, I would have checked.)

The PyShell structure tags include "stdin", "stdout", and "stderr".  These name are also used for corresponding rpc 'channels'.  Is "stdin" used for entered code, input() response, or both?  (An outstanding issue is that input response is visibly syntax marked.)  I am starting to review live use of the patch again.  First new bug is that input prompt and response is initially marked in the sidebar as a text continuation line with '...'.

taleinat:
>  4. I run coverage with htest lines ignored (as discussed in the test README), along with other exclusions, and conditionals checked both ways, and take the figure given.  Is 'htest ignored' what you meant by "(including htests)"?

Thanks for the reminder. I've run the coverage again exactly as specified in the README, and received 93% coverage, most of which is lines to be removed via refactoring. I've updated the doc-string.

(I made a Unix-y version of your batch script, @terryjreedy, let me know if you'd like me to put a copy of it somewhere.)

taleinat:
> 1. This PR is getting overly long and unwieldy.  Some comment threads seems to have been duplicated, so that responding on one copy is not reflected on another.  Close and open a new PR if you want.  In any case, please refresh with `git merge upstream/master`.

See new PR GH-22682. I'm closing this PR in favor of that.

E-Paine:
Could we possibly add the issue with `input` to the list?

> Getting it "just right" would require more refactoring than I'd like to add to this already large PR.

Is a "part-2" PR a suitable place for such a refactor?

taleinat:
Regarding the background color, here is a comparison of before/after this PR with the background for "normal code or text" changed to pink:
a

It seems that "stdin" hasn't been using the default background for a while (it's the same on 3.8.5). ISTM that we should simply fix the background for the "stdin" tag. @terryjreedy?

taleinat:
> Is a "part-2" PR a suitable place for such a refactor?

The "take 2" PR is just a continuation of this one. Followup PRs fixing remaining issues would certainly be welcome.

taleinat:
> Could we possibly add the issue with `input` to the list?

Yes, you're right, I missed that. Good catch!

taleinat:
> It seems that "stdin" hasn't been using the default background for a while (it's the same on 3.8.5). ISTM that we should simply fix the background for the "stdin" tag. @terryjreedy?

Actually, taking another look at this, I think the only problem in this example is the white bars until the end of the lines, which are caused by having the '\n' characters at the end of the lines have the "console" tag. Changing that to use the "normal" highlight config results in what I think is the expected behavior:

image

terryjreedy:
Documenting testing IDLE is https://bugs.python.org/issue30934.

Changing the tagging of stdout and stderr text is out of scope for this issue.   I also, at least at the moment, disagree with the change.  Consider
```
>>> 1/0
  File "<pyshell#1>", line 1, in <module>
    1/0
ZeroDivisionError: division by zero
```
The 4 traceback lines are evenly tagged from margin to margin with, in my theme, a subtle but distinctive pink background.  Without considerable testing, I think I would like a ragged background less.  Ditto for truncating the background on just the last line, if the traceback happens to not have a terminating newline.

I am continuing discussion on the new PR, starting with a copy of the 'unfinished'.

PR #15477, opened 2019-08-24
bpo-34679: Add conditional to BaseProactorEventLoop init

aeros:
Adds a conditional to BaseProactorEventLoop.\_\_init\_\_() to only call set_wakeup_fd() if the current thread is the main thread. Fixes asyncio Windows (this error only occurs on Windows) ValueError (see bpo issue comments).

https://bugs.python.org/issue34679

### DIFF
--- a/Lib/asyncio/proactor_events.py
+++ b/Lib/asyncio/proactor_events.py
@@ -12,6 +12,7 @@ import socket
 import warnings
 import signal
 import collections
+import threading
 
 from . import base_events
 from . import constants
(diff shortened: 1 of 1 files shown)


aeros167:
Since this PR is primarily a small internal change to the initialization for ``BaseProactorEventLoop``, I don't think it needs a news entry. If it's needed, I don't mind adding one though.

aeros167:
Normally I would wait a bit before sending out notifications, but I'll do so now since this issue was a release blocker for 3.8b4 (which is scheduled for release on 8/26/19).

/cc @python/asyncio

aeros167:
Since asletov submitted a similar PR (https://github.com/python/cpython/pull/15492) soon after but added tests, I close this one.

PR #15481, opened 2019-08-25
bpo-37914:Support total_hours() and total_minutes() in class timedelta

hongweipeng:
https://bugs.python.org/issue37914

### DIFF
--- a/Lib/datetime.py
+++ b/Lib/datetime.py
@@ -617,6 +617,14 @@ class timedelta:
         return ((self.days * 86400 + self.seconds) * 10**6 +
                 self.microseconds) / 10**6
 
+    def total_minutes(self):
+        """Total minutes in the duration."""
+        return self.total_seconds() / 60
+
+    def total_hours(self):
+        """Total hours in the duration."""
+        return self.total_seconds() / 3600
+
     # Read-only field accessors
     @property
     def days(self):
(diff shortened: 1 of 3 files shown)


nanjekyejoannah (commented):
Was a consesus reached to add these accessors @pganssle ? Looks like the discussion here : https://bugs.python.org/issue37914 was not in favor of adding them. Am I missing something ?

aeros167:
From Paul Ganssle:
>The reason there is no hours or minutes is that the normalized form of timedelta doesn't have those components. It would be inconsistent to have `hours` and `minutes` give the total duration of the timedelta in the chosen units while `.days` and `.seconds` return just the component of the normalized form.

>What's really being asked for here are `total_hours()` and `total_minutes()` methods, and when that has come up in the past (including recently on the python-dev mailing list), we've largely decided that the "divide by the units you want" idiom is sufficient (and in fact better in that you can choose any arbitrary units rather than just the ones that have specific names).

Since this has been discussed extensively, both on the ML and on the issue page from one of the ``datetime`` experts, I'm going to label to this PR as invalid. I'll leave the final decision for closing it to @pganssle though.

pganssle:
Thanks @aeros167 and @nanjekyejoannah for the comments and @serhiy-storchaka for closing the issue.

@hongweipeng Sorry if there was some miscommunication about this, we are indeed going to stick with the approach of dividing by `timedelta` objects to get out "total" units, but thank you for your contribution. I hope this does not discourage you from making further patches.

PR #15482, opened 2019-08-25
bpo-37664: Update ensurepip bundled wheels, again

pradyunsg:
https://bugs.python.org/issue37664


pradyunsg:
Oh, I should file against master. WHoops.

PR #15488, opened 2019-08-25
bpo-36670: fixed encoding issue with libregrtest on win systems

LorenzMende:
detailed: fixed decoding of typeperf in regrtest for win environments, added lookup for locations dependent typeperf key

### DIFF
--- a/Lib/test/libregrtest/win_utils.py
+++ b/Lib/test/libregrtest/win_utils.py
@@ -1,18 +1,19 @@
 import _winapi
+import winreg
 import msvcrt
-import os
 import subprocess
 import uuid
 from test import support
-
 
 # Max size of asynchronous reads
 BUFSIZE = 8192
 # Exponential damping factor (see below)
 LOAD_FACTOR_1 = 0.9200444146293232478931553241
+
 # Seconds per measurement
 SAMPLING_INTERVAL = 5
-COUNTER_NAME = r'\System\Processor Queue Length'
+# windows registry subkey of HKEY_LOCAL_MACHINE where the counter names of typeperf are registered
+COUNTER_REGISTRY_KEY = "SOFTWARE\\Microsoft\\Windows NT\\CurrentVersion\\Perflib\\CurrentLanguage"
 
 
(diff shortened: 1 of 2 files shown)


Lib/test/libregrtest/win_utils.py
```diff
@@ -93,7 +103,7 @@ def getloadavg(self):
             # "07/19/2018 01:32:26.605","3.000000"
             toks = line.split(',')
             # Ignore blank lines and the initial header
-            if line.strip() == '' or (COUNTER_NAME in line) or len(toks) != 2:
+            if line.strip() == '' or '\\\\' in line or len(toks) != 2:
                 continue
 
             load = float(toks[1].replace('"', ''))
```
zooba, line 109: While you're here, could you handle the case where `float` raises a ValueError (or maybe a TypeError? But I think we're guaranteed to only get ValueErrors at this point)? We can continue to the next line in that case.

LorenzMende: Hi zooba,
what about
```suggestion
            try:
                toks = line.strip().split(',')
                load = float(toks[-1].replace('"', ''))
            except:
                continue
```
this would handle the ValueError/ TypeError as well as take care of the line parsing.
Or is the bare exception handling unfavourable?

ammaraskar:
If it's not too much work, could you create a PR with your alternative solution from bpo whereby you raise early and the load average calculation gets disabled.

This really feels like a lot of extra complexity for not much gain, the primary reason system load averages are reported in regrtest is to track down flaky parallel/race condition/non deterministic tests. Those are primarily run on CI machines and build-bots which are almost always English machines anyway

Lib/test/libregrtest/win_utils.py
zooba: @LorenzMende That looks like a good option to me. You'll have to merge the change yourself though - I don't have permissions to your PR branch.

vstinner:
I merged PR #16511 which is contains a similar bug, but also multiple other regrtest bugfixes.

PR #15489, opened 2019-08-25
bpo-37805: Add tests for json.dump(..., skipkeys=True)

corona10:
https://bugs.python.org/issue37805

Automerge-Triggered-By: @methane

### DIFF
--- a/Lib/test/test_json/test_dump.py
+++ b/Lib/test/test_json/test_dump.py
@@ -11,6 +11,14 @@ class TestDump:
 
     def test_dumps(self):
         self.assertEqual(self.dumps({}), '{}')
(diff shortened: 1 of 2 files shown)


Julian (commented):
Thanks!

(Not a core dev) but would probably be nice to keep this as a separate test method (e.g. `test_dump_skipkeys` ) -- the original test here is small and easily understood in isolation whereas this new one needs a bit more knowledge to follow.

Otherwise lgtm though for whatever that counts.

corona10:
@Julian
Got it, Thanks for the review!

corona10:
@methane Can you please take a look?

epicfaace (approved):

methane (approved):

PR #15492, opened 2019-08-25
bpo-34679: Restore instantiation Windows IOCP event loop from non-main thread

asvetlov:
https://bugs.python.org/issue34679

### DIFF
--- a/Lib/asyncio/proactor_events.py
+++ b/Lib/asyncio/proactor_events.py
@@ -11,6 +11,7 @@ import os
 import socket
 import warnings
 import signal
+import threading
 import collections
 
 from . import base_events
@@ -627,7 +628,9 @@ class BaseProactorEventLoop(base_events.BaseEventLoop):
         proactor.set_loop(self)
         self._make_self_pipe()
         self_no = self._csock.fileno()
-        signal.set_wakeup_fd(self_no)
+        if threading.current_thread() is threading.main_thread():
+            # wakeup fd can be installed from main thread only
+            signal.set_wakeup_fd(self_no)
 
     def _make_socket_transport(self, sock, protocol, waiter=None,
                                extra=None, server=None):
--- a/Lib/asyncio/proactor_events.py
+++ b/Lib/asyncio/proactor_events.py
@@ -629,7 +629,7 @@ class BaseProactorEventLoop(base_events.BaseEventLoop):
         self._make_self_pipe()
         self_no = self._csock.fileno()
         if threading.current_thread() is threading.main_thread():
-            # wakeup fd can be installed from main thread only
+            # wakeup fd can only be installed to a file descriptor from the main thread
             signal.set_wakeup_fd(self_no)
 
     def _make_socket_transport(self, sock, protocol, waiter=None,
--- a/Lib/test/test_asyncio/test_windows_events.py
+++ b/Lib/test/test_asyncio/test_windows_events.py
@@ -59,6 +59,25 @@ class ProactorLoopCtrlC(test_utils.TestCase):
         thread.join()
 
 
+class ProactorMultithreading(test_utils.TestCase):
+    def test_run_from_nonmain_thread(self):
+        finished = False
+
+        async def coro():
+            await asyncio.sleep(0)
+
+        def func():
+            nonlocal finished
+            loop = asyncio.new_event_loop()
+            loop.run_until_complete(coro())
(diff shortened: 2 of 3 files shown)


Lib/test/test_asyncio/test_windows_events.py
```diff
@@ -59,6 +59,25 @@ def SIGINT_after_delay():
...
+        def func():
+            nonlocal finished
+            loop = asyncio.new_event_loop()
+            loop.run_until_complete(coro())
+            finished = True
+
+        thread = threading.Thread(target=func)
+        thread.start()
+        thread.join()
+        self.assertTrue(finished)
```
aeros, line 78: Good idea with regards to adding tests to ensure that further optimizations/changes don't accidentally recreate the bug.

Misc/NEWS.d/next/Library/2019-08-25-18-07-48.bpo-34679.HECzL7.rst
```diff
@@ -0,0 +1 @@
+Restore instantiation Windows IOCP event loop from non-main thread.
```
aeros: Minor grammar nit:
```suggestion
Restores instantiation of Windows IOCP event loops from the non-main thread.
```

I was also thinking that it might be helpful to add brief entry to the [glossary](https://docs.python.org/3.8/glossary.html) for IOCP. Readers unfamiliar with async terminology might not be aware of what the acronym stands for or what it's referring to. This could be done in a separately, in another PR.

```
Input/output completion ports. IOCP is an API used for handling
concurrent asynchronous IO operations by associating each port
object with a variable number of sockets or file descriptors.
```

If we add this in to master and backport it to 3.8 in a separate PR, I can later edit the "What's New" for 3.8 to link to the glossary term with a sphinx role: ``` :term:`IOCP` ```. Let me know what you think.

Lib/asyncio/proactor_events.py
```diff
@@ -627,7 +628,9 @@ def __init__(self, proactor):
         proactor.set_loop(self)
         self._make_self_pipe()
         self_no = self._csock.fileno()
-        signal.set_wakeup_fd(self_no)
+        if threading.current_thread() is threading.main_thread():
+            # wakeup fd can be installed from main thread only
```
aeros: Minor clarification/nit: 

```suggestion
            # wakeup fd can only be installed to a file descriptor from the main thread
```

aeros (changes requested):
Thanks for the PR @asvetlov, I think this PR is more extensive than the initial one I created (https://github.com/python/cpython/pull/15477, just included the bug fix), so I closed mine.

I have a few minor suggestions:

gvanrossum (approved):
Please change the news file per suggestion from @aeros167 (I wasn't able to commit it).

Misc/NEWS.d/next/Library/2019-08-25-18-07-48.bpo-34679.HECzL7.rst
asvetlov: I agree that adding IOCP to glossary makes sense in general. 
Sorry, don't want to do it here just for change note. The note has a high chance of dropping from release note: there is no python stable release where the regression is present, only 3.8 betas. Maybe I'm wrong, release notes editor should decide.
Anyway, adding IOCP term requires its usage in asyncio docs, not only news record. That's why the term requires a separate PR.

asvetlov:
@gvanrossum @aeros167 thanks the review.
All notes are fixed.

Misc/NEWS.d/next/Library/2019-08-25-18-07-48.bpo-34679.HECzL7.rst
aeros: >I agree that adding IOCP to glossary makes sense in general.

> Sorry, don't want to do it here just for change note. The note has a high chance of dropping from release note: there is no python stable release where the regression is present, only 3.8 betas. 

@asvetlov Sounds good, that's why I was thinking of adding the term in a separate PR. I seem to recall the term being used in the docs a couple of times. From a preliminary search of the repository using the one on GitHub, the term comes up around 6 times throughout various areas of asyncio not including news entries (including in code comments and as part of ``IocpProactor``). When I get around to opening the PR for it I'll probably include git grep results for a more thorough count of its usage.

PR #15496, opened 2019-08-26
bpo-32847: Add DirectoryNotEmptyError subclass of OSError

nanjekyejoannah:
Add `DirectoryNotEmptyError` subclass of  `OSError` .

https://bugs.python.org/issue32847


Include/pyerrors.h
```diff
@@ -122,6 +122,7 @@ PyAPI_DATA(PyObject *) PyExc_FileNotFoundError;
 PyAPI_DATA(PyObject *) PyExc_InterruptedError;
 PyAPI_DATA(PyObject *) PyExc_IsADirectoryError;
 PyAPI_DATA(PyObject *) PyExc_NotADirectoryError;
+PyAPI_DATA(PyObject *) PyExc_DirectoryNotEmptyError;
```
corona10: Is there any reason not to sort by alphabetic order?
Looks like `PyAPI_DATA(PyObject *) PyExc_DirectoryNotEmptyError` should be below the  `PyAPI_DATA(PyObject *) PyExc_ConnectionResetError`

Lib/test/test_exception_hierarchy.py
```diff
@@ -63,6 +63,7 @@ def test_select_error(self):
         +-- InterruptedError                                            EINTR
         +-- IsADirectoryError                                          EISDIR
         +-- NotADirectoryError                                        ENOTDIR
+        +-- DirectoryNotEmptyError                                  ENOTEMPTY
```
corona10: alphabetic order?

corona10 (changes requested):
IMHO, Error should be added by alphabetic order!

nanjekyejoannah:
@corona10 , I will fix that and other things on this PR. Thanks.

Doc/library/exceptions.rst
```diff
@@ -640,6 +640,12 @@ depending on the system error code.
    on something which is not a directory.
    Corresponds to :c:data:`errno` ``ENOTDIR``.
 
+.. exception:: DirectoryNotEmptyError
+
+   Raised when a directory operation (such as :func:`os.rmdir`) is requested
+   on a directory which is empty.
```
pppery:
```suggestion
   on a directory which is not empty.
```

Lib/test/test_exceptions.py
```diff
@@ -1170,6 +1170,12 @@ def inner():
             self.fail("RecursionError not raised")
         self.assertEqual(wr(), None)
 
+    def test_errno_ENOTEMPTY(self):
+        with self.assertRaises(OSError) as cm:
+            test_dir = os.path.dirname(__file__)
+            os.rmdir(test_dir)
```
vstinner: I would prefer to create a temporary directory rather than trying to remove os.path.dirname(__file__) which can be /usr/lib64/python3.7/test/ for example! You may get a permission error.

vstinner: I would prefer to use a temporary directory.

Lib/test/test_exceptions.py
```diff
@@ -1170,6 +1170,12 @@ def inner():
             self.fail("RecursionError not raised")
         self.assertEqual(wr(), None)
 
+    def test_errno_ENOTEMPTY(self):
+        with self.assertRaises(OSError) as cm:
```
vstinner: Wait, why not expecting DirectoryNotEmptyError?

nanjekyejoannah: I tried but I can't assert DirectoryNotEmptyError here. Am I missing something? I will pick it later as a subclass of `OSError`.

nanjekyejoannah:
I have done requested changes @vstinner PTAL.

Lib/test/test_exceptions.py
vstinner: If the test fails, it sounds like there is a bug in your PR. Moreover, "with self.assertRaises(OSError) as cm:" should only surround "os.rmdir(dirname)".

vstinner:
@pitrou: So, do you think that it's worth it to add DirectoryNotEmptyError for ENOTEMPTY?

@nanjekyejoannah: Is ENOTEMPTY available on Windows, Linux, FreeBSD and macOS at least?

Lib/test/test_exceptions.py
nanjekyejoannah: I will investigate.

pitrou:
> @nanjekyejoannah: Is ENOTEMPTY available on Windows, Linux, FreeBSD and macOS at least?

I would like to echo this question. If ENOTEMPTY isn't available, or isn't actually returned in any non-exotic situation, then there's no need adding an Exception subclass for it.

vstinner (changes requested):
I'm not excited by adding a new builtin symbol: IMHO this module already exists too many symbols.

I'm not convinced that it's worth it. There is exactly 0 line of code in CPython code base which expects ENOTEMPTY errno. So it looks like an artifical use case.

I recall that when PEP 3151 was implemented, the implementation replaced a lot of code using "except OSError as exc: if exc.errno == XXX: ... else: raise" with "except <new specific exception>: ..." which was neat.

But again, I don't see any usage of ENOTEMPTY in the Python stdlib.

Are you aware of 3rd party code expecting ENOTEMPTY? How many projects?

Adding a builtin symbol is a significant change, it should be well motived.

csabella:
@vstinner made a comment on the bug tracker suggesting to the close this issue.  Were there any additional opinions or should it be closed?

ned-deily:
> @vstinner made a comment on the bug tracker suggesting to the close this issue. Were there any additional opinions or should it be closed?

@warsaw opened the issue.  Barry?

PR #15499, opened 2019-08-26
bpo-37947: Adjust correctly the recursion level in symtable for named expressions

pablogsal:
https://bugs.python.org/issue37947

### DIFF
--- a/Python/symtable.c
+++ b/Python/symtable.c
@@ -1508,6 +1508,7 @@ symtable_handle_namedexpr(struct symtable *st, expr_ty e)
     }
     VISIT(st, expr, e->v.NamedExpr.value);
     VISIT(st, expr, e->v.NamedExpr.target);
+    VISIT_QUIT(st, 1);
 }
 
 static int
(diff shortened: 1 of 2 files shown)


pablogsal:
This is a fix also part of https://bugs.python.org/issue37954.

I will merge it to fix the reference leak for unblocking the upcoming beta release.

Python/symtable.c
```diff
@@ -1520,7 +1521,8 @@ symtable_visit_expr(struct symtable *st, expr_ty e)
     }
     switch (e->kind) {
     case NamedExpr_kind:
-        symtable_handle_namedexpr(st, e);
+        if(!symtable_handle_namedexpr(st, e))
+            VISIT_QUIT(st, 0);
```
ncoghlan, line 1525: This doesn't look right - there are `VISIT_QUIT(st, 0)` calls inside the handler, so won't this double-decrement the stack counter?

I agree there was definitely a bug on the success case, though.

ncoghlan: Actually, there's a VISIT_QUIT at the end of `symtable_visit_expr`, so the success case was also OK.

So perhaps the best option here will be to remove the original VISIT_QUIT calls from inside the helper function, and keep the new ones in the outer visit function?

ncoghlan: https://github.com/python/cpython/pull/15593

PR #15514, opened 2019-08-26
bpo-37954: Fix reference leak in the symtable

pablogsal:
https://bugs.python.org/issue37954


pablogsal:
I will merge it to fix the reference leak for unblocking the upcoming beta release.

PR #15521, opened 2019-08-26
bpo-37955: mock.patch incorrect reference to Mock

phsilva:
mock.patch mentions that keyword arguments are passed
to Mock when it passes to MagicMock by default.

This PR fixes that.

https://bugs.python.org/issue37955

### DIFF
--- a/Doc/library/unittest.mock.rst
+++ b/Doc/library/unittest.mock.rst
@@ -1369,7 +1369,7 @@ patch
     "as"; very useful if :func:`patch` is creating a mock object for you.
 
     :func:`patch` takes arbitrary keyword arguments. These will be passed to
-    :class:`AsyncMock` if the patched object is asynchronous, to 
+    :class:`AsyncMock` if the patched object is asynchronous, to
     :class:`MagicMock` otherwise or to *new_callable* if specified.
 
     ``patch.dict(...)``, ``patch.multiple(...)`` and ``patch.object(...)`` are
--- a/Doc/library/unittest.mock.rst
+++ b/Doc/library/unittest.mock.rst
@@ -1369,8 +1369,8 @@ patch
     "as"; very useful if :func:`patch` is creating a mock object for you.
 
     :func:`patch` takes arbitrary keyword arguments. These will be passed to
-    to construct a :class:`AsyncMock` if the patched object is an async function,
-    to :class:`MagicMock` otherwise or to *new_callable* if specified.
+    :class:`AsyncMock` if the patched object is an async function, to 
+    :class:`MagicMock` otherwise or to *new_callable* if specified.
 
     ``patch.dict(...)``, ``patch.multiple(...)`` and ``patch.object(...)`` are
     available for alternate use-cases.
--- a/Doc/library/unittest.mock.rst
+++ b/Doc/library/unittest.mock.rst
@@ -1369,7 +1369,7 @@ patch
     "as"; very useful if :func:`patch` is creating a mock object for you.
 
     :func:`patch` takes arbitrary keyword arguments. These will be passed to
-    :class:`AsyncMock` if the patched object is an async function, to 
+    :class:`AsyncMock` if the patched object is asynchronous, to 
     :class:`MagicMock` otherwise or to *new_callable* if specified.
 
     ``patch.dict(...)``, ``patch.multiple(...)`` and ``patch.object(...)`` are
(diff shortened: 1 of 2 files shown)


tirkarthi:
Changing the hyperlink to `Mock` sounds good to me. But also note like https://github.com/python/cpython/pull/13681 from 3.8 the default value could be `MagicMock` or `AsyncMock` depends on the target. cc: @mariocj89 @cjw296

phsilva:
Latest patch adjust the text to account the suggestions. Appreciate any reviews.

Doc/library/unittest.mock.rst
```diff
@@ -1369,7 +1369,8 @@ patch
     "as"; very useful if :func:`patch` is creating a mock object for you.
 
     :func:`patch` takes arbitrary keyword arguments. These will be passed to
-    the :class:`Mock` (or *new_callable*) on construction.
+    to construct a :class:`AsyncMock` if the patched object is an async function,
```
cjw296: double "to" now? 
a AsyncMock -> an AsyncMock.
I'm not sure this new wording is clearer.

Lib/unittest/mock.py
```diff
@@ -1651,7 +1651,8 @@ def patch(
     "as"; very useful if `patch` is creating a mock object for you.
 
     `patch` takes arbitrary keyword arguments. These will be passed to
-    the `Mock` (or `new_callable`) on construction.
+    to construct a `AsyncMock` if the patched object is an async function,
+    to `MagicMock` otherwise or to `new_callable` if specified.
```
cjw296: Is there any way we could not have this text duplicated between mock.py and mock.rst?

cjw296 (changes requested):

phsilva:

Lib/unittest/mock.py
phsilva: I am new to Python docs, but as far as I looked around there is a lot of duplication between .rst and .py in many modules, will need to ask around about if there is another way.

Doc/library/unittest.mock.rst
```diff
@@ -1369,7 +1369,8 @@ patch
     "as"; very useful if :func:`patch` is creating a mock object for you.
 
     :func:`patch` takes arbitrary keyword arguments. These will be passed to
-    the :class:`Mock` (or *new_callable*) on construction.
+    :class:`AsyncMock` if the patched object is an async function, to
```
cjw296: Is this true only for async functions? I thought this was true for all "async things"? @lisroach ?

Lib/unittest/mock.py
cjw296: It makes this a more substantial PR, but Sphinx autodoc should be able to pull in the .py's docstrings. Not sure if autodoc is enabled for the CPython docs, that would be some more digging on your part :-)

Doc/library/unittest.mock.rst
lisroach: Yes it would be more correct to say "if the patched object is asynchronous" to cover all async object types

phsilva: Glad to know, will adjust the wording on this. Just curious ab out what other async things there in Python? generators?!

phsilva:

Lib/unittest/mock.py
phsilva: Will take a look into it, but it looks like all modules actually repeats docs in `Doc` and in `Lib`, will ask around why.

Doc/library/unittest.mock.rst
lisroach: Yep, Async Generators: https://www.python.org/dev/peps/pep-0525/
Iterators, context managers: https://www.python.org/dev/peps/pep-049
Plus there are "coroutine objects" but those are going away, along with some older terminology. Best just to cover all the bases with a broad statement ;)

phsilva:
Hi, would be great if can get some review here... Thanks a lot! @lisroach @cjw296

Lib/unittest/mock.py
cjw296: How did the asking around go?

cjw296:
@csabella: waiting for a response to https://github.com/python/cpython/pull/15521#discussion_r366782609

cjw296 (approved):
Actually, let's leave the bigger change for another time.

csabella:
Thanks!

PR #15522, opened 2019-08-26
bpo-22347: Update mimetypes.guess_type to allow proper parsing of URLs

corona10:
https://bugs.python.org/issue22347

### DIFF
--- a/Lib/mimetypes.py
+++ b/Lib/mimetypes.py
@@ -23,10 +23,10 @@ init([files]) -- parse a list of files, default knownfiles (on Windows, the
 read_mime_types(file) -- parse one file, return a dictionary or None
 """
 
-import posixpath
 import os
 import sys
-import urllib
+import posixpath
+import urllib.parse
 try:
     import winreg as _winreg
 except ImportError:
--- a/Lib/mimetypes.py
+++ b/Lib/mimetypes.py
@@ -23,10 +23,10 @@ init([files]) -- parse a list of files, default knownfiles (on Windows, the
 read_mime_types(file) -- parse one file, return a dictionary or None
 """
 
+import posixpath
 import os
 import sys
-import posixpath
-import urllib.parse
+import urllib
 try:
     import winreg as _winreg
 except ImportError:
@@ -114,7 +114,8 @@ class MimeTypes:
         but non-standard types.
         """
         url = os.fspath(url)
-        scheme, url = urllib.parse._splittype(url)
+        p = urllib.parse.urlparse(url)
+        scheme, url = p.scheme, p.path
         if scheme == 'data':
             # syntax of data URLs:
             # dataurl   := "data:" [ mediatype ] [ ";base64" ] "," data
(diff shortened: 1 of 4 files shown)


corona10:
@zooba 
This issue is pretty old remained an issue. I'd like to close this issue.
Can you please take a look?

Misc/NEWS.d/next/Library/2019-08-27-01-03-26.bpo-22347._TRpYr.rst
```diff
@@ -0,0 +1,2 @@
+Remove urllib.parse._splittype from mimetypes.guess_type. Patch by Dong-hee
```
epicfaace: Can you include that this will allow for proper parsing of URLs with only a host name?

Lib/mimetypes.py
```diff
@@ -114,7 +114,8 @@ def guess_type(self, url, strict=True):
         but non-standard types.
         """
         url = os.fspath(url)
-        scheme, url = urllib.parse._splittype(url)
+        p = urllib.parse.urlparse(url)
```
epicfaace, line 117: You can now remove the `import urllib.parse` import at the beginning of this file.

Lib/test/test_mimetypes.py
```diff
@@ -51,6 +51,11 @@ def test_non_standard_types(self):
         eq(self.db.guess_type('foo.xul', strict=False), ('text/xul', None))
         eq(self.db.guess_extension('image/jpg', strict=False), '.jpg')
 
+    def test_url(self):
+        result = self.db.guess_type('http://host.html')
```
epicfaace, line 55: Can you add an example where it checks for `http://example.com/host.html`?

epicfaace (changes requested):

corona10:
@epicfaace
Thanks for the review. I 've updated the PR. Please take a look

epicfaace (approved):
Looks good!

corona10:
Thanks alot!

Lib/mimetypes.py
```diff
@@ -23,10 +23,10 @@
 read_mime_types(file) -- parse one file, return a dictionary or None
 """
 
+import posixpath
 import os
 import sys
-import posixpath
```
maxking: Any reason to move this to the top?

These are arranged in by length of the import size. 

Also, let's keep `urllib.parse` as the import instead of `urllib`.

Lib/mimetypes.py
maxking: Lets keep the `import urllib.parse` and not add `urilib` to the list of imports, I don't see any good reason to do that.

maxking (changes requested):
Thanks @corona10 , The changes look good to me, except for the import statements (which I don't think needs any changes).

Please make sure to not force-pus after the changes and just add another additional commit. Merging the PR will squash the commits.

It is easier to lookup what changes were made after reviews through the additional commits and it gets lost if you force-push.

corona10:
@maxking 
Thanks for the review, I will keep in mind not to use push force. :)
I've reverted the import.

corona10:

maxking (approved):
I think it looks good. 

Thanks @corona10 !

PR #15536, opened 2019-08-26
[typo] Fix disagreement between value, character, and name

nicktimko:
Used to be the hex value and description agreed, and the character was odd-one-out, but now the hex value is the odd-one after 77df9a1573.


fbidu (approved):

corona10:
I think that @ethanfurman can review this PR well.

PR #15541, opened 2019-08-27
Correct Roman-numeral example in Unicode HOWTO.

gnprice:
We had the character Ⅶ aka U+2166 ROMAN NUMERAL SEVEN, but the
code point given is 2167 and it's next to 2168.  Adjust so the
code point, character, and name all agree.  (This follows up on
77df9a157, which made the name match the character.)

We also had a mixture of `;` and `:` for a delimiter.
It doesn't really matter which, but it's good to be consistent.
The actual Unicode database files use `;`, so go with that.

### DIFF
--- a/Doc/howto/unicode.rst
+++ b/Doc/howto/unicode.rst
@@ -57,14 +57,14 @@ their corresponding code points:
    ...
    007B    '{'; LEFT CURLY BRACKET
    ...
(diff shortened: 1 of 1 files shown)


sir-sigurd:
#15536 also fixes this.

serhiy-storchaka (approved):

gnprice:
Thanks @serhiy-storchaka for the merge and the backports!

PR #15545, opened 2019-08-27
bpo-37961: tracemalloc: store the actual length of traceback

jd:
[bpo-37961](https://bugs.python.org/issue37961): tracemalloc: store the actual length of traceback

This adds a new field to the traceback_t data structure so it stores the
original length of the traceback that was recorded. This is useful to know if
the traceback has been truncated or not.

https://bugs.python.org/issue37961

### DIFF
--- a/Doc/library/tracemalloc.rst
+++ b/Doc/library/tracemalloc.rst
@@ -313,6 +313,9 @@ Functions
    frames. By default, a trace of a memory block only stores the most recent
    frame: the limit is ``1``. *nframe* must be greater or equal to ``1``.
 
+   You can still read the original number of total frames that composed the
+   traceback by looking at the :attr:`Traceback.total_nframe` attribute.
+
    Storing more than ``1`` frame is only useful to compute statistics grouped
    by ``'traceback'`` or to compute cumulative statistics: see the
    :meth:`Snapshot.compare_to` and :meth:`Snapshot.statistics` methods.
@@ -659,12 +662,24 @@ Traceback
 
    When a snapshot is taken, tracebacks of traces are limited to
    :func:`get_traceback_limit` frames. See the :func:`take_snapshot` function.
+   The original number of frames of the traceback is stored in the
+   :attr:`Traceback.total_nframe` attribute. That allows to know if a traceback
+   has been truncated by the traceback limit.
 
    The :attr:`Trace.traceback` attribute is an instance of :class:`Traceback`
    instance.
 
    .. versionchanged:: 3.7
       Frames are now sorted from the oldest to the most recent, instead of most recent to oldest.
+
+   .. attribute:: total_nframe
+
+      Total number of frames that composed the traceback before truncation.
+      This attribute can be set to ``None`` if the information is not
+      available.
+
+   .. versionchanged:: 3.9
+      The :attr:`Traceback.total_nframe` attribute was added.
 
    .. method:: format(limit=None, most_recent_first=False)
 
--- a/Lib/tracemalloc.py
+++ b/Lib/tracemalloc.py
@@ -182,15 +182,20 @@ class Traceback(Sequence):
     Sequence of Frame instances sorted from the oldest frame
     to the most recent frame.
     """
-    __slots__ = ("_frames",)
+    __slots__ = ("_frames", '_total_nframe')
 
-    def __init__(self, frames):
+    def __init__(self, frames, total_nframe=None):
         Sequence.__init__(self)
         # frames is a tuple of frame tuples: see Frame constructor for the
         # format of a frame tuple; it is reversed, because _tracemalloc
         # returns frames sorted from most recent to oldest, but the
         # Python API expects oldest to most recent
         self._frames = tuple(reversed(frames))
+        self._total_nframe = total_nframe
+
+    @property
+    def total_nframe(self):
+        return self._total_nframe
 
     def __len__(self):
         return len(self._frames)
@@ -221,7 +226,12 @@ class Traceback(Sequence):
         return str(self[0])
 
     def __repr__(self):
-        return "<Traceback %r>" % (tuple(self),)
+        s = "<Traceback %r" % tuple(self)
+        if self._total_nframe is None:
+            s += ">"
+        else:
+            s += f" total_nframe={self.total_nframe}>"
+        return s
 
     def format(self, limit=None, most_recent_first=False):
         lines = []
@@ -280,7 +290,7 @@ class Trace:
 
     @property
     def traceback(self):
-        return Traceback(self._trace[2])
+        return Traceback(*self._trace[2:])
 
     def __eq__(self, other):
         if not isinstance(other, Trace):
@@ -378,7 +388,7 @@ class Filter(BaseFilter):
             return self._match_frame(filename, lineno)
 
     def _match(self, trace):
-        domain, size, traceback = trace
+        domain, size, traceback, total_nframe = trace
         res = self._match_traceback(traceback)
         if self.domain is not None:
             if self.inclusive:
@@ -398,7 +408,7 @@ class DomainFilter(BaseFilter):
         return self._domain
 
     def _match(self, trace):
-        domain, size, traceback = trace
+        domain, size, traceback, total_nframe = trace
         return (domain == self.domain) ^ (not self.inclusive)
 
 
@@ -475,7 +485,7 @@ class Snapshot:
         tracebacks = {}
         if not cumulative:
             for trace in self.traces._traces:
-                domain, size, trace_traceback = trace
+                domain, size, trace_traceback, total_nframe = trace
                 try:
                     traceback = tracebacks[trace_traceback]
                 except KeyError:
@@ -496,7 +506,7 @@ class Snapshot:
         else:
             # cumulative statistics
             for trace in self.traces._traces:
-                domain, size, trace_traceback = trace
+                domain, size, trace_traceback, total_nframe = trace
                 for frame in trace_traceback:
                     try:
                         traceback = tracebacks[frame]
--- a/Modules/_tracemalloc.c
+++ b/Modules/_tracemalloc.c
@@ -78,7 +78,10 @@ __attribute__((packed))
 
 typedef struct {
     Py_uhash_t hash;
+    /* Number of frames stored */
     int nframe;
+    /* Total number of frames the traceback had */
+    uint16_t total_nframe;
     frame_t frames[1];
 } traceback_t;
 
@@ -308,6 +311,9 @@ hashtable_compare_traceback(_Py_hashtable_t *ht, const void *pkey,
     if (traceback1->nframe != traceback2->nframe)
         return 0;
 
+    if (traceback1->total_nframe != traceback2->total_nframe)
+        return 0;
+
     for (i=0; i < traceback1->nframe; i++) {
         frame1 = &traceback1->frames[i];
         frame2 = &traceback2->frames[i];
@@ -416,6 +422,7 @@ traceback_hash(traceback_t *traceback)
         /* the cast might truncate len; that doesn't change hash stability */
         mult += (Py_uhash_t)(82520UL + len + len);
     }
+    x ^= traceback->total_nframe;
     x += 97531UL;
     return x;
 }
@@ -436,11 +443,13 @@ traceback_get_frames(traceback_t *traceback)
     }
 
     for (pyframe = tstate->frame; pyframe != NULL; pyframe = pyframe->f_back) {
-        tracemalloc_get_frame(pyframe, &traceback->frames[traceback->nframe]);
-        assert(traceback->frames[traceback->nframe].filename != NULL);
-        traceback->nframe++;
-        if (traceback->nframe == _Py_tracemalloc_config.max_nframe)
-            break;
+        if (traceback->nframe < _Py_tracemalloc_config.max_nframe) {
+            tracemalloc_get_frame(pyframe, &traceback->frames[traceback->nframe]);
+            assert(traceback->frames[traceback->nframe].filename != NULL);
+            traceback->nframe++;
+        }
+        if (traceback->total_nframe < UINT16_MAX)
+            traceback->total_nframe++;
     }
 }
 
@@ -456,6 +465,7 @@ traceback_new(void)
     /* get frames */
     traceback = tracemalloc_traceback;
     traceback->nframe = 0;
+    traceback->total_nframe = 0;
     traceback_get_frames(traceback);
     if (traceback->nframe == 0)
         return &tracemalloc_empty_traceback;
@@ -1001,6 +1011,7 @@ tracemalloc_init(void)
     PyUnicode_InternInPlace(&unknown_filename);
 
     tracemalloc_empty_traceback.nframe = 1;
+    tracemalloc_empty_traceback.total_nframe = 1;
     /* borrowed reference */
     tracemalloc_empty_traceback.frames[0].filename = unknown_filename;
     tracemalloc_empty_traceback.frames[0].lineno = 0;
@@ -1234,7 +1245,7 @@ trace_to_pyobject(unsigned int domain, trace_t *trace,
     PyObject *trace_obj = NULL;
     PyObject *obj;
 
-    trace_obj = PyTuple_New(3);
+    trace_obj = PyTuple_New(4);
     if (trace_obj == NULL)
         return NULL;
 
@@ -1258,6 +1269,13 @@ trace_to_pyobject(unsigned int domain, trace_t *trace,
         return NULL;
     }
     PyTuple_SET_ITEM(trace_obj, 2, obj);
+
+    obj = PyLong_FromUnsignedLong(trace->traceback->total_nframe);
+    if (obj == NULL) {
+        Py_DECREF(trace_obj);
+        return NULL;
+    }
+    PyTuple_SET_ITEM(trace_obj, 3, obj);
 
     return trace_obj;
 }
--- a/Modules/_tracemalloc.c
+++ b/Modules/_tracemalloc.c
@@ -79,7 +79,7 @@ __attribute__((packed))
 typedef struct {
     Py_uhash_t hash;
     /* Number of frames stored */
-    int nframe;
+    uint16_t nframe;
     /* Total number of frames the traceback had */
     uint16_t total_nframe;
     frame_t frames[1];
@@ -88,8 +88,10 @@ typedef struct {
 #define TRACEBACK_SIZE(NFRAME) \
         (sizeof(traceback_t) + sizeof(frame_t) * (NFRAME - 1))
 
-#define MAX_NFRAME \
-        ((INT_MAX - (int)sizeof(traceback_t)) / (int)sizeof(frame_t) + 1)
+/* The maximum number of frames is either:
+ - The maximum number of frames we can store in `traceback_t.nframe`
+ - The maximum memory size_t we can allocate */
+static const unsigned long MAX_NFRAME = Py_MIN(UINT16_MAX, ((SIZE_MAX - sizeof(traceback_t)) / sizeof(frame_t) + 1));
 
 
 static PyObject *unknown_filename = NULL;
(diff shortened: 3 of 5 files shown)


Modules/_tracemalloc.c
```diff
@@ -436,11 +443,12 @@ traceback_get_frames(traceback_t *traceback)
...
-        assert(traceback->frames[traceback->nframe].filename != NULL);
-        traceback->nframe++;
-        if (traceback->nframe == _Py_tracemalloc_config.max_nframe)
-            break;
+        if (traceback->nframe < _Py_tracemalloc_config.max_nframe) {
+            tracemalloc_get_frame(pyframe, &traceback->frames[traceback->nframe]);
+            assert(traceback->frames[traceback->nframe].filename != NULL);
+            traceback->nframe++;
+        }
+        traceback->length++;
```
vstinner: You should handle integer overflow: stop if INT_MAX is reached. If should be unlikely, but I prefer to be safe.

Modules/_tracemalloc.c
```diff
@@ -78,7 +78,10 @@ __attribute__((packed))
 
 typedef struct {
     Py_uhash_t hash;
+    /* Number of frames stored */
     int nframe;
+    /* Total number of frames the traceback had */
+    int length;
```
vstinner: Hum, maybe rename it to "total" or "total_nframe"?

Modules/_tracemalloc.c
```diff
@@ -1259,6 +1269,13 @@ trace_to_pyobject(unsigned int domain, trace_t *trace,
     }
     PyTuple_SET_ITEM(trace_obj, 2, obj);
 
+    obj = PyLong_FromSize_t(trace->traceback->length);
```
vstinner: nitpick: I would prefer to use PyLong_FromLong() which is more common and closer to int type.

Lib/tracemalloc.py
```diff
@@ -282,6 +282,13 @@ def size(self):
     def traceback(self):
         return Traceback(self._trace[2])
 
+    @property
+    def length(self):
```
vstinner: I suggest to rename it total_nframe and move it to Traceback.

vstinner (changes requested):
You should also update Doc/library/tracemalloc.rst documentation, document Trace.length new attribute (I suggest to add Traceback.total_nframe attribute instead):
https://docs.python.org/dev/library/tracemalloc.html#trace

Is it possible to load a pickle file created by Python 3.7 on Python 3.9 with this change?

Modules/_tracemalloc.c
jd: Good point!

Modules/_tracemalloc.c
jd: Let's go for `total_nframe`.

Lib/tracemalloc.py
jd: Ok!

Modules/_tracemalloc.c
jd: ok!

jd:
> Is it possible to load a pickle file created by Python 3.7 on Python 3.9 with this change?

No, it failed with:
```
>>> s.traces[0]
  File "/Users/jd/Source/cpython/Lib/tracemalloc.py", line 299, in __repr__
    % (self.domain, _format_size(self.size, False), self.traceback))
  File "/Users/jd/Source/cpython/Lib/tracemalloc.py", line 284, in traceback
    return Traceback(self._trace[2], self._trace[3])
IndexError: tuple index out of range
```

I fixed it by setting a default `nframe` to `None` if the info is not available.

jd:
@vstinner I just saw your comment on the issue. I'll update this to also reduce the size of the already existing `nframe` attribute.

Modules/_tracemalloc.c
```diff
@@ -78,15 +78,18 @@ __attribute__((packed))
...
+    uint16_t total_nframe;
     frame_t frames[1];
 } traceback_t;
 
 #define TRACEBACK_SIZE(NFRAME) \
         (sizeof(traceback_t) + sizeof(frame_t) * (NFRAME - 1))
 
 #define MAX_NFRAME \
-        ((INT_MAX - (int)sizeof(traceback_t)) / (int)sizeof(frame_t) + 1)
+        ((UINT16_MAX - (int)sizeof(traceback_t)) / (int)sizeof(frame_t) + 1)
```
jd: @vstinner I'm not really sure about this change. I actually can't understand how `MAX_NFRAME` is calculated here. Can you give me a hint?

vstinner: It looks like I wrote to ensure that TRACEBACK_SIZE(nframe) will not overflow, but I am not sure that using INT_MAX makes sense.

TRACEBACK_SIZE() maximum should be smaller than or equal to SIZE_MAX, not INT_MAX.

Maybe Py_MIN(INT_MAX, ((SIZE_MAX - sizeof(traceback_t)) / sizeof(frame_t) + 1) would be more correct?

jd: > TRACEBACK_SIZE() maximum should be smaller than or equal to SIZE_MAX, not INT_MAX.

Agreed. `TRACEBACK_SIZE()` is actually a `size_t`.

> Py_MIN(INT_MAX, ((SIZE_MAX - sizeof(traceback_t)) / sizeof(frame_t) + 1) would be more correct?

Yes. I think you either want the maximum size you can allocate or the maximum number of frame you can store in a `uint16_t`.

I'll change that.

It'll make the result 65535 in all cases for now :)

jd:
That should be ready for review but I see @bedevere-bot might be confused with labels. Ping @vstinner?

Lib/tracemalloc.py
```diff
@@ -378,7 +382,7 @@ def _match_traceback(self, traceback):
             return self._match_frame(filename, lineno)
 
     def _match(self, trace):
-        domain, size, traceback = trace
+        domain, size, traceback, length = trace
```
vstinner: I understand that this line breaks if loading from a Python 3.8. Maybe you should add the 4th field in _Traces constructor, if len(trace) == 3, rather than fixing it in the traceback property.

For example:

```
# Convert Python 3.8 trace: add nframe field (set to None)
if traces and len(traces[0]) == 3:
   traces = tuple(trace + (None,) for trace in traces)
```

Modules/_tracemalloc.c
```diff
@@ -78,15 +78,21 @@ __attribute__((packed))
...
 
 #define TRACEBACK_SIZE(NFRAME) \
         (sizeof(traceback_t) + sizeof(frame_t) * (NFRAME - 1))
 
+/* The maximum number of frames is either:
+ - The maximum number of frames we can store in `nframe`
+ - The maximum memory size_t we can allocate */
 #define MAX_NFRAME \
-        ((INT_MAX - (int)sizeof(traceback_t)) / (int)sizeof(frame_t) + 1)
+        Py_MIN(UINT16_MAX, ((SIZE_MAX - sizeof(traceback_t)) / sizeof(frame_t) + 1))
```
vstinner: Can you try to convert to macro to a constant variable? Otherwise, the macro is expended each time it is used, and the compiler might recompute it each time.

```
static const int MAX_NFRAME = Py_MIN(UINT16_MAX, ((SIZE_MAX - sizeof(traceback_t)) / sizeof(frame_t) + 1));
```

I suggest to use int (signed). If you want to use an unsigned int, you need to modify more change.

You may need to update code around `if (max_nframe < 1 || (unsigned long) max_nframe > MAX_NFRAME) {`.

Modules/_tracemalloc.c
```diff
@@ -78,15 +78,21 @@ __attribute__((packed))
...
+    /* Total number of frames the traceback had */
+    uint16_t total_nframe;
     frame_t frames[1];
 } traceback_t;
 
 #define TRACEBACK_SIZE(NFRAME) \
         (sizeof(traceback_t) + sizeof(frame_t) * (NFRAME - 1))
 
+/* The maximum number of frames is either:
+ - The maximum number of frames we can store in `nframe`
```
vstinner:
```suggestion
 - The maximum number of frames we can store in `traceback_t.nframe`
```

Lib/tracemalloc.py
```diff
@@ -182,15 +182,16 @@ class Traceback(Sequence):
...
 
-    def __init__(self, frames):
+    def __init__(self, frames, nframe):
         Sequence.__init__(self)
         # frames is a tuple of frame tuples: see Frame constructor for the
         # format of a frame tuple; it is reversed, because _tracemalloc
         # returns frames sorted from most recent to oldest, but the
         # Python API expects oldest to most recent
         self._frames = tuple(reversed(frames))
+        self.nframe = nframe
```
vstinner: I would prefer to make it private and add a read-only property. The API is designed to make all data read-only to ensure that all data remain consistent.

Doc/library/tracemalloc.rst
```diff
@@ -659,12 +659,19 @@ Traceback
 
    When a snapshot is taken, tracebacks of traces are limited to
    :func:`get_traceback_limit` frames. See the :func:`take_snapshot` function.
+   The original number of frames of the traceback is stored in the
+   :attr:`Traceback.nframe` attribute. That allows to know if a traceback has
+   been truncated by the traceback limit.
 
    The :attr:`Trace.traceback` attribute is an instance of :class:`Traceback`
    instance.
```
vstinner, line 670: You should declare the new field using ".. attribute:: " markup.

vstinner (commented):
In C, you chose "total_nframe" name. In Python, it's "nframe" in the API, and sometimes "length" in the code. I would prefer to always use the same name everywhere.

I like "total_nframe", but I don't have a strong preference :-)

Lib/tracemalloc.py
jd: Sure!

Modules/_tracemalloc.c
jd: Not sure it's worth it, but fine. AFAIU the macro is expanded only when pre-processing `_tracemalloc.` which should only happen once.

The result is an `unsigned long`, no matter what. We can cast it to something else, but that should be ok — the code compiles without warning AFAICT.

jd:
@vstinner Updated with your comments :)

Lib/tracemalloc.py
```diff
@@ -280,7 +287,10 @@ def size(self):
 
     @property
     def traceback(self):
-        return Traceback(self._trace[2])
+        # Python < 3.9
+        if len(self._trace) == 3:
+            return Traceback(self._trace[2], None)
```
vstinner: This code path becomes deadcode with "traces = tuple(trace + (None,) for trace in traces)" below, no? I guess that you can now remove this code.

Modules/_tracemalloc.c
```diff
@@ -436,11 +445,13 @@ traceback_get_frames(traceback_t *traceback)
...
-        assert(traceback->frames[traceback->nframe].filename != NULL);
-        traceback->nframe++;
-        if (traceback->nframe == _Py_tracemalloc_config.max_nframe)
-            break;
+        if (traceback->nframe < _Py_tracemalloc_config.max_nframe) {
+            tracemalloc_get_frame(pyframe, &traceback->frames[traceback->nframe]);
+            assert(traceback->frames[traceback->nframe].filename != NULL);
+            traceback->nframe++;
+        }
+        if (traceback->total_nframe++ < UINT16_MAX)
```
vstinner: Hum, I don't think that this code properly handles integer overlow, since traceback->total_nframe++ has a side effect. It should be:

```suggestion
        if (traceback->total_nframe < UINT16_MAX)
```

Modules/_tracemalloc.c
```diff
@@ -1046,10 +1059,10 @@ tracemalloc_start(int max_nframe)
     PyMemAllocatorEx alloc;
     size_t size;
 
-    if (max_nframe < 1 || max_nframe > MAX_NFRAME) {
+    if (max_nframe < 1 || (unsigned long) max_nframe > MAX_NFRAME) {
         PyErr_Format(PyExc_ValueError,
-                     "the number of frames must be in range [1; %i]",
-                     (int)MAX_NFRAME);
+                     "the number of frames must be in range [1; %u]",
```
vstinner: unsigned long requires `%lu` format:

```suggestion
                     "the number of frames must be in range [1; %lu]",
```

vstinner (commented):
Thanks, the updated PR looks better. Another round of minor comments.

You may mention total_nfrace in start() documentation: https://docs.python.org/dev/library/tracemalloc.html#tracemalloc.start Users may want to use it to adjust the next call to start().

Lib/tracemalloc.py
jd: You're right! I forgot we added that.

Modules/_tracemalloc.c
jd: Woh good catch, this is clearly a bad typo.

jd:
Here you go @vstinner :)

Lib/tracemalloc.py
```diff
@@ -182,15 +182,20 @@ class Traceback(Sequence):
     Sequence of Frame instances sorted from the oldest frame
     to the most recent frame.
     """
-    __slots__ = ("_frames",)
+    __slots__ = ("_frames", '_total_nframe')
 
-    def __init__(self, frames):
+    def __init__(self, frames, total_nframe):
```
vstinner: What do you think of making this new parameter optional (total_nframe=None), and use len(frames) by default, if it's not specified? It would simplify get_object_traceback() and test_tracemalloc.

Modules/_tracemalloc.c
```diff
@@ -1259,6 +1271,13 @@ trace_to_pyobject(unsigned int domain, trace_t *trace,
     }
     PyTuple_SET_ITEM(trace_obj, 2, obj);
 
+    obj = PyLong_FromLong(trace->traceback->total_nframe);
```
vstinner: Since it's unsigned, I suggest to use PyLong_FromUnsignedLong() to avoid the risk of introducing a compiler warning.

vstinner (commented):
The overall change now LGTM, but I have one suggestion about the tracemalloc.Traceback constructor which may simplify the code and leaves the API kind of backward compatible.

Modules/_tracemalloc.c
jd: Good catch. Done.

Lib/tracemalloc.py
jd: That means you lose information when you load traces from Python <= 3.8 as there's no way to know if the `total_nframe` was captured or simply faked at load time.
I think it's better to say that you either that information or you don't (so set to `None`) rather than trying to make the field ambiguous.

jd:
Last update I hope 🤞 @vstinner :)

jd:
> I'm not sure that I understand your comment. Do you suggest to modify your PR to allow total_nframe to be None? I think that I'm fine with that.

You were suggesting to always set `total_nframe` to some value — either the one provided in the argument list, or, if set to `None`, to compute it with `len(frames)`. I'm just saying that doing so would make `None` an invalid value for `total_nframe` which removes a piece of information that might be important. :)

> _group_by() and loading a Python 3.8 file would use total_nframe=None if I understand correctly. In that case, I suggest to modify Traceback.repr() to omit "total_nframe=%s" if it's None.
> By the way, "<Traceback %r total_nframe=%s>" may be better to first show the most important first (the frames). So "<Traceback %r total_nframe=%s>" or "<Traceback %r>" depending if total_nframe is None.

Why not, done.

> In your implementation, total_nframe is lost in tracemalloc.get_object_traceback(). I'm not sure if it's a deliberate choice or not.

No, it's not. It's just a "feature not implemented yet". :)

> If you want, I can write this complex change, once this PR lands.

Yeah, I think it'd be safer to go in a new PR that leverage this one rather than making this one heavier. 👍

vstinner:
> You were suggesting to always set total_nframe to some value — either the one provided in the argument list, or, if set to None, to compute it with len(frames).

In your PR, total_nframe is always set to an integer. When you don't know the value, you put len(frames), but you do that in the Traceback caller. I suggest to do that in the constructor to simplify your PR: tests, get_object_traceback(), _group_by(), etc. would just call Traceback(frames).

> I'm just saying that doing so would make None an invalid value for total_nframe which removes a piece of information that might be important. :)

Again, your PR always set total_nframe to a number, so I don't understand your remark. Do you plan to modify your PR to set total_nframe to None in some cases (ex: _group_by())? If yes, you should update the documentation to mention that total_nframe can be None.

jd:
> In your PR, total_nframe is always set to an integer. When you don't know the value, you put len(frames), but you do that in the Traceback caller. I suggest to do that in the constructor to simplify your PR: tests, get_object_traceback(), _group_by(), etc. would just call Traceback(frames).

That's mainly because the argument *was* mandatory in an earlier version IIRC. I can just remove it from the tests now if you prefer.

> Again, your PR always set total_nframe to a number, so I don't understand your remark. Do you plan to modify your PR to set total_nframe to None in some cases (ex: _group_by())? If yes, you should update the documentation to mention that total_nframe can be None.

The initial plan was to always have it set to the number of frames — remember, it was named `nframes`. So it would either be the number of frames we captured or — if available — the number of total frames that were originally in place. So `None` was not possible value.

Now, the way the PR has changed, `None` could be a value. I've updated the PR in that sense. Some tests now uses `None` has a value. I've updated the attribute doc.

Lib/tracemalloc.py
```diff
@@ -221,7 +226,12 @@ def __str__(self):
         return str(self[0])
 
     def __repr__(self):
-        return "<Traceback %r>" % (tuple(self),)
+        s = "<Traceback %r" % tuple(self)
+        if self._total_nframe is None:
+            s += ">"
+        else:
+            s += "total_nframe=%d>" % self.total_nframe
```
vstinner: You forgot a space. I like f-string :-) No need to be pedantic with %d.

```suggestion
            s += f" total_nframe={self.total_nframe}>"
```

Lib/tracemalloc.py
```diff
@@ -247,7 +257,8 @@ def format(self, limit=None, most_recent_first=False):
 def get_object_traceback(obj):
     """
     Get the traceback where the Python object *obj* was allocated.
-    Return a Traceback instance.
+    Return a Traceback instance
+    .
```
vstinner: Is it a deliberate chose to put the dot on a newline? If not, please revert this change.

Lib/tracemalloc.py
```diff
@@ -280,7 +291,7 @@ def size(self):
 
     @property
     def traceback(self):
-        return Traceback(self._trace[2])
+        return Traceback(self._trace[2], self._trace[3])
```
vstinner: Since total_nframe can be None, maybe now it would make more sense to handle Python 3.8 compatibility on demand in this property:

```
# Call Traceback(frames, total_nframe).
# For a Python 3.8 trace, call Traceback(frames).
return Traceback(*self._trace[2:])
```

(and revert change in `_Traces.__init__`)

Well, you wrote something like that in a previous PR :-)

vstinner (commented):
> The initial plan was to always have it set to the number of frames — remember, it was named nframes. So it would either be the number of frames we captured or — if available — the number of total frames that were originally in place. So None was not possible value.
> Now, the way the PR has changed, None could be a value. I've updated the PR in that sense. Some tests now uses None has a value. I've updated the attribute doc.

Oh, I see, the PR evolved and then was longer consistent. I'm fine with having total_nframe=None. It's just that it wasn't clear if it could be None or not. It's better with your update PR.

New review.

Lib/tracemalloc.py
jd: Thanks!

Lib/tracemalloc.py
jd: Good idea!

Lib/tracemalloc.py
jd: No sorry, PEBKAC.

Doc/library/tracemalloc.rst
```diff
@@ -659,13 +662,24 @@ Traceback
...
 
    .. versionchanged:: 3.7
       Frames are now sorted from the oldest to the most recent, instead of most recent to oldest.
 
+   .. attribute:: total_nframe
+
+      Total number of frames that composed the traceback before truncation.
+      This attribute can be set to `None` if the information is not available.
```
vstinner: Travis CI fails with:

```
[2] library/tracemalloc.rst:678: default role used
````

I guess that it should be:

```suggestion
      This attribute can be set to ``None`` if the information is not available.
```

vstinner (approved):
LGTM. But the CI blocks the PR.

jd:
Thanks, fixed the doc!

vstinner:
> https://buildbot.python.org/all/#builders/103/builds/3247

That's https://bugs.python.org/issue35998

PR #15548, opened 2019-08-27
bpo-37064: add option -k to Tools/scripts/pathfix.py

PatrikKopkan:
- with this option it is possible to keep flags when changing shebang line

https://bugs.python.org/issue37064

### DIFF
--- a/Lib/test/test_tools/test_pathfix.py
+++ b/Lib/test/test_tools/test_pathfix.py
@@ -20,33 +20,39 @@ class TestPathfixFunctional(unittest.TestCase):
         with open(self.temp_file, 'r', encoding='utf8') as f:
             output = f.read()
             lines = output.split('\n')
-            self.assertEqual(lines[1:], ['print("Hello world")\n']
-            shebang = lines[0][:-1]  # strip the newline
+            self.assertEqual(lines[1:], ['print("Hello world")', ''])
+            shebang = lines[0]
             return shebang
-
-    def test_pathfix_keeping_flags(self):
-        self.assertEqual(
-            self.pathfix(
-                '#! /usr/bin/env python -R',
-                ['-i', '/usr/bin/python3', '-k',]),
-            '#! /usr/bin/python3 -R\n',
-        )
-
-    def test_pathfix_without_keeping_flags(self):
-        self.assertEqual(
-            self.pathfix(
-                '#! /usr/bin/env python -R',
-                ['-i', '/usr/bin/python3',]),
-            '#! /usr/bin/python3\n',
-        )
 
     def test_pathfix(self):
         self.assertEqual(
             self.pathfix(
                 '#! /usr/bin/env python',
                 ['-i', '/usr/bin/python3',]),
-            '#! /usr/bin/python3\n',
+            '#! /usr/bin/python3',
         )
+        self.assertEqual(
+            self.pathfix(
+                '#! /usr/bin/env python -R',
+                ['-i', '/usr/bin/python3', ]),
+            '#! /usr/bin/python3',
+        )
+
+    def test_pathfix_keeping_flags(self):
+        self.assertEqual(
+            self.pathfix(
+                '#! /usr/bin/env python -R',
+                ['-i', '/usr/bin/python3', '-k',]),
+            '#! /usr/bin/python3 -R',
+        )
+        self.assertEqual(
+            self.pathfix(
+                '#! /usr/bin/env python',
+                ['-i', '/usr/bin/python3', '-k',]),
+            '#! /usr/bin/python3',
+        )
+
+
 
 if __name__ == '__main__':
     unittest.main()
--- a/Lib/test/test_tools/test_pathfix.py
+++ b/Lib/test/test_tools/test_pathfix.py
@@ -7,28 +7,42 @@ from test import support
 class TestPathfixFunctional(unittest.TestCase):
     script = os.path.join(scriptsdir, 'pathfix.py')
 
-    @classmethod
-    def setupUpClass(cls):
-        self.addCleanup(support.unlink, support.TESTFN)
-
     def setUp(self):
         self.temp_file = support.TESTFN
+        self.addCleanup(support.unlink, support.TESTFN)
 
+    def pathfix(self, shebang, pathfix_flags):
         with open(self.temp_file, 'w') as f:
-            f.write('#! /usr/bin/env python -R\n' + 'print("Hello world")\n')
+            f.write(f'{shebang}\n' + 'print("Hello world")\n')
+        subprocess.call([sys.executable, self.script, *pathfix_flags, self.temp_file])
+        with open(self.temp_file) as f:
+            output = f.read()
+        print(output)
+        return output
 
     def test_pathfix_keeping_flags(self):
-        subprocess.call([sys.executable, self.script, '-i', '/usr/bin/python', '-k', '-n', self.temp_file])
-        with open(self.temp_file) as f:
-            output = f.read()
-        self.assertEqual(output, '#! /usr/bin/python -R\n' + 'print("Hello world")\n')
+        self.assertEqual(
+            self.pathfix(
+                '#! /usr/bin/env python -R',
+                ['-i', '/usr/bin/python', '-k', '-n']),
+            '#! /usr/bin/python -R\n' + 'print("Hello world")\n'
+        )
 
     def test_pathfix_without_keeping_flags(self):
-        subprocess.call([sys.executable, self.script, '-i', '/usr/bin/python', '-n', self.temp_file])
-        with open(self.temp_file) as f:
-            output = f.read()
-        self.assertEqual(output, '#! /usr/bin/python\n' + 'print("Hello world")\n')
+        self.assertEqual(
+            self.pathfix(
+                '#! /usr/bin/env python -R',
+                ['-i', '/usr/bin/python', '-n']),
+            '#! /usr/bin/python\n' + 'print("Hello world")\n'
+        )
 
+    def test_pathfix(self):
+        self.assertEqual(
+            self.pathfix(
+                '#! /usr/bin/env python',
+                ['-i', '/usr/bin/python', '-n']),
+            '#! /usr/bin/python\n' + 'print("Hello world")\n'
+        )
 
 if __name__ == '__main__':
     unittest.main()
--- a/Lib/test/test_tools/test_pathfix.py
+++ b/Lib/test/test_tools/test_pathfix.py
@@ -1,0 +1,34 @@
+import unittest, os, subprocess
+from test.test_tools import import_tool, scriptsdir
+import sys
+from test import support
+
+
+class TestPathfixFunctional(unittest.TestCase):
+    script = os.path.join(scriptsdir, 'pathfix.py')
+
+    @classmethod
+    def setupUpClass(cls):
+        self.addCleanup(support.unlink, support.TESTFN)
+
+    def setUp(self):
+        self.temp_file = support.TESTFN
+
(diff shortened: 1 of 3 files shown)


Tools/scripts/pathfix.py
```diff
@@ -33,16 +38,20 @@
 new_interpreter = None
 preserve_timestamps = False
 create_backup = True
+add_flag = None
```
vstinner: This variable looks unused.

Tools/scripts/pathfix.py
```diff
@@ -164,12 +180,28 @@ def fix(filename):
     # Return success
     return 0
 
+
+def parse_shebang(shebangline):
+    end = len(shebangline)
+    start = shebangline.find(b' -', 0, end)  # .find() returns index at space
```
vstinner: Please remove the comment, IMHO it's not really helpful.

", 0, end" are the default value and so can be removed.

Tools/scripts/pathfix.py
```diff
@@ -164,12 +180,28 @@ def fix(filename):
     # Return success
     return 0
 
+
+def parse_shebang(shebangline):
+    end = len(shebangline)
```
vstinner: Please add a comment that on Linux, the first whitespace (U+0020) indicates the beginning of options.

Tools/scripts/pathfix.py
```diff
@@ -164,12 +180,28 @@ def fix(filename):
...
+    start = shebangline.find(b' -', 0, end)  # .find() returns index at space
+    if start == -1:
+        return b'', b''
+    print(shebangline[(start + 2):end])
```
vstinner: Don't commit debug print please (remove it).

Tools/scripts/pathfix.py
```diff
@@ -164,12 +180,28 @@ def fix(filename):
...
+    return shebangline[(start + 2):end]
+
+
 def fixline(line):
+    global keep_flags
```
vstinner: I don't think that this global statement is useful. Variables are global by default.

Tools/scripts/pathfix.py
```diff
@@ -164,12 +180,28 @@ def fix(filename):
     # Return success
     return 0
 
+
+def parse_shebang(shebangline):
+    end = len(shebangline)
+    start = shebangline.find(b' -', 0, end)  # .find() returns index at space
+    if start == -1:
+        return b'', b''
```
vstinner: I don't understand why it returns a tuple. Why not returning b''? It looks like a bug.

Tools/scripts/pathfix.py
```diff
@@ -164,12 +180,28 @@ def fix(filename):
...
+    global keep_flags
     if not line.startswith(b'#!'):
         return line
+
     if b"python" not in line:
         return line
+    if keep_flags:
+        flags = parse_shebang(line)
+        return b'#! ' + new_interpreter + b" -" + flags
     return b'#! ' + new_interpreter + b'\n'
```
vstinner: I would prefer to have:

```
    if keep_flags:
        flags = parse_shebang(line)
    else:
        flags = b''
    return b'#! ' + new_interpreter + flags + b'\n'
```

So parse_shebang() must remove the trailing newline.

Lib/test/test_tools/test_pathfix.py
```diff
@@ -0,0 +1,34 @@
...
+        subprocess.call([sys.executable, self.script, '-i', '/usr/bin/python', '-k', '-n', self.temp_file])
+        with open(self.temp_file) as f:
+            output = f.read()
+        self.assertEqual(output, '#! /usr/bin/python -R\n' + 'print("Hello world")\n')
+
+    def test_pathfix_without_keeping_flags(self):
+        subprocess.call([sys.executable, self.script, '-i', '/usr/bin/python', '-n', self.temp_file])
+        with open(self.temp_file) as f:
+            output = f.read()
+        self.assertEqual(output, '#! /usr/bin/python\n' + 'print("Hello world")\n')
```
vstinner: Please add a test using -k when the shebang does not contain any flag.

Lib/test/test_tools/test_pathfix.py
```diff
@@ -0,0 +1,34 @@
...
+
+    @classmethod
+    def setupUpClass(cls):
+        self.addCleanup(support.unlink, support.TESTFN)
+
+    def setUp(self):
+        self.temp_file = support.TESTFN
+
+        with open(self.temp_file, 'w') as f:
+            f.write('#! /usr/bin/env python -R\n' + 'print("Hello world")\n')
```
vstinner: I would prefer to have an helper method which takes two arguments: (shebang, pathfix_flags) and return the new shebang. To get tests which look like:

```
def test_keep_flags(self):
  shebang = '#! /usr/bin/env python -R'
  self.assertEqual(
    self.pathfix(shebang, ['-i', '/usr/bin/python']),
    '#! /usr/bin/python')
  self.assertEqual(
    self.pathfix(shebang, ['-i', '/usr/bin/python', '-k']),
    '#! /usr/bin/python -R')
```

Lib/test/test_tools/test_pathfix.py
```diff
@@ -0,0 +1,34 @@
...
+import sys
+from test import support
+
+
+class TestPathfixFunctional(unittest.TestCase):
+    script = os.path.join(scriptsdir, 'pathfix.py')
+
+    @classmethod
+    def setupUpClass(cls):
+        self.addCleanup(support.unlink, support.TESTFN)
```
vstinner: It doesn't work: self doesn't exist. You may use setUp().

Misc/NEWS.d/next/Tools-Demos/2019-05-27-15-26-12.bpo-37064.k_SPW2.rst
```diff
@@ -0,0 +1 @@
+Add flag -k to pathscript.py. This flag preserve flags from shebang before change and adds them after change.
```
vstinner: I suggest: "Add flag -k to pathscript.py script: preserve shebang flags."

vstinner (changes requested):

Lib/test/test_tools/test_pathfix.py
```diff
@@ -0,0 +1,48 @@
+import unittest, os, subprocess
```
vstinner: Please put one import per line: see PEP 8.

Try also to sort imports.

Lib/test/test_tools/test_pathfix.py
```diff
@@ -0,0 +1,48 @@
...
+        self.temp_file = support.TESTFN
+        self.addCleanup(support.unlink, support.TESTFN)
+
+    def pathfix(self, shebang, pathfix_flags):
+        with open(self.temp_file, 'w') as f:
+            f.write(f'{shebang}\n' + 'print("Hello world")\n')
+        subprocess.call([sys.executable, self.script, *pathfix_flags, self.temp_file])
+        with open(self.temp_file) as f:
+            output = f.read()
+        print(output)
```
vstinner: Don't forget to remove debug print when you submit a PR.

Lib/test/test_tools/test_pathfix.py
```diff
@@ -0,0 +1,48 @@
...
+
+class TestPathfixFunctional(unittest.TestCase):
+    script = os.path.join(scriptsdir, 'pathfix.py')
+
+    def setUp(self):
+        self.temp_file = support.TESTFN
+        self.addCleanup(support.unlink, support.TESTFN)
+
+    def pathfix(self, shebang, pathfix_flags):
+        with open(self.temp_file, 'w') as f:
```
vstinner: I would prefer to specify the encoding: pass encoding="utf8" to this open, and the other one below.

Lib/test/test_tools/test_pathfix.py
```diff
@@ -0,0 +1,48 @@
...
+        with open(self.temp_file) as f:
+            output = f.read()
+        print(output)
+        return output
+
+    def test_pathfix_keeping_flags(self):
+        self.assertEqual(
+            self.pathfix(
+                '#! /usr/bin/env python -R',
+                ['-i', '/usr/bin/python', '-k', '-n']),
```
vstinner: Please remove -n from tests and always put it in pathfix() function.

Lib/test/test_tools/test_pathfix.py
```diff
@@ -0,0 +1,48 @@
...
+            '#! /usr/bin/python -R\n' + 'print("Hello world")\n'
```
vstinner: I don't think that it's useful to check for 'print("Hello world")\n' in each test. I would prefer to test it in pathfix() and just return the first line.

Something like this in pathfix():

lines = output.split('\n')
self.assertEqual(lines[1:], ['print("Hello world")\n']
shebang = lines[0][:-1]  # strip the newline
return shebang

Lib/test/test_tools/test_pathfix.py
```diff
@@ -0,0 +1,48 @@
...
+        subprocess.call([sys.executable, self.script, *pathfix_flags, self.temp_file])
+        with open(self.temp_file) as f:
+            output = f.read()
+        print(output)
+        return output
+
+    def test_pathfix_keeping_flags(self):
+        self.assertEqual(
+            self.pathfix(
+                '#! /usr/bin/env python -R',
```
vstinner, line 43: nitpick: instead of "/usr/bin/env python" and "/usr/bin/python" ... which is Python 2 in most Linux distributions... I would prefer to use "/usr/bin/env python" and "/usr/bin/python3" which is Python 3 in all Linux distributions :-)

Lib/test/test_tools/test_pathfix.py
```diff
@@ -0,0 +1,52 @@
...
+            f.write(f'{shebang}\n' + 'print("Hello world")\n')
+        subprocess.call([sys.executable, self.script, *pathfix_flags, '-n', self.temp_file])
+        with open(self.temp_file, 'r', encoding='utf8') as f:
+            output = f.read()
+            lines = output.split('\n')
+            self.assertEqual(lines[1:], ['print("Hello world")\n']
+            shebang = lines[0][:-1]  # strip the newline
+            return shebang
+
+    def test_pathfix_keeping_flags(self):
```
vstinner: I would prefer:
* put test_pathfix_without_keeping_flags() code into test_pathfix()
* move test_pathfix_keeping_flags() after test_pathfix()
* add a test in test_pathfix_keeping_flags(): use -k but the shebang doesn't contain any flag (the bug that you just fixed)

Tools/scripts/pathfix.py
```diff
@@ -164,12 +179,27 @@ def fix(filename):
     # Return success
     return 0
 
+
+def parse_shebang(shebangline):
+    shebangline = shebangline.strip()
+    end = len(shebangline)
+    start = shebangline.find(b' -', 0, end)
```
vstinner: "end" argument is useless: shebangline.find(b' -') behaves the same and is simpler. Remove 0 and end.

Tools/scripts/pathfix.py
vstinner: Ping @PatrikKopkan?

vstinner (changes requested):
Thanks, the code now looks way better! A few more comments.

Tools/scripts/pathfix.py
PatrikKopkan: I have zero idea how I deleted. In editor I still have:
        ```return b'#! ' + new_interpreter + b' -' + flags + b'\n'```

vstinner (approved):
LGTM.

PR #15549, opened 2019-08-27
Fixed a real little typo in the code comments

shashi278:


jdemeyer (changes requested):
This is not a typo. "sane" means valid, correct in this context.

shashi278:
My bad. Never mind :)

PR #15555, opened 2019-08-27
bpo-23878: Remove the key parameter of _Py_FindEnvConfigValue()

ZackerySpytz:
Suggested by @vstinner in GH-15424.

https://bugs.python.org/issue23878

### DIFF
--- a/Python/pathconfig.c
+++ b/Python/pathconfig.c
@@ -734,8 +734,7 @@ _PyPathConfig_ComputeSysPath0(const PyWideStringList *argv, PyObject **path0_p)
 /* Search for a prefix value in an environment file (pyvenv.cfg).
    If found, copy it into the provided buffer. */
 int
-_Py_FindEnvConfigValue(FILE *env_file, const wchar_t *key,
-                       wchar_t *value, size_t value_size)
+_Py_FindEnvConfigValue(FILE *env_file, wchar_t *value, size_t value_size)
 {
     int result = 0; /* meaning not found */
     char buffer[MAXPATHLEN * 2 + 1];  /* allow extra for key, '=', etc. */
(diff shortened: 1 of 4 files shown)


Python/pathconfig.c
```diff
@@ -734,8 +734,7 @@ _PyPathConfig_ComputeSysPath0(const PyWideStringList *argv, PyObject **path0_p)
 /* Search for a prefix value in an environment file (pyvenv.cfg).
```
vstinner, line 734: It doesn't search "for a prefix". It searchs for the the line "home=<path>" and return `<path>`. It tolerates whitespaces around "=".

vstinner:
IMO it's fine to reuse https://bugs.python.org/issue23878 for this cleanup change. It gives more context about this change, than no bpo number.

zooba (changes requested):
If we're going to remove the key argument, we should also rename the function to reflect that it only finds `home` somehow.

Personally, I see no benefit here, and we may as well leave it alone. But if others think it's cleaner to have `_Py_FindEnvConfigHomeValue` instead of specifying `L"home"` in the call, I'm not going to fight it.

csabella:
@ZackerySpytz, please resolve the merge conflicts and address the code review.  If this change is no longer relevant, please close the PR.  Thanks!

csabella:
@ZackerySpytz ping

iritkatriel:
https://bugs.python.org/issue23878 is closed. What is that status of this PR?

vstinner:
This PR is still relevant, but it's outdated. @ZackerySpytz: can you please rebase it on the main branch? Or create a new PR.

kumaraditya303:
Closing as it is outdated and bpo is closed, if required a new PR can be opened or this can reopened.

PR #15082, opened 2019-08-02
[3.8] Improve message for openssl lookup

vhquang:
I was trying to test the pre-release 3.8b3, and since I have multiple installation of OpenSSL, having this improved message with more specific on which type of OpenSSL header would be needed, will definitely make my hunt go easier.

Also, this is my first PR to Python. If there is anything else that can be improved on my PR, kindly let me know.

PR #15556, opened 2019-08-27
Improve message for openssl lookup

vhquang:
This is a re-open from #15082 , in which one of my reviewers suggested it go on `master`. It is a cosmetic change on the display message, so I don't think any issue number is needed.

Rationale for the change:
> I was trying to test the pre-release 3.8b3, and since I have multiple installation of OpenSSL, having this improved message with more specific on which type of OpenSSL header would be needed, will definitely make my hunt go easier.

It is still my first PR to Python, so kindly let me know if anything else that I can improve.

### DIFF
--- a/configure
+++ b/configure
@@ -17207,8 +17207,8 @@ fi
     if ! $found; then
         OPENSSL_INCLUDES=
         for ssldir in $ssldirs; do
-            { $as_echo "$as_me:${as_lineno-$LINENO}: checking for openssl/ssl.h in $ssldir" >&5
-$as_echo_n "checking for openssl/ssl.h in $ssldir... " >&6; }
+            { $as_echo "$as_me:${as_lineno-$LINENO}: checking for include/openssl/ssl.h in $ssldir" >&5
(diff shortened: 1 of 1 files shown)


brandtbucher (approved):

tiran (changes requested):
The configure file is auto-generated from configure.ac and m4 macro files. You cannot modify the generated output. It will be overwritten the next time the configure script is regeerated.

The OpenSSL detection code uses https://www.gnu.org/software/autoconf-archive/ax_check_openssl.html . Please take this up to the author of the m4 macro for ax_check_openssl. I don't want to deviate from upstream sources.

vhquang:
I didn't know the configure script is generated by `autoconf-archive/ax_check_openssl.m4`. I agree let take this upstream.
Thanks Christian and Brandt for the pointers!

vhquang:
Bring the change to `autoconf-archive` instead of here.

vhquang:
@tiran Just to close the loop, the change we discussed is merged upstream. Do I need to do anything in CPython so we will pick up the change? If so, can you give me few pointers?

PR #15558, opened 2019-08-28
bpo-37966: Fully implement the UAX #15 quick-check algorithm.

gnprice:
The purpose of the `unicodedata.is_normalized` function is to answer
the question `str == unicodedata.normalized(form, str)` more
efficiently than writing just that, by using the "quick check"
optimization described in the Unicode standard in UAX #15.

However, it turns out the code doesn't implement the full algorithm
from the standard, and as a result we often miss the optimization and
end up having to compute the whole normalized string after all.

Implement the standard's algorithm.  This greatly speeds up
`unicodedata.is_normalized` in many cases where our partial variant
of quick-check had been returning MAYBE and the standard algorithm
returns NO.

At a quick test on my desktop, the existing code takes about 4.4 ms/MB
(so 4.4 ns per byte) when the partial quick-check returns MAYBE and it
has to do the slow normalize-and-compare:

    $ build.base/python -m timeit -s 'import unicodedata; s = "\uf900"*500000' \
        -- 'unicodedata.is_normalized("NFD", s)'
    50 loops, best of 5: 4.39 msec per loop

With this patch, it gets the answer instantly (58 ns) on the same 1 MB
string:

    $ build.dev/python -m timeit -s 'import unicodedata; s = "\uf900"*500000' \
        -- 'unicodedata.is_normalized("NFD", s)'
    5000000 loops, best of 5: 58.2 nsec per loop

https://bugs.python.org/issue37966

### DIFF
--- a/Lib/test/test_unicodedata.py
+++ b/Lib/test/test_unicodedata.py
@@ -208,6 +208,8 @@ class UnicodeFunctionsTest(UnicodeDatabaseTest):
         self.assertEqual(self.db.normalize('NFC', u11a7_str_a), u11a7_str_b)
         self.assertEqual(self.db.normalize('NFC', u11c3_str_a), u11c3_str_b)
 
+    # For tests of unicodedata.is_normalized / self.db.is_normalized ,
+    # see test_normalization.py .
 
     def test_east_asian_width(self):
         eaw = self.db.east_asian_width
--- a/Modules/unicodedata.c
+++ b/Modules/unicodedata.c
@@ -19,6 +19,8 @@
 #include "ucnhash.h"
 #include "structmember.h"
 
+#include <stdbool.h>
+
 _Py_IDENTIFIER(NFC);
 _Py_IDENTIFIER(NFD);
 _Py_IDENTIFIER(NFKC);
@@ -793,7 +795,7 @@ typedef enum {YES = 0, MAYBE = 1, NO = 2} QuickcheckResult;
  */
 static QuickcheckResult
 is_normalized_quickcheck(PyObject *self, PyObject *input,
-                         int nfc, int k, int yes_only)
+                         int nfc, int k, bool yes_only)
 {
     /* An older version of the database is requested, quickchecks must be
        disabled. */
@@ -892,7 +894,7 @@ unicodedata_UCD_is_normalized_impl(PyObject *self, PyObject *form,
         return NULL;
     }
 
-    m = is_normalized_quickcheck(self, input, nfc, k, 0);
+    m = is_normalized_quickcheck(self, input, nfc, k, false);
 
     if (m == MAYBE) {
         cmp = (nfc ? nfc_nfkc : nfd_nfkd)(self, input, k);
@@ -938,28 +940,28 @@ unicodedata_UCD_normalize_impl(PyObject *self, PyObject *form,
     }
 
     if (_PyUnicode_EqualToASCIIId(form, &PyId_NFC)) {
-        if (is_normalized_quickcheck(self, input, 1, 0, 1) == YES) {
+        if (is_normalized_quickcheck(self, input, 1, 0, true) == YES) {
             Py_INCREF(input);
             return input;
         }
         return nfc_nfkc(self, input, 0);
     }
     if (_PyUnicode_EqualToASCIIId(form, &PyId_NFKC)) {
-        if (is_normalized_quickcheck(self, input, 1, 1, 1) == YES) {
+        if (is_normalized_quickcheck(self, input, 1, 1, true) == YES) {
             Py_INCREF(input);
             return input;
         }
         return nfc_nfkc(self, input, 1);
     }
     if (_PyUnicode_EqualToASCIIId(form, &PyId_NFD)) {
-        if (is_normalized_quickcheck(self, input, 0, 0, 1) == YES) {
+        if (is_normalized_quickcheck(self, input, 0, 0, true) == YES) {
             Py_INCREF(input);
             return input;
         }
         return nfd_nfkd(self, input, 0);
     }
     if (_PyUnicode_EqualToASCIIId(form, &PyId_NFKD)) {
-        if (is_normalized_quickcheck(self, input, 0, 1, 1) == YES) {
+        if (is_normalized_quickcheck(self, input, 0, 1, true) == YES) {
             Py_INCREF(input);
             return input;
         }
--- a/Modules/unicodedata.c
+++ b/Modules/unicodedata.c
@@ -791,8 +791,12 @@ is_normalized(PyObject *self, PyObject *input, int nfc, int k)
     if (self && UCD_Check(self))
         return NO;
 
-    /* The two quickcheck bits at this shift mean 0=Yes, 1=Maybe, 2=No,
-       as described in http://unicode.org/reports/tr15/#Annex8. */
+    /* This is an implementation of the following algorithm:
+       https://www.unicode.org/reports/tr15/#Detecting_Normalization_Forms
+       See there for background.
+    */
+
+    /* The two quickcheck bits at this shift mean 0=Yes, 1=Maybe, 2=No. */
     quickcheck_mask = 3 << ((nfc ? 4 : 0) + (k ? 2 : 0));
 
     i = 0;
--- a/Modules/unicodedata.c
+++ b/Modules/unicodedata.c
@@ -787,16 +787,14 @@ typedef enum {YES = 0, MAYBE = 1, NO = 2} QuickcheckResult;
  *
  * If `yes_only` is true, then return MAYBE as soon as we determine
  * the answer is not YES.
+ *
+ * For background and details on the algorithm, see UAX #15:
+ *   https://www.unicode.org/reports/tr15/#Detecting_Normalization_Forms
  */
 static QuickcheckResult
 is_normalized_quickcheck(PyObject *self, PyObject *input,
                          int nfc, int k, int yes_only)
 {
-    /* This is an implementation of the following algorithm:
-       https://www.unicode.org/reports/tr15/#Detecting_Normalization_Forms
-       See there for background.
-    */
-
     /* An older version of the database is requested, quickchecks must be
        disabled. */
     if (self && UCD_Check(self))
--- a/Modules/unicodedata.c
+++ b/Modules/unicodedata.c
@@ -783,9 +783,14 @@ typedef enum {YES = 0, MAYBE = 1, NO = 2} QuickcheckResult;
  *
  * Return YES or NO if quickcheck determines the input is certainly
  * normalized or certainly not, and MAYBE if quickcheck is unable to
- * tell. */
+ * tell.
+ *
+ * If `yes_only` is true, then return MAYBE as soon as we determine
+ * the answer is not YES.
+ */
 static QuickcheckResult
-is_normalized_quickcheck(PyObject *self, PyObject *input, int nfc, int k)
+is_normalized_quickcheck(PyObject *self, PyObject *input,
+                         int nfc, int k, int yes_only)
 {
     /* This is an implementation of the following algorithm:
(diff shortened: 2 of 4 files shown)


Modules/unicodedata.c
```diff
@@ -775,25 +775,42 @@ nfc_nfkc(PyObject *self, PyObject *input, int k)
...
 }
 
-typedef enum {YES, NO, MAYBE} NormalMode;
-
-/* Return YES if the input is certainly normalized, NO or MAYBE if it might not be. */
-static NormalMode
-is_normalized(PyObject *self, PyObject *input, int nfc, int k)
+// This needs to match the logic in makeunicodedata.py
+// which constructs the quickcheck data.
+typedef enum {YES = 0, MAYBE = 1, NO = 2} QuickcheckResult;
```
benjaminp, line 782: Maybe `makeunicodedata.py` should output this enum (with better name namespacing) then?

Lib/test/test_unicodedata.py
```diff
@@ -208,6 +208,8 @@ def test_issue29456(self):
         self.assertEqual(self.db.normalize('NFC', u11a7_str_a), u11a7_str_b)
         self.assertEqual(self.db.normalize('NFC', u11c3_str_a), u11c3_str_b)
 
+    # For tests of unicodedata.is_normalized / self.db.is_normalized ,
+    # see test_normalization.py .
```
benjaminp, line 212: ugh, I would support merging `test_normalization` into this file for clarity.

Modules/unicodedata.c
```diff
@@ -775,25 +775,42 @@ nfc_nfkc(PyObject *self, PyObject *input, int k)
...
+                         int nfc, int k, int yes_only)
 {
-    Py_ssize_t i, len;
-    int kind;
-    void *data;
-    unsigned char prev_combining = 0, quickcheck_mask;
+    /* This is an implementation of the following algorithm:
+       https://www.unicode.org/reports/tr15/#Detecting_Normalization_Forms
+       See there for background.
+    */
```
benjaminp: Seems like this belongs in the "doc comment" above the function.

Modules/unicodedata.c
```diff
@@ -775,25 +775,42 @@ nfc_nfkc(PyObject *self, PyObject *input, int k)
...
+ * Return YES or NO if quickcheck determines the input is certainly
+ * normalized or certainly not, and MAYBE if quickcheck is unable to
+ * tell.
+ *
+ * If `yes_only` is true, then return MAYBE as soon as we determine
+ * the answer is not YES.
+ */
+static QuickcheckResult
+is_normalized_quickcheck(PyObject *self, PyObject *input,
+                         int nfc, int k, int yes_only)
```
benjaminp: These "boolean int" parameters could be actual `bool`s.

Modules/unicodedata.c
```diff
@@ -775,25 +775,42 @@ nfc_nfkc(PyObject *self, PyObject *input, int k)
...
+// which constructs the quickcheck data.
+typedef enum {YES = 0, MAYBE = 1, NO = 2} QuickcheckResult;
+
+/* Run the Unicode normalization "quickcheck" algorithm.
+ *
+ * Return YES or NO if quickcheck determines the input is certainly
+ * normalized or certainly not, and MAYBE if quickcheck is unable to
+ * tell.
+ *
+ * If `yes_only` is true, then return MAYBE as soon as we determine
```
benjaminp, line 790: `yes_only` is a bizarre name for a parameter that causes the function to return `MAYBE`.

benjaminp (commented):
Thanks. Looks good. Here are a few nitty comments.

Modules/unicodedata.c
gnprice: Ah, good idea!

Do you think it'd be cleaner to add that to this PR, or make a small separate one?

Lib/test/test_unicodedata.py
gnprice: Yeah; as it is, it actually caused me to think this function had evaded being tested at all -- I'd gone as far as to write up some tests myself before I spotted that other file.

A bit of context on someone's thinking in having a separate file can be found above at line 178:
```
        # The rest can be found in test_normalization.py
        # which requires an external file.
```
I don't see why that means the test code should be in a separate file, though. There's already a try/skip mechanism to deal with the external file being unavailable. (I'm guessing that if I looked in the history I'd find that that mechanism wasn't there when the separate file was first added.)

Happy to send a PR to fold that file in.

Modules/unicodedata.c
gnprice: Hmm, yeah, good thought. The fine details of the algorithm make sense for an implementation comment... but really even the interface of this function only completely makes sense after at least skimming the high-level shape of that section.

(This is one of those spots where I followed what the existing code had, and didn't think too hard about whether something else would be better.)

Modules/unicodedata.c
gnprice: Oh wow it's the future! :slightly_smiling_face: I guess that's a C99 feature, and one of those we now accept using. Good to know.

I'm not sure I want to convert the existing two parameters within this PR; it seems like that could muddy the diff a bit with changes that are orthogonal to what it's really doing. But I guess I can see how it looks.

Modules/unicodedata.c
gnprice: Yeah, I'm not really a fan of this name; I'd be glad for a better one.

The idea is that this flag says the caller doesn't care about NO vs. MAYBE; it only cares whether the answer is YES.

Modules/unicodedata.c
gnprice: Pushed a change to `bool` for the new parameter.

I was a bit surprised to find that I needed to add an `#include <stdbool.h>` line to the file! Perhaps that should be in a central include file, in the same way as `limits.h` and `inttypes.h` are. That's certainly orthogonal to this PR, though.

gnprice: I made a quick further commit converting `nfc` and `k` to `bool` too (including in the callers, and using `true`/`false` rather than 1/0 for literal values.)

I'm ambivalent about adding it here. It expands the diff a bit -- was:
 1 file changed, 51 insertions(+), 24 deletions(-)
and becomes:
 1 file changed, 57 insertions(+), 30 deletions(-)
Slightly inclined to leave it out, but if you'd rather include it I'll do that.

Modules/unicodedata.c
benjaminp: Could be done latter, I think.

Modules/unicodedata.c
benjaminp: What makes the diffstat larger? Aren't you already touching all callers, since you added a parameter?

benjaminp: I think we want to avoid polluting public headers with a `stdbool.h` include—`bool` is a bit of a name grab in arbitrary mature C codebases. Using `stdbool.h` where it's needed is also consistent with IWYU.

Modules/unicodedata.c
benjaminp: Maybe invert the sense and call it `want_definite` meaning "process the string until we're sure it's not normalized or we reach the end"?

Modules/unicodedata.c
gnprice: > I think we want to avoid polluting public headers with a `stdbool.h` include—`bool` is a bit of a name grab in arbitrary mature C codebases.

Ah, fair enough. I think that's consistent with putting it in a generic internal-only header, though. Within the tree, the very few existing uses are either stdbool, or (in `Modules/expat/xmltok.c`) a sort of manual stdbool.

> Using `stdbool.h` where it's needed is also consistent with IWYU.

Hmm true. I guess IWYU doesn't feel like how this codebase is mostly written... but maybe it's best to make it more so.

gnprice: > What makes the diffstat larger? Aren't you already touching all callers, since you added a parameter?

Here's the extra commit: https://github.com/gnprice/cpython/commit/896c970f5
The lines not otherwise touched are where we declare and then fiddle with the `nfc` and `k` locals in `unicodedata_UCD_is_normalized_impl`.

Modules/unicodedata.c
gnprice: The algorithm in the standard is the `yes_only = false` (or `want_definite = true`) case, so I think things are simplest to describe if that's the baseline and the other case is described by contrast with it. Could certainly handle that with some added words, though.

Another aspect is that there is an asymmetry between a definite YES and a definite NO. In the `yes_only` / `!want_definite` case, what the caller really *most* wants is a definite YES... it's just a definite NO that is no more helpful than a MAYBE. Or another angle on that: if the caller really didn't care about getting a definite answer, it could just assume MAYBE and not bother calling the helper :wink: It's in the hope of getting a YES that it makes the call.

More ideas just now: perhaps `maybe_is_no`, to mean "a MAYBE is no worse than a NO"?

Or `yes_or_maybe`, for "there are only two distinct answers, YES and MAYBE". That's basically what I was going for with `yes_only`, but probably a less paradoxical way of putting it :slightly_smiling_face:

vstinner: The API also surprised me. I suggest tho rename is_normalized_quickcheck() to is_normalized_impl() and add 2 functions: is_normalized() (yes_only=false) and is_normalized_quickcheck()  (yes_only=true).

gnprice: > I suggest to rename is_normalized_quickcheck() to is_normalized_impl() and add 2 functions: is_normalized() (yes_only=false) and is_normalized_quickcheck() (yes_only=true).

Hmm, I like the idea of covering over some of the multiplication of flags with wrapper functions!

I don't like quite these names, though. First, I don't think `is_normalized` is a good name, for either value of the flag, because the function doesn't actually take full responsibility for finding out if the string is normalized -- it can return MAYBE, and then it's up to the caller to go and normalize the string and compare.

What the function does do is implement a quick but partial check to try to see if the string is normalized. Specifically it implements one that's defined in the standard, which the standard calls `quickCheck` in its sample code, and which uses a property in the UCD called `Quick_Check`.

More precisely, the yes_only=false case implements that standard quick-check algorithm; the yes_only=true (or yes_or_maybe=true, etc.) case (which is the version in master) is a variation on it. I think the name "quickcheck" or `is_normalized_quickcheck`, with no further qualifications, sounds like it will mean the standard's algorithm and it's confusing if it means a different algorithm instead.

Here's another possible set of names to use with the same idea:
* rename this function to `is_normalized_quickcheck_impl`
* `is_normalized_quickcheck` is the standard's quickcheck (yes_only/yes_or_maybe=false)
* `is_normalized_quickercheck` is the variant that's more impatient to `return` and hand responsibility back to the caller (yes_or_maybe=true)
  * Or `is_normalized_quickcheck_quicker`

Or if those names feel too long: `is_normalized_qc_impl`, `is_normalized_qc`, `is_normalized_qc_quicker`.

How does that sound?

gnprice: > Or `yes_or_maybe`, for "there are only two distinct answers, YES and MAYBE".

Another idea for the flag name (though its name becomes less important if we use Victor's idea of a pair of wrapper functions): `yes_vs_maybe`.

benjaminp: It doesn't seem like anything is standing out as dramatically better. At least the semantics are clearly documented and easily read from the function.

So, let's go with the current state.

Modules/unicodedata.c
benjaminp: Okay, we can fix up the rest in a sequel.

benjaminp (approved):

gnprice:
Thanks for all the reviews and the merge!

@benjaminp , I'll go on and send next some of those followups you suggested.

gnprice:
> @benjaminp , I'll go on and send next some of those followups you suggested.

Filed [bpo-38043](https://bugs.python.org/issue38043) for that, and sent PRs for two of them.

PR #15559, opened 2019-08-28
bpo-37806:Add check recursion in typeing.ForwardRef when get hash.

hongweipeng:
https://bugs.python.org/issue37806

### DIFF
--- a/Lib/typing.py
+++ b/Lib/typing.py
@@ -522,14 +522,13 @@ class ForwardRef(_Final, _root=True):
         return self.__forward_value__
 
     def _check_recursion(self, value, forward_args):
-        if isinstance(value, _GenericAlias):
-            for val in value.__args__:
-                if isinstance(val, _GenericAlias):
-                    self._check_recursion(val, forward_args)
-                elif self is val:
-                    forward_args.append(...)
-                else:
-                    forward_args.append(val)
+        for val in value.__args__:
+            if isinstance(val, _GenericAlias):
+                self._check_recursion(val, forward_args)
+            elif self is val:
+                forward_args.append(...)
+            else:
+                forward_args.append(val)
         return tuple(forward_args)
 
     def __eq__(self, other):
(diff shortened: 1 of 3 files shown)


Misc/NEWS.d/next/Library/2019-08-28-19-00-50.bpo-37806.HdrlIB.rst
```diff
@@ -0,0 +1 @@
+Fix infinite recursion from :func:`typing.get_type_hints`.Patch by hongweipeng.
```
ilevkivskyi:
```suggestion
Fix infinite recursion from :func:`typing.get_type_hints`.  Patch by hongweipeng.
```

Lib/typing.py
```diff
@@ -521,13 +521,26 @@ def _evaluate(self, globalns, localns):
             self.__forward_evaluated__ = True
         return self.__forward_value__
 
+    def _check_recursion(self, value, forward_args):
+        if isinstance(value, _GenericAlias):
+            for val in value.__args__:
+                if isinstance(val, _GenericAlias):
+                    self._check_recursion(val, forward_args)
+                elif self is val and self.__forward_evaluated__:
+                    forward_args.append(...)
```
ilevkivskyi: Is it possible to create an actually recursive object? Or is it too dangerous?

ilevkivskyi (commented):
Thanks! I have few comments. Also I think this requires some tests.

Lib/typing.py
hongweipeng: What does an actually recursive object mean? I added some test code in new commit. Can you have a look and see what other tests need to be added. Thanks.

hongweipeng:
I have some concerns, replacing the `ForwardRef` object with `...`,  may cause `get_type_hints(*) == get_type_hints(*)` return `False` .

Lib/typing.py
hongweipeng:
```
a = List['int']
for i in range(sys.getrecursionlimit() + 100):
    a = List[a]
```
It raises a `RecursionError` both in this PR and master branch. Is an actually recursive object mean this?

ilevkivskyi: No. Consider the initial example:
```python
ValueList = List['Value']
Value = Union[str, ValueList]
```
In this example I mean trying to construct an object `Value` such that `Value.__args__[1].__args__[0] is Value`. I understand this may be tricky, but I don't think it is impossible.

Also it indeed looks like this may be "dangerous" in the sense that we need to make `__repr__()`, `__eq__()`, and `__hash__()` aware of the possible recursion.

ilevkivskyi: On the other hand:
```
>>> x = [1]
>>> x.append(x)
>>> x
[1, [...]]
>>> y = [1]
>>> y.append(y)
>>> x == y
RecursionError: maximum recursion depth exceeded in comparison
```
So maybe it would be enough to just update the `__repr__()`.

hongweipeng:
PR #15650 can solve this issue.

PR #15565, opened 2019-08-28
bpo-37972: unittest.mock._Call now passes on __getitem__ to the __getattr__ chaining so that call() can be subscriptable

blhsing:
As reported on StackOverflow:
https://stackoverflow.com/questions/57636747/how-to-perform-assert-has-calls-for-a-getitem-call

The following code:

```
from unittest.mock import MagicMock, call
mm = MagicMock()
mm().foo()['bar']
print(mm.mock_calls)
```
would output:

`[call(), call().foo(), call().foo().__getitem__('bar')]`

But trying to use that list with:

`mm.assert_has_calls([call(), call().foo(), call().foo().__getitem__('bar')])`

would result in:

`TypeError: tuple indices must be integers or slices, not str`

This fix now makes `unittest.mock._Call.__getattribute__` raise an `AttributeError` when the given attribute name is `'__getitem__'` so that `__getitem__` can get the same chaining mechanism provided by the `__getattr__` method.

https://bugs.python.org/issue37972

### DIFF
--- a/Lib/unittest/mock.py
+++ b/Lib/unittest/mock.py
@@ -2428,6 +2428,12 @@ class _Call(tuple):
         return _Call(name=name, parent=self, from_kall=False)
 
 
+    def __getattribute__(self, attr):
+        if attr == '__getitem__':
+            raise AttributeError
+        return tuple.__getattribute__(self, attr)
+
+
     def count(self, /, *args, **kwargs):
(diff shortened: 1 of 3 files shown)


the-knights-who-say-ni:

voidspace (approved):
LGTM. Nice catch and fix.

PR #15592, opened 2019-08-29
bpo-37976: Prevent shadowing of TypeError in zip()

sir-sigurd:
https://bugs.python.org/issue37976


Python/bltinmodule.c
```diff
@@ -2549,8 +2549,8 @@ zip_new(PyTypeObject *type, PyObject *args, PyObject *kwds)
         PyObject *it = PyObject_GetIter(item);
         if (it == NULL) {
             if (PyErr_ExceptionMatches(PyExc_TypeError))
```
sir-sigurd: Might make sense to do the same for all exceptions types, but I'm not sure if it's OK for compatibility.

rhettinger (approved):

PR #15593, opened 2019-08-29
bpo-37947: Avoid double-decrement in symtable recursion counting

ncoghlan:
With `symtable_visit_expr` now correctly adjusting the recursion depth for named
expressions, `symtable_handle_namedexpr` should be leaving it alone.

https://bugs.python.org/issue37947

### DIFF
--- a/Python/symtable.c
+++ b/Python/symtable.c
@@ -1501,16 +1501,16 @@ symtable_handle_namedexpr(struct symtable *st, expr_ty e)
         PyErr_SyntaxLocationObject(st->st_filename,
                                     e->lineno,
                                     e->col_offset);
-        VISIT_QUIT(st, 0);
+        return 0;
     }
     if (st->st_cur->ste_comprehension) {
         /* Inside a comprehension body, so find the right target scope */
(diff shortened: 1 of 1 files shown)


pablogsal (approved):
LGTM

Thank you for correcting this!

ncoghlan:
I think I've found a good spot to add a sanity check that will make the test cases fail if a PR gets this accounting wrong - building that now (and I'll test it out by partially reverting the previous change).

ncoghlan:
Neat:
```
ERROR: test_named_expression_variable_reuse_in_comprehensions (test.test_named_expressions.NamedExpressionScopeTest) (case='Nested nonlocal')
  File "/home/ncoghlan/devel/cpython/Lib/test/test_named_expressions.py", line 469, in test_named_expression_variable_reuse_in_comprehensions
    exec(code, ns)
SystemError: symtable analysis recursion depth mismatch (before=96, after=95)
```

(This is with `VISIT_QUIT(st, 1)` at the end of `symtable_handle_namedexpr`)

PR #15595, opened 2019-08-29
bpo-37977: Warn more strongly and clearly about pickle security

lordmauve:
Rewrite the red warning at the top of the `pickle` module documentation.

* Simpler, more direct English.
* Explain the severity of vulnerability that doing this will cause.
* Link to the hmac module which can be used to prevent tampering.
* Link to the json module which is safer if less powerful.

https://bugs.python.org/issue37977

### DIFF
--- a/Doc/library/pickle.rst
+++ b/Doc/library/pickle.rst
@@ -30,9 +30,17 @@ avoid confusion, the terms used here are "pickling" and "unpickling".
 
 .. warning::
 
-   The :mod:`pickle` module is not secure against erroneous or maliciously
-   constructed data.  Never unpickle data received from an untrusted or
-   unauthenticated source.
+   The ``pickle`` module **is not secure**. Only unpickle data you trust.
+
+   It is possible to construct malicious pickle data which will **execute
+   arbitrary code during unpickling**. Never unpickle data that could have come
+   from an untrusted source, or that could have been tampered with.
+
+   Consider signing data with :mod:`hmac` if you need to ensure that it has not
+   been tampered with.
+
+   Safer serialization formats such as :mod:`json` may be more appropriate if
+   you are processing untrusted data. See :ref:`comparison-with-json`.
 
 
 Relationship to other Python modules
@@ -75,6 +83,9 @@ The :mod:`pickle` module differs from :mod:`marshal` in several significant ways
   pickling and unpickling code deals with Python 2 to Python 3 type differences
   if your data is crossing that unique breaking change language boundary.
 
+
+.. _comparison-with-json:
+
 Comparison with ``json``
 ^^^^^^^^^^^^^^^^^^^^^^^^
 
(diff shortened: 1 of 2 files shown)


pitrou:
There is already a section titled "Comparison with json", so perhaps move the sentence stating json is secure there?

lordmauve:
Hmm, good point. I don't think we should move it there because it wouldn't form part of the security warning, but we should link to that section - and that section should state again that JSON doesn't have the arbitrary code execution issue.

Doc/library/pickle.rst
```diff
@@ -30,9 +30,19 @@ avoid confusion, the terms used here are "pickling" and "unpickling".
 
 .. warning::
 
-   The :mod:`pickle` module is not secure against erroneous or maliciously
-   constructed data.  Never unpickle data received from an untrusted or
-   unauthenticated source.
+   The ``pickle`` module **is not secure**. Only unpickle data you trust.
+
+   It is possible to construct malicious pickle data which will **execute
+   arbitrary code during unpickling**. Never unpickle data that could have come
```
disconnect3d, line 36: Maybe:
```suggestion
   arbitrary code during pickle.load[s]**. Never unpickle data that could have come
```

rhettinger: I prefer the "unpickling version"

Doc/library/pickle.rst
```diff
@@ -30,9 +30,19 @@ avoid confusion, the terms used here are "pickling" and "unpickling".
...
+   from an untrusted source, or that could have been tampered with.
+
+   Consider signing data with :mod:`hmac` if you need to ensure that it has not
+   been tampered with.
+
+   The JSON data interchange format provided by the :mod:`json` module is more
```
rhettinger: I recommend leaving this out.  It goes well past giving an informative warning and moves into editorializing.  Also, for many uses of pickle, JSON is no substitute.

Doc/library/pickle.rst
moreati: To me, using "pickle.load[s]" here creates the false impression that it's just the module level convenience functions which are insecure. It incorrectly implies (by omission) that "p = pickle.Unpickler(..); p.load()" would be safe.

lordmauve: I agree. "unpickling" is defined immediately before this warning, in the previous paragraph.

Doc/library/pickle.rst
lordmauve: I have changed this sentence to be less discursive and simply point to the `json` module and the "Comparison with `json`" section lower down.

Doc/library/pickle.rst
disconnect3d: My initial thought was about making it _more_ explicit but since unpickling is already defined and `pickle.load[s]` doesn't cover all the cases it totally makes sense to reject this suggestion :).

Thanks for the great points!

rhettinger (approved):

lordmauve:
Thanks, everyone!

PR #15596, opened 2019-08-29
bpo-37979: Add alternative to fromisoformat in documentation

pganssle:
Adds a link to `dateutil.parser.isoparse` in the documentation.

It would be nice to set up intersphinx for things like this, but I think we can leave that for a separate PR.

CC: @pitrou 

[bpo-37979](https://bugs.python.org/issue37979)

https://bugs.python.org/issue37979


pitrou (approved):
LGTM. Thank you @pganssle :-)

PR #15602, opened 2019-08-29
bpo-37834: Prevent shutil.rmtree exception

ned-deily:
when built on non-Windows system without fd system call support,
like older versions of macOS.

https://bugs.python.org/issue37834


ambv:
Looks fine. Tests pass, manually tested on Windows, too.

PR #15612, opened 2019-08-30
bpo-37140: Fix StructUnionType_paramfunc()

vstinner:
Fix a ctypes regression of Python 3.8. When a ctypes.Structure is
passed by copy to a function, ctypes internals created a temporary
object which had the side effect of calling the structure finalizer
(__del__) twice. The Python semantics requires a finalizer to be
called exactly once. Fix ctypes internals to no longer call the
finalizer twice.

Create a new internal StructParam_Type which is only used by
_ctypes_callproc() to call PyMem_Free(ptr) on Py_DECREF(argument).
StructUnionType_paramfunc() creates such object.

https://bugs.python.org/issue37140

### DIFF
--- a/Lib/ctypes/test/test_structures.py
+++ b/Lib/ctypes/test/test_structures.py
@@ -3,7 +3,7 @@ from ctypes import *
 from ctypes.test import need_symbol
 from struct import calcsize
 import _ctypes_test
-import test.support
+from test import support
 
 class SubclassesTest(unittest.TestCase):
     def test_subclass(self):
@@ -202,7 +202,7 @@ class StructureTestCase(unittest.TestCase):
              "_pack_": -1}
         self.assertRaises(ValueError, type(Structure), "X", (Structure,), d)
 
-    @test.support.cpython_only
+    @support.cpython_only
     def test_packed_c_limits(self):
         # Issue 15989
         import _testcapi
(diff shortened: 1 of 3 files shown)


vstinner:
I wrote an unit test. I made sure that the unit test fails without the fix:

```
FAIL: test_pass_by_value_finalizer (ctypes.test.test_structures.StructureTestCase)
  File "/home/vstinner/python/master/Lib/ctypes/test/test_structures.py", line 449, in test_pass_by_value_finalizer
    self.assertEqual(finalizer_calls, [])
AssertionError: Lists differ: ['called'] != []

First list contains 1 additional elements.
First extra element 0:
'called'

- ['called']
+ []
```

Without the fix, Test.__del__() is called once internally during the call to the C function ("func()"), and then a second time when the object is destroyed (s = None).

Modules/_ctypes/_ctypes.c
```diff
@@ -403,35 +432,46 @@ static PyCArgObject *
...
     if ((size_t)self->b_size > sizeof(void*)) {
         void *new_ptr = PyMem_Malloc(self->b_size);
         if (new_ptr == NULL)
             return NULL;
         memcpy(new_ptr, self->b_ptr, self->b_size);
-        copied_self = (CDataObject *)PyCData_AtAddress(
-            (PyObject *)Py_TYPE(self), new_ptr);
-        copied_self->b_needsfree = 1;
+
+        obj = (&StructParam_Type)->tp_alloc(&StructParam_Type, 0);
```
serge-sans-paille, line 449: s/obj/arg_obj/ to increase readability ?

Modules/_ctypes/_ctypes.c
```diff
@@ -403,35 +432,46 @@ static PyCArgObject *
...
-    stgdict = PyObject_stgdict((PyObject *)copied_self);
+    stgdict = PyObject_stgdict((PyObject *)self);
     assert(stgdict); /* Cannot be NULL for structure/union instances */
     parg->pffi_type = &stgdict->ffi_type_pointer;
-    parg->value.p = copied_self->b_ptr;
-    parg->size = copied_self->b_size;
-    parg->obj = (PyObject *)copied_self;
+    parg->value.p = ptr;
+    parg->size = self->b_size;
+    parg->obj = obj;
```
serge-sans-paille, line 475: We spent some time to find that information, maybe add a comment here

```
parg->obj = obj; // obj will be deleted when parg is deleted
```

or something like that?

Lib/ctypes/test/test_structures.py
```diff
@@ -396,27 +396,66 @@ class Z(Y):
...
+        self.assertEqual(finalizer_calls, [])
+        self.assertEqual(s.first, 1)
+        self.assertEqual(s.second, 2)
+        self.assertEqual(s.third, 3)
+
+        # The finalizer must be called exactly once
+        s = None
+        support.gc_collect()
+        self.assertEqual(finalizer_calls, ["called"])
+
```
serge-sans-paille, line 458: <3 the test case

Modules/_ctypes/_ctypes.c
vstinner: I chose "obj" name to get "parg->obj = obj" which looks nice to me :-)

Modules/_ctypes/_ctypes.c
vstinner: done

PR #15623, opened 2019-08-30
bpo-37705: Improve the implementation of winerror_to_errno()

ZackerySpytz:
winerror_to_errno() is no longer automatically generated.
Do not rely on the old _dosmapperr() function.
Add ERROR_NO_UNICODE_TRANSLATION (1113) -> EILSEQ.

Many thanks go to Eryk Sun for their posts on the BPO issue.

https://bugs.python.org/issue37705


aeros (commented):
Thanks for the PR @ZackerySpytz.

I am very much in favor of using constants to give the winerrors more semantic meaning and grouping the returns. However, there's several of the constant names which don't describe their purpose in a meaningful way. I'm not familiar with our stance on linking to external documentation within code comments, but it seems like it might be helpful to provide a link to https://docs.microsoft.com/en-us/windows/win32/winsock/windows-sockets-error-codes-2.

eryksun:
Stever Downer (@zooba) mentioned adding a dict in Modules/errnomodule.c that exposes this mapping to Python code. I guess it could be called `winerror2errno`. Maybe we could also add a `winerrorcode` dict with the symbolic names of mapped errors, and errors of interest. For the latter, I think it would be better if `winerror_to_errno` were based on a counted or null-terminated table instead of a switch.  Each entry would have the symbolic name as a string (e.g. `"ERROR_FILE_NOT_FOUND"`), the error code, and the mapped errno value. This avoids needing to maintain supported error codes in two places. Otherwise if we map a new error in PC/errmap.h, we'll have to separately add the symbolic name in Modules/errnomodule.c.  

For Winsock errors, I think the `winerror2errno` and `winerrorcode` dicts would only need the six that are specially mapped to CRT errno values, since all others are already available as errno values -- both as module constants and in the `errorcode` dict. Given that, should these six Windows errors remain defined as errno values? For example, now that `WSAEINTR` maps to `EINTR`, does it make sense to have `errno.WSAEINTR == 10004` and `errorcode[10004] == 'WSAEINTR'`? In any case, I don't think it's a problem to leave them there.

zooba:
I'm merging and backporting this, as the additional error codes are general goodness that will improve all active versions.

Adding a new public API should be only for master.

PR #15628, opened 2019-08-31
bpo-37953:Improve ForwardRef equality checks

hongweipeng:

### DIFF
--- a/Lib/typing.py
+++ b/Lib/typing.py
@@ -524,8 +524,10 @@ class ForwardRef(_Final, _root=True):
     def __eq__(self, other):
         if not isinstance(other, ForwardRef):
             return NotImplemented
-        return (self.__forward_arg__ == other.__forward_arg__ and
-                self.__forward_value__ == other.__forward_value__)
+        if self.__forward_evaluated__ and other.__forward_evaluated__:
+            return (self.__forward_arg__ == other.__forward_arg__ and
+                    self.__forward_value__ == other.__forward_value__)
+        return self.__forward_arg__ == other.__forward_arg__
 
     def __hash__(self):
         return hash((self.__forward_arg__, self.__forward_value__))
--- a/Lib/typing.py
+++ b/Lib/typing.py
@@ -230,12 +230,7 @@ def _remove_dups_flatten(parameters):
         else:
             params.append(p)
     # Weed out strict duplicates, preserving the first of each occurrence.
-    new_params = []
-    for p in params:
-        if p not in new_params:
-            new_params.append(p)
-    params = new_params
-    return tuple(params)
+    return tuple(dict.fromkeys(params))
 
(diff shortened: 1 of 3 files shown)


Lib/typing.py
```diff
@@ -230,15 +230,11 @@ def _remove_dups_flatten(parameters):
...
-                new_params.append(t)
-                all_params.remove(t)
-        params = new_params
-        assert not all_params, all_params
+    new_params = []
+    for p in params:
+        if p not in new_params:
+            new_params.append(p)
+    params = new_params
     return tuple(params)
```
brandtbucher: This whole bit can be sped up by just replacing it with:

```py
return tuple(dict.fromkeys(params))
```

brandtbucher (changes requested):
This looks like a good change. I found one optimization that can be made, below.

Can you also add regression tests as well?

Lib/typing.py
hongweipeng: Thanks, I will add some tests. By the way, `params` is a list, `tuple(dict.fromkeys(params))` seems not faster than `tuple(params)`, I think it would be better to keep `tuple(params)`:

brandtbucher: Sorry, I should have been clearer! I meant replacing this whole list + loop bit with that one line. `dict.fromkeys` is a nice shortcut to remove duplicates while preserving order.

brandtbucher: Ah... I see now. You're trying to avoid the (broken) hashing.

I don't know how I feel about that. I think hashing should be fixed rather than just sidestepped here, since it's so tied to equality. At this point I'm actually in favor of a hybrid approach: combining your fix for `__eq__` and the other PR #15400’s fix for `__hash__`.

@gvanrossum and @ilevkivskyi, thoughts?

brandtbucher: @hongweipeng you can revert your most recent commit (3676a50, “code review”). My comment was misguided!

Lib/typing.py
```diff
@@ -230,16 +230,7 @@ def _remove_dups_flatten(parameters):
...
-    if len(all_params) < len(params):
-        new_params = []
-        for t in params:
-            if t in all_params:
-                new_params.append(t)
-                all_params.remove(t)
-        params = new_params
-        assert not all_params, all_params
-    return tuple(params)
+    return tuple(dict.fromkeys(params))
```
ilevkivskyi, line 233: TBH this looks like a hack. If at some point we would make sets ordered the same way as dicts are, it would be good to update the code, but currently I think I like the old way more.

ilevkivskyi:
See also my comment https://github.com/python/cpython/pull/15400#issuecomment-527081494 in the other PR.

hongweipeng:
PR #15650 is better.

PR #15633, opened 2019-09-01
bpo-24416: Return a IsoCalendarDate from date.isocalendar()

corona10:
https://bugs.python.org/issue24416

### DIFF
--- a/Doc/library/datetime.rst
+++ b/Doc/library/datetime.rst
@@ -670,7 +670,8 @@ Instance methods:
 
 .. method:: date.isocalendar()
 
-   Return a :class:`IsoCalendarDate`, (ISO year, ISO week number, ISO weekday).
+   Return a :term:`named tuple` object with three components: `year`, `week`
+   and `weekday`.
 
    The ISO calendar is a widely used variant of the Gregorian calendar. [#]_
 
@@ -689,8 +690,7 @@ Instance methods:
        (2004, 1, 7)
 
    .. versionchanged:: 3.9
-      Result changed from a tuple to a :class:`IsoCalendarDate`.
-      It can't be unpickled in previous versions.
+      Result changed from a tuple to a :term:`named tuple`.
 
 .. method:: date.isoformat()
 
@@ -1403,7 +1403,8 @@ Instance methods:
 
 .. method:: datetime.isocalendar()
 
-   Return a :class:`IsoCalendarDate`, (ISO year, ISO week number, ISO weekday).
+   Return a :term:`named tuple` object with three components: `year`, `week`
+   and `weekday`.
    The same as ``self.date().isocalendar()``.
 
 
--- a/Doc/library/datetime.rst
+++ b/Doc/library/datetime.rst
@@ -683,11 +683,11 @@ Instance methods:
    For example, 2004 begins on a Thursday, so the first week of ISO year 2004
    begins on Monday, 29 Dec 2003 and ends on Sunday, 4 Jan 2004::
 
-       >>> from datetime import date
-       >>> date(2003, 12, 29).isocalendar()
-       (2004, 1, 1)
-       >>> date(2004, 1, 4).isocalendar()
-       (2004, 1, 7)
+        >>> from datetime import date
+        >>> date(2003, 12, 29).isocalendar()
+        datetime.IsoCalendarDate(year=2004, week=1, weekday=1)
+        >>> date(2004, 1, 4).isocalendar()
+        datetime.IsoCalendarDate(year=2004, week=1, weekday=7)
 
    .. versionchanged:: 3.9
       Result changed from a tuple to a :term:`named tuple`.
--- a/Doc/library/datetime.rst
+++ b/Doc/library/datetime.rst
@@ -670,7 +670,7 @@ Instance methods:
 
 .. method:: date.isocalendar()
 
-   Return a named tuple, (ISO year, ISO week number, ISO weekday).
+   Return a :class:`IsoCalendarDate`, (ISO year, ISO week number, ISO weekday).
 
    The ISO calendar is a widely used variant of the Gregorian calendar. [#]_
 
@@ -689,7 +689,7 @@ Instance methods:
        (2004, 1, 7)
 
    .. versionchanged:: 3.9
-      Result changed from a tuple to a namedtuple.
+      Result changed from a tuple to a :class:`IsoCalendarDate`.
       It can't be unpickled in previous versions.
 
 .. method:: date.isoformat()
@@ -1403,7 +1403,7 @@ Instance methods:
 
 .. method:: datetime.isocalendar()
 
-   Return a named tuple, (ISO year, ISO week number, ISO weekday).
+   Return a :class:`IsoCalendarDate`, (ISO year, ISO week number, ISO weekday).
    The same as ``self.date().isocalendar()``.
 
 
--- a/Doc/library/datetime.rst
+++ b/Doc/library/datetime.rst
@@ -670,8 +670,8 @@ Instance methods:
 
 .. method:: date.isocalendar()
 
-   Return a :term:`named tuple` object with three components: `year`, `week`
-   and `weekday`.
+   Return a :term:`named tuple` object with three components: ``year``,
+   ``week`` and ``weekday``.
 
    The ISO calendar is a widely used variant of the Gregorian calendar. [#]_
 
@@ -1403,9 +1403,8 @@ Instance methods:
 
 .. method:: datetime.isocalendar()
 
-   Return a :term:`named tuple` object with three components: `year`, `week`
-   and `weekday`.
-   The same as ``self.date().isocalendar()``.
+   Return a :term:`named tuple` with three components: ``year``, ``week``
+   and ``weekday``. The same as ``self.date().isocalendar()``.
 
 
 .. method:: datetime.isoformat(sep='T', timespec='auto')
--- a/Doc/library/datetime.rst
+++ b/Doc/library/datetime.rst
@@ -670,7 +670,7 @@ Instance methods:
 
 .. method:: date.isocalendar()
 
-   Return a 3-tuple, (ISO year, ISO week number, ISO weekday).
+   Return a 3-namedtuple, (ISO year, ISO week number, ISO weekday).
 
    The ISO calendar is a widely used variant of the Gregorian calendar. [#]_
 
@@ -687,6 +687,10 @@ Instance methods:
        (2004, 1, 1)
        >>> date(2004, 1, 4).isocalendar()
        (2004, 1, 7)
+
+   .. versionchanged:: 3.9
+      Result changed from a tuple to a namedtuple.
+      It can't be unpickled in previous versions.
 
 .. method:: date.isoformat()
 
@@ -1399,8 +1403,8 @@ Instance methods:
 
 .. method:: datetime.isocalendar()
 
-   Return a 3-tuple, (ISO year, ISO week number, ISO weekday). The same as
-   ``self.date().isocalendar()``.
+   Return a 3-namedtuple, (ISO year, ISO week number, ISO weekday).
+   The same as ``self.date().isocalendar()``.
 
 
 .. method:: datetime.isoformat(sep='T', timespec='auto')
--- a/Doc/library/datetime.rst
+++ b/Doc/library/datetime.rst
@@ -670,7 +670,7 @@ Instance methods:
 
 .. method:: date.isocalendar()
 
-   Return a 3-namedtuple, (ISO year, ISO week number, ISO weekday).
+   Return a named tuple, (ISO year, ISO week number, ISO weekday).
 
    The ISO calendar is a widely used variant of the Gregorian calendar. [#]_
 
@@ -1403,7 +1403,7 @@ Instance methods:
 
 .. method:: datetime.isocalendar()
 
-   Return a 3-namedtuple, (ISO year, ISO week number, ISO weekday).
+   Return a named tuple, (ISO year, ISO week number, ISO weekday).
    The same as ``self.date().isocalendar()``.
 
 
--- a/Doc/whatsnew/3.9.rst
+++ b/Doc/whatsnew/3.9.rst
@@ -150,7 +150,6 @@ datetime
 The :meth:`~datetime.date.isocalendar()` of :class:`datetime.date`
 and :meth:`~datetime.datetime.isocalendar()` of :class:`datetime.datetime`
 methods now returns a :func:`~collections.namedtuple` instead of a :class:`tuple`.
-It can't be unpickled in previous versions.
 (Contributed by Dong-hee Na in :issue:`24416`.)
 
 fcntl
--- a/Doc/whatsnew/3.9.rst
+++ b/Doc/whatsnew/3.9.rst
@@ -145,6 +145,14 @@ Add :func:`curses.get_escdelay`, :func:`curses.set_escdelay`,
 :func:`curses.get_tabsize`, and :func:`curses.set_tabsize` functions.
 (Contributed by Anthony Sottile in :issue:`38312`.)
 
+datetime
+--------
+The :meth:`~datetime.date.isocalendar()` of :class:`datetime.date`
+and :meth:`~datetime.datetime.isocalendar()` of :class:`datetime.datetime`
+methods now returns a :func:`~collections.namedtuple` instead of a :class:`tuple`.
+It can't be unpickled in previous versions.
+(Contributed by Dong-hee Na in :issue:`24416`.)
+
 fcntl
 -----
 
--- a/Lib/datetime.py
+++ b/Lib/datetime.py
@@ -1120,7 +1120,7 @@ class date:
             if today >= _isoweek1monday(year+1):
                 year += 1
                 week = 0
-        return IsoCalendarDate((year, week+1, day+1))
+        return _IsoCalendarDate((year, week+1, day+1))
 
     # Pickle support.
 
@@ -1231,11 +1231,16 @@ class IsoCalendarDate(tuple):
     def weekday(self):
         return self[2]
 
+    def __reduce__(self):
+        return (tuple, (tuple(self),))
+
     def __repr__(self):
         return (f'{self.__class__.__name__}'
             f'(year={self[0]}, week={self[1]}, weekday={self[2]})')
 
 
+_IsoCalendarDate = IsoCalendarDate
+del IsoCalendarDate
 _tzinfo_class = tzinfo
 
 class time:
@@ -2541,7 +2546,7 @@ else:
          _format_time, _format_offset, _is_leap, _isoweek1monday, _math,
          _ord2ymd, _time, _time_class, _tzinfo_class, _wrap_strftime, _ymd2ord,
          _divide_and_round, _parse_isoformat_date, _parse_isoformat_time,
-         _parse_hh_mm_ss_ff)
+         _parse_hh_mm_ss_ff, _IsoCalendarDate)
     # XXX Since import * above excludes names that start with _,
     # docstring does not get overwritten. In the future, it may be
     # appropriate to maintain a single module level docstring and
--- a/Lib/datetime.py
+++ b/Lib/datetime.py
@@ -1120,7 +1120,7 @@ class date:
             if today >= _isoweek1monday(year+1):
                 year += 1
                 week = 0
-        return _IsoCalendarDate((year, week+1, day+1))
+        return IsoCalendarDate((year, week+1, day+1))
 
     # Pickle support.
 
@@ -1237,9 +1237,6 @@ class IsoCalendarDate(tuple):
     def __repr__(self):
         return (f'{self.__class__.__name__}'
                 f'(year={self[0]}, week={self[1]}, weekday={self[2]})')
-
-_IsoCalendarDate = IsoCalendarDate
-del IsoCalendarDate
 
 
 _tzinfo_class = tzinfo
@@ -2547,7 +2544,7 @@ else:
          _format_time, _format_offset, _is_leap, _isoweek1monday, _math,
          _ord2ymd, _time, _time_class, _tzinfo_class, _wrap_strftime, _ymd2ord,
          _divide_and_round, _parse_isoformat_date, _parse_isoformat_time,
-         _parse_hh_mm_ss_ff, _IsoCalendarDate)
+         _parse_hh_mm_ss_ff)
     # XXX Since import * above excludes names that start with _,
     # docstring does not get overwritten. In the future, it may be
     # appropriate to maintain a single module level docstring and
--- a/Lib/datetime.py
+++ b/Lib/datetime.py
@@ -1096,7 +1096,7 @@ class date:
         return self.toordinal() % 7 or 7
 
     def isocalendar(self):
-        """Return a IsoCalendarDate containing ISO year, week number, and weekday.
+        """Return a named tuple containing ISO year, week number, and weekday.
 
         The first ISO week of the year is the (Mon-Sun) week
         containing the year's first Thursday; everything else derives
@@ -1121,7 +1121,7 @@ class date:
             if today >= _isoweek1monday(year+1):
                 year += 1
                 week = 0
-        return _IsoCalendarDate((year, week+1, day+1))
+        return IsoCalendarDate((year, week+1, day+1))
 
     # Pickle support.
 
@@ -1212,7 +1212,7 @@ class tzinfo:
             return (self.__class__, args, state)
 
 
-class _IsoCalendarDate(tuple):
+class IsoCalendarDate(tuple):
 
     def __new__(cls, seq):
         if len(seq) != 3:
@@ -1233,12 +1233,8 @@ class _IsoCalendarDate(tuple):
         return self[2]
 
     def __repr__(self):
-        return ('datetime.IsoCalendarDate'
+        return (f'{self.__class__.__name__}'
             f'(year={self[0]}, week={self[1]}, weekday={self[2]})')
-
-
-_IsoCalendarDate.__name__ = "IsoCalendarDate"
-_IsoCalendarDate.__qualname__ = "IsoCalendarDate"
 
 
 _tzinfo_class = tzinfo
@@ -2546,7 +2542,7 @@ else:
          _format_time, _format_offset, _is_leap, _isoweek1monday, _math,
          _ord2ymd, _time, _time_class, _tzinfo_class, _wrap_strftime, _ymd2ord,
          _divide_and_round, _parse_isoformat_date, _parse_isoformat_time,
-         _parse_hh_mm_ss_ff, _IsoCalendarDate)
+         _parse_hh_mm_ss_ff)
     # XXX Since import * above excludes names that start with _,
     # docstring does not get overwritten. In the future, it may be
     # appropriate to maintain a single module level docstring and
--- a/Lib/datetime.py
+++ b/Lib/datetime.py
@@ -1120,7 +1120,7 @@ class date:
         return self.toordinal() % 7 or 7
 
     def isocalendar(self):
-        """Return a named tuple containing ISO year, week number, and weekday.
+        """Return a IsoCalendarDate containing ISO year, week number, and weekday.
 
         The first ISO week of the year is the (Mon-Sun) week
         containing the year's first Thursday; everything else derives
--- a/Lib/datetime.py
+++ b/Lib/datetime.py
@@ -12,7 +12,6 @@ import time as _time
 import math as _math
 import sys
 
-
 def _cmp(x, y):
     return 0 if x == y else 1 if x > y else -1
 
--- a/Lib/datetime.py
+++ b/Lib/datetime.py
@@ -1229,11 +1229,13 @@ class IsoCalendarDate(tuple):
         return self[2]
 
     def __reduce__(self):
+        # This code is intended to pickle the object without making the
+        # class public. See https://bugs.python.org/msg352381
         return (tuple, (tuple(self),))
 
     def __repr__(self):
         return (f'{self.__class__.__name__}'
-            f'(year={self[0]}, week={self[1]}, weekday={self[2]})')
+                f'(year={self[0]}, week={self[1]}, weekday={self[2]})')
 
 
 _IsoCalendarDate = IsoCalendarDate
--- a/Lib/datetime.py
+++ b/Lib/datetime.py
@@ -15,6 +15,9 @@ import sys
 from collections import namedtuple
 
 
+IsoCalendarDate = namedtuple('IsoCalendarDate', 'year week weekday')
+
+
 def _cmp(x, y):
     return 0 if x == y else 1 if x > y else -1
 
@@ -1098,7 +1101,7 @@ class date:
         return self.toordinal() % 7 or 7
 
     def isocalendar(self):
-        """Return a 3-named tuple containing ISO year, week number, and weekday.
+        """Return a named tuple containing ISO year, week number, and weekday.
 
         The first ISO week of the year is the (Mon-Sun) week
         containing the year's first Thursday; everything else derives
@@ -1561,8 +1564,6 @@ _time_class = time  # so functions w/ args named "time" can get at the class
 time.min = time(0, 0, 0)
 time.max = time(23, 59, 59, 999999)
 time.resolution = timedelta(microseconds=1)
-
-IsoCalendarDate = namedtuple('IsoCalendarDate', 'year week weekday')
 
 
 class datetime(date):
--- a/Lib/datetime.py
+++ b/Lib/datetime.py
@@ -1232,9 +1232,12 @@ class IsoCalendarDate(tuple):
     def weekday(self):
         return self[2]
 
+    def __reduce__(self):
+        return (tuple, (tuple(self),))
+
     def __repr__(self):
         return (f'{self.__class__.__name__}'
-            f'(year={self[0]}, week={self[1]}, weekday={self[2]})')
+                f'(year={self[0]}, week={self[1]}, weekday={self[2]})')
 
 
 _tzinfo_class = tzinfo
--- a/Lib/datetime.py
+++ b/Lib/datetime.py
@@ -1231,12 +1231,9 @@ class IsoCalendarDate(tuple):
     def weekday(self):
         return self[2]
 
-    def __reduce__(self):
-        return (tuple, (tuple(self),))
-
     def __repr__(self):
         return (f'{self.__class__.__name__}'
-                f'(year={self[0]}, week={self[1]}, weekday={self[2]})')
+            f'(year={self[0]}, week={self[1]}, weekday={self[2]})')
 
 
 _tzinfo_class = tzinfo
--- a/Lib/datetime.py
+++ b/Lib/datetime.py
@@ -12,10 +12,29 @@ import time as _time
 import math as _math
 import sys
 
-from collections import namedtuple
 
+class IsoCalendarDate(tuple):
+    def __new__(cls, seq):
+        if len(seq) != 3:
+            raise TypeError(f'{cls.__name__}() takes a 3-sequence '
+                    f'({len(seq)}-sequence given)')
+        return tuple.__new__(cls, seq)
 
-IsoCalendarDate = namedtuple('IsoCalendarDate', 'year week weekday')
+    @property
+    def year(self):
+        return self[0]
+
+    @property
+    def week(self):
+        return self[1]
+
+    @property
+    def weekday(self):
+        return self[2]
+
+    def __repr__(self):
+        return (f'{self.__class__.__name__}'
+            f'(year={self[0]}, week={self[1]}, weekday={self[2]})')
 
 
 def _cmp(x, y):
@@ -1126,7 +1145,7 @@ class date:
             if today >= _isoweek1monday(year+1):
                 year += 1
                 week = 0
-        return IsoCalendarDate(year, week+1, day+1)
+        return IsoCalendarDate((year, week+1, day+1))
 
     # Pickle support.
 
--- a/Lib/datetime.py
+++ b/Lib/datetime.py
@@ -1120,7 +1120,7 @@ class date:
             if today >= _isoweek1monday(year+1):
                 year += 1
                 week = 0
-        return _IsoCalendarDate((year, week+1, day+1))
+        return _IsoCalendarDate(year, week+1, day+1)
 
     # Pickle support.
 
@@ -1213,11 +1213,8 @@ class tzinfo:
 
 class IsoCalendarDate(tuple):
 
-    def __new__(cls, seq):
-        if len(seq) != 3:
-            raise TypeError(f'{cls.__name__}() takes a 3-sequence '
-                    f'({len(seq)}-sequence given)')
-        return super().__new__(cls, seq)
+    def __new__(cls, year, week, weekday):
+        return super().__new__(cls, (year, week, weekday))
 
     @property
     def year(self):
--- a/Lib/datetime.py
+++ b/Lib/datetime.py
@@ -13,30 +13,6 @@ import math as _math
 import sys
 
 
-class IsoCalendarDate(tuple):
-    def __new__(cls, seq):
-        if len(seq) != 3:
-            raise TypeError(f'{cls.__name__}() takes a 3-sequence '
-                    f'({len(seq)}-sequence given)')
-        return tuple.__new__(cls, seq)
-
-    @property
-    def year(self):
-        return self[0]
-
-    @property
-    def week(self):
-        return self[1]
-
-    @property
-    def weekday(self):
-        return self[2]
-
-    def __repr__(self):
-        return (f'{self.__class__.__name__}'
-            f'(year={self[0]}, week={self[1]}, weekday={self[2]})')
-
-
 def _cmp(x, y):
     return 0 if x == y else 1 if x > y else -1
 
@@ -1145,7 +1121,7 @@ class date:
             if today >= _isoweek1monday(year+1):
                 year += 1
                 week = 0
-        return IsoCalendarDate((year, week+1, day+1))
+        return _IsoCalendarDate((year, week+1, day+1))
 
     # Pickle support.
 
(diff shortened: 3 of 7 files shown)


Doc/library/datetime.rst
```diff
@@ -590,7 +590,7 @@ Instance methods:
 
 .. method:: date.isocalendar()
 
-   Return a 3-tuple, (ISO year, ISO week number, ISO weekday).
+   Return a 3-namedtuple, (ISO year, ISO week number, ISO weekday).
```
taleinat: I don't believe "3-namedtuple" is in customary use, used elsewhere in our docs, nor clear enough in this context. Just use "named tuple".

Regardless, we'll need to describe the new `IsoCalendarDate` class explicitly (somewhere). So perhaps reference that instead.

Doc/whatsnew/3.9.rst
```diff
@@ -109,6 +109,14 @@ New Modules
 Improved Modules
 ================
 
+datetime
+--------
+The :meth:`~datetime.date.isocalendar()` of :class:`datetime.date`
+and :meth:`~datetime.datetime.isocalendar()` of :class:`datetime.datetime`
+methods now returns a :func:`~collections.namedtuple` instead of a :class:`tuple`.
```
taleinat, line 152: Both here and in the NEWS entry:

```suggestion
methods now return a :term:`named tuple` instead of a :class:`tuple`.
```

Lib/datetime.py
```diff
@@ -1091,7 +1094,7 @@ def isoweekday(self):
         return self.toordinal() % 7 or 7
 
     def isocalendar(self):
-        """Return a 3-tuple containing ISO year, week number, and weekday.
+        """Return a 3-named tuple containing ISO year, week number, and weekday.
```
taleinat: See above comment.

```suggestion
        """Return a named tuple containing ISO year, week number, and weekday.
```

Lib/datetime.py
```diff
@@ -1555,6 +1558,9 @@ def __reduce__(self):
 time.max = time(23, 59, 59, 999999)
 time.resolution = timedelta(microseconds=1)
 
+IsoCalendarDate = namedtuple('IsoCalendarDate', 'year week weekday')
```
taleinat: Why is this here in the middle of the file, after its first use? This seems a rather strange location. Perhaps nearer the beginning of the file?

Lib/test/datetimetester.py
```diff
@@ -1363,6 +1364,13 @@ def test_isocalendar(self):
             d = self.theclass(2010, 1, 4+i)
             self.assertEqual(d.isocalendar(), (2010, 1, i+1))
 
+    def test_isocalendar_named(self):
+        d = self.theclass(2008, 12, 3)
+        t = d.isocalendar()
+        self.assertEqual(t.year, 2008)
+        self.assertEqual(t.week, 49)
+        self.assertEqual(t.weekday, 3)
+
```
taleinat, line 1385: Great job adding this test!

I suggest also checking equality with:
1. an explicitly created `IsoCalendarDate` instance
2. an equivalent 3-tuple (for backwards compatibility)

Doc/whatsnew/3.9.rst
```diff
@@ -109,6 +109,14 @@ New Modules
 Improved Modules
 ================
 
+datetime
+--------
+The :meth:`~datetime.date.isocalendar()` of :class:`datetime.date`
```
taleinat, line 150: This wording seems overly verbose. I suggest:

```ReST
The ``isocalendar()`` methods of :class:`datetime.date` and :class:`datetime.datetime` now return ...
```

taleinat (changes requested):
In addition to the in-line comments, note also that the new `IsoCalendarDate` class needs its own (small) section in the docs.

Modules/_datetimemodule.c
```diff
@@ -3244,7 +3267,16 @@ date_isocalendar(PyDateTime_Date *self, PyObject *Py_UNUSED(ignored))
...
         week = 0;
     }
-    return Py_BuildValue("iii", year, week + 1, day + 1);
+
+    PyStructSequence_SET_ITEM(v, 0, PyLong_FromLong(year));
+    PyStructSequence_SET_ITEM(v, 1, PyLong_FromLong(week + 1));
+    PyStructSequence_SET_ITEM(v, 2, PyLong_FromLong(day + 1));
+
+    if (PyErr_Occurred()) {
+        Py_XDECREF(v);
```
ZackerySpytz:
```suggestion
        Py_DECREF(v);
```

Modules/_datetimemodule.c
```diff
@@ -3234,6 +3254,9 @@ date_isocalendar(PyDateTime_Date *self, PyObject *Py_UNUSED(ignored))
     int  week;
     int  day;
 
+    PyObject* v = PyStructSequence_New(&StructIsoCalendarDateType);
```
ZackerySpytz:
```suggestion
    PyObject *v = PyStructSequence_New(&StructIsoCalendarDateType);
```
Also, I think this `PyStructSequence_New()` call should be moved down (so that it is closer to the `PyStructSequence_SET_ITEM()` calls) .

Modules/_datetimemodule.c
```diff
@@ -3234,6 +3254,9 @@ date_isocalendar(PyDateTime_Date *self, PyObject *Py_UNUSED(ignored))
...
+    if (!v) return NULL;
```
ZackerySpytz: The `return` should not be on the same line as the `if`.
```c
if (v == NULL) {
    return NULL;
}
```
Please see the [PEP 7](https://www.python.org/dev/peps/pep-0007/) style guide.

corona10:
@taleinat @ZackerySpytz cc @pganssle @abalkin

First of all, thanks for the code review.
I 've reflected the review for this PR.

However, other comments have been raised by the core developer on whether this patch is acceptable. 
IMHO, It is a good idea to review the patch after the discussion is complete.

Again, thank you for the code reviews. :)

Doc/library/datetime.rst
```diff
@@ -122,6 +122,14 @@ Available Types
 
    .. versionadded:: 3.2
 
+.. class:: IsoCalendarDate
```
pganssle: Perhaps @taleinat will disagree with me, but if we do make this switch (and as I've said in the BPO thread, I'm not in favor of it), I consider the name and details of the return type class to be a bit of an implementation detail.

Practically speaking we would need to document that it is a `namedtuple` of some sort and what its attributes are, but ideally this would be a purely private class. I would never want to see code like `isinstance(x, datetime.IsoCalendarDate)` or `x = IsoCalendarDate(2019, 1, 1)`.

As such, if you document it anywhere it should probably not be referred to by name and just document its properties in the `isocalendar` documentation.

corona10:
@rhettinger 
Is there anything can review with this PR except updating PR naming and updating documentation?

Lib/test/datetimetester.py
```diff
@@ -1363,6 +1364,14 @@ def test_isocalendar(self):
             d = self.theclass(2010, 1, 4+i)
             self.assertEqual(d.isocalendar(), (2010, 1, i+1))
 
+    def test_isocalendar_named(self):
```
pganssle: How about refactoring this test and `test_isocalendar` together into a single parametrized test:

```python
def test_isocalendar(self):
    # Check examples from
    # https://www.phys.uu.nl/~vgent/calendar/isocalendar.htm
    week_mondays = [
        ((2003, 12, 22), (2003, 52, 1)),
        ((2003, 12, 29), (2004, 1, 1)),
        ((2004, 1, 5), (2004, 2, 1)),
        ((2009, 12, 21), (2009, 52, 1)),
        ((2009, 12, 28), (2009, 53, 1)),
        ((2010, 1, 4), (2010, 1, 1)),
    ]

    test_cases = []
    for cal_date, iso_date in week_mondays:
        base_date = self.theclass(*cal_date)

        # Adds one test case for every day of the specified weeks
        for i in range(7):
            new_date = base_date + timedelta(i)
            new_iso = iso_date[0:2] + (iso_date[2] + 1,)

            test_cases.append((new_date, new_iso))

    for d, exp_iso in test_cases:
        with self.subTest(d=d, comparison="tuple"):
            self.assertEqual(d.isocalendar(), exp_iso)

        # Check that the tuple contents are accessible by field name
        with self.subTest(d=d, comparison="fields"):
            iso_tuple = d.isocalendar()
            self.assertEqual((iso_tuple.year, iso_tuple.week, iso_tuple.day),
                             exp_iso)
```

I recognize that it adds a bit of complexity to the code, but it will give more granular output than we have now while also allowing us to easily add new examples if needed.

pganssle (changes requested):
I haven't had time to do a full review yet, but I've added one comment about refactoring a test (feel free to dispute that if you disagree), and I do think that we should make the particular class used a private implementation detail and *not* expose it, per my comment [here](https://github.com/python/cpython/pull/15633/files#r319766276).

@taleinat I know that you asked for `IsoCalendarDate` to be documented - are you OK with us only guaranteeing that the return value has `year`, `week` and `day` fields but not exposing the actual class for `isinstance` or construction purposes? If so, I think we can just document those properties in the `isocalendar()` documentation.

Lib/test/datetimetester.py
corona10: Great idea

rhettinger:
Marking this as closed.  See the discussion on BPO where we've decided that collections.namedtuple() wasn't the way to go here.  Instead, C functions should use structseq.

Thanks for PR and sorry the effort didn't come to fruition.

pganssle:
@rhettinger This PR uses a structseq, it's what we used for the benchmarks, and I see no reason to close it. Dong-Hee can just push more commits with any changes you might request.

tim-one:
I changed the title to replace "named tuple" with "structseq".  All the code and doc  changes checked in here should do something similar - the correct name is "struct sequence" (see, e.g., the existing docs for `sys.float_info`).

rhettinger:
Paul, are you still wanting to go down the path of using *collections.namedtuple()* instead of *structseq*?   If not, this PR needs to be closed.

tim-one:
Raymond, the actual current code here uses `PyStructSequence_XXX` calls.  Not named tuples anymore.  That's why I changed the title :smile:.

rhettinger:
I had make a quick scan of the latest version and didn't expect to see `from collections import namedtuple` on line 11 of `Lib/datetime.py`.

tim-one:
Yes, the Python and C implementations differ in the current state.  I just assumed Paul was keen to change the C version to structseqs first, to get his benchmarks done.

rhettinger:
Also, I had thought we were going to use "named tuple" (two words) to cover the generic concept including both *collections.namedtuple()* and *structseq*.  See https://docs.python.org/3/glossary.html#term-named-tuple

I don't think the user ever sees something called a "structseq".  

```
>>> import time
>>> time.localtime()
time.struct_time(tm_year=2019, tm_mon=9, tm_mday=8, tm_hour=16, tm_min=50, tm_sec=17, tm_wday=6, tm_yday=251, tm_isdst=1)
>>> type(_)
<class 'time.struct_time'>
```

I don't think the API of a *structseq* is publicly documented anywhere, so it may not make sense to refer to it at all.  For example, see what was done for https://docs.python.org/3/library/time.html#time.struct_time

pganssle:
> I had make a quick scan of the latest version and didn't expect to see `from collections import namedtuple` on line 11 of `Lib/datetime.py`.

In any case, I suggest giving this a full review and give specifics of how to change things. I have been assuming that because the docs say that a `structseq` is the "C version of a `namedtuple`" that we would be using a struct sequence in C and a `namedtuple` in the pure Python version, because I can't find any documentation on how to create a struct seq in pure Python.

> Yes, the Python and C implementations differ in the current state. I just assumed Paul was keen to change the C version to structseqs first, to get his benchmarks done.

I didn't make any changes, actually. I've never seen any version of this PR that *didn't* use a structseq in the C code. I think we should do whatever keeps the C and Python versions as close as possible without sacrificing performance is what we should do. If there's a way to create a structseq in pure Python, we should do that in `datetime.py`.

> I changed the title to replace "named tuple" with "structseq". All the code and doc changes checked in here should do something similar - the correct name is "struct sequence" (see, e.g., the existing docs for `sys.float_info`).

This may come down to how it ends up getting implemented, but if the pure python version is a `collections.namedtuple`, then I think we should use a generic term that covers both (and even if the two implementations use the same type, we should maybe still do so). The main concern is that I think that as the reference implementation we should probably not be overly prescriptive in how it gets implemented - the interface is what's more important here, and we should document that.

pganssle:
> I don't think the API of a _structseq_ is publicly documented anywhere, so it may not make sense to refer to it at all.

The only documentation I can find for struct sequences is [the C API documentation](https://docs.python.org/3/c-api/tuple.html#struct-sequence-objects). It seems to indicate that the C API is the only API for constructing these things and seems to imply that the pure Python version is `collections.namedtuple`. The two examples you and Tim have mentioned so far don't have pure Python implementations, so perhaps this is the first module that has to solve this problem?

tim-one:
I'm not going to argue about the docs.  The very same glossary you pointed at also has an entry for "struct sequence":

> struct sequence
> A tuple with named elements. Struct sequences expose an interface similar to named tuple in that elements can be accessed either by index or as an attribute. However, they do not have any of the named tuple methods like _make() or _asdict(). Examples of struct sequences include sys.float_info and the return value of os.stat().

and, as already noted, `sys.float_info` is documented as

> A struct sequence holding information about the float type. 

Paul, I believe you're right - alas - that the C-level `structseq` implementation isn't available to Python-level code.

rhettinger:
The name doesn't matter much to me either.

tim-one:
"It would be nice" if the docs were consistent, but I don't care much either :wink:.

Assuming you agree there's no way to build a new flavor of structseq from Python code, do you still object to `datetime.py` using a `namedtuple` instead?  It would be at best weird if the Python version had to rely on the C version to supply an `IsoCalendarDate` constructor 😦 .

corona10:
@rhettinger @tim-one @pgagnon 
I will follow the final decision. 
It looks like 3 options to deal with this issue.
1. Use namedtuple only and remove c extension api
2. Use namedtuple for pure python version and Use for structseq for c version
3. Use structseq only remove pure python version.

taleinat:
> It looks like 3 options to deal with this issue.
> 
>     1. Use namedtuple only and remove c extension api
> 
>     2. Use namedtuple for pure python version and Use for structseq for c version
> 
>     3. Use structseq only remove pure python version.

There is another option: 

4. The pure-Python version defines a simple tuple sub-class with attribute access to the items, closely mimicking the structseq used by the C version. This should even allow seamlessly pickling one type and unpickling to the other, in both directions.

corona10:
@taleinat 

> 4. The pure-Python version defines a simple tuple sub-class with attribute access to the items

In that case, should we port _structseq from pypy?
https://bitbucket.org/pypy/pypy/src/default/lib_pypy/_structseq.py

pganssle:
> * Use namedtuple only and remove c extension api
> * Use namedtuple for pure python version and Use for structseq for c version
> * Use structseq only remove pure python version.

1 and 3 are non-starters, this feature is *definitely* not important enough to break the separation between the C and Python versions of datetime. I think @taleinat's suggestion is the correct thing to do.

> In that case, should we port _structseq from pypy?
> https://bitbucket.org/pypy/pypy/src/default/lib_pypy/_structseq.py

That seems like overkill for the initial implementation, to me. It may be a good idea to add a pure python `structseq` implementation, but since this is the only thing that actually needs it right now, I think a one-off custom `tuple` subclass is the simplest thing to do.

corona10:
@pganssle @taleinat 
Okay, then I will start to implement `IsoCalendarDate `through subclassing tuple to mimicking the `structseq` used by the C version.

cc @rhettinger

corona10:
@pganssle @taleinat
By the way, is there any document or code that I can reference for implementing capability between subclass of tuple and structseq?

pganssle:
@corona10 Something like this should suffice:

```python
class IsoCalendarDate(tuple):
    def __new__(cls, /, year, week, weekday):
        return super().__new__(cls, (year, week, weekday))

    @property
    def year(self):
        return self[0]

    @property
    def week(self):
        return self[1]

    @property
    def weekday(self):
        return self[2]

    def __repr__(self):
        return (
            f"{self.__class__.__name__}"
            + f"(year={self[0]}, week={self[1]}, weekday={self[2]})"
        )
```

I will note that this has a slightly different constructor from the generated struct sequence, but that is not really important since I think we should make `IsoCalendarDate` a private implementation detail and not a class that end-users can construct. If you wanted the constructor to be the same for whatever reason, you would change `__new__` to look like this:

```python
    def __new__(cls, seq):
        if len(seq) != 3:
            raise TypeError(f"{cls.__name__}() takes a 3-sequence " +
                            f"({len(seq)}-sequence given)")

        return super().__new__(cls, seq)
```

In either case, please modify the documentation so that `IsoCalendarDate` is no longer documented as a top-level class and instead document the fields of the struct in `date.isocalendar` and/or `datetime.isocalendar`.

corona10:
@pganssle Please take a look. I've updated the PR.

corona10:

Lib/datetime.py
```diff
@@ -1091,7 +1116,7 @@ def isoweekday(self):
         return self.toordinal() % 7 or 7
 
     def isocalendar(self):
-        """Return a 3-tuple containing ISO year, week number, and weekday.
+        """Return a IsoCalendarDate containing ISO year, week number, and weekday.
```
pganssle:
```suggestion
        """Return a named tuple containing ISO year, week number, and weekday.
```

Lib/datetime.py
```diff
@@ -8,6 +8,31 @@
 import math as _math
 import sys
 
+
+class IsoCalendarDate(tuple):
+    def __new__(cls, seq):
```
pganssle: Is there a reason you chose this implementation rather than the other one? I suspect the other one will be somewhat more robust (even if the error message is different).

Doc/whatsnew/3.9.rst
```diff
@@ -116,6 +116,12 @@ Added the *indent* option to :func:`~ast.dump` which allows it to produce a
 multiline indented output.
 (Contributed by Serhiy Storchaka in :issue:`37995`.)
 
+datetime
+--------
+The ``isocalendar()`` methods of :class:`datetime.date` and
+:class:`datetime.datetime` now return a :class:`IsoCalendarDate`
```
pganssle:
```suggestion
:class:`datetime.datetime` now return a named tuple
```

Or

```suggestion
:class:`datetime.datetime` now return a :term:`named tuple`
```

Doc/library/datetime.rst
```diff
@@ -1230,8 +1233,8 @@ Instance methods:
 
 .. method:: datetime.isocalendar()
 
-   Return a 3-tuple, (ISO year, ISO week number, ISO weekday).  The same as
-   ``self.date().isocalendar()``.
+   Return a :class:`IsoCalendarDate`, (ISO year, ISO week number, ISO weekday).
```
pganssle:
```suggestion
   Return a named tuple, (ISO year, ISO week number, ISO weekday).
```

Modules/_datetimemodule.c
```diff
@@ -3383,7 +3420,7 @@ static PyMethodDef date_methods[] = {
      PyDoc_STR("Return time tuple, compatible with time.localtime().")},
 
     {"isocalendar", (PyCFunction)date_isocalendar,  METH_NOARGS,
-     PyDoc_STR("Return a 3-tuple containing ISO year, week number, and "
+     PyDoc_STR("Return a IsoCalendarDate containing ISO year, week number, and "
```
pganssle:
```suggestion
     PyDoc_STR("Return a named tuple containing ISO year, week number, and "
```

Doc/library/datetime.rst
```diff
@@ -590,7 +590,7 @@ Instance methods:
 
 .. method:: date.isocalendar()
 
-   Return a 3-tuple, (ISO year, ISO week number, ISO weekday).
+   Return a :class:`IsoCalendarDate`, (ISO year, ISO week number, ISO weekday).
```
pganssle: This link to `IsoCalendarDate` will or should be dead, as this is an undocumented class. I have made suggestions in many places to change it to "named tuple", in this documentation I suggest you reword it like this:

Return a :term:`named tuple` object with three components: ``year``, ``week`` and ``weekday``.

Doc/library/datetime.rst
```diff
@@ -606,6 +606,9 @@ Instance methods:
    ``date(2003, 12, 29).isocalendar() == (2004, 1, 1)`` and ``date(2004, 1,
    4).isocalendar() == (2004, 1, 7)``.
 
+   .. versionchanged:: 3.9
+      Result changed from a tuple to a :class:`IsoCalendarDate`.
```
pganssle:
```suggestion
      Result changed from a tuple to a :term:`named tuple`.
```

Doc/library/datetime.rst
```diff
@@ -606,6 +606,9 @@ Instance methods:
...
+      It can't be unpickled in previous versions.
```
pganssle: I think I would remove this line. In general pickling something in one version of Python and unpickling it in another version is not a supported operation and I think the fact that the class of the return value changed should be enough of a hint to people debugging this issue if it comes up.

pganssle (changes requested):

Doc/library/datetime.rst
corona10: @pganssle 

> The pure-Python version defines a simple tuple sub-class with attribute access to the 
>  items, closely mimicking the structseq used by the C version. 

`:term:named tuple` is the right choice?
We are actually mimicking structseq aren't we?

tim-one: Unfortunately, the Glossary has entries for both "named tuple" and "struct sequence", and it's clear as mud what the difference is intended to be.  At this point I'd be in favor of getting rid of the phrase "struct sequence" entirely everywhere in the docs, using only "named tuple" instead.  But that's outside the scope of this PR. 

The abbreviated name "structseq" doesn't mean anything to anyone outside CPython developers - it's an internal name used in CPython's C implementation code.

So, bottom line:  Paul's "`named tuple` object" looks best to me for now, although it will become more obviously clearer later, when the current "struct sequence"/"named tuple" confusion is purged from the docs.

Lib/test/datetimetester.py
```diff
@@ -1349,19 +1350,32 @@ def test_weekday(self):
...
+                new_date = base_date + timedelta(i)
+                new_iso = iso_date[0:2] + (iso_date[2] + i,)
+                test_cases.append((new_date, new_iso))
+
+        for d, exp_iso in test_cases:
+            with self.subTest(d=d, comparison="tuple"):
+                self.assertEqual(d.isocalendar(), exp_iso)
+
+            # Check that the tuple contents are accessible by field name
+            with self.subTest(d=d, comparison="fields"):
```
taleinat, line 1382: The two loops here can be combined, which will also reduce the number of sub-tests.

Modules/_datetimemodule.c
```diff
@@ -6516,6 +6553,17 @@ PyInit__datetime(void)
     PyModule_AddIntMacro(m, MINYEAR);
     PyModule_AddIntMacro(m, MAXYEAR);
 
+    /* IsoCalendarDate */
+    if (!initialized) {
```
taleinat: In existing uses of structseq, I see both this initialization idiom used as well as checking whether `<type object>.tp_name == NULL`. The latter avoids using a module-level static variable, keeping the relevant logic all in one place. Would someone like to weigh in on which is preferred here?

tim-one: > Would someone like to weigh in on which is preferred here?

Don't really care.  As a (very) old-timer here, I recognize the intent of the module-level static instantly, because that's how it was done from the start.  But the other way is fine too, provided there's a comment clarifying the intent of the otherwise-weird-looking test.

tim-one:
FYI, I opened a different issue about the "struct sequence" / "named tuple" doc stuff:

https://bugs.python.org/issue38096

Modules/_datetimemodule.c
```diff
@@ -6516,6 +6553,17 @@ PyInit__datetime(void)
...
+        if (PyStructSequence_InitType2(&StructIsoCalendarDateType,
+                                       &struct_iso_calendar_date_desc) < 0) {
+            return NULL;
+        }
+        initialized = 1;
+    }
+    Py_INCREF((PyObject *) &StructIsoCalendarDateType);
+    PyModule_AddObject(m, "IsoCalendarDate", (PyObject *) &StructIsoCalendarDateType);
```
pganssle: I think we can drop this line - there's no particular need to expose this as a public, constructable object in the C module. You'll also need to add `IsoCalendarDate` into [this list of cleaned up symbols](https://github.com/python/cpython/blob/3f4db4a0bab073b768fae958e93288bd5d24eadd/Lib/datetime.py#L2513) so it isn't still exposed.

Lib/datetime.py
```diff
@@ -8,6 +8,31 @@
 import math as _math
 import sys
 
+
+class IsoCalendarDate(tuple):
```
pganssle: I think the name of this symbol should be `_IsoCalendarDate`, because it turns out that `datetime` doesn't have an `__all__`, which would make `IsoCalendarDate` presumptively public. We want the repr to continue to say `IsoCalendarDate`, though, so I see three options:

1. Declare the class name as `IsoCalendarDate` then rename the symbol and delete the old symbol, like so:

```python
class IsoCalendarDate(tuple):
    ...

# The class name shouldn't start with an underscore, but the symbol should
_IsoCalendarDate = IsoCalendarDate
del IsoCalendarDate
```

2. Set `__name__` and `__qualname__`, like so:

```python
class _IsoCalendarDate(tuple):
    ...
_IsoCalendarDate.__name__ = "IsoCalendarDate"
_IsoCalendarDate.__qualname__ = "IsoCalendarDate"
```

3. Change the `__repr__` implementation to hard-code the class name, like so:

```python
    def __repr__(self):
        return ('datetime.IsoCalendarDate'
            f'(year={self[0]}, week={self[1]}, weekday={self[2]})')
```

There are positives and negatives from each approach and, ultimately, it doesn't really matter much. I think the first approach is the most natural.

pganssle: Oh, I should also note, this is not a great place in the file for this class's definition. It's at the top, way before all the other class definitions. I would define it [right below the definition for `date`](https://github.com/python/cpython/blob/3f4db4a0bab073b768fae958e93288bd5d24eadd/Lib/datetime.py#L1140).

pganssle (changes requested):

Lib/datetime.py
corona10: @pganssle 
This makes `python attribute lookup IsoCalendarDate on datetime failed` when pickling dump.

Lib/datetime.py
corona10: @pganssle This implementation allows the pickling exchange between python version and C version.

corona10:
@pganssle 
To support pickle IsoCalendar, https://github.com/python/cpython/pull/15633/commits/20b10a8bff9da750c0de2ecbf7b216855729fc4a is needed. Please leave a feedback for this.

Lib/test/datetimetester.py
pganssle: @taleinat It's not clear to me what you mean by this - in my original proposal I deliberately used two loops because I thought it was more readable to 1. separate "test case creation" and "evaluating test cases" and 2. keep a lower level of nesting in each of the separate loops.

I also am not sure how combining the loops would reduce the number of sub-tests - without modifying precisely what you loop over the subtests are going to stay the same.

There's definitely a case to be made for not doing the full matrix of "every day of the week in each of these example weeks", but each pass of the loop is relatively inexpensive, and there's only 6 x 7 = 42 calls, tested twice. If we're concerned about it we could refactor like this, maybe:

```python
for d, exp_iso in test_cases:
    with self.subTest(d=d, subtest="call"):
        t = d.isocalendar()

    with self.subTest(d=d, subtest="tuple_comparison"):
        self.assertEqual(t, exp_iso)

    with self.subTest(d=d, subtest="field_comparison"):
        self.assertEqual((t.year, t.week, t.weekday), exp_iso)
```

Though `.isocalendar()` is fast, I suspect faster than entering and leaving a `subTest` context - it would only make sense to make this change to cleanly separate the three separate things we are actually testing here.

Lib/test/datetimetester.py
```diff
@@ -1349,19 +1350,29 @@ def test_weekday(self):
...
+        for cal_date, iso_date in week_mondays:
+            base_date = self.theclass(*cal_date)
+            # Adds one test case for every day of the specified weeks
+            for i in range(7):
+                new_date = base_date + timedelta(i)
+                new_iso = iso_date[0:2] + (iso_date[2] + i,)
+                test_cases.append((new_date, new_iso))
+
+        for d, exp_iso in test_cases:
+            with self.subTest(d=d, comparison="fields and tuple"):
```
pganssle: I think these are at *least* two different tests, combining them into one subtest sorta defeats the point of subtest, can you revert to the old version, with two separate `subTest`s?

pganssle (changes requested):
I think the pickling interface is causing some problems, so I think the right thing to do at this point is to just pickle the underlying `tuple` object rather than trying to retain the `namedtuple`.

We can always remove that extension later (it won't be any more breaking than changing it now would be), and this allows us to keep `IsoCalendarDate` an implementation detail. I highly doubt that anyone will notice this, as it's very unlikely that anyone wants to pickle an object, pass it somewhere and then access its value by field (it's certainly possible, but part of the motivation for this change is that people are mostly just immediately accessing one or another of the fields of this tuple rather than passing around the result as a whole).

pganssle:
@corona10 This has merge conflicts now, can you pull the current version (including my commits), then rebase against master?

corona10:
@pganssle Sure, I am working on this.
It looks like the PR is affected by https://github.com/python/cpython/commit/3fb1363fe87a24cdb2ee1dd9746f1c49046af958

corona10:
@pganssle 
Rebased! 

corona10:
@pganssle `test_os`is failed..

pganssle:
@corona10 Not sure what I was thinking, but what I did in the C code in 6690903deecd9ebc1973e8944ea75281 was *not* the way you modify a struct sequence's `__reduce__` method. I don't have time at the moment to deal with the fiddly details of creating a subclass in C, so I've reverted my changes, but I think that having `isocalendar().__reduce__` return `(tuple, (tuple(self),))` is the right thing to do to minimize the changes to the public API.

If you have time, can you change `StructIsoCalendarDateType` to a struct sequence subclass with a custom `__reduce__`, as I inexpertly tried to do in my commit?

corona10:
@pganssle 
How about just creating type as `_ IsoCalendarDate`?
The current implementation is well working to be able to exchange the python version and c version.

pganssle:
@corona10 The big tension I have is that I would very much like the repr to say `IsoCalendarDate` rather than `_IsoCalendarDate`. Will pickling still work on a StructSequence if we set `__name__` to something other than the name of the attribute?

If not, we will either have to override `__reduce__` or `__repr__`, and if we're already overriding methods, we should do `__reduce__`.

corona10:
@pganssle Since https://github.com/python/cpython/pull/16203 is merged. We can proceed this PR?

pganssle:
@corona10 #16203 makes just merging this and relying on `__all__` to advertise that `IsoCalendarDate` is not public a possibility, but I still prefer the idea of at least an initial release that pickles these things as normal tuples. Once we release something with a pickleable `IsoCalendarDate`, we can't really go back, but it's backwards compatible to switch the class that gets pickled to `IsoCalendarDate`.

Supporting round-trips through pickling in `IsoCalendarDate` also *requires* us to maintain `datetime.IsoCalendarDate` as a constructable class, otherwise pickle won't work.

I recognize that it's a bit of a pain to do the subclassing in C, but given that this is targeting Python 3.9, we have plenty of time to finish the feature before the next release.

corona10:
@pganssle 
Okay, I am going to write the subclass of the tuple in C ASAP.
I will update this PR when the code is ready :)

corona10:
@pganssle cc @taleinat 

I 've updated the C version of IsoCalendarDate from structseq to tuple based.
And I check the reference leak test and it passed.

Please take a look 
Thanks alot
```
./python.exe -m test test_datetime -R 3:3
Run tests sequentially
0:00:00 load avg: 3.07 [1/1] test_datetime
beginning 6 repetitions
123456
......
test_datetime passed in 1 min 7 sec

== Tests result: SUCCESS ==

1 test OK.

Total duration: 1 min 7 sec
Tests result: SUCCESS
```

Modules/_datetimemodule.c
```diff
@@ -3225,6 +3227,154 @@ date_isoweekday(PyDateTime_Date *self, PyObject *Py_UNUSED(ignored))
...
+static PyGetSetDef iso_calendar_date_getset[] = {
+    {"year",        (getter)iso_calendar_date_year},
+    {"week",      (getter)iso_calendar_date_week},
+    {"weekday",      (getter)iso_calendar_date_weekday},
+    {NULL}
+};
+
+static PyTypeObject PyDateTime_IsoCalendarDateType = {
+    PyVarObject_HEAD_INIT(NULL, 0)
+    "datetime.IsoCalendarDate",                         /* tp_name */
```
corona10: @pganssle 
Now, we can use `datetime._IsoCalendarDate` for tp_name
and `datetime.IsoCalendarDate` for repr.

Do you want it?

corona10:
@pganssle
Now, this PR is working as you expect.
My one concern is if the user write the code in below way.
```python

pickle.dump(d.isocalendar(), f)
f.seek(0)
res = pickle.load(f)

print(res.year)
```
It will not work as the user's expected.
Should we document about it?

corona10:
Since the current implementation has no way to recover the origin type `IsoCalendarDate`
I'd like to suggest it `IsoCalendarDate` as a public type.
So that user can recover ``IsoCalendarDate` by this way.
```python
        d = self.theclass(2019, 1, 1)
        f = io.BytesIO()
        pickle.dump(d.isocalendar(), f)
        f.seek(0)
        res = pickle.load(f)
        self.assertEqual(res, (2019, 1, 2))
        p = datetime_module.IsoCalendarDate(res)
        self.assertEqual(p.year, 2019)
````

IMHO, It can catch both backward compatible and the new feature.

**Comment by corona10** [comment]
I ve noticed that the above issue could be solved through `datetime.date.fromisocalendar`
So no issue yet all.
Please take a look :)

**Comment by pganssle** [comment]
@corona10 It is indeed true that pickling and unpickling the result of calling `isocalendar()` will not give you an `IsoCalendarDate` object, it will give a `tuple` object. I don't think this will be a big deal, because one of the main justifications for switching to a named tuple response is the empirical evidence that most people are *not* passing around the result of `isocalendar()` calls, but rather immediately accessing one or two components and passing those around.

I suspect that very few people will notice that these things pickle as if they were normal tuples, and considering that they have always actually *been* normal tuples, it's unlikely that anyone will immediately start relying on the fact that they can tuple an `IsoCalendarDate` object.

If this starts causing real bugs, we can just make `IsoCalendarDate` public, but really, I doubt they will.

We should definitely document it, though.

File: Modules/_datetimemodule.c [hunk 26]
```diff
@@ -3225,6 +3227,174 @@ date_isoweekday(PyDateTime_Date *self, PyObject *Py_UNUSED(ignored))
     return PyLong_FromLong(dow + 1);
 }
 
+PyDoc_STRVAR(iso_calendar_date__doc__,
+"The result of date.isocalendar() or datetime.isocalendar()\n\n\
+This object may be accessed either as a tuple of\n\
+  ((year, week, weekday)\n\
+or via the object attributes as named in the above tuple.");
+
+typedef struct {
```
pganssle, line 3236: I have not spent enough time trying to implement this to know, but is there a reason why you've implemented a whole new class here rather than creating a `StructSeq` and then subclassing that rather than subclassing `tuple` directly?

Admittedly it's a bit neater to not complicate the MRO, but it does end up with a lot more implementation code when we really just want to override `__reduce__`.

Modules/_datetimemodule.c
```diff
@@ -3225,6 +3227,174 @@ date_isoweekday(PyDateTime_Date *self, PyObject *Py_UNUSED(ignored))
...
+};
+
+/*[clinic input]
+@classmethod
+datetime.IsoCalendarDate.__new__ as iso_calendar_date_new
+    sequence as seq: object
+[clinic start generated code]*/
+
+static PyObject *
+iso_calendar_date_new_impl(PyTypeObject *type, PyObject *seq)
```
pganssle: If we stick with this version of the implementation, it's probably better to have a constructor that takes 3 C integers rather than a tuple of `PyLong_Long` objects, since the only place this should ever actually be constructed is in the code for `isocalendar()` and in the current form of this implementation, `isocalendar()` is first constructing an unnecessary intermediate `tuple` object, which is then broken apart to create the `IsoCalendarDate` object.

Modules/_datetimemodule.c
corona10: @pganssle 
I actually intend to subclassing tuple directly with current implementation.
However, as I am not an expert of C API extension. 
if this is the wrong way, please let me know the standard way of subclassing tuple in C API.

other will work as same as tuple.

I am just overriding tuple with
`__repr__ `-> iso_calendar_date_repr
`__reduce__` -> isocalendardate_reduce
`__getset__` -> iso_calendar_date_methods
as same as pure python version.

tp_base is also PyTuple_Type,

corona10: You can see that IsoCalendarDate support basic tuple's operation without any define

pganssle: Yes I noticed you are subclassing `tuple` directly, but that means you have to re-implement `StructSequence` yourself here. It's not a terribly complicated thing to implement, so it could be a reasonable thing to do for many reasons, I just wanted to check if you had deliberately ruled out using a StructSequence as the base instead of `tuple`.

Modules/_datetimemodule.c
corona10: https://github.com/python/cpython/blob/0c989db57ec189511dc876973850bbd31c547ec0/Lib/datetime.py#L1216
So it means that we should update the pure python version of `__new__`?

Modules/_datetimemodule.c
corona10: reference https://docs.python.org/3/extending/newtypes_tutorial.html#subclassing-other-types

corona10: @pganssle 
Oh, now I understood what you want to say (Sorry for my poor English)
As you know when we create this type with structsequence there was a test failure with test_os.
So I try to re-implement it as same as pure python and I notice that it works well.

> that means you have to re-implement StructSequence yourself here.

Yes, this is some kinds of complicated works but since we re-implement named tuple in pure python version. this works looks like same work around.

corona10:
@pganssle 
Hi, I've update the constructor to accept 3 ints without creating unnecessary intermediate tuple object.
Please take a look. 

And should we try to create a structseq based type? We can try but
I think the current implementation is the right way from the point of view of symmetry between Python and C versions.
What do you think?

corona10:
@pganssle gentle ping cc @taleinat

Lib/datetime.py
```diff
@@ -1210,6 +1210,34 @@ def __reduce__(self):
...
+    @property
+    def weekday(self):
+        return self[2]
+
+    def __reduce__(self):
+        return (tuple, (tuple(self),))
+
+    def __repr__(self):
+        return (f'{self.__class__.__name__}'
+            f'(year={self[0]}, week={self[1]}, weekday={self[2]})')
```
vstinner: nitpick: try to align it with the f on the first line ;-) put it on 3 lines if needed.

Lib/datetime.py
```diff
@@ -1210,6 +1210,34 @@ def __reduce__(self):
...
+    @property
+    def week(self):
+        return self[1]
+
+    @property
+    def weekday(self):
+        return self[2]
+
+    def __reduce__(self):
+        return (tuple, (tuple(self),))
```
vstinner, line 1234: Is it deliberate to deserialize it as a regular tuple? If yes, please add a comment to explain the rationale.

Lib/test/datetimetester.py
```diff
@@ -1355,19 +1356,45 @@ def test_weekday(self):
...
+
+        The result of a round trip should be a plain tuple.
+        """
+        d = self.theclass(2019, 1, 1)
+        f = io.BytesIO()
+        pickle.dump(d.isocalendar(), f)
+        f.seek(0)
+        res = pickle.load(f)
+
+        self.assertEqual(res, (2019, 1, 2))
```
vstinner, line 1395: Please test also *res* type.

Lib/test/datetimetester.py
```diff
@@ -1355,19 +1356,45 @@ def test_weekday(self):
...
+    def test_isocalendar_pickling(self):
+        """Test that the result of datetime.isocalendar() can be pickled.
+
+        The result of a round trip should be a plain tuple.
+        """
+        d = self.theclass(2019, 1, 1)
+        f = io.BytesIO()
+        pickle.dump(d.isocalendar(), f)
+        f.seek(0)
+        res = pickle.load(f)
```
vstinner: You could just use dumps() and loads() to avoid a file, no?

Modules/_datetimemodule.c
vstinner: Can't reuse structseq type but override the type __reduce__ method?

corona10: @vstinner I am going to find a way to do this.
It does not look easy to me, if you can give a hint for me. It will be helpful :)

corona10:

Modules/_datetimemodule.c
corona10: cc @pablogsal

corona10:
cc @vstinner

Modules/_datetimemodule.c
vstinner: If you write a PR using structseq, I suggest to create a separated PR. So we can compare both approaches. Create [WIP] PR without overriding `__reduce__()` if you want help.

corona10: Great idea

Lib/datetime.py
```diff
@@ -1210,6 +1210,36 @@ def __reduce__(self):
         else:
             return (self.__class__, args, state)
 
+
+class IsoCalendarDate(tuple):
+
```
pppery, line 1215:
```suggestion

__slots__ = ()
```

corona10: @pppery Can you explain the reason about this suggestion?

Modules/_datetimemodule.c
corona10: @vstinner @pablogsal 
https://github.com/python/cpython/pull/17651 is the experimental PR for comparing both approach.
I'd like to discuss how we will finalize this PR. (Do not change is the one of the options.)

pganssle:
@corona10 Hey, sorry I never got back to this, I've been swamped with PEP 615.

Do you mind letting me know the current state of things? Do you have a recommendation between this implementation and the other one?

I'm also increasingly amenable to the idea of just having this thing named `_IsoCalendar` or something (particularly if we can overload the `__repr__` to have it return `IsoCalendar(...)` instead?) if it's too complicated to do anything else.

corona10:
@pganssle 

Hmm, I should remember what I tried before for this issue :)

AFAIK(??) I still prefer the current PR implementation.  
Since https://github.com/python/cpython/pull/17651 is too verbose. (It's just my opinion)
he current implementation is able to compatible with pure python code.
This PR can be the same level of verbose but we can bring compatible issues easily.
(You know namedtuple and structseq is quite different.)

But if you decide specific requirements then we can change the implementation.
And I prefer the start with new PR.
We(or You) can create a new PR to start.

corona10:
or we have to think that this issue is should be solved?

PR #15634, opened 2019-09-01
bpo-35771: IDLE: Fix and optimize flaky tool-tip hover delay tests

taleinat:
A 50ms delay will often causes failures on slower machines.

This PR is an alternative to PR GH-14926. In this PR, both of the tests with a delay have been merged into one, to avoid multiple `time.sleep()` calls, reducing the overall run time.

For reference, the total run time for `test_idle` on my machine is about 7.5 seconds. 0.15 seconds, which is the delay time added by these tests, is 2% of that.

https://bugs.python.org/issue35771

### DIFF
--- a/Lib/idlelib/idle_test/test_tooltip.py
+++ b/Lib/idlelib/idle_test/test_tooltip.py
@@ -1,26 +1,30 @@
+from functools import wraps
 from idlelib.tooltip import TooltipBase, Hovertip
 from test.support import requires
-requires('gui')
-
-from functools import wraps
 import time
 from tkinter import Button, Tk, Toplevel
 import unittest
+
+
+requires('gui')
 
 
 def setUpModule():
     global root
     root = Tk()
 
+
 def root_update():
     global root
     root.update()
+
 
 def tearDownModule():
     global root
     root.update_idletasks()
     root.destroy()
     del root
+
 
 def add_call_counting(func):
     @wraps(func)
--- a/Lib/idlelib/idle_test/test_tooltip.py
+++ b/Lib/idlelib/idle_test/test_tooltip.py
@@ -108,7 +108,7 @@ class HovertipTest(unittest.TestCase):
         # Run multiple tests requiring an actual delay simultaneously.
 
         # Test #1: A hover tip with a non-zero delay appears after the delay.
-        tooltip1 = Hovertip(self.button, 'ToolTip text', hover_delay=50)
+        tooltip1 = Hovertip(self.button, 'ToolTip text', hover_delay=100)
         self.addCleanup(tooltip1.hidetip)
         tooltip1.showtip = add_call_counting(tooltip1.showtip)
         root_update()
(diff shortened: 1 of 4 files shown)


taleinat:
@ZackerySpytz, your review would be welcome.

@terryjreedy, this is not the approach you recommended in [your comment on GH-14926](https://github.com/python/cpython/pull/14926#issuecomment-518039720). IMO, in this case, it is better to actually test against what happens with Tk rather than with a mock.

Lib/idlelib/idle_test/test_tooltip.py
```diff
@@ -1,27 +1,31 @@
+from functools import wraps
```
terryjreedy: This first few changes are irrelevant to the issue and are a regression from the 'fail fast (as soon as possible)' design used in IDLE tests.  If importing the tested module fails, doing anything else is useless.  If the whole module requires gui and that fails (as is does for a majority of buildbots) doing anything else is useless.  I have changed back to this.

I am also adding module docstring with coverage.

Lib/idlelib/idle_test/test_tooltip.py
```diff
@@ -1,27 +1,31 @@
...
 def setUpModule():
     global root
     root = Tk()
 
+
 def root_update():
     global root
     root.update()
 
+
```
terryjreedy: idlelib is not consitently double spaced around functions.  To me, it is less needed than for classes.  And PEP8 has this, which I think applies here: "Blank lines may be omitted between a bunch of related one-liners".

That aside, root_update just adds call overhead.  I am replacing it.

Lib/idlelib/idle_test/test_tooltip.py
```diff
@@ -65,81 +69,92 @@ class HovertipTest(unittest.TestCase):
     def setUp(self):
         self.top, self.button = _make_top_and_button(self)
 
+    def is_tipwindow_shown(self, tooltip):
+        return tooltip.tipwindow and tooltip.tipwindow.winfo_viewable()
```
terryjreedy, line 73: I might have used an even shorten name ('tip_shown'), but even as is, this is a definite improvement in readability.

Lib/idlelib/idle_test/test_tooltip.py
```diff
@@ -65,81 +69,92 @@ class HovertipTest(unittest.TestCase):
...
         self.button.event_generate('<Enter>', x=0, y=0)
         root_update()
         self.button.event_generate('<Leave>', x=0, y=0)
         root_update()
-        self.assertFalse(tooltip.tipwindow and tooltip.tipwindow.winfo_viewable())
-        self.assertGreater(len(tooltip.showtip.call_args_list), 0)
 
-    def test_dont_show_on_mouse_leave_before_delay(self):
-        tooltip = Hovertip(self.button, 'ToolTip text', hover_delay=50)
+        time.sleep(0.15)
```
terryjreedy, line 136: One sleep of .15 (to Windows accuracy of around .017) is faster than 2 sleeps of .10 (ditto).

Misc/NEWS.d/next/IDLE/2019-09-01-10-22-55.bpo-35771.tdbmbP.rst
```diff
@@ -0,0 +1,2 @@
+Increase the ``hover_delay`` in IDLE's test_tooltip, and run both tests
+waiting for a delay to expire simultaneously, to reduce total run time.
```
terryjreedy: I changed this to what users are more likely to want to know.  "To avoid occasional spurious test_idle failures on slower machines,  increase the ``hover_delay`` in test_tooltip."

terryjreedy (approved):
An average speedup of .004 seconds and greatly decreased chance of spurious failure is a win.  I will merge with my revisions when CI passes.

PR #15637, opened 2019-09-01
Correct minor gramatical mistake in sys.settrace doc

andresdelfino:

### DIFF
--- a/Doc/library/sys.rst
+++ b/Doc/library/sys.rst
@@ -1253,7 +1253,8 @@ always available.
 
    The trace function is invoked (with *event* set to ``'call'``) whenever a new
    local scope is entered; it should return a reference to a local trace
-   function to be used for the new scope, or ``None`` if the scope shouldn't be traced.
+   function to be used for the new scope, or ``None`` if the scope shouldn't be
+   traced.
 
    The local trace function should return a reference to itself (or to another
    function for further tracing in that scope), or ``None`` to turn off tracing
--- a/Doc/library/sys.rst
+++ b/Doc/library/sys.rst
@@ -1253,7 +1253,7 @@ always available.
 
    The trace function is invoked (with *event* set to ``'call'``) whenever a new
    local scope is entered; it should return a reference to a local trace
-   function to be used that scope, or ``None`` if the scope shouldn't be traced.
+   function to be used for the new scope, or ``None`` if the scope shouldn't be traced.
 
    The local trace function should return a reference to itself (or to another
    function for further tracing in that scope), or ``None`` to turn off tracing


corona10:
@vstinner We can add a `skip news` or  `skip issue` label for this kind of PR right?

Doc/library/sys.rst
```diff
@@ -1253,7 +1253,7 @@ always available.
 
    The trace function is invoked (with *event* set to ``'call'``) whenever a new
    local scope is entered; it should return a reference to a local trace
-   function to be used that scope, or ``None`` if the scope shouldn't be traced.
+   function to be used for the new scope, or ``None`` if the scope shouldn't be traced.
```
aeros: From what I can tell, it looks like the original author intended to write this as:

> function to be used for that scope ...

Your suggestion provides an improvement though by making it a bit more explicit. 👍

aeros (approved):
Thanks for the PR @andresdelfino!

> @vstinner We can add a skip news or skip issue label for this kind of PR?

The changes look good to me, and this PR doesn't require an issue or news entry since it's just involving a typo correction. There are other minor changes which can use ``skip news`` and ``skip issue`` labels, but this one is a perfect example.

corona10 (approved):
@aeros167 Thanks for the mentoring :)
This PR looks good to me also.

Doc/library/sys.rst
aeros: The suggested changes causes this line to go above the maximum of 80 for chars per line (currently at 87). This can be easily fixed by splitting the last word into the next line:
```suggestion
   function to be used for the new scope, or ``None`` if the scope shouldn't be
   traced.
```

aeros (changes requested):
Although I approve of the changes made, I just realized that this will cause the length of line 1256 to go beyond the reST maximum (https://devguide.python.org/documenting/#use-of-whitespace):

> All reST files use an indentation of 3 spaces; no tabs are allowed. The maximum line length is 80 characters for normal text

We'll have to trim down this line to be at or below 80 chars before merging it. An easy way to measure the length is by copy/pasting the text into the python interpreter and using the builtin ``len()`` function. There are some areas of the documentation that exceed the 80 char line limit, but it would be preferable if we could fix this one before merging it.

aeros167:
> @aeros167 Thanks for the mentoring :)

No problem! Feel free to @mention me if you have any general triaging or documentation questions. @vstinner is far more knowledgeable, but I should be able to help with more simple issues.

vstinner:
@corona10: "skip news" is fine for documentation-only changes. "skip issue" is ok-ish: I don't think that it's worth it to bother with an issue to track backports. Please add "needs 3.7 backport" and "needs 3.8 backport" labels: I like to continue to update the 3.7 doc, but I don't think that it's worth it to update 2.7 doc. The reference for branches open for bugfixes: https://devguide.python.org/#status-of-python-branches

Doc/library/sys.rst
andresdelfino: Done! I just commited your suggestion.

aeros167:
@vstinner 
> "skip news" is fine for documentation-only changes.

IMO, typo corrections. clarifications, or other minor improvements are definitely appropriate for a ``skip news`` label, but I think it can be beneficial to have a news entry when there's a substantial new section added. From what Brett has told me in the past, it's more of an "opt-out" system where it's up to the author, but I think we should encourage news entries for the more substantial documentation changes. It certainly doesn't hurt.

vstinner:
> Since we're approaching EoL for 2, are we generally avoiding backports for documentation changes for 2.7? I completely agree that it's probably not worthwhile, just wanted to check if there was an official stance on this topic.

In my experience, backports from master to 2.7 branch are painful: lot of conflicts. IMHO it's not worth it.

I'm not aware of an official stance.

ned-deily:
> I think we should encourage news entries for the more substantial documentation changes. It certainly doesn't hurt.

It doesn't hurt but I can think of few instances when it is really helpful to have a NEWS entry for a doc-only change, especially in a case where the change is only grammatical or formatting.  A NEWS entry *might* be appropriate in a case where, say, the original doc entry was just flat-out wrong and had been the cause of users queries.  The intent of NEWS entries is to give a quick hint to problems that have been fixed and new features that have been added to the code base.  It is not meant to document every change in a release; people interested in that level of detail can easily peruse the git change log on GitHub.

I also agree that, in general, we should not be backporting changes to 2.7 at this point.  Users have been living with 2.7's quirks for ten years now.

andresdelfino:
@aeros167 could you review the change I made after your request?

aeros (approved):

aeros167:
> It doesn't hurt but I can think of few instances when it is really helpful to have a NEWS entry for a doc-only change, especially in a case where the change is only grammatical or formatting. A NEWS entry might be appropriate in a case where, say, the original doc entry was just flat-out wrong and had been the cause of users' queries. 

Yeah I was mostly referring to changes that made significant modifications to existing sections, which would change the meaning or significantly increase the amount of information provided. Grammatical or formatting changes definitely wouldn't need news entries. 

> In my experience, backports from master to 2.7 branch are painful: lot of conflicts. IMHO it's not worth it.

> I also agree that, in general, we should not be backporting changes to 2.7 at this point. Users have been living with 2.7's quirks for ten years now.

Thanks for the clarification @vstinner and @ned-deily. Even if it's not an official stance, it's still very useful to have a rough idea of where we stand on 2.7 backports.

vstinner:
AppVeyor job didn't run on the first 3.7 backport PR. So I closed it and created a new one.

PR #15641, opened 2019-09-02
bpo-38006: Avoid closure in weakref.WeakValueDictionary

vstinner:
weakref.WeakValueDictionary defines a local remove() function used as
callback for weak references. This function was created with a
closure.  Modify the implementation to avoid the closure.


tiran (approved):
The refactorization solves my segfault issue.

PR #15645, opened 2019-09-02
bpo-38009: Do not call weakref callbacks that are being collected

pablogsal:
https://bugs.python.org/issue38009

### DIFF
--- a/Include/cpython/objimpl.h
+++ b/Include/cpython/objimpl.h
@@ -97,6 +97,8 @@ typedef struct {
 #define _PyGC_SET_FINALIZED(o) \
     _PyGCHead_SET_FINALIZED(_Py_AS_GC(o))
 
+#define _PyObject_GC_IS_COLLECTING(op) \
+    ((((PyGC_Head*)op)->_gc_prev & _PyGC_PREV_MASK_COLLECTING) != 0)
 
 PyAPI_FUNC(PyObject *) _PyObject_GC_Malloc(size_t size);
 PyAPI_FUNC(PyObject *) _PyObject_GC_Calloc(size_t size);
--- a/Include/cpython/objimpl.h
+++ b/Include/cpython/objimpl.h
@@ -97,8 +97,7 @@ typedef struct {
 #define _PyGC_SET_FINALIZED(o) \
     _PyGCHead_SET_FINALIZED(_Py_AS_GC(o))
 
-#define _PyObject_GC_IS_COLLECTING(op) \
-    ((((PyGC_Head*)op)->_gc_prev & _PyGC_PREV_MASK_COLLECTING) != 0)
+PyAPI_FUNC(int) _PyObject_GC_IS_COLLECTING(PyObject *op);
 
 PyAPI_FUNC(PyObject *) _PyObject_GC_Malloc(size_t size);
 PyAPI_FUNC(PyObject *) _PyObject_GC_Calloc(size_t size);
--- a/Objects/weakrefobject.c
+++ b/Objects/weakrefobject.c
@@ -1,4 +1,5 @@
 #include "Python.h"
+#include "objimpl.h"
 #include "structmember.h"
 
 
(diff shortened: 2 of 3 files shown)


pablogsal:
I will change the PR to add a warning and fix the comment explaining the problem

Include/cpython/objimpl.h
```diff
@@ -97,6 +97,8 @@ typedef struct {
 #define _PyGC_SET_FINALIZED(o) \
     _PyGCHead_SET_FINALIZED(_Py_AS_GC(o))
 
+#define _PyObject_GC_IS_COLLECTING(op) \
+    ((((PyGC_Head*)op)->_gc_prev & _PyGC_PREV_MASK_COLLECTING) != 0)
```
vstinner: Does this macro expect a PyObject* or a PyGC_Head*? In gc_is_collecting() you pass PyGC_Head*, in handle_callback you pass a PyObject*.

Usually, _PyObject_xxx() functions expects a Python object. I suggest to move PyObject_IS_GC() check into _PyObject_GC_IS_COLLECTING() and convert it to a regular function. You can define the function in gcmodule.c.

Objects/weakrefobject.c
```diff
@@ -874,6 +875,31 @@ PyWeakref_GetObject(PyObject *ref)
...
+     * possible to end in this function if the tp_clear of a function
+     * decrements the references of some internal object that is
+     * weak-referenced, invoking the weak reference callback that will
+     * try to call the function, which is in an incosistent state as
+     * is in the middle of its tp_clear and some internal fields may
+     * be NULL. */
+
+    if (PyObject_IS_GC(callback) && _PyObject_GC_IS_COLLECTING(callback)) {
+        PyErr_WarnEx(PyExc_RuntimeWarning, "A weak reference"
+                " was triying to execute a callback to a function that is being cleared by"
```
Yhg1s: triying -> trying

Objects/weakrefobject.c
```diff
@@ -874,6 +875,31 @@ PyWeakref_GetObject(PyObject *ref)
...
+                " the garbage collector.\n Some C extension class in the dependence"
+                " chain is probably not implementing correctly the garbage collector support.", 1);
```
Yhg1s: "probably not implementing the garbage collector support correctly". (Or, "not correctly implementing".)

Objects/weakrefobject.c
```diff
@@ -874,6 +875,31 @@ PyWeakref_GetObject(PyObject *ref)
...
+     * decrements the references of some internal object that is
+     * weak-referenced, invoking the weak reference callback that will
+     * try to call the function, which is in an incosistent state as
+     * is in the middle of its tp_clear and some internal fields may
+     * be NULL. */
+
+    if (PyObject_IS_GC(callback) && _PyObject_GC_IS_COLLECTING(callback)) {
+        PyErr_WarnEx(PyExc_RuntimeWarning, "A weak reference"
+                " was triying to execute a callback to a function that is being cleared by"
+                " the garbage collector.\n Some C extension class in the dependence"
```
Yhg1s: I would say "A C extension class" rather than "Some C extension class". There is _some_ correlation between developers triggering this warning and having written the faulty C extension class, at least.

Objects/weakrefobject.c
```diff
@@ -874,6 +875,31 @@ PyWeakref_GetObject(PyObject *ref)
 static void
 handle_callback(PyWeakReference *ref, PyObject *callback)
 {
+    /* If the garbage collector support is not properly implemented on
+     * some classes that are involved in a reference cycle, a weak
+     * reference may try to invoke a callback object that is being
+     * cleaned (tp_clear) by the garbage collector and it may be in an
+     * inconsistent state. As the garbage collector explicitly does
+     * not invoke callbacks that are part of the same cycle isolate as
```
Yhg1s: "cycle isolate" doesn't appear as a common term elsewhere in the source (if it is meant to be a term and not a typo). Perhaps rephrase it as "does not invoke callbacks when they are being destroyed at the same time as ..."? Or explain the term somewhere.

Objects/weakrefobject.c
```diff
@@ -874,6 +875,31 @@ PyWeakref_GetObject(PyObject *ref)
...
+     * try to call the function, which is in an incosistent state as
+     * is in the middle of its tp_clear and some internal fields may
+     * be NULL. */
+
+    if (PyObject_IS_GC(callback) && _PyObject_GC_IS_COLLECTING(callback)) {
+        PyErr_WarnEx(PyExc_RuntimeWarning, "A weak reference"
+                " was triying to execute a callback to a function that is being cleared by"
+                " the garbage collector.\n Some C extension class in the dependence"
+                " chain is probably not implementing correctly the garbage collector support.", 1);
+        return;
```
Yhg1s, line 904: What is the actual effect of just returning here? The callback isn't being called, but it wouldn't have been called had all the objects in the cycle had done the right thing, right? So what's the actual value of the warning?

pablogsal: > So what's the actual value of the warning?

The reason we return is to not crash and then we warn that the interpreter has avoided a crash due to a bad implementation but we need to issue the warning so this does not pass silently.

Would you prefer to just crash?

Yhg1s: No, I'm trying to find out if there are _other_ reasons that people should fix their code, besides the specific crash that you're now turning into a warning. If the _only_ reason to fix the code is to avoid this warning, the warning just becomes noise -- especially because very often the warning isn't seen by the developer responsible for the incorrect C code, but some helpless user.

pablogsal: Oh, I understand. Well, after this fix there is no hard crash, but an incorrect implementation of the garbage collector can have undefined consequences. These range from leaks because tp_traverse is not correctly implemented or some other crashes elsewhere.

What do you think we should do here? Maybe a more comprehensive message?

Yhg1s: I'm thinking *less* of a message, really :) The message is useful for people who have control over the C code that does the wrong thing, but to end users of the module. Maybe the exception should be a DeprecationWarning, so it's not shown by default? Perhaps we should discuss this on python-dev or discourse.

pablogsal:
I think I am closing this now that https://github.com/python/cpython/pull/16495 is merged, as I agree is a more resilient solution.

PR #15647, opened 2019-09-02
bpo-38008: Move builtin protocol whitelist to mapping instead of list

jivid:
https://bugs.python.org/issue38008

https://bugs.python.org/issue38008


the-knights-who-say-ni:

Lib/typing.py
```diff
@@ -989,10 +989,21 @@ def _allow_reckless_class_cheks():
         return True
 
 
-_PROTO_WHITELIST = ['Callable', 'Awaitable',
-                    'Iterable', 'Iterator', 'AsyncIterable', 'AsyncIterator',
-                    'Hashable', 'Sized', 'Container', 'Collection', 'Reversible',
-                    'ContextManager', 'AsyncContextManager']
+_PROTO_WHITELIST = {
+    'Callable': 'collections.abc',
```
corona10: I'd like to write code in this way.
`'Callable': collections.abc.__name__,`
It can catch typo error.

corona10 (commented):
Please add news entries by using blurb tool.

https://devguide.python.org/committing/#what-s-new-and-news-entries

Lib/typing.py
ilevkivskyi: Alternatively I would just avoid the repetition by making `_PROTO_WHITELIST` a dictionary mapping module to set of names in it. So that the code below would be
```python
    base.__module__ in _PROTO_WHITELIST and
    base.__name__ in _PROTO_WHITELIST[base.__module__]
```

ilevkivskyi (commented):
Thanks for PR!

I have just one suggestion, could you please add some tests? (and a news entry as @corona10 suggested)

ilevkivskyi (approved):
Thanks, LGTM!

(Also there is a tool called blurb, so that you don't need to create the news items manually.)

PR #15650, opened 2019-09-02
bpo-37953:Improve ForwardRef equality check

hongweipeng:

### DIFF
--- a/Lib/test/test_typing.py
+++ b/Lib/test/test_typing.py
@@ -2353,6 +2353,65 @@ class ForwardRefTests(BaseTestCase):
         self.assertEqual(fr, typing.ForwardRef('int'))
         self.assertNotEqual(List['int'], List[int])
 
+    def test_forward_equality_gth(self):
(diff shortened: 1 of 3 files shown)


hongweipeng:
This PR can solve the recursion problem incidentally [https://bugs.python.org/issue37806](https://bugs.python.org/issue37806). I add some tests for that.

aeros (commented):
Based on @ilevkivskyi's [comment in the related PR](https://github.com/python/cpython/pull/15400#issuecomment-527110764):

> Generally, I would propose to keep only one PR open, because currently the discussion is scattered between two PRs.

Let's hold off on moving forward with this one until we reach a clear consensus on what direction this should be going. Also, it looks like the author of https://github.com/python/cpython/pull/15400 is still interested in working on their PR.

ilevkivskyi:
Closing in favour of the original PR https://github.com/python/cpython/pull/15400 (see my comment there).

PR #15659, opened 2019-09-03
Use standard integer types instead of Python aliases

sir-sigurd:
I'd attribute this to the [old issue](https://bugs.python.org/issue17884), but for some reason, some developers seem to be annoyed by using old issues.


sir-sigurd:
Should we leave this for compatibility?
https://github.com/python/cpython/blob/132acaba5a7f01373ca624b1a5975b190fe866f5/Include/pyport.h#L73-L78

Modules/_sha3/sha3module.c
```diff
@@ -40,18 +40,18 @@
...
   /* opt32 is used for the remaining 32 and 64bit platforms */
   #define KeccakOpt 32
 #endif
 
-#if KeccakOpt == 64 && defined(PY_UINT64_T)
+#if KeccakOpt == 64
   /* 64bit platforms with unsigned int64 */
-  typedef PY_UINT64_T UINT64;
-  typedef unsigned char UINT8;
+  typedef uint64_t UINT64;
```
sir-sigurd, line 53: Probably we can remove these `typedef`s and restore the original ones?

https://github.com/python/cpython/blob/132acaba5a7f01373ca624b1a5975b190fe866f5/Modules/_sha3/kcp/KeccakP-1600-opt64.c#L21-L24

csabella:
> I'd attribute this to the old issue, but for some reason, some developers seem to be annoyed by using old issues.

You can always open a new issue and then reference the old issue in the description.  :-)

sir-sigurd:
Closing this as the same changes were committed in 1a1bd2e23871619adc1405e1cdc7c1e61252fd2c which for some reason is attributed to bpo-40302 :man_shrugging: .

vstinner:
> Closing this as the same changes were committed in 1a1bd2e which for some reason is attributed to bpo-40302 man_shrugging .

It's just that I noticed a few remainaing PY_INT64_T types while working on https://bugs.python.org/issue40302

PR #15662, opened 2019-09-03
bpo-38018: increase code coverage for multiprocessing.shared_memory

vinay0410:
https://bugs.python.org/issue38018

### DIFF
--- a/Lib/test/_test_multiprocessing.py
+++ b/Lib/test/_test_multiprocessing.py
@@ -3719,6 +3719,30 @@ class _TestSharedMemory(BaseTestCase):
         self.assertLess(same_sms.size, 20*sms.size)  # Size was ignored.
         same_sms.close()
 
+        # Creating Shared Memory Segment with -ve size
+        with self.assertRaises(ValueError):
+            shared_memory.SharedMemory(create=True, size=-2)
+
+        # Attaching Shared Memory Segment without a name
+        with self.assertRaises(ValueError):
+            shared_memory.SharedMemory(create=False)
+
+        # Test if shared memory segment is created properly,
+        # when _make_filename returns an existing shared memory segment name
+        with unittest.mock.patch(
+            'multiprocessing.shared_memory._make_filename') as mock_make_filename:
+
+            names = ['test01_fn', 'test02_fn']
+            mock_make_filename.side_effect = names
+            shm1 = shared_memory.SharedMemory(create=True, size=1)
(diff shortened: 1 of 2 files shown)


brandtbucher (approved):

applio (commented):
The first two checks look like good additions.
The use of mock to control the name of the shared memory segments does not look right to me -- have you verified that this test passes?

vinay0410:
@applio , I have verified that the test passes. Also, I have verified that it covers [these](https://github.com/python/cpython/blob/9d607061c9c888913ae2c18543775cf360d55f27/Lib/multiprocessing/shared_memory.py#L94-L95) lines.
I wanted to add a test case which verifies the fact that nothing breaks, if `_make_filename` generates name of already existing segment.
I only could ensure that using mock.

applio (commented):
The test involving mock appears use two distinct names for the two `SharedMemory(create=True, size=1)` invocations.  Crudely adding a `print()` to show `shm1.name` and `shm2.name` seems to confirm this.  Hence the two tests that pass are against `names[0]` and later `names[1]`.

Perhaps I misunderstood but I thought the intent was to verify that `_make_filename` creating duplicate names does not trigger a problem in and of itself.

vinay0410:
@applio, The intent of the test is to verify if a problem is caused when `_make_filename` returns as **existing** name. So, the first call `shm1 = shared_memory.SharedMemory(create=True, size=1)` will create a shared memory with name `test01_fn`, since such a name(shared memory) doesn't exists yet.

Then I reset the side effect function of the mock object by `mock_make_filename.side_effect = names`

So, _make_filename will now again return `test01_fn` on first call. And, `test01_fn` is now a name of existing shared memory segment, which was created by the first call.
This will result in a `FileExistsError` [here](https://github.com/python/cpython/blob/9d607061c9c888913ae2c18543775cf360d55f27/Lib/multiprocessing/shared_memory.py#L94-L95)

And then the loop will continue, calling `_make_filename` again, which will then return `test02_fn`, and since no shared memory with such name exists, it will be created.

So, therefore it verifies existing segment names returned by `_make_filename` will be skipped, and it will be called again.

applio:
Ahhh, thank you for the clarification.  Got it.

pablogsal:
This PR introduced a regression in FreeBSD buildbots:

https://buildbot.python.org/all/#/builders/168/builds/1417

PR #15664, opened 2019-09-03
bpo-38022: IDLE: upgrade help.html to sphinx 2.x HTML5 output

taleinat:
This is done by making `HelpParser` properly handle the slightly different HTML. This required two changes:

1. Avoid additional empty lines for `<p>` tags inside list items (e.g. `<li>`). This is achieved by avoiding starting a new block of text for `<p>` tags if they come after an opening tag.
2. Avoid an extra empty line after `<pre>` tags. This is requried because `<pre>` tags need a newline at the end for them to be displayed properly. Achieved by adding only a single newline if the previous line is empty or entirely white-space.

Also, note the removal of the `class_ != 'first'` check for `<p>` tags. This is because the new HTML output no longer includes "first" classes on tags.

https://bugs.python.org/issue38022

### DIFF
--- a/Lib/idlelib/help.html
+++ b/Lib/idlelib/help.html
@@ -1,11 +1,9 @@
 
-<!DOCTYPE html PUBLIC "-//W3C//DTD XHTML 1.0 Transitional//EN"
-  "http://www.w3.org/TR/xhtml1/DTD/xhtml1-transitional.dtd">
+<!DOCTYPE html>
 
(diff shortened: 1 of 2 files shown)


terryjreedy (approved):
Thank you for this.  I compared new output to 3.8 and the only change is the fix of not double spacing after code (pre) blocks.  The code change is clear.  I will merge and backport.

`# I will follow with a separate no-issue, no-news PR to fix comments with caps and periods.`

PR #15666, opened 2019-09-03
bpo-38059: Using sys.exit() over exit() in inspect.py

alanyee:
Constants added by the site module like exit() "should not be used in programs"
https://docs.python.org/3/library/constants.html#exit

https://bugs.python.org/issue38059


brandtbucher (changes requested):
Thanks, this looks good!

I understand that the change is arguably trivial... but given that this is a code change in a commonly-used module, do you mind creating an issue and NEWS entry for it?

zooba:
Thanks for the PR!

PR #15669, opened 2019-09-03
Fix idlelib.help comments

terryjreedy:
Format comments as sentences and use sentences when possible.

### DIFF
--- a/Lib/idlelib/help.py
+++ b/Lib/idlelib/help.py
@@ -130,12 +130,13 @@ class HelpParser(HTMLParser):
     def handle_endtag(self, tag):
         "Handle endtags in help.html."
         if tag in ['h1', 'h2', 'h3']:
-            self.indent(0)  # Clear tag, reset indent.
+            assert self.level == 0
             if self.show:
                 indent = ('        ' if tag == 'h3' else
                           '    ' if tag == 'h2' else
                           '')
                 self.toc.append((indent+self.header, self.text.index('insert')))
+            self.tags = ''
         elif tag in ['span', 'em']:
             self.chartags = ''
         elif tag == 'a':
@@ -210,7 +211,6 @@ class HelpFrame(Frame):
     "Display html text, scrollbar, and toc."
     def __init__(self, parent, filename):
         Frame.__init__(self, parent)
-        # Keep references to widgets for test access.
         self.text = text = HelpText(self, filename)
         self['background'] = text['background']
         self.toc = toc = self.toc_menu(text)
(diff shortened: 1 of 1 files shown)


terryjreedy:
@taleinat I made minor changes to comment texts and will consider a few more.

I think I found a mismatch between indent() and one of its calls.
```
    def indent(self, amt=1):
        "Change indent (+1, 0, -1) and tags."
        self.level += amt
        self.tags = '' if self.level == 0 else 'l'+str(self.level)
    ...    
        if tag in ['h1', 'h2', 'h3']:
            self.indent(0)  # Clear tag, reset indent.
```
The comment asserts that self.indent(0) will set self.level to 0 and then self.tags to ''.  But it leaves self.level alone and only resets self.tags if self.level was already 0.  As it turns out, the latter is always true, as I would expect it to be if +1s and -1s are properly matched.  So I think the call should be replaced with `self.tags = ''`, possible preceded by `assert self.level == 0`.

taleinat:
> I think I found a mismatch between indent() and one of its calls.

I noticed that too but didn't investigate.

I suggest separating into three methods: `indent()`, `dedent()` and `reset_indent()`.

Lib/idlelib/help.py
```diff
@@ -209,15 +210,15 @@ class HelpFrame(Frame):
...
+        # Keep references to widgets for test access.
         self.text = text = HelpText(self, filename)
         self['background'] = text['background']
         self.toc = toc = self.toc_menu(text)
         self.scroll = scroll = Scrollbar(self, command=text.yview)
         text['yscrollcommand'] = scroll.set
 
         self.rowconfigure(0, weight=1)
-        self.columnconfigure(1, weight=1)  # text
+        self.columnconfigure(1, weight=1)  # Only expand text.
```
taleinat: Perhaps: "Only expand the text widget."

Lib/idlelib/help.py
```diff
@@ -209,15 +210,15 @@ class HelpFrame(Frame):
     "Display html text, scrollbar, and toc."
     def __init__(self, parent, filename):
         Frame.__init__(self, parent)
-        # keep references to widgets for test access.
+        # Keep references to widgets for test access.
```
taleinat: I think this comment should be removed. Keeping references like this is done in most of IDLE's classes, and it serves for more than just enabling testing.

taleinat (commented):
The comment changes look fine to me.

Lib/idlelib/help.py
terryjreedy: Done.

Lib/idlelib/help.py
terryjreedy: Done.  It was written years ago and become more routine since.

terryjreedy:
After more thought, I went with `assert` at the beginning of the block, as I am pretty sure level != 0 would indicate an error somewhere.  The set tags at the end of the block, as done for other blocks.

PR #15677, opened 2019-09-04
Improve clarity of try-return-finally-return

toonarmycaptain:
Clarify execution of try-return-finally-return case.
Per discussion https://discuss.python.org/t/pep-601-forbid-return-break-continue-breaking-out-of-finally

### DIFF
--- a/Doc/tutorial/errors.rst
+++ b/Doc/tutorial/errors.rst
@@ -388,17 +388,17 @@ example::
      File "<stdin>", line 2, in <module>
    KeyboardInterrupt
 
-A *finally clause* is always executed before leaving the :keyword:`try`
-statement, whether an exception has occurred or not. When an exception has
-occurred in the :keyword:`!try` clause and has not been handled by an
-:keyword:`except` clause (or it has occurred in an :keyword:`!except` or
-:keyword:`!else` clause), it is re-raised after the :keyword:`!finally` clause has
-been executed.  The :keyword:`!finally` clause is also executed penultimately
-when the :keyword:`!try` statement is left via a :keyword:`break`,
-:keyword:`continue` or :keyword:`return` statement (thus a :keyword:`return`
-in a :keyword:`finally` clause will execute before and instead of a
-:keyword:`return` statement in a :keyword:`try` clause). For
-example::
+If a :keyword:`finally` clause is present, the :keyword:`finally` clause will execute as the last task before the :keyword:`try` statement completes. The :keyword:`finally` clause runs whether or not the :keyword:try statement produces an exception. The following points discuss more complex cases when an exception occurs:
+
+* If an exception occurs during execution of the :keyword:`!try` clause, the exception may be handled by an :keyword:`except` clause. In all cases, the exception is re-raised after the :keyword:`!finally` clause has been executed.
+
+* An exception could occur during execution of an :keyword:`!except` or :keyword:`!else` clause. Again, the exception is re-raised after the :keyword:`!finally` clause has been executed.
+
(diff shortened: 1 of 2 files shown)


Doc/tutorial/errors.rst
```diff
@@ -393,10 +393,12 @@ statement, whether an exception has occurred or not. When an exception has
 occurred in the :keyword:`!try` clause and has not been handled by an
 :keyword:`except` clause (or it has occurred in an :keyword:`!except` or
 :keyword:`!else` clause), it is re-raised after the :keyword:`finally` clause has
```
epicfaace:
```suggestion
:keyword:`!else` clause), it is re-raised after the :keyword:`!finally` clause has
```

Doc/tutorial/errors.rst
```diff
@@ -393,10 +393,12 @@ statement, whether an exception has occurred or not. When an exception has
...
-been executed.  The :keyword:`!finally` clause is also executed "on the way out"
-when any other clause of the :keyword:`!try` statement is left via a
-:keyword:`break`, :keyword:`continue` or :keyword:`return` statement.  A more
-complicated example::
+been executed.  The :keyword:`!finally` clause is also executed penultimately
```
epicfaace: Reading this for the first time still didn't make this clearer for me. Maybe it would be better to include an example such as https://discuss.python.org/t/pep-601-forbid-return-break-continue-breaking-out-of-finally/2239/3?

Doc/tutorial/errors.rst
```diff
@@ -392,11 +392,24 @@ A *finally clause* is always executed before leaving the :keyword:`try`
 statement, whether an exception has occurred or not. When an exception has
 occurred in the :keyword:`!try` clause and has not been handled by an
 :keyword:`except` clause (or it has occurred in an :keyword:`!except` or
-:keyword:`!else` clause), it is re-raised after the :keyword:`finally` clause has
-been executed.  The :keyword:`!finally` clause is also executed "on the way out"
-when any other clause of the :keyword:`!try` statement is left via a
-:keyword:`break`, :keyword:`continue` or :keyword:`return` statement.  A more
-complicated example::
+:keyword:`!else` clause), it is re-raised after the :keyword:`!finally` clause has
```
willingc: Since this is a tutorial, I would recommend being explicit with a direct tone and minimal jargon. Some suggestions for change of this PR:

If a :keyword:`finally` clause is present, the :keyword:`finally` clause will execute as the last task before the :keyword:`try` statement completes. The :keyword:`finally` clause runs whether or not the :keyword:`try` statement produces an exception. The following points discuss more complex cases when an exception occurs:

* If an exception occurs during execution of the :keyword:`!try` clause, the exception may be handled by an :keyword:`except` clause. In all cases, the exception is re-raised after the :keyword:`:`!finally` clause has been executed.

* An exception could occur during execution of an :keyword:`!except` or :keyword:`!else` clause. Again, the exception is re-raised after the :keyword:`:`!finally` clause has been executed. 

* If the :keyword:`!try` statement reaches a :keyword:`break`, :keyword:`continue` or :keyword:`return` statement, the :keyword:`finally` clause will execute just prior to the :keyword:`break`, :keyword:`continue` or :keyword:`return` statement's execution.

* If a :keyword:`finally` clause includes a :keyword:`return` statement, the :keyword:`finally` clause's  :keyword:`return` statement will execute before, and instead of,  the :keyword:`return` statement in a :keyword:`try` clause.

For example:

willingc (commented):
Hi @toonarmycaptain, thanks for the PR. I have left some suggested wording to help be more explicit and user friendly in tone for this tutorial.

willingc (changes requested):
Thanks and let me know if you need any clarification on the suggested changes.

willingc (approved):
Thanks @toonarmycaptain. Just waiting for CI to pass.

PR #15673, opened 2019-09-04
gh-82039: relax cookiejar.py case-sensitive regex for the inconsequential first line of the cookie file

globalshrug:
https://bugs.python.org/issue37858

* Issue: gh-82039

PR #15679, opened 2019-09-04
[3.8] bpo-37858: relax cookiejar.py case-sensitive regex for the inconsequential first line of the cookie file (GH-15673)

globalshrug:
https://bugs.python.org/issue37858

### DIFF
--- a/Lib/http/cookiejar.py
+++ b/Lib/http/cookiejar.py
@@ -1999,7 +1999,7 @@ class MozillaCookieJar(FileCookieJar):
     header by default (Mozilla can cope with that).
 
     """
-    magic_re = re.compile("#( Netscape)? HTTP Cookie File")
+    magic_re = re.compile("#( Netscape)? HTTP Cookie File", re.IGNORECASE)
     header = """\
 # Netscape HTTP Cookie File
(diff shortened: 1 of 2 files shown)


ned-deily:
Hi, @ashleyharvey.  When proposing a fix like this, you should only open a PR against the master branch, which you did with #15673.  Once it's been reviewed and merged, the core developer handling the merge will decide if it needs to be backported to other branches like 3.8 and 3.7 and we have tools that will normally automatically do the backports.  To avoid further confusion, I'm closing this PR.

ashleyharvey:
> Hi, @ashleyharvey. When proposing a fix like this, you should only open a PR against the master branch, which you did with #15673. Once it's been reviewed and merged, the core developer handling the merge will decide if it needs to be backported to other branches like 3.8 and 3.7 and we have tools that will normally automatically do the backports. To avoid further confusion, I'm closing this PR.

Understood, thank you.  Will it go all the way back to 2.7, or is that up to the core dev?

ned-deily:
That's generally up to the core dev but 2.7 is about to reach end-of-life so we are trying to minimize the number of changes going there.

PR #15680, opened 2019-09-04
[3.7] bpo-37858: relax cookiejar.py case-sensitive regex for the inconsequential first line of the cookie file (GH-15673)

globalshrug:
…ejar.py

(the comment/header); add self to acks.

https://bugs.python.org/issue37858


ned-deily:
Closing this PR awaiting review of #15673 and autobackport from there.

PR #15681, opened 2019-09-04
bpo-38030: Fix os.stat failures on block devices on Windows

zooba:
https://bugs.python.org/issue38030

### DIFF
--- a/Modules/posixmodule.c
+++ b/Modules/posixmodule.c
@@ -1794,13 +1794,13 @@ win32_xstat_impl(const wchar_t *path, struct _Py_stat_struct *result,
             case ERROR_INVALID_PARAMETER:
             case ERROR_INVALID_FUNCTION:
             case ERROR_NOT_SUPPORTED:
-                retval = -1;
+                /* Volumes and physical disks are block devices, e.g.
+                   \\.\C: and \\.\PhysicalDrive0. */
+                memset(result, 0, sizeof(*result));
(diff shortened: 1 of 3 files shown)


Modules/posixmodule.c
```diff
@@ -1827,7 +1827,14 @@ win32_xstat_impl(const wchar_t *path, struct _Py_stat_struct *result,
 
 cleanup:
     if (hFile != INVALID_HANDLE_VALUE) {
-        CloseHandle(hFile);
+        /* Preserve last error if we are failing */
+        error = retval ? GetLastError() : 0;
+        if (!CloseHandle(hFile)) {
+            retval = -1;
+        } else if (retval) {
```
eryksun, line 1834: Have you had a problem with this? The normal path for the internal implementation of `CloseHandle` is to call `status = NtClose(handle)`. If the return status is successful, it just returns `TRUE`. Otherwise Windows sets the last error from the status code (i.e. `RtlNtStatusToDosError`, etc) and returns `FALSE`. It's not going to modify our thread's last error value if it succeeds.

zooba: No, just being overly defensive. There are other functions that (incorrectly) change the last error, and it's easier to be consistent about preserving it than to rely on a not-strictly-followed convention.

zooba: In particular, for others who come in and copy my code without necessarily checking severe edge cases like this :)

PR #15701, opened 2019-09-05
bpo-38037: Fix reference counters in signal module

ghost:
Only 3.8 and 3.9 branches are affected.

https://bugs.python.org/issue38037

### DIFF
--- a/Misc/NEWS.d/next/Library/2019-09-05-22-09-38.bpo-38037.xA66nY.rst
+++ b/Misc/NEWS.d/next/Library/2019-09-05-22-09-38.bpo-38037.xA66nY.rst
@@ -1,1 +1,1 @@
-Fix ref leaks in signal module.
+Fix reference counters in the :mod:`signal` module. 
--- a/Misc/NEWS.d/next/Library/2019-09-05-22-09-38.bpo-38037.xA66nY.rst
+++ b/Misc/NEWS.d/next/Library/2019-09-05-22-09-38.bpo-38037.xA66nY.rst
@@ -1,0 +1,1 @@
+Fix ref leaks in signal module.
--- a/Modules/signalmodule.c
+++ b/Modules/signalmodule.c
@@ -1351,10 +1351,12 @@ PyInit__signal(void)
     d = PyModule_GetDict(m);
 
     x = DefaultHandler = PyLong_FromVoidPtr((void *)SIG_DFL);
+    Py_XINCREF(x);
     if (PyModule_AddObject(m, "SIG_DFL", x))
         goto finally;
 
     x = IgnoreHandler = PyLong_FromVoidPtr((void *)SIG_IGN);
+    Py_XINCREF(x);
     if (PyModule_AddObject(m, "SIG_IGN", x))
         goto finally;
 
--- a/Modules/signalmodule.c
+++ b/Modules/signalmodule.c
@@ -1351,8 +1351,9 @@ PyInit__signal(void)
     d = PyModule_GetDict(m);
 
     DefaultHandler = PyLong_FromVoidPtr((void *)SIG_DFL);
-    if (!DefaultHandler)
+    if (!DefaultHandler) {
         goto finally;
+    }
     Py_INCREF(DefaultHandler);
     if (PyModule_AddObject(m, "SIG_DFL", DefaultHandler)) {
         Py_DECREF(DefaultHandler);
@@ -1360,8 +1361,9 @@ PyInit__signal(void)
     }
 
     IgnoreHandler = PyLong_FromVoidPtr((void *)SIG_IGN);
-    if (!IgnoreHandler)
+    if (!IgnoreHandler) {
         goto finally;
+    }
     Py_INCREF(IgnoreHandler);
     if (PyModule_AddObject(m, "SIG_IGN", IgnoreHandler)) {
         Py_DECREF(IgnoreHandler);
@@ -1578,8 +1580,9 @@ PyInit__signal(void)
 #if defined (HAVE_SETITIMER) || defined (HAVE_GETITIMER)
     ItimerError = PyErr_NewException("signal.ItimerError",
             PyExc_OSError, NULL);
-    if (!ItimerError)
+    if (!ItimerError) {
         goto finally;
+    }
     Py_INCREF(ItimerError);
     if (PyModule_AddObject(m, "ItimerError", ItimerError)) {
         Py_DECREF(ItimerError);
--- a/Modules/signalmodule.c
+++ b/Modules/signalmodule.c
@@ -1351,14 +1351,22 @@ PyInit__signal(void)
     d = PyModule_GetDict(m);
 
     DefaultHandler = PyLong_FromVoidPtr((void *)SIG_DFL);
-    Py_XINCREF(DefaultHandler);
-    if (PyModule_AddObject(m, "SIG_DFL", DefaultHandler))
+    if (!DefaultHandler)
         goto finally;
+    Py_INCREF(DefaultHandler);
+    if (PyModule_AddObject(m, "SIG_DFL", DefaultHandler)) {
+        Py_DECREF(DefaultHandler);
+        goto finally;
+    }
 
     IgnoreHandler = PyLong_FromVoidPtr((void *)SIG_IGN);
-    Py_XINCREF(IgnoreHandler);
-    if (PyModule_AddObject(m, "SIG_IGN", IgnoreHandler))
+    if (!IgnoreHandler)
         goto finally;
+    Py_INCREF(IgnoreHandler);
+    if (PyModule_AddObject(m, "SIG_IGN", IgnoreHandler)) {
+        Py_DECREF(IgnoreHandler);
+        goto finally;
+    }
 
     if (PyModule_AddIntMacro(m, NSIG))
         goto finally;
@@ -1570,9 +1578,13 @@ PyInit__signal(void)
 #if defined (HAVE_SETITIMER) || defined (HAVE_GETITIMER)
     ItimerError = PyErr_NewException("signal.ItimerError",
             PyExc_OSError, NULL);
-    Py_XINCREF(ItimerError);
-    if (PyModule_AddObject(m, "ItimerError", ItimerError))
+    if (!ItimerError)
         goto finally;
+    Py_INCREF(ItimerError);
+    if (PyModule_AddObject(m, "ItimerError", ItimerError)) {
+        Py_DECREF(ItimerError);
+        goto finally;
+    }
 #endif
 
 #ifdef CTRL_C_EVENT
@@ -1618,6 +1630,9 @@ finisignal(void)
     Py_CLEAR(IntHandler);
     Py_CLEAR(DefaultHandler);
     Py_CLEAR(IgnoreHandler);
+#ifdef HAVE_GETITIMER
+    Py_CLEAR(ItimerError);
+#endif
 }
 
 
--- a/Modules/signalmodule.c
+++ b/Modules/signalmodule.c
@@ -1570,6 +1570,7 @@ PyInit__signal(void)
 #if defined (HAVE_SETITIMER) || defined (HAVE_GETITIMER)
     ItimerError = PyErr_NewException("signal.ItimerError",
             PyExc_OSError, NULL);
+    Py_XINCREF(ItimerError);
     if (PyModule_AddObject(m, "ItimerError", ItimerError))
         goto finally;
 #endif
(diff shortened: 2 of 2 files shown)


brandtbucher (changes requested):
I'm not familiar with this file, but it looks like these `Py_XINCREF` calls should be made *after* the calls to `PyModule_AddObject`, correct?

If `PyModule_AddObject` fails, the additional reference isn't stolen... so it isn't needed.

animalize:
> I'm not familiar with this file, but it looks like these `Py_XINCREF` calls should be made _after_ the calls to `PyModule_AddObject`, correct?
> 
> If `PyModule_AddObject` fails, the additional reference isn't stolen... so it isn't needed.

Search `PyModule_AddObject` in CPython code.
It seems always `Py_INCREF` first, then `PyModule_AddObject`.

If `Py_INCREF` after calls, maybe the object is deallocated inside `PyModule_AddObject`?
https://github.com/python/cpython/blob/3.8/Python/modsupport.c#L654

animalize:
No need `needs backport to 3.7` label, only 3.8 and 3.9 branches are affected.

This is a regression caused by commit https://github.com/python/cpython/commit/9541bd321a94f13dc41163a5d7a1a847816fac84 (22 Apr 2019)

brandtbucher:
> Search PyModule_AddObject in CPython code.
> It seems always Py_INCREF first, then call PyModule_AddObject.

`PyModule_AddObject` basically looks like a wrapper around `insertdict` with an added `Py_DECREF` after. I don't see any way that waiting until after the call could end up deallocating the object out from under you. Perhaps those other locations are just `Py_INCREF`'ing first out of convention?

> If `Py_INCREF` after calls, maybe the object is already deallocated inside `PyModule_AddObject` function?
> https://github.com/python/cpython/blob/3.8/Python/modsupport.c#L654

I'm not sure we have to worry about that. The code you link to shows that the object is only `Py_DECREF`'d on success, after a new reference has already been inserted into the module dict by `PyDict_SetItemString`.

If the increase in refcount has to happen before the `PyModule_AddObject` call (and I'm not yet convinced it does), I at least think that we should `Py_XDECREF` in the failure condition (before `goto finally;`).

brandtbucher:
The failure looks like it's due to https://bugs.python.org/issue34037, unrelated.

animalize:
> Ah, yes. I at least think that we should `Py_XDECREF` in the failure condition (before `goto finally;`) though, right?

Search in CPython code, some sites don't check `PyModule_AddObject`'s return value, some sites don't do `Py_DECREF` if `PyModule_AddObject` fails. Looks less normative.

I did all in the last commit.

Modules/signalmodule.c
```diff
@@ -1350,13 +1350,23 @@ PyInit__signal(void)
     /* Add some symbolic constants to the module */
     d = PyModule_GetDict(m);
 
-    x = DefaultHandler = PyLong_FromVoidPtr((void *)SIG_DFL);
-    if (PyModule_AddObject(m, "SIG_DFL", x))
+    DefaultHandler = PyLong_FromVoidPtr((void *)SIG_DFL);
+    if (!DefaultHandler)
```
vstinner: Would you mind to add { ... } to the 3 modified if blocks?

Misc/NEWS.d/next/Library/2019-09-05-22-09-38.bpo-38037.xA66nY.rst
```diff
@@ -0,0 +1 @@
+Fix ref leaks in signal module.
```
vstinner: I suggest:

```suggestion
Fix reference counters in the :mod:`signal` module. 
```

vstinner (approved):
LGTM, but I have a few minor comments.

@nanjekyejoannah: Would you mind to also review this change? I would prefer to have a second reviewer.

brandtbucher (approved):

brandtbucher:
@vstinner It looks like the idiom of calling `PyModule_AddObject` without `Py_DECREF`'ing on the error condition (or even checking for it at all) has spread quite a bit since https://bugs.python.org/issue26868 was first reported. I'll prepare a PR to fix the other call sites!

Modules/signalmodule.c
```diff
@@ -1350,13 +1350,23 @@ PyInit__signal(void)
...
+    }
 
-    x = IgnoreHandler = PyLong_FromVoidPtr((void *)SIG_IGN);
-    if (PyModule_AddObject(m, "SIG_IGN", x))
+    IgnoreHandler = PyLong_FromVoidPtr((void *)SIG_IGN);
+    if (!IgnoreHandler)
         goto finally;
+    Py_INCREF(IgnoreHandler);
+    if (PyModule_AddObject(m, "SIG_IGN", IgnoreHandler)) {
+        Py_DECREF(IgnoreHandler);
```
nanjekyejoannah, line 1369: The  ``Py_INCREF()`` is needed because PyModule_AddObject() PyModule_AddObject leaves the reference counter unchanged as @vstinner said but IIRC, the ``Py_DECREF()`` was called outside the conditionals. Shouldn't we keep it there still?  [See this diff ](https://github.com/python/cpython/commit/9541bd321a94f13dc41163a5d7a1a847816fac84)

Modules/signalmodule.c
```diff
@@ -1568,8 +1578,13 @@ PyInit__signal(void)
 #if defined (HAVE_SETITIMER) || defined (HAVE_GETITIMER)
     ItimerError = PyErr_NewException("signal.ItimerError",
             PyExc_OSError, NULL);
-    if (PyModule_AddObject(m, "ItimerError", ItimerError))
+    if (!ItimerError)
         goto finally;
+    Py_INCREF(ItimerError);
+    if (PyModule_AddObject(m, "ItimerError", ItimerError)) {
+        Py_DECREF(ItimerError);
```
nanjekyejoannah, line 1588: same here.

animalize:
Comments addressed.

animalize:
Should initialize these static variables as below?
If `Py_CLEAR(DefaultHandler)` on an uninitialized `DefaultHandler`, it will `Py_DECREF` an uninitialized object.

```diff
--- a/Modules/signalmodule.c
+++ b/Modules/signalmodule.c
@@ -125,16 +125,16 @@ static volatile struct {
 /* Speed up sigcheck() when none tripped */
 static _Py_atomic_int is_tripped;
 
-static PyObject *DefaultHandler;
-static PyObject *IgnoreHandler;
-static PyObject *IntHandler;
+static PyObject *DefaultHandler = NULL;
+static PyObject *IgnoreHandler = NULL;
+static PyObject *IntHandler = NULL;
 
 #ifdef MS_WINDOWS
 static HANDLE sigint_event = NULL;
 #endif
 
 #ifdef HAVE_GETITIMER
-static PyObject *ItimerError;
+static PyObject *ItimerError = NULL;
 
 /* auxiliary functions for setitimer */
 static int
```

animalize:
If you don't mind, I want to reuse `PyDict_SetItemString()`.
It keeps the code neat, and saves a pair of `Py_INCREF()`/`Py_INCREF()`.

compare
```c
    DefaultHandler = PyLong_FromVoidPtr((void *)SIG_DFL);
    if (!DefaultHandler ||
        PyDict_SetItemString(d, "SIG_DFL", DefaultHandler)) {
        goto finally;
    }
```
to
```c
    DefaultHandler = PyLong_FromVoidPtr((void *)SIG_DFL);
    if (!DefaultHandler) {
        goto finally;
    }
    Py_INCREF(DefaultHandler);
    if (PyModule_AddObject(m, "SIG_DFL", DefaultHandler)) {
        Py_DECREF(DefaultHandler);
        goto finally;
    }
```

nanjekyejoannah (commented):
> If you don't mind, I want to reuse `PyDict_SetItemString()`.
> It keeps the code neat, and saves a pair of `Py_INCREF()`/`Py_DECREF()`.
> 
> compare
> 
> ```c
>     DefaultHandler = PyLong_FromVoidPtr((void *)SIG_DFL);
>     if (!DefaultHandler ||
>         PyDict_SetItemString(d, "SIG_DFL", DefaultHandler) < 0) {
>         goto finally;
>     }
> ```
> 
> to
> 
> ```c
>     DefaultHandler = PyLong_FromVoidPtr((void *)SIG_DFL);
>     if (!DefaultHandler) {
>         goto finally;
>     }
>     Py_INCREF(DefaultHandler);
>     if (PyModule_AddObject(m, "SIG_DFL", DefaultHandler)) {
>         Py_DECREF(DefaultHandler);
>         goto finally;
>     }
> ```

Follow the discussion here : https://bugs.python.org/issue24011 on replacing PyDict_SetItemString() calls with PyModule_AddIntMacro() . By the time I made this change, all the PyModule_AddIntMacro() changes had been made save for this in this commit https://github.com/python/cpython/commit/6782b14bcba612e4a39e41992c77306217b91e30

nanjekyejoannah:
Moving this to a comment : Follow the discussion here : https://bugs.python.org/issue24011 on replacing PyDict_SetItemString() calls with PyModule_AddIntMacro() . By the time I made this change, all the PyModule_AddIntMacro() changes had been made save for this in this commit 6782b14

Modules/signalmodule.c
ghost: `PyModule_AddObject()` function only `Py_DECREF(o)` when the object is successfully added, see:
https://github.com/python/cpython/blob/v3.8.0b4/Python/modsupport.c#L654

So it's correct to stay here, it balances the `Py_INCREF(o)` before `PyModule_AddObject()`.

Other call sites also do it like this:
https://github.com/python/cpython/blob/v3.8.0b4/Modules/_io/_iomodule.c#L678-L685
https://github.com/python/cpython/blob/v3.8.0b4/Modules/_asynciomodule.c#L3392-L3408

animalize:
I still prefer to use `PyDict_SetItemString()`.

IMO it's suitable for this situation, `PyModule_AddObject()` is more suitable for adding an object that doesn't need to be held by other variable.

I have another problem.
In [issue24011](https://bugs.python.org/issue24011)'s `signalmodule.patch` by Christian Heimes, there is a change:
```diff
+  finally:
     if (PyErr_Occurred()) {
         Py_DECREF(m);
         m = NULL;
     }
 
-  finally:
     return m;
 }
```
But this change has not been adopted, is it intentional or negligent? @tiran

vstinner:
I don't think that PyModule_AddIntMacro() can be used in this PR. I like the removal of "x" variable, and reuse DefaultHandler, IgnoreHandler and ItimerError.

vstinner:
> @vstinner It looks like the idiom of calling PyModule_AddObject without Py_DECREF'ing on the error condition (or even checking for it at all) has spread quite a bit since https://bugs.python.org/issue26868 was first reported. I'll prepare a PR to fix the other call sites! I don't want other contributors to think that this is the correct way to use it.

I suggest you to open a new issue at bugs.python.org for that. Before fixing the usage, I suggest to enhance the documentation. Maybe add a short code example:
https://docs.python.org/dev/c-api/module.html?highlight=pymodule_addobject#c.PyModule_AddObject

Modules/signalmodule.c
vstinner: @nanjekyejoannah: Py_DECREF(IgnoreHandler); here is correct.

vstinner:
@nanjekyejoannah: PyModule_AddObject() has terrible API, it's easy to misuse it...

nanjekyejoannah:
@vstinner @animalize agreed.

vstinner:
> @vstinner @animalize agreed.

If the PR looks good to you, would you mind to use GitHub UI to approve the PR?

vstinner:
@animalize: If you want to use PyDict_SetItemString(), please open a separated PR so I can compare the two approaches. Using PyDict_SetItemString() sounds like a good idea: internally, PyModule_AddObject calls PyDict_SetItemString, but in PyInit__signal() we already have have "d" (module dictionary).

nanjekyejoannah:
In terms of fixing the leaks, this LGTM.

nanjekyejoannah (approved):

animalize:
> @animalize: If you want to use PyDict_SetItemString(), please open a separated PR so I can compare the two approaches. Using PyDict_SetItemString() sounds like a good idea: internally, PyModule_AddObject calls PyDict_SetItemString, but in PyInit__signal() we already have have "d" (module dictionary).

Please see this PR, it uses `PyDict_SetItemString`: https://github.com/python/cpython/pull/15753

vstinner:
I merged PR #15753 which is simpler instead.

animalize:
Thanks for your reviews!

brandtbucher:
> I suggest you to open a new issue at bugs.python.org for that. Before fixing the usage, I suggest to enhance the documentation. Maybe add a short code example:
> https://docs.python.org/dev/c-api/module.html?highlight=pymodule_addobject#c.PyModule_AddObject

I'll open an issue about live code changes. I've already got a doc patch at #15725!

PR #15709, opened 2019-09-06
 bpo-38041: Refine IDLE Shell restart lines. 

terryjreedy:
Restart lines now always start with '=' and never end with ' ' and fill the width of the window unless that would require ending with ' ', which could be wrapped by itself and possible confusing the user.

https://bugs.python.org/issue38041

### DIFF
--- a/Lib/idlelib/NEWS.txt
+++ b/Lib/idlelib/NEWS.txt
@@ -3,9 +3,9 @@ Released on 2019-10-20?
 ======================================
 
 
-bpo-38401: Shell restart lines now fill the window width and always
-start with '='.  Filenames still wrap if long relative to the width,
-but there is no longer a trailing space that can wrap by itself.
+bpo-38401: Shell restart lines now fill the window width, always start
+with '=', and avoid wrapping unnecessarily. The line will still wrap
+if the included file name is long relative to the width.
 
 bpo-37092: Add mousewheel scrolling for IDLE module, path, and stack
 browsers.  Patch by George Zhang.
--- a/Lib/idlelib/NEWS.txt
+++ b/Lib/idlelib/NEWS.txt
@@ -3,6 +3,10 @@ Released on 2019-10-20?
 ======================================
 
 
+bpo-38401: Shell restart lines now fill the window width and always
+start with '='.  Filenames still wrap if long relative to the width,
+but there is no longer a trailing space that can wrap by itself.
+
 bpo-37092: Add mousewheel scrolling for IDLE module, path, and stack
 browsers.  Patch by George Zhang.
 
--- a/Lib/idlelib/idle_test/test_pyshell.py
+++ b/Lib/idlelib/idle_test/test_pyshell.py
@@ -9,7 +9,7 @@ from tkinter import Tk
 
 class FunctionTest(unittest.TestCase):
     # Test stand-alone module level non-gui functions.
-    def test_restart_line(self):
+    def test_restart_line_wide(self):
         eq = self.assertEqual
         for file, mul, extra in (('', 22, ''), ('finame', 21, '=')):
             width = 60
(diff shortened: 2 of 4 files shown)


terryjreedy:
Tal, I believe that this is ready and expect to merge it tomorrow, as indicated by the idlelib news entry.  But take a look if you want and have time.

Lib/idlelib/pyshell.py
```diff
@@ -387,6 +387,16 @@ def handle_EOF(self):
         "Override the base class - just re-raise EOFError"
         raise EOFError
 
+def restart_line(width, filename):  # See bpo-38141.
+    if filename == '':
+        filename = 'Shell'
+    tag = f"= RESTART: {filename} ="
+    if width >= 13 + len(filename):  # The length of tag.
```
taleinat:
```suggestion
    if width >= len(tag):
```

Lib/idlelib/pyshell.py
```diff
@@ -387,6 +387,16 @@ def handle_EOF(self):
         "Override the base class - just re-raise EOFError"
         raise EOFError
 
+def restart_line(width, filename):  # See bpo-38141.
```
taleinat, line 390: I know you like short names, but I find this unclear. I suggest "format_restart_line". Otherwise, let's add a doc-string.

Lib/idlelib/pyshell.py
```diff
@@ -387,6 +387,16 @@ def handle_EOF(self):
...
+    if filename == '':
+        filename = 'Shell'
+    tag = f"= RESTART: {filename} ="
+    if width >= 13 + len(filename):  # The length of tag.
+        div, mod = divmod((width -len(tag)), 2)
+        return f"\n{(div+mod)*'='}{tag}{div*'='}"
```
taleinat: For better separation of concerns, IMO this function should return the line without a `'\n'` at the beginning.

Lib/idlelib/NEWS.txt
```diff
@@ -3,6 +3,10 @@ Released on 2019-10-20?
 ======================================
 
 
+bpo-38401: Shell restart lines now fill the window width and always
```
taleinat: I find the second sentence hard to grok without more context, and the part about a trailing space overly detailed and technical. How about:

"Shell restart lines now fill the window width, always start with '=', and avoid wrapping unnecessarily. The line will still wrap if the included file name is long relative to the width."

Lib/idlelib/idle_test/test_pyshell.py
```diff
@@ -7,6 +7,28 @@
...
+class FunctionTest(unittest.TestCase):
+    # Test stand-alone module level non-gui functions.
+
+    def test_restart_line_wide(self):
+        eq = self.assertEqual
+        for file, mul, extra in (('', 22, ''), ('finame', 21, '=')):
+            width = 60
+            bar = mul * '='
+            with self.subTest(file=file, bar=bar):
+                file = file if file else 'Shell'
```
taleinat:
```suggestion
                file = file or 'Shell'
```

taleinat (commented):
I like this change overall. I made a few comments but nothing major.

Lib/idlelib/pyshell.py
taleinat: Also, this might better be a classmethod or staticmethod rather than a module-level function.

terryjreedy: Docstrings for function are a given. I added one giving the specification being implemented.

Lib/idlelib/pyshell.py
terryjreedy: Agreed.  I should not worry about the extra microseconds of 2 writes.  The tests will be clearer also.  

Done, with a manual check that '\n' is still written.

Lib/idlelib/NEWS.txt
terryjreedy: Agreed.  I already reworded in the prototype commit message.  Your version is nice.

terryjreedy:
I am up much later than I should be.  Will finish revisions after sleeping.  Thank you for the improvements.

Lib/idlelib/pyshell.py
terryjreedy: Beside which, '\n' is usually needed to end a current line with an unused '>>> ' prompt.  We could instead have removed the prompt, or for 'Restart Shell', inserted the restart line above the prompt line.   In any case, we will have to reexamine what to do when the sidebar is added (the next thing to review).

Lib/idlelib/NEWS.txt
terryjreedy: Both news items replaced with your improvement.  Thank you for the help polishing the PR.

PR #15710, opened 2019-09-06
bpo-38015: replace inline function `is_small_int` with a macro version

ghost:
There is an unnecessary type casting (to long long), may slightly reduce performance:
```c
static inline int
is_small_int(long long ival)
{
    return -NSMALLNEGINTS <= ival && ival < NSMALLPOSINTS;
}
```

https://bugs.python.org/issue38015


rhettinger (approved):

vstinner:
This change introduced a regression: https://bugs.python.org/issue38205 Python no longer builds without small integer singletons :-(

PR #15615, opened 2019-08-30
bpo-36409: Remove old plistlib API deprecated in 3.4

bigfootjon:
Also cleanup the documentation

https://bugs.python.org/issue36409

PR #15727, opened 2019-09-07
bpo-38053 Update documentation for plistlib

bigfootjon:
* Update "Mac OS X" to "Apple" since plists are used more widely than just macOS
* Re-add the UID class documentation (oops, removed in GH-15615)

https://bugs.python.org/issue38053

### DIFF
--- a/Doc/library/plistlib.rst
+++ b/Doc/library/plistlib.rst
@@ -34,6 +34,9 @@ Values can be strings, integers, floats, booleans, tuples, lists, dictionaries
 (but only with string keys), :class:`bytes`, :class:`bytearray`
 or :class:`datetime.datetime` objects.
 
+.. versionchanged:: 3.4
+   New API, old API deprecated.  Support for binary format plists added.
+
 .. versionchanged:: 3.8
    Support added for reading and writing :class:`UID` tokens in binary plists as used
    by NSKeyedArchiver and NSKeyedUnarchiver.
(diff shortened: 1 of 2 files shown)


Doc/library/plistlib.rst
```diff
@@ -112,16 +106,25 @@ This module defines the following functions:
...
-   .. versionadded:: 3.4
+
+The following classes are available:
+
+.. class:: UID(data)
+
+   Wraps an :class:`int`.  This is used when reading or writing NSKeyedArchiver
+   encoded data, which contains UID (see PList manual).
+
+   It has one attribute, :attr:`data` which can be used to retrieve the int value
```
epicfaace:
```suggestion
   It has one attribute, :attr:`data`, which can be used to retrieve the int value
```

Doc/library/plistlib.rst
```diff
@@ -76,16 +74,12 @@ This module defines the following functions:
    The parser for the binary format raises :exc:`InvalidFileException`
    when the file cannot be parsed.
 
-   .. versionadded:: 3.4
```
shihai1991: why remove this `versionadded`? it's a new api from 3.4 in https://bugs.python.org/file32754/issue-14455-v10.txt. the downstairs `versionadded` have this problem too.

bigfootjon: The `versionadded` tags relate to the deprecation of the old plistlib API after the new one superseded it. Since the old API has been removed (#15615) I thought it would be a good idea to clean up the documentation. Also, Python versions older than 3.5 are not supported, so I figured the tags didn't mean much any way.

I'm happy to re-add all the `versionadded` tags if desired?

Apologies if I'm not following the standard conventions for making changes to the documentation!

ronaldoussoren: I agree, the versionadded directives for the various functions should not be removed.

ronaldoussoren (approved):
LGTM

PR #15135, opened 2019-08-05
Correct typos in the codecs module documentation

geryogam:
This PR will correct typos and punctuation in the [`codecs` module documentation](https://docs.python.org/3/library/codecs.html).

PR #15732, opened 2019-09-08
bpo-38056: overhaul Error Handlers section in codecs documentation

ghost:
* `replace`/`backslashreplace`/`surrogateescape` were wrongly described as encoding only, in fact they can also be used in decoding.
* clarify the description of `surrogatepass`.
* add more descriptions to each handler.
* add two REPL examples.
* add indexes for Error Handler's name.
* add default parameter values in codecs.rst
* improve term "text encoding".

I read the source code, hope there is no mistake and misleading.

The indexes were tested in .chm documentation.

The term "text encoding" appeared 38 times in the entire document, the change in `glossary.rst` should be correct.

Also verified these terms with Wikipedia and Google:
1. XML/HTML numeric character reference
2. Name property from Unicode Character Database

October 2nd effect:

https://bugs.python.org/issue38056


aeros (commented):
Thanks for the PR @animalize.

As someone who is not particularly experienced with using the ``codec`` module (for my purposes, I've never had much of a need for it), these examples are quite helpful for visually representing the purpose of each of the error handlers. 

I'm not entirely certain that we need to provide an example for all four of them, but in general I think that providing an example or two would be very helpful to readers.

aeros167:
If these changes are approved, I think it would be appropriate to backport the changes to the 3.7 and 3.8 docs. Also, I've added a ``skip news`` label. You can add one if you'd like to, but it shouldn't be required for adding examples to the documentation.

animalize:
> I'm not entirely certain that we need to provide an example for all four of them, but in general I think that providing an example or two would be very helpful to readers.

Thanks for your review.
My idea is to demonstrate these things:
* `appropriate XML character reference`
* `backslashed escape sequences`
* `\N{...} escape sequences`

So I would like to include at least these three.

I prefered to use this example, it covers a non-BMP character 🏫.
```python
>>> '42 µ ♪ 🏫'.encode(encoding='ascii', errors='replace')
b'42 ? ? ?'
>>> '42 µ ♪ 🏫'.encode(encoding='ascii', errors='xmlcharrefreplace')
b'42 &#181; &#9834; &#127979;'
>>> '42 µ ♪ 🏫'.encode(encoding='ascii', errors='backslashreplace')
b'42 \\xb5 \\u266a \\U0001f3eb'
>>> '42 µ ♪ 🏫'.encode(encoding='ascii', errors='namereplace')
b'42 \\N{MICRO SIGN} \\N{EIGHTH NOTE} \\N{SCHOOL}'
```

But if paste the statement into the REPL on Windows, the non-BMP character is broken, and get a invalid result `\\ufffd`. Maybe this will confuse the user, so I abandoned this example.
```python
>>> '42 µ ♪ �'.encode(encoding='ascii', errors='backslashreplace')
b'42 \\xb5 \\u266a \\ufffd'
```

aeros167:
Not that I'm opposed to the changes per say, but I think the examples alone would be significantly easier and less controversial to review. Also, it's usually more efficient from a workflow perspective to have more multiple focused PRs than a single large overhaul at once.

animalize:
> 
> 
> Not that I'm opposed to the changes, but I think the examples alone would be significantly easier and less controversial to review. Also, it's usually more efficient from a workflow perspective to have more multiple focused PRs than a single large overhaul at once.

Ah, just like suddenly getting excited for this PR. 😄

There are obvious errors, for example the `backslashreplace` error handler works with decoding since Python 3.5, but it is still in "encode only table".

`replace` error handler also works with both encode and decode codecs.

aeros167:
@animalize:
> If codec experts want to do this, or who thinks he/she can do better than me, free feel to create a new PR.

I'm not recommending for anyone else to do the other parts of the current one, you can still apply the all of the changes. I just mean having only the examples in this PR and then splitting the other sections into their own PRs. As the scope of a PR gets larger, the time required to fully review it prior to a merge significantly increases. From my experience, the core developers tend to prefer smaller and more focused PRs when it's possible.

Doc/library/codecs.rst
```diff
@@ -290,27 +290,42 @@ codec will handle encoding and decoding errors.
 Error Handlers
 ^^^^^^^^^^^^^^
 
-To simplify and standardize error handling,
-codecs may implement different error handling schemes by
-accepting the *errors* string argument.  The following string values are
-defined and implemented by all standard Python codecs:
+To simplify and standardize error handling, codecs may implement different
```
JulienPalard, line 303: Please refrain to rewrap paragraphs that don't need to. Can you put those few lines like they were before?

Doc/library/codecs.rst
```diff
@@ -290,27 +290,42 @@ codec will handle encoding and decoding errors.
...
 +-------------------------+-----------------------------------------------+
 | Value                   | Meaning                                       |
 +=========================+===============================================+
-| ``'strict'``            | Raise :exc:`UnicodeError` (or a subclass);    |
+| ``'strict'``            | Raise :exc:`UnicodeError` (or a subclass),    |
 |                         | this is the default.  Implemented in          |
 |                         | :func:`strict_errors`.                        |
 +-------------------------+-----------------------------------------------+
-| ``'ignore'``            | Ignore the malformed data and continue        |
-|                         | without further notice.  Implemented in       |
```
JulienPalard: Why rewraping this one? Without an important reason, please revert it back to avoid useless diff in history and PRs.

JulienPalard (dismissed):
I'm not sure reorganizing the table is worth it, what did you try to enhance by replacing the separation "all codecs vs text codecs" by the separation "both encode/decode vs encode only"?

Doc/library/codecs.rst
ghost: This is just an intermediate version.
Please see the effect picture at the top of this issue, and the latest commit.
I'm reading `codecs`'s source code, and going to complete this PR in a week or two.

Doc/glossary.rst
```diff
@@ -1044,7 +1044,16 @@ Glossary
       as :keyword:`if`, :keyword:`while` or :keyword:`for`.
 
    text encoding
-      A codec which encodes Unicode strings to bytes.
+      Strings are stored internally as sequences of Unicode code points in
```
JulienPalard: I'd skip over the "internal" things, what do you think of something a bit lighter like:

```
Strings in Python are a sequence of Unicode code points which need to be serialized as a sequence of bytes to be stored or transfered.
```

(keeping 2nd and 3rd paragraphs).

JulienPalard (commented):
LGTM with nits, but not codec expert here.

Doc/glossary.rst
ghost: How about this:
```rst
      String in Python is a sequence of Unicode code points (in range
      ``0x0``--``0x10FFFF``). To store or transfer a string, it need to be
      serialized as a sequence of bytes.
```
As a non-English person, reading a too long sentence may be difficult.
I'd keep the range of Unicode, this Unicode knowledge may help the reader understand the context.

animalize:
> LGTM with nits, but not codec expert here.

I am reading the source code these days, hope there will be no mistakes.

This PR bases on https://github.com/python/cpython/pull/15135.
We need to wait for it to be backported to 3.7/3.8 branches.

Doc/glossary.rst
aeros: A couple of minor grammar fixes for that section:
```
      A string in Python is a sequence of Unicode code points (in range
      ``0x0``--``0x10FFFF``). To store or transfer a string, it needs to be
      serialized as a sequence of bytes.
```

ghost: @aeros167  Thank you very much!

animalize:
It looks fine, I removed `[WIP]` prefix in the title.

Doc/glossary.rst
```diff
@@ -1044,7 +1044,16 @@ Glossary
...
-      A codec which encodes Unicode strings to bytes.
+      A string in Python is a sequence of Unicode code points (in range
+      ``0x0``--``0x10FFFF``). To store or transfer a string, it needs to be
+      serialized as a sequence of bytes.
+
+      Serializing a string into a sequence of bytes is known as "encoding", and
+      recreating the string from the sequence of bytes is known as "decoding".
+
+      There are a variety of different text serialization
+      :ref:`codecs <standard-encodings>`, which are collectivity referred to as
```
aeros: Grammar fix:
```suggestion
      :ref:`codecs <standard-encodings>`, which are collectively referred to as
```

Doc/library/codecs.rst
```diff
@@ -23,10 +23,10 @@
 This module defines base classes for standard Python codecs (encoders and
 decoders) and provides access to the internal Python codec registry, which
 manages the codec and error handling lookup process. Most standard codecs
-are :term:`text encodings <text encoding>`, which encode text to bytes,
-but there are also codecs provided that encode text to text, and bytes to
-bytes. Custom codecs may encode and decode between arbitrary types, but some
-module features are restricted to use specifically with
+are :term:`text encodings <text encoding>`, which encode text to bytes (and
+reverse), but there are also codecs provided that encode text to text, and
```
aeros:
```suggestion
the reverse), but there are also codecs provided that encode text to text, and
```

Doc/library/codecs.rst
```diff
@@ -23,10 +23,10 @@
...
+bytes to bytes. Custom codecs may encode and decode between arbitrary types,
+but some module features are restricted to use specifically with
```
aeros:
```suggestion
but some module features are restricted to be used specifically with
```

Doc/library/codecs.rst
```diff
@@ -23,10 +23,10 @@
...
 :term:`text encodings <text encoding>`, or with codecs that encode to
```
aeros:
```suggestion
 :term:`text encodings <text encoding>` or with codecs that encode to
```

Doc/library/codecs.rst
```diff
@@ -290,58 +290,56 @@ codec will handle encoding and decoding errors.
...
+   pair: backslashreplace; error handler's name
+   pair: surrogateescape; error handler's name
    single: ? (question mark); replacement character
    single: \ (backslash); escape sequence
    single: \x; escape sequence
    single: \u; escape sequence
    single: \U; escape sequence
-   single: \N; escape sequence
+
+The following error handlers can be used with all :ref:`standard-encodings`
```
aeros: Recommended change:
```
The following error handlers can be used with all codecs that are considered
:ref:`standard-encodings`:
```

Doc/library/codecs.rst
```diff
@@ -414,42 +438,58 @@ functions:
...
-   Implements the ``'replace'`` error handling (for :term:`text encodings
-   <text encoding>` only): substitutes ``'?'`` for encoding errors
-   (to be encoded by the codec), and ``'\ufffd'`` (the Unicode replacement
-   character) for decoding errors.
+.. function:: ignore_errors(exception)
 
+   Implements the ``'ignore'`` error handling.
 
-.. function:: ignore_errors(exception)
+   Malformed data is ignored and encoding or decoding is continued without
```
aeros:
```suggestion
   Malformed data is ignored; encoding or decoding is continued without
```

Doc/library/codecs.rst
```diff
@@ -414,42 +438,58 @@ functions:
...
 
+.. function:: replace_errors(exception)
 
-.. function:: xmlcharrefreplace_errors(exception)
+   Implements the ``'replace'`` error handling.
 
-   Implements the ``'xmlcharrefreplace'`` error handling (for encoding with
-   :term:`text encodings <text encoding>` only): the
-   unencodable character is replaced by an appropriate XML character reference.
+   Substitutes ``?`` (ASCII character) for encoding errors, or ``U+FFFD`` (the
```
aeros:
```suggestion
   Substitutes ``?`` (ASCII character) for encoding errors or ``U+FFFD`` (the
```

Doc/library/codecs.rst
```diff
@@ -414,42 +438,58 @@ functions:
...
 
 .. function:: backslashreplace_errors(exception)
 
-   Implements the ``'backslashreplace'`` error handling (for
-   :term:`text encodings <text encoding>` only): malformed data is
-   replaced by a backslashed escape sequence.
+   Implements the ``'backslashreplace'`` error handling.
+
+   Malformed data is replaced by a backslashed escape sequence.
+   On encoding, use hexadecimal form of Unicode code point with formats
```
aeros:
```suggestion
   On encoding, use the hexadecimal form of Unicode code point with formats
```

Doc/library/codecs.rst
```diff
@@ -414,42 +438,58 @@ functions:
...
+   ``\xhh`` ``\uxxxx`` ``\Uxxxxxxxx``. On decoding, use hexadecimal form of
```
aeros:
```suggestion
   ``\xhh`` ``\uxxxx`` ``\Uxxxxxxxx``. On decoding, use the hexadecimal form of
```

Doc/library/codecs.rst
```diff
@@ -414,42 +438,58 @@ functions:
...
+   byte value with format ``\xhh``.
```
aeros, line 481:
```suggestion
   byte values with format ``\xhh``.
```

Doc/library/codecs.rst
```diff
@@ -414,42 +438,58 @@ functions:
...
+
+   .. versionchanged:: 3.5
+      Now works with decoding and translating.
```
aeros:
```suggestion
      Works with decoding and translating.
```

Doc/library/codecs.rst
```diff
@@ -414,42 +438,58 @@ functions:
...
+
+
+.. function:: xmlcharrefreplace_errors(exception)
+
+   Implements the ``'xmlcharrefreplace'`` error handling (for encoding within
+   :term:`text encoding` only).
+
+   The unencodable character is replaced by an appropriate XML/HTML numeric
+   character reference, which is a decimal form of Unicode code point with
+   format ``&#num;``
```
aeros:
```suggestion
   format ``&#num;``.
```

Doc/library/codecs.rst
```diff
@@ -414,42 +438,58 @@ functions:
...
 
 .. function:: namereplace_errors(exception)
 
-   Implements the ``'namereplace'`` error handling (for encoding with
-   :term:`text encodings <text encoding>` only): the
-   unencodable character is replaced by a ``\N{...}`` escape sequence.
+   Implements the ``'namereplace'`` error handling (for encoding within
+   :term:`text encoding` only).
+
+   The unencodable character is replaced by a ``\N{...}`` escape sequence,
```
aeros:
```suggestion
   The unencodable character is replaced by a ``\N{...}`` escape sequence.
```

Doc/library/codecs.rst
```diff
@@ -414,42 +438,58 @@ functions:
...
+   what appears in the brace is the Name property from Unicode Character
```
aeros:
```suggestion
   The sequence that appears in the brace is the *name* property from Unicode Character
```

Misc/NEWS.d/next/Documentation/2019-09-12-08-28-17.bpo-38056.6ktYkc.rst
```diff
@@ -0,0 +1 @@
+Overhaul :ref:`error-handlers` section in :mod:`codecs` module documentation.
```
aeros:
```suggestion
Overhaul the :ref:`error-handlers` documentation in :mod:`codecs`.
```

aeros167:
After applying the above suggestions, also address the wrapping of each section changed to fit within the 80 character line width limit for reST.

Doc/library/codecs.rst
ghost: I changed to:
```rst
The following error handlers can be used with all Python
:ref:`standard-encodings` codecs:
```
Hope it's consistent with [here](https://docs.python.org/3/library/stdtypes.html#str.encode) (the last sentence):
```rst
.. method:: str.encode(encoding="utf-8", errors="strict")

   Return an encoded version of the string as a bytes object. Default encoding
   is ``'utf-8'``. *errors* may be given to set a different error handling scheme.
   The default for *errors* is ``'strict'``, meaning that encoding errors raise
   a :exc:`UnicodeError`. Other possible
   values are ``'ignore'``, ``'replace'``, ``'xmlcharrefreplace'``,
   ``'backslashreplace'`` and any other name registered via
   :func:`codecs.register_error`, see section :ref:`error-handlers`. For a
   list of possible encodings, see section :ref:`standard-encodings`.
```

Doc/library/codecs.rst
ghost: `\xhh` is a single byte's value, should we keep it as is?

Doc/library/codecs.rst
ghost: I omitted it intentionally, so that not to cause visual confusion with the previous `&#num;`

Doc/library/codecs.rst
ghost: I'm afraid the reader may confuse, what is the "sequence" refered to in this paragraph?
`\N{LATIN SMALL LETTER SHARP S}`
or
`LATIN SMALL LETTER SHARP S`

Doc/library/codecs.rst
aeros: @animalize:
> \xhh is a single byte's value, should we keep it as is?

Oh, I see. Yes, this part can be kept as is then.

Doc/library/codecs.rst
aeros: @animalize 
> I'm afraid the reader may confuse, what is the "sequence" refered to in this paragraph?

Hmm, I see your point, what do you think of "The sequence of characters that appear within the braces ..."? This would differentiate it from the escape sequence without being as vague as the original "what appears" .

Doc/library/codecs.rst
aeros: @animalize 
> I omitted it intentionally, so that not to cause visual confusion with the previous &#num;

The visual confusion shouldn't be an issue since you used the code blocks. Could you take a look at the section with the code blocks and the period after using ``make html`` to build the docs? It may look a bit confusing in the plaintext version, but after adding the Sphinx and reST formatting it shouldn't be an issue.

Doc/library/codecs.rst
```diff
@@ -23,11 +23,11 @@
...
-are :term:`text encodings <text encoding>`, which encode text to bytes,
-but there are also codecs provided that encode text to text, and bytes to
-bytes. Custom codecs may encode and decode between arbitrary types, but some
-module features are restricted to use specifically with
-:term:`text encodings <text encoding>`, or with codecs that encode to
+are :term:`text encodings <text encoding>`, which encode text to bytes (and
+the reverse), but there are also codecs provided that encode text to text, and
+bytes to bytes. Custom codecs may encode and decode between arbitrary types,
+but some module features are restricted to be used specifically with
+ :term:`text encodings <text encoding>` or with codecs that encode to
```
aeros: It looks like the indentation on this line is causing the Travis doc build issues:
```
/home/travis/build/python/cpython/Doc/library/codecs.rst:30:Unexpected indentation.
```

```suggestion
:term:`text encodings <text encoding>` or with codecs that encode to
```

ghost: Fixed in local repository.

Doc/library/codecs.rst
ghost: I would like to avoid using "sequence" again, any idea?

Doc/library/codecs.rst
ghost: The background color is not very obvious:


How about add a space before`.`?
```rst
format ``&#num;`` .
```

ghost: effect:

Doc/library/codecs.rst
aeros: Perhaps "The set of characters ..."?

aeros: Also note that "brace" should be "braces", since it is enclosed within a pair of braces. I did not include that in my initial suggestion.

Doc/library/codecs.rst
aeros: LGTM, adding the space looks fine.

Doc/library/codecs.rst
ghost: I'd keep ``Name``, usually ``*name*`` style is used for function parameter in Python documentation.

Add an example to this item.
After adding this example, any advice for "what appears"?


The above table also has these contents, but I hope the sentences in the table keep short.

aeros: @animalize I was suggesting for "The set of characters" to replace "what appears".
 
Here are all of the changes I would suggest for the second paragraph:
```
All unencodeable characters are replaced with a ``\N{..}`` escape sequence. The set of
characters that appear in the braces represent the ``Name`` property from the Unicode
Character Database. For example, the German lowercase letter ... will be converted to
the byte sequence ...
```
Note: Includes a few new suggestions and doesn't include the line wrapping. Also, I omitted a few parts using ellipses (...).

ghost: To be honest, I want the documentation to be simple. 
When people are busy with task, they can get the knowledge at a glance.

Long texts may be more standardized, but add a little cost of understanding.
Maybe the tutorial part needs a more standard language, but the reference part can be simple a bit.

aeros: @animalize:
> To be honest, I want the documentation to be simple.

Is there a particular part of the recommendation that you thought was not phrased in a simple manner? If so, I can provide an alternative with only the grammar corrections. I'd be glad to further work on improving it, that was just a first idea. 

>  Maybe the tutorial part needs a more standard language, but the reference part can be simple a bit.

Based on my experience from the documentation changes and reviews I've made so far, it seems to be the other way around for the most part.

The tutorial tends to be the less formalized part and more simple, while the language reference focuses more on technical accuracy. @JulienPalard and @willingc might be able to provide more accurate information on that though (since they're both documentation experts).

ghost: I mean "The set of characters" is a little longer than "what appears", but this is only a minor problem. 😉

I adopted your suggestion in the commit.

animalize:
Thanks for your reviews, the current state looks fine.

animalize:

aeros (approved):
As far as the general grammar and formatting goes, LGTM. The actual contents of the PR will need to be reviewed by someone experienced with the ``codecs`` module, such as @malemburg or @doerwalter.

animalize:
> The actual contents of the PR will need to be reviewed by someone experienced with the codecs

#### I checked the source code, these statements don't have mistakes:
* “To simplify and standardize error handling, codecs may implement different error handling schemes by accepting the errors string argument:”

* “The following error handlers can be used with all Python Standard Encodings codecs”

* `surrogatepass` error handler's applicable codecs in the table

* function `codecs.register_error()`'s description

#### Also verified these terms with Wikipedia and Google:
* `XML/HTML numeric character reference`

* `Name property from Unicode Character Database`

#### The term "text encoding" appeared 38 times in the entire document, our change in `\Doc\glossary.rst` is correct.

aeros167:
To clarify what I meant by "someone experienced", I was referring to one of the core devs that are considered an [expert for the codecs module](https://devguide.python.org/experts/#stdlib). Although the latest iteration of the changes look correct as far as I can tell, they'll likely be the ones to determine whether or not to merge it. For large stdlib documentation sections being changed, it typically requires final assessment from the respective expert(s) of the module. That's why I @mentioned them.

If they are unable to be reached or aren't able to find the time within a month from the latest commit made to the PR, you can open a new thread in python-dev to see if any other available core developers have enough expertise in the module to review the changes.

animalize:
After reading the [source code](https://github.com/python/cpython/blob/v3.8.0b4/Python/codecs.c#L1130) and some experiments, update the description of `surrogatepass`.

I don't expect this PR to be merged in a short time. 😉

animalize:
If there is enough time, there is no end to improvement. :)

Doc/glossary.rst
```diff
@@ -1044,7 +1044,16 @@ Glossary
       as :keyword:`if`, :keyword:`while` or :keyword:`for`.
 
    text encoding
-      A codec which encodes Unicode strings to bytes.
+      A string in Python is a sequence of Unicode code points (in range
+      ``0x0``--``0x10FFFF``). To store or transfer a string, it needs to be
```
vstinner:
```suggestion
      ``U+0000``--``U+10FFFF``). To store or transfer a string, it needs to be
```

Doc/glossary.rst
```diff
@@ -1044,7 +1044,16 @@ Glossary
...
+      serialized as a sequence of bytes.
```
vstinner, line 1130: Why talking about "serialization" in one way and "recreating [the string]" in the other way? I would prefer to use the terms "encode" and "decode". Here is my definition of an encoding:
https://unicodebook.readthedocs.io/definitions.html#encoding

Doc/library/codecs.rst
```diff
@@ -290,58 +290,56 @@ codec will handle encoding and decoding errors.
...
-|                         | without further notice. Implemented in        |
-|                         | :func:`ignore_errors`.                        |
-+-------------------------+-----------------------------------------------+
-
-The following error handlers are only applicable to
-:term:`text encodings <text encoding>`:
+      >>> 'German ß, ♬'.encode(encoding='ascii', errors='backslashreplace')
+      b'German \\xdf, \\u266c'
+      >>> 'German ß, ♬'.encode(encoding='ascii', errors='xmlcharrefreplace')
+      b'German &#223;, &#9836;'
```
vstinner, line 309: I like the idea of examples. Would you mind to add an example for all available error handlers?

It may be interesting to add an example for surrogatepass which is an uncommon case.

Doc/library/codecs.rst
```diff
@@ -23,11 +23,11 @@
 This module defines base classes for standard Python codecs (encoders and
 decoders) and provides access to the internal Python codec registry, which
 manages the codec and error handling lookup process. Most standard codecs
-are :term:`text encodings <text encoding>`, which encode text to bytes,
-but there are also codecs provided that encode text to text, and bytes to
-bytes. Custom codecs may encode and decode between arbitrary types, but some
-module features are restricted to use specifically with
-:term:`text encodings <text encoding>`, or with codecs that encode to
+are :term:`text encodings <text encoding>`, which encode text to bytes (and
+the reverse), but there are also codecs provided that encode text to text, and
```
vstinner: (and the reverse) => and decode bytes to text

vstinner (commented):
cc @serhiy-storchaka

Doc/glossary.rst
ghost: This explanation was basically copied from here (the first paragraph):
https://docs.python.org/3/library/codecs.html#encodings-and-unicode

Julien Palard suggested not to mention internal details (CPU, memory, endianness) and use the word "store or transfer a string", Kyle Stanley fixed some English grammar issues, then it became like this.

It seems the original author wanted to explain the term "encode"/"decode", and tried to avoid using the word "encode"/"decode" in the explanation.

Moreover, maybe this book can be mentioned in `codecs` or `unicodedata` documentation?

Doc/library/codecs.rst
ghost: Currently there are only two REPL examples, just demonstrate this sentence:
> codecs may implement different error handling schemes by accepting the *errors* string argument

If add example for all available error handlers, will the page become ugly?
all

`surrogatepass` is so uncommon, maybe people who need it know it naturally.
IMHO we can just describe `surrogatepass` clearly. Most readers don't need it, so don't have to be disturbed by an example.

In addition, `surrogatepass` example involves encoding algorithm, I'm afraid the readers will not see the clue from it:

```python
>>> '\uD8AA'.encode(encoding='utf-8', errors='surrogatepass')
b'\xed\xa2\xaa'
```

Doc/library/codecs.rst
ghost: fixed

Doc/glossary.rst
aeros: @animalize 
> It seems the original author wanted to explain the term "encode"/"decode", and tried to avoid using the word "encode"/"decode" in the explanation.

That was my understanding as well. It seems like it would defeat the purpose of having a glossary entry for "text encoding" if the term "encoding" were to be used in the definition itself.

Doc/library/codecs.rst
aeros: @vstinner 
> I like the idea of examples. Would you mind to add an example for all available error handlers?

I'm in agreement with adding an example for anything commonly utilized, but I don't think we should necessarily add one for all of the error handlers. 

> It may be interesting to add an example for surrogatepass which is an uncommon case.

IMO, we should try to focus on having examples for the common cases. Code examples can be very helpful, but in excess they can become distracting to readers.

malemburg:
On 17.10.2019 07:02, Kyle Stanley wrote:
> *@aeros* commented on this pull request.
> 
> ------------------------------------------------------------------------
> 
> In Doc/glossary.rst
> <https://github.com/python/cpython/pull/15732#discussion_r335810307>:
> 
>> @@ -1044,7 +1044,16 @@ Glossary
>        as :keyword:`if`, :keyword:`while` or :keyword:`for`.
>  
>     text encoding
> -      A codec which encodes Unicode strings to bytes.
> +      A string in Python is a sequence of Unicode code points (in range
> +      ``0x0``--``0x10FFFF``). To store or transfer a string, it needs to be
> +      serialized as a sequence of bytes.
> 
>     It seems the original author wanted to explain the term
>     "encode"/"decode", and tried to avoid using the word
>     "encode"/"decode" in the explanation.
> 
> That was my understanding as well. It seems like it would defeat the
> purpose of having a glossary entry for "text encoding" if the term
> "encoding" were to be used in the definition itself.

You have to differentiate between the action of encoding text and
the encoding itself. The above refers to the encoding, which is a
definition of how to translate text into bytes.

-- 
Marc-Andre Lemburg
http://www.malemburg.com/

animalize:
> You have to differentiate between the action of encoding text and the encoding itself. The above refers to the encoding, which is a definition of how to translate text into bytes.

I searched `[Tt]ext\s+[Ee]ncoding` in entire documentation, it seems the change in `glossary.rst` should be correct.

For example, in built-in [open()](https://docs.python.org/3/library/functions.html#open) function doc:

> ...
> *encoding* is the name of the encoding used to decode or encode the file. This should only be used in text mode. The default encoding is platform dependent (whatever locale.getpreferredencoding() returns), but any **text encoding** supported by Python can be used. See the codecs module for the list of supported encodings.
> ...

In [Codec.decode()](https://docs.python.org/3/library/codecs.html#codecs.Codec.decode) function doc:

> ...
> Decodes the object input and returns a tuple (output object, length consumed). For instance, for a **text encoding**, decoding converts a bytes object encoded using a particular character set encoding to a string object.
> ...

In [xml.etree.ElementInclude.default_loader()](https://docs.python.org/3/library/xml.etree.elementtree.html#xml.etree.ElementTree.xml.etree.ElementInclude.default_loader) function doc:

> ...
> *encoding* is an optional **text encoding**. If not given, encoding is utf-8. 
> ...

All places refer to "text encoding", seem talking about the current change.
The `codecs` [documentation](https://docs.python.org/3/library/codecs.html#encodings-and-unicode) also has a definition of "text encoding":
> Strings are stored internally as sequences of code points in range 0x0–0x10FFFF. (See PEP 393 for more details about the implementation.) Once a string object is used outside of CPU and memory, endianness and how these arrays are stored as bytes become an issue. As with other codecs, serialising a string into a sequence of bytes is known as encoding, and recreating the string from the sequence of bytes is known as decoding. There are a variety of different text serialisation codecs, which are collectivity referred to as **text encodings**.
> ...

Doc/library/codecs.rst
willingc: I like the idea of example errors as well. I wonder if the table is making this more cluttered. 

Perhaps something like:

```
Error Name
    Definition
    Example
```
would be more helpful.

Alternatively, a blank line between REPL examples would increase readability.

ghost: It seems you all like the examples, I will try to plan a new layout to suit this idea.

My idea about the current change:

> If an error handleris is very easy to understand, maybe no need to give an example.
IMHO `strict`/`ignore`/`replace` are the cases.
> 
> Or if an error handler is uncommon, maybe the reader doesn't need to be disturbed by an example, we can just use the text to describe it clearly.
> IMHO these are not very common in real code: `namereplace`/`surrogateescape`/`surrogatepass`.
> 
> Then only two remaining. (`backslashreplace`/`xmlcharrefreplace`)
> These two examples also teach some Unicode knowledges imperceptibly: the value of code point, `ß` is not an ASCII character.

aeros: @animalize 
> If an error handleris is very easy to understand, maybe no need to give an example.

I think the simple ones could still benefit from an example, just to show the basics of how it works. Even if it's fairly simple, it may not be quite as easy to understand for someone reading over the codecs documentation for the first time.

> Or if an error handler is uncommon, maybe the reader doesn't need to be disturbed by an example, we can just use the text to describe it clearly.
IMHO these are not very common in real code: namereplace/surrogateescape/surrogatepass.

> Then only two remaining. (backslashreplace/xmlcharrefreplace)
These two examples also teach some Unicode knowledges imperceptibly: the value of code point, ß is not an ASCII character.

:+1:

ghost: > It seems you all like the examples, I will try to plan a new layout to suit this idea.

Sorry, I've been too busy recently, very intense.
I will do this when I have time.

animalize:
>> @animalize on 4 Nov 2019 
>> It seems you all like the examples, I will try to plan a new layout to suit this idea.
>
> @animalize on 8 Dec 2019 
> Sorry, I've been too busy recently, very intense.
> I will do this when I have time.

Could you consider merging this patch?
Although there are not many examples, it was very carefully arranged, it took me almost a month.
I'm a bit afraid that it will be difficult to find out so much time to rearrange a new layout.

aeros:
@animalize With this large of a semantical change to the `encoding` docs, I think it's necessary for @malemburg to have a final look-over; I'll request a review.

Doc/library/codecs.rst
```diff
@@ -361,27 +359,55 @@ The following error handlers are only applicable to
...
 +-------------------+------------------------+-------------------------------------------+
 | Value             | Codecs                 | Meaning                                   |
 +===================+========================+===========================================+
-|``'surrogatepass'``| utf-8, utf-16, utf-32, | Allow encoding and decoding of surrogate  |
-|                   | utf-16-be, utf-16-le,  | codes. These codecs normally treat the    |
-|                   | utf-32-be, utf-32-le   | presence of surrogates as an error.       |
+|``'surrogatepass'``| utf-8, utf-16, utf-32, | Allow encoding and decoding Surrogate code|
+|                   | utf-16-be, utf-16-le,  | point (``U+D800`` - ``U+DFFF``) as normal |
+|                   | utf-32-be, utf-32-le   | code point. Otherwise these codecs treat  |
+|                   |                        | the presence of Surrogate code point in   |
```
JelleZijlstra:
```suggestion
|``'surrogatepass'``| utf-8, utf-16, utf-32, | Allow encoding and decoding surrogate code|
|                   | utf-16-be, utf-16-le,  | point (``U+D800`` - ``U+DFFF``) as normal |
|                   | utf-32-be, utf-32-le   | code point. Otherwise these codecs treat  |
|                   |                        | the presence of surrogate code point in   |
```

https://unicode.org/faq/utf_bom.html#:~:text=A%3A%20Surrogates%20are%20code%20points,DC0016%20to%20DFFF16. doesn't capitalize it.

Doc/library/codecs.rst
```diff
@@ -473,7 +516,7 @@ The base :class:`Codec` class defines these methods which also define the
 function interfaces of the stateless encoder and decoder:
 
 
-.. method:: Codec.encode(input[, errors])
+.. method:: Codec.encode(input[, errors='strict'])
```
JelleZijlstra:
```suggestion
.. method:: Codec.encode(input, errors='strict')
```

If we give the default we don't need the square brackets to say that the param is optional.

Doc/library/codecs.rst
```diff
@@ -491,7 +534,7 @@ function interfaces of the stateless encoder and decoder:
    of the output object type in this situation.
 
 
-.. method:: Codec.decode(input[, errors])
+.. method:: Codec.decode(input[, errors='strict'])
```
JelleZijlstra:
```suggestion
.. method:: Codec.decode(input, errors='strict')
```

Doc/library/codecs.rst
```diff
@@ -558,7 +601,7 @@ define in order to be compatible with the Python codec registry.
    object.
 
 
-   .. method:: encode(object[, final])
+   .. method:: encode(object[, final=False])
```
JelleZijlstra:
```suggestion
   .. method:: encode(object, final=False)
```

Doc/library/codecs.rst
```diff
@@ -615,7 +658,7 @@ define in order to be compatible with the Python codec registry.
    object.
 
 
-   .. method:: decode(object[, final])
+   .. method:: decode(object[, final=False])
```
JelleZijlstra:
```suggestion
   .. method:: decode(object, final=False)
```

Doc/library/codecs.rst
```diff
@@ -749,7 +792,7 @@ compatible with the Python codec registry.
    :func:`register_error`.
 
 
-   .. method:: read([size[, chars, [firstline]]])
+   .. method:: read([size=-1[, chars=-1, [firstline=False]]])
```
JelleZijlstra:
```suggestion
   .. method:: read(size=-1, chars=-1, firstline=False)
```

Doc/library/codecs.rst
```diff
@@ -775,7 +818,7 @@ compatible with the Python codec registry.
       available on the stream, these should be read too.
 
 
-   .. method:: readline([size[, keepends]])
+   .. method:: readline([size=None[, keepends=True]])
```
JelleZijlstra:
```suggestion
   .. method:: readline(size=None, keepends=True)
```

Doc/library/codecs.rst
```diff
@@ -786,7 +829,7 @@ compatible with the Python codec registry.
       returned.
 
 
-   .. method:: readlines([sizehint[, keepends]])
+   .. method:: readlines([sizehint=None[, keepends=True]])
```
JelleZijlstra:
```suggestion
   .. method:: readlines(sizehint=None, keepends=True)
```

JelleZijlstra (commented):
I like this change except for a few small nits. Seems like it ran out of steam after a lot of review activity, but I'm planning to merge it soon now.

JelleZijlstra (approved):

animalize:
Thanks for your reviews, which has led to these revisions being improved in every aspects.

I just didn't add more examples, but I think too many examples will disturb the readers.

PR #15735, opened 2019-09-09
bpo-34037: Fix test_asyncio failure and add loop.shutdown_default_executor()

aeros:
This PR was based heavily upon @vstinner's [PR](https://github.com/python/cpython/pull/13786) and the suggestions from @asvetlov to improve upon it:

>  I support the idea but have a question about implementation.
asyncio provides an interface class for loop called AbstractEventLoop.
I believe that the new property should be reflected there.
The next question is: should it be a property or get_wait_executor_on_close() / set_wait_executor_on_close() pair of methods? Asyncio loop is designed to don't use properties but getters / setters. 

@vstinner's PR was reverted due to it making too drastic of an API change for 3.8 and the concerns from @asvetlov about adding the property without providing public getter and setter methods.

However, the changes proposed still had significant support, and **most importantly**, it fixes a bug which has been causing intermittent failures within the CI tests for ``test_asyncio`` due to dangling threads:
```
test_run_in_executor_cancel (test.test_asyncio.test_events.KqueueEventLoopTests) ...

  Warning -- threading_cleanup() failed to cleanup 1 threads (count: 1, dangling: 2)
  Dangling thread: <Thread(ThreadPoolExecutor-13_0, started daemon 34471091200)>
  Dangling thread: <_MainThread(MainThread, started 34393318400)>
```

Since this is involving changes to the public API, I fully plan on adding the appropriate changes to the documentation. But, I would prefer to wait until I receive feedback from @vstinner and @asvetlov about the implementation itself before doing so since they were both significantly involved in @vstinner's previous PR.

I'll cc the asyncio team as well in case anyone else has feedback to provide.

/cc @python/asyncio 


Misc/NEWS.d/next/Library/2019-09-09-04-43-26.bpo-34037.kX3ZqZ.rst
```diff
@@ -0,0 +1 @@
+For :mod:`asyncio`, add attribute ``asyncio.BaseEventLoop.wait_executor_on_close`` to provide an option for waiting for the executor before closing the event loop. Also adds methods ``loop.get_wait_executor_on_close`` and ``loop.set_wait_executor_on_close``.
```
aeros: I'll fix the number of characters per line and add the appropriate Sphinx roles after I finish adding the documentation changes.

aeros167:
Adding a ``DO-NOT-MERGE`` label until I'm finished with the documentation changes. This PR should not be merged in it's current state, as it only includes the implementation at the moment.

Lib/asyncio/base_events.py
```diff
@@ -1867,3 +1870,13 @@ def set_debug(self, enabled):
...
         if self.is_running():
             self.call_soon_threadsafe(self._set_coroutine_origin_tracking, enabled)
+
+    def get_wait_executor_on_close(self):
+        return self._wait_executor_on_close
+
+    def set_wait_executor_on_close(self, enabled):
+        if not isinstance(enabled, bool):
+            raise TypeError(f'Expected enabled to be boolean,'
+                            f'got {enabled!r}')
```
vstinner: please add a space before "got"

Misc/NEWS.d/next/Library/2019-09-09-04-43-26.bpo-34037.kX3ZqZ.rst
vstinner: it's not an attribute but 2 new methods.

vstinner (dismissed):
You have to update the documentation:

* Document the 2 new event loop methods, around https://docs.python.org/dev/library/asyncio-eventloop.html
* Document the behavior change in What's New in Python 3.9, around: https://docs.python.org/dev/whatsnew/3.9.html#changes-in-the-python-api

Misc/NEWS.d/next/Library/2019-09-09-04-43-26.bpo-34037.kX3ZqZ.rst
aeros: @vstinner:
> it's not an attribute but 2 new methods.

When I initially created the news entry, the ``wait_executor_on_close`` attribute/property (within Python, I believe property would probably be the more correct term) was public, as it was in your PR. But then I realized that it should probably be private since the getter and setter methods were created. 

I'll update it accordingly.

aeros167:
Thanks for the feedback @vstinner.

> You have to update the documentation

Yep, I was definitely planning on doing so within this PR. I just wanted to make sure the implementation details were satisfactory first, so that I wouldn't have to add the documentation changes and then completely rewrite them from different specifications. I'll start working on that next.

I appreciate that you specified the locations. I figured the documentation for the methods would be on asyncio event loop page, but I wasn't certain where to place the changes in What's New (or whether that should be done within this PR). This will be my first time adding an entry in there, as the majority of my contributions thus far have been PR reviews or documentation changes. 

I have to admit, adding my own What's New entry for 3.9 is both a little intimidating and exciting. (:

1st1 (changes requested):
I still don't like this API and I'd suggest to implement it the other way around. We have a very similar problem with asynchronous generators -- closing them isn't an easy process and can be a lengthy blocking operation.  The way we do that is via the `loop.shutdown_asyncgens()` method, that users are supposed to call *manually* before `loop.close()`.  This is a low-level API, and the expectation is that `asyncio.run()` function would do that for you.

Therefore I propose the following:

* Add a new `loop.shutdown_threadpool()` method.  Just like with `shutdown_asyncgens()` -- it would be invalid to call `loop.run_in_executer()` after `loop.shutdown_threadpool()` is called.

* Make `asyncio.run()` to call the new `loop.shutdown_threadpool()`.

1st1:
@aeros167 Kyle, do you have time to work on this issue this week?

aeros167:
@1st1:
> @aeros167 Kyle, do you have time to work on this issue this week?

Yeah, I definitely should have time to work on it. Would you recommend making the changes directly to this PR or opening a separate one? 

I think your proposal would be far more robust and provide a significantly better user experience. But this change might be a decent fallback, mainly for the purpose of fixing the CI tests in case there are unforeseen complications with properly implementing the proposed ``loop.shutdown_threadpool()``. The intermittent failures for ``test_asyncio`` are quite problematic.

I would be glad to help with working on implementing it though. I've been interested in becoming more involved with improving the CI tests and asyncio development in general, and I think this process will provide a great introduction to both. Thanks for the in-depth feedback.

asvetlov:
I support @1st1 proposal: `loop.shutdown_threadpool()` is more robust than a flag (plus we have the similar `shutdown_asyncgens()` already.

1st1:
> But this change might be a decent fallback, mainly for the purpose of fixing the CI tests in case there are unforeseen complications with properly implementing the proposed loop.shutdown_threadpool(). The intermittent failures for test_asyncio are quite problematic.

Well, it's a bit too late to add new APIs in the 3.8 branch, no matter if they are public or private.  I suggest to go ahead with the new idea so that we can fix the CI for the master branch.

FWIW @vstinner suggested the `shutdown_default_executor()` name in the bpo, which I think is a better name indeed.

aeros167:
@1st1 

My current plan:

1) Add boolean flag ``self._executor_shutdown_called`` to BaseEventLoop. This would be used in ``loop.run_in_executer()`` to check whether or not the executor shutdown has been called.  

2) Create method ``loop.shutdown_default_executor()`` (both for AbstractEventLoop and BaseEventLoop). Implementation:

```python
def shutdown_default_executor(self, wait=True):
    executor = self._default_executor
    # Nothing to shut down if there is no executor
    if executor is None: 
        return
    # Used by loop.run_in_executer() to check if the the executor has been terminated
    self._executor_shutdown_called = True 
    executor.shutdown(wait) # alternatively, wait=wait, but that seems redundant
```

3) Modify ``loop.run_in_executer()`` to raise a RuntimeError if ``self._executor_shutdown_called`` is True. This would be done through an internal ``loop._check_executor_shutdown()`` method, called within ``loop.run_in_executer()``, similar to the existing ``loop._check_closed``. Implementation:

```python
def _check_executor_shutdown(self):
    if self._executor_shutdown_called:
        raise RuntimeError("Executor has been shut down")
```

4) Modify ``asyncio.run()`` to use ``loop._check_executor_shutdown()``. This is the part that I'm somewhat unclear about. Since ``asyncio.run()`` is already calling ``loop.close()``, can we just modify ``loop.close()`` to call the new ``loop.shutdown_default_executor()``? Implementation:

```python
def close(self):
    ...
    self._closed = True
    self._ready.clear()
    self._scheduled.clear()
#    executor = self._default_executor
#    if executor is not None:
#        self._default_executor = None
#        executor.shutdown(wait=False)
#        executor.shutdown(wait=self._wait_executor_on_close)
    self.shutdown_default_executor() # default is True
```

In the above implementation, users could directly utilize ``loop.shutdown_default_executor()`` instead of ``loop.close()`` if they don't want to wait for executor to finish joining threads before shutting down. From my understanding, ``wait=True`` fits the majority of use cases and would be the correct choice to use in ``asyncio.run()``. 

If further customization is required (or there's issues with the tests timing out in the future), we could add an optional argument *thread_timeout* to ``executor.shutdown``. If wait is True, this would assign a maximum time to live for each of the threads joined in ``executor._threads``.

5) Update the documentation. Similar to before, I would like to do this step last so that I don't have to rewrite the documentation changes if a different implementation is desired.

If the above is satisfactory, +1 this comment, otherwise let me know if anything is incorrect.

Note: Several of the comments in the example implementations above are for the purpose of explaining my thought process, and would not be included in the PR. Particularly the inline comments.

aeros167:
I'll change the PR title as well after I commit the changes to the PR branch. I've got some college work to do for now, but I should be able to work on that later tonight or tomorrow.

asvetlov:
1. Assign `self._executor_shutdown_called = True` unconditionally.
2. Don't call `shutdown_default_executor()` from `close()`; left it for user as we do for `shutdown_asyncgens()`

aeros167:
@asvetlov:
> Assign ``self._executor_shutdown_called = True`` unconditionally.

That would make more sense, especially since this is done in ``self._asyncgens_shutdown_called``. The status of the flag should change on call to the method rather than only when there's a default executor set.

> Don't call ``shutdown_default_executor()`` from ``close()``; left it for user as we do for ``shutdown_asyncgens()``

Hmm, okay. It just seemed odd to me to leave in the part in the last 4 lines of ``close()``, since ``shutdown_default_executor()`` is effectively a more robust version of that. Should those lines be removed from ``close()`` or remain as is?

I'm not certain where the best place to call ``shutdown_default_executor()`` would be from within ``asyncio.run()``. Perhaps like this?

```python
    ...
    loop = events.new_event_loop()
    try:
        events.set_event_loop(loop)
        loop.set_debug(debug)
        return loop.run_until_complete(main)
    finally:
        try:
            _cancel_all_tasks(loop)
            loop.run_until_complete(loop.shutdown_asyncgens())
            loop.shutdown_default_executor()
        finally:
            events.set_event_loop(None)
            loop.close()
```

Thanks for the feedback, and I greatly appreciate that you guys have allowed me to help with working on this issue. I've been learning a lot about the internals of asyncio during this process. (:

asvetlov:
I think your last comment is correct.
Would be nice to get @1st1 feedback as well

aeros167:
Updating the PR title to better reflect the changes. I'll update the PR itself after receiving feedback from Yury.

aeros167:
I just realized that I forgot to set ``self._default_executor = None`` within my example for ``shutdown_default_executor``. I'll update it accordingly and include Andrew's suggestion regarding the flag as well.

1st1:
> Hmm, okay. It just seemed odd to me to leave in the part in the last 4 lines of close(), since shutdown_default_executor() is effectively a more robust version of that. Should those lines be removed from close() or remain as is?

They should remain as-is.

The reasoning:

1. We must not change the behaviour of the `loop.close()` method.

2. Right now `loop.close()` shuts down the executor without waiting; keep it that way.

3. The new `loop.shutdown_default_executor()` should always wait.  Drop the `wait` parameter.  The reason for this function is to let users explicitly shut executor down *and wait* until it's shutdown.

The `asyncio.run()` function should call `loop.shutdown_default_executor()` *after* `loop.shutdown_asyncgens()`.

Also, I'd make `loop.shutdown_default_executor()` a coroutine, even though we don't need that now.  That would give us room to add additional functionality that might involve awaiting on something, and will make the new function symmetric with `loop.shutdown_asyncgens()`.

asvetlov:
> Also, I'd make `loop.shutdown_default_executor()` a coroutine, even though we don't need that now. That would give us room to add additional functionality that might involve awaiting on something, and will make the new function symmetric with `loop.shutdown_asyncgens()`.

@1st1 I understand the motivation but curious how it can be done in practice. Waiting is a synchronous call. 
`loop.shutdown_default_executor()` can spawn a thread and wait a future that is set from the thread. Doesn't it look overcomplicated?

1st1:
> loop.shutdown_default_executor() can spawn a thread and wait a future that is set from the thread. Doesn't it look overcomplicated?

I'd just keep it simple and let the coroutine block.  We can implement the "wait in a thread" solution later if we need to.  Do you think this would be a bad idea?

aeros167:
@1st1:
> Also, I'd make loop.shutdown_default_executor() a coroutine, even though we don't need that now.

I was actually thinking about this when adding ``shutdown_default_executor()`` to ``asyncio.run()`` in the example implementation, the asymmetry with ``shutdown_asyncgens()`` did seem slightly off. 

But, I wasn't sure what this would involve. Can we just directly convert the method into a coroutine without changing the implementation details to implement this in the most simple manner? I wasn't certain if this would work properly with ``executor.shutdown()`` without introducing new potential issues.

>  We can implement the "wait in a thread" solution later if we need to.

Would this involve awaiting on a ``tasks.gather()`` on a list created from ``executor._threads`` and joining each thread within ``loop.shutdown_default_executor()`` instead of using ``executor.shutdown()``? That seems like it would be quite a bit more involved. 

For now, I'll start working on updating this PR to include the changes without turning ``loop.shutdown_default_executor()`` into a coroutine. I would glad to try helping with that as well, but that should probably be done in a separate PR.

Thanks for the detailed feedback.

asvetlov:
> I'd just keep it simple and let the coroutine block. We can implement the "wait in a thread" solution > later if we need to. Do you think this would be a bad idea?

I think a blocking coroutine is very confusing, it will produce `'Executing %s took %.3f seconds'` warning in debug mode.

asvetlov:
@aeros167 yes, please go ahead with non-coroutine version, it makes a value in itself.

aeros167:
The recent batch of commits were the full code changes. The only thing I changed from the examples and from the suggestions was the name of the internal method to check if the default executor was closed. When I was looking into the best place to use it for ``loop.run_in_executor()``, I realized that the name ``loop._check_executor_shutdown()`` might be misleading, as it could imply that the method is checking for any closed executor. So, I renamed it to ``_check_default_executor()``.

It might be useful at some point in the future to consider adding a public method called ``loop.check_executor()`` or ``loop.check_executor_shutdown()`` that can take any executor as an argument and checks to see if executor.shutdown has been called (by checking the status of the flag ``executor._shutdown`` within the ``ThreadPoolExecutor`` class. This could be quite useful for ``loop.run_in_executor`` to ensure that the executor shutdown has not been called when the user passes in their own executor. The current implementation in this PR will only check the default executor.

aeros167:
Next, I'll work on adding the documentation changes, the Misc/NEWS entry, and the What's New 3.9 entry. I just wanted to give you guys a chance to look at the implementation details first.

Lib/asyncio/base_events.py
```diff
@@ -404,8 +404,11 @@ def __init__(self):
         # A weak set of all asynchronous generators that are
         # being iterated by the loop.
         self._asyncgens = weakref.WeakSet()
+
```
1st1: Please remove this empty line; it wasn't there to visually group all stuff related to asyncgens.

Lib/asyncio/base_events.py
```diff
@@ -543,6 +550,16 @@ def _asyncgen_firstiter_hook(self, agen):
                     'asyncgen': agen
                 })
 
+    def shutdown_default_executor(self):
+        """Shutdown the default executor, but wait for the threads
```
1st1: You need to fit the first sentence into one line.

Lib/asyncio/events.py
```diff
@@ -249,6 +249,12 @@ def close(self):
         """Shutdown all active asynchronous generators."""
         raise NotImplementedError
 
+    def shutdown_default_executor(self):
+        """Shutdown the default executor, but wait for the threads
+        in the threadpool to finish joining.
+        """
```
1st1: ditto

Lib/asyncio/base_events.py
1st1: To avoid code duplication, you can add a `def _shutdown_default_executor(self, wait):` and call it from `loop.close()` and `loop. shutdown_default_executor`

Lib/asyncio/base_events.py
aeros: Oh okay, I just thought it would be useful to group the shutdown flags visually. I'll remove it.

Lib/asyncio/base_events.py
aeros: @1st1:
> You need to fit the first sentence into one line.

I think that would exceed the recommended 80 char limit per line, would you like me to fit the max words within the limit on each line or reduce the docstring sentence? I can make it a bit shorter:

> """Shutdown the default executor, but wait for the threadpool to finish."""

Lib/asyncio/events.py
aeros: See above comment

Lib/asyncio/base_events.py
1st1: Yeah, sorry for the nitpicking :)

Lib/asyncio/base_events.py
1st1: > """Shutdown the default executor, but wait for the threadpool to finish."""

That works.  In general, we try to use this rule: keep the first sentence short and on one line.  Add another paragraph if needed to provide more details.

A lot of tools that work with docstrings/etc like when the first line is complete.

Lib/asyncio/base_events.py
aeros: @1st1:
> Yeah, sorry for the nitpicking :)

No problem, I don't mind nitpicking. I can definitely understand why you're tight on the formatting changes, it helps to keep the code more organized. (:

Lib/asyncio/base_events.py
aeros: Ah, good to know. Thanks.

1st1 (approved):

1st1:
@vstinner Victor, would you be OK if we go ahead with this?

asvetlov (approved):
LGTM
Should the following PR make `shutdown_default_executor()` an async method?

asvetlov:
News entry is required for the change

aeros167:
@asvetlov 
> News entry is required for the change

Yep, I'm currently working on the documentation changes, Misc/NEWS entry, and What's New for 3.9. I figured it would help to expedite the process if I pushed the batch of commits containing the implementation details first, to give you all a chance to review the code changes.

aeros167:
@asvetlov:
> Should the following PR make shutdown_default_executor() an async method?

I would happy to work on that next, unless you or Yury would prefer to. (: 

This PR has been an especially enjoyable experience for me, I've learned a lot in the last few days about asyncio and am looking forward to working on it more in the future as well.

1st1:
> Should the following PR make shutdown_default_executor() an async method?

If we want that them I'm strongly in favor of making it async now (even though it will be blocking) before we merge this.

aeros167:
@1st1:
> If we want that them I'm strongly in favor of making it async now (even though it will be blocking) before we merge this.

I'm good with either option, but an argument in favor of adding it for this PR would be that we don't have to update the documentation twice. I'm currently adding ``loop.shutdown_default_executor()`` to the docs in using the ``.. method::`` directive, and it will have to be changed to ``.. coroutinemethod::`` when changed to a coroutine (not to mention the wording of the section). 

The downside would be that it would slightly increase the amount of time that we're getting the intermittent ``test_asyncio`` failures and have a chance of introducing new complications.

asvetlov:
Ok, let's make async. It looks slightly better in my eyes.
@aeros167 the threading trick can look like:
```
async def shutdown_default_executor(self):
    fut = self.create_future()
    th = threading.Thread(func=self._do_shutdown, args=(fut,))
    th.start()
    await fut
    th.join()

def _do_shutdown(self, fut):
   try:
       self._executor.shutdown(wait=True)
       self.call_soon_threadsafe(fut.set_result, None)
   except Exception as ex:
       self.call_soon_threadsafe(fut.set_exception, ex) 
```

The real code is slightly more complex but you've got my idea.

1st1:
> The real code is slightly more complex but you've got my idea.

What's would that extra complexity do?  I mean your code looks fine to me as it is :)

1st1:
> This function runs the passed coroutine, taking care of managing the asyncio event loop, finalizing asynchronous generators, and closing the threadpool.

That one works.  I'd say "closing the default threadpool."

aeros167:
@1st1 
> What's would that extra complexity do? I mean your code looks fine to me as it is :)

I think @asvetlov was referring to me understanding the idea of it, but having some extra complexity beyond what I was stating. That would be correct. I got it for the most part, but didn't have an exact idea of what it would look like.

aeros167:
Also, with @asvetlov's coroutine implementation, that means we need to undo the change to ``loop.close()``, correct? From my understanding ``loop.close()`` should still use ``executor.shutdown(wait=False)``.

asvetlov:
correct

aeros167:
>  th = threading.Thread(func=self._do_shutdown, args=(fut,))

@asvetlov Unless I'm mistaken, shouldn't the named argument being passed to ``threading.Thread`` be ``target`` instead of ``func``?

>  th = threading.Thread(target=self._do_shutdown, args=(fut,))

I've seen the two terms used interchangeably in similar parameters (since they effectively mean the same thing), but based on the docs, ``threading.Thread`` uses ``target``: https://docs.python.org/3/library/threading.html#threading.Thread.

aeros167:
Also, adding ``self._executor_shutdown_called = True`` to the first line for `loop.shutdown_default_executor``.

asvetlov:
@aeros167 sorry, it is `target` of course. My memory is weak sometimes.

Also, `th.join()` should be in finally block:
```
async def shutdown_default_executor(self):
    fut = self.create_future()
    th = threading.Thread(func=self._do_shutdown, args=(fut,))
    th.start()
    try:
        await fut
    finally:
        th.join()
```

aeros167:
> @aeros167 sorry, it is target of course. My memory is weak sometimes.

No problem, I would have probably done the same if I hadn't read the docs for threading recently. I very much appreciate the help. (:

aeros167:
>        self._executor.shutdown(wait=True)

I believe that should be ``self._default_executor``. Also, based on some of the errors I'm receiving  ``test_asyncio``, we'll have to add back that conditional check at the start to make sure that ``self._default_executor`` isn't None. The error was the good old "Nonetype has no method x" (x being shutdown in this case). There were a total of 8 exceptions initially, but after making a few changes I've resolved all of them and test_asyncio is passing again. 

I'll pushed the latest commits I've made to the PR branch. The only thing the current version is missing is the What's New 3.9 addition and the Misc/NEWS entry, I believe that I've made all of the appropriate documentation changes though. Let me know if anything else is missing or incorrect.

Lib/asyncio/runners.py
```diff
@@ -45,6 +45,7 @@ def run(main, *, debug=False):
         try:
             _cancel_all_tasks(loop)
             loop.run_until_complete(loop.shutdown_asyncgens())
+            loop.run_until_complete(loop.shutdown_default_executor())
```
aeros, line 48: Since this is now a coroutine, it should be called with ``loop.run_until_complete()`` instead of by itself as it was previously.

Lib/asyncio/base_events.py
```diff
@@ -543,6 +549,26 @@ def _asyncgen_firstiter_hook(self, agen):
                     'asyncgen': agen
                 })
 
+    async def shutdown_default_executor(self):
+        """Schedule the shutdown of the default executor."""
+        self._executor_shutdown_called = True
+        if self._default_executor is None:
```
aeros, line 555: This check is needed, otherwise numerous exceptions occur from ``test_asyncio``. There's nothing to shutdown if the default executor isn't set.

Lib/asyncio/events.py
```diff
@@ -249,6 +249,10 @@ def close(self):
         """Shutdown all active asynchronous generators."""
         raise NotImplementedError
 
+    async def shutdown_default_executor(self):
+        """Schedule the shutdown of the default executor."""
```
aeros, line 253: I changed the docstrings to reflect that this method is now a coroutine. There might be a better way to word this, but I had trouble thinking of something more descriptive that didn't go beyond a single line. Let me know if the previous one is preferable over this one.

Doc/library/asyncio-task.rst
```diff
@@ -229,6 +229,8 @@ Running an asyncio Program
        **Important:** this function has been added to asyncio in
        Python 3.7 on a :term:`provisional basis <provisional api>`.
```
aeros, line 230: @1st1 Is ``asyncio.run()`` still going to be considered provisional in 3.9?

aeros167:
Since the scope of these changes are rather large, should the "What's New" entry be under an asyncio section for "Improved Modules" or "Changes in the Python API"? I'm not overly familiar with the appropriate usage of the different "What's New" sections. 

The initial version of this PR which just added a single and fairly straight forward class method seemed more appropriate to go under "Changes in the Python API". But since the latest version of the PR is adding a new coroutine that significantly alters ``asyncio.run()``, I was thinking that it might be better placed under "Improved Modules".

Lib/asyncio/runners.py
asvetlov: That's fine

Lib/asyncio/base_events.py
```diff
@@ -406,6 +406,8 @@ def __init__(self):
         self._asyncgens = weakref.WeakSet()
         # Set to True when `loop.shutdown_asyncgens` is called.
         self._asyncgens_shutdown_called = False
+        # Set to True when `loop.shutdown_default_executor` is called.
+        self._executor_shutdown_called = False
```
asvetlov, line 410: You don't need the flag.
Assign `self._default_executor = None` after shutdown to make `shutdown_default_executor()` idempotent, that's it.

Doc/library/asyncio-task.rst
asvetlov: Surely no. 
It should be fixed in a separate PR.

Lib/asyncio/base_events.py
```diff
@@ -768,6 +795,8 @@ def run_in_executor(self, executor, func, *args):
             self._check_callback(func, 'run_in_executor')
         if executor is None:
             executor = self._default_executor
+            # Only check when the default executor is being used
+            self._check_default_executor()
```
asvetlov, line 799: I doubt if we need the check.
`shutdown_default_executor()` is called at the very end of event loop lifecycle, all tasks are stopped and killed at this point.
There is no code that can call `run_in_executor()` at the moment.

Lib/asyncio/events.py
asvetlov: The docstring is fine

Lib/asyncio/base_events.py
aeros: @asvetlov 
> I doubt if we need the check.

> You don't need the flag.

Okay, I'll try removing both of them, running the tests again, and observing what happens. I think the main thing that might have been causing the earlier issues was the lack of a ``if executor is None`` check for ``shutdown_default_executor()``, which was addressed.

aeros: @asvetlov 
>  shutdown_default_executor() is called at the very end of event loop lifecycle, all tasks are stopped and killed at this point.
There is no code that can call run_in_executor() at the moment.

I think the concern earlier from @1st1 was users directly calling ``shutdown_default_executor()`` and then trying to call ``run_in_executor()`` at a later point:

>  Just like with shutdown_asyncgens() -- it would be invalid to call loop.run_in_executer() after loop.shutdown_threadpool() is called.

(this was before the rename was suggested)

Does this no longer apply?

asvetlov: Well, makes sense. Sorry

asvetlov (approved):
LGTM

Lib/asyncio/base_events.py
aeros: @asvetlov 
> Well, makes sense. Sorry

No problem. If we ever phase out the lower level API from the public in the distant future and make it internal, it might be worth removing the flag and check, but as long as ``run_in_executor()`` is part of the public API, it probably needs to remain.

Lib/asyncio/base_events.py
aeros: Resolving this comment based on my other comment: https://github.com/python/cpython/pull/15735#discussion_r323894385. Feel free to reopen it if you think this still needs to be addressed.

Doc/library/asyncio-task.rst
aeros: I'll leave this unresolved so it's easier to find later.

Doc/library/asyncio-eventloop.rst
```diff
@@ -167,6 +167,18 @@ Running and stopping the loop
 
    .. versionadded:: 3.6
 
+.. coroutinemethod:: loop.shutdown_default_executor()
+
+   Schedule the closure of the default executor, but wait for the
```
1st1:
```suggestion
   Schedule the closure of the default executor and wait for it to join all threads
```

1st1 (approved):
Please fix the last remaining nit I pointed out.  Thanks!  Looks great

aeros167:
@1st1 So do you think a "What's New" entry should be added as Victor was suggesting earlier, or should we skip it as @asvetlov was suggesting? 

If so, should it go under the API section or the section for the asyncio module? I think we just at least add it to the API section, since it's added a new coroutine method. It might be relevant to add it to the module section as well since it changes the behavior of ``asyncio.run()``, which may have some affect on users.

aeros167:
Oops, I have no idea why I named that commit "Update Misc/NEWS" entry. That should be "Adjust doc wording for loop.shutdown_default_executor()". Not sure if GitHub allows the updating of commit message titles within the PR (without undoing them). It shouldn't matter though since it'll all be squashed into a single commit, just wanted to clarify.

aeros167:
Oh crap, I just messed that up. I was attempting to update my master branch in my fork, I did not mean to affect this PR branch. I'll try to revert this branch back to https://github.com/python/cpython/pull/15735/commits/dbc08ae5bac719634ee3c22fa65e0d5915ca48f1. Apologies for the spam messages, I'm still learning how to use Git properly.

aeros167:
Fixed, I successfully reverted the PR branch to https://github.com/python/cpython/commit/dbc08ae5bac719634ee3c22fa65e0d5915ca48f1. I would appreciate it if someone could remove the additional reviewers added, I don't have permissions to do that. Sorry about the mess.

asvetlov:
done

aeros167:
Normally, I haven't left the "Patch by <name>" bit in the news entries for my own PRs, but since this one involved a significant amount of effort on my part, I'd like to do so. I'll update the news entry accordingly.

asvetlov:
Let's merge it, there is no reason for waiting

aeros167:
@asvetlov 
> Let's merge it, there is no reason for waiting

The only reason to wait was getting feedback from @1st1 and @vstinner on whether or not the "What's New" entry should be added, but that can be done at a later point if it's  useful/necessary. All of the actual changes were finalized.

vstinner:
> The only reason to wait was getting feedback from @1st1 and @vstinner on whether or not the "What's New" entry should be added, but that can be done at a later point if it's useful/necessary. All of the actual changes were finalized. Thanks for merging it!

New features deserve an entry in the What's New (In Python 3.8) doc, yes. You can open a new PR for that.

PR #15739, opened 2019-09-09
Doc: Fix PDF build (NoUri).

JulienPalard:
This is fixing an exception in the PDF builds of documentation. I fixed it the same way it's fixed in Sphinx: https://github.com/sphinx-doc/sphinx/commit/456abdefbe5167dfde70ecefd1b5ace09bb03ca9.

I'm currently running a test build to visually check it's OK.

### DIFF
--- a/Doc/tools/extensions/pyspecific.py
+++ b/Doc/tools/extensions/pyspecific.py
@@ -22,6 +22,7 @@ from docutils import nodes, utils
 
 from sphinx import addnodes
 from sphinx.builders import Builder
+from sphinx.errors import NoUri
 from sphinx.locale import translators
 from sphinx.util import status_iterator, logging
 from sphinx.util.nodes import split_explicit_title
@@ -569,10 +570,13 @@ def process_audit_events(app, doctree, fromdocname):
         for i, (doc, label) in backlinks:
             if isinstance(label, str):
                 ref = nodes.reference("", nodes.Text("[{}]".format(i)), internal=True)
-                ref['refuri'] = "{}#{}".format(
-                    app.builder.get_relative_uri(fromdocname, doc),
-                    label,
-                )
+                try:
+                    ref['refuri'] = "{}#{}".format(
+                        app.builder.get_relative_uri(fromdocname, doc),
+                        label,
+                    )
+                except NoUri:
+                    pass
                 node += ref
         row += nodes.entry('', node)
 
--- a/Doc/tools/extensions/pyspecific.py
+++ b/Doc/tools/extensions/pyspecific.py
@@ -22,7 +22,10 @@ from docutils import nodes, utils
 
 from sphinx import addnodes
 from sphinx.builders import Builder
-from sphinx.errors import NoUri
+try:
+    from sphinx.errors import NoUri
+except ImportError:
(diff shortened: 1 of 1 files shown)


Doc/tools/extensions/pyspecific.py
```diff
@@ -569,10 +570,13 @@ def process_audit_events(app, doctree, fromdocname):
...
-                    app.builder.get_relative_uri(fromdocname, doc),
-                    label,
-                )
+                try:
+                    ref['refuri'] = "{}#{}".format(
+                        app.builder.get_relative_uri(fromdocname, doc),
+                        label,
+                    )
+                except NoUri:
+                    pass
```
zooba: I would make it `continue` in this case, otherwise we'll end up with `[2]` that isn't a link but also has no information about where it'd be going.

If it ends up that none of the links work (fairly likely), then it'll show up with nothing at all.

JulienPalard: I trust you: anyway both PDF are rendered the same: the links are out of the page :(

zooba: No no, don't trust me on this :) I'm trusting you to get it to work

Doc/tools/extensions/pyspecific.py
```diff
@@ -22,6 +22,7 @@
 
 from sphinx import addnodes
 from sphinx.builders import Builder
+from sphinx.errors import NoUri
```
zooba: Looks like this is not available on the version of Sphinx that's used in CI. I'm pretty sure it's pinned somewhere as well.

Maybe we should just warn on all exceptions and skip the reference?

JulienPalard: It moved recently, we could try both imports.

Doc/tools/extensions/pyspecific.py
```diff
@@ -569,10 +570,13 @@ def process_audit_events(app, doctree, fromdocname):
...
-                ref['refuri'] = "{}#{}".format(
-                    app.builder.get_relative_uri(fromdocname, doc),
-                    label,
-                )
+                try:
+                    ref['refuri'] = "{}#{}".format(
+                        app.builder.get_relative_uri(fromdocname, doc),
+                        label,
+                    )
+                except NoUri:
```
zooba, line 581: Perhaps to avoid the import and provide useful feedback?

```suggestion
                except Exception as ex:
                    logging.getLogger("pyspecific.AuditEvent").warn(
                        "Failed to generate link: {}".format(ex))
```

JulienPalard: I prefer being precise and ignore `NoUri`, as Sphinx does interally. It's fully expected to get a `NoUri` here, it's not "exceptional": we're buildin a PDF file per section (library, c-api, tutorial, ...) but we're trying to link from a section to another (from library to c-api) so the links can't work. Other exceptions are *not* normal (I don't yet know them), so I prefer for them to bubble up explicitly.

zooba: Agreed. Having both imports is the better option.

zooba (approved):

PR #15751, opened 2019-09-09
bpo-38063: test_socket uses unittest.main()

vstinner:
Remove test_main() from test_socket: use setUpModule() with
tearDownModule() and unittest.main() to discover automatically test
cases.

Reorder also imports.

https://bugs.python.org/issue38063

### DIFF
--- a/Lib/test/test_socket.py
+++ b/Lib/test/test_socket.py
@@ -1,30 +1,35 @@
 import unittest
 from test import support
 
+import _thread as thread
+import array
+import contextlib
 import errno
 import io
 import itertools
-import socket
+import math
+import os
+import pickle
+import platform
+import queue
+import random
 import select
+import shutil
+import signal
+import socket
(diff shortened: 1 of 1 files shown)


vstinner:
I tested manually that I get the same number of test methods with and without this change:

```
# this PR
vstinner@apu$ ./python -m test test_socket --list-cases|wc -l
819

# without this change (reference)
vstinner@apu$ git checkout master
vstinner@apu$ ./python -m test test_socket --list-cases|wc -l
819
```

zware (approved):
LGTM!

vstinner:
Oh. "./python -m test test_socket --list-cases" doesn't work as expected. On the master branch, it lists a lot of tests which are *not* run...

Many tests of test_socket fail with this change. I will work on a fix.

Seluj78 (approved):

PR #15753, opened 2019-09-09
bpo-38037: Fix reference counters in signal module

ghost:
Only 3.8 and 3.9 branches are affected.

Use `PyDict_SetItemString()` in this PR.

https://bugs.python.org/issue38037

### DIFF
--- a/Misc/NEWS.d/next/Library/2019-09-09-18-39-23.bpo-38037.B0UgFU.rst
+++ b/Misc/NEWS.d/next/Library/2019-09-09-18-39-23.bpo-38037.B0UgFU.rst
@@ -1,0 +1,1 @@
+Fix reference counters in the :mod:`signal` module.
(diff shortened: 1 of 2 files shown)


vstinner (approved):
LGTM.

@nanjekyejoannah: Would you mind to review it as well?

nanjekyejoannah (approved):

nanjekyejoannah:
LGTM.

vstinner:
I managed to reproduce https://bugs.python.org/issue38037#msg351196 crash using this change:

```
index 4cf2b0726e..0ad2609882 100644
--- a/Objects/longobject.c
+++ b/Objects/longobject.c
@@ -16,10 +16,10 @@ class int "PyObject *" "&PyLong_Type"
 /*[clinic end generated code: output=da39a3ee5e6b4b0d input=ec0275e3422a36e3]*/
 
 #ifndef NSMALLPOSINTS
-#define NSMALLPOSINTS           257
+#define NSMALLPOSINTS           0
 #endif
 #ifndef NSMALLNEGINTS
-#define NSMALLNEGINTS           5
+#define NSMALLNEGINTS           0
 #endif
 
 _Py_IDENTIFIER(little);
```

Start Python, exit Python: Python does crash at exit.

I confirm that this change fix the crash.

PR #15762, opened 2019-09-09
bpo-38066: Hide internal Stream methods

asvetlov:
feed_eof(), feed_data(), set_exception(), and set_transport() are prefixed with underscore now.

https://bugs.python.org/issue38066

### DIFF
--- a/Lib/asyncio/streams.py
+++ b/Lib/asyncio/streams.py
@@ -1133,9 +1133,9 @@ class _BaseStreamProtocol(FlowControlMixin, protocols.Protocol):
         stream = self._stream
         if stream is not None:
             if exc is None:
-                stream.feed_eof()
+                stream._feed_eof()
             else:
-                stream.set_exception(exc)
+                stream._set_exception(exc)
         if not self._closed.done():
             if exc is None:
                 self._closed.set_result(None)
@@ -1147,12 +1147,12 @@ class _BaseStreamProtocol(FlowControlMixin, protocols.Protocol):
     def data_received(self, data):
         stream = self._stream
         if stream is not None:
-            stream.feed_data(data)
+            stream._feed_data(data)
 
     def eof_received(self):
         stream = self._stream
         if stream is not None:
-            stream.feed_eof()
+            stream._feed_eof()
         if self._over_ssl:
             # Prevent a warning in SSLProtocol.eof_received:
             # "returning true from eof_received()
@@ -1219,7 +1219,7 @@ class _StreamProtocol(_BaseStreamProtocol):
         stream = self._stream
         if stream is None:
             return
-        stream.set_transport(transport)
+        stream._set_transport(transport)
         stream._protocol = self
 
     def connection_lost(self, exc):
@@ -1449,7 +1449,7 @@ class Stream:
     def exception(self):
         return self._exception
 
-    def set_exception(self, exc):
+    def _set_exception(self, exc):
         self._exception = exc
 
         waiter = self._waiter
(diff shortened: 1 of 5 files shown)


1st1 (approved):

1st1:
@asvetlov Andrew, can we also remove the `Stream.transport` property?

asvetlov:
I think we can but it makes subprocess API a little less backward compatible.
Really wonder how much code is broken.
I'm working on backward-compatible approach: keep all mentioned method plus `transport` with underscore-prefixed version for internal usage and public version for backward compatibility that raises `DeprecationWarding`. It is the safest approach that we can do.

1st1:
But `asyncio.Stream` is a new thing - what backward compatibility problems do we have with it?

asvetlov:
`create_subprocess_exec` / `create_subprocess_exec` used to work with `StreamReader`/`StreamWriter`, now they are use `asyncio.Stream`. The replacement is drop-in, `Stream` has all API methods from reader and writer.

1st1:
Ah, right! Then your approach with deprecating them gracefully works.  Don't forget to also update the docs.

asvetlov:
Legacy methods are restored with emitting deprecation warnings.

Lib/asyncio/streams.py
```diff
@@ -1351,6 +1351,11 @@ def is_server_side(self):
 
     @property
     def transport(self):
+        warnings.warn("Stream.transport attribute is deprecated "
```
asvetlov, line 1354: asyncio itself never uses this property, there is no need for readonly internal API version

Lib/asyncio/streams.py
```diff
@@ -1366,7 +1371,7 @@ def writelines(self, data):
     def _fast_drain(self):
         # The helper tries to use fast-path to return already existing
         # complete future object if underlying transport is not paused
-        #and actual waiting for writing resume is not needed
+        # and actual waiting for writing resume is not needed
```
asvetlov, line 1374: I think it's ok to fix this unrelated comment spacing here. Don't worth a separate PR

Lib/asyncio/streams.py
```diff
@@ -1351,6 +1351,11 @@ def is_server_side(self):
 
     @property
     def transport(self):
+        warnings.warn("Stream.transport attribute is deprecated "
+                      "since Python 3.8, and scheduled for removal in 3.10; "
```
aeros: Minor grammar nit. The comma isn't needed here and there should be an "is" before "scheduled". 
```suggestion
                      "since Python 3.8 and is scheduled for removal in 3.10; "
```

Lib/asyncio/streams.py
```diff
@@ -1366,7 +1371,7 @@ def writelines(self, data):
     def _fast_drain(self):
         # The helper tries to use fast-path to return already existing
         # complete future object if underlying transport is not paused
```
aeros, line 1373: Unrelated to the PR, but could make a minor grammar fix here? I could make further suggestions to the phrasing of the code comments for this method, but I figured this suggestion would be easier and uncontroversial. 

```suggestion
         # complete future object if the underlying transport is not paused
```

aeros (commented):
Thanks for the PR @asvetlov.

I have a couple of minor grammar suggestions.

Lib/asyncio/streams.py
asvetlov: fast drain works with the single object, not multiple ones.

1st1 (approved):

1st1:
Great!

Lib/asyncio/streams.py
aeros: > fast drain works with the single object, not multiple ones.

Oops, I'll adjust the suggestion.

PR #10870, opened 2018-12-03
bpo-9004: Recommend against using utctimetuple in the docs.

abalkin:
Closes [bpo-9004](https://bugs.python.org/issue9004).

Original patch by Gaurav Tatke.

https://bugs.python.org/issue9004

PR #13410, opened 2019-05-18
bpo-36960: Make datetime docs more user-friendly

bsolomon1124:
https://bugs.python.org/issue36960

----

Appreciate any core reviewer who can take the time to sit down to read through the updated version.  While the diff is fairly large in terms of number of commits and lines changed, there are no drastic changes made here and I've made an effort to keep the commits modular and self-documenting.

Last rebased off master as of 2019-06-22.

PR #15773, opened 2019-09-09
bpo-37488 : Document a warning for datetime.utcnow() and utcfromtimestamp()

nanjekyejoannah:
https://bugs.python.org/issue37488

Automerge-Triggered-By: @pganssle

### DIFF
--- a/Doc/library/datetime.rst
+++ b/Doc/library/datetime.rst
@@ -862,13 +862,10 @@ Other constructors, all class methods:
    (for example, this may be possible on platforms supplying the C
    :c:func:`gettimeofday` function).
 
-   If *tz* is not ``None``, it must be an instance of a :class:`tzinfo` subclass, and the
-   current date and time are converted to *tz*’s time zone. In this case the
-   result is equivalent to::
+   If *tz* is not ``None``, it must be an instance of a :class:`tzinfo` subclass,
+   and the current date and time are converted to *tz*’s time zone.
 
-     tz.fromutc(datetime.utcnow().replace(tzinfo=tz))
-
-   See also :meth:`today`, :meth:`utcnow`.
+   This function is preferred over :meth:`today` and :meth:`utcnow`.
 
 
 .. classmethod:: datetime.utcnow()
@@ -878,6 +875,14 @@ Other constructors, all class methods:
    This is like :meth:`now`, but returns the current UTC date and time, as a naive
    :class:`.datetime` object. An aware current UTC datetime can be obtained by
    calling ``datetime.now(timezone.utc)``. See also :meth:`now`.
+
+   .. warning::
+
+      Because naive ``datetime`` objects are treated by many ``datetime`` methods
+      as local times, it is preferred to use aware datetimes to represent times
+      in UTC. As such, the recommended way to create an object representing the
+      current time in UTC  by calling ``datetime.now(timezone.utc)``.
+
 
 .. classmethod:: datetime.fromtimestamp(timestamp, tz=None)
 
@@ -889,10 +894,6 @@ Other constructors, all class methods:
    If *tz* is not ``None``, it must be an instance of a :class:`tzinfo` subclass, and the
    timestamp is converted to *tz*’s time zone.
 
-   In this case the result is equivalent to::
-
-     tz.fromutc(datetime.utcfromtimestamp(timestamp).replace(tzinfo=tz))
-
    :meth:`fromtimestamp` may raise :exc:`OverflowError`, if the timestamp is out of
    the range of values supported by the platform C :c:func:`localtime` or
    :c:func:`gmtime` functions, and :exc:`OSError` on :c:func:`localtime` or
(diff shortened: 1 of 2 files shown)


pganssle:
@nanjekyejoannah I don't think we can (or possibly should) put an actual warning in the code here - people will likely complain about it and we're not actually deprecating the function yet, it will likely just cause a lot of uproar and spam (as much as I'd like to...).

What I was referring to in bpo-15733 was a warning box in the documentation, like:

```rst
.. warning::

    Because naive datetimes are treated by many functions as local times,
    it is recommended that you create a timezone-aware datetime using
    ``datetime.now(tz=timezone.utc)`` instead.
```

I have not spent much time thinking about this. We may want a little footnote showing one of the gotchas to be referenced by both `utcfromtimestamp` and `utcnow`.

Doc/library/datetime.rst
```diff
@@ -29,8 +29,9 @@ is used to represent a specific moment in time that is not open to
 interpretation [#]_.
 
 A naive object does not contain enough information to unambiguously locate
-itself relative to other date/time objects.  Whether a naive object represents
-Coordinated Universal Time (UTC), local time, or time in some other timezone is
+itself relative to other date/time objects.  Native objects are no longer
```
pganssle: Was this an intentional change? It looks like you kept in the beginnings of an explanation of the problem, but also the problematic old text.

This is actually a bit complicated to reword, because the sentiment behind it has changed, and we may need to hedge a bit. Let me think about how best to communicate the new intent....

nanjekyejoannah:
Okay. I had thought you wanted us to introduce a warning. I will update sometime back.

Doc/library/datetime.rst
```diff
@@ -804,6 +811,11 @@ Other constructors, all class methods:
    except the latter formula always supports the full years range: between
    :const:`MINYEAR` and :const:`MAXYEAR` inclusive.
 
+   .. warning::
+
+      Because naive datetimes are treated by many functions as local times.
```
epicfaace: Can you change this wording to be of a similar sentence structure as in the earlier section (for `datetime.utcnow()`)?

Misc/NEWS.d/next/Library/2019-09-09-12-48-10.bpo-37488.kteuwA.rst
```diff
@@ -0,0 +1 @@
+Add warning to datetime.utcnow() and datetime.utcfromtimestamp()
```
epicfaace:
```suggestion
Add warning to :meth:`datetime.utcnow` and :meth:`datetime.utcfromtimestamp`
```

epicfaace (changes requested):

Doc/library/datetime.rst
```diff
@@ -752,6 +752,13 @@ Other constructors, all class methods:
    :class:`.datetime` object.  An aware current UTC datetime can be obtained by
    calling ``datetime.now(timezone.utc)``.  See also :meth:`now`.
 
+   .. warning::
+
+      Because naive datetimes are treated by many functions as local times,
```
aeros: This should be a bit more specific. Also, code comments or Sphinx roles should be used when referencing objects (such as classes, methods, or functions). In this case, code comments would be more appropriate since this is already within the section for ``datetime``. 

```suggestion
      Because naive ``datetime`` objects are treated by many ``datetime`` methods as local times,
```

This line will need to be split off into the next one since it will exceed the reST maximum width of 80 chars (94 chars after the suggestion).

Doc/library/datetime.rst
```diff
@@ -752,6 +752,13 @@ Other constructors, all class methods:
...
+      it is recommended that you create a timezone-aware datetime
```
aeros: Phrasing improvement, it's preferable to avoid the usage of pronouns such as "you" when possible within technical writing, and I think "timezone specific" makes a bit more sense than "timezone-aware". 
```suggestion
      it is recommended to create a timezone specific ``datetime`` object
```

Doc/library/datetime.rst
```diff
@@ -804,6 +811,11 @@ Other constructors, all class methods:
    except the latter formula always supports the full years range: between
    :const:`MINYEAR` and :const:`MAXYEAR` inclusive.
 
+   .. warning::
+
+      Because naive datetimes are treated by many functions as local times.
+      To get an aware :class:`.datetime` object, use :meth:`fromtimestamp`.
```
aeros: The Sphinx role here for ``datetime`` can be removed here since this section is already within the ``datetime`` class. The main purpose of the roles is to provide links to other sections, so there's not much of a purpose in a role that links to a section it is already contained within.

```suggestion
      To get a timezone specific datetime object, use :meth:`fromtimestamp`.
```

aeros (commented):

I have a few suggestions:

pganssle:
@nanjekyejoannah I believe this will have merge conflicts now that #13410 has been merged. Can you rebase against master? I also think adding a warning in `utctimetuple()` will allow us to close #10870.

Doc/library/datetime.rst
nanjekyejoannah: Thanks for the review. I consulted @pganssle at the sprints and he says we should keep to aware here. I hope I have addressed the other suggestions.

aeros: @nanjekyejoannah:
> Thanks for the review. I consulted @pganssle at the sprints and he says we should keep to aware here. I hope I have addressed the other suggestions.

Sounds good to me. If Paul prefers using "timezone-aware" over "timezone specific", I have no objections.

aeros (approved):
@nanjekyejoannah For consistency, can provide the warning message for each of them?

> Because naive ``datetime`` objects are treated by many ``datetime`` methods
as local times, the recommended way to create aware datetime objects for
the current time in UTC is by calling ``datetime.now(timezone.utc)``.

Otherwise, everything else looks good to me.

pganssle (approved):

PR #15782, opened 2019-09-09
bpo-38070: visit_decref() calls _PyObject_IsFreed()

vstinner:
In debug mode, visit_decref() now calls _PyObject_IsFreed() to ensure
that the object is not freed. If it's freed, the program fails with
an assertion error and Python dumps informations about the freed
object.

### DIFF
--- a/Modules/gcmodule.c
+++ b/Modules/gcmodule.c
@@ -376,6 +376,8 @@ static int
 visit_decref(PyObject *op, void *data)
 {
     assert(op != NULL);
+    _PyObject_ASSERT(op, !_PyObject_IsFreed(op));
+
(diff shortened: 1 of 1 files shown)


pablogsal (approved):
LGTM

vstinner:
I compared performances without/with the change.

(*) Test 1: ./configure --with-pydebug; make; ./python -m test -j0 # using 8 logical CPUs and gcc -Og

* ref: Total duration: 3 min 19 sec
* patch: Total duration: 3 min 24 sec (almost the same)

(*) Test 2: bm_deltablue.py of pyperformance

"Benchmark hidden because **not significant** (1): deltablue"

Ok, it seems like this change has no significant impact on a debug build of Python.

PR #15791, opened 2019-09-09
[3.8] bpo-20490: Improve circular import error message (GH-15308)

zooba:
https://bugs.python.org/issue20490


asottile (approved):
backport looks good -- thanks for fixing the conflicts!

PR #15798, opened 2019-09-09
bpo-38075: Port _randommodule.c to PEP-384

DinoV:
Summary:
- Migrate `Random_Type` to `PyType_FromSpec`
- To simulate an old use of `PyLong_Type.tp_as_number->nb_absolute`, I added
  code to the module init function to stash `int.__abs__` for later
  use. Ideally we'd use `PyType_GetSlot()` instead, but it doesn't currently
  work for static types in CPython, and implementing it just for this case
  doesn't seem worth it.

https://bugs.python.org/issue38075

### DIFF
--- a/Modules/_randommodule.c
+++ b/Modules/_randommodule.c
@@ -80,14 +80,22 @@
 #define LOWER_MASK 0x7fffffffU  /* least significant r bits */
 
 typedef struct {
+    PyObject *Random_Type;
+    PyObject *Long___abs__;
+} _randomstate;
+
+#define _randomstate(o) ((_randomstate *)PyModule_GetState(o))
+
+static struct PyModuleDef _randommodule;
+
+#define _randomstate_global _randomstate(PyState_FindModule(&_randommodule))
+
+typedef struct {
     PyObject_HEAD
     int index;
     uint32_t state[N];
 } RandomObject;
 
-static PyTypeObject Random_Type;
-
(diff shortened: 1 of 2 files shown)


the-knights-who-say-ni:

Modules/_randommodule.c
```diff
@@ -275,7 +283,8 @@ random_seed(RandomObject *self, PyObject *arg)
     if (PyLong_Check(arg)) {
         /* Calling int.__abs__() prevents calling arg.__abs__(), which might
            return an invalid value. See issue #31478. */
-        n = PyLong_Type.tp_as_number->nb_absolute(arg);
+        n = PyObject_CallFunctionObjArgs(_randomstate_global->Long___abs__, arg,
```
Yhg1s: This looks... ugh. It also *looks* like it could potentially change behaviour... and potentially be a lot slower than calling the slot function directly. Perhaps we need a new _Py_ conveniency function for this?

markshannon: The fast way to make this call is to use `_PyObject_Vectorcall`. Technically it is not in the public API yet, but `PyObject_Vectorcall` will be in 3.9

DinoV: I've switched this to doing a PyLong_CheckExact for the common case w/ a dispatch to PyNumber_Absolute if we match.  And in the case that we don't we'll fallback to the vector call.

PR #15809, opened 2019-09-09
Remove macOS tests from Travis.

benjaminp:
Azure runs macOS, so we don't need Travis to do it.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -73,16 +73,6 @@ matrix:
         - make -C Doc/ PYTHON=../python venv
       script:
         xvfb-run make -C Doc/ PYTHON=../python SPHINXOPTS="-q -W -j4" doctest
-    - name: "Mac OS X tests"
-      os: osx
-      language: c
-      compiler: clang
-      # Testing under macOS is optional until testing stability has been demonstrated.
-      env: OPTIONAL=true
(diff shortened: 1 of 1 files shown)


ammaraskar:
Is overlap the only motivation for removal here? I don't think overlap is necessarily bad, after all we run Linux on both Travis and Azure as well. Windows is run on Azure and AppVeyor. The redundancy helps to figure out flaky tests/race conditions and provides some actionable information in case a particular provider goes down.

benjaminp:
On Mon, Sep 9, 2019, at 18:11, Ammar Askar wrote:
> Is overlap the only motivation for removal here? I don't think overlap 
> is necessarily bad, after all we run Linux on both Travis and Azure as 
> well.

The overlap isn't free, though. Duplicated work takes up finite CI resources. It's also true that the more things we run the more likely we are to see flakes.

zooba:
I agree with this particular one, since the Travis resources are limited.

I'd prefer to keep as many parallel builds as we can though, especially cross-CI system, as that gives us better representation of different user environments.

ned-deily (commented):
I think that running the tests under macOS on Travis has not proved to be very effective at finding real problems so I support making this change.  Thanks!

PR #15814, opened 2019-09-09
Fix typo in dictobject comment

dalgarno:
Fixes a small typo in the dictobject comment. No`bpo` as it's a trivial fix.

### DIFF
--- a/Objects/dictobject.c
+++ b/Objects/dictobject.c
@@ -39,7 +39,7 @@ Size of indices is dk_size.  Type of each index in indices is vary on dk_size:
 * int32 for 2**16 <= dk_size <= 2**31
 * int64 for 2**32 <= dk_size
 
-dk_entries is array of PyDictKeyEntry.  It's size is USABLE_FRACTION(dk_size).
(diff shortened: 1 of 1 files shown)


brandtbucher (approved):

aeros (approved):
Thanks for the PR @dalgarno and welcome!

LGTM, we just need to wait for a core developer to review this.

Since this is involving a code comment change, I'll classify this as a documentation PR.

DinoV (approved):

methane:
Oh, I just created a thread for this PR...

https://discuss.python.org/t/fixing-simple-typo-in-comment-not-document/2297

aeros167:
@methane:
> Oh, I just created a thread for this PR...

Hmm, I can understand the concern about the messy commit logs, but I would strongly disagree that we should not accept these types of PRs. I don't think we should explicitly encourage them, but code comment typo fixing PRs provide a great introduction into the CPython workflow for new contributors. Many first time contributors start out with typo fixes, and then move on to making more impactful changes once they become more familiar with the workflow.

Also, if you want to go through the git log without seeing the typo fixes, you can simply use:

``git log --grep="typo" --invert-grep``

PR #15815, opened 2019-09-09
WIP: bpo-1615158 : POSIX capabilities support (In a separate module)

nanjekyejoannah:
WIP. Dont bother reviewing now.

Co-Author : Matt Kern 

https://bugs.python.org/issue1615158

### DIFF
--- a/Modules/posixcapmodule.c
+++ b/Modules/posixcapmodule.c
@@ -1,0 +1,645 @@
+/* POSIXCAP module implementation */
+
+#define PY_SSIZE_T_CLEAN
+
+#include "Python.h"
+#ifdef MS_WINDOWS
+   /* include <windows.h> early to avoid conflict with pycore_condvar.h:
+
+        #define WIN32_LEAN_AND_MEAN
+        #include <windows.h>
+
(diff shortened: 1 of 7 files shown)


Modules/posixcapmodule.c
```diff
@@ -0,0 +1,645 @@
+/* POSIXCAP module implementation */
+
+#define PY_SSIZE_T_CLEAN
+
+#include "Python.h"
+#ifdef MS_WINDOWS
```
eryksun, line 6: AFAICT, this module has nothing to do with Windows, in which case there's no reason to include windows.h or winreparse.h.

aeros167:
@nanjekyejoannah:
> WIP. Dont bother reviewing now.

Could you please add [WIP] to the title, just after the "bpo-#:" part? This makes it significantly easier to spot.

Modules/posixcapmodule.c
nanjekyejoannah: Noted. This was a result of a copy paste I think. I am revising this patch. Like I said it is still draft. Reviewing it now will waste anyone's time.

PR #15823, opened 2019-09-10
bpo-37936: Systematically distinguish rooted vs. unrooted in .gitignore.

gnprice:
A root cause of [bpo-37936](https://bugs.python.org/issue37936) is that it's easy to write a .gitignore
rule that's intended to apply to a specific file (e.g., the
`pyconfig.h` generated by `./configure`) but actually applies to all
similarly-named files in the tree (e.g., `PC/pyconfig.h`.)

Specifically, any rule with no non-trailing slashes is applied in an
"unrooted" way, to files anywhere in the tree.  This means that if we
write the rules in the most obvious-looking way, then

 * for specific files we want to ignore that happen to be in
   subdirectories (like `Modules/config.c`), the rule will work
   as intended, staying "rooted" to the top of the tree; but

 * when a specific file we want to ignore happens to be at the root of
   the repo (like `platform`), then the obvious rule (`platform`) will
   apply much more broadly than intended: if someone tries to add a
   file or directory named `platform` somewhere else in the tree, it
   will unexpectedly get ignored.

That's surprising behavior that can make the .gitignore file's
behavior feel finicky and unpredictable.

To avoid it, we can simply always give a rule "rooted" behavior when
that's what's intended, by systematically using leading slashes.

Further, to help make the pattern obvious when looking at the file and
minimize any need for thinking about the syntax when adding new rules:
separate the rules into one group for each type, with brief comments
identifying them.

For most of these rules it's clear whether they're meant to be rooted
or unrooted, but in a handful of cases I've only guessed.  In that
case the safer default (the choice that won't hide information) is the
narrower, rooted meaning, with a leading slash.  If for some of these
the unrooted meaning is desired after all, it'll be easy to move them
to the unrooted section at the top.

While here, also made some rules more general where that seems
most natural, and deleted some rules that appear long obsolete.

Co-Authored-By: Zachary Ware <zach@python.org>

https://bugs.python.org/issue37936

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -23,6 +23,8 @@
 .gdb_history
 .purify
 __pycache__
+.hg/
+.svn/
 .idea/
 tags
 TAGS
@@ -96,7 +98,6 @@ Tools/unicode/data/
 /config.log
 /config.status
 /config.status.lineno
-/.hg/
 /platform
 /pybuilddir.txt
 /pyconfig.h
@@ -106,7 +107,6 @@ Tools/unicode/data/
 /python-gdb.py
 /python.exe-gdb.py
 /reflog.txt
-/.svn/
 /.coverage
 /coverage/
 /externals/
--- a/.gitignore
+++ b/.gitignore
@@ -93,7 +93,6 @@ Tools/unicode/data/
 /config.status
 /config.status.lineno
 /.hg/
-/ipch/
 /libpython*.a
 /libpython*.so*
 /libpython*.dylib
--- a/.gitignore
+++ b/.gitignore
@@ -92,7 +92,6 @@ Tools/unicode/data/
 /config.log
 /config.status
 /config.status.lineno
-/db_home
 /.hg/
 /ipch/
 /libpython*.a
--- a/.gitignore
+++ b/.gitignore
@@ -1,7 +1,6 @@
-# Two-trick pony for OSX and other case insensitive file systems:
-# Ignore ./python binary on Unix but still look into ./Python/ directory.
-/python
-!/Python/
+#####
+# First, rules intended to apply in all subdirectories.
+# These contain no slash, or only a trailing slash.
 
 *.cover
 *.iml
@@ -18,6 +17,25 @@
 *.profraw
 *.dyn
 .gdb_history
+.purify
+__pycache__
+.idea/
+tags
+TAGS
+.vs/
+.vscode/
+gmon.out
+.mypy_cache/
+
+# Ignore core dumps... but not Tools/msi/core/ or the like.
+core
+!core/
+
+
+#####
+# Then, rules meant for a specific location relative to the repo root.
+# These must contain a non-trailing slash (and may also have a trailing slash.)
+
 Doc/build/
 Doc/venv/
 Doc/.venv/
@@ -29,7 +47,7 @@ Lib/lib2to3/*.pickle
 Lib/test/data/*
 !Lib/test/data/README
 /Makefile
-Makefile.pre
+/Makefile.pre
 Misc/python.pc
 Misc/python-embed.pc
 Misc/python-config.sh
@@ -69,46 +87,40 @@ PCbuild/arm64/
 PCbuild/obj/
 PCbuild/win32/
 Tools/unicode/data/
-.purify
-__pycache__
-autom4te.cache
-build/
-buildno
-config.cache
-config.log
-config.status
-config.status.lineno
-core
-!Tools/msi/core/
-db_home
-.hg/
-.idea/
-ipch/
-libpython*.a
-libpython*.so*
-libpython*.dylib
-libpython*.dll
-platform
-pybuilddir.txt
+/autom4te.cache
+/build/
+/buildno
+/config.cache
+/config.log
+/config.status
+/config.status.lineno
+/db_home
+/.hg/
+/ipch/
+/libpython*.a
+/libpython*.so*
+/libpython*.dylib
+/libpython*.dll
+/platform
+/pybuilddir.txt
 /pyconfig.h
-python-config
-python-config.py
-python.bat
-python.exe
-python-gdb.py
-python.exe-gdb.py
-reflog.txt
-.svn/
-tags
-TAGS
-.coverage
-coverage/
-externals/
-htmlcov/
+/python-config
+/python-config.py
+/python.bat
+/python.exe
+/python-gdb.py
+/python.exe-gdb.py
+/reflog.txt
+/.svn/
+/.coverage
+/coverage/
+/externals/
+/htmlcov/
 Tools/msi/obj
 Tools/ssl/amd64
 Tools/ssl/win32
-.vs/
-.vscode/
-gmon.out
-.mypy_cache/
+
+# Two-trick pony for OSX and other case insensitive file systems:
+# Ignore ./python binary on Unix but still look into ./Python/ directory.
+/python
+!/Python/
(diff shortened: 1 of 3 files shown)


.gitignore
```diff
@@ -69,46 +87,40 @@ PCbuild/arm64/
...
-pybuilddir.txt
+/autom4te.cache
+/build/
+/buildno
+/config.cache
+/config.log
+/config.status
+/config.status.lineno
+/db_home
+/.hg/
```
zware: This can safely be ignored globally; we're not going to check in a mercurial repository anywhere :)

.gitignore
```diff
@@ -69,46 +87,40 @@ PCbuild/arm64/
...
-externals/
-htmlcov/
+/python-config
+/python-config.py
+/python.bat
+/python.exe
+/python-gdb.py
+/python.exe-gdb.py
+/reflog.txt
+/.svn/
```
zware: This should also be globally ignored.

.gitignore
```diff
@@ -69,46 +87,40 @@ PCbuild/arm64/
...
+/.coverage
```
zware: And this.

.gitignore
```diff
@@ -69,46 +87,40 @@ PCbuild/arm64/
...
+/autom4te.cache
+/build/
+/buildno
+/config.cache
+/config.log
+/config.status
+/config.status.lineno
+/db_home
+/.hg/
+/ipch/
```
zware: I don't think this was meant to be rooted at `/`; IIRC it's intermediate files in the Windows build rooted in `PCbuild/<something>`, and thus can probably just go away since those entire directories are ignored now.

.gitignore
```diff
@@ -69,46 +87,40 @@ PCbuild/arm64/
...
+/libpython*.a
+/libpython*.so*
+/libpython*.dylib
+/libpython*.dll
```
zware: Each of these 4 can be generalized to ignore anything with these extensions; we don't have anything with these extensions checked in and likely never will.

.gitignore
```diff
@@ -69,46 +87,40 @@ PCbuild/arm64/
 PCbuild/obj/
```
zware, line 93: 4 lines up (thank you, GitHub, for not letting me comment there directly...), there's a `PCbuild/.vs/` rule, but the `.vs/` rule on line 25 should cover that.

.gitignore
```diff
@@ -69,46 +87,40 @@ PCbuild/arm64/
...
-ipch/
-libpython*.a
-libpython*.so*
-libpython*.dylib
-libpython*.dll
-platform
-pybuilddir.txt
+/autom4te.cache
+/build/
+/buildno
```
zware: I believe this is an ancient leftover from before the `buildno` section of `sys.version` referred to a VCS hash; it can just go away.

.gitignore
```diff
@@ -69,46 +87,40 @@ PCbuild/arm64/
...
-tags
-TAGS
-.coverage
-coverage/
-externals/
-htmlcov/
+/python-config
+/python-config.py
+/python.bat
+/python.exe
```
zware: We could globally ignore `*.exe`, with a specific exception for `!Lib/distutils/command/*.exe`.

.gitignore
```diff
@@ -69,46 +87,40 @@ PCbuild/arm64/
...
-platform
-pybuilddir.txt
+/autom4te.cache
+/build/
+/buildno
+/config.cache
+/config.log
+/config.status
+/config.status.lineno
+/db_home
```
zware: This is a holdover from the `bsddb` package, and can go away (so can the explicit callout of `db_home` in Lib/test/libregrtest/runtest.py).

.gitignore
```diff
@@ -69,46 +87,40 @@ PCbuild/arm64/
...
-libpython*.dll
-platform
-pybuilddir.txt
+/autom4te.cache
+/build/
+/buildno
+/config.cache
+/config.log
+/config.status
+/config.status.lineno
```
zware, line 101: Maybe combine these into `/config.status*`?

zware (changes requested):
This mostly looks good, but I have several suggestions for further cleanup.

.gitignore
gnprice: Good catch, thanks!

.gitignore
gnprice: Ah, indeed!

Looks like that disappeared in b98eb875d , in the Python 3.0 period (and the bit that made a `db_home` directory went away just before that, in d098ff232 .)

.gitignore
gnprice: Sure.

.gitignore
gnprice: Good idea, thanks!

.gitignore
gnprice: I think `config.status.lineno` is the only one like this -- it looks like it comes from [these lines of `configure`](https://github.com/python/cpython/blob/3f4db4a0bab073b768fae958e93288bd5d24eadd/configure#L468)... where it checks if the current shell has the special variable `$LINENO`, and if not, then it runs a sed script on *itself* (?), turning all the `$LINENO` references into the actual line number, and sources the resulting script instead.

Or something like that; this is some fairly dense shell-scripting. Not sure the exact pathway that causes that to end up making a `config.status.lineno` rather than `configure.lineno`. In any case, I'm not seeing signs of a more general pattern of doing the same with other suffixes.

Given that just `config.status` is the main thing, and `config.status.lineno` is a fairly obscure portability hack and the only one of its kind, I think it's probably more confusing than it's worth to write a general pattern `/config.status*` for them.

.gitignore
gnprice: Solid prediction, I think :)

gnprice (commented):
Thanks @zware for the many helpful suggestions!

Just pushed an update making all your suggested changes, with the exception of the one on `config.status`; details below.

gnprice:
@zware Also just edited the PR description (aka draft commit message) to credit you as a coauthor, because that seems appropriate given all the new substantive changes you contributed :slightly_smiling_face: 

(If you disagree, though, please feel free to take it out when merging.)

zware (approved):
LGTM, thanks!

zware:
The 3.7 backport is just complicated enough that I'm going to leave it alone.

Thanks for the patch, @gnprice!

.gitignore
```diff
@@ -1,11 +1,14 @@
...
-!/Python/
+#####
+# First, rules intended to apply in all subdirectories.
+# These contain no slash, or only a trailing slash.
 
 *.cover
 *.iml
 *.o
+*.a
+*.so*
```
asottile, line 9: fwiw, this is causing a minor inconvenience in packaging for debian as this now matches `debian/README.source`

the original pattern was `libpython*.so*` could that be restored?

zware: `debian/README.source` does not exist in the cpython repo, so presumably you can unignore it wherever you're patching it in.  I would think you would want to explicitly unignore anything added to avoid issues like this.

I prefer to keep `*.so*`, because we we don't want to accidentally check in any compiled extension modules (even though that shouldn't happen anyway and any `*.so` other than `libpython*` should theoretically only appear in the ignored `build` directory, ignoring them anyway seems better to me).

However, I could be persuaded to expand that rule to

```
*.so
*.so.*
```

If that works for you @asottile, please submit a PR and ping me.

asottile: will send a PR :+1:

PR #15830, opened 2019-09-10
bpo-37251: Removes __code__ check from _is_async_obj.

lisroach:
It should be possible that an awaitable object does not have a `__code__`, so technically checking for the `__code__` attribute in `_is_async_obj` is incorrect. This faulty code check is also why things like: `mock.MagicMock(spec=lambda x: x)` evaluate to True, even though they are not asynchronous.

https://bugs.python.org/issue37251


Misc/NEWS.d/next/Library/2019-09-10-10-59-50.bpo-37251.8zn2o3.rst
```diff
@@ -0,0 +1,3 @@
+Removes `__code__` check in AsyncMock's `_is_async_obj` that incorrectly
```
tirkarthi: Nit: I think we can remove `_is_async_obj` since it's just a private function.

```suggestion
Remove `__code__` check in AsyncMock that incorrectly
```

tirkarthi (approved):
LGTM. Thanks Lisa.

asvetlov (commented):
LGTM

asvetlov:
Added backport label

PR #15839, opened 2019-09-10
bpo-22273: Update ctypes to correctly handle arrays in small structur…

vsajip:
…es on x86_64.

https://bugs.python.org/issue22273

### DIFF
--- a/Modules/_ctypes/stgdict.c
+++ b/Modules/_ctypes/stgdict.c
@@ -350,6 +350,9 @@ PyCStructUnionType_update_stgdict(PyObject *type, PyObject *fields, int isStruct
     int pack;
     Py_ssize_t ffi_ofs;
     int big_endian;
+#if defined(X86_64)
+    int arrays_seen = 0;
+#endif
 
     /* HACK Alert: I cannot be bothered to fix ctypes.com, so there has to
        be a way to use the old, broken semantics: _fields_ are not extended
@@ -501,6 +504,10 @@ PyCStructUnionType_update_stgdict(PyObject *type, PyObject *fields, int isStruct
             Py_XDECREF(pair);
             return -1;
         }
+#if defined(X86_64)
+        if (PyCArrayTypeObject_Check(desc))
+            arrays_seen = 1;
+#endif
         dict = PyType_stgdict(desc);
         if (dict == NULL) {
             Py_DECREF(pair);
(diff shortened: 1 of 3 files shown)


vsajip:
@eryksun Would appreciate your comments on this PR. Here, I only added support for arrays in small structures - I will address the union and bitfield issues in bpo-16575 and bpo-16576 in a separate PR.

Modules/_ctypes/stgdict.c
```diff
@@ -641,6 +648,103 @@ PyCStructUnionType_update_stgdict(PyObject *type, PyObject *fields, int isStruct
...
+         * By small structures, we mean ones that are 16 bytes or less. In that
+         * case, there can't be more than 16 elements after unrolling arrays,
+         * as we (will) disallow bitfields. So we can collect the true ffi_type
+         * values in a fixed-size local array on the stack and, if any arrays
+         * were seen, replace the ffi_type_pointer.elements with a more
+         * accurate set, to allow libffi to marshal them into registers
+         * correctly. It means one more loop over the fields, but if we got
+         * here, the structure is small, so there aren't too many of those.
+         */
+        ffi_type * actual_types[MAX_ELEMENTS + 1];
```
ZackerySpytz:
```suggestion
        ffi_type *actual_types[MAX_ELEMENTS + 1];
```

Modules/_ctypes/stgdict.c
```diff
@@ -641,6 +648,103 @@ PyCStructUnionType_update_stgdict(PyObject *type, PyObject *fields, int isStruct
...
+        int actual_type_index = 0;
+
+        memset(actual_types, 0, sizeof(actual_types));
+        for (i = 0; i < len; ++i) {
+            PyObject *name = NULL, *desc = NULL;
```
ZackerySpytz: It doesn't seem like it's necessary to initialize `name` and `desc` to `NULL`.

Modules/_ctypes/stgdict.c
```diff
@@ -641,6 +648,103 @@ PyCStructUnionType_update_stgdict(PyObject *type, PyObject *fields, int isStruct
...
+            PyObject *pair = PySequence_GetItem(fields, i);
+            StgDictObject *dict;
+            int bitsize = 0;
+
+            if (!pair || !PyArg_ParseTuple(pair, "UO|i", &name, &desc, &bitsize)) {
```
ZackerySpytz: According to PEP 7, lines should be limited to 79 chars.

Modules/_ctypes/stgdict.c
```diff
@@ -641,6 +648,103 @@ PyCStructUnionType_update_stgdict(PyObject *type, PyObject *fields, int isStruct
...
+                PyErr_SetString(PyExc_TypeError,
```
ZackerySpytz, line 687: This code will overwrite any possible exception that is raised by `PySequence_GetItem()` (if it fails). I'm not sure if that should be done.

Modules/_ctypes/stgdict.c
vsajip: IMO we shouldn't apply this rule _too_ strictly - it's intended to avoid too-deep nesting in code. In just about all cases in this file, the longer lines are due to verbose error messages or long variable names (neither of which are a bad thing).

Modules/_ctypes/stgdict.c
```diff
@@ -641,6 +648,106 @@ PyCStructUnionType_update_stgdict(PyObject *type, PyObject *fields, int isStruct
     stgdict->align = total_align;
     stgdict->length = len;      /* ADD ffi_ofs? */
 
+#if defined(X86_64)
+
+#define MAX_ELEMENTS 16
+
+    if (arrays_seen && (size <= 16)) {
```
Kentzo, line 655: @vsajip Is it intentional that `16` is used here instead of `MAX_ELEMENTS`?

vsajip: Whoops, no. That's a slip-up. Looks like I'll need to find a fix for the buildbot failures, anyway ... so I'll deal with this soon.

Modules/_ctypes/stgdict.c
```diff
@@ -641,6 +648,106 @@ PyCStructUnionType_update_stgdict(PyObject *type, PyObject *fields, int isStruct
...
+                    Py_DECREF(pair);
+                    PyErr_Format(PyExc_TypeError,
+                                 "second item in _fields_ tuple (index %zd) must be a C type",
+                                 i);
+                    return -1;
+                }
+                /* Copy over the element's type, length times. */
+                while (length > 0) {
+                    actual_types[actual_type_index++] = &edict->ffi_type_pointer;
+                    assert(actual_type_index <= MAX_ELEMENTS);
```
Kentzo, line 720: Assertion can be checked just once before the while loop.

vsajip: But `actual_type_index` is incremented in the loop, and in theory the increment could take it past the end of the array, which is what the assertion is guarding against.

PR #15848, opened 2019-09-10
bpo-38090: Fix reference leak in ceval.c

pablogsal:
https://bugs.python.org/issue38090


pablogsal:
Tested locally:

```
❯ ./python.exe -m test test_ctypes -R :
Run tests sequentially
0:00:00 load avg: 1.24 [1/1] test_ctypes
beginning 9 repetitions
123456789
.........

== Tests result: SUCCESS ==

1 test OK.

Total duration: 8 sec 596 ms
Tests result: SUCCESS
```

benjaminp (approved):

PR #15849, opened 2019-09-10
bpo-38088: Fixes distutils not finding vcruntime140.dll with only v142 toolset installed

zooba:
https://bugs.python.org/issue38088


zware (approved):
LGTM

PR #15851, opened 2019-09-10
bpo-38089: Move Azure Pipelines to latest VM versions and make macOS tests optional

zooba:
https://bugs.python.org/issue38089

### DIFF
--- a/.azure-pipelines/ci.yml
+++ b/.azure-pipelines/ci.yml
@@ -12,7 +12,7 @@ jobs:
   displayName: Pre-build checks
 
   pool:
-    vmImage: ubuntu-latest
+    vmImage: ubuntu-16.04
 
   steps:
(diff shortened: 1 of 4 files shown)


zooba:
Note that this will fail until #15850  and #15849  are merged.

zware:
I'm a bit concerned about this one; this leaves us open to being broken by any image update.  With Azure being required, that means any update that breaks us forces us to manually intervene in any PR opened before the breakage is fixed.  We've had a few instances of that otherwise; I didn't like them :)

zooba:
I guess given I found two other bugs by doing this, I can't really argue with the possibility :)

I'll see if I can figure out the current names of the images and use them explicitly.

zooba:
@zware Waiting on your approval

zware (approved):
LGTM

PR #5053, opened 2017-12-30
gh-72680: Fix false positives when using zipfile.is_zipfile()

jjolly:
Fix zipfile validation issue by ... providing more validation!

Originally, zipfile.is_zipfile() only checked the End Central Directory
signature. If the signature could be found in the last 64k of the file,
success! This produced false positives on any file with 'PK\x05\x06' in the
last 64k of the file - including PDFs and PNGs.

This is now corrected by actually validating the Central Directory location
and size based on the information provided by the End Central Directory
along with verifying the Central Directory signature of the first entry.

This should be sufficient for the vast number of zipfiles, but more could be
done to absolutely validate the zipfile content - such as validating all
local file headers and Central Directory entries.

https://bugs.python.org/issue28494

* Issue: gh-72680

PR #15853, opened 2019-09-10
bpo-28494: Test existing zipfile working behavior.

gpshead:
Adds unittests for executables with a zipfile appended to test_zipfile
as zipfile.is_zipfile and zipfile.ZipFile work properly on these today.

This is being added as a regression test that https://github.com/python/cpython/pull/5053 would otherwise fail today.

https://bugs.python.org/issue28494


gpshead:
duh, fixing the test to skip the execute one on windows :)

Lib/test/test_zipfile.py
```diff
@@ -2496,11 +2496,15 @@ def test_read_zip64_with_exe_prepended(self):
         self._test_zip_works(self.exe_zip64)
 
     @unittest.skipUnless(sys.executable, 'sys.executable required.')
+    @unittest.skipUnless(os.access('/bin/bash', os.X_OK),
```
Yhg1s, line 2499: How about using shutil.which('bash') instead? There are some windows machines with bash installed now... will the test work there?

gpshead: I didn't think windows supported #! line executables as a concept.  executable is based on filename ending in bat/cmd/com/exe, not an execute bit there.

Yhg1s:
Buildbot failure should have been fixed by GH-15902.

PR #15877, opened 2019-09-10
bpo-36876: Add a tool that identifies unsupported global C variables.

ericsnowcurrently:
This PR does 2 things:

* make the globals checker tool more robust and maintainable
* add a test that invokes the checker and fails if any unsupported globals have been added

For now the test is marked as a known failure.  I'll be working to fix all of unsupported globals, at which point I'll unmark the test.

Note that this PR adds code to `Tools/c-analyzer` or adds new tests.  The only existing code I touch is where I factor out an existing testing helper from tests for another tool.


vstinner:
115 commits, 51 files changes, and wow, that's a huge PR.

vstinner:
" bpo-36876: Make sure no new globals get added."

So far, the C API don't provide any solution for function "static PyObject *local_global_var = ...;" :-)

ericsnowcurrently:
Yeah, it's a lot, but it's all under Tools/. :)  Also, I've updated the title to be more clear.

ericsnowcurrently:
FYI, all the failures are strictly with patchcheck, which is complaining about whitespace in some files.  I cannot reproduce that locally.

PR #15883, opened 2019-09-10
bpo-38092: Correctly pass __PYVENV_LAUNCHER__ in when creating processes in windows

pierreglaser:
I guess it also needs a backport to 3.8 and 3.7.

Also, I still haven't a proper reproducer and this bothers me...

pinging @zooba

https://bugs.python.org/issue38092


zooba (changes requested):
Please also add the change to the default Python executable as we discussed on the issue.

zooba:
Thanks for the PR! I completed it in #16098 with the rest of the change

PR #15889, opened 2019-09-10
bpo-36373: Fix deprecation warnings

asvetlov:

### DIFF
--- a/Lib/unittest/async_case.py
+++ b/Lib/unittest/async_case.py
@@ -89,8 +89,9 @@ class IsolatedAsyncioTestCase(TestCase):
         else:
             return ret
 
-    async def _asyncioLoopRunner(self):
-        queue = self._asyncioCallsQueue
+    async def _asyncioLoopRunner(self, fut):
+        self._asyncioCallsQueue = queue = asyncio.Queue()
+        fut.set_result(None)
         while True:
             query = await queue.get()
             queue.task_done()
@@ -113,8 +114,9 @@ class IsolatedAsyncioTestCase(TestCase):
         asyncio.set_event_loop(loop)
         loop.set_debug(True)
         self._asyncioTestLoop = loop
-        self._asyncioCallsQueue = asyncio.Queue(loop=loop)
-        self._asyncioCallsTask = loop.create_task(self._asyncioLoopRunner())
+        fut = loop.create_future()
+        self._asyncioCallsTask = loop.create_task(self._asyncioLoopRunner(fut))
+        loop.run_until_complete(fut)
 
     def _tearDownAsyncioLoop(self):
         assert self._asyncioTestLoop is not None
--- a/Lib/asyncio/locks.py
+++ b/Lib/asyncio/locks.py
@@ -332,7 +332,7 @@ class Condition(_ContextManagerMixin):
                           DeprecationWarning, stacklevel=2)
 
         if lock is None:
-            lock = Lock(loop=self._loop)
+            lock = Lock(loop=loop)
         elif lock._loop is not self._loop:
             raise ValueError("loop argument must agree with lock")
 
--- a/Lib/asyncio/queues.py
+++ b/Lib/asyncio/queues.py
@@ -45,7 +45,7 @@ class Queue:
         # Futures.
         self._putters = collections.deque()
         self._unfinished_tasks = 0
-        self._finished = locks.Event(loop=self._loop)
+        self._finished = locks.Event(loop=loop)
         self._finished.set()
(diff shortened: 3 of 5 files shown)


asvetlov:
@tirkarthi that's what I mean

Lib/unittest/async_case.py
```diff
@@ -89,8 +89,9 @@ def _callMaybeAsync(self, func, /, *args, **kwargs):
         else:
             return ret
 
-    async def _asyncioLoopRunner(self):
-        queue = self._asyncioCallsQueue
+    async def _asyncioLoopRunner(self, fut):
+        self._asyncioCallsQueue = queue = asyncio.Queue()
```
tirkarthi, line 93: Just curious why we moved it here. Is there a difference in constructing it in `_setupAsyncioLoop` itself without passing loop and having it here?

tirkarthi (approved):
LGTM. As I understand previously we were always passing `self._loop` which would always be initialized with a loop object and would mean we are passing the loop value to the other callers though the initial caller had `loop=None` only. This looks more correct and simple to me given `asyncio.Queue()` is a valid call. Thanks Andrew.

1st1 (changes requested):
Please don't merge this until we discuss it.

Lib/unittest/async_case.py
asvetlov: Well, it is possible to just drop the `loop` parameter from `Queue` construction because we installed the default event loop a few lines above but I did two steps further.

Instantiation on asyncio loop coupled objects from an async function is more idiomatic, it always grabs a currently executed loop (the loop which is used to run a coroutine).
It prevents things like this:
```
import asyncio
queue = asyncio.Queue()

asyncio.run(...)  # a code that works with queue
```
In the example `queue` variable is instantiated with the default asyncio loop but `run()` creates a new loop version. So the queue doesn't work and hangs eventually because a future created by one loop cannot be awaited by another one.

In Python 3.9 I have a plan to add another deprecation warning if `not self._loop.is_running()` for queues and locks and eventually replace `get_event_loop()` with `get_running_loop()` as a final fix.

asvetlov:
`asyncio.run()` is a nice API, everybody should use it (and people do as I see).

On Python 3.6 where `asyncio.run()` doesn't exist the difference is not drastic though, instead of `asyncio.run(main())` you can use `loop.run_until_complete(main())`. 
The loop preparation and especially finalization is missing here but roughly the code is the same.

Also, please read [comment above](https://github.com/python/cpython/pull/15889#discussion_r322860377). In aiohttp we forbade the creation of `ClientSession` outside of async function context step by step because we have a long gory list of reports when "aiohttp hangs for some reason", and the reason was misusage of a loop and an object that belongs to a different loop by accident.

1st1 (approved):

1st1:
Alright.

dstansby:
Is there a drop in replacement for code that was previously using the `loop` parameter to `Queue`? I'm struggling to understand how to modify code to have the same behaviour without using the `loop` argument.

aaugustin:
I had the same issue and found this workaround.

```diff
     # Create an event loop that will run in a background thread.
     loop = asyncio.new_event_loop()

+    # Due to zealous removal of the loop parameter in the Queue constructor,
+    # we need a factory coroutine to run in the freshly created event loop.
+    async def queue_factory():
+        return asyncio.Queue()
+
     # Create a queue of user inputs. There's no need to limit its size.
-    inputs: asyncio.Queue[str] = asyncio.Queue(loop=loop)
+    inputs: asyncio.Queue[str] = loop.run_until_complete(queue_factory())

     # Create a stop condition when receiving SIGINT or SIGTERM.
     stop: asyncio.Future[None] = loop.create_future()
```

Forcing the introduction of factory functions counts as progress towards Java, I guess...

cladmi:
I ended up here after following deprecation warning -> git blame -> ticket -> comments.
So reviving it as may be useful for others following the path.

I have a case of an asynchronous loop running on another thread already, and a synchronous program that creates tasks to run on it. Basically like background workers.

Trying to go a generic way here. Maybe they would be some existing helpers but I did not find it yet.

```
async def as_coroutine(cb, *args, **kwargs):
    return cb(*args, **kwargs)


def call_on_loop(cb, *args, loop, **kwargs):
    """Run a callback on the async loop and return its return value."""
    coro = as_coroutine(cb, *args, **kwargs)
    return asyncio.run_coroutine_threadsafe(coro, loop).result()


# Loop is running already somewhere
queue = call_on_loop(asyncio.Queue, loop=loop)
```

PR #15890, opened 2019-09-10
bpo-35784: Include optional kwargs in hashlib.new() documentation

ta1hia:
Updating documentation for hashlib.new(), including Doc/library/hashlib.rst and the docstring in Lib/hashlib.py. 

https://bugs.python.org/issue35784

https://bugs.python.org/issue35784

### DIFF
--- a/Doc/library/hashlib.rst
+++ b/Doc/library/hashlib.rst
@@ -99,7 +99,7 @@ More condensed:
    >>> hashlib.sha224(b"Nobody inspects the spammish repetition").hexdigest()
    'a4337bc45a8fc544c03f52dc550cd6e1e87021bc896588bd79e901e2'
 
-.. function:: new(name[, data])
+.. function:: new(name[, data, **kwargs])
 
    Is a generic constructor that takes the string *name* of the desired
    algorithm as its first parameter.  It also exists to allow access to the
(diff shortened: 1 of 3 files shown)


gpshead (approved):
A news entry isn't really necessary for a documentation update, but is harmless regardless. :)

rhettinger:
> additional keyword arguments that are specific to the selected hashing algorithm

This is going to raise the question of which keywords are available for each algorithm and what they do.  It would be nice if we had a summary table (like that at the top of the *itertools()* documentation).

gpshead:
> This is going to raise the question of which keywords are available for each algorithm and what they do. It would be nice if we had a summary table (like that at the top of the itertools() documentation).

Lets not make such a table.  People shouldn't use `hashlib.new()` in _most_ code.  It is a conveinence function for the use case where someone needs to choose a hash algorithm at runtime.  I don't want to duplicate documentation for the same thing in two places in hashlib.rst.

The one true canonical place for arguments for any given hash function to be defined is in the documentation for the `hashlib.$algorithmname()` constructor. Or individual documentation specific to each algorithm should we wind up in a situation where a `$algorithmname` constructor no longer makes sense (as some of the plethora of suffixed blake$mumble variants might do).  Not as part of `hashlib.new()`.

tahia-khan:
@tiran should this PR be closed then, since the semantic might change anyway after you finish integrating OpenSSL?

tiran:
> @tiran should this PR be closed then, since the semantic might change anyway after you finish integrating OpenSSL?

Makes sense, let's close it. Thanks for your contribution. :+1:

PR #11661, opened 2019-01-23
bpo-35810: Incref heap-allocated types in PyObject_Init

eduardo-elizondo:
https://bugs.python.org/issue35810

Correctly incref an intance's type. Currently, if a heap-allocated type creates an instance through PyObject_{,GC}_New{Var}, the type won't incref. This adds a change to pull the incref to the core PyObject_Init{INIT} function to correctly incref heap-allocated types.

This now means that heap-allocated types that add a custom tp_dealloc, should decref the instance types - just like the default tp_dealloc does.

Currently there are 10 heap-allocated types in CPython:
* PyCursesPanel_Type in _curses_panel.c
* sslerror_type in _ssl.c
* Example_Type in _testmultiphase.c
* Str_Type in _testmultiphase.c
* Tkapp_Type in _tkinter.c
* Tktt_Type in _tkinter.c
* PyTclObject_Type in _tkinter.c
* Xxo_Type in xxlimited.c
* Str_Type in xxlimited.c
* Null_Type in xxlimited.c
* Struct Sequences in structseq.c

Out of those only the following 5 types allocate instances through PyObject_{GC}_New{Var}:
* PyCursesPanel_Type. Action: Added a decref in its tp_dealloc
* Tkapp_Type. Action: Removed the manual incref after allocation.
* Tktt_Type. Action: Removed the manual incref after allocation.
* PyTclObject_Type. Action: Removed the manual incref after allocation.
* Xxo_Type. No Action: It inherits the default tp_dealloc which already decrefs the type.
* Struct Sequences. Action: Removed decref the type in the deallocation function.

PR #10854, opened 2018-12-03
bpo-35381 Make all posixmodule types heap-allocated

eduardo-elizondo:
After #9665, this moves the remaining types in posixmodule to be heap-allocated to make it compatible with PEP384

https://bugs.python.org/issue35381

PR #9665, opened 2018-10-02
bpo-34784: Posixmodule: Heap StructSequence

eduardo-elizondo:
https://bugs.python.org/issue34784

PR #15892, opened 2019-09-11
bpo-35381 Remove all static state from posixmodule

eduardo-elizondo:
After #9665, this moves the remaining types in posixmodule to be heap-allocated to make it compatible with PEP384 as well as modifying all the type accessors to fully make the type opaque.

The original PR that got messed up a rebase: https://github.com/python/cpython/pull/10854. All the issues in that commit have now been addressed since https://github.com/python/cpython/pull/11661 got committed.

This change also removes any state from the data segment and onto the module state itself.

https://bugs.python.org/issue35381


Modules/posixmodule.c
```diff
@@ -6157,33 +6161,38 @@ os_sched_getscheduler_impl(PyObject *module, pid_t pid)
...
-/*[clinic end generated code: output=48f4067d60f48c13 input=ab4de35a9a7811f2]*/
-{
-    PyObject *res;
-
-    res = PyStructSequence_New(type);
+PyDoc_STRVAR(os_sched_param__doc__,
+"sched_param(sched_priority)\n"
+"--\n"
+"\n"
+"Current has only one field: sched_priority\");\n"
```
encukou: This has bad grammar and some extra characters. Also, it has only duplicate information; can it be just removed?
```suggestion
```

Modules/posixmodule.c
```diff
@@ -6157,33 +6161,38 @@ os_sched_getscheduler_impl(PyObject *module, pid_t pid)
...
-[clinic start generated code]*/
-
-static PyObject *
-os_sched_param_impl(PyTypeObject *type, PyObject *sched_priority)
-/*[clinic end generated code: output=48f4067d60f48c13 input=ab4de35a9a7811f2]*/
-{
-    PyObject *res;
-
-    res = PyStructSequence_New(type);
+PyDoc_STRVAR(os_sched_param__doc__,
```
encukou: The reason this is changed to manual argument parsing is that it's `#if`-ed,  right?

Modules/posixmodule.c
```diff
@@ -14462,9 +14463,11 @@ INITFUNC(void)
         }
 
         /* initialize scandir types */
-        if (PyType_Ready(&ScandirIteratorType) < 0)
+        ScandirIteratorType = (PyTypeObject *)PyType_FromSpec(&ScandirIteratorType_spec);
```
encukou: This assumes the `INITFUNC` is run exactly once per process. The classes will be shared across subinterpreters, losing most benefits of heap types.

Modules/posixmodule.c
```diff
@@ -14428,8 +14423,11 @@ INITFUNC(void)
         if (StatResultType == NULL) {
             return NULL;
         }
-        structseq_new = StatResultType->tp_new;
-        StatResultType->tp_new = statresult_new;
+        /* Add a custom __new__ to the structsequence */
+        structseq_new = (newfunc)PyType_GetSlot((PyTypeObject *)StatResultType, Py_tp_new);
+        dunder_new = PyDescr_NewClassMethod((PyTypeObject *)StatResultType, &StatResultType_dunder_new);
```
encukou: `PyDescr_NewClassMethod` can fail; this is missing error handling.

Modules/posixmodule.c
```diff
@@ -14428,8 +14423,11 @@ INITFUNC(void)
...
+        PyObject_SetAttrString((PyObject *)StatResultType, "__new__", dunder_new);
```
encukou: I don't quite understand this. Why is the `__new__` switched?

encukou (commented):
As Jeroen pointed out in the issue, PEP 384 (Stable ABI) isn't very relevant to the standard library. PEP 489 (Multi-phase init) would bring benefits (better subinterpreter support), but this isn't quite there.
Is there a practical benefit to this change?

encukou:
I pushed a few changes directly rather than writing a comment about them. Hope that's OK with you :)

eduardo-elizondo:
Note: I will address the pending comments in a bit

encukou:
With C static globals left in, we unfortunately don't quite get those benefits.

eduardo-elizondo:
@encukou I indeed have a local patch with all static state removed from posixmodule. I had it split up to make it easier to review the two things independently. I was planning to follow up this change with the static state removal. Would you want me to merge that here instead?

Modules/posixmodule.c
eduardo-elizondo: This is now addressed with the module state

Modules/posixmodule.c
eduardo-elizondo: This is the current behavior, so my change preserves this. If the question is: why was the code was written like this in the first place, that's a whole other topic!!

The TL;DR is that this wanted to add some extra checks / error handling so it pulled tp_new out and wrapped it around another function

Modules/posixmodule.c
eduardo-elizondo: I believe this was copy paste from the clinic. Let me update it.

Modules/posixmodule.c
eduardo-elizondo: Correct

Modules/posixmodule.c
eduardo-elizondo: Thanks! Fixed

encukou:
I'm not done reviewing, but comments so far:
* Module-level constant strings should use `PyUnicode_InternFromString` rather than `PyUnicode_FromString`
* I'd rather bring `_Py_IDENTIFIER(__fspath__)` back (for now) than reimplement `_PyObject_LookupSpecial`. The workaround is worse than the problem; we should work on a more systematic solution.
* In the stdlib, I'd prefer using `tp_free` and `tp_new` directly rather than using the workarounds here.
* `ScandirIterator_new` and `DirEntry_new` are the same function.

encukou:
I'm writing a patch to address those comments.

encukou:
I added some commits here. The core dev is ending, and I won't have time to look at this issue for the next few days (or weeks, unfortunately, depending on work load).
Please take these as comments/conversation starters, rather than the one true way of doing things. Feel free to rebase/modify them.

eduardo-elizondo:
@encukou thanks for all the reviews! I'll update this to address all the comments

eduardo-elizondo:
> Module-level constant strings should use PyUnicode_InternFromString rather than PyUnicode_FromString
Agreed

> I'd rather bring _Py_IDENTIFIER(__fspath__) back (for now) than reimplement _PyObject_LookupSpecial.
A more systematic solution would be to implement another C-API to hide the abstraction there. I rather not expand the C-API given that this pattern is barely present in the stdlib.

> In the stdlib, I'd prefer using tp_free and tp_new directly rather than using the workarounds here.
Under PEP384, `tp_free` and `tp_new` are not public api therefore, they can't be used directly. Also, with the introduced changes, os.stat_result now fails. I will revert that back to what I originally had

> ScandirIterator_new and DirEntry_new are the same function.
Let's merge them

encukou:
> I think there's two options: 1) Solve this by creating a new C-API that handles this without a _Py_IDENTIFIER. Unfortunately, this will add yet another C-API function. 2) Not expand the C-API and fix the pattern at the extension level. There's only 2 C Extensions in the stdlib that use this pattern.
Given that this is not a very common pattern, I think it's best to go for approach # 2. I didn't modify anything just yet but I'd like to hear your thoughts!

As we remove statics, there will need to be a generic solution for _Py_IDENTIFIER. Eric & the team are thinking about how to handle them.
Let's not change things in a single module.

> Unfortunately, under PEP384, tp_free and tp_new are not public api so they can't be used

PEP 384, the stable ABI, is not relevant to the standard library.
It is good to „dogfood“ the stable ABI in the stdlib, to make sure the ABI is usable – and to catch places where it can be improved, like this. We should definitely expand the stable ABI to allow handling slots more elegantly.
But I don't want to work around the stable ABI limitations.

eduardo-elizondo:
Hey Petr, good to hear back from you!

> As we remove statics, there will need to be a generic solution for _Py_IDENTIFIER. Eric & the team are thinking about how to handle them. Let's not change things in a single module.

Outside of the special case of `_PyObject_LookupSpecial `, we solved the `_Py_IDENTIFIER` problem by setting a unicode object in the module state and using it with non `_Py_IDENTIFIER` C-APIs. I.e. `PyObject_GetAttr` instead of `_PyObject_GetAttrId`.

> PEP 384, the stable ABI, is not relevant to the standard library.

That's something that I've been trying to get across but I'm probably not doing a good job at it 😛. The really nice benefit of PEP384 is that it hides all the implementation details of CPython. This allows any other implementation to use any of these modules only by using the C-API. Using any type (i.e. `tp_free`) member ties the module to the CPython implementation only. That's why, this is relevant to any C Extension module.

Let me know if it makes sense and/or if you have other questions!

> We should definitely expand the stable ABI to allow handling slots more elegantly.

Agreed! Any suggestions on the proper avenues to do that?

encukou:
Sorry for the delay – I needed to cancel my CPython day last week :(

> The really nice benefit of PEP384 is that it hides all the implementation details of CPython. This allows any other implementation to use these modules as well. So adding these changes here allows 1) to set a pattern for other C Extensions to use PEP384, and 2) enables these extensions to be used by any other implementation.

Agreed. However, the *set a pattern for other C Extensions* is an issue here.
I find this PR quite elegant, and I'm happy that the stable ABI is nice to use – with one exception, the setting of structseq slots via `PyDescr_NewClassMethod` and `PyObject_SetAttrString`. All the other pieces set a *good* example, but this one is an example of how to get around a limited API.
I want to find a better API *before* this is in the standard library and sets an example for everyone else.

Can we merge this PR with commit c1a6d030b199aba4fb7599ff8bc5191ffbdc7e37 reverted, and leave that bit for future work?
It's a big PR and I think it would help to have the majority of it merged.

>> We should definitely expand the stable ABI to allow handling slots more elegantly.
>
> Agreed! Any suggestions on the proper avenues to do that?

Brainstorming ideas:
* Add a new version of `PyStructSequence_NewType` that would take additional slot information (setting a precedent for other `*_NewType` functions)
* Add a "set" counterpart to `PyType_GetSlot` (and think through the implications & limitations of that)

eduardo-elizondo:
> Can we merge this PR with commit c1a6d03 reverted, and leave that bit for future work?

Sounds like a plan! I'll update this PR in a bit :)

eduardo-elizondo:
@encukou done!

encukou (approved):
I had a few cosmetic nitpicks, but I don't think they need a review round-trip, so I'll just merge with them.

Thank you very much for this change!

Modules/posixmodule.c
```diff
@@ -537,7 +538,7 @@ _Py_Uid_Converter(PyObject *obj, void *p)
     if (index == NULL) {
         PyErr_Format(PyExc_TypeError,
                      "uid should be integer, not %.200s",
-                     Py_TYPE(obj)->tp_name);
+                     _PyType_Name(Py_TYPE(obj)));
```
vstinner, line 541: tp_name and _PyType_Name() are not the same. There were some discussions between fully qualified names (module.name1.name2.name3) vs short name (name3), but no consensus was found. Well, I should reopen a discussion ;-) I prefer to provide longest names in error messages to ease debug.

eduardo-elizondo: @vstinner I actually debated over that for a bit! I think we are missing something like: `PyType_QualName`. I can do a quick PR for that but I don't know if there will be push back from other devs. Thoughts?

encukou: Reopening the discussion sounds like a good idea.

encukou: Was it reopened? A discussion could be useful in https://bugs.python.org/issue42035.

PR #15895, opened 2019-09-11
bpo-38096: Clean up the "struct sequence" / "named tuple" docs

rhettinger:
https://bugs.python.org/issue38096

### DIFF
--- a/Doc/glossary.rst
+++ b/Doc/glossary.rst
@@ -739,17 +739,28 @@ Glossary
       also :term:`immutable`.
 
    named tuple
-      Any tuple-like class whose indexable elements are also accessible using
-      named attributes (for example, :func:`time.localtime` returns a
-      tuple-like object where the *year* is accessible either with an
-      index such as ``t[0]`` or with a named attribute like ``t.tm_year``).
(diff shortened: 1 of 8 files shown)


pganssle (approved):
LGTM.

Doc/glossary.rst
```diff
@@ -739,17 +739,28 @@ Glossary
...
+           1024
+           >>> sys.float_info.max_exp              # named field access
+           1024
+           >>> isinstance(sys.float_info, tuple)   # kind of tuple
+           True
+
+      Some named tuples are built-in types (such as the above examples).
+      Alternatively, a named tuple can be created from a regular class
+      definition that inherits from :class:`tuple` and that defines named
+      fields.  Such as class can be written by hand or it can be created with
```
tim-one, line 760: Typo:  "as" in "Such as class" should be "a".

tim-one (commented):
Looks great!  Thank you.  Some comments:

- Left a comment pointing out a typo.

- Would have liked to see a blurb added to `structseq.c` pointing out that structseq is an internal CPython implementation detail, and that docs for modules using structseqs should call them "named tuples".

PR #15907, opened 2019-09-11
bpo-35066: Make trailing percent test more portable.

benjaminp:
Different libc implementations have different behavior when presented with trailing % in strftime strings. To make test_strftime_trailing_percent more portable, compare the output of datetime.strftime directly to that of time.strftime rather than hardcoding.

https://bugs.python.org/issue35066

Automerge-Triggered-By: @pganssle

Automerge-Triggered-By: @pganssle

### DIFF
--- a/Lib/test/datetimetester.py
+++ b/Lib/test/datetimetester.py
@@ -1449,15 +1449,20 @@ class TestDate(HarmlessMixedComparison, unittest.TestCase):
         t.strftime("%f")
 
     def test_strftime_trailing_percent(self):
-        # bpo-35066: make sure trailing '%' doesn't cause
-        # datetime's strftime to complain
+        # bpo-35066: Make sure trailing '%' doesn't cause datetime's strftime to
+        # complain. Different libcs have different handling of trailing
+        # percents, so we simply check datetime's strftime acts the same as
+        # time.strftime.
         t = self.theclass(2005, 3, 2)
         try:
             _time.strftime('%')
         except ValueError:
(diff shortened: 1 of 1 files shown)


pganssle (approved):

Lib/test/datetimetester.py
```diff
@@ -1456,8 +1456,11 @@ def test_strftime_trailing_percent(self):
             _time.strftime('%')
         except ValueError:
             self.skipTest('time module does not support trailing %')
-        self.assertEqual(t.strftime('%'), '%')
-        self.assertEqual(t.strftime("m:%m d:%d y:%y %"), "m:03 d:02 y:05 %")
+        self.assertEqual(t.strftime('%'), _time.strftime('%'), t.timetuple())
```
pganssle:
```suggestion
        self.assertEqual(t.strftime('%'), _time.strftime('%', t.timetuple()))
```

pganssle (changes requested):

benjaminp:

pganssle (approved):

pganssle:
Per conversation at the core dev sprint - I'm mildly uncomfortable about changing this because the aspiration is that this behavior should work on all platforms, and we should probably set platform-specific expected failures instead of changing this, but *practically speaking*, this might break a bunch of platforms while we're not really prepared to actually create a platform-independent version of this, so for now I agree that we should do this more-portable version of the test. :disappointed:

PR #15921, opened 2019-09-11
bpo-37326: Include libffi license when available

zooba:
https://bugs.python.org/issue37326


zooba:
I also pushed the current LICENSE file to our binaries repo, so it'll be included in the generated LICENSE.txt file.

zware (approved):

PR #15924, opened 2019-09-11
bpo-36634: Fixes activate.bat when existing values contain double quotes

zooba:
https://bugs.python.org/issue36634


zware (approved):
LGTM

PR #15947, opened 2019-09-11
bpo-38093: Correctly returns AsyncMock for async subclasses.

lisroach:
We noticed some funny behavior when mocking a async context manager and async iterator. Mainly that you needed to mock the CM or iterator as a MagicMock and additionally set `__aenter__` and `__aexit__` on the mock manually. Trying to use an AsyncMock would cause your code to fail. This is counterintuitive and makes the code hard to write.

Fixing the bad `if` statement in `_def_get_child` fixes AsyncMocks, now they work like you would think and don't crash. 
You can now mock a async context manager or async iterator with either AsyncMock or MagicMock and both function the same.

https://bugs.python.org/issue38093


Lib/unittest/test/testmock/testasync.py
```diff
@@ -363,22 +363,58 @@ def test_add_side_effect_iterable(self):
...
+        def __enter__(self, *args, **kwargs):
+            return self
+
+        def __exit__(self, *args, **kwargs):
+            pass
+
+    class ProductionCode:
+        # Example real-world(ish) code
+        async def main(self):
+            async with self.session.post('https://python.org') as response:
```
tirkarthi, line 400: Do we need to declare `self.session`?

lisroach: Oops, yeah I took that code out incorrectly.

Lib/unittest/mock.py
```diff
@@ -1904,6 +1904,8 @@ def method(self, /, *args, **kw):
     '__str__': lambda self: object.__str__(self),
     '__sizeof__': lambda self: object.__sizeof__(self),
     '__fspath__': lambda self: f"{type(self).__name__}/{self._extract_mock_name()}/{id(self)}",
+    '__aenter__': lambda self: AsyncMockMixin._mock_call(self),
```
tirkarthi: `AsyncMockMixin._mock_call` is an async def so I guess it should be awaited. There are some warnings but not sure if they are related to this.

```
./python.exe -m unittest Lib.unittest.test.testmock.testasync
........../Users/karthikeyansingaravelan/stuff/python/cpython/Lib/unittest/mock.py:740: RuntimeWarning: coroutine 'AsyncMockMixin._mock_call' was never awaited
  return object.__setattr__(self, name, value)
RuntimeWarning: Enable tracemalloc to get the object allocation traceback
...../Users/karthikeyansingaravelan/stuff/python/cpython/Lib/unittest/mock.py:2084: RuntimeWarning: coroutine 'AsyncMockMixin._mock_call' was never awaited
  self.name = name
RuntimeWarning: Enable tracemalloc to get the object allocation traceback
.....................................
Ran 52 tests in 0.508s

OK
```

lisroach:
Ah the doctest failure is what I added the #TODO about, I'll need to update the docs as well. I'll fix that TODO today or tomorrow so the failures aren't super opaque. 

Looking into fixing the errors now.

asvetlov (approved):
please backport to 3.8 on merging

Lib/unittest/mock.py
```diff
@@ -1867,7 +1870,7 @@ def _patch_stopall():
     '__reduce__', '__reduce_ex__', '__getinitargs__', '__getnewargs__',
     '__getstate__', '__setstate__', '__getformat__', '__setformat__',
     '__repr__', '__dir__', '__subclasses__', '__format__',
-    '__getnewargs_ex__', '__aenter__', '__aexit__', '__anext__', '__aiter__',
```
tirkarthi, line 1870: Can we have a test for the default values of these functions? There is a behavior change in 3.8b4 and now where `__aexit__` for MagicMock returns False and now it returns a coroutine. The docs can also be updated in the "mock and default values" section since these don't have default values : https://docs.python.org/3.8/library/unittest.mock.html?highlight=asyncmock#unittest.mock.NonCallableMagicMock . I guess this also makes updating `_return_values` dictionary since the default values are not needed now for these and can be removed.

```
➜  cpython git:(pr_15947) ✗ python3.8
Python 3.8.0b4 (v3.8.0b4:d93605de72, Aug 29 2019, 21:47:47)
[Clang 6.0 (clang-600.0.57)] on darwin
>>> from unittest.mock import MagicMock
>>> m = MagicMock()
>>> m.__aexit__()
False
➜  cpython git:(pr_15947) ✗ ./python.exe
Python 3.9.0a0 (heads/pr_15947:f476b5845a, Sep 12 2019, 10:52:24)
>>> from unittest.mock import MagicMock
>>> m = MagicMock()
>>> m.__aexit__()
<coroutine object AsyncMockMixin._mock_call at 0x107005d40>
```

Doc/library/unittest.mock-examples.rst
```diff
@@ -280,36 +280,37 @@ function returns is what the call returns:
...
 
 
 Mocking asynchronous context manager
 ~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
 
-Since Python 3.8, ``MagicMock`` has support to mock
+Since Python 3.8, ``AsyncMock`` has support to mock
 :ref:`async-context-managers` through ``__aenter__`` and ``__aexit__``. The
-return value of ``__aenter__`` is an :class:`AsyncMock`.
+return value of ``__aenter__`` is an async function.
```
ezio-melotti: I think it would be better to say something like:
```
By default, ``__aenter__`` is set to an :class:`AsyncMock` that return an async function.
```
We want to make the distinction between what the methods are and what they return clear.

Lib/unittest/test/testmock/testasync.py
```diff
@@ -378,22 +378,67 @@ def test_add_side_effect_iterable(self):
...
+        self.assertTrue(hasattr(cm, "__aenter__"))
+        self.assertTrue(hasattr(cm, "__aexit__"))
+        self.assertTrue(asyncio.iscoroutinefunction(cm.__aexit__))
+
+    def test_set_return_value_of_aenter_magicmock(self):
+        pc = self.ProductionCode()
+        pc.session = MagicMock(name='sessionmock')
+        cm = MagicMock(name='magic_cm')
+        response = AsyncMock(name='response')
+        response.json = AsyncMock(return_value={'json':123})
```
tirkarthi: PEP8 nit regarding space here and in below dict declarations.

```suggestion
        response.json = AsyncMock(return_value={'json': 123})
```

Lib/unittest/test/testmock/testasync.py
```diff
@@ -378,22 +378,67 @@ def test_add_side_effect_iterable(self):
...
+        cm.__aenter__.return_value = response
+        pc.session.post.return_value = cm
+        result = asyncio.run(pc.main())
+        self.assertEqual(result, {'json':123})
 
     def test_mock_supports_async_context_manager(self):
         called = False
         instance = self.WithAsyncContextManager()
-        mock_instance = MagicMock(instance)
+        # TODO: Horrible error message if you use a MagicMock here
```
tirkarthi: Please open a bpo in case it will not be part of the PR.

Lib/unittest/test/testmock/testasync.py
```diff
@@ -474,7 +518,38 @@ def __aiter__(self):
...
+        instance = self.WithAsyncIterator()
+        mock_instance = AsyncMock(instance)
+
+        self.assertEqual(asyncio.iscoroutine(instance.__aiter__),
+                         asyncio.iscoroutine(mock_instance.__aiter__))
+        self.assertEqual(asyncio.iscoroutine(instance.__anext__),
+                         asyncio.iscoroutine(mock_instance.__anext__))
+
+        iterator = instance.__aiter__()
+        if asyncio.iscoroutine(iterator):
```
tirkarthi: The test is not parameterized or have subtests so would it be better to have an assertion here like `self.assertTrue(asyncio.iscoroutine(iterator))` ?

Lib/unittest/test/testmock/testasync.py
```diff
@@ -474,7 +518,38 @@ def __aiter__(self):
...
+            iterator = asyncio.run(iterator)
+
+        mock_iterator = mock_instance.__aiter__()
+        if asyncio.iscoroutine(mock_iterator):
```
tirkarthi: The test is not parameterized or have subtests so would it be better to have an assertion here like self.assertTrue(asyncio.iscoroutine(iterator)) ?

tirkarthi (commented):
Docs could be updated about `return_value` change for MagicMock with respect to async functions and also a test to make sure we don't regress in the future.

Lib/unittest/test/testmock/testasync.py
```diff
@@ -505,20 +580,34 @@ def test_mock_async_for(self):
...
         expected = ["FOO", "BAR", "BAZ"]
-        with self.subTest("iterate through default value"):
+        with self.subTest("iterate through default value with magicmock"):
             mock_instance = MagicMock(self.WithAsyncIterator())
             self.assertEqual([], asyncio.run(iterate(mock_instance)))
 
-        with self.subTest("iterate through set return_value"):
+        with self.subTest("iterate through default value using asyncmock"):
+            mock_instance = AsyncMock(self.WithAsyncIterator())
+            self.assertEqual([], asyncio.run(iterate(mock_instance)))
```
ezio-melotti: The usual convention is to have `assertEqual(actual, expected)`, but here and in the rest of the function the values are inverted.  The fix for this might belong to a different PR.

Lib/unittest/mock.py
lisroach: Hmm this is a good point about `__aexit__`, I think it should have a default `return_value` of False (which is in the `_return_values` dict, but maybe needs to be added to the `_calculate_return_value` instead.

This section specifically is the "non-default" values, so I don't think it make sense to update that section of the docs, and there is a unit test for the default values that are set in `test_magicmock_defaults`. I will update `__aexit__` and add it there.

Lib/unittest/test/testmock/testasync.py
```diff
@@ -505,20 +580,34 @@ def test_mock_async_for(self):
             return accumulator
 
         expected = ["FOO", "BAR", "BAZ"]
-        with self.subTest("iterate through default value"):
+        with self.subTest("iterate through default value with magicmock"):
             mock_instance = MagicMock(self.WithAsyncIterator())
             self.assertEqual([], asyncio.run(iterate(mock_instance)))
 
-        with self.subTest("iterate through set return_value"):
+        with self.subTest("iterate through default value using asyncmock"):
```
ezio-melotti: The only difference between this `with` and the one above is the use of `AsyncMock` instead of `MagickMock`.  Can we use `for MockClass in MagicMock, AsyncMock: ...` and remove the duplication?

Lib/unittest/test/testmock/testasync.py
```diff
@@ -474,7 +518,38 @@ def __aiter__(self):
...
+        self.assertEqual(result, [1, 2, 3])
+
+    def test_mock_aiter_and_anext_asyncmock(self):
+        instance = self.WithAsyncIterator()
+        mock_instance = AsyncMock(instance)
+
+        self.assertEqual(asyncio.iscoroutine(instance.__aiter__),
+                         asyncio.iscoroutine(mock_instance.__aiter__))
+        self.assertEqual(asyncio.iscoroutine(instance.__anext__),
+                         asyncio.iscoroutine(mock_instance.__anext__))
```
ezio-melotti: I think it would be better to use `assertTrue` here, and maybe just add a comment to clarify that we are checking that the `instance` and the `mock_instance` behave the same.

Lib/unittest/test/testmock/testasync.py
```diff
@@ -378,22 +378,67 @@ def test_add_side_effect_iterable(self):
...
+    def test_set_return_value_of_aenter_asyncmock(self):
+        pc = self.ProductionCode()
+        pc.session = MagicMock(name='sessionmock')
+        cm = AsyncMock(name='async_cm')
+        response = AsyncMock(name='response')
+        response.json = AsyncMock(return_value={'json':123})
+        cm.__aenter__.return_value = response
+        pc.session.post.return_value = cm
+        result = asyncio.run(pc.main())
+        self.assertEqual(result, {'json':123})
```
ezio-melotti: I would turn this in a helper method that accepts a `MockClass` as argument and call it from both these methods passing `MagicMock` and `AsyncMock`.

ezio-melotti (changes requested):

Lib/unittest/mock.py
lisroach: Ah well, this is actually another case of "called" vs "awaited". Once you await `__aexit__` it returns False correctly, just calling prints the type of `__aexit__` which is a coroutine object. 

So I want to leave the `__aexit__` in _return_value and I will update the test. Do you think that would resolve this?

Lib/unittest/test/testmock/testasync.py
lisroach: Actually I think this is fixed, will remove comment and make it a subTest to test both classes.

tirkarthi: Thanks for the clarification. Mine was due to the comment over todo https://github.com/python/cpython/pull/15947#issuecomment-530401838. I would be happy if it was fixed then.

Lib/unittest/mock.py
tirkarthi: Ah okay so it's a case where once `__aexit__` is awaited then it returns the calculated value of False. Sorry, I am not sure of async internals to see if `__aexit__` can be used like `await __aexit__()` or makes sense but I can understand in terms of the PR context with mock that the coroutine needs to be always awaited.

asvetlov: `await cm.__aexit__()` is a legal call.

Doc/library/unittest.mock-examples.rst
```diff
@@ -280,36 +280,38 @@ function returns is what the call returns:
...
 
 Mocking asynchronous context manager
 ~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
 
-Since Python 3.8, ``MagicMock`` has support to mock
-:ref:`async-context-managers` through ``__aenter__`` and ``__aexit__``. The
-return value of ``__aenter__`` is an :class:`AsyncMock`.
+Since Python 3.8, ``AsyncMock`` and ``MagicMock`` have support to mock
+:ref:`async-context-managers` through ``__aenter__`` and ``__aexit__``.
+By default, ``__aenter__`` is an ``AsyncMock`` that returns an async function.
```
asvetlov: Heh, `__aexit__` is an `AsyncMock` too, isn't it?
```suggestion
By default, ``__aenter__`` and ``__aexit__`` are ``AsyncMock`` instances that return an async function.
```

asvetlov (approved):
LGTM.
A small nitpick for documentation wording, feel free to ignore if you disagree.

Lib/unittest/test/testmock/testasync.py
```diff
@@ -378,35 +378,85 @@ def test_add_side_effect_iterable(self):
...
+        cm = MagicMock(name='magic_cm')
+        self.assertTrue(hasattr(cm, "__aenter__"))
+        self.assertTrue(hasattr(cm, "__aexit__"))
+
+    def test_magic_methods_are_async_functions(self):
+        cm = MagicMock(name='magic_cm')
+        self.assertTrue(asyncio.iscoroutinefunction(cm.__aenter__))
+        self.assertTrue(asyncio.iscoroutinefunction(cm.__aexit__))
+        # These should pass but cause warnings to be raised
+        # self.assertTrue(inspect.isawaitable(cm.__aenter__()))
```
tirkarthi: Can we just await on the method for the tests? Would prefer not having commented code.

```
+        for meth in [cm.__aenter__, cm.__aexit__]:
+            awaitable = meth()
+            self.assertTrue(inspect.isawaitable(awaitable))
+            await awaitable
```

tirkarthi:
Slightly offtopic but there is `unittest.IsolatedAsyncioTestCase` (https://docs.python.org/3.8/library/unittest.html#unittest.IsolatedAsyncioTestCase) so it can remove a bunch of boilerplate to have async functions just to test await statements and using `asyncio.run`. I would try opening a separate issue for this for discussion.

Lib/unittest/test/testmock/testasync.py
lisroach: This will fail because we are trying to await outside of an async function :( 

I haven't looked at IsolatedAsyncioTestCase yet! Would that allow us to have an `async def test_blah` that would solve this?

Doc/library/unittest.mock-examples.rst
lisroach: You are right, I will add that 😆

Lib/unittest/test/testmock/testasync.py
tirkarthi: > I haven't looked at IsolatedAsyncioTestCase yet! Would that allow us to have an async def test_blah that would solve this?

I just documented it yesterday ;) yes, you can write `async def test_blah` and that would be collected by test runner. It's like a drop in replacement to `TestCase` and in addition also takes coroutines/async functions as test cases.

lisroach: Perfect! That would make testing AsyncMock more robust, I've found a good number of places I would like to test with `await` but haven't been able to. Let's make it another issue, I can take these commented lines out and they will eventually get added back in with the new `IsolatedAsyncioTestCase`.

Lib/unittest/test/testmock/testasync.py
```diff
@@ -474,27 +521,31 @@ def __aiter__(self):
...
-            iterator = asyncio.run(iterator)
-
-        mock_iterator = mock_instance.__aiter__()
-        if asyncio.iscoroutine(mock_iterator):
-            mock_iterator = asyncio.run(mock_iterator)
+    def test_aiter_set_return_value(self):
+        mock_iter = AsyncMock(name="tester")
+        mock_iter.__aiter__.return_value = [1, 2, 3]
+        async def main():
+            async for i in mock_iter:
```
tirkarthi: Slightly confused over if this is needed since we just return a list comprehension and assert only once.

Lib/unittest/test/testmock/testasync.py
asvetlov: `IsolatedAsyncioTestCase` is not used by any CPython test yet (except self-testing) but I want to rewrite the most of asyncio tests with its usage eventually.

Doc/library/unittest.mock-examples.rst
```diff
@@ -280,36 +280,39 @@ function returns is what the call returns:
...
-    >>> mock_instance = MagicMock(AsyncContextManager())
+    ...
+    >>> mock_instance = MagicMock(AsyncContextManager())  # AsyncMock also works here
     >>> async def main():
     ...     async with mock_instance as result:
     ...         pass
+    ...
     >>> asyncio.run(main())
     >>> mock_instance.__aenter__.assert_called_once()
     >>> mock_instance.__aexit__.assert_called_once()
```
ezio-melotti: Eventually we probably want to replace these either with `assert_aiwated_once()` or by returning a result and checking for that.

Lib/unittest/mock.py
```diff
@@ -988,9 +988,13 @@ def _get_child_mock(self, /, **kw):
         _type = type(self)
         if issubclass(_type, MagicMock) and _new_name in _async_method_magics:
             klass = AsyncMock
-        if issubclass(_type, AsyncMockMixin):
+        elif _new_name  in _sync_async_magics:
+            # Special case these ones b/c users will assume they are async,
+            # but they are actually sync
```
ezio-melotti: An example of such method in the comment would probably help.  Is this about e.g. `__aiter__` that sounds async but is actually sync?
Edit: now I saw where `_sync_async_magics` is defined that `__aiter__` is the only one.

Lib/unittest/mock.py
ezio-melotti: For reference, this is how async cms work: https://www.python.org/dev/peps/pep-0492/#new-syntax
Both `__aenter__` and `__aexit__` are awaited.

Lib/unittest/test/testmock/testasync.py
```diff
@@ -378,35 +378,82 @@ def test_add_side_effect_iterable(self):
...
+
+    def test_magicmock_has_async_magic_methods(self):
+        cm = MagicMock(name='magic_cm')
+        self.assertTrue(hasattr(cm, "__aenter__"))
+        self.assertTrue(hasattr(cm, "__aexit__"))
+
+    def test_magic_methods_are_async_functions(self):
+        cm = MagicMock(name='magic_cm')
+        self.assertTrue(asyncio.iscoroutinefunction(cm.__aenter__))
+        self.assertTrue(asyncio.iscoroutinefunction(cm.__aexit__))
```
ezio-melotti, line 420: These are `AsyncMock`s, right?
So this is indirectly testing that `AsyncMock`  are coroutine functions?
(not saying that the test is useless, but maybe a short comment telling that those are `AsyncMock`s and that `AsyncMock`s are coroutines would help)

Lib/unittest/test/testmock/testasync.py
```diff
@@ -378,35 +378,82 @@ def test_add_side_effect_iterable(self):
...
+                inner_test(mock_type)
 
     def test_mock_supports_async_context_manager(self):
-        called = False
-        instance = self.WithAsyncContextManager()
-        mock_instance = MagicMock(instance)
+        def inner_test(mock_type):
+            called = False
+            instance = self.WithAsyncContextManager()
+            mock_instance = mock_type(instance)
```
ezio-melotti: What about calling these `cm` and `mock_cm`?

Lib/unittest/test/testmock/testasync.py
```diff
@@ -378,35 +378,82 @@ def test_add_side_effect_iterable(self):
...
+
+            async def use_context_manager():
+                nonlocal called
+                async with mock_instance as result:
```
ezio-melotti: `async with mock_cm:`

Lib/unittest/test/testmock/testasync.py
```diff
@@ -378,35 +378,82 @@ def test_add_side_effect_iterable(self):
...
+                    called = True
+                return result
```
ezio-melotti, line 448: `return mock_cm`

Lib/unittest/test/testmock/testasync.py
```diff
@@ -378,35 +378,82 @@ def test_add_side_effect_iterable(self):
...
 
-        async def use_context_manager():
-            nonlocal called
-            async with mock_instance as result:
-                called = True
-            return result
+            result = asyncio.run(use_context_manager())
+            self.assertTrue(called)
+            self.assertTrue(mock_instance.__aenter__.called)
+            self.assertTrue(mock_instance.__aexit__.called)
```
ezio-melotti: This should probably check if they have been awaited.

Lib/unittest/test/testmock/testasync.py
```diff
@@ -474,27 +521,31 @@ def __aiter__(self):
...
-        iterator = instance.__aiter__()
-        if asyncio.iscoroutine(iterator):
-            iterator = asyncio.run(iterator)
-
-        mock_iterator = mock_instance.__aiter__()
-        if asyncio.iscoroutine(mock_iterator):
-            mock_iterator = asyncio.run(mock_iterator)
+    def test_aiter_set_return_value(self):
+        mock_iter = AsyncMock(name="tester")
+        mock_iter.__aiter__.return_value = [1, 2, 3]
```
ezio-melotti, line 532: We talked about briefly this at the sprint:
* check what happens with a genexp
* check what happens without setting a return_value (should still be iterable)

Lib/unittest/test/testmock/testasync.py
```diff
@@ -474,27 +521,31 @@ def __aiter__(self):
...
+        async def main():
+            async for i in mock_iter:
+                return [i async for i in mock_iter]
+        result = asyncio.run(main())
+        self.assertEqual(result, [1, 2, 3])
+
+    def test_mock_aiter_and_anext_asyncmock(self):
+        def inner_test(mock_type):
+            instance = self.WithAsyncIterator
```
ezio-melotti: Here you are missing the `()`, you have them at line 438 459.
Is this intentional?

Lib/unittest/test/testmock/testmagicmethods.py
```diff
@@ -286,6 +287,8 @@ def test_magicmock_defaults(self):
         self.assertEqual(math.trunc(mock), mock.__trunc__())
         self.assertEqual(math.floor(mock), mock.__floor__())
         self.assertEqual(math.ceil(mock), mock.__ceil__())
+        self.assertTrue(asyncio.iscoroutinefunction(mock.__aexit__))
+        self.assertTrue(asyncio.iscoroutinefunction(mock.__aenter__))
```
ezio-melotti, line 319: Maybe check `self.assertIsInstance(mock.__aexit__, AsyncMock)` too?

All these tests should also be applied to `AsyncMock`.

ezio-melotti (changes requested):
More comments, mostly minor things, some unrelated things that we discussed briefly IRL -- you know what to do.

Doc/library/unittest.mock-examples.rst
lisroach: Agreed, I will update them in this PR if that's okay.

Lib/unittest/test/testmock/testasync.py
lisroach: Ah, you're right, didn't notice the double loop

Lib/unittest/test/testmock/testasync.py
lisroach: Fixed, it makes no difference so tests don't fail but want to keep them consistent.

Lib/unittest/test/testmock/testasync.py
lisroach: Leaving as result as discussed because the result object is different than the original object since it has been mocked.

Lib/unittest/test/testmock/testasync.py
lisroach: Added a test for directly checking AsyncMock instances and a comment to make this more clear.

Lib/unittest/test/testmock/testasync.py
lisroach: This is not working quite right yet, but it should be fixed with issue38163

ezio-melotti (approved):

Misc/NEWS.d/next/Library/2019-09-11-14-45-30.bpo-38093.yQ6k7y.rst
```diff
@@ -0,0 +1,2 @@
+Fixes AsyncMock so it doesn't crash when used with AsynContextManagers
```
ezio-melotti:
```suggestion
Fixes AsyncMock so it doesn't crash when used with AsyncContextManagers
```

PR #15949, opened 2019-09-11
bpo-33166: Change os.cpu_count to return active (real) processors

zooba:
https://bugs.python.org/issue33166

### DIFF
--- a/Modules/posixmodule.c
+++ b/Modules/posixmodule.c
@@ -12179,23 +12179,8 @@ os_cpu_count_impl(PyObject *module)
 {
     int ncpu = 0;
 #ifdef MS_WINDOWS
-    /* Vista is supported and the GetMaximumProcessorCount API is Win7+
-       Need to fallback to Vista behavior if this call isn't present */
-    HINSTANCE hKernel32;
-    static DWORD(CALLBACK *_GetMaximumProcessorCount)(WORD) = NULL;
-    Py_BEGIN_ALLOW_THREADS
-    hKernel32 = GetModuleHandleW(L"KERNEL32");
-    *(FARPROC*)&_GetMaximumProcessorCount = GetProcAddress(hKernel32,
-        "GetMaximumProcessorCount");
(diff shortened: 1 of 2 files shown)


zooba:
Note that I added the prototype rather than changing the Windows version in pyconfig.h. This is so it's safe to backport to 3.8.

In a later PR I'll update the API version.

Modules/posixmodule.c
```diff
@@ -12179,23 +12179,8 @@ os_cpu_count_impl(PyObject *module)
...
-    Py_END_ALLOW_THREADS
-    if (_GetMaximumProcessorCount != NULL) {
-        ncpu = _GetMaximumProcessorCount(ALL_PROCESSOR_GROUPS);
-    }
-    else {
-        SYSTEM_INFO sysinfo;
-        GetSystemInfo(&sysinfo);
-        ncpu = sysinfo.dwNumberOfProcessors;
-    }
+    DWORD WINAPI GetActiveProcessorCount(WORD group);
```
zware, line 12183: This could use the comment from the PR discussion in the code; it seems out of place.

zware: The other option would be to go ahead and set the minimum to Windows 7 in pyconfig.h and backport it to 3.8; as far as I know there are no other things that we want 8+ for yet.

zware (changes requested):

Modules/posixmodule.c
zooba:
```suggestion
    /* Declare prototype here to avoid pulling in all of the Win7 APIs in 3.8 */
    DWORD WINAPI GetActiveProcessorCount(WORD group);
```

zooba:
I don't want to backport the WINNT_VER change because we have so much stuff that tries to auto-detect macros, and that may introduce new macros. It's a little late in the release cycle for that.

zooba:
I have made requested changes; please review again

zware (approved):
LGTM

PR #15951, opened 2019-09-11
bpo-32592: Set Windows 8 as the minimum required version for API support

zooba:
https://bugs.python.org/issue32592

### DIFF
--- a/Misc/NEWS.d/next/Windows/2019-09-11-14-51-56.bpo-32592.jvQMD9.rst
+++ b/Misc/NEWS.d/next/Windows/2019-09-11-14-51-56.bpo-32592.jvQMD9.rst
@@ -1,0 +1,1 @@
+Set Windows 8 as the minimum required version for API support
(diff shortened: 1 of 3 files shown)


zware:
This is going to require a change to the buildbot config to turn off the 3.x builder on the Windows 7 bot(s).  I'd rather wait on this until that change is in than to have to make the change to fix the breakage.

zooba:
It shouldn't, as the minimum API isn't actually enforced anywhere except at compile time, and all the Win10 SDKs have the Win8 APIs available.

zware:
I've started a buildbot-custom build to check that.

zooba:
Don't merge this before #15949 or it'll break the other one. (That PR is being backported, and then I'll have to rebase this one...)

zware:
The [Win7 SP1 buildbot](https://buildbot.python.org/all/#/builders/42/builds/58) is happy, so I'll merge this after #15949.

zware (approved):

vstinner:
Thanks!

PR #15955, opened 2019-09-11
gh-60716: Update judgment method for jpeg.

derodero24:
Changed to judgment method for jpeg based on file signature.
https://en.wikipedia.org/wiki/List_of_file_signatures

https://bugs.python.org/issue37629

* Issue: gh-60716


the-knights-who-say-ni:

## Recognized GitHub username

@derodero24

derodero24:
> Don’t add a jpeg file, create a fake jpeg file during the test and you remove it after.

Excuse me.
I couldn't find the method creating a fake jpeg file.
Could you tell me where it is?
Thanks.

matrixise:
Yep, open a binary file and store the magic number in the file, or the header.

Lib/imghdr.py
```diff
@@ -36,7 +36,7 @@ def what(file, h=None):
 
 def test_jpeg(h, f):
     """JPEG data in JFIF or Exif format"""
-    if h[6:10] in (b'JFIF', b'Exif'):
+    if h[:2] == b'\xff\xd8':
```
matrixise: And what’s the behaviour with the existing files?

matrixise (changes requested):

Lib/imghdr.py
derodero24: It returns True only for `python.jpg` and False otherwise.
This is the same as the previous code.

derodero24:
> Yep, open a binary file and store the magic number in the file, or the header.

The magic number I'm focusing on is `\xff\xd8`.
How should I store this?

matrixise:
store it into a file, but I would like to be sure that all the tests pass with any jpg file.

derodero24:
You mean, I can replace the header of existing jpg file?
I am concerned that this is a slightly special jpeg file.

I've already tested with hundreds of randomly prepared jpg files, but there is no mistake.
At least it should be better than current code.

derodero24:

derodero24:
@matrixise Could you check for changes?

StanFromIreland:
This module was removed in 3.13.

PR #15957, opened 2019-09-11
bpo-38113: Update the Python-ast.c generator to PEP384

DinoV:
Summary: This mostly migrates Python-ast.c to PEP384 and removes all statics from the whole file. This modifies the generator itself that generates the Python-ast.c.  It leaves in the usage of _PyObject_LookupAttr even though it's not fully PEP384 compatible (this could always be shimmed in by anyone who needs it).

https://bugs.python.org/issue38113

### DIFF
--- a/Parser/asdl_c.py
+++ b/Parser/asdl_c.py
@@ -78,14 +78,18 @@ class EmitVisitor(asdl.VisitorBase):
     def __init__(self, file):
         self.file = file
         self.identifiers = set()
+        self.singletons = set()
+        self.types = set()
         super(EmitVisitor, self).__init__()
 
     def emit_identifier(self, name):
-        name = str(name)
-        if name in self.identifiers:
-            return
-        self.emit("_Py_IDENTIFIER(%s);" % name, 0)
-        self.identifiers.add(name)
+        self.identifiers.add(str(name))
+
+    def emit_singleton(self, name):
+        self.singletons.add(str(name))
+
+    def emit_type(self, name):
+        self.types.add(str(name))
 
     def emit(self, s, depth, reflow=True):
         # XXX reflow long lines?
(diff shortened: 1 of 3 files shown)


ericsnowcurrently (approved):
LGTM

Parser/asdl_c.py
```diff
@@ -391,9 +395,8 @@ def sumTrailer(self, name, add_label=False):
     def simpleSum(self, sum, name):
         self.funcHeader(name)
         for t in sum.types:
-            line = ("isinstance = PyObject_IsInstance(obj, "
-                    "(PyObject *)%s_type);")
-            self.emit(line % (t.name,), 1)
+            self.emit("isinstance = PyObject_IsInstance(obj, "
+                      f"astmodulestate_global->{t.name}_type);", 1)
```
tiran: You cannot use f-strings here. Some platforms like old Ubuntu have python3 3.5 or older. Travis CI is failing with a SyntaxError.

tiran (changes requested):

Parser/asdl_c.py
```diff
@@ -940,12 +915,22 @@ def visitModule(self, mod):
...
+        self.emit("m = PyModule_Create(&_astmodule);", 2)
+        self.emit("if (!m) return 0;", 2)
+        self.emit("PyState_AddModule(m, &_astmodule);", 2)
+        self.emit("}", 1)
+        self.emit("astmodulestate *state = astmodulestate_global;", 1)
+        self.emit("if (state->initialized) return 1;", 1)
+        self.emit("if (init_identifiers() < 0) return 0;", 1)
+        self.emit("state->AST_type = PyType_FromSpec(&AST_type_spec);", 1)
+        self.emit("if (!state->AST_type) return 0;", 1)
+        self.emit("((PyTypeObject*)state->AST_type)->tp_dictoffset = offsetof(AST_object, dict);", 1)
```
eduardo-elizondo, line 929: This is still accessing the dict offset, we need to remove this by letting the type system handle it

PR #15970, opened 2019-09-11
bpo-38115: fix invalid lnotab after optimization

Yhg1s:
Fix [bpo-38115](https://bugs.python.org/issue38115), a bug in the peephole optimizer that caused lnotab to have invalid bytecode offsets when the optimizer got rid of the last line(s) of a function. Also fix a bug where the optimizer bails out of optimization but leaves the lnotab modified. This is a very old bug, but before 3.8 the optimizer is more likely to successfully optimize away the last line of a function.

These don't change the *bytecode* that's generated, but they do change the *code objects* (the co_lnotab attribute), so though they theoretically affect Python 3.7 and earlier, we shouldn't backport this beyond 3.8. (Also, if you want to verify this manually, don't forget to touch the .py file or delete the .pyc file... not that I'm speaking of personal experience or anything.)

https://bugs.python.org/issue38115

### DIFF
--- a/Lib/test/test_peepholer.py
+++ b/Lib/test/test_peepholer.py
@@ -40,6 +40,20 @@ class TestTranforms(BytecodeTestCase):
                 self.fail(f'{instr.opname} at {instr.offset} '
                           f'jumps to {tgt.opname} at {tgt.offset}')
 
+    def check_lnotab(self, code):
+        "Check that the lnotab byte offsets are sensible."
+        code = dis._get_code_object(code)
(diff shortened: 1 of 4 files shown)


Yhg1s:
@markshannon might also be interested in this, I guess ;-P

Yhg1s:
Looks like the new assertion is getting triggered by code in test_scope, I'm digging into it.

pablogsal:
Thanks @Yhg1s for helping me and for trying to solve this. I will update #14068 as we talked.

PR #15989, opened 2019-09-11
bpo-38119: Fix shmem resource tracking

applio:
Eliminates the use of the resource tracker on shared memory segments but preserves its use on semaphores, etc. where it is appropriately applied.

Adds a test that fails with the current code (an independent Python process started via subprocess accesses shared memory created by a different Python process, then exits normally, expecting the original creator process to continue to be able to access that shared memory) but passes with this patch to defend against future regressions.

https://bugs.python.org/issue38119

### DIFF
--- a/Lib/multiprocessing/managers.py
+++ b/Lib/multiprocessing/managers.py
@@ -1309,14 +1309,6 @@ if HAS_SHMEM:
         _Server = SharedMemoryServer
 
         def __init__(self, *args, **kwargs):
-            if os.name == "posix":
-                # bpo-36867: Ensure the resource_tracker is running before
-                # launching the manager process, so that concurrent
-                # shared_memory manipulation both in the manager and in the
-                # current process does not create two resource_tracker
-                # processes.
-                from . import resource_tracker
-                resource_tracker.ensure_running()
             BaseManager.__init__(self, *args, **kwargs)
             util.debug(f"{self.__class__.__name__} created by pid {getpid()}")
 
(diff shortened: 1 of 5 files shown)


vinay0410:
[Updated PR](https://github.com/python/cpython/pull/21516) fixing merge conflicts.

col-in-coding (approved):

vinay0410:
@lihaoxiang1989 , this PR has merge conflicts, I have created a PR fixing those, containing the same changes. Can you approve that instead.
[Updated PR](https://github.com/python/cpython/pull/21516) fixing merge conflicts.

lthakur007:
Hello Team,
As am facing the same issue for which this PR has a solution.

May I know, once the PR is merged, how can I get the fix on Ubuntu 18.04?

vinay0410:
@lthakur007 , these changes are now being tracked by https://github.com/python/cpython/pull/21516, once those changes are merged you can either wait for an alpha or beta tag release, or you can directly build python from source (master).

lthakur007:
> @lthakur007 , these changes are now being tracked by #21516, once those changes are merged you can either wait for an alpha or beta tag release, or you can directly build python from source (master).

Thank you for the response.

zhuowater (commented):
good idea,it works,nice job!

atarashansky:
Astonishing that this is still not merged... I've been manually unregistering shared memory from child processes for the last year. What's the point of shared memory if it doesn't persist between processes by default...

buzmeg:
@jcarlitooliveira Does #82300 help you in any way?

gpshead:
there's an updated PR mentioned above: https://github.com/python/cpython/pull/21516 - it has a lot more discussion on it so i'm closing this earlier version of the proposed change.

PR #15993, opened 2019-09-11
bpo-38121: Sync importlib.metadata with 0.22 backport

jaraco:
https://bugs.python.org/issue38121


Lib/importlib/_bootstrap_external.py
```diff
@@ -1379,54 +1379,8 @@ def find_distributions(self, context=None):
...
-    @classmethod
-    def _search_path(cls, root, pattern):
-        if not root.is_dir():
-            return ()
-        normalized = pattern.replace('-', '_')
-        return (item for item in root.iterdir()
-                if cls._matches_info(normalized, item)
-                or cls._matches_legacy(normalized, item))
+        from importlib.metadata import MetadataPathFinder
+        MetadataPathFinder.find_distributions(*args, **kwargs)
```
pganssle: This needs to `return` the result of this call, I've pushed a commit fixing it.

Lib/importlib/metadata.py
```diff
@@ -363,6 +365,58 @@ def find_distributions(self, context=Context()):
...
+    def _search_paths(cls, pattern, paths):
+        """Find metadata directories in paths heuristically."""
+        return itertools.chain.from_iterable(
+            cls._search_path(path, pattern)
+            for path in map(cls._switch_path, paths)
+            )
+
+    @staticmethod
+    def _switch_path(path):
+        PYPY_OPEN_BUG = False
```
pganssle, line 392: What is this?

jaraco: This disables a workaround present in the importlib_metadata code for a bug in PyPy. I’ve coded it this way to avoid interacting with the code line itself to aid in merges.

pganssle: Ah, that makes sense. Was just checking to see if this got checked in by accident.

pganssle (approved):
LGTM! Will merge per our conversation at the sprint.

PR #16004, opened 2019-09-11
bpo-37879: Fix warnings in _testcapimodule

encukou:
Commit ff023ed36ea260ab64be5895f1f1f087c798987a introduced some warnings that got through my review. Here is a fix.

https://bugs.python.org/issue37879

### DIFF
--- a/Modules/_testcapimodule.c
+++ b/Modules/_testcapimodule.c
@@ -6304,7 +6304,7 @@ static void
 heapctypesubclasswithfinalizer_finalize(PyObject *self)
 {
     PyObject *error_type, *error_value, *error_traceback, *m;
-    PyObject *oldtype=NULL, *newtype=NULL;
+    PyObject *oldtype = NULL, *newtype = NULL;
 
     /* Save the current exception, if any. */
     PyErr_Fetch(&error_type, &error_value, &error_traceback);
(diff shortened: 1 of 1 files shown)


Modules/_testcapimodule.c
```diff
@@ -6303,7 +6303,8 @@ heapctypesubclasswithfinalizer_init(PyObject *self, PyObject *args, PyObject *kw
 static void
 heapctypesubclasswithfinalizer_finalize(PyObject *self)
 {
-    PyObject *error_type, *error_value, *error_traceback, *m, *oldtype, *newtype;
+    PyObject *error_type, *error_value, *error_traceback, *m;
+    PyObject *oldtype=NULL, *newtype=NULL;
```
corona10: nit
```C
PyObject *oldtype = NULL, *newtype = NULL;
```

Always put spaces around assignment, Boolean and comparison operators.
from PEP7 :)

corona10 (changes requested):

Modules/_testcapimodule.c
```diff
@@ -6233,7 +6233,7 @@ static PyType_Slot HeapCType_slots[] = {
     {Py_tp_init, heapctype_init},
     {Py_tp_members, heapctype_members},
     {Py_tp_dealloc, heapctype_dealloc},
-    {Py_tp_doc, heapctype__doc__},
+    {Py_tp_doc, (char*)heapctype__doc__},
```
corona10, line 6236: https://github.com/python/cpython/blob/2ec70102066fe5534f1a62e8f496d2005e1697db/Modules/_struct.c#L2048

There are some cases that handling type casting with Py_tp_doc
They are using void*. How about casting to void*?

encukou: Either will work.
Ideally we'd want all the slots to be `const`, but some of them are PyObjects with mutable refcounts :(

corona10 (approved):

PR #16005, opened 2019-09-11
bpo-36871: Handle spec errors in assert_has_calls

sfreilich:
The fix in PR #13261 handled the underlying issue about the spec for specific methods not being applied correctly, but it didn't fix the issue that was causing the misleading error message.

The code currently grabs a list of responses from _call_matcher (which may include exceptions). But it doesn't reach inside the list when checking if the result is an exception. This results in a misleading error message when one of the provided calls does not match the spec.

https://bugs.python.org/issue36871

Automerge-Triggered-By: @gpshead

### DIFF
--- a/Lib/unittest/mock.py
+++ b/Lib/unittest/mock.py
@@ -936,12 +936,12 @@ class NonCallableMock(Base):
         if not any_order:
             if expected not in all_calls:
                 if cause is None:
-                  problem = 'Calls not found.'
+                    problem = 'Calls not found.'
                 else:
-                  problem = ('Error processing expected calls.\n'
-                             'Errors: {}').format(
-                                 [e if isinstance(e, Exception) else None
-                                  for e in expected])
+                    problem = ('Error processing expected calls.\n'
+                               'Errors: {}').format(
+                                   [e if isinstance(e, Exception) else None
+                                    for e in expected])
                 raise AssertionError(
                     f'{problem}\n'
                     f'Expected: {_CallList(calls)}\n'
@@ -2227,12 +2227,12 @@ class AsyncMockMixin(Base):
         if not any_order:
             if expected not in all_awaits:
                 if cause is None:
-                  problem = 'Awaits not found.'
+                    problem = 'Awaits not found.'
                 else:
-                  problem = ('Error processing expected awaits.\n'
-                             'Errors: {}').format(
-                                 [e if isinstance(e, Exception) else None
-                                  for e in expected])
+                    problem = ('Error processing expected awaits.\n'
+                               'Errors: {}').format(
+                                   [e if isinstance(e, Exception) else None
+                                    for e in expected])
                 raise AssertionError(
                     f'{problem}\n'
                     f'Expected: {_CallList(calls)}\n'
--- a/Lib/unittest/mock.py
+++ b/Lib/unittest/mock.py
@@ -935,9 +935,17 @@ class NonCallableMock(Base):
         all_calls = _CallList(self._call_matcher(c) for c in self.mock_calls)
         if not any_order:
             if expected not in all_calls:
+                if cause is None:
+                  problem = 'Calls not found.'
+                else:
+                  problem = ('Error processing expected calls.\n'
+                             'Errors: {}').format(
+                                 [e if isinstance(e, Exception) else None
+                                  for e in expected])
                 raise AssertionError(
-                    'Calls not found.\nExpected: %r%s'
-                    % (_CallList(calls), self._calls_repr(prefix="Actual"))
+                    f'{problem}\n'
+                    f'Expected: {_CallList(calls)}\n'
+                    f'Actual: {self._calls_repr(prefix="Actual")}'
                 ) from cause
             return
 
@@ -2218,8 +2226,16 @@ class AsyncMockMixin(Base):
         all_awaits = _CallList(self._call_matcher(c) for c in self.await_args_list)
         if not any_order:
             if expected not in all_awaits:
+                if cause is None:
+                  problem = 'Awaits not found.'
+                else:
+                  problem = ('Error processing expected awaits.\n'
+                             'Errors: {}').format(
+                                 [e if isinstance(e, Exception) else None
+                                  for e in expected])
                 raise AssertionError(
-                    f'Awaits not found.\nExpected: {_CallList(calls)}\n'
+                    f'{problem}\n'
+                    f'Expected: {_CallList(calls)}\n'
                     f'Actual: {self.await_args_list}'
                 ) from cause
             return
--- a/Lib/unittest/mock.py
+++ b/Lib/unittest/mock.py
@@ -2214,7 +2214,7 @@ class AsyncMockMixin(Base):
         they must all appear in :attr:`await_args_list`.
         """
         expected = [self._call_matcher(c) for c in calls]
-        cause = expected if isinstance(expected, Exception) else None
+        cause = next((e for e in expected if isinstance(e, Exception)), None)
         all_awaits = _CallList(self._call_matcher(c) for c in self.await_args_list)
         if not any_order:
             if expected not in all_awaits:
--- a/Lib/unittest/mock.py
+++ b/Lib/unittest/mock.py
@@ -931,7 +931,7 @@ class NonCallableMock(Base):
         If `any_order` is True then the calls can be in any order, but
         they must all appear in `mock_calls`."""
         expected = [self._call_matcher(c) for c in calls]
-        cause = expected if isinstance(expected, Exception) else None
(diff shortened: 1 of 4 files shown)


sfreilich:
Ah, wait, I made my fix to the wrong line!

Also, assert_has_awaits has the exact same bug.

sfreilich:
This error handling bug isn't a huge problem _most_ of the time. But when there's a bug in the creation or application of the mock specs (like the bugs fixed in earlier PRs for this issue), the resulting error message is _really_ confusing, since it incorrectly says that the expected calls were not present, even when the expected and actual calls are in fact identical.

Lib/unittest/mock.py
```diff
@@ -931,7 +931,7 @@ def assert_has_calls(self, calls, any_order=False):
         If `any_order` is True then the calls can be in any order, but
         they must all appear in `mock_calls`."""
         expected = [self._call_matcher(c) for c in calls]
-        cause = expected if isinstance(expected, Exception) else None
+        cause = next((e for e in expected if isinstance(e, Exception)), None)
```
tirkarthi, line 934: Good change. Thanks. One suggestion is that this will only show the first exception on a series of exceptions. Thoughts on below?

```python
def f(): pass

mock = Mock(spec=f)
mock.assert_has_calls([call(a='wrong'), call(1)])
```

```
TypeError: got an unexpected keyword argument 'a'

  File "/tmp/foo.py", line 15, in <module>
    mock.assert_has_calls([call(a='wrong'), call(1)])
  File "/Users/karthikeyansingaravelan/stuff/python/cpython/Lib/unittest/mock.py", line 939, in assert_has_calls
    raise AssertionError(
Expected: [call(a='wrong'), call(1)]
```

Something like `cause = [e for e in expected if isinstance(e, Exception)] or None` would show multiple exceptions but this would mean it should be an `Exception` and we can't get the type. 

```
Exception: [TypeError("got an unexpected keyword argument 'a'"), TypeError('too many positional arguments')]

  File "/tmp/foo.py", line 15, in <module>
    mock.assert_has_calls([call(a='wrong'), call(1)])
  File "/Users/karthikeyansingaravelan/stuff/python/cpython/Lib/unittest/mock.py", line 939, in assert_has_calls
    raise AssertionError(
Expected: [call(a='wrong'), call(1)]
```

lisroach: @1st1 here's another example of where MultiError could come in handy. 

I do like the showing of the multiple arguments, it's not something we often see in Python but in the case of testing the more exceptions you can find the better it seems to me.

sfreilich: I feel like I'm missing some context re MultiError, what is that?

I'm somewhat surprised putting a list of exceptions in the from clause works, the [documentation](https://docs.python.org/3.9/reference/simple_stmts.html#the-raise-statement) implies that's not correct:

> The from clause is used for exception chaining: if given, the second expression must be another exception class or instance

Could always put more information in the text of the message, though? (Instead of or in addition to setting the cause to the first exception encountered?)

tirkarthi: I hope there will be a PEP soon for the idea. I guess this trio issue helps and the closest reference I can find with google : https://github.com/python-trio/trio/issues/611

sfreilich: I added a change to improve the error messages to include all the exceptions encountered while processing the expected calls, while still setting the cause of the AssertionError to the first exception.

sfreilich:
Is this ready to merge?

gpshead:
I think this is ready to merge, but can you use blurb to add a news entry about the fixed error message? It's worth calling out.

gpshead (approved):

gpshead (approved):

Lib/unittest/mock.py
```diff
@@ -931,13 +931,21 @@ def assert_has_calls(self, calls, any_order=False):
...
+                    problem = ('Error processing expected calls.\n'
+                               'Errors: {}').format(
+                                   [e if isinstance(e, Exception) else None
+                                    for e in expected])
                 raise AssertionError(
-                    'Calls not found.\nExpected: %r%s'
-                    % (_CallList(calls), self._calls_repr(prefix="Actual"))
+                    f'{problem}\n'
+                    f'Expected: {_CallList(calls)}\n'
+                    f'Actual: {self._calls_repr(prefix="Actual")}'
```
tirkarthi, line 948: This causes a difference in error message with Actual line printed twice. Using `self._calls_repr` alone takes care of it. `self._calls_repr` already prepends a newline so please remove the "\n" in `f'Expected: {_CallList(calls)}\n'` to avoid an extra new line.

```python
from unittest.mock import *

def f(): pass

m = Mock(spec=f)
m()
m.assert_has_calls([call(), call('wrong')])
```

```
➜  cpython git:(master) ✗ ./python.exe /tmp/foo.py
TypeError: too many positional arguments

  File "/tmp/foo.py", line 7, in <module>
    m.assert_has_calls([call(), call('wrong')])
  File "/Users/karthikeyansingaravelan/stuff/python/cpython/Lib/unittest/mock.py", line 945, in assert_has_calls
    raise AssertionError(
AssertionError: Error processing expected calls.
Errors: [None, TypeError('too many positional arguments')]
Expected: [call(), call('wrong')]
Actual:
Actual: [call()].
```

```suggestion
                    f'{self._calls_repr(prefix="Actual")}'
```

tirkarthi:
Sorry for the last minute change. Just spotted it. Would request removing automerge.

sfreilich:
I'll create another PR for that, sorry.

Lib/unittest/mock.py
gpshead: whooops. thanks!  i missed that. :)  follow up with a new PR with that little change.  looks like we need to do the backport manually and can lump these two changes together for that anyways.

tirkarthi:
No problem. Please also add a test too with the PR as below which would fail without the error message change suggested. Normally we don't added very specific asserts for error message but I think it helps in this case. I think `assert_has_awaits` is not affected since it doesn't use `self._calls_repr`.

```python
with self.assertRaisesRegex(
        AssertionError,
        re.escape("Error processing expected calls.\n"
                  "Errors: [None, TypeError('too many positional "
                  "arguments')]\n"
                  "Expected: [call(), call('wrong')]")) as cm:
    mock.assert_has_calls([call(), call('wrong')])

with self.assertRaisesRegex(
        AssertionError,
        re.escape("Error processing expected calls.\n"
                  "Errors: [None, TypeError('too many positional "
                  "arguments')]\n"
                  "Expected: [call(), call('wrong')]\n"
                  "Actual: [call()]")) as cm:
    mock()
    mock.assert_has_calls([call(), call('wrong')])

```

Lib/unittest/mock.py
sfreilich: Good catch, sorry. I was making this more similar to cleaner logic in assert_has_awaits, and I missed that repetition.

sfreilich:
PR #16361

gpshead:
so much for automation.  i'll take care of the backports of both of these PRs.

tirkarthi:
I am not sure of the backport to 3.7 . Though this fixes the bug in detecting Exception it also enhances and changes error message which some package might be asserting in their test. IMO we can backport only exception detecting code if needed to 3.7 instead of error message enhancement.

sfreilich:
> might be asserting in their test

The specific error message produced by unittest.mock.Mock.assert_has_calls seems like a strange thing to be asserting about in tests. Unless the test is testing a subclass of mock.Mock or something. Basically, I see your concern, but I think the tradeoffs are less towards interface stability when the error messages in question are generated by test assertions.

That said, the more important part is the one that prevents the exceptions from spec.bind from being swallowed entirely and ensures at lest the first one ends up reported to the user somehow (e.g. by setting it as the cause of the AssertionError).

tirkarthi:
There are packages that assert exact error messages that are mostly built on top of mock module. Sample issue over 3.8 error message alignment change : https://github.com/pytest-dev/pytest-mock/issues/139

> That said, the more important part is the one that prevents the exceptions from spec.bind from being swallowed entirely and ensures at lest the first one ends up reported to the user somehow (e.g. by setting it as the cause of the AssertionError).

Okay, I see it as a partly improvement in error message with the bug fix to detect the exception object. I will leave it to the core dev for the backport to 3.7 decision.

gpshead:
PR #16374 has a backport to 3.7.

I see the incorrect error message as a bug so I'm leaning towards merging it.

That pytest has some of its own test suite depend on specifics of error message isn't _super_ surprising given the nature of pytest, but it is within its own test suite and they seem to understand how to fix fragile golden value data issues in those tests if/when they do come up.

I wouldn't expect any normal actual user code to reasonably depend on specific formatting of error messages from mock assertions.

PR #16012, opened 2019-09-12
bpo-38112: Compileall improvements

frenzymadness:
Improvements of compileall module (released as compileall2 on [PyPI](https://pypi.org/project/compileall2/)|[GitHub](https://github.com/fedora-python/compileall2)) which makes the module better for Linux distributions where some manipulation with a baked path into a byte-compiled file is necessary. This patchset contains tests but needs to add documentation.

https://bugs.python.org/issue38112

### DIFF
--- a/Doc/library/compileall.rst
+++ b/Doc/library/compileall.rst
@@ -57,7 +57,7 @@ compile Python sources.
 
    Remove (``-s``) or append (``-p``) the given prefix of paths
    recorded in the ``.pyc`` files.
-   Raises :exc:`ValueError` if combined with ``-d``.
+   Cannot be combined with ``-d``.
 
 .. cmdoption:: -x regex
 
@@ -126,7 +126,7 @@ compile Python sources.
 
 .. versionchanged:: 3.9
    Added the ``-s``, ``-p``, ``-e`` options.
-   Raised the default default recursion limit from 10 to
+   Raised the default recursion limit from 10 to
    :py:func:`sys.getrecursionlimit()`.
    Added the possibility to specify the ``-o`` option multiple times.
 
--- a/Doc/library/compileall.rst
+++ b/Doc/library/compileall.rst
@@ -52,6 +52,13 @@ compile Python sources.
    cases where the source file does not exist at the time the byte-code file is
    executed.
 
+.. cmdoption:: -s strip_prefix
+.. cmdoption:: -p prepend_prefix
+
+   Remove (``-s``) or append (``-p``) the given prefix of paths
+   recorded in the ``.pyc`` files.
+   Raises :exc:`ValueError` if combined with ``-d``.
+
 .. cmdoption:: -x regex
 
    regex is used to search the full path to each file considered for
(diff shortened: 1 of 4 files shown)


the-knights-who-say-ni:

## CLA Missing

@frenzymadness

frenzymadness:
CLA signed.

Doc/library/compileall.rst
```diff
@@ -107,6 +124,12 @@ compile Python sources.
 .. versionchanged:: 3.7
    Added the ``--invalidation-mode`` option.
 
+.. versionchanged:: 3.9
+   Added the ``-s``, ``-p``, ``-e`` options.
+   Raised the default default recursion limit from 10 to
```
frenzymadness: Word "default" two times

encukou:
I'm still not sure if "`-d` takes precedence" means that the `stripdir`/`prependdir` options are somehow applied after `ddir`, or if they should be ignored entirely if `ddir` is given.

What's the expected behavior of: ddir and prependdir, ddir and stripdir?
Does it make sense to allow these combinations?

frenzymadness:
If `ddir` is specified, `stripdir` and `prependdir` are ignored. It makes sense to use `ddir` or a combination of `stripdir` and `prependdir`, for example:
* If you want to modify a path, for example, remove /buildroot from it and add a leading /, you can do it in with a combination of `stripdir` and `prependdir` and the rest of a path after /buildroot will stay there untouched.
* But if you know that you want to replace the whole path with something else, is much easier to use `ddir` than a combination of `stripdir` and `prependdir` because it can simply replace a whole path from beginning to the name of a file with specified path.

encukou:
But now, if you use `ddir` and `prependdir`, they're both applied.

Since combining `-s`/`-p` with `ddir` is not useful, would it make sense to reject those combinations?

frenzymadness:
> But now, if you use `ddir` and `prependdir`, they're both applied.

I don't think so. At leat, I did some tests and from results it seems to me that if you use ddir, prependdir is ignored. For example:

```
$ echo "1/0" > comp_test/test.py
$ ../python -m compileall -f -d "/foo/bar/baz" -p "/prependthis" comp_test
$ ../python comp_test/__pycache__/test.cpython-39.pyc
  File "/foo/bar/baz/test.py", line 1, in <module>
ZeroDivisionError: division by zero
```

This is btw covered by the latest test where it checks that the content of prependdir is not in the error output. Am I missing something?

> Since combining `-s`/`-p` with `ddir` is not useful, would it make sense to reject those combinations?

Yes, I can imagine something like `ValueError("ddir cannot be used in a combination with stripdir and prependdir")`. This exception can be altered for the functions where it'd report argument names and for CLI usage where it'd report command line arguments.

frenzymadness:
The latest commit adds ValueError in case `ddir` and `stripdir`/`prependdir` are used together. It's only in `compile_file` function which makes it simple and DRY but it also means that this error is not raised when the directory to compile is empty or does not exists.

If you are okay with this form, I'll rework conditions related to `ddir` precedence and also documentation.

encukou:
Looks good

frenzymadness:
Rebased, squashed, added news entry, all green. From my point of view, unless I forgot to change something, it's ready.

Doc/library/compileall.rst
encukou:
```suggestion
   Raised the default recursion limit from 10 to
```

encukou (approved):
Looks good now, except the CLI interface should raise CLI error instead of showing a traceback. I added a commit for that, let's merge if it's green.

Workflow nitpick: we squash all changes when we merge, so rebasing is unnecessary. After you rebase, the commits need to be reviewed again, in case something changed.

frenzymadness:
Looks good. Thank you!

encukou:
I'm looking into it.

PR #16013, opened 2019-09-12
Cut disused recode_encoding logic in _PyBytes_DecodeEscape.

gnprice:
All call sites pass NULL for `recode_encoding`, so this path is
completely untested.  That's been true since before Python 3.0.
It adds significant complexity to this logic, so it's best to
take it out.

All call sites now have a literal NULL, and that's been true since
commit 768921cf3 eliminated a conditional (`foo ? bar : NULL`) at
the call site in Python/ast.c where we're parsing a bytes literal.
But even before then, that condition `foo` had been a constant
since unadorned string literals started meaning Unicode, in commit
572dbf8f1 aka v3.0a1~1035 .

The `unicode` parameter is already unused, so mark it as unused too.
The code that acted on it was also taken out before Python 3.0, in
commit 8d30cc014 aka v3.0a1~1031 .

The function (PyBytes_DecodeEscape) is exposed in the API, but it's
never been documented.

### DIFF
--- a/Include/bytesobject.h
+++ b/Include/bytesobject.h
@@ -77,9 +77,7 @@ PyAPI_FUNC(PyObject *) PyBytes_DecodeEscape(const char *, Py_ssize_t,
 #ifndef Py_LIMITED_API
 /* Helper for PyBytes_DecodeEscape that detects invalid escape chars. */
 PyAPI_FUNC(PyObject *) _PyBytes_DecodeEscape(const char *, Py_ssize_t,
-                                             const char *, Py_ssize_t,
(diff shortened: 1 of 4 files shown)


gvanrossum:
> Is there a reason not to eliminate the unicode and recode_encoding parameters from _PyBytes_DecodeEscape completely?

Well, it was also called from typed_ast/ast3/Python/ast.c (which is an external backport of ast.c and uses some internals like this one), but we're fixing that now, so no need to roll this back or anything. See https://github.com/python/typed_ast/pull/128

PR #16016, opened 2019-09-12
bpo-38130: Fix error in explaining when an exception is re-raised

prodehghan:
https://bugs.python.org/issue38130

### DIFF
--- a/Doc/tutorial/errors.rst
+++ b/Doc/tutorial/errors.rst
@@ -390,7 +390,7 @@ example::
 
 If a :keyword:`finally` clause is present, the :keyword:`finally` clause will execute as the last task before the :keyword:`try` statement completes. The :keyword:`finally` clause runs whether or not the :keyword:`try` statement produces an exception. The following points discuss more complex cases when an exception occurs:
 
-* If an exception occurs during execution of the :keyword:`!try` clause, the exception may be handled by an :keyword:`except` clause. In all cases, the exception is re-raised after the :keyword:`!finally` clause has been executed.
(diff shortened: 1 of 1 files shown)


Doc/tutorial/errors.rst
```diff
@@ -390,7 +390,7 @@ example::
 
 If a :keyword:`finally` clause is present, the :keyword:`finally` clause will execute as the last task before the :keyword:`try` statement completes. The :keyword:`finally` clause runs whether or not the :keyword:`try` statement produces an exception. The following points discuss more complex cases when an exception occurs:
 
-* If an exception occurs during execution of the :keyword:`!try` clause, the exception may be handled by an :keyword:`except` clause. In all cases, the exception is re-raised after the :keyword:`!finally` clause has been executed.
+* If an exception occurs during execution of the :keyword:`!try` clause, the exception may be handled by an :keyword:`except` clause. If the exception is not handled by any of the :keyword:`except` clauses, the exception is re-raised after the :keyword:`!finally` clause has been executed.
```
epicfaace: Because we won't necessarily have multiple "except clauses".

```suggestion
* If an exception occurs during execution of the :keyword:`!try` clause, the exception may be handled by an :keyword:`except` clause. If the exception is not handled by an :keyword:`except` clause, the exception is re-raised after the :keyword:`!finally` clause has been executed.
```

matrixise (approved):

matrixise:
Thank you for your PR, I am going to merge into master, 3.8 and 3.7

PR #16023, opened 2019-09-12
bpo-38132: Simplify _hashopenssl code

tiran:

https://bugs.python.org/issue38132


Misc/NEWS.d/next/Library/2019-09-12-12-47-35.bpo-38132.KSFx1F.rst
```diff
@@ -0,0 +1,2 @@
+The OpenSSL hashlib wrapper uses a simpler implementation. Several Macros
+and pointless caches are gone. The hash name now comes from OpenSSL's EVP.
```
gpshead, line 2: From a news entry reader perspective they'll want to know what this means in terms of API changes.  Do the strings returned by `.name` on any existing standard supported hashes change from what our hardcoded values were before?  If not, state not, if so, list what changes.

gpshead (changes requested):
less code is better code! :)

gpshead (approved):

PR #16025, opened 2019-09-12
bpo-38133: Allow py.exe launcher to locate installations from the Microsoft Store

zooba:
Also improve display of active virtual environments with `--list` options.

https://bugs.python.org/issue38133


matrixise (approved):
Reading the source code, I don't find any problem, I am not a Win32 expert but the algos seem be okay.

matrixise:
I should try to find a VM of Windows and try to help you for the #windows bugs.

PR #16042, opened 2019-09-12
bpo-38135: Fix depth-first-search computation in compile.c

markshannon:
With the current compiler it is impossible to expose the bug in the block ordering computation. 
Consequently, there is no observable change in behavior and thus no news item is necessary.

https://bugs.python.org/issue38135


gpshead (commented):
At a glance I believe this is correct.  Looping Thomas in as a reviewer as I believe he's been working in the compile code recently.

PR #16053, opened 2019-09-12
bpo-38138: Fix memory leak introduced by interned strings

DinoV:
Interned string needs to be decref'd

https://bugs.python.org/issue38138

Automerge-Triggered-By: @matrixise


matrixise (approved):
Checked on my laptop, I have no leak with your patch
```
./python -m test test_importlib -R 3:3
```

Thank you

PR #16060, opened 2019-09-12
bpo-38108: Makes mock objects inherit from Base

lisroach:
Updates the few Mock classes that do not inherit from Base to properly inherit from base. Updates the super() calls that should go with them as well.

That was all I planned on doing with this diff, but the changes to get AsyncMock inheritance working right with subclasses was so linked with this diff I merged it into one to prevent temporary breaking changes. 

https://bugs.python.org/issue38108


asvetlov (approved):
LGTM

lisroach:
Oops, didn't mean to do that

PR #16070, opened 2019-09-12
bpo-38013: make async_generator_athrow object tolerant to throwing exceptions

asvetlov:
Even when the helper is not started yet.

This behavior follows conventional generator one.
There is no reason for `async_generator_athrow` to handle `gen.throw()` differently.

https://bugs.python.org/issue38013

### DIFF
--- a/Lib/test/test_asyncgen.py
+++ b/Lib/test/test_asyncgen.py
@@ -1125,6 +1125,27 @@ class AsyncGenAsyncioTest(unittest.TestCase):
         res = self.loop.run_until_complete(run())
         self.assertEqual(res, [i * 2 for i in range(1, 10)])
 
+    def test_asyncgen_nonstarted_hooks_are_cancellable(self):
+        messages = []
+
+        def exception_handler(loop, context):
+            messages.append(context)
+
+        async def async_iterate():
+            yield 1
+            yield 2
+
+        async def main():
+            loop = asyncio.get_running_loop()
+            loop.set_exception_handler(exception_handler)
+
+            async for i in async_iterate():
+                break
+
+        asyncio.run(main())
+
+        self.assertEqual([], messages)
+
 
 if __name__ == "__main__":
     unittest.main()
--- a/Lib/test/test_asyncgen.py
+++ b/Lib/test/test_asyncgen.py
@@ -1126,6 +1126,7 @@ class AsyncGenAsyncioTest(unittest.TestCase):
         self.assertEqual(res, [i * 2 for i in range(1, 10)])
 
     def test_asyncgen_nonstarted_hooks_are_cancellable(self):
+        # See https://bugs.python.org/issue38013
         messages = []
 
         def exception_handler(loop, context):
--- a/Misc/NEWS.d/next/Core and Builtins/2019-09-12-19-50-01.bpo-38013.I7btD0.rst
+++ b/Misc/NEWS.d/next/Core and Builtins/2019-09-12-19-50-01.bpo-38013.I7btD0.rst
@@ -1,0 +1,3 @@
+Allow to call ``async_generator_athrow().throw(...)`` even for non-started
+async generator helper. It fixes annoying warning at the end of
+:func:`asyncio.run` call.
--- a/Objects/genobject.c
+++ b/Objects/genobject.c
@@ -1884,11 +1884,6 @@ async_gen_athrow_throw(PyAsyncGenAThrow *o, PyObject *args)
 {
     PyObject *retval;
 
-    if (o->agt_state == AWAITABLE_STATE_INIT) {
-        PyErr_SetString(PyExc_RuntimeError, NON_INIT_CORO_MSG);
-        return NULL;
-    }
-
     if (o->agt_state == AWAITABLE_STATE_CLOSED) {
         PyErr_SetNone(PyExc_StopIteration);
         return NULL;


1st1 (changes requested):
I don't think this is the right way, although it might be. I'll need some time to carefully review the implementation. Please don't merge this.

Lib/test/test_asyncgen.py
```diff
@@ -1125,6 +1125,27 @@ def make_arange(n):
         res = self.loop.run_until_complete(run())
         self.assertEqual(res, [i * 2 for i in range(1, 10)])
 
+    def test_asyncgen_nonstarted_hooks_are_cancellable(self):
+        messages = []
```
aeros, line 1130: Would it be helpful to add a brief comment mentioning the purpose of this test? I don't think the purpose is immediately obvious from the name of the test alone (without the context of the bpo issue and PR comments).

asvetlov: I've added a link to the issue, should be enough for digging into the context

Lib/test/test_asyncgen.py
```diff
@@ -1125,6 +1125,28 @@ def make_arange(n):
...
+        async def main():
+            loop = asyncio.get_running_loop()
+            loop.set_exception_handler(exception_handler)
+
+            async for i in async_iterate():
+                break
+
+        asyncio.run(main())
+
+        self.assertEqual([], messages)
```
aeros, line 1148: Isn't it preferable to use `assertFalse` when checking if a container is empty? Unless I'm mistaken, this has better performance than comparing an empty container to an existing one (due to not having to initialize an empty container).
```suggestion
        self.assertTrue(len(messages), 0)
```

Edit: Changed suggestion based on comments below.

asvetlov: No. The initial `messages` value is precisely the empty list `[]`.
The assertion is correct. It doesn't check for `None` for example.

aeros: Oh okay, thanks for the clarification @asvetlov. In the past, I've seen other developers using a customized ``assertEmpty`` for containers, where the logic is one of the following:

```
def assertEmpty(self, obj):
    self.assertFalse(obj)
```

```
def assertEmpty(self, obj):
    self.assertFalse(len(obj))
```

```
def assertEmpty(self, obj):
    self.assertEqual(len(obj), 0)
```

(Basically the same as the above, since False is a const for 0)

Since we don't have an ``assertEmpty`` method directly included in the unit test framework, I had assumed that within CPython, we had used one of the above (without creating a dedicated method).

Is it instead preferable to use an ``assertEquals`` on an empty container and the container being tested as empty? Example: if we were testing if ``test_set`` was empty, we would use ``self.assertEquals(set(), test_set)``.

Apologies if any of this is fairly obvious. I've recently been trying to learn our standard testing practices better, so that I can help with improving the tests in the future.

ammaraskar: Personally I find `self.assertEqual([], empty_list)` to be far more readable than relying on the implicit boolean coercion of `self.assertFalse(empty_list)`, and adding an `assertEmpty` when it doesn't already exist in `unittest` seems like extra complexity for not much extra utility.

aeros: @ammaraskar To clarify, I was not arguing for the addition of an ``assertEmpty`` for the ``unitttest`` module, I was just explaining where I had the idea from. I don't think it would be worth adding to the public API for ``unittest``. 

Out of curiosity, do you find ``self.assertEqual([], empty_list)`` to be more readable than ``self.assertTrue(len(empty_list), 0)``? I can understand that the implicit boolean conversion might not be overly explicit in stating that the container is empty.

The main disadvantages with using ``self.assertEqual([], empty_list)`` I can think of are the additional resources required to initialize the empty list, and having to perform the equality comparison between the two objects. From my understanding, accessing the length property and comparing to an int with ``self.assertTrue(len(empty_list), 0)`` would be faster. A single change between the two would not make a substantial difference, but I could see this causing a difference in the duration over time for the tests if it's done repeatedly.

My main question to @asvetlov was asking what the standard practice was for the CPython tests regarding empty container checks. I am also curious if the performance is considered to be substantial concern enough concern for this to be worth using. 

My guess would be that performance in the tests is a concern, but not the highest priority. If it is a concern, I have no idea as to the extent. In my experience, the order of priority within tests is typically something along the lines of:

coverage > consistency > readability >= performance

But that can vary significantly between different projects, which is why I'm asking.

asvetlov: @aeros167 think about what error you have if the check fails.
`self.assertEqual(len(obj), 0)` is the worst in my case: `1 != 0` doesn't provide an idea what messages are present in the list which is expected to be empty.
A comparison with the empty list generates a clear error text on failure.

asvetlov: Performance matters. Say, 6 years ago I worked hard to reduce the time of asyncio tests execution in the factor of two. For tests, it usually means that a test sleeps less and does as small work as possible for demonstration of the idea. Any check method has equal time from this perspective regardless of the check's algorithmic complexity (O(1) vs O(N) for `bool(lst)` vs `[] == lst`). The test algorithmic complexity may be important sometimes.

aeros: @asvetlov Thanks for the detailed response, I certainly appreciate it.

> think about what error you have if the check fails.
self.assertEqual(len(obj), 0) is the worst in my case: 1 != 0 doesn't provide an idea what messages are present in the list which is expected to be empty.

Ah, this is a good point that I had not considered. Although the optional message argument could be added "list/container should be empty" to elaborate, arguably ``self.assertEqual([], empty_list)`` is already self-explanatory. 

> Performance matters. Say, 6 years ago I worked hard to reduce the time of asyncio tests execution in the factor of two. 

On behalf of everyone who has ran through the test suite manually, I thank you for this. ``test_asyncio`` is currently one of longer duration tests on there, double the current duration is pretty hard to imagine. It's good to know that performance is a concern.

> Any check method has equal time from this perspective regardless of the check's algorithmic complexity (O(1) vs O(N) for bool(lst) vs [] == lst). The test algorithmic complexity may be important sometimes.

Does this occur in the case of a check that is only performed once, a small number of times, or when it takes up a small portion of the execution time, N is effectively considered negligible enough to be considered constant time (O(1)), even in the case of O(N)? I can understand that it might considered negligible in comparison to the other components of the ``test_asyncio``. Perhaps this would be considered more significant of a difference if this check was being performed repeatedly, or in a context where there weren't external delays.

+1 if I'm understanding this correctly, otherwise feel free to correct me.

asvetlov: You are correct

1st1 (approved):

PR #16071, opened 2019-09-12
bpo-38142: Updated _hashopenssl.c to be PEP 384 compliant

tiran:
The _hashlib OpenSSL wrapper extension module is now PEP 384 compliant.

Also remove refleak test from test_hashlib. The updated type no longer accepts random arguments to `__init__`.

https://bugs.python.org/issue38142

### DIFF
--- a/Modules/_hashopenssl.c
+++ b/Modules/_hashopenssl.c
@@ -34,7 +34,12 @@
 
 #define MUNCH_SIZE INT_MAX
 
+<<<<<<< HEAD
 #ifdef NID_sha3_224
+=======
+<<<<<<< HEAD
+#if defined(NID_sha3_224) && defined(EVP_MD_FLAG_XOF)
+>>>>>>> bpo-38142: Updated _hashopenssl.c to be PEP 384 compliant
 #define PY_OPENSSL_HAS_SHA3 1
 #endif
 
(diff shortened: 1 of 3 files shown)


Modules/_hashopenssl.c
```diff
@@ -364,45 +374,25 @@ PyDoc_STRVAR(hashtype_doc,
...
-    0,                  /* tp_base */
-    0,                  /* tp_dict */
-    0,                  /* tp_descr_get */
-    0,                  /* tp_descr_set */
-    0,                  /* tp_dictoffset */
+    Py_TPFLAGS_DEFAULT | Py_TPFLAGS_BASETYPE,
+    EVPtype_slots
 };
 
+
```
eduardo-elizondo: Remove extra ws? :)

eduardo-elizondo (changes requested):
Since EVP is now a heap allocated type, you should now properly refcount the type as well. You can read the details in: https://docs.python.org/3.8/whatsnew/3.8.html#changes-in-the-c-api, search for [bpo-35810](https://bugs.python.org/issue35810)

Anyways, the TL;DR is that EVP_dealloc should now be:
```
static void
EVP_dealloc(EVPobject *self)
{
    PyTypeObject *tp = Py_TYPE(self);
    if (self->lock != NULL)
        PyThread_free_lock(self->lock);
    EVP_MD_CTX_free(self->ctx);
    PyObject_Del(self);
    Py_DECREF(tp);
}
```

tiran:
> Since EVP is now a heap allocated type, you should now properly refcount the type as well. You can read the details in: [docs.python.org/3.8/whatsnew/3.8.html#changes-in-the-c-api](https://docs.python.org/3.8/whatsnew/3.8.html#changes-in-the-c-api), search for [bpo-35810](https://bugs.python.org/issue35810)

Thanks! I missed the note. :)

eduardo-elizondo (approved):
Perfect! This LGTM now!

eduardo-elizondo:
Agh I forget I don't have the commit bit here. Let's ping someone else to merge, cc @ericsnowcurrently  @DinoV

PR #16077, opened 2019-09-13
bpo-38148: Add slots to asyncio transports

asvetlov:
https://bugs.python.org/issue38148

### DIFF
--- a/Misc/NEWS.d/next/Library/2019-09-13-08-55-43.bpo-38148.Lnww6D.rst
+++ b/Misc/NEWS.d/next/Library/2019-09-13-08-55-43.bpo-38148.Lnww6D.rst
@@ -1,1 +1,1 @@
-Add slots to asyncio transports.
+Add slots to :mod:`asyncio` transport classes, which can reduce memory usage.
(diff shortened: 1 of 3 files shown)


Misc/NEWS.d/next/Library/2019-09-13-08-55-43.bpo-38148.Lnww6D.rst
```diff
@@ -0,0 +1 @@
+Add slots to asyncio transports.
```
aeros: Thanks for the PR @asvetlov.

Could we also slightly mention the purpose/benefit of the slots? Also, it might help to explicit mention that we're specifically talking about the transport *classes*, so it makes more sense as to what is being referred to. 

```suggestion
Add slots to :mod:`asyncio` transport classes, which can reduce memory usage.
```

Also added a Sphinx role to link to the module itself. Not 100% needed for a Misc/NEWS entry, but it doesn't hurt.

asvetlov: Agree

aeros (approved):

PR #16078, opened 2019-09-13
bpo-34533: Remove all static state from _csv module

eduardo-elizondo:
Add all the changes needed to make _csv compatible with PEP384 as well as removing all static state from it

https://bugs.python.org/issue34533

### DIFF
--- a/Lib/test/test_csv.py
+++ b/Lib/test/test_csv.py
@@ -454,7 +454,7 @@ class TestDialectRegistry(unittest.TestCase):
     def test_pickle(self):
         for name in csv.list_dialects():
             dialect = csv.get_dialect(name)
-            for proto in range(pickle.HIGHEST_PROTOCOL + 1):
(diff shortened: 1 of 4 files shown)


encukou:
The PoC/WIP implementation of PEP 573 should do this plus also remove reliance `PyState_FindModule`.
I'll concentrate on that PEP rather than on reviewing this.

eduardo-elizondo:
Yup! PEP573 should be the long term solution for getting the module state in C Extensions. We can always go back and update this module once the PEP is finalized. On that note, I have a couple of comments/questions that I've been meaning to post in capi-sig about PEP573. I should prioritize that as well :-) 

For now though, this should be ready to be merged as is (pending any in-line comments reviewers might have).

ericsnowcurrently:
Sorry, for the delay, @eduardo-elizondo.  I plan on taking a look next Friday.

encukou:
Here is my attempt: https://github.com/python/cpython/pull/22838

PR #16086, opened 2019-09-13
bpo-38145: Fix short option d for bdist_dumb

sinoroc:
In the 'distutils' command 'bdist_dumb', the short option 'd' is used
for both the 'bdist-dir' and 'dist-dir' options.

The bdist-dir option appeared first on 2000-05-13 in commit: ba0506b3492a13f5c8cec7598bf2b5f9735ac7b2

The dist-dir option appeared then on 2000-07-05 in commit: c4eb84accb1ac1e09ea78bc3af81acdbe9499fb8

There appears to have been no version released between these two
commits, so most likely the global behaviour of the command has
stayed consistent.

The short option d actually triggers the dist-dir option, not the
bdist-dir option.

It is therefore safe to change the short option for bdist-dir. A
choice consistent with other similar distutils commands is 'b'.

https://bugs.python.org/issue38145

### DIFF
--- a/Lib/distutils/command/bdist_dumb.py
+++ b/Lib/distutils/command/bdist_dumb.py
@@ -16,7 +16,7 @@ class bdist_dumb(Command):
 
     description = "create a \"dumb\" built distribution"
 
-    user_options = [('bdist-dir=', 'd',
+    user_options = [('bdist-dir=', 'b',
                      "temporary directory for creating the distribution"),
                     ('plat-name=', 'p',
                      "platform name to embed in generated filenames "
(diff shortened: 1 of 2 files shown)


merwok:
In the absence of automated tests, could you copy/paste a shell session that demonstrates usage of the new short option?

sinoroc:
Not sure what a meaningful shell session should look like in this case. So let's start with this and we can eventually improve from there:

```
$ .venv/bin/python -c "import platform; print(platform.python_build());"
('heads/bpo-38145:ff8057a03c', 'Oct  8 2019 16:11:35')
$ .venv/bin/python setup.py -v bdist_dumb -d alt-dist -b alt-build
running bdist_dumb
running build
installing to alt-build
running install
running install_egg_info
Creating alt-build/tmp/project/.venv/lib/python3.9/site-packages/
Writing alt-build/tmp/project/.venv/lib/python3.9/site-packages/Project-1.0.0-py3.9.egg-info
changing into 'alt-build'
Creating tar archive
changing back to '/tmp/project'
removing 'alt-build' (and everything under it)
$ ls alt-dist/
Project-1.0.0.linux-x86_64.tar.gz
```

merwok:
Transfered to https://github.com/pypa/distutils/issues/18

PR #16090, opened 2019-09-13
bpo-38157: Add example about per file output for mock_open.

tirkarthi:
* Add an example with using `side_effect` to return content per file.

https://bugs.python.org/issue38157


matrixise:
@tirkarthi Do you want to add a blurb entry?

tirkarthi:
I feel it's not necessary since it's adding only documentation example.

corona10 (approved):
@tirkarthi  
Looks good to me.
Let's wait for core developers review :)

Thanks for your contribution.

cjw296:
Would it be possible to add a unit test that exercises this pattern too?
(I wish we could do literal includes with the cpython Sphinx docs, so the examples actually get tested, thoughts?)

tirkarthi:
> Would it be possible to add a unit test that exercises this pattern too?

unit test as a test in code or as doctest itself like below?

```
>>> DEFAULT = "default"
>>> data_dict = {"file1": "data1", "file2": "data2"}
>>> def open_side_effect(name):
...     return mock_open(read_data=data_dict.get(name, DEFAULT))()
...
>>> with patch("builtins.open", side_effect=open_side_effect):
...     with open("file1") as file1:
...         assert file1.read() == "data1"
...     with open("file2") as file2:
...         assert file2.read() == "data2"
...     with open("file3") as file2:
...         assert file2.read() == "default"
```

cjw296:
If we could do literal includes, then we could include the actual unit test that gets run. Not sure that's possible?

A doctest such as the one you show: where would that get executed (and so tested...)?

Doc/library/unittest.mock-examples.rst
```diff
@@ -360,6 +360,30 @@ of arbitrary attributes as well as the getting of them then you can use
 *spec_set* instead of *spec*.
 
 
+Using side_effect to return per file content
+~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
+
+:func:`mock_open` is used to patch :func:`open` method. :attr:`~Mock.side_effect`
+can be used to return a new Mock object per call. This can be used to return different
+content per file stored in a dictionary::
```
slateny:
```suggestion
contents per file stored in a dictionary::
```

Doc/library/unittest.mock-examples.rst
```diff
@@ -360,6 +360,30 @@ of arbitrary attributes as well as the getting of them then you can use
...
+
+    DEFAULT = "default"
```
slateny: This should probably be initially indented 3 spaces as convention (see the example `def f(a, b, c): pass` and [devguide](https://devguide.python.org/documentation/style-guide/#use-of-whitespace)) above.

hugovk (approved):
Thanks!

hugovk:
@corona10:

> @tirkarthi
> Looks good to me.
> Let's wait for core developers review :)

You're both core devs now :) :tada:

tirkarthi:
Thanks @hugovk

PR #16091, opened 2019-09-13
bpo-36046: Fix buildbot failures from the PR.

gpshead:
Varying user/group/permission check needs on platforms.

https://bugs.python.org/issue36046

### DIFF
--- a/Lib/test/test_subprocess.py
+++ b/Lib/test/test_subprocess.py
@@ -1762,8 +1762,10 @@ class POSIXProcessTestCase(BaseTestCase):
                             [sys.executable, "-c",
                              "import os; print(os.getuid())"],
                             user=user)
+                except PermissionError:  # errno.EACCES
+                    pass
                 except OSError as e:
-                    if e.errno != errno.EPERM:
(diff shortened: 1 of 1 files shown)


Lib/test/test_subprocess.py
```diff
@@ -1830,7 +1840,15 @@ def test_group_error(self):
     def test_extra_groups(self):
         gid = os.getegid()
         group_list = [65534 if gid != 65534 else 65533]
-        name_group = "nogroup" if sys.platform != 'darwin' else "staff"
+        for name_group in ('staff', 'nogroup', 'grp'):
```
Yhg1s: Since this is repeated twice, factor it out.

Lib/test/test_subprocess.py
```diff
@@ -1785,11 +1787,23 @@ def test_user_error(self):
         with self.assertRaises(ValueError):
             subprocess.check_call([sys.executable, "-c", "pass"], user=65535)
 
+    @staticmethod
```
Yhg1s: ffs, function, not staticmethod.

PR #12818, opened 2019-04-13
[WIP] bpo-17013: Implement WaitableMock to create Mock objects that can wait until called

tirkarthi:
This is an initial implementation with preliminary docs and tests to see if it's worthy enough of addition.

Some notes : 

* Currently, it doesn't support waiting for calls with keyword arguments.
* It seems the calls to event object are also recorded in `mock_calls` . I think these should be filtered out or maybe I am doing something wrong.
* There is per call event object and per mock object event to track if mock was called with given arguments and to store just mock is called or not. Is there a better way to store this?

cc : @mariocj89 

https://bugs.python.org/issue17013

PR #16094, opened 2019-09-13
gh-61215: New mock to wait for multi-threaded events to happen

mariocj89:
This PR implements the change proposed in [bpo-17013](https://bugs.python.org/issue17013) and takes over https://github.com/python/cpython/pull/12818 (agreed offline with @tirkarthi).

https://bugs.python.org/issue17013

CC: @tirkarthi @voidspace @lisroach 

* Issue: gh-61215

### DIFF
--- a/Doc/library/unittest.mock.rst
+++ b/Doc/library/unittest.mock.rst
@@ -205,8 +205,10 @@ The Mock Class
     import asyncio
     import inspect
     import unittest
+    import threading
     from unittest.mock import sentinel, DEFAULT, ANY
     from unittest.mock import patch, call, Mock, MagicMock, PropertyMock, AsyncMock
+    from unittest.mock import ThreadingMock
     from unittest.mock import mock_open
 
 :class:`Mock` is a flexible mock object intended to replace the use of stubs and
@@ -1097,6 +1099,51 @@ object::
       >>> asyncio.run(main('bar'))
       >>> mock.await_args_list
       [call('foo'), call('bar')]
+
+
+.. class:: ThreadingMock(spec=None, side_effect=None, return_value=DEFAULT, wraps=None, name=None, spec_set=None, unsafe=False, *, timeout=UNSET, **kwargs)
+
+  A version of :class:`MagicMock` for multithreading tests. The
+  :class:`ThreadingMock` object provides extra methods to wait for a call to
+  be invoked, rather than assert on it immediately.
+
+  The default timeout is specified by the ``timeout`` argument, or if unset by the
+  :attr:`ThreadingMock.DEFAULT_TIMEOUT` attribute, which defaults to blocking (``None``).
+
+  You can configure the global default timeout by setting :attr:`ThreadingMock.DEFAULT_TIMEOUT`.
+
+  .. method:: wait_until_called(*, timeout=UNSET)
+
+      Waits until the mock is called.
+
+      If a timeout was passed at the creation of the mock or if a timeout
+      argument is passed to this function, the function raises an
+      :exc:`AssertionError` if the call is not performed in time.
+
+        >>> mock = ThreadingMock()
+        >>> thread = threading.Thread(target=mock)
+        >>> thread.start()
+        >>> mock.wait_until_called(timeout=1)
+        >>> thread.join()
+
+  .. method:: wait_until_any_call(*args, **kwargs)
+
+      Waits until the the mock is called with the specified arguments.
+
+      If a timeout was passed at the creation of the mock
+      the function raises an :exc:`AssertionError` if the call is not performed in time.
+
+        >>> mock = ThreadingMock()
+        >>> thread = threading.Thread(target=mock, args=("arg1", "arg2",), kwargs={"arg": "thing"})
+        >>> thread.start()
+        >>> mock.wait_until_any_call("arg1", "arg2", arg="thing")
+        >>> thread.join()
+
+  .. attribute:: DEFAULT_TIMEOUT
+
+    Global default timeout in seconds to create instances of :class:`ThreadingMock`.
+
+  .. versionadded:: 3.13
 
 
 Calling
--- a/Lib/test/test_unittest/testmock/testthreadingmock.py
+++ b/Lib/test/test_unittest/testmock/testthreadingmock.py
@@ -1,0 +1,278 @@
+import time
+import unittest
+import concurrent.futures
+
+from unittest.mock import patch, ThreadingMock, call
+
+
+class Something:
+    def method_1(self):
+        pass
+
+    def method_2(self):
+        pass
+
+
+class TestThreadingMock(unittest.TestCase):
+    def _call_after_delay(self, func, /, *args, **kwargs):
+        time.sleep(kwargs.pop("delay"))
+        func(*args, **kwargs)
+
+    def setUp(self):
+        self._executor = concurrent.futures.ThreadPoolExecutor(max_workers=5)
+
+    def tearDown(self):
+        self._executor.shutdown()
+
+    def run_async(self, func, /, *args, delay=0, **kwargs):
+        self._executor.submit(
+            self._call_after_delay, func, *args, **kwargs, delay=delay
+        )
+
+    def _make_mock(self, *args, **kwargs):
+        return ThreadingMock(*args, **kwargs)
+
+    def test_spec(self):
+        waitable_mock = self._make_mock(spec=Something)
+
+        with patch(f"{__name__}.Something", waitable_mock) as m:
+            something = m()
+
+            self.assertIsInstance(something.method_1, ThreadingMock)
+            self.assertIsInstance(something.method_1().method_2(), ThreadingMock)
+
+            with self.assertRaises(AttributeError):
+                m.test
+
+    def test_side_effect(self):
+        waitable_mock = self._make_mock()
+
+        with patch(f"{__name__}.Something", waitable_mock):
+            something = Something()
+            something.method_1.side_effect = [1]
+
+            self.assertEqual(something.method_1(), 1)
+
+    def test_instance_check(self):
+        waitable_mock = self._make_mock()
+
+        with patch(f"{__name__}.Something", waitable_mock):
+            something = Something()
+
+            self.assertIsInstance(something.method_1, ThreadingMock)
+            self.assertIsInstance(something.method_1().method_2(), ThreadingMock)
+
+    def test_dynamic_child_mocks_are_threading_mocks(self):
+        waitable_mock = self._make_mock()
+        self.assertIsInstance(waitable_mock.child, ThreadingMock)
+
+    def test_dynamic_child_mocks_inherit_timeout(self):
+        mock1 = self._make_mock()
+        self.assertIs(mock1._mock_wait_timeout, None)
+        mock2 = self._make_mock(timeout=2)
+        self.assertEqual(mock2._mock_wait_timeout, 2)
+        mock3 = self._make_mock(timeout=3)
+        self.assertEqual(mock3._mock_wait_timeout, 3)
+
+        self.assertIs(mock1.child._mock_wait_timeout, None)
+        self.assertEqual(mock2.child._mock_wait_timeout, 2)
+        self.assertEqual(mock3.child._mock_wait_timeout, 3)
+
+        self.assertEqual(mock2.really().__mul__().complex._mock_wait_timeout, 2)
+
+    def test_no_name_clash(self):
+        waitable_mock = self._make_mock()
+        waitable_mock._event = "myevent"
+        waitable_mock.event = "myevent"
+        waitable_mock.timeout = "mytimeout"
+        waitable_mock("works")
+        waitable_mock.wait_until_called()
+        waitable_mock.wait_until_any_call("works")
+
+    def test_wait_success(self):
+        waitable_mock = self._make_mock(spec=Something)
+
+        with patch(f"{__name__}.Something", waitable_mock):
+            something = Something()
+            self.run_async(something.method_1, delay=0.01)
+            something.method_1.wait_until_called()
+            something.method_1.wait_until_any_call()
+            something.method_1.assert_called()
+
+    def test_wait_success_with_instance_timeout(self):
+        waitable_mock = self._make_mock(timeout=1)
+
+        with patch(f"{__name__}.Something", waitable_mock):
+            something = Something()
+            self.run_async(something.method_1, delay=0.01)
+            something.method_1.wait_until_called()
+            something.method_1.wait_until_any_call()
+            something.method_1.assert_called()
+
+    def test_wait_failed_with_instance_timeout(self):
+        waitable_mock = self._make_mock(timeout=0.01)
+
+        with patch(f"{__name__}.Something", waitable_mock):
+            something = Something()
+            self.run_async(something.method_1, delay=0.5)
+            self.assertRaises(AssertionError, waitable_mock.method_1.wait_until_called)
+            self.assertRaises(
+                AssertionError, waitable_mock.method_1.wait_until_any_call
+            )
+
+    def test_wait_success_with_timeout_override(self):
+        waitable_mock = self._make_mock(timeout=0.01)
+
+        with patch(f"{__name__}.Something", waitable_mock):
+            something = Something()
+            self.run_async(something.method_1, delay=0.05)
+            something.method_1.wait_until_called(timeout=1)
+
+    def test_wait_failed_with_timeout_override(self):
+        waitable_mock = self._make_mock(timeout=1)
+
+        with patch(f"{__name__}.Something", waitable_mock):
+            something = Something()
+            self.run_async(something.method_1, delay=0.1)
+            with self.assertRaises(AssertionError):
+                something.method_1.wait_until_called(timeout=0.05)
+            with self.assertRaises(AssertionError):
+                something.method_1.wait_until_any_call(timeout=0.05)
+
+    def test_wait_success_called_before(self):
+        waitable_mock = self._make_mock()
+
+        with patch(f"{__name__}.Something", waitable_mock):
+            something = Something()
+            something.method_1()
+            something.method_1.wait_until_called()
+            something.method_1.wait_until_any_call()
+            something.method_1.assert_called()
+
+    def test_wait_magic_method(self):
+        waitable_mock = self._make_mock()
+
+        with patch(f"{__name__}.Something", waitable_mock):
+            something = Something()
+            self.run_async(something.method_1.__str__, delay=0.01)
+            something.method_1.__str__.wait_until_called()
+            something.method_1.__str__.assert_called()
+
+    def test_wait_until_any_call_positional(self):
+        waitable_mock = self._make_mock(spec=Something)
+
+        with patch(f"{__name__}.Something", waitable_mock):
+            something = Something()
+            self.run_async(something.method_1, 1, delay=0.1)
+            self.run_async(something.method_1, 2, delay=0.2)
+            self.run_async(something.method_1, 3, delay=0.3)
+            self.assertNotIn(call(1), something.method_1.mock_calls)
+
+            something.method_1.wait_until_any_call(1)
+            something.method_1.assert_called_with(1)
+            self.assertNotIn(call(2), something.method_1.mock_calls)
+            self.assertNotIn(call(3), something.method_1.mock_calls)
+
+            something.method_1.wait_until_any_call(3)
+            self.assertIn(call(2), something.method_1.mock_calls)
+            something.method_1.wait_until_any_call(2)
+
+    def test_wait_until_any_call_keywords(self):
+        waitable_mock = self._make_mock(spec=Something)
+
+        with patch(f"{__name__}.Something", waitable_mock):
+            something = Something()
+            self.run_async(something.method_1, a=1, delay=0.1)
+            self.run_async(something.method_1, b=2, delay=0.2)
+            self.run_async(something.method_1, c=3, delay=0.3)
+            self.assertNotIn(call(a=1), something.method_1.mock_calls)
+
+            something.method_1.wait_until_any_call(a=1)
+            something.method_1.assert_called_with(a=1)
+            self.assertNotIn(call(b=2), something.method_1.mock_calls)
+            self.assertNotIn(call(c=3), something.method_1.mock_calls)
+
+            something.method_1.wait_until_any_call(c=3)
+            self.assertIn(call(b=2), something.method_1.mock_calls)
+            something.method_1.wait_until_any_call(b=2)
+
+    def test_wait_until_any_call_no_argument_fails_when_called_with_arg(self):
+        waitable_mock = self._make_mock(timeout=0.01)
+
+        with patch(f"{__name__}.Something", waitable_mock):
+            something = Something()
+            something.method_1(1)
+
+            something.method_1.assert_called_with(1)
+            with self.assertRaises(AssertionError):
+                something.method_1.wait_until_any_call()
+
+            something.method_1()
+            something.method_1.wait_until_any_call()
+
+    def test_wait_until_any_call_global_default(self):
+        with patch.object(ThreadingMock, "DEFAULT_TIMEOUT"):
+            ThreadingMock.DEFAULT_TIMEOUT = 0.01
+            m = self._make_mock()
+            with self.assertRaises(AssertionError):
+                m.wait_until_any_call()
+            with self.assertRaises(AssertionError):
+                m.wait_until_called()
+
+            m()
+            m.wait_until_any_call()
+        assert ThreadingMock.DEFAULT_TIMEOUT != 0.01
+
+    def test_wait_until_any_call_change_global_and_override(self):
+        with patch.object(ThreadingMock, "DEFAULT_TIMEOUT"):
+            ThreadingMock.DEFAULT_TIMEOUT = 0.01
+
+            m1 = self._make_mock()
+            self.run_async(m1, delay=0.1)
+            with self.assertRaises(AssertionError):
+                m1.wait_until_called()
+
+            m2 = self._make_mock(timeout=1)
+            self.run_async(m2, delay=0.1)
+            m2.wait_until_called()
+
+            m3 = self._make_mock()
+            self.run_async(m3, delay=0.1)
+            m3.wait_until_called(timeout=1)
+
+            m4 = self._make_mock()
+            self.run_async(m4, delay=0.1)
+            m4.wait_until_called(timeout=None)
+
+            m5 = self._make_mock(timeout=None)
+            self.run_async(m5, delay=0.1)
+            m5.wait_until_called()
+
+        assert ThreadingMock.DEFAULT_TIMEOUT != 0.01
+
+    def test_reset_mock_resets_wait(self):
+        m = self._make_mock(timeout=0.01)
+
+        with self.assertRaises(AssertionError):
+            m.wait_until_called()
+        with self.assertRaises(AssertionError):
+            m.wait_until_any_call()
+        m()
+        m.wait_until_called()
+        m.wait_until_any_call()
+        m.assert_called_once()
+
+        m.reset_mock()
+
+        with self.assertRaises(AssertionError):
+            m.wait_until_called()
+        with self.assertRaises(AssertionError):
+            m.wait_until_any_call()
+        m()
+        m.wait_until_called()
+        m.wait_until_any_call()
+        m.assert_called_once()
+
+
+if __name__ == "__main__":
+    unittest.main()
--- a/Lib/unittest/mock.py
+++ b/Lib/unittest/mock.py
@@ -14,6 +14,7 @@ __all__ = (
     'call',
     'create_autospec',
     'AsyncMock',
+    'ThreadingMock',
     'FILTER_DIR',
     'NonCallableMock',
     'NonCallableMagicMock',
@@ -32,6 +33,7 @@ import sys
 import builtins
 import pkgutil
 from asyncio import iscoroutinefunction
+import threading
 from types import CodeType, ModuleType, MethodType
 from unittest.util import safe_repr
 from functools import wraps, partial
@@ -3001,6 +3003,98 @@ class PropertyMock(Mock):
         return self()
     def __set__(self, obj, val):
         self(val)
+
+
+_timeout_unset = sentinel.TIMEOUT_UNSET
+
+class ThreadingMixin(Base):
+
+    DEFAULT_TIMEOUT = None
+
+    def _get_child_mock(self, /, **kw):
+        if "timeout" in kw:
+            kw["timeout"] = kw.pop("timeout")
+        elif isinstance(kw.get("parent"), ThreadingMixin):
+            kw["timeout"] = kw["parent"]._mock_wait_timeout
+        elif isinstance(kw.get("_new_parent"), ThreadingMixin):
+            kw["timeout"] = kw["_new_parent"]._mock_wait_timeout
+        return super()._get_child_mock(**kw)
+
+    def __init__(self, *args, timeout=_timeout_unset, **kwargs):
+        super().__init__(*args, **kwargs)
+        if timeout is _timeout_unset:
+            timeout = self.DEFAULT_TIMEOUT
+        self.__dict__["_mock_event"] = threading.Event()  # Event for any call
+        self.__dict__["_mock_calls_events"] = []  # Events for each of the calls
+        self.__dict__["_mock_calls_events_lock"] = threading.Lock()
+        self.__dict__["_mock_wait_timeout"] = timeout
+
+    def reset_mock(self, /, *args, **kwargs):
+        """
+        See :func:`.Mock.reset_mock()`
+        """
+        super().reset_mock(*args, **kwargs)
+        self.__dict__["_mock_event"] = threading.Event()
+        self.__dict__["_mock_calls_events"] = []
+
+    def __get_event(self, expected_args, expected_kwargs):
+        with self._mock_calls_events_lock:
+            for args, kwargs, event in self._mock_calls_events:
+                if (args, kwargs) == (expected_args, expected_kwargs):
+                    return event
+            new_event = threading.Event()
+            self._mock_calls_events.append((expected_args, expected_kwargs, new_event))
+        return new_event
+
+    def _mock_call(self, *args, **kwargs):
+        ret_value = super()._mock_call(*args, **kwargs)
+
+        call_event = self.__get_event(args, kwargs)
+        call_event.set()
+
+        self._mock_event.set()
+
+        return ret_value
+
+    def wait_until_called(self, *, timeout=_timeout_unset):
+        """Wait until the mock object is called.
+
+        `timeout` - time to wait for in seconds, waits forever otherwise.
+        Defaults to the constructor provided timeout.
+        Use None to block undefinetively.
+        """
+        if timeout is _timeout_unset:
+            timeout = self._mock_wait_timeout
+        if not self._mock_event.wait(timeout=timeout):
+            msg = (f"{self._mock_name or 'mock'} was not called before"
+                   f" timeout({timeout}).")
+            raise AssertionError(msg)
+
+    def wait_until_any_call(self, *args, **kwargs):
+        """Wait until the mock object is called with given args.
+
+        Waits for the timeout in seconds provided in the constructor.
+        """
+        event = self.__get_event(args, kwargs)
+        if not event.wait(timeout=self._mock_wait_timeout):
+            expected_string = self._format_mock_call_signature(args, kwargs)
+            raise AssertionError(f'{expected_string} call not found')
+
+
+class ThreadingMock(ThreadingMixin, MagicMixin, Mock):
+    """
+    A mock that can be used to wait until on calls happening
+    in a different thread.
+
+    The constructor can take a `timeout` argument which
+    controls the timeout in seconds for all `wait` calls of the mock.
+
+    You can change the default timeout of all instances via the
+    `ThreadingMock.DEFAULT_TIMEOUT` attribute.
+
+    If no timeout is set, it will block undefinetively.
+    """
+    pass
 
 
 def seal(mock):
--- a/Misc/NEWS.d/next/Library/2019-09-13-13-28-10.bpo-17013.NWcgE3.rst
+++ b/Misc/NEWS.d/next/Library/2019-09-13-13-28-10.bpo-17013.NWcgE3.rst
@@ -1,0 +1,3 @@
+Add ``ThreadingMock`` to :mod:`unittest.mock` that can be used to create
+Mock objects that can wait until they are called. Patch by Karthikeyan
+Singaravelan and Mario Corchero.


mariocj89:
>  Support for multiprocessing? If not, maybe rename to MultithreadingMock?

Spoke offline with @voidspace and suggested to keep this PR for the multithreading use case only as it is not trivial to get this working for multiprocess. It was also suggested to rename the mock to make it clear it is about multithreading and keep the `Event` as private as we are not exposing it anymore.


@tirkarthi @voidspace @lisroach if this looks good I'll get started with the docs.

Lib/unittest/mock.py
```diff
@@ -32,9 +33,11 @@
 import pprint
 import sys
 import builtins
+import threading
 from types import CodeType, ModuleType, MethodType
 from unittest.util import safe_repr
 from functools import wraps, partial
+from collections import defaultdict
```
tirkarthi: `defaultdict` is not used and can be removed. I started my original PR to store call object as key mapped to event but call objects are not always hashable.

Lib/unittest/mock.py
```diff
@@ -2816,6 +2819,59 @@ def __set__(self, obj, val):
...
 
+class ThreadingMock(MagicMock):
+    """
+    A mock that can be used to wait until on calls happening
+    in a different thread.
+    """
+
+    def __init__(self, *args, **kwargs):
+        _safe_super(ThreadingMock, self).__init__(*args, **kwargs)
+        self._event = threading.Event()
```
tirkarthi: Are we going to allow customization of the event class?

Lib/unittest/mock.py
```diff
@@ -2816,6 +2819,59 @@ def __set__(self, obj, val):
         self(val)
 
 
+class ThreadingMock(MagicMock):
```
tirkarthi: My discussion with @voidspace was to use Mixins instead of directly inheriting `MagicMock`.

Lib/unittest/mock.py
mariocj89: Wops, good catch, forgot to remove it in https://github.com/python/cpython/pull/16094/commits/28c3974d37b665a026af53c9e106c42db93541e9

Lib/unittest/mock.py
mariocj89: From a chat with @voidspace (see https://github.com/python/cpython/pull/16094#issuecomment-531314327), passing `multiprocessing.Event` does not work. It needs further support especially if we want to have the `wait_until_any_call`. That is also why we renamed the mock to be specific to threading. Once we get this through we can look at supporting multiprocessing if there is a need.

Lib/unittest/test/testmock/testthreadingmock.py
```diff
@@ -0,0 +1,173 @@
...
+
+            self.assertIsInstance(something.method_1, ThreadingMock)
+            self.assertIsInstance(
+                something.method_1().method_2(), ThreadingMock)
+
+
+    def test_side_effect(self):
+        waitable_mock = self._make_mock()
+
+        with patch(f'{__name__}.Something', waitable_mock):
```
mariocj89: For the reviewer: I left these tests with the patch function. All tests are using `patch` but I can change this to have a single test verifying patch and use the mock directly on the rest to simplify the test scenarios.

Lib/unittest/mock.py
mariocj89: I can change this to have a `ThreadingMixin` and then have `ThreadingMock` to inherit from `MagicMock` and `ThreadingMixing`.

@tirkarthi @lisroach @voidspace do you prefer that?

mariocj89:
Gentle ping @tirkarthi @voidspace @lisroach! Let me know if the general idea/naming looks good and I'll get to work on the documentation.

tirkarthi:
Sorry, got caught up with new job. I will try to get to the PR after two weeks.

tirkarthi:
I am fine with the PR. The primary concern in my implementation was to have mock_timeout fit in the API along with either to have a default or wait forever and hope we have reached some consensus on it as per the PR. It would be helpful if one of the core dev accepts this idea for merge. Also does it needs some more discussion from other channels like discourse, mailing list etc since I don't have experience in proposing new API. Thanks Mario for picking up the PR.

Doc/library/unittest.mock.rst
```diff
@@ -1078,6 +1080,37 @@ object::
...
+      If ``mock_timeout`` is set, after that number of seconds waiting,
+      it raises an :exc:`AssertionError`, waits forever otherwise.
+
+        >>> mock = ThreadingMock()
+        >>> thread = threading.Thread(target=mock)
+        >>> thread.start()
+        >>> mock.wait_until_called(mock_timeout=1)
+        >>> thread.join()
+
+  .. method:: wait_until_any_call(*args, mock_timeout=None, **kwargs)
```
vstinner: I would prefer to put the timeout as the first parameter, maybe as a positional-only parameter to avoid any risk of conflict with **kwargs. If you want to block, pass explicitly None. If having to pass None bothers you, maybe add a second function which always block?

mariocj89: We tried to namespace it as it is done with other arguments in `mock` to avoid conflicts indeed.

I did not think about potentially having 2 different methods though, that might be nice indeed.

Something like:

`wait_until_any_call(*args, **kwargs)` and `wait_until_any_call_with_timeout(timeout, /, *args, **kwargs)`?

Thoughts @lisroach @voidspace @tirkarthi ?

Alternatively, as mentioned before, we can take the timeout in the constructor.

vstinner: Usually timeout=None is called "blocking" in Python, "timeout=0.0" is called "non-blocking" and "timeout=value" has no clear name... If we choose to have 2 methods, maybe "blocking" could be used for the timeout=None case: block_until_called(...) vs wait_until_called(timeout, ...).

I'm not strongly opposed to mock_timeout=value parameter.

Is it a good practice to use a wait without upper bound in an unit test? For example, is it fine if your test completes in 1 minute? 1 hour? 1 day? 1 week?

My experience with the Python test suite is a that a timeout helps to get an output when a test hangs.

mariocj89: I am slightly more inclined to just having `mock_timeout`, I think it is the simplest and easiest to discover, the methods are similar to the existing ones and across the two of them.
The "only" downside is that they might collide with an existing argument, but we are namespacing it already, it is `mock_timeout`, not `timeout`. Having [a quick search in github](https://github.com/search?l=Python&q=mock_timeout+NOT+test&type=Code) seems to indicate it might not be an issue. Additionally, if users wanted to fake a function that had an argument `mock_timeout` they can do it so by using `side_effect` (basically not being able to use this function). I think @voidspace thought similarly (namespacing `timeout` with `mock` should be good enough).

In terms of having a default timeout, I agree with [your previous comment here](https://github.com/python/cpython/pull/12818#discussion_r275276284) which prefers to not have a default at all. Any default is trying to guess what the user is mocking, which can be result in flaky tests. Say you are working with a function that takes 1 second and we set the 1 second as a default for example.
The "only harm" with blocking forever is that the test suite will not finish and it will be eventually terminated by something else (user, CI, pytest, etc.). I think it is much better to have a default that blocks forever (so there are no surprises for the user based on the environment for example) and let them specify a timeout if they really want, as they will know what is the right value.

mariocj89: As a note, IIRC when speaking with @voidspace, he liked the idea of having the timeout namespaced. He can further comment.

vstinner:
> As a note, IIRC when speaking with @voidspace, he liked the idea of having the timeout namespaced. He can further comment.

What does it mean to have timeout namespaced? Have a parameter called "mock_timeout"?

mariocj89:
Yeah, just prepending it with “mock_”, mothibg fancy.

 having just ‘timeout’ can easily collide with existing functions that take a timeout parameter.

Lib/unittest/test/testmock/testthreadingmock.py
```diff
@@ -0,0 +1,173 @@
+import threading
+import multiprocessing
+import time
+import unittest
+import concurrent.futures
+
+from test.support import start_threads
```
Kentzo: Doesn't seem to be used.

mariocj89: Good catch, I refactored this to work with futures.ThreadPoolExecutor, I need neither that nor multiprocessing nor threading.

mariocj89:
Gentle ping @lisroach @voidspace, just wished yesterday that I had this mock in a codebase :).

cjw296:
@mariocj89 - might be worth emailing or otherwise contacting @lisroach and @voidspace offline, I know GitHub notifications can be noisy for some people and a ping on here may not get to them.

Doc/library/unittest.mock.rst
```diff
@@ -1078,6 +1080,37 @@ object::
       [call('foo'), call('bar')]
 
 
+.. class:: ThreadingMock(spec=None, side_effect=None, return_value=DEFAULT, wraps=None, name=None, spec_set=None, unsafe=False, **kwargs)
```
cjw296: Threaded rather than Threading maybe?

mariocj89: Happy to look for a better name. We discussed EventMock as well in the sprints, but could not find a better name. I'm not a native English speaker, if ThreadedMock "is more correct", happy to rename it.

cjw296: I guess it depends what it is, if it's a mock that helps when dealing with threading, then ThreadingMock feels right. If it's a mock that is threaded, then ThreadedMock instead. I'm afraid I don't understand the use case enough to give you a definitive answer.

mariocj89: It is the former, so I guess the naming is fine :). Thanks!

Kentzo:
I don't think adding an additional class is justified enough. It does not really make Mock anymore "threaded" than it already is, just adds a convenience method on top of the threading facility. @cjw296 was already confused with the name above. Recall that unittest is somewhat hard to navigate already.

I'm also not convinced that implementation is convenient enough: e.g. the implementation does not allow to wait for arbitrary conditions.

One way or [another](https://github.com/python/cpython/pull/17133) let's get it through.

Kentzo:
Perhaps with the release of 3.9 it's time to consider this change for 3.10?

Lib/unittest/mock.py
lisroach: @mariocj89 sorry for such a late reply, but yes I agree with using mixins to keep it consistent with the rest of the module

Kentzo:
@lisroach Yes, the issue that was pointed out (I changed the `called` property from a boolean to a boolean-alike object) was resolved: I added the `call_event` property instead.

cjw296:
Just one plea here: many people are stuck on older versions of Python, and so only get to use new stuff via the backport, if at all possible, please have a thought to writing code in such a way that it'll run on Python 3.6. We already had to skip the "position-only args" commit in the backport, and that was pretty painful.

mariocj89:
> I think @Kentzo makes a good point, does it add enough functionality to justify creating a new class

I'm absolutely fine with that. A year ago there was a discussion where it seems that you and @voidspace where leaning more towards having specific mocks for different functionality, where it was discussed to have this implemented as such. But if things have changed I'm 100% fine with that :).

Should we then close this PR go with @Kentzo proposal?

Kentzo:
@cjw296 I believe changes in https://github.com/python/cpython/pull/20759 are backward compatible as is or, in worst case, after trivial modifications.

Doc/library/unittest.mock.rst
cjw296: I'm a little confused: can you give an example where naming the parameter `timeout` would conflict?

Lib/unittest/mock.py
cjw296: @mariocj89 - I'm confused, I thought I saw a commit where you made this change? Sounds like it should happen before we land this...

Lib/unittest/mock.py
```diff
@@ -3003,6 +3005,59 @@ def __set__(self, obj, val):
...
+    """
+    A mock that can be used to wait until on calls happening
+    in a different thread.
+    """
+
+    def __init__(self, *args, **kwargs):
+        _safe_super(ThreadingMock, self).__init__(*args, **kwargs)
+        self.__dict__["_event"] = threading.Event()
+        self.__dict__["_expected_calls"] = []
+        self.__dict__["_events_lock"] = threading.Lock()
```
cjw296: I wonder if it might be pragmatic to prefix these with `_mock_` to further reduce the chance of conflicting with child attributes of the mocked class?

Lib/unittest/mock.py
```diff
@@ -3003,6 +3005,59 @@ def __set__(self, obj, val):
...
+
+    def __get_event(self, expected_args, expected_kwargs):
+        with self._events_lock:
+            for args, kwargs, event in self._expected_calls:
+                if (args, kwargs) == (expected_args, expected_kwargs):
+                    return event
```
cjw296: This doesn't appear to consume the call, so how do you wait on more than one matching call?

Lib/unittest/mock.py
```diff
@@ -3003,6 +3005,59 @@ def __set__(self, obj, val):
...
+        """Wait until the mock object is called.
+
+        `mock_timeout` - time to wait for in seconds, waits forever otherwise.
+        """
+        if not self._event.wait(timeout=mock_timeout):
+            msg = (f"{self._mock_name or 'mock'} was not called before"
+                   f" timeout({mock_timeout}).")
+            raise AssertionError(msg)
+
+    def wait_until_any_call(self, *args, mock_timeout=None, **kwargs):
```
cjw296: If this took a `Call` instance, could we then safely name the other parameter just `timeout`?

Doc/library/unittest.mock.rst
mariocj89: Say you want to mock the following function:

```python
def my_cool_socket_func(timeout=3):
    pass
```

You then wait for it to be called as:
```
m = ThreadingMock(my_cool_socket_func)
# some other thread runs: m(timeout=4)
m.wait_until_any_call(timeout=4)
```

The user might expect the `wait_until_any_call` to pass, but it will fail after 4 seconds, as the timeout keyword matches our waiting timeout, not the parameter timeout. Not a huge concern though.

We could split the function as Victor was suggesting and just rename it to timeout if you prefer, having:

```
def block_until_any_call(*args, **kwargs):
    pass
def wait_until_any_call(timeout, /, *args, **kwargs):
    pass
```

This way:
- Block function is kept clean and just takes the arguments as the function takes them when it is called.
- Wait function always takes timeout as the first argument and as it is possitional only, it cannot collide with a function parameter "timeout".

In short:

```
m = ThreadingMock(my_cool_socket_func)
# some other thread runs: m(timeout=4)
m.wait_until_any_call(timeout=4)  # Fails, missing positional arg timeout
m.block_until_any_call(timeout=4)  # Works, blocks until it is called
m.wait_until_any_call(4)  # Works, blocks for 4 seconds
```

cjw296: The trouble with positional-only stuff is it makes backporting really hard :-( 
(I'm also not sure it doesn't cause other problems of the a similar type...)

How about having `waitaible_for_call(*args, **kw)` that returns an object that has a `wait(timeout=None)` method?
(a bit like `mock_calls`, and with maybe `waitable_for_any_call()` that return an object that waits for the next call of any type?)

Lib/unittest/mock.py
mariocj89: You can just wait on them, that way you know which one failed easily:
```python
block_until_any_call("arg1", kwarg2="no")
block_until_any_call("arg2", kwarg2="yes")
```

If you want to specify order, you can use assert:

```python
block_until_any_call("arg2", kwarg2="yes")
assert_calls(call("arg1", kwarg2="no"), call("arg2", kwarg2="yes"))
```

My worry of implementing `wait_until_calls` is that it will continue waiting until the calls happen in the specific order.

If we want full completeness, I'd suggest also implementing `block_until_call`, to wait for *the first call* to match the specified arguments. My worry is how to define this behaviour though. Do we match only the next call or do we it allow to match the last one as well? Think it'd be hard to coordinate this on a multithreading program.

mariocj89:
Missing work to land this (I can get this done before EOW):

# Create a Mixin for MultithreadingMock.
To ease composing Mocks if desired as well.

# Prepend attributes with `_mock`
To prevent collisions

# Add block/wait_untill calls
So that block takes no timeout and wait_until always takes it as the first one, avoiding potential confusions.
With this, we will rename `mock_timeout` to just `timeout` and have it as poossitional only arg in the `wait_until` functions.

# Implement wait_until_calls?
So a user can ask to wait for a list of calls to happen in order. Not sure if it adds much value but I can add it. Let me know if you want me to add that in this PR @cjw296 .

# Implement `block_until_call/wait_until_call`?
To match the next call to happen (or maybe the last recent one). The logic seems a bit to define though, so I explicitly left it out for now. (As no one has asked for it yet TBH). Again, let me know if you want me to add this, but I thought YAGNI.

Doc/library/unittest.mock.rst
mariocj89: > The trouble with positional-only stuff is it makes backporting really hard :-(

Have you considered drooping support for Python < 3.7?
3.7 will be marked EOL this very month.

mariocj89:
@cjw296 let me know what you think about "Implement wait_until_calls?" and "Implement block_until_call/wait_until_call?" from https://github.com/python/cpython/pull/16094#issuecomment-1592582392

Lib/unittest/mock.py
mariocj89: Changed it :)

Doc/library/unittest.mock.rst
cjw296: The backport still supports 3.6!
Unfortunately, cpython's much shortened support windows don't work for everyone...

cjw296:
> @cjw296 let me know what you think about "Implement wait_until_calls?" and "Implement block_until_call/wait_until_call?" from [#16094 (comment)](https://github.com/python/cpython/pull/16094#issuecomment-1592582392)

I agree that I think we can YAGNI this until a good case comes up. I can think of getting the same effect by calling `block_until_call` / `wait_until_call` `n` times and then inspecting `mock_calls` to check the calls were as they should be, but remembering that they might be out of order.

Doc/library/unittest.mock.rst
```diff
@@ -1099,6 +1101,58 @@ object::
...
+  happen on a different thread, rather than assert on it immediately.
+
+  .. method:: block_until_called()
+
+      Blocks until the the mock is called.
+
+        >>> mock = ThreadingMock()
+        >>> thread = threading.Thread(target=mock)
+        >>> thread.start()
+        >>> mock.wait_until_called(1)
```
cjw296: I find this pretty hard to read... what is `1` here? I have to go dig...
```
 >>> mock.wait_until_called(timeout=1)
```
...is clearer, but then we have the name clash.

How about:

```
>>> mock.until_called().wait(timeout=1)
```

Doc/library/unittest.mock.rst
```diff
@@ -1099,6 +1101,58 @@ object::
...
+        >>> thread.join()
+
+  .. method:: block_until_any_call(*args, **kwargs)
+
+      Block until the the mock is called with the specified arguments.
+
+        >>> mock = ThreadingMock()
+        >>> thread = threading.Thread(target=mock, args=("arg1", "arg2",), kwargs={"arg": "thing"})
+        >>> thread.start()
+        >>> mock.wait_until_any_call(1, "arg1", "arg2", arg="thing")
```
cjw296: The above then becomes:
```
>>> mock.until_any_call("arg1", "arg2", arg="thing").wait(timeout=1)
```

I'm not sure we actually need the blocking variants, since that could just be spelled:
```
>>> mock.until_any_call("arg1", "arg2", arg="thing").wait()
```
...which I think is the behaviour you proposed (and which I really like!) for the default timeout being `None` and meaning "don't timeout".

Doc/library/unittest.mock.rst
```diff
@@ -1099,6 +1101,58 @@ object::
       [call('foo'), call('bar')]
 
 
+.. class:: ThreadingMock(spec=None, side_effect=None, return_value=DEFAULT, wraps=None, name=None, spec_set=None, unsafe=False, **kwargs)
+
+  A version of :class:`MagicMock` for multithreading tests. The
+  :class:`ThreadingMock` object provides extra methods to wait for a call to
+  happen on a different thread, rather than assert on it immediately.
+
+  .. method:: block_until_called()
```
cjw296: Looks like this and the following definition are documenting `wait_until_called`?

Doc/library/unittest.mock.rst
mariocj89: `mock.until_called().wait(timeout=1)`

What does `mock.until_called()` do? It worries me that the syntax is quite different from the rest of the mock methods.

mariocj89: > I find this pretty hard to read... what is 1 here? I have to go dig...

Absolutely agree, which is why I was pushing back on it initially. I'd expect most users to either use `block_until_called` or `wait_until_called` with a single timeout for all method. Two ideas I can think of are:

1. Document `wait_until_called` with a constant:
```python
>>> TIMEOUT = 1
>>> mock.wait_until_called(TIMEOUT)
```
2. Find a way to inject this in a different place (I was thinking on the `ThreadingMock` constructor (but again, risk of clash).
```python
>>> mock = ThreadingMock(timeout=1)
>>> mock.wait_until_called()
```

cjw296: I like option 2 here, can you give me an example of a clash when using `ThreadingMock(timeout=1)`?

cjw296: > What does `mock.until_called()` do? It worries me that the syntax is quite different from the rest of the mock methods.

Yeah, I definitely don't love it ;-) `mock.until_called()` would return an object that you can call `wait()` on, nothing more. It's just syntactic sugar to avoid the potential argument name clash. it sort-of has equivalents in the way you set side effect and return values through the existing API, but only if you squint in a certain way...

mariocj89: > I like option 2 here, can you give me an example of a clash when using ThreadingMock(timeout=1)?

Because I was thinking that we'd need to expose that in a property and it could clash with the mocked object properties, but it actually doesnt! We could just take it as the constructor and not expose it in the class. Let me work on a proposal.

mariocj89:
@cjw296 [4f88c29](https://github.com/python/cpython/pull/16094/commits/4f88c29fe48acac88e3d0dc2d9a51d0e4d867059) implements the `timeout` parameter and gives an option for a global timeout. let me know what you think and if you like it I'll cleanup the commit and prepare the PR for a final review.

Doc/library/unittest.mock.rst
```diff
@@ -1101,48 +1101,37 @@ object::
...
-      Raises an :exc:`AssertionError` if the call is not performed.
+      If a timeout was passed at the creation of the mock or if a timeout
+      argument is passed to this function, the function raises an
+      :exc:`AssertionError` if the call is not performed.
 
         >>> mock = ThreadingMock()
         >>> thread = threading.Thread(target=mock)
         >>> thread.start()
-        >>> mock.wait_until_called(1)
+        >>> mock.wait_until_called(timeout=1)
```
cjw296, line 1126: Doesn't this have potential for parameter clashes?

Lib/test/test_unittest/testmock/testthreadingmock.py
```diff
@@ -145,14 +176,14 @@ def test_wait_until_any_call_positional(self):
             self.run_async(something.method_1, 3, delay=0.3)
             self.assertNotIn(call(1), something.method_1.mock_calls)
 
-            something.method_1.block_until_any_call(1)
+            something.method_1.wait_until_any_call(1)
```
cjw296, line 171: The smaller API surface of losing the separate `block_` method is certainly appealing!

Lib/unittest/mock.py
```diff
@@ -3004,13 +3004,32 @@ def __get__(self, obj, obj_type=None):
     def __set__(self, obj, val):
         self(val)
 
+
+_timeout_unset = sentinel.TIMEOUT_UNSET
```
cjw296, line 3008: Nice! Although I think `None` works just as well here:
"What timeout is set?"
"None, it blocks forever"

Lib/unittest/mock.py
```diff
@@ -3004,13 +3004,32 @@ def __get__(self, obj, obj_type=None):
...
 class ThreadingMixin(Base):
 
-    def __init__(self, *args, **kwargs):
+    DEFAULT_TIMEOUT = _timeout_unset
+
+    def _get_child_mock(self, /, **kw):
+        if "timeout" in kw:
+            kw["timeout"] = kw.pop("timeout")
+        elif isinstance(kw.get("parent"), ThreadingMixin):
+            kw["timeout"] = kw["parent"]._mock_wait_timeout
```
cjw296, line 3018: Do we have a test that verifies this? I think it's important functionality :-)

Lib/unittest/mock.py
```diff
@@ -3032,45 +3050,43 @@ def _mock_call(self, *args, **kwargs):
 
         return ret_value
 
-    def wait_until_called(self, timeout, /):
+    def wait_until_called(self, *, timeout=_timeout_unset):
```
cjw296, line 3059: nice!

Lib/unittest/mock.py
```diff
@@ -3032,45 +3050,43 @@ def _mock_call(self, *args, **kwargs):
...
             msg = (f"{self._mock_name or 'mock'} was not called before"
                    f" timeout({timeout}).")
             raise AssertionError(msg)
 
-    def wait_until_any_call(self, timeout, /, *args, **kwargs):
+    def wait_until_any_call(self, *args, **kwargs):
         """Wait until the mock object is called with given args.
 
-        `timeout` - time to wait for in seconds, waits forever otherwise.
+        Waits for the timeout in seconds provided in the constructor.
```
cjw296, line 3076: Yeah, this feels much better!

cjw296 (commented):
This feels like we're on the right track :-)

owillebo:
Should reset_mock re-instantiate the event instances if any?

This allows for waiting for same call multiple times.

mariocj89:
> Should reset_mock re-instantiate the event instances if any?

Absolutely! Let me add that.

Lib/unittest/mock.py
mariocj89: What I was trying to achieve is a user changing the class timeout to be `3` then passing a timeout of `None` to mean "block". If this is None, how do we distinguish a user not passing anything and we getting the default vs they specifically saying "block"?

Lib/unittest/mock.py
mariocj89: Yup, note all tests that invoke `m.method1`. Calling that generates a child mock.
Happy to add more tests if there is anything else you want covered.

Lib/unittest/mock.py
cjw296: Sure, that makes sense :-)

Lib/unittest/mock.py
```diff
@@ -3003,6 +3005,98 @@ def __set__(self, obj, val):
...
+        self.__dict__["_mock_calls_events"] = []  # Events for each of the calls
+        self.__dict__["_mock_events_lock"] = threading.Lock()
+        self.__dict__["_mock_wait_timeout"] = timeout
+
+    def reset_mock(self, /, *args, **kwargs):
+        """
+        See :func:`.Mock.reset_mock()`
+        """
+        super().reset_mock(*args, **kwargs)
+        self.__dict__["_mock_event"] = threading.Event()
```
Kentzo, line 3037: Wouldn't it make sense to set existing events before resetting them in case user's code is holding onto them? A condition seems to be necessary to disambiguate call-set events from reset-set events.

Lib/unittest/mock.py
```diff
@@ -3003,6 +3005,98 @@ def __set__(self, obj, val):
...
+            kw["timeout"] = kw["_new_parent"]._mock_wait_timeout
+        return super()._get_child_mock(**kw)
+
+    def __init__(self, *args, timeout=_timeout_unset, **kwargs):
+        super().__init__(*args, **kwargs)
+        if timeout is _timeout_unset:
+            timeout = self.DEFAULT_TIMEOUT
+        self.__dict__["_mock_event"] = threading.Event()  # Event for any call
+        self.__dict__["_mock_calls_events"] = []  # Events for each of the calls
+        self.__dict__["_mock_events_lock"] = threading.Lock()
```
Kentzo: This lock seems to protect `_mock_calls_events`, perhaps it should named after it as `_mock_calls_events_lock` instead?

Lib/unittest/mock.py
mariocj89: > Wouldn't it make sense to set existing events before resetting them in case user's code is holding onto them?

But that would invalidate previous `wait` calls that the user has requested isn't it? Why would a user want that?

I might have misunderstood it though, if so, can you write some code that illustrates how this would look like from a user perspective?

Lib/unittest/mock.py
mariocj89: Makes total sense! Thanks

cjw296 (approved):
This looks good to go!

Lib/unittest/mock.py
Kentzo: The invalidation is the point. If a user resets a mock while the test is still waiting on it then it’s a bug IMHO. However, this is an opportunity to handle it gracefully telling the user where they messed up.

I don’t have a legible example in mind, only a misuse.

Lib/unittest/mock.py
```diff
@@ -3003,6 +3005,98 @@ def __set__(self, obj, val):
...
+        Use None to block undefinetively.
+        """
+        if timeout is _timeout_unset:
+            timeout = self._mock_wait_timeout
+        if not self._mock_event.wait(timeout=timeout):
+            msg = (f"{self._mock_name or 'mock'} was not called before"
+                   f" timeout({timeout}).")
+            raise AssertionError(msg)
+
+    def wait_until_any_call(self, *args, **kwargs):
```
willemvanoort, line 3073: In terms of consistency with `Mock.assert_called_with`, would `wait_until_called_with` not be a more suitable name here? Or any particular reason that this method name was chosen?

Lib/unittest/mock.py
owillebo: > The invalidation is the point. **If a user resets a mock while the test is still waiting on it** then it’s a bug IMHO. However, this is an opportunity to handle it gracefully telling the user where they messed up.
> 
> I don’t have a legible example in mind, only a misuse.

That would mean that the user test script is multithreaded, one thread that is waiting for the mock to be called and one thread that does the reset_mock.

Must we facilitate multithreaded test scripts?

If so, the methods used to wait for the mock to be called must raise an exception when reset_mock is called.

Kentzo: > Must we facilitate multithreaded test scripts?

No, but we should attempt to give a sensible error. It's a testing library, verbosity is expected, at least I'd appreciate it as a user. But by no means it's a blocker.

> If so, the methods used to wait for the mock to be called must raise an exception when reset_mock is called.

Agree.

mariocj89: That's interesting. We could add that if desired indeed.
The way I saw `reset_mock` is that it resets the recorded calls, not the expectations I had on the the object.

I'd start with YAGNI TBH as even if we were to implement this I'd suggest to add a kw arg to `reset_mock` to invalidate previous `await_X` calls.

mariocj89:
OK! Let's see the results:

# 1. wasm32-emscripten node: 

```RuntimeError: can't start new thread```

# 2. wasm32-emscripten node (pthreads) 3.x 

Fails with:
```
FAIL: test_wait_until_any_call_keywords (test.test_unittest.testmock.testthreadingmock.TestThreadingMock.test_wait_until_any_call_keywords)
  File "/opt/buildbot/bcannon-wasm/3.x.bcannon-wasm.emscripten-node-pthreads/build/Lib/test/test_unittest/testmock/testthreadingmock.py", line 188, in test_wait_until_any_call_keywords
    self.assertNotIn(call(a=1), something.method_1.mock_calls)
AssertionError: call(a=1) unexpectedly found in [call(a=1)]
```


# 3. wasm32-wasi 3.x

`RuntimeError: can't start new thread`

**Comment by mariocj89** [comment]
I'm worried 2 might come from flaky tests, will review them all to make sure we can make them more resilient (as you know, with timeouts we are exposed to the host).

Need also to investigate what is going on with the error to start a thread.

**Comment by tirkarthi** [comment]
@mariocj89  Since the failures are in wasm related builds where threads might not be supported maybe you need to wrap the test file to skip when threads are not working. I see `requires_working_threading` being used in many tests. The docstring also notes threading related tests might fail in wasm builds. I guess wasm related builds are only during full test run and is not present in GitHub actions.

https://github.com/python/cpython/blob/58906213cc5d8f2be311664766b4923ef29dae1f/Lib/test/support/threading_helper.py#L216-L247

**Comment by mariocj89** [comment]
Indeed, just spoke with @pablogsal who told me the same :P.

Will send a PR before EOD.

**Comment by tirkarthi** [comment]
It seems below tests are totally skipped and maybe a related one needs to be added to the mock related tests as well. Since the test will be backported to mock backport project maybe the helper function needs to be inlined in case mock backport runs on wasm builds to not rely on `test.support`

```
rg '^threading_helper.requires_working_threading' Lib/test
Lib/test/test_queue.py
14:threading_helper.requires_working_threading(module=True)

Lib/test/test_thread.py
13:threading_helper.requires_working_threading(module=True)

Lib/test/test_threading_local.py
15:threading_helper.requires_working_threading(module=True)

Lib/test/test_threading.py
29:threading_helper.requires_working_threading(module=True)

Lib/test/test_threadedtempfile.py
24:threading_helper.requires_working_threading(module=True)

Lib/test/test_importlib/test_locks.py
15:threading_helper.requires_working_threading(module=True)

Lib/test/test_importlib/test_threaded_import.py
22:threading_helper.requires_working_threading(module=True)
```

**Comment by mariocj89** [comment]
https://github.com/python/cpython/pull/106366

Will followup trying to harden the rests of the test.

**Comment by mariocj89** [comment]
@cjw296 even though the API looks neat to me, I showed it to a colleague who does not have a lot of mocking experiencet and suggested to use `wait_for_called_with` rather than `wait_for_any_call`.

Whilst I like how `wait_for_any_call` mirrors the naming convention with `assert_any_call` and leaves the door open to use `wait_for_called_with` to fail if the next call does not match the specified call, it might be easier to read to just call it `wait_for_called_with` and if we ever need a "wait for called with X and fail if different" we could name it "wait_for_next_call".

What do you think?

```
mock = ThreadingMock()
mock.wait_for_any_call("arg1", kwarg2=2)   #1
mock.wait_for_call_with("arg1", kwarg2=2)  #2
```

1 or 2?

I'm fine with one, but if we want to change it, it is possibly now or never.

**Comment by cjw296** [comment]
Maybe it should be `wait_for_any_call_with`?

**Comment by mariocj89** [comment]
Sounds good. PR ready: https://github.com/python/cpython/pull/106414

File: Lib/unittest/mock.py [hunk 23]
```diff
@@ -3003,6 +3005,98 @@ def __set__(self, obj, val):
...
 
+_timeout_unset = sentinel.TIMEOUT_UNSET
+
+class ThreadingMixin(Base):
+
+    DEFAULT_TIMEOUT = None
+
+    def _get_child_mock(self, /, **kw):
+        if "timeout" in kw:
+            kw["timeout"] = kw.pop("timeout")
```
cjw296, line 3016: @mariocj89 - this path has no test coverage:
https://app.circleci.com/pipelines/github/testing-cabal/mock/1428/workflows/7b4dab2f-46de-4dd7-bb8b-5d2000975547/jobs/20351

Please can you PR to add some? (and feel free to do the ones from @graingert's [PR](https://github.com/python/cpython/pull/94962) if you have energy ;-)

Lib/test/test_unittest/testmock/testthreadingmock.py
```diff
@@ -0,0 +1,278 @@
...
+
+from unittest.mock import patch, ThreadingMock, call
+
+
+class Something:
+    def method_1(self):
+        pass
+
+    def method_2(self):
+        pass
```
cjw296, line 13: @mariocj89 - when you're doing test coverage for [this](https://github.com/python/cpython/pull/16094/files#r1257929837), please can you tack `  #pragma: no-cover` on these two pass statements as per https://mock.readthedocs.io/en/latest/#backporting-rules?

Lib/unittest/mock.py
mariocj89: Good catch, this is indeed no longer used as we moved to `_mock_wait_timeout`

Lib/test/test_unittest/testmock/testthreadingmock.py
mariocj89: I don't see any other `#pragma: no-cover` in the codebase. Are you sure we should we add this to `cpython`?

Lib/unittest/mock.py
mariocj89: https://github.com/python/cpython/pull/106591

Lib/test/test_unittest/testmock/testthreadingmock.py
cjw296: I don't see any harm, and it makes for less chance of conflicts when backporting stuff in future.
(I _wish_ cpython actually checked code coverage, including in tests, but that's a whole other discussion :-/)

cjw296: https://github.com/python/cpython/pull/106617

gvanrossum:
There seems to be another problem with this -- I just got a test flake, apparently due to the use of short fixed timeouts in the test added here. See https://github.com/python/cpython/actions/runs/5559403783/jobs/10155484726.

cjw296:
Not sure what a "test flake" means in this context?

Looking at the test again, I'm unclear it's sensible:
https://github.com/python/cpython/blob/0db85eeba762e72f9f3c027e432cdebc627aac6c/Lib/test/test_unittest/testmock/testthreadingmock.py#L166-L180

Why the delays of 0.2, 0.5 and 0.6?
The assertions on lines 169, 175 and 176 seem likely but not guaranteed?

Lines 178-180 suppose that because `call(3)` has occurred, then `call(2)` will have occurred. Again, that seems likely given the 0.1s delay but not guaranteed, and indeed seems to have not happened here. What could lead these to be out of order? Some funny OS-level scheduler thing on that macOS runner? Something on @gvanrossum's branch?

FWIW, this is the first time I'm aware of this test failing, and it's run nightly on a bunch of different python versions in the backport CI: https://app.circleci.com/pipelines/github/testing-cabal/mock?branch=master
Those are all on Linux, mind...

@mariocj89 - what are your thoughts on this? What can we do to make this test reliable. A bonus would be to make them faster to run, these threading tests are noticeably much slower than the other mock tests...

mariocj89:
It is a hard balance to keep the tests fast but reliable.

I will try changing this to have short retrying timeouts to make that happen. Will send a PR on Monday if there is not rush as I don't have access to a PC this weekend

gvanrossum:
> Not sure what a "test flake" means in this context?

A flaky test is a test that usually passes but sometimes fails. Tests that rely on timeouts are often flaky (but there are other causes). Any large complex system has flaky tests. In this case, the test failed in CI on macOS only; Windows and Ubuntu were fine. Re-running the test on my machine locally it passes. That makes it flaky in my book.

Note that, as defense against flaky tests, our CI machinery re-runs a failed test once and if it passes the second time, it is accepted. In this case we were apparently unlucky with the timeouts twice in a row.

Thanks for having a second look at these tests!

mariocj89:
https://github.com/python/cpython/pull/106822 should prevent any future failure.

Kentzo:
Is this change going to be included in 3.12?

gvanrossum:
I haven’t checked the source but I see no evidence of a backport to the 3.12 branch, so presumably not.

cjw296:
@Kentzo - it doesn't look like `ThreadingMock` made it into https://github.com/python/cpython/blob/3.12/Lib/unittest/mock.py, so no need for this fix.

`ThreadingMock` is available in the the mock rolling backport if you want to use it before 3.13 is out:
https://mock.readthedocs.io/en/latest/changelog.html#id1
https://pypi.org/project/mock/5.1.0/

owillebo:
Is it possible to use ThreadingMock with auto spec feature?

i.e. let patch or create_autospec create a ThreadingMock instance instead of a MagicMock instance.

mariocj89:
Per [the mocks docs](https://docs.python.org/3/library/unittest.mock.html#the-mock-class):

You can specify an alternative class of [`Mock`](https://docs.python.org/3/library/unittest.mock.html#unittest.mock.Mock) using the `new_callable` argument to [`patch()`](https://docs.python.org/3/library/unittest.mock.html#unittest.mock.patch).

That should work in theory, give it a try!

owillebo:
> Per [the mocks docs](https://docs.python.org/3/library/unittest.mock.html#the-mock-class):
> 
> You can specify an alternative class of [`Mock`](https://docs.python.org/3/library/unittest.mock.html#unittest.mock.Mock) using the `new_callable` argument to [`patch()`](https://docs.python.org/3/library/unittest.mock.html#unittest.mock.patch).
> 
> That should work in theory, give it a try!

This is not possible with autospec. The #1 reason for using patch :-(.

`ValueError: Cannot use 'autospec' and 'new_callable' together`

PR #16101, opened 2019-09-13
bpo-38159: Clarify documentation of PyState_AddModule

encukou:
This was never intented to be called manually from PyInit_*.

Also, clarify PyState_RemoveModule return value.

https://bugs.python.org/issue38159

### DIFF
--- a/Doc/c-api/module.rst
+++ b/Doc/c-api/module.rst
@@ -487,9 +487,11 @@ since multiple such modules can be created from a single definition.
 
    Python calls ``PyState_AddModule`` automatically after importing a module,
    so it is unnecessary (but harmless) to call it from module initialization
-   code. An explicit call is needed only if the module's own init code subsequently calls ``PyState_FindModule``.
+   code. An explicit call is needed only if the module's own init code
+   subsequently calls ``PyState_FindModule``.
    The function is mainly intended for implementing alternative import
-   mechanisms (either by calling it directly, or by referring to its implementation for details of the required state updates).
+   mechanisms (either by calling it directly, or by referring to its
+   implementation for details of the required state updates).
 
    Return 0 on success or -1 on failure.
 
--- a/Doc/c-api/module.rst
+++ b/Doc/c-api/module.rst
@@ -487,9 +487,9 @@ since multiple such modules can be created from a single definition.
 
    Python calls ``PyState_AddModule`` automatically after importing a module,
    so it is unnecessary (but harmless) to call it from module initialization
-   code, unless that code itself needs the module to be registered.
+   code. An explicit call is needed only if the module's own init code subsequently calls ``PyState_FindModule``.
    The function is mainly intended for implementing alternative import
-   mechanisms.
+   mechanisms (either by calling it directly, or by referring to its implementation for details of the required state updates).
 
    Return 0 on success or -1 on failure.
 
--- a/Doc/c-api/module.rst
+++ b/Doc/c-api/module.rst
@@ -485,10 +485,17 @@ since multiple such modules can be created from a single definition.
 
    Only effective on modules created using single-phase initialization.
 
+   The function is intended for implementing alternative import mechanisms.
+   Python calls PyState_AddModule automatically when importing a module; it
+   is generally not necessary to call it at module initialization time.
+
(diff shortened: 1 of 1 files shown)


eduardo-elizondo:
Just to clarify "implementing an alternative import mechanism" meant someone changing `__import__` right? Otherwise, could you expand a bit more on what you meant by that? :)

Btw for context, these were added since the documentation currently specifies that `PyState_FindModule` needs to be paired up with `PyState_AddModule` to work. Since the import documentation does not offer a guarantee that your module will be attached to the interpreter state (through `PyState_FindModule`), it seemed like the right call to have that in the init functions.

As a matter of fact, in our system we do rely on this signal to correctly work with `PyState_FindModule`. Our import mechanism does not do it by default.

Maybe an alternative can be to update the import documentation that it should offer this guarantee? Thoughts?

encukou:
Changing `__import__`, creating module objects dynamically, things like that. I'm not 100% sure on this expected behavior myself, unfortunately. But it's currently redundant in the common case.

encukou:
I asked over at at capi-sig.

matrixise:
@encukou maybe you could add "DO-NOT-MERGE"?

encukou:
Note to reviewer: See the [capi-sig discussion](https://mail.python.org/archives/list/capi-sig@python.org/thread/Z3K57PM54Q6OO4ZZMTL5LLQ4TSC4F2J6/)

encukou:
cc @ncoghlan

eduardo-elizondo (approved):
LGTM with a small comment

Doc/c-api/module.rst
```diff
@@ -485,10 +485,19 @@ since multiple such modules can be created from a single definition.
 
    Only effective on modules created using single-phase initialization.
 
+   Python calls ``PyState_AddModule`` automatically after importing a module,
```
eduardo-elizondo, line 488: An alternate implementation might not necessarily call ``PyState_AddModule`` but still achieve the same end result. Perhaps just say that Python attaches the module to the interpreter state and should be accessible through ``PyState_FindModule``. How they achieve that should be up to the implementation

encukou: All ways I can find of wording this more precisely make it significantly harder to understand.

I think it's understood that alternate implementations are free to do an equivalent of calling `PyState_AddModule`, instead of actually calling it. They should still implement the `PyState_AddModule` semantics (unless they decide to differ, of course).

Another thing that's implied here is that Python only calls this on single-phase-init modules. And another is that it's idempotent.

Should we document all that? How much hand-holding do authors of alternative import mechanisms need? (This is a real question, which you are in a good position to answer.)

encukou: But my main issue is:

> All ways I can find of wording this more precisely make it significantly harder to understand.

If you can think of a better way, let me know! Otherwise I can merge this.

Doc/c-api/module.rst
```diff
@@ -485,10 +485,19 @@ since multiple such modules can be created from a single definition.
...
+   so it is unnecessary (but harmless) to call it from module initialization
+   code, unless that code itself needs the module to be registered.
+   The function is mainly intended for implementing alternative import
+   mechanisms.
```
ncoghlan:
```suggestion
   mechanisms (either by calling it directly, or by referring to its implementation for details of the required state updates).
```

Doc/c-api/module.rst
```diff
@@ -485,10 +485,19 @@ since multiple such modules can be created from a single definition.
 
    Only effective on modules created using single-phase initialization.
 
+   Python calls ``PyState_AddModule`` automatically after importing a module,
+   so it is unnecessary (but harmless) to call it from module initialization
+   code, unless that code itself needs the module to be registered.
```
ncoghlan:
```suggestion
   code. An explicit call is needed only if the module's own init code subsequently calls ``PyState_FindModule``.
```

ncoghlan (commented):
Added a couple of specific suggestions that I think improve precision without making it too much harder to read.

encukou:
Azure Pipelines didn't start, I'll try closing & reopening.

PR #10441, opened 2018-11-09
bpo-22393: Fix multiprocessing.Pool hangs if a worker process dies unexpectedly

oesteban:
This PR fixes [issue22393](https://bugs.python.org/issue22393).

Three new unittests have been added.

https://bugs.python.org/issue22393

PR #16103, opened 2019-09-13
gh-66587: Fix deadlock from pool worker death without communication

applio:
Adds tracking of which worker process in the pool takes which job from the queue.

When a worker process dies without communication, its task/job is also lost.  By tracking what job that worker took off the job queue as its task, upon detecting the death, the parent process can add an item to the result queue indicating the failure of that task/job.

In case of a future regression, the supplied test uses subprocess to constrain the test with a timeout to ensure an indefinite hang does not interfere with the running of tests.

https://bugs.python.org/issue22393

* Issue: gh-66587


pierreglaser:
This looks good to me, simply a few remarks:
- out of curiosity, are you aware of #10441, that implements a fix for this issue?
- what I like about #10441 is that the pool errors out as soon as the worker crashes, whereas in this PR, the pool waits until all tasks are complete before erroring.

Also pinging @tomMoral

zooba:
For mine, I think this fix seems more elegant than #10441, but the tests in that PR seem to have more coverage.

I personally prefer to just have the task fail, and the pool continue. The current behaviour is that the broken worker is immediately replaced and other work continues, but if you wait on the failed task then it will never complete. Now it does complete (with a failure), which means robust code can re-queue it if appropriate. I don't see any reason to tear down the entire pool.

Few comments on the PR incoming.

Lib/multiprocessing/pool.py
```diff
@@ -587,12 +597,18 @@ def _handle_results(outqueue, get, cache):
                 util.debug('result handler got sentinel')
                 break
 
-            job, i, obj = task
-            try:
-                cache[job]._set(i, obj)
-            except KeyError:
-                pass
-            task = job = obj = None
+            job, i, (task_info, value) = task
```
zooba, line 604: We know that `value` is always `pid` here, right? Can we just call it that? (I spent a minute or two trying to figure out whether `value` is hashable and eventually traced it back to the source...)

zooba: Similarly, `task_info` is really the result, yes? Or `task_is_complete`?

zooba: Obviously I'm not very familiar with this module, but given you've got such wonderfully descriptive names everywhere else in the PR, these ones stand out as a bit vague :)

Lib/multiprocessing/pool.py
```diff
@@ -294,8 +296,15 @@ def _join_exited_workers(pool):
             if worker.exitcode is not None:
                 # worker exited
                 util.debug('cleaning up worker %d' % i)
+                pid =  worker.ident
```
zooba:
```suggestion
                pid = worker.ident
```

Lib/multiprocessing/pool.py
```diff
@@ -294,8 +296,15 @@ def _join_exited_workers(pool):
...
                 worker.join()
                 cleaned = True
+                if pid in job_assignments:
```
zooba:
```suggestion
                job = job_assignments.pop(pid, None)
                if job:
                    outqueue.put((job, i, (False, RuntimeError("Worker died"))))
```

And some additional simplification below, of course.

Lib/multiprocessing/pool.py
```diff
@@ -294,8 +296,15 @@ def _join_exited_workers(pool):
...
+                    # If the worker process died without communicating back
+                    # while running a job, add a default result for it.
+                    job = job_assignments[pid]
+                    outqueue.put((job, i, (False, RuntimeError("Worker died"))))
```
zooba: Can we put the exit code into the error? And presumably this should at least be a `ProcessError` if not a new subclass (`WorkerExitedError`?).

zooba: Documenting the exception is probably sufficient documentation as well.

Lib/multiprocessing/pool.py
```diff
@@ -671,7 +691,8 @@ def join(self):
     def _help_stuff_finish(inqueue, task_handler, size):
         # task_handler may be blocked trying to put items on inqueue
         util.debug('removing tasks from inqueue until task handler finished')
-        inqueue._rlock.acquire()
+        if inqueue._reader.poll():
+            inqueue._rlock.acquire()
```
zooba, line 697: If the `poll()` returns False, that means we should exit this function, right? Given the next part used to be done under lock, it hardly seems right that now we can do it without the lock (but maybe it's all been cleaned up here?)

Lib/multiprocessing/pool.py
```diff
@@ -587,12 +597,18 @@ def _handle_results(outqueue, get, cache):
...
-            task = job = obj = None
+            job, i, (task_info, value) = task
+            if task_info is None:
+                # task_info is True or False when a task has completed but
+                # None indicates information about which process has
+                # accepted a job from the queue.
+                job_assignments[value] = job
+            else:
+                try:
+                    cache[job]._set(i, (task_info, value))
```
tomMoral, line 612: Why don't you remove the job from `job_assignement` here? It would avoid unecessary operation when a worker died gracefully.

Lib/multiprocessing/pool.py
tomMoral: It is necessary to get the `inqueue._rlock` even if `poll()` returns `False` as the `task_handler` might still concurrently be trying to put some work in the `inqueue`.

Lib/multiprocessing/pool.py
tomMoral: Why not directly set the value of the job instead of putting it in the `outqueue`? Using the `outqueue` will require using the `lock` + serialization while you are already in the main process. But to do that, it requires to clean the `job_assignment` when a job is complete.

tomMoral: Also, the `i` here is not correct. This might make the imap hang forever.

tomMoral (commented):
Here is a batch of comments.

I have to say that I like this solution as it is the most robust way of handling this, (a kind of scheduler). But it also comes with more complexity and increase communication needs -> more changes for deadlocks.

One of the main argument for the fail on error design is that there is no way there is no way to know in the main process if the worker that died had a lock on one of the communication queue. In this situation, the only way to recover the system and avoid a deadlock is to kill the `Pool` and re-spawn one.

Lib/multiprocessing/pool.py
applio: When `task_info` is `None`, then `value` will be the pid but if `task_info` is `True` then `value` will be the result of the task or if `False` then `value` will be an exception object.  I'm totally agreed that `value` is vague but I struggled to come up with a more appropriate name.  I do take some small consolation that `(task_info, value)` is at least a little more helpful than `obj`. :)

Lib/test/_test_multiprocessing.py
```diff
@@ -2712,6 +2712,27 @@ def errback(exc):
         p.close()
         p.join()
 
+    def test_pool_worker_died_without_communicating(self):
+        # Issue22393: test fix of indefinite hang caused by worker processes
+        # exiting abruptly (such as via os._exit()) without communicating
+        # back to the pool at all.
+        prog = (
```
taleinat, line 3343: This can be written much more clearly using a multi-line string. See for example a very similar case in `test_shared_memory_cleaned_after_process_termination` in this file.

Lib/test/_test_multiprocessing.py
```diff
@@ -2712,6 +2712,27 @@ def errback(exc):
...
+            "import os, multiprocessing as mp; "
+            "is_main = (__name__ == '__main__'); "
+            "p = mp.Pool(1) if is_main else print('worker'); "
+            "p.map(os._exit, [1]) if is_main else None; "
+            "(p.close() or p.join()) if is_main else None"
+        )
+        # Only if there is a regression will this ever trigger a
+        # subprocess.TimeoutExpired.
+        completed_process = subprocess.run(
+            [sys.executable, '-E', '-S', '-O', '-c', prog],
```
taleinat, line 3353: The '-O' flag probably shouldn't be used here, but '-S' and '-E' seem fine.

taleinat (commented):
Additional tests would certainly be a good idea.

Lib/test/_test_multiprocessing.py
taleinat: Also, consider calling `test.support.script_utils.interpreter_requires_environment()`, and only use the '-E' flag if that returns `False`, as done by the other Python script running utils in `test.support.script_utils`.

Or just use `test.support.script_utils.run_python_until_end()` instead of `subprocess.run()`.

csabella:
@applio, I'm not sure where this one is at, but I believe there are some comments that still need to be addressed.  I don't know if it's waiting on anything else, but it would probably be nice to get this merged.

ambv:
Closing and re-opening to re-trigger CI.

ambv:

Lib/multiprocessing/pool.py
```diff
@@ -297,8 +299,16 @@ def _join_exited_workers(pool):
             if worker.exitcode is not None:
                 # worker exited
                 util.debug('cleaning up worker %d' % i)
+                pid = worker.ident
                 worker.join()
                 cleaned = True
+                job_info = job_assignments.pop(pid, None)
```
pablogsal, line 305: This depends on `_handle_results()` recording the assignment before `_join_exited_workers()` runs. If the worker dies first, we miss the job and the result can still hang.

Lib/multiprocessing/pool.py
```diff
@@ -590,12 +601,18 @@ def _handle_results(outqueue, get, cache):
...
-                cache[job]._set(i, obj)
-            except KeyError:
-                pass
-            task = job = obj = None
+            job, i, (task_info, value) = task
+            if task_info is None:
+                # task_info is True or False when a task has completed but
+                # None indicates information about which process has
+                # accepted a job from the queue.
+                job_assignments[value] = (job, i)
```
pablogsal, line 609: This entry is never cleared when the worker sends the real result, so `maxtasksperchild` can report `Worker died` for a completed map chunk. Can we clear it when handling the result?

pablogsal (changes requested):
.

Lib/multiprocessing/pool.py
gpshead: i think there's still a race condition regardless.

gpshead:
FWIW I don't think this PR is workable to solve this.  shutting down the Pool as suggested in https://github.com/python/cpython/issues/66587#issuecomment-1093664340 like concurrent.futures does with BrokenProcessPool on unclean worker death would make more sense...

PR #16126, opened 2019-09-13
bpo-38165: Fix tempfile.py

evgeniysolntsev:
freezes when loading a program on imports because user rights are insufficient
python installed on c:/ and requires administrator rights
when run program from not root user

### DIFF
--- a/Lib/tempfile.py
+++ b/Lib/tempfile.py
@@ -242,26 +242,16 @@ def _mkstemp_inner(dir, pre, suf, flags, output_type):
     if output_type is bytes:
         names = map(_os.fsencode, names)
 
-    for seq in range(TMP_MAX):
-        name = next(names)
-        file = _os.path.join(dir, pre + name + suf)
-        _sys.audit("tempfile.mkstemp", file)
-        try:
-            fd = _os.open(file, flags, 0o600)
-        except FileExistsError:
-            continue    # try again
-        except PermissionError:
-            # This exception is thrown when a directory with the chosen name
-            # already exists on windows.
-            if (_os.name == 'nt' and _os.path.isdir(dir) and
-                _os.access(dir, _os.W_OK)):
-                continue
-            else:
-                raise
-        return (fd, _os.path.abspath(file))
+    name = next(names)
+    file = _os.path.join(dir, pre + name + suf)
+    try:
+        fd = _os.open(file, flags, 0o600)
+    except FileExistsError:
+        raise FileExistsError(_errno.EEXIST, "No usable temporary file name found")
+    except PermissionError:
+        raise PermissionError(_errno.EACCES, "No permission to create temp file")
 
-    raise FileExistsError(_errno.EEXIST,
-                          "No usable temporary file name found")
+    return (fd, _os.path.abspath(file))
 
 
 # User visible interfaces.


eduardo-elizondo (changes requested):
All the builds are failing. Please look at the failing builds and re-submit your change

aeros (commented):
Thanks for the PR @evgeniymischuk. Please adjust the title of the PR according the standard conventions. The format is: ``bpo-<issue_number>: <summary_of_PR>``. This allows the PR to be directly linked to the bpo issue. If you haven't already, I would highly recommend reading through the [devguide's quick reference page](https://devguide.python.org/#quick-reference). 

Travis CI is currently failing because of a error in ``test.test_tempfile.TestMkstempInner``, specifically for the tests ``test_collision_with_existing_directory`` and ``test_collision_with_existing_file ``. Both failures are occurring during the handling of ``FileNotExistsError``. Here's the relevant traceback, for debugging purposes:
image

I didn't include the other one since they both contained the same exact traceback. Essentially, when the ``FileExistsError`` is being raised, there is no suitable replacement name found for the temporary file created during the tests. I would recommend trying to address this issue first.

Before you update this PR with a new batch of commits, you should [run the tests locally](https://devguide.python.org/runtests/#running) to ensure everything is passing. The CI can still potentially fail if the tests are not passing in other environments, but this will help to rule out the majority of issues.

aeros167:
For now, I am adding a ``DO-NOT-MERGE`` label since there are issues to resolve with the PR. Once the problems have been addressed, either someone else can remove the label, or you can @mention me to take a look at the PR again.

Lib/tempfile.py
```diff
@@ -242,26 +242,16 @@ def _mkstemp_inner(dir, pre, suf, flags, output_type):
     if output_type is bytes:
         names = map(_os.fsencode, names)
 
-    for seq in range(TMP_MAX):
```
aeros, line 245: Why is this being removed? This block is supposed to continue trying to create the file until the maximum number of tries are reached, ``TMP_MAX``.

Lib/tempfile.py
```diff
@@ -242,26 +242,16 @@ def _mkstemp_inner(dir, pre, suf, flags, output_type):
...
-                continue
-            else:
-                raise
-        return (fd, _os.path.abspath(file))
+    name = next(names)
+    file = _os.path.join(dir, pre + name + suf)
+    try:
+        fd = _os.open(file, flags, 0o600)
+    except FileExistsError:
+        raise FileExistsError(_errno.EEXIST, "No usable temporary file name found")
```
aeros, line 250: When this exception occurs, this block is supposed try to *handle* the exception, not just raise it.

Lib/tempfile.py
```diff
@@ -242,26 +242,16 @@ def _mkstemp_inner(dir, pre, suf, flags, output_type):
...
+    except PermissionError:
+        raise PermissionError(_errno.EACCES, "No permission to create temp file")
```
aeros, line 252: See my [other comment](https://github.com/python/cpython/pull/16126/files#r324393420).

matrixise:
Hi @aeros167 

Thank you for your review and your time, but I think we can close this PR and the associated issues.

aeros167:
> Hi @aeros167
Thank you for your review and your time, but I think we can close this PR and the associated issues.

No problem, I think they could be closed as well.

PR #8934, opened 2018-08-25
bpo-34503: Fix refleak in PyErr_SetObject()

izbyshev:
https://bugs.python.org/issue34503

PR #16136, opened 2019-09-14
bpo-38168: Fix possible refleaks in setint() of mmapmodule.c

shihai1991:
https://bugs.python.org/issue38168

### DIFF
--- a/Modules/mmapmodule.c
+++ b/Modules/mmapmodule.c
@@ -1468,7 +1468,8 @@ static void
 setint(PyObject *d, const char *name, long value)
 {
     PyObject *o = PyLong_FromLong(value);
-    if (o && PyDict_SetItemString(d, name, o) == 0) {
+    if (o) {
+        PyDict_SetItemString(d, name, o);
         Py_DECREF(o);
     }
 }


matrixise:
Could you reproduce with a test? and via this command?
```
./python -m test test_XYZ -R 3:3
```
replace test_XYZ by a test where you test the memory leak.

shihai1991:
> Could you reproduce with a test? and via this command?
> 
> ```
> ./python -m test test_XYZ -R 3:3
> ```
> 
> replace test_XYZ by a test where you test the memory leak.

I have the same question too, but i found some PR or bpo no regression test against it :(
[1] https://github.com/python/cpython/pull/8934
[2] https://bugs.python.org/issue27248

animalize:
In current code, if `PyDict_SetItemString()` failes (return -1), `o` will not be released.
It's hard to reproduce a leak, it requires harsh condition.

Another way is to use `PyModule_AddIntMacro()`, and no longer use `setint()` in `Modules/mmapmodule.c`.
IMO `PyModule_AddIntMacro()`is more suitable.

shihai1991:
> In current code, if `PyDict_SetItemString()` failes (return -1), `o` will not be released.
> It's hard to reproduce a leak, it requires harsh condition.

Agree.
 
> Another way is to use `PyModule_AddIntMacro()`, and no longer use `setint()` in `Modules/mmapmodule.c`.
> IMO `PyModule_AddIntMacro()`is more suitable, but maybe a bit slower.

I review the PyModule_AddIntMacro code, looks your are right. But as your said, it MAYBE a bit slower.

animalize:
I would suggest use "possible" in the title, "Fix possible refleaks..."

shihai1991:
> I would suggest use "possible" in the title, "Fix possible refleaks..."

done :)

zhangyangyu (approved):
I think for this tiny and clear fix, which also hard to reproduce, it's okay to omit test.

PR #16099, opened 2019-09-13
bpo-38122: minor fixes to AsyncMock spec handling

voidspace:
Simplification of the spec handling for AsyncMock and correction for how spec_set is handled.

https://bugs.python.org/issue38122

PR #16137, opened 2019-09-14
[3.8] bpo-38122: minor fixes to AsyncMock spec handling (GH-16099).

matrixise:
(cherry picked from commit 14fd925a18fe3db0922a7d798e373102fe7a8a9c)

Co-authored-by: Michael Foord <voidspace@users.noreply.github.com>

https://bugs.python.org/issue38122

### DIFF
--- a/Lib/unittest/mock.py
+++ b/Lib/unittest/mock.py
@@ -1016,23 +1016,6 @@ class NonCallableMock(Base):
 _MOCK_SIG = inspect.signature(NonCallableMock.__init__)
 
 
-class _AnyComparer(list):
-    """A list which checks if it contains a call which may have an
-    argument of ANY, flipping the components of item and self from
-    their traditional locations so that ANY is guaranteed to be on
-    the left."""
-    def __contains__(self, item):
-        for _call in self:
-            if len(item) != len(_call):
-                continue
-            if all([
-                expected == actual
-                for expected, actual in zip(item, _call)
-            ]):
-                return True
-        return False
-
-
 def _try_iter(obj):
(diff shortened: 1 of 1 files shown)


matrixise:
Hi @voidspace 

There is problem with the #14700 because that one has not been backported to 3. and your code use `_AnyComparer`. I try to find the right solution with @pganssle

matrixise:
@voidspace `_AnyComparer` is specific to master, not backported to 3.8, so, I have removed that class from your PR and run the tests. wait & see.

lisroach (approved):
LGTM

tirkarthi:
Seems this missed the RC build :( It more looks like a code refactor. If I understand it correctly in previous case there was list comprehension over arguments starting with "spec" to take first argument and perhaps depending on the order they are passed. Now with PR it checks for `spec_set` and falls back to `spec` making the choice predictable to detect. I guess it could cause problem if someone depends on the order in 3.8 RC or 3.8.0 and upgrades to 3.8.1 as it's merged later. Also there is a removed comment about what to do when spec_set and spec are different so could it be tested to ensure this choice is explicit?

matrixise:
I worked on this issue in september but 3.8 has been released in october, do you think we could merge it into 3.8?

csabella:
@matrixise, should this change still be backported to 3.8?  Thanks!

matrixise:
@csabella for this PR, yep, I think we should backport it to 3.8.

ambv:
Sorry, closing this stale backport as 3.8 is closing to security fixes only mode.

PR #16139, opened 2019-09-14
bpo-38169: Increase code coverage for SharedMemory and ShareableList

vinay0410:
Increase Code Coverage, and add check in SharedMemory when user wants to create shm with size 0, to throw ValueError.

https://bugs.python.org/issue38169

### DIFF
--- a/Lib/test/_test_multiprocessing.py
+++ b/Lib/test/_test_multiprocessing.py
@@ -3705,6 +3705,18 @@ class _TestSharedMemory(BaseTestCase):
         self.assertGreaterEqual(sms.size, 512)
         self.assertGreaterEqual(len(sms.buf), sms.size)
 
+        # Verify __repr__
+        self.assertIn(sms.name, str(sms))
+        self.assertIn(str(sms.size), str(sms))
+
+        # Test pickling
+        sms.buf[0:6] = b'pickle'
+        pickled_sms = pickle.dumps(sms)
+        sms2 = pickle.loads(pickled_sms)
+        self.assertEqual(sms.name, sms2.name)
+        self.assertEqual(sms.size, sms2.size)
+        self.assertEqual(bytes(sms.buf[0:6]), bytes(sms2.buf[0:6]), b'pickle')
+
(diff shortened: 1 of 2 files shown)


Lib/multiprocessing/shared_memory.py
```diff
@@ -73,6 +73,8 @@ class SharedMemory:
     def __init__(self, name=None, create=False, size=0):
         if not size >= 0:
             raise ValueError("'size' must be a positive integer")
+        if create and size == 0:
```
corona10: IMHO, raising ValueError need to be discussed with core developers.

vinay0410: Actually, if `size == 0 and create is True`, then the shared memory segment is created but a `ValueError` is raised by `mmap`  [here](https://github.com/python/cpython/blob/0e2c856e74d8702d5903ff9f5a6fbbaa436b2bf3/Lib/multiprocessing/shared_memory.py#L113).
So, previously also, ValueError was being raised. But, shared memory segment was being created and it wasn't unlinked.

I could have added `except ValueError` at [L114](https://github.com/python/cpython/blob/0e2c856e74d8702d5903ff9f5a6fbbaa436b2bf3/Lib/multiprocessing/shared_memory.py#L113) with `except OSError` to call `self.unlink` and the raise.
But I think that the implemented approach in PR would be better, since this way there wouldn't be any hassle of linking and unlinking the shared memory, if invalid create size is being passed.

vinay0410:
Hi @applio , Could you please review this ?

vinay0410:
Hi @applio , Could you please review this.

vinay0410:
HI, @applio , Can you please review this ?

vinay0410:
@tiran , Could you please review this ?

corona10 (changes requested):
I am okay with all changes.
But please separate this PR into 2 PRs
- PR for Increasing code coverage
- Create a new issue for raising ValueError and create a new PR for this. (Please ping me this.)

vinay0410:

corona10 (approved):

PR #16149, opened 2019-09-14
bpo-37798: Prevent undefined behavior in direct calls to the C helper…

rhettinger:


corona10 (approved):
@rhettinger LGTM for the code.
2 questions for this PR.

1. Don't we have to add this logic for pure python version?
https://github.com/python/cpython/blob/ff603f6c3d3dc0e9ea8c1c51ce907c4821f42c54/Lib/statistics.py#L821

2. We don't need unit tests for _normal_dist_inv_cdf with error handling?

corona10:
@rhettinger 
When I think about it, the two cases above seem to be no problem
It's good to go :)

PR #16152, opened 2019-09-14
Doc: Add link to interpreter tutorial

adorilson:
Just a link to GNU Readline site.

### DIFF
--- a/Doc/tutorial/interpreter.rst
+++ b/Doc/tutorial/interpreter.rst
@@ -154,7 +154,7 @@ declaration should be added as the second line of the file.  For example::
 
    #!/usr/bin/env python3
    # -*- coding: cp1252 -*-
-   
+
 .. _GNU Readline: https://tiswww.case.edu/php/chet/readline/rltop.html
 .. rubric:: Footnotes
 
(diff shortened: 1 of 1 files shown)


aeros (changes requested):
Thanks for the PR @adorilson.

+1 on adding the link, I just have a few minor suggestions.

1) Grammar fix: "systems that support GNU Readline library" should be "systems that support the GNU Readline library".

2) Use a direct inline link instead of the footer link. The footer links are typically used for links that are likely to be used multiple times within the page, but this one is only being used once. A direct link can be added using:
```
`GNU Readline <https://tiswww.case.edu/php/chet/readline/rltop.html>`_
```

(can be split across multiple lines)

3)  Adjust the number of characters per line to fit within the maximum reST 80 character width. Currently, line 37 is at 81 characters and line 40 is at 82 characters. I would suggest adjusting it after making the other adjustments, so that it doesn't have to be done multiple times.

aeros167:
I would also recommend making an adjustment to the PR title, so that it's easier to tell where the changes are being made from the title alone. Something along the lines of "Doc: Add link to interpreter tutorial" should work.

aeros (approved):
Thanks, LGTM.

matrixise (approved):

PR #16156, opened 2019-09-15
bpo-38081: Add more non-fatal error codes for ntpath.realpath

zooba:
https://bugs.python.org/issue38081

### DIFF
--- a/Lib/ntpath.py
+++ b/Lib/ntpath.py
@@ -530,15 +530,28 @@ else:
         if seen is None:
             seen = set()
 
+        # These error codes indicate that we should stop reading links and
+        # return the path we currently have.
+        # 1: ERROR_INVALID_FUNCTION
+        # 2: ERROR_FILE_NOT_FOUND
+        # 3: ERROR_DIRECTORY_NOT_FOUND
+        # 5: ERROR_ACCESS_DENIED
+        # 21: ERROR_NOT_READY (implies drive with no media)
+        # 32: ERROR_SHARING_VIOLATION
+        # 50: ERROR_NOT_SUPPORTED (implies no file system)
+        # 67: ERROR_BAD_NET_NAME (implies remote server unavailable)
+        # 87: ERROR_INVALID_PARAMETER
+        # 4390: ERROR_NOT_A_REPARSE_POINT
+        # 4392: ERROR_INVALID_REPARSE_DATA
+        # 4393: ERROR_REPARSE_TAG_INVALID
+        allowed_winerror = 1, 2, 3, 5, 21, 32, 50, 67, 87, 4390, 4392, 4393
+
         while normcase(path) not in seen:
             seen.add(normcase(path))
             try:
                 path = _nt_readlink(path)
             except OSError as ex:
-                # Stop on incorrect function (1), file (2) or
-                # directory (3) not found, or paths that are
-                # not reparse points (4390)
-                if ex.winerror in (1, 2, 3, 4390):
+                if ex.winerror in allowed_winerror:
                     break
                 raise
             except ValueError:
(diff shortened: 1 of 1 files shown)


Lib/ntpath.py
```diff
@@ -554,9 +567,20 @@ def _getfinalpathname_nonstrict(path):
...
-        # invalid syntax (123), and symlinks that cannot be followed (1921)
-        allowed_winerror = 2, 3, 87, 123, 1921
+        # These error codes indicate that we should stop resolving the path
+        # and return the value we currently have.
+        # 1: ERROR_INVALID_FUNCTION
+        # 2: ERROR_FILE_NOT_FOUND
+        # 3: ERROR_DIRECTORY_NOT_FOUND
+        # 5: ERROR_ACCESS_DENIED
+        # 21: ERROR_NOT_READY (implies drive with no media)
+        # 32: ERROR_SHARING_VIOLATION
```
eryksun: Getting a sharing violation is highly unusual in this case, so I'd include a parenthetical note such as "(implies an NTFS paging file)". 

If I recall correctly, FAT32 fails the paging-file case with a more sensible `STATUS_ACCESS_DENIED` instead of `STATUS_SHARING_VIOLATION`. Normally, a sharing violation error can only occur if we request data access (i.e. read/execute, write/append, delete) on a volume device or a file-system file or directory and the desired access isn't shared by existing opens. We never request data access in these cases, so NTFS returning a sharing violation is unexpected. (Well, since it has behaved like this for over 25 years, it's not unexpected historically, just unexpected for a non-data `CreateFileW` call to fail with this error.)

Anyway, what we do in `os.stat` for `ERROR_ACCESS_DENIED` and `ERROR_SHARING_VIOLATION` is to query the parent directory for basic stat info via `FindFirstFileW`. Do you want to try to get the normalized filename from the parent directory in this case -- e.g. "PAGEFILE.SYS" -> "pagefile.sys", or "C:\Users\JOHNSM\~1" -> "C:\Users\JohnSmith"? We'd need a private function for this such as `nt._get_normalized_name`. In particular, we may have a short alias such as "JOHNSM\~1", which requires `FindFirstFileW` to map it to the real name "JohnSmith". Either name is 'real', but by default `GetFinalPathNameByHandleW` tries to get a normalized name, so we'd be more consistent with its behavior in general.

Lib/ntpath.py
```diff
@@ -554,9 +567,20 @@ def _getfinalpathname_nonstrict(path):
...
+        # 50: ERROR_NOT_SUPPORTED (implies no file system)
```
eryksun: Actually, for a device that's not a file system or not mounted by a file system, the error is typically either `ERROR_INVALID_FUNCTION` (e.g. from `STATUS_INVALID_DEVICE_REQUEST` for the file-system control `NtFsControlFile`: `FSCTL_GET_REPARSE_POINT`) or `ERROR_INVALID_PARAMETER` (e.g. from `STATUS_INVALID_INFO_CLASS` for the file-system query  `NtQueryInformationFile`: `FileNameInformation`). 

Getting `ERROR_NOT_SUPPORTED` (from `STATUS_NOT_SUPPORTED`) is usually for a specific request that's not supported, as opposed to a class of request that's not supported. For example, the named-pipe file system doesn't support `FSCTL_GET_REPARSE_POINT`.  A driver may also return `STATUS_NOT_IMPLEMENTED` in this case, but that maps to `ERROR_INVALID_FUNCTION`.

Lib/ntpath.py
zooba: I'll update the comment now, but I think we can leave the additional check for a separate bug.

PR #16157, opened 2019-09-15
bpo-27218 : improve tracing performance with f_trace set to Py_None

nanjekyejoannah:
I am not adding a new test as I think this test :

```
def test_set_and_retrieve_none(self):
        sys.settrace(None)
        assert sys.gettrace() is None

```
Works for this case too. No ?

https://bugs.python.org/issue27218


nanjekyejoannah:
Armin had a comment on pdb that seems to make sense. I close this until we have looked at why pdb fails.

PR #16159, opened 2019-09-15
bpo-38178: Don't explicitly pass "loop" to EchoClientProtocol.

hniksic:
https://bugs.python.org/issue38178


asvetlov (approved):
The change is correct and reflects the current best practice.

PR #16163, opened 2019-09-15
Doc: remove duplicate word in controlflow tutorial

adorilson:
Just remove duplicate word in sentence:

> use positional-only to prevent prevent breaking API changes

### DIFF
--- a/Doc/tutorial/controlflow.rst
+++ b/Doc/tutorial/controlflow.rst
@@ -695,7 +695,7 @@ As guidance:
 * Use keyword-only when names have meaning and the function definition is
   more understandable by being explicit with names or you want to prevent
   users relying on the position of the argument being passed.
-* For an API, use positional-only to prevent prevent breaking API changes
+* For an API, use positional-only to prevent breaking API changes
   if the parameter's name is modified in the future.
 
 .. _tut-arbitraryargs:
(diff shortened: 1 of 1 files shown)


JulienPalard (changes requested):
Hi! Thanks for improving the documentation!

In this commit you rewrapped a lot of paragraphs without changing them, which is bad for the log readability. (It mess with git blame, it's hard to review as it's long to find the actual modification, ...).

Can you please rollback those changes and do a *minimal* commit with just the removal of the duplicate word? I bet the paragraph containing the removed word don't needs rewrapping neither after the removal.

You can simply add a new commit to your branch and push it, as anyway both commits will be squashed while being merged: you don't have to force push, just add a new commit.

JulienPalard:
As you did it in two separate commit (good) you can simply run `git revert 89dbf912798b8440d67f4ab8184c85bafa29dd2b` (and `git push`).

adorilson:

But, I made a revert wrong. Then I reverted the revert. :/

If you prefer, I can redo the PR from scratch. My bad.

JulienPalard (approved):

JulienPalard:
@adorilson thanks! The series of revert if not a problem as we squash before merging, see: https://github.com/python/cpython/commit/b7af4e75657e6478920d00260c48a1a3020002fc it's clean.

PR #16166, opened 2019-09-15
bpo-37828: Fix default mock_name in unittest.mock.assert_called error

categulario:
I used parenthesis to change precedence so the correct message is displayed. This is my very fist contribution to python (:

https://bugs.python.org/issue37828

### DIFF
--- a/Lib/unittest/test/testmock/testmock.py
+++ b/Lib/unittest/test/testmock/testmock.py
@@ -397,9 +397,10 @@ class MockTest(unittest.TestCase):
 
 
     def test_assert_called_exception_message(self):
-        with self.assertRaisesRegex(AssertionError, "Expected 'mock' to have been called"):
+        msg = "Expected '{0}' to have been called"
+        with self.assertRaisesRegex(AssertionError, msg.format('mock')):
             Mock().assert_called()
-        with self.assertRaisesRegex(AssertionError, "Expected 'test_name' to have been called"):
+        with self.assertRaisesRegex(AssertionError, msg.format('test_name')):
             Mock(name="test_name").assert_called()
 
 
--- a/Lib/unittest/test/testmock/testmock.py
+++ b/Lib/unittest/test/testmock/testmock.py
@@ -396,6 +396,13 @@ class MockTest(unittest.TestCase):
         _check(mock)
 
 
+    def test_assert_called_exception_message():
+        with self.assertRaisesRegexp(AssertionError, "Expected 'mock' to have been called"):
+            Mock().assert_called()
+        with self.assertRaisesRegexp(AssertionError, "Expected 'test_name' to have been called"):
+            Mock(name="test_name").assert_called()
+
+
     def test_assert_called_once_with(self):
         mock = Mock()
         mock()
--- a/Lib/unittest/test/testmock/testmock.py
+++ b/Lib/unittest/test/testmock/testmock.py
@@ -396,10 +396,10 @@ class MockTest(unittest.TestCase):
         _check(mock)
 
 
-    def test_assert_called_exception_message():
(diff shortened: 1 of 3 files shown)


the-knights-who-say-ni:

## CLA Missing

@categulario

mariocj89 (commented):
Quite a nice fix @categulario!

Would be great if you can add a test to make sure this does not happen again. You can do so by adding a test like

```python
def test_assert_called_exception_message():
  with self.assertRaisesRegexp(AssertionError, "Expected 'mock' to have been called"):
      Mock().assert_called()
  with self.assertRaisesRegexp(AssertionError, "Expected 'test_name' to have been called"):
      Mock(name="test_name").assert_called()
```

Also don't forget to go through [the process for the CLA](https://github.com/python/cpython/pull/16166#issuecomment-531597290) so the contribution can be accepted.

You should also include a news entry, see [here](https://blurb-it.herokuapp.com/).

This patch should be probably backported as well (for a core dev to mark as such).

categulario:
Thanks for the review @mariocj89!

* I added the test in a follow up commit
* I signed it the CLA like an hour ago, I'm told it takes a day to settle
* I just did the blurb thing
* _It works in my computer_ haha

Misc/NEWS.d/next/Library/2019-09-15-21-31-18.bpo-37828.gLLDX7.rst
```diff
@@ -0,0 +1 @@
+Properly report the mock name in :meth:`unittest.mock.assert_called` error
```
mariocj89: I think this needs to be ``` :meth:`unittest.mock.Mock.assert_called` ```, as the method belongs to the Mock class.

You can test this out running `make html` in the `Docs` folder and seeing the output.

mariocj89: As a nitpick, you might want to be more precise on what is fixed, so people can quickly see the fix in the "what's new" section.

Something like:
```
Fix incorrect mock name being generated in :meth:`unittest.mock.assert_called` exceptions
for mocks without a name. Patch by Abraham Toriz Cruz.
```

mariocj89 (commented):
I've made some suggestions over the *what's new* but the code looks good to me!

Lib/unittest/test/testmock/testmock.py
```diff
@@ -396,6 +396,13 @@ def _check(mock):
         _check(mock)
 
 
+    def test_assert_called_exception_message(self):
+        with self.assertRaisesRegex(AssertionError, "Expected 'mock' to have been called"):
```
tirkarthi: Nit : Please follow PEP 8 guidelines to ensure the line length is not more than 80 characters here and in below. You can probably have it as a format string with `message` and fill in the mock name like below.

```python
message = "Expected '{0}' to have been called"

with self.assertRaisesRegex(AssertionError, message.format('mock')):
    Mock().assert_called()
```

categulario: Thanks! What is the preferred way to run pep8 agains the modified code?

mariocj89 (approved):
Thanks @categulario !

Lib/unittest/test/testmock/testmock.py
tirkarthi: I only do manual checking. There are some parts of the code that predates PEP8 and PEP8 is enforced only for new code so using pep8 libraries on whole file might give some false information.

tirkarthi (approved):
LGTM. Thanks :)

mariocj89:
ping @pablogsal

pablogsal (approved):
LGTM

Thanks @tirkarthi and @mariocj89  for the reviews!

PR #15761, opened 2019-09-09
bpo-37383: Updates docs to reflect AsyncMock call_count after await.

lisroach:
Updates documentation to make it more clear that call_count is not updated _unless_ the AsyncMock has been awaited.

https://bugs.python.org/issue37383

PR #16192, opened 2019-09-16
bpo-38136: Updates await_count and call_count to be different things

lisroach:
After a lot of discussion we have decided that "calls" and "awaits" should be counted as two different things. 

To accomplish this I separated the call counting from the actual execution of the call, that way they are counted at two different times. For synchronous calls the increment and execute calls happen one immediately after another, but for asynchronous calls the increment count for call count happens first, and then only after `await` is called does the call execute and increment the await counters.

https://bugs.python.org/issue38136

### DIFF
--- a/Lib/unittest/test/testmock/testasync.py
+++ b/Lib/unittest/test/testmock/testasync.py
@@ -3,7 +3,7 @@ import inspect
 import unittest
 
 from unittest.mock import (call, AsyncMock, patch, MagicMock, create_autospec,
-                           _AwaitEvent)
+                           _AwaitEvent, sentinel)
 
 
 def tearDownModule():
@@ -524,11 +524,158 @@ class AsyncMockAssert(unittest.TestCase):
     def setUp(self):
         self.mock = AsyncMock()
 
-    async def _runnable_test(self, *args):
-        if not args:
+    async def _runnable_test(self, *args, **kwargs):
+        if not args and not kwargs:
             await self.mock()
         else:
-            await self.mock(*args)
+            await self.mock(*args, **kwargs)
+
+    async def _await_coroutine(self, coroutine):
+        return await coroutine
+
+    def test_assert_called_but_not_awaited(self):
+        mock = AsyncMock(AsyncClass)
+        with self.assertWarns(RuntimeWarning):
+            # Will raise a warning because never awaited
+            mock.async_method()
+        self.assertTrue(asyncio.iscoroutinefunction(mock.async_method))
+        mock.async_method.assert_called()
+        mock.async_method.assert_called_once()
+        mock.async_method.assert_called_once_with()
+        with self.assertRaises(AssertionError):
+            mock.assert_awaited()
+
+    def test_assert_called_then_awaited(self):
+        mock = AsyncMock(AsyncClass)
+        mock_coroutine = mock.async_method()
+        mock.async_method.assert_called()
+        mock.async_method.assert_called_once()
+        mock.async_method.assert_called_once_with()
+        with self.assertRaises(AssertionError):
+            mock.async_method.assert_awaited()
+
+        asyncio.run(self._await_coroutine(mock_coroutine))
+        # Assert we haven't re-called the function
+        mock.async_method.assert_called_once()
+        mock.async_method.assert_awaited_once()
+        mock.async_method.assert_awaited_once_with()
+
+    def test_assert_called_and_awaited_at_same_time(self):
+        with self.assertRaises(AssertionError):
+            self.mock.assert_awaited()
+
+        with self.assertRaises(AssertionError):
+            self.mock.assert_called()
+
+        asyncio.run(self._runnable_test())
+        self.mock.assert_called_once()
+        self.mock.assert_awaited_once()
+
+    def test_assert_called_twice_and_awaited_once(self):
+        mock = AsyncMock(AsyncClass)
+        coroutine = mock.async_method()
+        with self.assertWarns(RuntimeWarning):
+            # Will raise a warning because never awaited
+            mock.async_method()
+        mock.async_method.assert_called()
+        asyncio.run(self._await_coroutine(coroutine))
+        mock.async_method.assert_awaited_once()
+
+    def test_assert_called_once_and_awaited_twice(self):
+        mock = AsyncMock(AsyncClass)
+        coroutine = mock.async_method()
+        mock.async_method.assert_called_once()
+        asyncio.run(self._await_coroutine(coroutine))
+        with self.assertRaises(RuntimeError):
+            # Cannot reuse already awaited coroutine
+            asyncio.run(self._await_coroutine(coroutine))
+        mock.async_method.assert_awaited()
+
+    def test_assert_awaited_but_not_called(self):
+        with self.assertRaises(AssertionError):
+            self.mock.assert_awaited()
+        with self.assertRaises(AssertionError):
+            self.mock.assert_called()
+        with self.assertRaises(TypeError):
+            # You cannot await an AsyncMock, it must be a coroutine
+            asyncio.run(self._await_coroutine(self.mock))
+
+        with self.assertRaises(AssertionError):
+            self.mock.assert_awaited()
+        with self.assertRaises(AssertionError):
+            self.mock.assert_called()
+
+    def test_assert_has_calls_not_awaits(self):
+        kalls = [call('NormalFoo')]
+        with self.assertWarns(RuntimeWarning):
+            self.mock('NormalFoo')
+        self.mock.assert_has_calls(kalls)
+        with self.assertRaises(AssertionError):
+            self.mock.assert_has_awaits(kalls)
+
+    def test_assert_has_mock_calls_on_async_mock(self):
+        mock = AsyncMock()
+        with self.assertWarns(RuntimeWarning):
+            mock()
+        kalls_empty = [('', (), {})]
+        self.assertEqual(mock.mock_calls, kalls_empty)
+
+        kalls_args = ([call('NormalFoo'), call('baz')])
+        with self.assertWarns(RuntimeWarning):
+            self.mock('NormalFoo')
+            self.mock('baz')
+        self.assertEqual(self.mock.mock_calls, kalls_args)
+
+        a_mock = AsyncMock(AsyncClass)
+        with self.assertWarns(RuntimeWarning):
+            a_mock.async_method()
+        self.assertEqual(a_mock.async_method.mock_calls, kalls_empty)
+        self.assertEqual(a_mock.mock_calls, [call.async_method()])
+
+        kalls_args_kwargs = [call(), call(1, 2, 3, a=4, b=5)]
+        with self.assertWarns(RuntimeWarning):
+            a_mock.async_method(1, 2, 3, a=4, b=5)
+        self.assertEqual(a_mock.async_method.mock_calls, kalls_args_kwargs)
+
+    def test_async_method_calls_recorded(self):
+        with self.assertWarns(RuntimeWarning):
+            self.mock.something(3, fish=None)
+            self.mock.something_else.something(6, cake=sentinel.Cake)
+
+        self.assertEqual(self.mock.something_else.method_calls,
+                        [("something", (6,), {'cake': sentinel.Cake})],
+                        "method calls not recorded correctly")
+        self.assertEqual(self.mock.method_calls, [
+            ("something", (3,), {'fish': None}),
+            ("something_else.something", (6,), {'cake': sentinel.Cake})
+        ],
+            "method calls not recorded correctly")
+
+    def test_async_arg_lists(self):
+        def assert_attrs(mock):
+            names = 'call_args_list', 'method_calls', 'mock_calls'
+            for name in names:
+                attr = getattr(mock, name)
+                self.assertIsInstance(attr, _CallList)
+                self.assertIsInstance(attr, list)
+                self.assertEqual(attr, [])
+
+            assert_attrs(self.mock)
+
+            with self.assertWarns(RuntimeWarning):
+                self.mock()
+                self.mock(1, 2)
+                self.mock(a=3)
+
+                self.mock.reset_mock()
+                assert_attrs(self.mock)
+
+                a_mock = AsyncMock(AsyncClass)
+                a_mock.async_method()
+                a_mock.async_method(1, a=3)
+
+                a_mock.reset_mock()
+                assert_attrs(a_mock)
 
     def test_assert_awaited(self):
         with self.assertRaises(AssertionError):
--- a/Lib/unittest/test/testmock/testasync.py
+++ b/Lib/unittest/test/testmock/testasync.py
@@ -633,8 +633,8 @@ class AsyncMockAssert(unittest.TestCase):
             # Will raise a warning because never awaited
             self.mock('foo')
             self.mock('baz')
-        kalls_args = ([call(), call('foo'), call('baz')])
-        self.assertEqual(self.mock.mock_calls, kalls_args)
+        mock_kalls = ([call(), call('foo'), call('baz')])
+        self.assertEqual(self.mock.mock_calls, mock_kalls)
 
     def test_assert_has_mock_calls_on_async_mock_with_spec(self):
         a_class_mock = AsyncMock(AsyncClass)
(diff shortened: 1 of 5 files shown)


Lib/unittest/test/testmock/testasync.py
```diff
@@ -3,7 +3,7 @@
 import unittest
 
 from unittest.mock import (ANY, call, AsyncMock, patch, MagicMock,
-                           create_autospec, _AwaitEvent)
+                            create_autospec, _AwaitEvent, sentinel)
```
matrixise: why the space?

tirkarthi:
Does this mean we have to revert the doc updated added in https://github.com/python/cpython/pull/15761 ?

Lib/unittest/test/testmock/testasync.py
lisroach: Oh, I did a merge in the UI, didn't notice the space

lisroach:
@tirkarthi yes, I can actually just remove that doc change with this PR.

Lib/unittest/test/testmock/testmock.py
```diff
@@ -2017,7 +2018,7 @@ def trace(frame, event, arg):  # pragma: no cover
             )
 
             mocks = [
-                Mock, MagicMock, NonCallableMock, NonCallableMagicMock
+                Mock, MagicMock, NonCallableMock, NonCallableMagicMock, AsyncMock
```
tirkarthi, line 2021: I have been thinking to add this. Thanks for doing it. It also needs to be added for `EventMock` in the other PR just to make sure there is no regression.

Lib/unittest/test/testmock/testasync.py
```diff
@@ -528,11 +528,158 @@ class AsyncMockAssert(unittest.TestCase):
...
         self.mock = AsyncMock()
 
-    async def _runnable_test(self, *args):
-        if not args:
+    async def _runnable_test(self, *args, **kwargs):
+        if not args and not kwargs:
             await self.mock()
         else:
-            await self.mock(*args)
+            await self.mock(*args, **kwargs)
```
ezio-melotti: Unless I'm missing something, I believe you can remove the `if`/`else` and just `await self.mock(*args, **kwargs)` here.

Lib/unittest/test/testmock/testasync.py
```diff
@@ -528,11 +528,158 @@ class AsyncMockAssert(unittest.TestCase):
...
+            mock.assert_awaited()
+
+    def test_assert_called_then_awaited(self):
+        mock = AsyncMock(AsyncClass)
+        mock_coroutine = mock.async_method()
+        mock.async_method.assert_called()
+        mock.async_method.assert_called_once()
+        mock.async_method.assert_called_once_with()
+        with self.assertRaises(AssertionError):
+            mock.async_method.assert_awaited()
```
ezio-melotti, line 558: IIUC this will be `False` for both `mock.async_method.assert_awaited()` and `mock.assert_awaited()`, but in this test you check for `mock.async_method.assert_awaited()` and in the previous for `mock.assert_awaited()`
I think it would be better to make sure both are correct, and make the rest consistent (i.e. either always check both everywhere, or check that the values are propagated correctly in a specific test and then just check `mock.async_method.assert_awaited()` in the other tests).

Lib/unittest/test/testmock/testasync.py
```diff
@@ -528,11 +528,158 @@ class AsyncMockAssert(unittest.TestCase):
...
+
+        asyncio.run(self._await_coroutine(mock_coroutine))
+        # Assert we haven't re-called the function
+        mock.async_method.assert_called_once()
+        mock.async_method.assert_awaited_once()
+        mock.async_method.assert_awaited_once_with()
```
ezio-melotti, line 565: Shouldn't this also check for `mock.async_method.assert_awaited()` here?

Lib/unittest/test/testmock/testasync.py
```diff
@@ -528,11 +528,158 @@ class AsyncMockAssert(unittest.TestCase):
...
+        asyncio.run(self._runnable_test())
+        self.mock.assert_called_once()
+        self.mock.assert_awaited_once()
+
+    def test_assert_called_twice_and_awaited_once(self):
+        mock = AsyncMock(AsyncClass)
+        coroutine = mock.async_method()
+        with self.assertWarns(RuntimeWarning):
+            # Will raise a warning because never awaited
+            mock.async_method()
```
ezio-melotti, line 584: Why doesn't the first call raise a warning? (maybe add a comment that answers this question)

Lib/unittest/test/testmock/testasync.py
```diff
@@ -528,11 +528,158 @@ class AsyncMockAssert(unittest.TestCase):
...
+        mock.async_method.assert_called()
+        asyncio.run(self._await_coroutine(coroutine))
```
ezio-melotti, line 588: Maybe check that it's still not awaited after the two calls, before awaiting it.

Lib/unittest/test/testmock/testasync.py
```diff
@@ -528,11 +528,158 @@ class AsyncMockAssert(unittest.TestCase):
...
+            # You cannot await an AsyncMock, it must be a coroutine
+            asyncio.run(self._await_coroutine(self.mock))
+
+        with self.assertRaises(AssertionError):
+            self.mock.assert_awaited()
+        with self.assertRaises(AssertionError):
+            self.mock.assert_called()
+
+    def test_assert_has_calls_not_awaits(self):
+        kalls = [call('NormalFoo')]
```
ezio-melotti: Why the k?  Also why `'NormalFoo'`?

Lib/unittest/test/testmock/testasync.py
```diff
@@ -528,11 +528,158 @@ class AsyncMockAssert(unittest.TestCase):
...
+        with self.assertWarns(RuntimeWarning):
+            self.mock('NormalFoo')
```
ezio-melotti: Why is this `RuntimeWarning` raised?

Lib/unittest/test/testmock/testasync.py
```diff
@@ -528,11 +528,158 @@ class AsyncMockAssert(unittest.TestCase):
...
+            self.mock.assert_has_awaits(kalls)
+
+    def test_assert_has_mock_calls_on_async_mock(self):
+        mock = AsyncMock()
+        with self.assertWarns(RuntimeWarning):
+            mock()
+        kalls_empty = [('', (), {})]
+        self.assertEqual(mock.mock_calls, kalls_empty)
+
+        kalls_args = ([call('NormalFoo'), call('baz')])
```
ezio-melotti: Why not just `foo`?

Lib/unittest/test/testmock/testasync.py
```diff
@@ -528,11 +528,158 @@ class AsyncMockAssert(unittest.TestCase):
...
+        with self.assertWarns(RuntimeWarning):
+            self.mock('NormalFoo')
+            self.mock('baz')
+        self.assertEqual(self.mock.mock_calls, kalls_args)
```
ezio-melotti: IIUC in this test you are first getting the `mock`, then calling it to get the coro (and that raises a `RuntimeWarning`), then you test that the mock has been called with no args.  Then you call it two more times with two args (and that raises a warning again) and check that it's been called twice with two different args.

If this is correct I would add a comment about why the error is raised and move the definition of `kalls_args` after the `with` block.

Lib/unittest/test/testmock/testasync.py
```diff
@@ -528,11 +528,158 @@ class AsyncMockAssert(unittest.TestCase):
...
+        mock = AsyncMock()
+        with self.assertWarns(RuntimeWarning):
+            mock()
+        kalls_empty = [('', (), {})]
+        self.assertEqual(mock.mock_calls, kalls_empty)
+
+        kalls_args = ([call('NormalFoo'), call('baz')])
+        with self.assertWarns(RuntimeWarning):
+            self.mock('NormalFoo')
+            self.mock('baz')
```
ezio-melotti, line 635: Why are you using `mock()` in the first part and `self.mock()` here, but later you use `a_mock()` and let all calls accumulate?

Lib/unittest/test/testmock/testasync.py
```diff
@@ -528,11 +528,158 @@ class AsyncMockAssert(unittest.TestCase):
...
+        a_mock = AsyncMock(AsyncClass)
+        with self.assertWarns(RuntimeWarning):
+            a_mock.async_method()
+        self.assertEqual(a_mock.async_method.mock_calls, kalls_empty)
+        self.assertEqual(a_mock.mock_calls, [call.async_method()])
+
+        kalls_args_kwargs = [call(), call(1, 2, 3, a=4, b=5)]
+        with self.assertWarns(RuntimeWarning):
+            a_mock.async_method(1, 2, 3, a=4, b=5)
+        self.assertEqual(a_mock.async_method.mock_calls, kalls_args_kwargs)
```
ezio-melotti: IIUC in this test you are first getting the `mock`, then calling it to get the coro (and that raises a `RuntimeWarning`), then you test that the mock has been called with no args.  Then you call it two more times with two args (and that raises a warning again) and check that it's been called twice with two different args.  Then you do it again with an `async_method` with an empty call and a call with args.

If this is correct I would:
* add a comment about why the error is raised before the first `with`
* add a comment to separate the part that checks calling the mock to the part about calling the `async_method`
* move the definition of `kalls_args` and `kalls_args_kwargs` after the `with` block like you did in the first part.
* maybe split this test into two, since it's getting confused with `mock`, `self.mock` and `a_mock`
* at the end check that the kalls propagated to `a_mock.mock_calls` too?

Lib/unittest/test/testmock/testasync.py
```diff
@@ -528,11 +528,158 @@ class AsyncMockAssert(unittest.TestCase):
...
+
+    def test_async_method_calls_recorded(self):
+        with self.assertWarns(RuntimeWarning):
+            self.mock.something(3, fish=None)
+            self.mock.something_else.something(6, cake=sentinel.Cake)
+
+        self.assertEqual(self.mock.something_else.method_calls,
+                        [("something", (6,), {'cake': sentinel.Cake})],
+                        "method calls not recorded correctly")
```
ezio-melotti: Last two line should be indented one space to the right, aligned under the `self`.

Lib/unittest/test/testmock/testasync.py
```diff
@@ -528,11 +528,158 @@ class AsyncMockAssert(unittest.TestCase):
...
+        self.assertEqual(self.mock.method_calls, [
+            ("something", (3,), {'fish': None}),
+            ("something_else.something", (6,), {'cake': sentinel.Cake})
+        ],
+            "method calls not recorded correctly")
```
ezio-melotti, line 666: Why you call `something` then `something_else.something`, but the tests check `something_else` first and `something`+`something_else_something` afterwards?

Lib/unittest/test/testmock/testasync.py
```diff
@@ -528,11 +528,158 @@ class AsyncMockAssert(unittest.TestCase):
...
+
+    def test_async_arg_lists(self):
+        def assert_attrs(mock):
+            names = 'call_args_list', 'method_calls', 'mock_calls'
```
ezio-melotti: I would add `()` here.

Lib/unittest/test/testmock/testasync.py
```diff
@@ -528,11 +528,158 @@ class AsyncMockAssert(unittest.TestCase):
...
+
+                self.mock.reset_mock()
+                assert_attrs(self.mock)
+
+                a_mock = AsyncMock(AsyncClass)
+                a_mock.async_method()
+                a_mock.async_method(1, a=3)
+
+                a_mock.reset_mock()
+                assert_attrs(a_mock)
```
ezio-melotti: I think the second half should be dedented.
AFAICS, this is just a test that defines a function that is never called.

Misc/NEWS.d/next/Library/2019-09-16-09-54-42.bpo-38136.MdI-Zb.rst
```diff
@@ -0,0 +1,3 @@
+Changes AsyncMock call count and await count to be two different counters.
```
ezio-melotti, line 1: What about docs updates?

ezio-melotti (changes requested):

Lib/unittest/test/testmock/testasync.py
lisroach: Good catch, it should be `mock.async_method.assert_awaited()` in the previous test. I can check both in the previous test, I think it might be overkill to test both in every single test.

Lib/unittest/test/testmock/testasync.py
lisroach: Following the convention from `testmock.py` things like "klass" and "kalls" are used. I like it because it makes it a little clearer that these are test objects and not the same as the "call" objects from `mock`. 

No specific reason for `NormalFoo`, it used to be the name of NormalClass and just stuck around. Would just `foo` be better?

Lib/unittest/test/testmock/testasync.py
lisroach: Do you mean just swap the order of the asserts?

Lib/unittest/test/testmock/testasync.py
lisroach: Oh, yeah ignore my earlier comment about whether I should just change this to `foo` :)

Lib/unittest/test/testmock/testasync.py
```diff
@@ -610,76 +614,92 @@ def test_assert_awaited_but_not_called(self):
...
+            # Will raise a warning because never awaited
+            a_class_mock.async_method(1, 2, 3, a=4, b=5)
+        kalls_args_kwargs = [call(), call(1, 2, 3, a=4, b=5)]
+        self.assertEqual(a_class_mock.async_method.mock_calls, kalls_args_kwargs)
+        self.assertEqual(
+            a_class_mock.mock_calls,
+            [
+                call.async_method(),
+                call.async_method(1, 2, 3, a=4, b=5)
+            ]
```
ezio-melotti: I think it's better to do `method_calls = [call(), call(1, 2, 3, a=4, b=5)]` and `mock_kalls = [call.async_method(), call.async_method(1, 2, 3, a=4, b=5)]` and then use these two vars in the `assertEqual`s to save a few lines and maintain consistency

Lib/unittest/test/testmock/testasync.py
ezio-melotti: Yes, it looked backwards and led me to wonder if there was something going on that required checking `something_else` before `something`.

Lib/unittest/test/testmock/testasync.py
```diff
@@ -610,76 +614,92 @@ def test_assert_awaited_but_not_called(self):
...
+        )
 
     def test_async_method_calls_recorded(self):
         with self.assertWarns(RuntimeWarning):
+            # Will raise a warning because never awaited
```
ezio-melotti: Nitpick: `will raise *warnings* because never awaited, since here there are two calls, and presumably two warnings raised.

The same apply to the same comment below.

Lib/unittest/test/testmock/testasync.py
```diff
@@ -610,76 +614,92 @@ def test_assert_awaited_but_not_called(self):
...
 
-            assert_attrs(self.mock)
-
-            with self.assertWarns(RuntimeWarning):
-                self.mock()
-                self.mock(1, 2)
-                self.mock(a=3)
+        assert_attrs(self.mock)
+        with self.assertWarns(RuntimeWarning):
+            # Will raise a warning because never awaited
```
ezio-melotti: `s/a warning/warnings/`

Misc/NEWS.d/next/Library/2019-09-16-09-54-42.bpo-38136.MdI-Zb.rst
ezio-melotti: This comment was poorly placed.
I didn't mean to ask that the doc updates should be included in the news entry -- I was asking why I didn't see any changes in the documentation to reflect the changes you did in the code.

ezio-melotti (changes requested):

Lib/unittest/test/testmock/testasync.py
ezio-melotti: Ack for the `k`, and thanks for fixing the `foo`:  I like the normal `foo` much better than the `NormalFoo`

Misc/NEWS.d/next/Library/2019-09-16-09-54-42.bpo-38136.MdI-Zb.rst
lisroach: Oh sorry that makes more sense. I just added something to the assert_awaited() call, I'm not sure if that is enough. Would a full example on the example page be overkill or a good idea?

ezio-melotti (approved):

PR #16198, opened 2019-09-16
bpo-38183: Test_idle ignores user config directory.

terryjreedy:
It no longer tries to create or access .idlerc or any files within.
Users must run IDLE to discover problems with saving settings.

https://bugs.python.org/issue38183


terryjreedy:
Writing 'recent-files.lst' instead of "$HOME/.idlerc/recent-files.lst" resulted in a new ENV CHANGED failure.
@vstinner Is there documentation on what changes cause a failure?

New patch should prevent writing that file.

PR #16202, opened 2019-09-16
bpo-38192: Fix remaining passing of "loop" in the protocol examples

hniksic:
See https://bugs.python.org/issue38192 .

https://bugs.python.org/issue38192

### DIFF
--- a/Doc/library/asyncio-protocol.rst
+++ b/Doc/library/asyncio-protocol.rst
@@ -870,8 +870,8 @@ method, sends data and closes the transport when it receives the answer::
     class EchoClientProtocol:
         def __init__(self, message, on_con_lost):
             self.message = message
-            self.transport = None
             self.on_con_lost = on_con_lost
+            self.transport = None
 
         def connection_made(self, transport):
             self.transport = transport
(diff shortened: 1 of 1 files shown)


aeros (changes requested):
Thanks for the PR @hniksic.

Since this is just removing the explicit passing of *loop* which was is going to be receiving deprecation warnings in 3.8 for several parts of the public API, this looks correct for the most part. The consensus from the asyncio experts seems to be that there is no need for *loop* to be explicitly passed. 

The only suggestion I would make is to modify the ordering within ``main()`` (line 949) and removing the comment where the future is created.
Current:
```
async def main():
    # Get a reference to the event loop as we plan to use
    # low-level APIs.
    loop = asyncio.get_running_loop()

    # Create a pair of connected sockets
    rsock, wsock = socket.socketpair()

    # Prepare a future to inform us when the connection is closed.
    on_con_lost = loop.create_future()
```
Suggested:
```
async def main():
    # Get a reference to the event loop as we plan to use
    # low-level APIs.
    loop = asyncio.get_running_loop()
    on_con_lost = loop.create_future()

    # Create a pair of connected sockets
    rsock, wsock = socket.socketpair()
```
It makes more sense to set up ``on_con_lost = loop.create_future()`` right after acquiring a reference to the event loop, rather than after creating the pair of sockets. Even if it provides the same effect functionally, this makes the coroutine's structuring more logical. 

The removal of the comment might be a bit more subjective, but personally I don't think it provides additional insight. Especially since it wasn't seen as necessary when it was used in the previous two examples, I don't think it's needed for this one either.

But, we can wait for feedback from @asvetlov and @1st1 before moving forward with the suggested changes.

hniksic:
@aeros167 Thanks for the review. I like your suggestion and have updated the PR.

asvetlov (approved):
LGTM, thanks!

PR #16203, opened 2019-09-16
bpo-38155: Adding __all__ to datetime module

ta1hia:
Co-Authored-By: Paul Ganssle <paul@ganssle.io>

https://bugs.python.org/issue38155

### DIFF
--- a/Lib/datetime.py
+++ b/Lib/datetime.py
@@ -4,6 +4,8 @@ See http://www.iana.org/time-zones/repository/tz-link.html for
 time zone and DST data sources.
 """
 
+__all__ = ("date", "datetime", "time", "timedelta", "timezone", "tzinfo", "MINYEAR", "MAXYEAR")
+
 import time as _time
 import math as _math
 import sys
--- a/Lib/datetime.py
+++ b/Lib/datetime.py
@@ -4,7 +4,9 @@ See http://www.iana.org/time-zones/repository/tz-link.html for
 time zone and DST data sources.
 """
 
-__all__ = ("date", "datetime", "time", "timedelta", "timezone", "tzinfo", "MINYEAR", "MAXYEAR")
+__all__ = ("date", "datetime", "time", "timedelta", "timezone", "tzinfo",
+           "MINYEAR", "MAXYEAR")
+
 
 import time as _time
 import math as _math
--- a/Lib/test/datetimetester.py
+++ b/Lib/test/datetimetester.py
@@ -19,7 +19,7 @@ from array import array
 from operator import lt, le, gt, ge, eq, ne, truediv, floordiv, mod
 
 from test import support
-from test.support import is_resource_enabled, ALWAYS_EQ, LARGEST, SMALLEST
+from test.support import import_fresh_module, is_resource_enabled, ALWAYS_EQ, LARGEST, SMALLEST
 
 import datetime as datetime_module
 from datetime import MINYEAR, MAXYEAR
@@ -61,6 +61,19 @@ class TestModule(unittest.TestCase):
         datetime = datetime_module
         self.assertEqual(datetime.MINYEAR, 1)
         self.assertEqual(datetime.MAXYEAR, 9999)
+
+    def test_all(self):
+        """Test that __all__ only points to valid attributes."""
+        all_attrs = dir(datetime_module)
+        for attr in datetime_module.__all__:
+            self.assertIn(attr, all_attrs)
+
+    def test_c_all(self):
+        """Test that __all__ symbols between the c datetime module and
+        the python datetime library are equivalent."""
+        c_datetime = import_fresh_module('datetime', fresh=['_datetime'])
+        py_datetime = import_fresh_module('datetime', blocked=['_datetime'])
+        self.assertEqual(c_datetime.__all__, py_datetime.__all__)
 
     def test_name_cleanup(self):
         if '_Pure' in self.__class__.__name__:
--- a/Lib/test/datetimetester.py
+++ b/Lib/test/datetimetester.py
@@ -19,7 +19,7 @@ from array import array
 from operator import lt, le, gt, ge, eq, ne, truediv, floordiv, mod
 
 from test import support
-from test.support import import_fresh_module, is_resource_enabled, ALWAYS_EQ, LARGEST, SMALLEST
+from test.support import is_resource_enabled, ALWAYS_EQ, LARGEST, SMALLEST
 
 import datetime as datetime_module
 from datetime import MINYEAR, MAXYEAR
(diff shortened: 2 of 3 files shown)


Lib/test/datetimetester.py
```diff
@@ -62,6 +62,19 @@ def test_constants(self):
         self.assertEqual(datetime.MINYEAR, 1)
         self.assertEqual(datetime.MAXYEAR, 9999)
 
+    def test_all(self):
+        """Test that __all__ only points to valid attributes."""
```
ammaraskar, line 66: Hmm, I think there's a simpler way to test this, just do:

```python
from datetime import *
```

This will throw an AttributeError if you have an invalid attribute defined there.

Example (`test.py`):
```python
__all__ = ['a']
b = 1
```
```python
>>> from test import *
AttributeError: module 'test' has no attribute 'a'
```

cc @pganssle

aeros (commented):
Thanks for the PR @tahia-khan.

I have some suggestions regarding the news entry, but overall +1 on the PR. This seems to be well aligned with what @pganssle was asking for in the issue comments.

Misc/NEWS.d/next/Library/2019-09-16-21-47-48.bpo-38155.d92lRc.rst
```diff
@@ -0,0 +1 @@
+Add __all__ to the datetime module, including unit tests.
```
aeros: Made a couple of minor formatting adjustments and removed the part about the tests. News entries typically don't contain information about the tests, as they aren't part of the public API. 
```suggestion
Add ``__all__`` to  :mod:`datetime`.
```

aeros (changes requested):

Lib/datetime.py
```diff
@@ -4,6 +4,8 @@
 time zone and DST data sources.
 """
 
+__all__ = ("date", "datetime", "time", "timedelta", "timezone", "tzinfo", "MINYEAR", "MAXYEAR")
```
corona10: Please make sure it follows to limit all lines to a maximum of 79 characters.
PEP8: https://www.python.org/dev/peps/pep-0008/#maximum-line-length

corona10 (changes requested):

Lib/datetime.py
aeros: In case this isn't clear, this can easily be fixed by changing it to:

```suggestion
__all__ = ("date", "datetime", "time", "timedelta", "timezone", "tzinfo",
                 "MINYEAR", "MAXYEAR")
```

Lib/test/datetimetester.py
```diff
@@ -19,7 +19,7 @@
 from operator import lt, le, gt, ge, eq, ne, truediv, floordiv, mod
 
 from test import support
-from test.support import is_resource_enabled, ALWAYS_EQ, LARGEST, SMALLEST
+from test.support import import_fresh_module, is_resource_enabled, ALWAYS_EQ, LARGEST, SMALLEST
```
aeros: This section is also going over the 79 char limit. Please adjust it accordingly.

Lib/test/datetimetester.py
pganssle: @ammaraskar `from x import *` is only valid at the top level of a file, so this cannot be executed in a test.

Lib/test/datetimetester.py
```diff
@@ -62,6 +62,19 @@ def test_constants(self):
         self.assertEqual(datetime.MINYEAR, 1)
         self.assertEqual(datetime.MAXYEAR, 9999)
 
+    def test_all(self):
+        """Test that __all__ only points to valid attributes."""
+        all_attrs = dir(datetime_module)
+        for attr in datetime_module.__all__:
+            self.assertIn(attr, all_attrs)
+
+    def test_c_all(self):
```
pganssle: I do not think this is the right place for this test - `datetimetester.py` is executed twice - once with the C module and once with the Python module. This may also mess with the state of `sys.modules`, see [this comment](https://github.com/python/cpython/blob/master/Lib/test/test_datetime.py#L20).

I think just drop the `test_c_all` test and just leave the other test.

Misc/NEWS.d/next/Library/2019-09-16-21-47-48.bpo-38155.d92lRc.rst
pganssle: @tahia-khan You may also want to (optionally) add your name to the news - "Patch by Tahia Khan." (or whatever you prefer to be credited as).

pganssle (changes requested):
Thanks for the PR @tahia-khan, and thanks for the reviews @aeros167 and @ammaraskar! With the formatting changes and dropping `test_c_all`, I think this looks pretty good.

I was mildly taken aback that you've defined `__all__` as a tuple rather than a list, which is traditional, but I don't see any reason why it *shouldn't* be a `list`, and in fact `__all__` being immutable makes more sense than mutable, so I'm not sure why it's not more common.

tahia-khan:
Thanks everybody for the detailed change suggestions. I'll wait a bit for the final reviewer and submit all changes in one commit.

> I was mildly taken aback that you've defined `__all__` as a tuple rather than a list, which is traditional, but I don't see any reason why it _shouldn't_ be a `list`, and in fact `__all__` being immutable makes more sense than mutable, so I'm not sure why it's not more common.

@pganssle Ah - I didn't actually realize lists were the usual choice. I was peeking around other files that had an `__all__` defined and it seemed like both lists and tuples were being used, so I went with the immutable choice :)

pganssle:
> Ah - I didn't actually realize lists were the usual choice. I was peeking around other files that had an `__all__` defined and it seemed like both lists and tuples were being used, so I went with the immutable choice :)

Yes, I agree that you made the sensible choice here. I tried looking it up and I could see no one arguing that you *shouldn't* use a `tuple`, and some people arguing that you shouldn't use a `list`, so I say we keep it as `tuple`.

Lib/test/datetimetester.py
aeros: @pganssle Ah, is this already the most viable way to test the validity of the attributes then?

pganssle: Well, it's the best one I could think of. There may be better tests for `__all__`, but given that this module already does some funky and apparently fragile stuff with the import mechanism, I think it's probably best not to try to get too fancy - particularly since it's not *that* important of a test.

Lib/test/datetimetester.py
ta1hia: For this I'll go with parenthesis notation from the [PEP328](https://www.python.org/dev/peps/pep-0328/#rationale-for-parentheses) rather than backslashes.

pganssle: I think you won't need to change this line at all once you have dropped `test_c_all`.

tahia-khan:

pganssle (approved):
Thanks everyone, this looks good, so I've rebased it and I'm going to merge now.

The alignment in the latest version also looked a little off for whatever reason, I aligned the second line in the `__all__` to the parenthesis as part of the rebase.

PR #16204, opened 2019-09-16
bpo-28556: Update the opening note in typing docs

ilevkivskyi:
This PR replaces the old note mentioning that `typing` is a provisional module with a new one mentioning types are not enforced at runtime. I am not sure if there was any official announcement about making `typing` non-provisional, but _de-facto_ no new features were added during Python 3.7, and no backwards incompatible changes were made except for few small things that were considered bugs.

https://bugs.python.org/issue28556

### DIFF
--- a/Doc/library/typing.rst
+++ b/Doc/library/typing.rst
@@ -10,7 +10,7 @@
 
 .. note::
 
-   Python runtime does not enforce the function and variable type annotations.
+   The Python runtime does not enforce function and variable type annotations.
    They can be used by third party tools such as type checkers, IDEs, linters,
    etc.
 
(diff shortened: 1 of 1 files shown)


Doc/library/typing.rst
```diff
@@ -10,10 +10,9 @@
 
 .. note::
 
-   The typing module has been included in the standard library on a
-   :term:`provisional basis <provisional api>`. New features might
-   be added and API may change even between minor releases if deemed
-   necessary by the core developers.
+   Python runtime does not enforce the function and variable type annotations.
```
gvanrossum: *The* Python runtime does not enforce ~the~ function and ...

gvanrossum (approved):
I don't recall exactly how it became non-provisional, I think we already bent the rules for provisional modules by keeping it provisional longer than one release cycle. :-)

PR #16209, opened 2019-09-17
Doc: add list(dict) in stdtypes library

adorilson:
Doc about list(dict) was missing in stdtypes library.

Maybe we must add something in Dictionary view objects section, too.
But, I'm not sure.


aeros (changes requested):
Thanks for the PR @adorilson 

I think this could be a helpful addition, but it should be below the entry for ``iter(d)`` rather than at the top of the list. Those two are functionally similar to each other.

matrixise (approved):
Thank you for your PR.

matrixise:
Thank you for your PR.

PR #16233, opened 2019-09-17
bpo-38187: Fix reference leak in test_tools

pablogsal:
This PR fixes the reference leak in test_tools (there is still a problem with some tests when running -R). The problem is that the `Variable` class in `Tools/c-analyzer/c_parser/info.py` has a descriptor that stores state in the class-level. Therefore when running the tests multiple times, the state is leaked as there and never released.

Probably this needs a better architecture so we don't need to remember to call `Variable._isglobal.instances.clear()` after the tests that use Variable, but for now it fixes the leak.

https://bugs.python.org/issue38187

This one was tricky to find.


ericsnowcurrently (approved):
This is fine as a quick fix.  I'll work on fixing the Slot class, probably using a WeakKeyDictionary.

PR #16242, opened 2019-09-17
No-BPO: Remove unused import in hmac.py

eduardo-elizondo:
Simple one line change to remove an unused import in hmac.py


eduardo-elizondo:
False alarm, it is indeed used :)

PR #16244, opened 2019-09-17
bpo-38070: _Py_DumpTraceback() writes <no Python frame>

vstinner:
When a Python thread has no frame, _Py_DumpTraceback() and
_Py_DumpTracebackThreads() now write "<no Python frame>", rather than
writing nothing.

### DIFF
--- a/Python/traceback.c
+++ b/Python/traceback.c
@@ -797,12 +797,15 @@ dump_traceback(int fd, PyThreadState *tstate, int write_header)
     PyFrameObject *frame;
     unsigned int depth;
 
-    if (write_header)
+    if (write_header) {
         PUTS(fd, "Stack (most recent call first):\n");
+    }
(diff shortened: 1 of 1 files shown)


vstinner:
Example with the change:

```
$ PYTHONHOME=/ ./python 
Fatal Python error: init_fs_encoding: failed to get the Python codec of the filesystem encoding
ModuleNotFoundError: No module named 'encodings'

Current thread 0x00007f8a5a7f4740 (most recent call first):
<no Python frame>
```

without the change:

```
$ PYTHONHOME=/ ./python 
Fatal Python error: init_fs_encoding: failed to get the Python codec of the filesystem encoding
ModuleNotFoundError: No module named 'encodings'

Current thread 0x00007fd955873740 (most recent call first):
```

PR #16127, opened 2019-09-13
bpo-38152: Fix refleak in the finalizer of AST type

eduardo-elizondo:
The `AST_object` was leaking its reference to the `AST` type. This change properly decreases the reference to the type at deallocation time. Also, this modifies `tp_free` to use `PyType_GetSlot` to make `PyTypeObject` opaque.

https://bugs.python.org/issue38152

PR #16248, opened 2019-09-18
bpo-38206: Clarify tp_dealloc requirements for heap allocated types.

ammaraskar:
As mentioned in the bpo ticket, this mistake came up on two reviews:
- https://github.com/python/cpython/pull/16127#pullrequestreview-288312751
- https://github.com/python/cpython/pull/16071#pullrequestreview-287819525

Would be nice to have it documented in a more permanent place than 3.8's whatsnew entry.

https://bugs.python.org/issue38206

### DIFF
--- a/Doc/c-api/type.rst
+++ b/Doc/c-api/type.rst
@@ -118,7 +118,8 @@ The following functions and structs are used to create
 
 .. c:function:: PyObject* PyType_FromSpecWithBases(PyType_Spec *spec, PyObject *bases)
 
-   Creates and returns a heap type object from the *spec*.
+   Creates and returns a heap type object from the *spec*
+   (:const:`Py_TPFLAGS_HEAPTYPE`).
 
    If *bases* is a tuple, the created heap type contains all types contained
    in it as base types.
--- a/Doc/c-api/typeobj.rst
+++ b/Doc/c-api/typeobj.rst
@@ -654,9 +654,9 @@ and :c:type:`PyType_Type` effectively act as defaults.)
    the instance is still in existence, but there are no references to it.  The
    destructor function should free all references which the instance owns, free all
    memory buffers owned by the instance (using the freeing function corresponding
-   to the allocation function used to allocate the buffer), and then call the
-   type's :c:member:`~PyTypeObject.tp_free` function.  If the type is not
-   subtypable (doesn't have the :const:`Py_TPFLAGS_BASETYPE` flag bit set), it is
+   to the allocation function used to allocate the buffer), and call the type's
+   :c:member:`~PyTypeObject.tp_free` function.  If the type is not subtypable
+   (doesn't have the :const:`Py_TPFLAGS_BASETYPE` flag bit set), it is
    permissible to call the object deallocator directly instead of via
    :c:member:`~PyTypeObject.tp_free`.  The object deallocator should be the one used to allocate the
    instance; this is normally :c:func:`PyObject_Del` if the instance was allocated
@@ -664,9 +664,20 @@ and :c:type:`PyType_Type` effectively act as defaults.)
    :c:func:`PyObject_GC_Del` if the instance was allocated using
    :c:func:`PyObject_GC_New` or :c:func:`PyObject_GC_NewVar`.
 
-   Finally, if the type is heap allocated (:const:`Py_TPFLAGS_HEAPTYPE` or
-   built with :c:func:`PyType_FromSpec`), the deallocator should decrement the
-   reference count for its type object.
+   Finally, if the type is heap allocated (:const:`Py_TPFLAGS_HEAPTYPE`), the
+   deallocator should decrement the reference count for its type object after
+   calling the type deallocator. In order to avoid dangling pointers, the
+   recommended way to achieve this is:
+
+   .. code-block:: c
+
+     static void foo_dealloc(foo_object *self) {
+         PyTypeObject *tp = Py_TYPE(self);
+         // free references and buffers here
+         tp->tp_free(self);
+         Py_DECREF(tp);
+     }
+
 
    **Inheritance:**
 
@@ -1025,7 +1036,8 @@ and :c:type:`PyType_Type` effectively act as defaults.)
 
    .. data:: Py_TPFLAGS_HEAPTYPE
 
-      This bit is set when the type object itself is allocated on the heap.  In this
+      This bit is set when the type object itself is allocated on the heap, for
+      example, types created dynamically using :c:func:`PyType_FromSpec`.  In this
       case, the :attr:`ob_type` field of its instances is considered a reference to
       the type, and the type object is INCREF'ed when a new instance is created, and
       DECREF'ed when an instance is destroyed (this does not apply to instances of
--- a/Doc/c-api/typeobj.rst
+++ b/Doc/c-api/typeobj.rst
@@ -654,8 +654,8 @@ and :c:type:`PyType_Type` effectively act as defaults.)
    the instance is still in existence, but there are no references to it.  The
    destructor function should free all references which the instance owns, free all
    memory buffers owned by the instance (using the freeing function corresponding
-   to the allocation function used to allocate the buffer), and finally (as its
-   last action) call the type's :c:member:`~PyTypeObject.tp_free` function.  If the type is not
+   to the allocation function used to allocate the buffer), and then call the
+   type's :c:member:`~PyTypeObject.tp_free` function.  If the type is not
    subtypable (doesn't have the :const:`Py_TPFLAGS_BASETYPE` flag bit set), it is
    permissible to call the object deallocator directly instead of via
    :c:member:`~PyTypeObject.tp_free`.  The object deallocator should be the one used to allocate the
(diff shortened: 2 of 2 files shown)


ammaraskar:
cc @eduardo-elizondo 

Happy to take any suggestions on the wording.

Doc/c-api/type.rst
```diff
@@ -118,7 +118,8 @@ The following functions and structs are used to create
 
 .. c:function:: PyObject* PyType_FromSpecWithBases(PyType_Spec *spec, PyObject *bases)
 
-   Creates and returns a heap type object from the *spec*.
+   Creates and returns a heap type object from the *spec*
+   (:const:`Py_TPFLAGS_HEAPTYPE`).
```
eduardo-elizondo, line 122: Probably not needed, I don't see anything else in type.rst specifying the flags that it sets to the type

Doc/c-api/typeobj.rst
```diff
@@ -654,8 +654,8 @@ and :c:type:`PyType_Type` effectively act as defaults.)
    the instance is still in existence, but there are no references to it.  The
    destructor function should free all references which the instance owns, free all
    memory buffers owned by the instance (using the freeing function corresponding
-   to the allocation function used to allocate the buffer), and finally (as its
-   last action) call the type's :c:member:`~PyTypeObject.tp_free` function.  If the type is not
+   to the allocation function used to allocate the buffer), and then call the
+   type's :c:member:`~PyTypeObject.tp_free` function.  If the type is not
```
eduardo-elizondo: Remove the `then`. Perhaps even include a short blurb saying that the instance might be resurrected by `tp_del` or `tp_finalize`.

Doc/c-api/typeobj.rst
```diff
@@ -664,6 +664,10 @@ and :c:type:`PyType_Type` effectively act as defaults.)
    :c:func:`PyObject_GC_Del` if the instance was allocated using
    :c:func:`PyObject_GC_New` or :c:func:`PyObject_GC_NewVar`.
 
+   Finally, if the type is heap allocated (:const:`Py_TPFLAGS_HEAPTYPE` or
+   built with :c:func:`PyType_FromSpec`), the deallocator should decrement the
```
eduardo-elizondo: I would remove the parenthesis here as it's already implied that a heap allocated type can only come from a managed class or a type created through `PyType_FromSpec`.

That being said, this might be obvious other readers so maybe we can do the following:
* Leave the reference to `:const:Py_TPFLAGS_HEAPTYPE`.
* Modify the documentation of `Py_TPFLAGS_HEAPTYPE` to include that this is set in any python class from `type.__new__` file as well as those created through `PyType_FromSpec`. This follow what `Py_TPFLAGS_READY` specifies with `PyType_Ready`.
With this change people should now be able to click on the `Py_TPFLAGS_HEAPTYPE` flag which will now clearly indicate which types end up with that flag.

Either works for me!

Doc/c-api/typeobj.rst
```diff
@@ -664,6 +664,10 @@ and :c:type:`PyType_Type` effectively act as defaults.)
...
+   reference count for its type object.
```
eduardo-elizondo: Also add that this should specifically happen `after the instance has already been deallocated`. There are two scenarios where this is important:
1) The instance was resurrected. In which you should not decref the type.
2) The instance owns the last reference to the type. If you decref the type (which triggers a `tp_dealloc` on the type itself), now you will have a dangling pointer in the instance's `ob_type` member which might trigger some segfaults.

eduardo-elizondo:
Yup! Looks mostly good, just a couple of comments.

Doc/c-api/type.rst
ammaraskar: The intention behind making this explicit here is to help guide people towards what a heap allocated type actually is. By providing a pointer to `Py_TPFLAGS_HEAPTYPE` it lets them reference the part of the docs that mention its reference counting as well as cross-reference the portion in the `tp_dealloc` docs.

Doc/c-api/typeobj.rst
ammaraskar: Removed the `then`, I really don't know enough about type resurrection to feel confident talking about it so I'm gonna punt that to someone more capable (maybe yourself :))

Doc/c-api/typeobj.rst
ammaraskar: I like that idea, I left the reference to `:const:Py_TPFLAGS_HEAPTYPE` and that mentions `PyType_FromSpec`. Regarding this part: 

>from `type.__new__` file

Could you help me out with the wording? Not sure if it's the best place to mention it as this document deals entirely with the C API but it still might be useful.

Doc/c-api/typeobj.rst
ammaraskar: Done, I also left a suggested implementation like in `tp_finalize` to help explain better (similar to your whatsnew entry). Please look it over

Doc/c-api/typeobj.rst
```diff
@@ -654,16 +654,31 @@ and :c:type:`PyType_Type` effectively act as defaults.)
...
+   deallocator should decrement the reference count for its type object after
+   calling the type deallocator. In order to avoid dangling pointers, the
+   recommended way to achieve this is:
+
+   .. code-block:: c
+
+     static void foo_dealloc(foo_object *self) {
+         PyTypeObject *tp = Py_TYPE(self);
+         // free references and buffers here
+         tp->tp_free(self);
```
eduardo-elizondo, line 677: Since this is a heap type, change this to use the slot API to be compliant with PEP384.

```
freefunc free_func = PyType_GetSlot(tp, Py_tp_free);
free_func(self);
```

eduardo-elizondo:
One last comment. This looks good now! Ping @DinoV , could you help getting this merged?

Doc/c-api/typeobj.rst
```diff
@@ -654,16 +654,32 @@ and :c:type:`PyType_Type` effectively act as defaults.)
...
+   calling the type deallocator. In order to avoid dangling pointers, the
+   recommended way to achieve this is:
+
+   .. code-block:: c
+
+     static void foo_dealloc(foo_object *self) {
+         PyTypeObject *tp = Py_TYPE(self);
+         // free references and buffers here
+         freefunc free_func = PyType_GetSlot(tp, Py_tp_free);
+         free_func(self);
```
encukou: This can also be written simply as `tp->tp_free(self);`.
The text above references `PyTypeObject.tp_free` directly, so `PyType_GetSlot` may be confusing. An API function is also most likely slower than following the pointer directly.

ammaraskar: Hmm, that's what I had it as originally but Eduardo said it's not PEP384 compliant: https://github.com/python/cpython/pull/16248#discussion_r327243422

Doc/c-api/typeobj.rst
encukou: [PEP 384](https://www.python.org/dev/peps/pep-0384/) defines a stable ABI – an API subset you can use when you can't recompile extensions for each minor version of CPython.

How is it relevant here?

eduardo-elizondo: @encukou it is very relevant! What you said is correct, but this PEP has the added benefit of making `PyTypeObject` opaque. This gives the type system control over the implementation of the type object. For instance, in our system, the type object has a very different implementation. Yet, we can correctly interface with C Extensions. Using `tp_free` directly ties the code to CPython's implementation.

By the way, this is only relevant for C extensions, that is, `cpython/Modules` + third-party libs. Internal code CPython code should be free to use all the implementation details it wants! Given that the updated doc has the context of heap types, it's a fair assumption to make :-)

Doc/c-api/typeobj.rst
eduardo-elizondo: See my reply above

Doc/c-api/typeobj.rst
encukou: I agree that using the stable ABI is good, and we all migrate to it eventually. However, it's orthogonal to this change, which is about making the documentation *correct*. Pushing a stable ABI agenda in a fix for "Clarify that tp_dealloc must decref for heap allocated type" seems too sneaky for me – even though I count myself as a proponent of the stable ABI :)

In CPython, `PyType_GetSlot` is likely* much slower than using the slot directly, so it is not a clear win.
I'm for avoiding direct slot access (both in stdlib and extension modules), but only after a wider discussion and perhaps after also adding a fast (and unstable) API.

And again: the text above talks about `PyTypeObject.tp_free`. The example is confusingly out of sync with that. I think we should change both occurences, but please, do it in a different PR.

---

* (I haven't measured, but there's a bunch of extra checks in CPython's `PyType_GetSlot`)

ammaraskar: Reverted back to `tp->tp_free(self);`

eduardo-elizondo: Woops, didn't mean to try any sneaky stuff! 😛 

I agree with your point, let's get to anything related to PEP384 independently.

eduardo-elizondo (approved):
LGTM

encukou (approved):

PR #16254, opened 2019-09-18
gh-62436: Adjust str.isspace to use Unicode's White_Space property.

gnprice:
When Unicode support was first added to Python, there was no Unicode
property identifying whitespace, so we approximated it by putting
together a couple of other properties.

Now there is a White_Space property, so let's use it.

Happily, the difference from our original approximation is only in the
four rare control characters 001C..001F.

As a bonus, `isspace` now joins all similar methods in giving exactly
matching results for ASCII characters represented as `str` or as
`bytes`.  Add a test for that nice property.

https://bugs.python.org/issue18236

* Issue: gh-62436


Lib/test/test_unicode.py
```diff
@@ -625,13 +625,21 @@ def test_isspace(self):
...
         for codepoint in range(sys.maxunicode + 1):
             char = chr(codepoint)
             bidirectional = unicodedata.bidirectional(char)
             category = unicodedata.category(char)
             self.assertEqual(char.isspace(),
-                             (bidirectional in ('WS', 'B', 'S')
-                              or category == 'Zs'))
+                             ((bidirectional in ('WS', 'B', 'S')
+                               or category == 'Zs')
+                              and codepoint not in range(0x1c, 0x20)))
```
Numerlor, line 642: Is this guaranteed to hold in future unicode versions? i.e. should it be tested on the string method if it's only testing the unicode database parsing and generation?

Numerlor (commented):
Ran into an issue with this when I noticed int's stripping is different to str's, so it's nice to see that there's already a PR for a fix.

The introduced changes seem good to my eye that's not entirely familiar with the codebase, apart from the outdated version notices.

I've also noticed that the added docstrings to test_unicode are using single quote marks while others are using double quotes so changing that to be consistent would be nice

PR #16264, opened 2019-09-18
bpo-17258: use sha256 instead of md5 within multiprocessing.connection

stratakis:
Within multiprocessing.connection, deliver_challenge() and
answer_challenge() use hmac for a challenge/response, however
hmac implicitly defaults to using MD5, which fails under FIPS
mode. Hardcode the digest value to sha256.

This PR is adapted from the patch provided at [bpo-17258](https://bugs.python.org/issue17258) and pulled from https://github.com/encukou/cpython/commit/b819ca654873ea36144ce1cb6f8c422a6d2f5d7d

https://bugs.python.org/issue17258

### DIFF
--- a/Lib/multiprocessing/connection.py
+++ b/Lib/multiprocessing/connection.py
@@ -42,6 +42,10 @@ BUFSIZE = 8192
 # A very generous timeout when it comes to local connections...
 CONNECTION_TIMEOUT = 20.
 
+# The hmac module implicitly defaults to using MD5.
+# Support using a stronger algorithm for the challenge/response code:
+HMAC_DIGEST_NAME='sha256'
+
 _mmap_counter = itertools.count()
 
 default_family = 'AF_INET'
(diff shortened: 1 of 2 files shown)


stratakis:
cc @tiran

matrixise (approved):

tiran (changes requested):
I would prefer a more flexible solution that allows us to change algorithms more easily in the future and perhaps allows different auth mechanisms than HMAC challenge and response.

encukou:
> I would prefer a more flexible solution that allows us to change algorithms more easily in the future 

As fas as I can see, changing the algorithm would mean setting `HMAC_DIGEST_NAME` to a new value. That sounds very easy.
What am I missing?

> and perhaps allows different auth mechanisms than HMAC challenge and response.

I don't see how this PR is blocking someone from working on that.

tiran:
The change breaks backward compatibility. multiprocessing supports distributed computing across multiple machines and works with multiple Python versions. With the change a controller with Python 3.9 would no longer be able to talk to a 3.8 server or the other way around.

stratakis:
@tiran What would you propose to move this forward, essentially moving away from md5? Any approaches that would make sense? Something like implementing a client-server handshake negotiation to agree on a hash?

stratakis:
Closing this. Superseded by #20380

PR #16266, opened 2019-09-18
Doc: Clarify dict equality irrespective of ordering.

toonarmycaptain:
Despite dicts now maintaining insertion order, `{'a':1, 'b':2} == {'b':2, 'a':1} >>> True` - and presumably will stay accurate due to the magnitude of such a breaking change.


tirkarthi:
I think by same here it's implicit that order is not considered for dict comparison but making it explicit doesn't seem to hurt though. cc: @rhettinger

toonarmycaptain:
Agreed, but the question was posed in the Python room on stackoverflow by a maintainer, so I figure it doesn't hurt.

matrixise (approved):

matrixise:
Thank you for your PR, this one has been merged into master, 3.8 and 3.7

PR #16267, opened 2019-09-18
Fix or remove dead assignments identified by scan-build

alexhenrie:
[scan-build](https://clang-analyzer.llvm.org/scan-build.html) identified several variable assignments that are overwritten before they can be used. To make the code easier to understand and to ensure that it is fully optimized, let's fix or get rid of these dead assignments.

### DIFF
--- a/Modules/_datetimemodule.c
+++ b/Modules/_datetimemodule.c
@@ -2489,7 +2489,6 @@ delta_new(PyTypeObject *type, PyObject *args, PyObject *kw)
         int x_is_odd;
         PyObject *temp;
 
-        whole_us = round(leftover_us);
         if (fabs(whole_us - leftover_us) == 0.5) {
             /* We're exactly halfway between two integers.  In order
              * to do round-half-to-even, we must determine whether x
--- a/Modules/_xxsubinterpretersmodule.c
+++ b/Modules/_xxsubinterpretersmodule.c
@@ -221,8 +221,9 @@ _sharedexception_bind(PyObject *exctype, PyObject *exc, PyObject *tb)
     if (err->name == NULL) {
         if (PyErr_ExceptionMatches(PyExc_MemoryError)) {
             failure = "out of memory copying exception type name";
+        } else {
+            failure = "unable to encode and copy exception type name";
         }
-        failure = "unable to encode and copy exception type name";
         goto finally;
     }
 
@@ -237,8 +238,9 @@ _sharedexception_bind(PyObject *exctype, PyObject *exc, PyObject *tb)
         if (err->msg == NULL) {
             if (PyErr_ExceptionMatches(PyExc_MemoryError)) {
                 failure = "out of memory copying exception message";
+            } else {
+                failure = "unable to encode and copy exception message";
             }
-            failure = "unable to encode and copy exception message";
             goto finally;
         }
     }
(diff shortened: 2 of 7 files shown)


Modules/_datetimemodule.c
```diff
@@ -2489,7 +2489,6 @@ delta_new(PyTypeObject *type, PyObject *args, PyObject *kw)
         int x_is_odd;
         PyObject *temp;
 
-        whole_us = round(leftover_us);
```
pganssle, line 2492: For anyone else reading, the way this diff is displayed is a bit misleading - line 2488 is not shown, but the same value is assigned when `whole_us` is declared, then overwritten here.

An alternative would be to delete 2488 and change this into a declaration, or to keep the declaration and assignment lines separate, but I'm indifferent between the three.

Python/pylifecycle.c
```diff
@@ -680,7 +680,6 @@ pyinit_config(_PyRuntimeState *runtime,
     if (_PyStatus_EXCEPTION(status)) {
         return status;
     }
-    config = &tstate->interp->config;
```
pganssle, line 683: Hm, this one I'm having trouble understanding how it happened and if it was actually *intended* to have some effect.

Modules/socketmodule.c
```diff
@@ -1639,7 +1639,6 @@ idna_converter(PyObject *obj, struct maybe_idna *data)
         return 1;
     }
     data->obj = NULL;
-    len = -1;
```
pganssle, line 1642: This is dead code now, but `len` gets set in a bunch of different conditional branches. I am mildly worried that something might change in the future where someone forgets to set `len` in one of those branches and we end up using uninitialized memory.

Doesn't look *terribly* dangerous, but maybe it would make sense to leave it as `len = -1` and then add `assert(len >= 0)` directly before `len` is used?

Modules/_xxsubinterpretersmodule.c
```diff
@@ -221,8 +221,9 @@ _sharedexception_bind(PyObject *exctype, PyObject *exc, PyObject *tb)
     if (err->name == NULL) {
         if (PyErr_ExceptionMatches(PyExc_MemoryError)) {
             failure = "out of memory copying exception type name";
+        } else {
+            failure = "unable to encode and copy exception type name";
```
pganssle, line 225: This appears to be an actual bugfix.

eamanu (commented):
Hmm interest PR. Seems to be good. But I can make a depth review

csabella:
@alexhenrie, please open a ticket on the bug tracker for this issue and update the title to include the ticket number.  Thank you!

alexhenrie:
This patch cleans up various totally unrelated functions, so it'd probably be best to split it up and open a bug report for each function that needs changes.

Python/pylifecycle.c
vstinner: This change is correct. I wrote the config, but it's no longer needed to update the config variable.

Python/pathconfig.c
```diff
@@ -265,7 +265,7 @@ config_init_module_search_paths(PyConfig *config, _PyPathConfig *pathconfig)
 
     const wchar_t *sys_path = pathconfig->module_search_path;
     const wchar_t delim = DELIM;
-    const wchar_t *p = sys_path;
+    const wchar_t *p;
```
vstinner, line 268: Would you mind to move the variable declaration inside the loop to even show better its scope?

PR #16269, opened 2019-09-19
Doc: errors tutorial improvements

adorilson:
Improvements:

- Improvements in how try clause works section
  This suggestion is because the execution continues after *except*, not after *try* but before *except*. I guess this form more clear.

- Surrounding some keywords with \*...\*
  For uniformity the highlighted terms  

- Adjust the number of chars per line to 80

Automerge-Triggered-By: GH:aeros


Doc/tutorial/errors.rst
```diff
@@ -101,29 +101,29 @@ The :keyword:`try` statement works as follows.
 * If no exception occurs, the *except clause* is skipped and execution of the
   :keyword:`try` statement is finished.
 
-* If an exception occurs during execution of the try clause, the rest of the
+* If an exception occurs during execution of the *try clause*, the rest of the
```
aeros: Example of where the Sphinx role or code blocks should be used.
```suggestion
* If an exception occurs during execution of the :keyword:`try` clause, the rest of the
```

Doc/tutorial/errors.rst
```diff
@@ -101,29 +101,29 @@ The :keyword:`try` statement works as follows.
...
   clause is skipped.  Then if its type matches the exception named after the
```
aeros: This is not part of the PR, but this minor grammar fix should be made anyways. Wouldn't be worth it's own PR.
```suggestion
   clause is skipped.  Then, if its type matches the exception named after the
```

Doc/tutorial/errors.rst
```diff
@@ -101,29 +101,29 @@ The :keyword:`try` statement works as follows.
...
-  :keyword:`except` keyword, the except clause is executed, and then execution
-  continues after the :keyword:`try` statement.
+  :keyword:`except` keyword, the *except clause* is executed, and then execution
+  continues after the *try/except* block.
```
aeros: Try/except is an appropriate name for the block, but it is not a literal keyword in the code.
```suggestion
  continues after the try/except block.
```

aeros (changes requested):
Thanks for the PR @adorilson.

Regarding the usage of asterisks, they are typically reserved for argument names, so I do not support the idea of using *try clause* or *except clause*. If anything, the emphasis should only surround "try" or "except". Additionally, it should be using either the Spinx ``keyword`` role or inline code blocks. The Sphinx roles provide a specific functional purpose of linking to the documentation, so they are more beneficial when appropriately used.

Sphinx role example:
```
:keyword:`try` clause
```
Code block example:
```
``try`` clause
```
Also, while we typically require that any PRs changing or adding documentation fit within the 80 character line width limit for reST, it is discouraged to change an existing section solely for the purpose of fixing the wrapping, as this can add unnecessary clutter to the commit logs.

Doc/tutorial/errors.rst
```diff
@@ -101,29 +101,29 @@ The :keyword:`try` statement works as follows.
 * If no exception occurs, the *except clause* is skipped and execution of the
   :keyword:`try` statement is finished.
 
-* If an exception occurs during execution of the try clause, the rest of the
-  clause is skipped.  Then if its type matches the exception named after the
-  :keyword:`except` keyword, the except clause is executed, and then execution
-  continues after the :keyword:`try` statement.
+* If an exception occurs during execution of the :keyword:`try` clause, the rest of the
+   clause is skipped.  Then, if its type matches the exception named after the
```
aeros: This extra space is likely causing the doc build failures. 
```suggestion
  clause is skipped.  Then, if its type matches the exception named after the
```

adorilson:
Hi, folks.

What exactly I need to do here?

aeros (approved):
Sorry for the delay, LGTM. (For clarification, I did not have commit privileges when I last left a review)

PR #16270, opened 2019-09-19
bpo-38205: replace get_small_int(ival) macro with a function version

ghost:
Use Py_UNREACHABLE() macro in a safe way.

### DIFF
--- a/Objects/longobject.c
+++ b/Objects/longobject.c
@@ -42,6 +42,8 @@ PyObject *_PyLong_One = NULL;
 */
 static PyLongObject small_ints[NSMALLNEGINTS + NSMALLPOSINTS];
 
+/* If use inline functions here, may lose performance due to
+   unnecessary type casting. */
 #define IS_SMALL_INT(ival) (-NSMALLNEGINTS <= (ival) && (ival) < NSMALLPOSINTS)
 #define IS_SMALL_UINT(ival) ((ival) < NSMALLPOSINTS)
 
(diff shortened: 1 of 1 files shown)


Objects/longobject.c
```diff
@@ -42,6 +42,8 @@ PyObject *_PyLong_One = NULL;
 */
 static PyLongObject small_ints[NSMALLNEGINTS + NSMALLPOSINTS];
 
+/* If use inline functions here, may lose performance due to
+   unnecessary type casting. */
 #define IS_SMALL_INT(ival) (-NSMALLNEGINTS <= (ival) && (ival) < NSMALLPOSINTS)
```
ghost, line 47: Address STINNER Victor's advice: 
> Morever, if using a static inline function is causing issues, it would be nice to add a comment to explain why, so the issue will be avoided in the future.
https://bugs.python.org/issue37812#msg352670

vstinner (changes requested):
" bpo-38205: replace get_small_int(ival) macro with a function version #16270 "

Wait, https://bugs.python.org/issue38205 is about a macro which became a function and then was revert to a macro again... I don't understand why you make it a function... one more time.

Py_UNREACHABLE() should be fixed, not longobject.c.

animalize:
> I like the idea of implementing Py_UNREACHABLE() as a function.

+1, close this PR.

animalize:
>  I don't understand why you make it a function... one more time.

The current code is not consistent:
If "preallocated small integers" enalbed, `get_small_int(ival)` is a function.
If "preallocated small integers" disabled, `get_small_int(ival)` is a macro.

PR #16280, opened 2019-09-19
bpo-38205: Convert Py_UNREACHABLE() macro to a function

vstinner:

### DIFF
--- a/Python/errors.c
+++ b/Python/errors.c
@@ -1612,7 +1612,7 @@ Py_UNREACHABLE(void)
         "The limits were in our heads all along. Follow your dreams.\n"
         "https://xkcd.com/2200");
 #else
-    Py_FatalError("Python internals reached an unreachable code path");
+    Py_FatalError("Unreachable C code path reached");
 #endif
 }
 
--- a/Python/errors.c
+++ b/Python/errors.c
@@ -1603,6 +1603,19 @@ PyErr_ProgramTextObject(PyObject *filename, int lineno)
     return err_programtext(tstate, fp, lineno);
 }
 
+void _Py_NO_RETURN
+Py_UNREACHABLE(void)
+{
+#ifdef Py_DEBUG
+    Py_FatalError(
+        "We've reached an unreachable state. Anything is possible.\n"
+        "The limits were in our heads all along. Follow your dreams.\n"
+        "https://xkcd.com/2200");
+#else
+    Py_FatalError("Python internals reached an unreachable code path");
+#endif
+}
+
 #ifdef __cplusplus
 }
 #endif
--- a/Doc/c-api/intro.rst
+++ b/Doc/c-api/intro.rst
@@ -105,15 +105,6 @@ defined closer to where they are useful (e.g. :c:macro:`Py_RETURN_NONE`).
 Others of a more general utility are defined here.  This is not necessarily a
 complete listing.
 
-.. c:macro:: Py_UNREACHABLE()
-
-   Use this when you have a code path that you do not expect to be reached.
-   For example, in the ``default:`` clause in a ``switch`` statement for which
-   all possible values are covered in ``case`` statements.  Use this in places
-   where you might be tempted to put an ``assert(0)`` or ``abort()`` call.
-
-   .. versionadded:: 3.7
-
 .. c:macro:: Py_ABS(x)
 
    Return the absolute value of ``x``.
--- a/Doc/c-api/sys.rst
+++ b/Doc/c-api/sys.rst
@@ -380,3 +380,17 @@ Process Control
    function registered last is called first. Each cleanup function will be called
    at most once.  Since Python's internal finalization will have completed before
    the cleanup function, no Python APIs should be called by *func*.
+
+.. c:function:: void Py_UNREACHABLE(void)
+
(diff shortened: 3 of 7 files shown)


Python/errors.c
```diff
@@ -1603,6 +1603,19 @@ PyErr_ProgramTextObject(PyObject *filename, int lineno)
...
+{
+#ifdef Py_DEBUG
+    Py_FatalError(
+        "We've reached an unreachable state. Anything is possible.\n"
+        "The limits were in our heads all along. Follow your dreams.\n"
+        "https://xkcd.com/2200");
+#else
+    Py_FatalError("Python internals reached an unreachable code path");
+#endif
+}
```
ghost, line 1617: `abort();` missed

vstinner:
This change:
* Now dump the Python traceback of all threads when Py_UNREACHABLE() is called: call Py_FatalError()
* Call DebugBreak() on Windows in debug mode (Py_FatalError)
* Should make Python binary a few bytes smaller in debug mode: replace fputs()+abort() calls with a single call to Py_UNREACHABLE()
* Keep the short version of the xkcd joke in debug mode

I didn't keep the long version of the joke since RANDALL_WAS_HERE compilation flag is not documented anyway. If someone wants to keep it, I suggest to document it in Misc/SpecialBuilds.txt. Moreover, I dislike untested code :-/ I'm also fine with merging the two xkcd texts into a single longer one if you prefer.

Recently, @serhiy-storchaka fixed "from __future__ import barry_as_FLUFL": this easteregg doesn't need to rebuild Python, so I'm fine with keep it.

Python/errors.c
vstinner: Py_FatalError() exit Python: https://docs.python.org/dev/c-api/sys.html#c.Py_FatalError

vstinner:
You can test this change by (1) applying this change:

```
diff --git a/Python/bltinmodule.c b/Python/bltinmodule.c
index a40eb421f5..a85c3fbaae 100644
--- a/Python/bltinmodule.c
+++ b/Python/bltinmodule.c
@@ -1552,6 +1552,7 @@ builtin_len(PyObject *module, PyObject *obj)
         assert(PyErr_Occurred());
         return NULL;
     }
+    if (res == 123456789) { Py_UNREACHABLE(); }
     return PyLong_FromSsize_t(res);
 }
 
```

(2) run this script:

```
class HaveLength:
    def __len__(self):
        return 123456789

def func(obj):
    print("len", len(obj))

def main():
    func(HaveLength())

main()
```

Output with this change (in debug mode):

```
$ ./python x.py 
Fatal Python error: We've reached an unreachable state. Anything is possible.
The limits were in our heads all along. Follow your dreams.
https://xkcd.com/2200

Current thread 0x00007f81b54ec740 (most recent call first):
  File "/home/vstinner/python/master/x.py", line 6 in func
  File "/home/vstinner/python/master/x.py", line 9 in main
  File "/home/vstinner/python/master/x.py", line 11 in <module>
Aborted (core dumped)
```

Output without this change (in debug mode):

```
$ ./python x.py
ERROR:

We've reached an unreachable state. Anything is possible.
The limits were in our heads all along. Follow your dreams.

https://xkcd.com/2200
Aborted (core dumped)
```

This main enhancement is that you now **get the Python traceback where the bug occurred**.

vstinner:
This change fix https://bugs.python.org/issue38205

Python/errors.c
```diff
@@ -1603,6 +1603,19 @@ PyErr_ProgramTextObject(PyObject *filename, int lineno)
...
+void _Py_NO_RETURN
+Py_UNREACHABLE(void)
+{
+#ifdef Py_DEBUG
+    Py_FatalError(
+        "We've reached an unreachable state. Anything is possible.\n"
+        "The limits were in our heads all along. Follow your dreams.\n"
+        "https://xkcd.com/2200");
+#else
+    Py_FatalError("Python internals reached an unreachable code path");
```
zware: "Python internals" isn't quite accurate; since this is a public function it may be used in third-party code (which is probably more likely to actually hit this function :)), but "Python internals" makes it sound like it's definitely our bug and so we'll get the report.  Maybe something like "Unreachable C code path reached"?

zware (commented):
One nit about the release-mode message, but the overall change looks reasonable to me.

Python/errors.c
vstinner: I changed the message.

vstinner:
@zware: @serhiy-storchaka prefers to keep a macro rather converting it to a function: https://bugs.python.org/issue38205#msg352782 What's the opinion?

zware:
I'm somewhat in favor of keeping it as a macro just for complete backwards compatibility, even if it's implemented as a private `_Py_Unreachable` function.  I will note that about the only third-party usage of `Py_UNREACHABLE` that comes up in a quick GitHub search is in `py3c` which checks `#if defined(Py_UNREACHABLE)`, which may not play nicely with a function.

vstinner:
I wrote PR #16290 which fix https://bugs.python.org/issue38205 but keeps Py_UNREACHABLE() as a macro.

matrixise:
@vstinner I think you can close this one and merge PR #16290

vstinner:
It was decided to merge PR #16290 instead.

PR #16282, opened 2019-09-19
bpo-37531: Skip test_regrtest.test_multiprocessing_timeout() on all platforms

vstinner:


vstinner:
"Azure Pipelines PR — #20190919.22 failed" it seems like the job has been cancelled, but I don't understand why.

I closed/reopened my PR to schedule a new Azure Pipelines run.

PR #16288, opened 2019-09-19
bpo-33694: Remove test_asyncio ProactorDatagramTransportTests

vstinner:
ProactorDatagramTransportTests tests are too close to the
implementation.

https://bugs.python.org/issue33694


vstinner:
Since my commit 79790bc35fe722a49977b52647f9b5fe1deda2b7, nobody came up with a better way to test this code, so I propose to simply remove it.

On buildbots, when test_asyncio fails, it is re-run in verbose mode. If the overall job fails, a bot looks into buildot logs and copy them to the pull request which introduced the regression. Because of "... FIXME: bpo-33694: ..." messages, tons of pull requests have be associated to https://bugs.python.org/issue33694 whereas they are unrelated to asyncio...

cc @pablogsal

asvetlov (approved):
Good to me. 
@1st1 hates mock based asyncio tests for the same reason, I agree with him.

These tests were important in making the first implementation of UDP proactor sockets; the tests helped me to figure out how the code actually works.
Now the tests are useless

pablogsal (approved):

vstinner:
> @1st1 hates mock based asyncio tests for the same reason, I agree with him.

Yeah, me too. I hate tests which basically tests mock rather than testing the actual code.

PR #16291, opened 2019-09-19
bpo-38045: Avoid quadratic complexity when creating composite IntFlags.

anntzer:
When one creates all possible combos of an IntFlag, e.g. with

    from enum import Flag
    F = Flag("F", list("abcdefghijklm"))
    for idx in range(2**len(F) - 1):
        F(idx)

the previous version would exhibit quadratic complexity because
`_value2member_map_` would grow bigger over each iteration.

Iterate over "true" flag members rather than `_value2member_map_`,
to avoid this problem.  For a 13-member flag (so 8192 combinations),
the speedup is more than 100x.

https://bugs.python.org/issue38045

### DIFF
--- a/Lib/enum.py
+++ b/Lib/enum.py
@@ -860,29 +860,23 @@ def _decompose(flag, value):
     """Extract all members from the value."""
     # _decompose is only called if the value is not named
     not_covered = value
-    negative = value < 0
-    # issue29167: wrap accesses to _value2member_map_ in a list to avoid race
-    #             conditions between iterating over it and having more pseudo-
-    #             members added to it
-    if negative:
-        # only check for named flags
-        flags_to_check = [
-                (m, v)
-                for v, m in list(flag._value2member_map_.items())
-                if m.name is not None
-                ]
-    else:
-        # check for named flags and powers-of-two flags
-        flags_to_check = [
-                (m, v)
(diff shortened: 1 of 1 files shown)


the-knights-who-say-ni:

## Recognized GitHub username

@anntzer

hongweipeng:
I think it's better not to create members in `_decompose()`. Just like its name, it only does decomposition work.

hongweipeng:
I created a commit without creating members https://github.com/hongweipeng/cpython/commit/a5f06010962cafa457a876835e7789ec30eb783e . the speedup is aslo more than 100x in the example and passes test_enum.py.

anntzer:
Your patch fixes the problem with Flag, but not with IntFlag (replace Flag by IntFlag in the original example and the slow behavior still exists with your patch, but not with mine).

hongweipeng:
@anntzer Thanks for your response. You are right. The previous commit did not resolve `IntFlag`, but I think there are still ways to solve `Flag` and `IntFlag` without creating members. So I create another commit, see in https://github.com/hongweipeng/cpython/commit/9ae31f416b0d4be4e12f746da08dfe750a5a3d7b .

anntzer:
Superseded by https://github.com/python/cpython/pull/16483.

PR #16298, opened 2019-09-20
bpo-38234: Fix PyConfig_Read() when Py_SetPath() was called

vstinner:
* If Py_SetPath() has been called, _PyConfig_InitPathConfig() now
  uses its value.
* Py_Initialize() now longer copies path configuration from PyConfig
  to the global path configuration (_Py_path_config).


pablogsal (approved):
LGTM

vstinner:
I rebased my PR on top of commit c422167749f92d4170203e996a2c619c818335ea.

PR #16299, opened 2019-09-20
[3.8] bpo-38093: Correctly returns AsyncMock for async subclasses. (GH-15947)

lisroach:


Lib/unittest/mock.py
```diff
@@ -988,9 +988,13 @@ def _get_child_mock(self, /, **kw):
         _type = type(self)
         if issubclass(_type, MagicMock) and _new_name in _async_method_magics:
             klass = AsyncMock
-        if issubclass(_type, AsyncMockMixin):
+        elif _new_name  in _sync_async_magics:
+            # Special case these ones b/c users will assume they are async,
```
matrixise, line 992: `b/c` ? sorry, I don't know that term, what's that?

Lib/unittest/test/testmock/testasync.py
```diff
@@ -377,43 +377,88 @@ def test_add_side_effect_iterable(self):
...
+            response = AsyncMock(name='response')
+            response.json = AsyncMock(return_value={'json': 123})
+            cm.__aenter__.return_value = response
+            pc.session.post.return_value = cm
+            result = asyncio.run(pc.main())
+            self.assertEqual(result, {'json': 123})
+
+        for mock_type in [AsyncMock, MagicMock]:
+            with self.subTest(f"test set return value of aenter with {mock_type}"):
+                inner_test(mock_type)
```
matrixise, line 436: :+1:

Lib/unittest/test/testmock/testasync.py
```diff
@@ -481,27 +526,30 @@ def __aiter__(self):
...
-        if asyncio.iscoroutine(iterator):
-            iterator = asyncio.run(iterator)
-
-        mock_iterator = mock_instance.__aiter__()
-        if asyncio.iscoroutine(mock_iterator):
-            mock_iterator = asyncio.run(mock_iterator)
+    def test_aiter_set_return_value(self):
+        mock_iter = AsyncMock(name="tester")
+        mock_iter.__aiter__.return_value = [1, 2, 3]
+        async def main():
```
matrixise, line 532: maybe old school, but what do you think if you add one empty line before the definition of main and after?

Lib/unittest/test/testmock/testasync.py
```diff
@@ -512,19 +560,30 @@ def test_mock_async_for(self):
...
 
-        with self.subTest("iterate through set return_value iterator"):
-            mock_instance = MagicMock(self.WithAsyncIterator())
+        def test_set_return_value_iter(mock_type):
+            mock_instance = mock_type(self.WithAsyncIterator())
             mock_instance.__aiter__.return_value = iter(expected[:])
-            self.assertEqual(expected, asyncio.run(iterate(mock_instance)))
+            self.assertEqual(asyncio.run(iterate(mock_instance)), expected)
+
+        for mock_type in [AsyncMock, MagicMock]:
```
matrixise, line 578: I understand this construction but I would prefer something like that, but we don't use pytest. (maybe a new feature for the unittest lib?)
```python
# @pytest.mark.parametrize('mock_type', (AsyncMock, MagicMock))
@unittest.parametrize('mock_type', (AsyncMock, MagicMock))
def test_default(self, mock_type):
    pass
```

matrixise (approved):
Thanks @lisroach. but if you can read my comments. maybe we could add a feature like `pytest.mark.parametrize` in unittest.py?

Lib/unittest/test/testmock/testasync.py
lisroach: Not a bad idea! I think there are a number of things we could borrow from pytest. Parameterization would make this kind of construction much nicer.

PR #16300, opened 2019-09-20
bpo-38236: Dump path config at first import error

vstinner:
Python now dumps path configuration if it fails to import the Python
codecs of the filesystem and stdio encodings.

https://bugs.python.org/issue38236


vstinner:
See https://bugs.python.org/issue38236#msg352873 for an output example.

matrixise (approved):
Thanks for this feature, sometimes, I have this error. Need to fix the conflict? Thank you

methane (approved):

PR #16304, opened 2019-09-20
bpo-38187: Fix a refleak in Tools/c-analyzer.

ericsnowcurrently:
The "Slot" helper (descriptor) is leaking references due to its caching mechanism.  The PR includes a partial fix to Slot, but also adds Variable.storage to replace the problematic use of Slot.

https://bugs.python.org/issue38187

### DIFF
--- a/Tools/c-analyzer/c_analyzer_common/known.py
+++ b/Tools/c-analyzer/c_analyzer_common/known.py
@@ -34,34 +34,41 @@ def from_file(infile, *,
         id = ID(filename, funcname, name)
         if kind == 'variable':
             values = known['variables']
-            value = Variable(id, declaration)
-            value._isglobal = _is_global(declaration) or id.funcname is None
+            if funcname:
+                storage = _get_storage(declaration) or 'local'
+            else:
+                storage = _get_storage(declaration) or 'implicit'
+            value = Variable(id, storage, declaration)
         else:
             raise ValueError(f'unsupported kind in row {row}')
-        if value.name == 'id' and declaration == UNKNOWN:
-            # None of these are variables.
-            declaration = 'int id';
-        else:
-            value.validate()
+        value.validate()
+#        if value.name == 'id' and declaration == UNKNOWN:
+#            # None of these are variables.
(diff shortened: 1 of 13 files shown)


Tools/c-analyzer/c_analyzer_common/known.py
```diff
@@ -34,34 +34,41 @@ def from_file(infile, *,
...
-        return True
-    if vartype.startswith(('_Py_IDENTIFIER(', '_Py_static_string(')):
-        return True
-    if vartype.startswith('PyDoc_VAR('):
-        return True
-    if vartype.startswith(('SLOT1BINFULL(', 'SLOT1BIN(')):
-        return True
-    if vartype.startswith('WRAP_METHOD('):
-        return True
+    if decl.startswith('static '):
```
aeros, line 56: What is the purpose of including the last space character in the string ``'static '``? I noticed this was also present in ``'extern '`` below.

Tools/c-analyzer/c_parser/info.py
```diff
@@ -63,18 +89,17 @@ def validate(self):
...
     @property
     def isglobal(self):
-        try:
-            return self._isglobal
-        except AttributeError:
-            # XXX Include extern variables.
-            # XXX Ignore functions.
-            self._isglobal = ('static' in self.vartype.split())
-            return self._isglobal
+        return self.storage != 'local'
```
aeros, line 102: Wouldn't an implicit also not be considered a global?

```suggestion
        return self.storage != 'local' or self.storage != 'implicit'
```

Tools/c-analyzer/c_analyzer_common/known.py
1st1: So that you don't accidentally parse identifiers like `static_something`

Tools/c-analyzer/c_parser/info.py
ericsnowcurrently: In the context of the tool, "global" is any variable that maintains state at a process-global level.  That includes non-local implicits.

PR #16319, opened 2019-09-21
bpo-38244: [Documentation] Fix unreliable link to sever in ftplib.rst

Kriyszig:
* The link to `ftp.debian.org` is unreliable due to a failed redirect. Instead the main URL to `ftp.us.debian.org` is preferred

https://bugs.python.org/issue38244

### DIFF
--- a/Doc/library/ftplib.rst
+++ b/Doc/library/ftplib.rst
@@ -22,7 +22,7 @@ as mirroring other FTP servers.  It is also used by the module
 Here's a sample session using the :mod:`ftplib` module::
 
    >>> from ftplib import FTP
-   >>> ftp = FTP('ftp.us.debian.org')     # connect to host, default port
+   >>> ftp = FTP('ftp.us.debian.org')  # connect to host, default port
    >>> ftp.login()                     # user anonymous, passwd anonymous@
    '230 Login successful.'
    >>> ftp.cwd('debian')               # change into "debian" directory
--- a/Doc/library/ftplib.rst
+++ b/Doc/library/ftplib.rst
@@ -22,7 +22,7 @@ as mirroring other FTP servers.  It is also used by the module
 Here's a sample session using the :mod:`ftplib` module::
 
    >>> from ftplib import FTP
-   >>> ftp = FTP('ftp.debian.org')     # connect to host, default port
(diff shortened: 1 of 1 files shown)


Doc/library/ftplib.rst
```diff
@@ -22,7 +22,7 @@ as mirroring other FTP servers.  It is also used by the module
 Here's a sample session using the :mod:`ftplib` module::
 
    >>> from ftplib import FTP
-   >>> ftp = FTP('ftp.debian.org')     # connect to host, default port
+   >>> ftp = FTP('ftp.us.debian.org')     # connect to host, default port
```
ghost: Please remove 3 spaces before the comment, so that these `#` can be aligned.

Kriyszig: Addressed in 971d84f 
Sorry I didn't notice that :sweat_smile: Thanks for pointing it out

aeros (commented):
Thanks for the PR @Kriyszig.

While I can understand the intent of this PR, I don't think we should recommend a server that's specific to a single country (if we can avoid doing so). This would not work as well for those on the opposite side of the world from the US. According to https://www.debian.org/mirror/list, it is recommended to use the following format to access their primary FTP mirrors:

```
 ftp.<country>.debian.org 
```

The only problem with using the above directly is that it wouldn't be valid in literal code, so that's up to the experts of ``ftplib`` to decide what should be used instead. But either way, we should stop recommending ``ftp.debian.org``. From the [debian-mirrors-anounce ML](https://lists.debian.org/debian-devel-announce/2007/12/msg00002.html):
```
In the future, it may get services reduced, or shut down, or converted into
a globally load-balanced name, or whatever. Please don't use it.

If you're using it now, please switch to a country-based DNS name
such as ftp.us.debian.org, ftp.ca.debian.org, ftp.uk.debian.org, ...
The list of those servers is at http://www.debian.org/mirror/list
```

aeros167:
Expert for ``ftplib``:
/cc @giampaolo

giampaolo:
As long as ftp.us.debian.org is up, I'm OK with the PR as-is. If in the future we can find a better alternative we can just change it again.

PR #16321, opened 2019-09-21
bpo-38216, bpo-36274: Allow subclasses to override validation and encoding behavior

jaraco:
This patch introduces a new semi-private hook for subclasses to override the validation and encoding behavior in `http.client.HTTPConnection.putrequest`, allowing select clients to customize the behavior and implement invalid requests as was possible prior to the patch in [bpo-30458](https://bugs.python.org/issue30458) (for control characters) and prior to Python 3.0 (for encoding).

### DIFF
--- a/Lib/http/client.py
+++ b/Lib/http/client.py
@@ -1172,15 +1172,20 @@ class HTTPConnection:
             # For HTTP/1.0, the server will assume "not chunked"
             pass
 
+
+    # The encoding URLs are constrained to before hitting the wire.
+    _prepare_path_encoding = 'ascii'
+
     def _prepare_path(self, url):
         """Validate a url for putrequest and return encoded bytes."""
         # Prevent CVE-2019-9740.
-        if match := _contains_disallowed_url_pchar_re.search(url):
+        match = _contains_disallowed_url_pchar_re.search(url)
+        if match:
             raise InvalidURL(f"URL can't contain control characters. {url!r} "
                              f"(found at least {match.group()!r})")
 
-        # Require ASCII characters only
-        return url.encode('ascii')
+        # Require a specific character set (normally ASCII).
+        return url.encode(self._prepare_path_encoding)
 
     def putheader(self, header, *values):
         """Send a request header line to the server.
--- a/Lib/http/client.py
+++ b/Lib/http/client.py
@@ -1094,7 +1094,6 @@ class HTTPConnection:
             self._http_vsn_str.encode('ascii')
         ))
 
-        # Non-ASCII characters should have been eliminated earlier
         self._output(request)
 
         if self._http_vsn == 11:
--- a/Lib/http/client.py
+++ b/Lib/http/client.py
@@ -1173,8 +1173,9 @@ class HTTPConnection:
             pass
 
 
-    # The encoding URLs are constrained to before hitting the wire.
-    _prepare_path_encoding = 'ascii'
+    def _encode_prepared_path(self, str_url):
+        # ASCII also helps prevent CVE-2019-9740.
+        return str_url.encode('ascii')
 
     def _prepare_path(self, url):
         """Validate a url for putrequest and return encoded bytes."""
@@ -1184,8 +1185,7 @@ class HTTPConnection:
             raise InvalidURL(f"URL can't contain control characters. {url!r} "
                              f"(found at least {match.group()!r})")
 
-        # Require a specific character set (normally ASCII).
-        return url.encode(self._prepare_path_encoding)
+        return self._encode_prepared_path(url)
 
     def putheader(self, header, *values):
         """Send a request header line to the server.
--- a/Lib/http/client.py
+++ b/Lib/http/client.py
@@ -1089,10 +1089,7 @@ class HTTPConnection:
         self._method = method
         if not url:
             url = '/'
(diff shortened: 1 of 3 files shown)


tirkarthi:
So the fix would be that projects like cherrypy can override this private function to make sure crlf is not validated? I understand it's a private hook but not sure how it could be documented if some other project stumbles upon this and wants to bypass the validation. IIRC, the original discussion was about introducing a parameter to turn off validation in putrequest but was given up due to introducing a new parameter in a maintenance release.

jaraco:
> So the fix would be that projects like cherrypy can override this private function to make sure crlf is not validated?

Essentially, yes. I specifically wanted to avoid adding a parameter due to the public exposure that brings (documentation, explanation, additional interfaces). Almost no clients are going to want to override this behavior. Those that do will very likely trace the code to this place and see that the behavior can be overridden. The tests enforce that contract so it isn't accidentally removed, but otherwise it's tucked away with little documentation. Anyone asking about it may refer to the tests for examples of how it might be used.

Lib/test/test_httplib.py
```diff
@@ -1155,6 +1155,63 @@ def run_server():
...
+        """
+        It should be possible to override the default encoding
+        to transmit bytes in another encoding even if invalid
+        (bpo-36274).
+        """
+        class UnsafeHTTPConnection(client.HTTPConnection):
+            def _prepare_path(self, url):
+                # Prevent CVE-2019-9740.
+                if match := client._contains_disallowed_url_pchar_re.search(
+                        url):
```
webknjaz: This little hanging thing looks weird. Maybe do smth like
```python
pattern = client._contains_disallowed_url_pchar_re
if match := pattern.search(url):
```

webknjaz (approved):

Lib/test/test_httplib.py
zooba: Or just ignore the line length _guideline_ for this one case.

gpshead: it's fine as is.  using `m` as the name instead of `match` would've been my suggestion. ;)

Lib/test/test_httplib.py
```diff
@@ -1155,6 +1155,63 @@ def run_server():
...
+                    raise InvalidURL(
+                        f"URL can't contain control characters. {url!r} "
+                        f"(found at least {match.group()!r})")
+
+                return url.encode('utf-8')
```
gpshead: Rather than requiring people to override the entire def _prepare_path method just to change the encoding it allows (which, as this example shows, means copy and pasting the first part of the implementation), how about making a `._prepare_path_encoding` attribute to the class that could be overridden.  it'd default to `'ascii'` but you would set it to `'utf-8'` in this subclass.  instead of overriding the method.

Lib/http/client.py
```diff
@@ -1085,18 +1085,17 @@ def putrequest(self, method, url, skip_host=False,
...
-                             f"(found at least {match.group()!r})")
-        request = '%s %s %s' % (method, url, self._http_vsn_str)
+
+        request = b' '.join((
+            method.encode('ascii'),
+            self._prepare_path(url or '/'),
+            self._http_vsn_str.encode('ascii')
+        ))
 
         # Non-ASCII characters should have been eliminated earlier
```
gpshead: We can get rid of this comment now.  it was referring to the `.encode('ascii')` that is no longer below.

gpshead (changes requested):

gpshead (changes requested):
also, please use blurb to add a NEWS entry for this change.

gpshead:
I'm looking into addressing my suggested changes now so that this can go in.

jaraco:
I'll look at this later this evening.

Lib/test/test_httplib.py
jaraco: I considered this, but the proposed change that inspired this change wished to change the invocation of `encode('ascii')` to `encode('ascii', errors='surrogateescape')`, so that seemed like enough of a signal that one might wish to override the encoding process entirely.

jaraco: Does it still make sense to use `self._prepare_path_encoding` so it's easier to override if that's what someone wants?

gpshead: i suppose _prepare_path_encoding could be its own method in that case.  I wanted to avoid people needing to duplicate the control characters check for this situation.

gpshead (approved):

gpshead:
Waiting on the CI results here.  I don't anticipate problems.  Please take a look and let me know what you think.  I want to get this in over the weekend.

jaraco:
This approach looks acceptable to me. I'm still not super-happy with it (not your changes in particular, just the problem); I hate how the _prepare_path and _encode_prepared_path are entangled. I took a stab at a slightly less aggressive approach that once again separates the concerns of _prepare_path and _encode_prepared_path in [this branch](https://github.com/jaraco/cpython/tree/feature/38216-putrequest-hooks-separate) (currently one commit ahead of this PR). @gpshead, if you have time or energy, would you comment on that approach? Otherwise, we can move forward with this approach.

gpshead:
i'm inclined to go forward with your new PR #16448

ned-deily:
> i'm inclined to go forward with your new PR #16448 
Me, too.

PR #16328, opened 2019-09-22
bpo-36876: Moved Parser/listnode.c statics to interpreter state.

vsajip:

### DIFF
--- a/Include/internal/pycore_pystate.h
+++ b/Include/internal/pycore_pystate.h
@@ -139,8 +139,8 @@ struct _is {
         struct {
             int level;
             int atbol;
-        } listnode_data;
-    } parser_data;
+        } listnode;
+    } parser;
 };
 
 PyAPI_FUNC(struct _is*) _PyInterpreterState_LookUpID(PY_INT64_T);
--- a/Include/internal/pycore_pystate.h
+++ b/Include/internal/pycore_pystate.h
@@ -132,6 +132,15 @@ struct _is {
     struct _warnings_runtime_state warnings;
 
     PyObject *audit_hooks;
+/*
+ * See bpo-36876: miscellaneous ad hoc statics have been moved here.
+ */
+    struct {
+        struct {
+            int level;
+            int atbol;
+        } listnode_data;
+    } parser_data;
 };
 
 PyAPI_FUNC(struct _is*) _PyInterpreterState_LookUpID(PY_INT64_T);
--- a/Parser/listnode.c
+++ b/Parser/listnode.c
@@ -21,8 +21,8 @@ listnode(FILE *fp, node *n)
 {
     PyInterpreterState *interp = _PyInterpreterState_GET_UNSAFE();
 
-    interp->parser_data.listnode_data.level = 0;
-    interp->parser_data.listnode_data.atbol = 1;
+    interp->parser.listnode.level = 0;
+    interp->parser.listnode.atbol = 1;
     list1node(fp, n);
 }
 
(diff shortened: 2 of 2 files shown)


Include/internal/pycore_pystate.h
```diff
@@ -132,6 +132,15 @@ struct _is {
...
 
     PyObject *audit_hooks;
+/*
+ * See bpo-36876: miscellaneous ad hoc statics have been moved here.
+ */
+    struct {
+        struct {
+            int level;
+            int atbol;
+        } listnode_data;
```
ericsnowcurrently: "_data" is a little redundant here.

```suggestion
        } listnode;
```

Include/internal/pycore_pystate.h
```diff
@@ -132,6 +132,15 @@ struct _is {
...
+    } parser_data;
```
ericsnowcurrently: Likewise with "_data".

```suggestion
    } parser;
```

Include/internal/pycore_pystate.h
```diff
@@ -132,6 +132,15 @@ struct _is {
     struct _warnings_runtime_state warnings;
 
     PyObject *audit_hooks;
+/*
+ * See bpo-36876: miscellaneous ad hoc statics have been moved here.
+ */
+    struct {
```
ericsnowcurrently, line 138: If you think there will be any other parser/compiler-related state to move to PyInterpreterState then consider making this struct stand-alone and perhaps even creating pycore_compiler.h (or pycore_parser.h) for it.  That way the code doesn't get too cluttered.

Parser/listnode.c
```diff
@@ -15,19 +16,21 @@ PyNode_ListTree(node *n)
...
-static int level, atbol;
-
 static void
 listnode(FILE *fp, node *n)
 {
-    level = 0;
-    atbol = 1;
+    PyInterpreterState *interp = _PyInterpreterState_GET_UNSAFE();
+
+    interp->parser_data.listnode_data.level = 0;
```
ericsnowcurrently: I was going to point to the advantage of having stand-alone structs for places like this (so there isn't so much going on with each line).  However, I actually kind of like how the line captures the indirection clearly. :)

ericsnowcurrently (approved):
mostly LGTM

I have left just a few small suggestions to consider.

Include/internal/pycore_pystate.h
vsajip: Done.

Include/internal/pycore_pystate.h
vsajip: Done.

vstinner:
```
interp = _PyInterpreterState_GET_UNSAFE(); 
```

You can add listnode or parser variable to avoid having to write "interp->parser.listnode". But for that, you need to name the structure in pycore_pystate.h.

vsajip:
I'm not quite sure what you mean. The change was specifically to move those variables into the interpreter state rather than stand-alone variables, as they were previously.

vstinner:
I suggest to add `parser = &interp->parser;` and then use `parser` instead of `interp`. Or something similar with `listnode`.

vsajip:
Ah, OK, I understand now :-)

Not sure what an additional variable buys in this case. I agree, if readability were adversely affected, it would be worth doing. As it is, it seems readable enough.

vstinner:
It was just a minor remark for readability. You're free to ignore it ;-)

PR #16331, opened 2019-09-23
bpo-38255: Replace "method" with "attribute" in the description of super()

geryogam:
This PR will make the following change to the [_Built-in Functions_ chapter](https://docs.python.org/3.8/library/functions.html#super) of the library documentation:

- replace the word "method" with the more general word "attribute" in the description of `super()`, since `super()` is not restricted to method access but can also do data attribute access:

```
>>> class A:
...     x = True
... 
>>> class B(A):
...     x = False
... 
>>> B().x
False
>>> super(B, B()).x
True
```

https://bugs.python.org/issue38255

### DIFF
--- a/Doc/library/functions.rst
+++ b/Doc/library/functions.rst
@@ -1594,7 +1594,7 @@ are always available.  They are listed here in alphabetical order.
 .. function:: super([type[, object-or-type]])
 
    Return a proxy object that delegates attribute access to a parent or sibling
-   class of *type*.  This is useful for accessing inherited methods that have
+   class of *type*.  This is commonly used for accessing inherited methods that have
    been overridden in a class.
 
    The *object-or-type* determines the :term:`method resolution order`
--- a/Doc/library/functions.rst
+++ b/Doc/library/functions.rst
@@ -1593,8 +1593,8 @@ are always available.  They are listed here in alphabetical order.
 
 .. function:: super([type[, object-or-type]])
 
-   Return a proxy object that delegates method calls to a parent or sibling
-   class of *type*.  This is useful for accessing inherited methods that have
+   Return a proxy object that delegates attribute accesses to a parent or sibling
+   class of *type*.  This is useful for accessing inherited attributes that have
    been overridden in a class.
 
    The *object-or-type* determines the :term:`method resolution order`
--- a/Doc/library/functions.rst
+++ b/Doc/library/functions.rst
@@ -1594,7 +1594,7 @@ are always available.  They are listed here in alphabetical order.
 .. function:: super([type[, object-or-type]])
 
    Return a proxy object that delegates attribute access to a parent or sibling
-   class of *type*.  This is useful for accessing inherited methods (or data attributes) that have
+   class of *type*.  This is useful for accessing inherited methods that have
    been overridden in a class.
 
    The *object-or-type* determines the :term:`method resolution order`
--- a/Doc/library/functions.rst
+++ b/Doc/library/functions.rst
@@ -1594,7 +1594,7 @@ are always available.  They are listed here in alphabetical order.
 .. function:: super([type[, object-or-type]])
 
    Return a proxy object that delegates attribute access to a parent or sibling
-   class of *type*.  This is useful for accessing inherited attributes that have
+   class of *type*.  This is useful for accessing inherited methods (or data attributes) that have
    been overridden in a class.
 
    The *object-or-type* determines the :term:`method resolution order`
--- a/Doc/library/functions.rst
+++ b/Doc/library/functions.rst
@@ -1593,7 +1593,7 @@ are always available.  They are listed here in alphabetical order.
 
 .. function:: super([type[, object-or-type]])
 
-   Return a proxy object that delegates attribute accesses to a parent or sibling
+   Return a proxy object that delegates attribute access to a parent or sibling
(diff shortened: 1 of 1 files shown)


asottile (approved):

Doc/library/functions.rst
```diff
@@ -1593,8 +1593,8 @@ are always available.  They are listed here in alphabetical order.
 
 .. function:: super([type[, object-or-type]])
 
-   Return a proxy object that delegates method calls to a parent or sibling
-   class of *type*.  This is useful for accessing inherited methods that have
+   Return a proxy object that delegates attribute accesses to a parent or sibling
```
freddrake: I'd be inclined to just say "attribute access" instead of "attribute accesses".

geryogam: Done. Thanks @freddrake.

rhettinger (changes requested):
I don't want to do this.  Delegating methods is the primary use case for *super()*, perhaps by a factor of over a hundred to one.

We can add "(or attributes)" but should keep the wording focused on methods.

freddrake (approved):

Doc/library/functions.rst
```diff
@@ -1593,8 +1593,8 @@ are always available.  They are listed here in alphabetical order.
 
 .. function:: super([type[, object-or-type]])
 
-   Return a proxy object that delegates method calls to a parent or sibling
-   class of *type*.  This is useful for accessing inherited methods that have
+   Return a proxy object that delegates attribute access to a parent or sibling
```
rhettinger, line 1596: This one needs to be restored as well.

rhettinger (changes requested):

Doc/library/functions.rst
maggyero: The 1st sentence is the general definition and the 2nd sentence is the particular usage. So if we restrict the general definition to methods but widen the particular usage to attributes, isn't it weird? What about doing the opposite: general definition about attributes, particular application about methods? Since you want to stress that `super()` is most often _used_ for methods, it seems to be the most sensible. I have made the edit, please tell me what you think.

geryogam: @rhettinger  The 1st sentence is the general definition and the 2nd sentence is the particular usage. So if we restrict the general definition to methods but widen the particular usage to attributes, isn't it weird? What about doing the opposite: the general definition about attributes, the particular usage about methods? Since you want to stress that `super()` is most often _used_ for methods, it seems to be the most sensible. I have made an edit, tell me what you think.

freddrake: I think this should remain.  The general facility provided by ``super`` is centered on attribute resolution; the fact that it's most commonly used for methods is more about the usage pattern.

rhettinger: The only known example of this is exotic, weird, and arguably an anti-pattern:

```
>>> class A:
...     x = True
... 
>>> class B(A):
...     x = False
... 
>>> B().x
False
>>> super(B, B()).x
True
```

I'm now thinking that the core description should be left as-is and that we could add a sentence at the bottom mentioning that attribute access is also possible.

geryogam: @freddrake What do you think of @rhettinger's suggestion? I am not very satisfied with it, but it is always better than not even mentioning that attribute access is possible. And @rhettinger is the expert so he should have the last word.

freddrake: The example shown is clearly not realistic, though it does work that way.

It's far more interesting when implementing an attribute using a descriptor that delegates a portion of the work to an implementation provided by a base class.  I've found this useful on a number of occasions, though I don't think any of that code is public.

While the motivation for ``super`` was definitely it's use in extending methods, the mechanism used to implement it really has nothing to do with the type of the attribute value.

rhettinger: Presumably there is a reason that MRO stands for method resolution order even though it can be used for attributes as well :-)

I don't think this patch should go forward as is.  Readers are best served by emphasizing the typical and intended use cases.  Python's *super()* is weird and atypical compared to other languages.  People read this doc section to learn how to use it. They are critically dependent on these docs. Accordingly, any change for a rare use case should be in a single, separate sentence in the last paragraph.

These docs have a long history of being confusing and their present state is the result of many iterations.  It is easy to make them worse.

rhettinger:
See the alternative PR at https://github.com/python/cpython/pull/16368 which adds a sentence to the section on the various ways to use *super()*.

When people are first taught about *super()* (something I do frequently), it isn't a good idea to lead-off the discussion with attribute access.  If user's first experiment is with attribute access, their odds of success will be low and it will distract from learning the primary, intended use of the tool.

rhettinger (changes requested):
Let's not merge this until the alternative PR has been discussed.

willingc:
Thanks @maggyero for this PR and your efforts on documentation.

After reading this PR and #16368, I believe #16368 covers the most common usage and mentions the possibility of attributes as well. I recommend merging #16368 since I believe it gives more clarity to the reader.

willingc (commented):
I recommend merging #16368 and closing this PR. Thanks all.

maggyero:
Thanks @rhettinger for the review and the alternative PR, I am fine with it. Can I close this one?

Doc/library/functions.rst
geryogam: Alright, that makes sense.

PR #16335, opened 2019-09-23
bpo-38234: Fix _PyConfig_InitPathConfig()

vstinner:
* _PyConfig_InitPathConfig() now starts by copying the global path
  configuration, and then override values set in PyConfig.
* _PyPathConfig_Calculate() implementations no longer override
  _PyPathConfig fields which are already computed. For example,
  if _PyPathConfig.prefix is not NULL, leave it unchanged.
* If Py_SetPath() has been called, _PyConfig_InitPathConfig() doesn't
  _PyPathConfig_Calculate() anymore.
* _PyPathConfig_Calculate() now only uses PyConfig to initialize
  PyCalculatePath structure.
* pathconfig_calculate(): remove useless temporary
  "_PyPathConfig new_config" variable.
* calculate_module_search_path(): remove hack to workaround memory
  allocation failure, call Py_FatalError() instead.
* Fix get_program_full_path(): handle memory allocation failure.

### DIFF
--- a/Python/pathconfig.c
+++ b/Python/pathconfig.c
@@ -51,12 +51,12 @@ pathconfig_clear(_PyPathConfig *config)
         ATTR = NULL; \
     } while (0)
 
-    CLEAR(config->prefix);
     CLEAR(config->program_full_path);
+    CLEAR(config->prefix);
     CLEAR(config->exec_prefix);
     CLEAR(config->module_search_path);
-    CLEAR(config->home);
     CLEAR(config->program_name);
+    CLEAR(config->home);
     CLEAR(config->base_executable);
 #undef CLEAR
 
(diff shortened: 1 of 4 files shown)


Python/pathconfig.c
```diff
@@ -283,6 +246,54 @@ config_init_module_search_paths(PyConfig *config, _PyPathConfig *pathconfig)
 }
 
 
+/* Calculate the path configuration:
+
+   - base_executable (Windows only)
```
zooba: Might be worth noting that it is used on all platforms, but only requires calculation at this point for Windows (which I think is what you're implying?)

vstinner: I completed the comment.

vstinner:
> If Py_SetPath() has been called, _PyConfig_InitPathConfig() doesn't _PyPathConfig_Calculate() anymore.

_PyConfig_InitPathConfig() doesn't **call** _PyPathConfig_Calculate() anymore.

PR #16337, opened 2019-09-23
bpo-38260: Add Docs on asyncio.run

eamanu:
Add docs about return and raise exception on asyncio.run

https://bugs.python.org/issue38260

### DIFF
--- a/Doc/library/asyncio-task.rst
+++ b/Doc/library/asyncio-task.rst
@@ -225,10 +225,9 @@ Running an asyncio Program
     the end.  It should be used as a main entry point for asyncio
     programs, and should ideally only be called once.
 
-    Return the `base_events.run_until_complete()` that this return the
-    Future's result, or raise a RuntimeError if `asyncio.run()`is called
-    from a running event loop, or a ValueError if `main` is not a
-    courutine.
+    Return a result of *coro* execution, or raise a RuntimeError if
+    ``asyncio.run()`` is called from a running event loop, or a
+    ValueError if *coro* is not a courutine.
 
     Example::
 
--- a/Doc/library/asyncio-task.rst
+++ b/Doc/library/asyncio-task.rst
@@ -225,9 +225,9 @@ Running an asyncio Program
     the end.  It should be used as a main entry point for asyncio
     programs, and should ideally only be called once.
 
-    Return a result of *coro* execution, or raise a RuntimeError if
-    ``asyncio.run()`` is called from a running event loop, or a
-    ValueError if *coro* is not a courutine.
+    Return a result of *coro* execution, or raise a :exc:`RuntimeError`
+    if ``asyncio.run()`` is called from a running event loop, or a
+    :exc:`ValueError` if *coro* is not a courutine.
 
     Example::
 
--- a/Doc/library/asyncio-task.rst
+++ b/Doc/library/asyncio-task.rst
@@ -225,6 +225,19 @@ Running an asyncio Program
     the end.  It should be used as a main entry point for asyncio
     programs, and should ideally only be called once.
 
+    Return the `base_events.run_until_complete()` that this return the
+    Future's result, or raise a RuntimeError if `asyncio.run()`is called
+    from a running event loop, or a ValueError if `main` is not a
+    courutine.
+
+    Example::
+
+        async def main():
+            await asyncio.sleep(1)
+            print('hello')
+
+        asyncio.run(main())
+
     .. versionadded:: 3.7
 
     .. versionchanged:: 3.9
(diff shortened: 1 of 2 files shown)


Doc/library/asyncio-task.rst
```diff
@@ -225,6 +225,19 @@ Running an asyncio Program
     the end.  It should be used as a main entry point for asyncio
     programs, and should ideally only be called once.
 
+    Return the `base_events.run_until_complete()` that this return the
```
asvetlov: I think we don't need to mention a future and `run_until_complete()` here.
`asyncio.run()` returns a result of *coro* execution, that's it.

eamanu: Hi @asvetlov Thanks for the review. I make the changes

Doc/library/asyncio-task.rst
```diff
@@ -225,6 +225,18 @@ Running an asyncio Program
     the end.  It should be used as a main entry point for asyncio
     programs, and should ideally only be called once.
 
+    Return a result of *coro* execution, or raise a RuntimeError if
```
corona10: RuntimeError should be 
```
:exc:`RuntimeError`
```

Doc/library/asyncio-task.rst
```diff
@@ -225,6 +225,18 @@ Running an asyncio Program
...
+    ``asyncio.run()`` is called from a running event loop, or a
+    ValueError if *coro* is not a courutine.
```
corona10: This also should be
````
:exc:`ValueError`
````

**Review by corona10 — CHANGES_REQUESTED** [review]
I left some comments for you.
Please take a look

File: Doc/library/asyncio-task.rst (hunk 2 above)
Comment (eamanu) [reply Doc/library/asyncio-task.rst]: Thanks I change it

File: Doc/library/asyncio-task.rst (hunk 3 above)
Comment (eamanu) [reply Doc/library/asyncio-task.rst]: Change it too. thanks!

**Review by asvetlov — APPROVED** [review]
Thanks!

File: Doc/library/asyncio-task.rst [hunk 4]
```diff
@@ -225,6 +225,18 @@ Running an asyncio Program
     the end.  It should be used as a main entry point for asyncio
     programs, and should ideally only be called once.
 
+    Return a result of *coro* execution, or raise a :exc:`RuntimeError`
+    if ``asyncio.run()`` is called from a running event loop, or a
+    :exc:`ValueError` if *coro* is not a courutine.
```
aeros, line 230: Thanks for the adding the additional information @eamanu, but I noticed a couple of issues with the section added. 

The more apparent one was a spelling typo, "courutine" should be "coroutine" (misspelled here and in the docstring). Also, it's helpful to link to the glossary using:
```
:term:`coroutine`
```

Also, while this section is helpful (particularly with specifying the `RuntimeError` and `ValueError` scenarios), I think it could be phrased significantly better. I would suggest the following changes:

```
Within the new event loop, *coro* is executed, returning the result. If
``asyncio.run()`` is called within a running event loop, a :exc:`RuntimeError`
is raised. If *coro* is not a coroutine, a :exc:`ValueError` is raised.
```

Personally, I think it's especially important to phrase this section of the documentation as well as possible, since it's likely one of the most read sections for the entire asyncio module.

Let me know what you think @asvetlov and @1st1. If you approve of the suggestions, I'll open a new PR that makes the changes I mentioned above in the docs for `ascynio.run()` and in the docstring as well (since this one was already merged and backported). 

Grammar tip: When chaining together multiple instances of "or", each of the items should be separated by a comma, but only the last one is supposed to have the "or". For example: "item1, item2, or item3" instead of "item1, or item2, or item3". The section added to the documentation in this PR used the latter.

aeros167:
Also in case it wasn't obvious, the test_venv failure is completely unrelated to the PR. This failure has been happening intermittently for the buildbots recently. I believe it recently happened in a PR from @corona10 and he created a bpo issue for it.

asvetlov:
PR with improvement is welcome! @aeros167 your proposals make sense

aeros167:
Sounds good! I'll open one soon.

Doc/library/asyncio-task.rst
1st1: I'm not sure I follow "Within the new event loop, *coro* is executed, returning the result." -- I can't parse it.  I'd keep the original one "Return the result of *coro* execution."

IMO it's not necessary to document RuntimeError and ValueError; "This function cannot be called when another asyncio event loop is running in the same thread." already covers that.

aeros: @1st1 Okay, I'll make the suggested changes in https://github.com/python/cpython/pull/16403.

Edit: Actually, I think specifying that ValueError is raised when something other than a coroutine is passed might be useful (for basic debugging purposes). Do you think that can remain since it's not covered by the existing section?

PR #16339, opened 2019-09-23
[3.8] bpo-38234: Fix Windows _PyPathConfig_Calculate()

vstinner:
On Windows, _PyPathConfig.home is now preferred over PyConfig.home.


vstinner:
I backported the whole PR #16338 instead: PR #16340.

PR #4819, opened 2017-12-12
bpo-30256: pass all BaseProxy arguments through AutoProxy

uSpike:
The AutoProxy method did not accept the "manager_owned" keyword argument which broke when nested AutoProxy objects were created.

https://bugs.python.org/issue30256

PR #16341, opened 2019-09-24
bpo-30256: Add test for nested queues

finefoot:
I stumbled upon this problem: https://stackoverflow.com/questions/56716470/python-multiprocessing-nested-shared-objects-doesnt-work-with-queue

Right now, Python docs state

> Shared objects are capable of being nested.  For example, a shared container object such as a shared list can contain other shared objects which will all be managed and synchronized by the `SyncManager`.

Source: https://docs.python.org/3/library/multiprocessing.html

These changes have been introduced with https://hg.python.org/cpython/rev/39e7307f9aee along with some tests. However, the tests only use nested `ListProxy` and `DictProxy`.

Because a test for `AutoProxy` is missing, the mentioned issue from the StackOverflow post remained undetected. This pull request adds a test for `AutoProxy[Queue]` for a queue nested inside a list and inside a dict.

Hence, the test added by this pull request will currently fail due to the existing bug regarding nested `AutoProxy` which already has a fix in #4819. After applying the patch from #4819, this test will succeed. So this pull request should be merged together with #4819.

https://bugs.python.org/issue30256

https://bugs.python.org/issue30256


finefoot:
@brandtbucher Thanks for adding the labels. :) Do you mind adding the "skip news" label to this pull request and the bugfix one #4819, too? It's the only CI check missing from #4819 and as far as I understand, a news entry for that small patch is not necessary. The failed CI check might be the only thing missing from getting #4819 merged.

brandtbucher (changes requested):
Hi @finefoot! Thanks and welcome to CPython! :sunglasses:

I think that this PR (and certainly the other one) still need NEWS entries. "Trivial" changes and "small" changes don't always mean the same thing, and my personal litmus test is whether or not the changes touch live code. A NEWS entry only takes [seconds to write](https://blurb-it.herokuapp.com/), but can be extremely valuable when digging through changes or debugging regressions later. It seems, at least in my experience, that first-timers are usually a bit *too* hesitant to include them.

I'll take a look over at #4819 and see what the holdup is.

brandtbucher (approved):

finefoot:
Thanks for the response :) Sorry, you're right - I had a look at only the most recent pull request which introduced a change to `_test_multiprocessing.py` which doesn't have a NEWS entry and makes use of the "skip news" label. I should have also looked at a few old pull requests which all *do* have a NEWS entry. ;) So https://github.com/python/cpython/pull/16341/commits/3f6cd1549500b761a657d0f49e1551b6a9b5437d should fix that now, I think.

pitrou:
Sorry for the delay. @finefoot is it possible for you to integrate the #4819 fix in this PR? It will make it easier to review and merge.

finefoot:
@pitrou I thought, I'd just adopt the changes directly from master once #4819 got merged. (To show the tests will pass, then.) But you want me to add the changes to managers.py with an individual commit before that, yes?

pitrou:
The fix for a bug and the tests for it should be logically contained in a single PR, so I'd like this PR (or the other one) to contain both.

finefoot:
@pitrou :+1: Done, I've merged the changes from #4819 and the tests just passed.

finefoot:
Okay, when I said the tests passed, I was just looking at Travis CI. I'm not sure why Codecov fails... :thinking: Just let me know, if I should do or add anything else. :+1:

finefoot:
@pitrou Friendly ping since it's been 16 days :) Like I said in the previous comment, just let me know, if I should do or add anything else - including but not exclusive to the failed codecov check.

pitrou:
@finefoot You can leave the codecov failure alone, it's not relevant. The patch is in my todo list, it's just that I need to familiarize a bit with the code around that feature.

finefoot:
@pitrou Perfect, thank you :) I didn't want to hurry you regarding the patch itself - I just wanted to make sure that there's nothing I can do.

applio:
I first left his comment on #4819 but I perhaps should have added it here instead:

When testing this fix, this post on StackOverflow suggests they experienced a seg fault: https://stackoverflow.com/questions/56716470/multiprocessing-manager-nested-shared-objects-doesnt-work-with-queue

Has this been investigated and resolved?

finefoot:
@applio The post on Stack Overflow that you're linking to is exactly the one from my original post. :wink: If I understand correctly, the seg faults only appeared when using a different solution that has been referred to in the comments of that Stack Overflow post, not the solution from #4891.

finefoot:
> Note that the solution I referenced in comments in @Darkonaut's answer produced seg faults in my own testing, but this solution did not.

From https://stackoverflow.com/a/58668088

uSpike:
Thank you @finefoot for creating this MR with additional tests.  As I mentioned in #4819 and @finefoot noted above, it seems to me that the noted stackoverflow comment was not suggesting that *this* change causes crashes.  And, two other commenters in #4819  including myself are unable to reproduce it.  It'd be really, really, really nice to have this in 3.10.

gvanrossum (approved):
LGTM. I'm not sure how the test relates to the fix, but I'm too lazy to run all the tests and I trust this. I'll add a `Co-Authored-By: uSpike` header since the patch is @uSpike's.

gvanrossum:
I'll close and reopen this PR to force the tests to re-run, otherwise I can't merge it.

gvanrossum:
...Aaaaand we're back!

gvanrossum:
- What I should add:
```
Co-Authored-By:  Jordan Speicher <uSpike@users.noreply.github.com>
```
- Waiting for tests to pass.
- Hopefully I can merge despite the failing codecov test.

PR #16346, opened 2019-09-24
closes bpo-38174: Update vendored expat library to 2.2.8.

benjaminp:
Fixes CVE-2019-15903. See full changelog at https://github.com/libexpat/libexpat/blob/R_2_2_8/expat/Changes.

https://bugs.python.org/issue38174


vstinner (changes requested):
Using my https://github.com/vstinner/misc/blob/master/cpython/cpython_rebuild_expat_dir.sh script (modify TAG= at the first line to put: TAG=R_2_2_8), I get a very similar result, except that my script also removed Modules/expat/loadlibrary.c. loadlibrary.c has been removed from libexpat: https://github.com/libexpat/libexpat/commit/5d34a708e752ccea53fc6f97bf7a14933587d8b9

"replace loadlibrary.c and RtlGenRandom call with rand_s() on windows"

benjaminp:

vstinner (approved):
LGTM, thanks.

PR #16355, opened 2019-09-24
bpo-13743: Add some documentation strings to xml.dom.minidom

alexitkes:
Add  documentation strings to some methods of the `xml.dom.minidom.Element` class. It would be possibly better to be able to know some small but  important features of these methods directly from `pydoc`. The `Element.getAttribute` method returns an empty string both if there  is no such attribute and if an empty string is an explicitly given attribute value. It never returns None. The `Element.getElementsByTagName` method returns all descendants, not direct children only.

https://bugs.python.org/issue13743

### DIFF
--- a/Lib/xml/dom/minidom.py
+++ b/Lib/xml/dom/minidom.py
@@ -718,6 +718,14 @@ class Element(Node):
         Node.unlink(self)
 
     def getAttribute(self, attname):
+        """Returns the value of the specified attribute.
+
+        Returns the value of the element's attribute named attname as
+        a string. An empty string is returned if the element does not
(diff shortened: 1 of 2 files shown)


corona10:
Thanks for the PR @alexitkes
Please add a whats news through blurb tools.

You can follow the guide via this link. 
https://devguide.python.org/committing/#what-s-new-and-news-entries

corona10:
@alexitkes 
Please leave an issue on https://bugs.python.org also :)

p.s
I can find a similar issue on https://bugs.python.org  
https://bugs.python.org/issue13743.

PR #16360, opened 2019-09-24
bpo-38267: Add thread timeout parameter to `loop.shutdown_default_executor()`

aeros:
/cc @1st1 @asvetlov 

https://bugs.python.org/issue38267


aeros167:
Adding documentation changes and Misc/NEWS next. Before doing so, I wanted to make sure the implementation itself was correct.

Lib/asyncio/base_events.py
```diff
@@ -549,7 +551,7 @@ def _asyncgen_firstiter_hook(self, agen):
                     'asyncgen': agen
                 })
 
-    async def shutdown_default_executor(self):
+    async def shutdown_default_executor(self, timeout=THREAD_JOIN_TIMEOUT):
```
aeros: @1st1 Which option would you prefer?

1) In the current form, with the constant as a default parameter for *timeout*.

2) In another form, having the default for *timeout* as `None`, and passing the constant as an argument when it's called from `asyncio.run()`.

Lib/asyncio/base_events.py
```diff
@@ -66,6 +66,8 @@
 # Maximum timeout passed to select to avoid OS limitations
 MAXIMUM_SELECT_TIMEOUT = 24 * 3600
 
+# Default timeout for joining each thread in `shutdown_default_executor()`
+THREAD_JOIN_TIMEOUT = 300
```
aeros: Let me know if this name is suitable, or if anyone has a better suggestion.

Lib/asyncio/base_events.py
1st1: I'd have the default set to `None` and fix `asyncio.run()` to pass `THREAD_JOIN_TIMEOUT`.  This matches the default behaviour of `thread.join()` -- i.e. wait indefinitely.

Lib/asyncio/base_events.py
```diff
@@ -549,7 +551,7 @@ def _asyncgen_firstiter_hook(self, agen):
                     'asyncgen': agen
                 })
 
-    async def shutdown_default_executor(self):
+    async def shutdown_default_executor(self, timeout=THREAD_JOIN_TIMEOUT):
         """Schedule the shutdown of the default executor."""
```
1st1: You'll need to update the docstring and documentation of this method & asyncio.run to reflect the change.

aeros: Will do, just wanted to verify the implementation details first. I'll apply your suggestion and then update the docstring and docs accordingly.

Lib/asyncio/base_events.py
aeros: I'll move the constant over to `runners.py` then. If we're using it only within `asynio.run()`, it wouldn't make sense to have it within `base_events.py`.

Lib/asyncio/base_events.py
aeros: While updating the docstrings, I realized that the docstring for `asyncio.run()` wasn't updated to reflect the addition of `shutdown_default_executor()`. I'll address that as well, it was only a small change.

Lib/asyncio/runners.py
```diff
@@ -21,6 +25,10 @@ def run(main, *, debug=False):
     It should be used as a main entry point for asyncio programs, and should
     ideally only be called once.
 
+    Each thread within the threadpool is given a timeout duration of 5
+    minutes. If the thread hasn't finishing joining within that duration,
```
aeros: @1st1 While writing the documentation, I was thinking that @asvetlov's suggestion to add *timeout* as parameter for ``asyncio.run()`` might be beneficial. Although it would add an additional knob for users to adjust, forcing them to strictly use a 5 minute timeout duration seems a bit restrictive.

Do you think it might be useful to give users the option to adjust the timeout duration by adding a *timeout* parameter to `asynio.run()`? Since you regularly interact with daily users of asyncio, you might have a better idea of whether or not that would be useful.

1st1: I think that adding knobs is only justified when there's request for that from users. In this particular case it's a bit tricky.

The way I think about it is that it's easy to implement whatever timeout is necessary with the following pattern:

```python
asyncio.run(main())

# versus

async def runner():
  try:
    return await main()
  finally:
    loop = asyncio.get_running_loop()
    await loop.shutdown_default_executor(3600)

asyncio.run(runner())
```

The argument for this pattern is that in more or less advanced asyncio applications there's always a "runner()" like in the above.  It usually sets up signal handlers, manages global state, etc.  `asyncio.run()` in those cases is simply providing a set of reasonable defaults on how you run your event loop and how you shut it down.  And 5 minutes is a reasonable default in this case, I guess.

Long story short, I'd go with this design and learn from our users if we need to make it configurable.

This is just my opinion on this and I'd love to hear what @asvetlov thinks about it.

aeros: > The argument for this pattern is that in more or less advanced asyncio applications there's always a "runner()" like in the above.

Ah, I actually hadn't thought as much about how relatively simple it would be for the users to create their own runner. I wonder if it would be worthwhile to document this pattern somewhere in `asyncio-task.rst`, for those who desire further customization.

>  And 5 minutes is a reasonable default in this case, I guess.

Yeah 5 minutes seems reasonable for most purposes. If it's an issue, we can easily adjust it by modifying the constant.

> Long story short, I'd go with this design and learn from our users if we need to make it configurable.

Good point. It might be a better idea to leave out the parameter initially and then add it in later if the users request it. It would be easy to add the new parameter to the API for `asyncio.run()` if it's desired, but it's incredibly difficult to remove parameters in the future.

> This is just my opinion on this and I'd love to hear what @asvetlov thinks about it.

Personally I have no issue with either of the options, I think both of them would be viable. After reading your latest response, I'm leaning more in favor of keeping the current iteration though.

The more I've been involved with API design, I've increasingly come to realize that there's frequently some degree of disconnect between what the users actually want and what the developers *think* the users want. It can definitely be worthwhile to start with a minimalist approach and add features based on user feedback, rather than making assumptions. Making too many assumptions can easily lead to feature bloat.

aeros167:
I can work on resolving the conflicts later tonight or tomorrow after someone gets the chance to look over https://github.com/python/cpython/pull/16403. The conflicts were probably created from the recently merged https://github.com/python/cpython/pull/16337. They should be fairly straightforward to resolve though.

aeros167:
@1st1 I resolved all of the merge conflicts. I think we were just waiting on feedback from @asvetlov about his thoughts on adding a *timeout*  parameter to `asyncio.run()` for adjusting the thread timeout duration. He initially proposed it in https://github.com/python/cpython/pull/16284#issuecomment-533703595.

My position on it currently is that it can be addressed later if enough users request the feature. As we discussed earlier, adding undesired extra knobs to tweak in the API can easily result in feature bloat. It's much less costly to add a requested option at a later point than it is to remove unused features, due to backwards compatibility concerns and deprecation periods.

Are we ready to move forward with this PR or would you like to wait for Andrew's feedback first?

aeros167:
@1st1 I just noticed an important concern that we may have overlooked (or at least I did). Currently, the *timeout* parameter for shutdown_default_executor() is specifying the total duration for *all* of the threads to close, not each one:
```python
    async def shutdown_default_executor(self, timeout=None):
        self._executor_shutdown_called = True
        if self._default_executor is None:
            return
        future = self.create_future()
        thread = threading.Thread(target=self._do_shutdown, args=(future,))
        thread.start()
        try:
            await future
        finally:
            #the timeout for this thread is for shutting down all threads in the thread pool
            thread.join(timeout)

    def _do_shutdown(self, future):
        try:
            self._default_executor.shutdown(wait=True)
            self.call_soon_threadsafe(future.set_result, None)
        except Exception as ex:
            self.call_soon_threadsafe(future.set_exception, ex)
```

In order for it to be for *each* thread, this would have to be implemented at a much lower level, by joining each thread in `self._executor._threads`, effectively re-implementing `ThreadPoolExecutor.shutdown()`. Alternatively `ThreadPoolExecutor.shutdown()` would have to be provided a timeout parameter:
```
    async def shutdown_default_executor(self, timeout=None):
        self._executor_shutdown_called = True
        if self._default_executor is None:
            return
        future = self.create_future()
        thread = threading.Thread(target=self._do_shutdown, args=(future,timeout))
        thread.start()
        try:
            await future
        finally:
            thread.join()

    def _do_shutdown(self, future, timeout):
        try:
            self._default_executor.shutdown(wait=True, timeout)
            self.call_soon_threadsafe(future.set_result, None)
        except Exception as ex:
            self.call_soon_threadsafe(future.set_exception, ex)
```
Here would be the modified `ThreadPoolExecutor.shutdown()`:
```
    def shutdown(self, wait=True, timeout=None):
        with self._shutdown_lock:
            self._shutdown = True
            self._work_queue.put(None)
        if wait:
            for t in self._threads:
                t.join(timeout)
```

The other option is that we change the documentation to specify that the *timeout* parameter is for all of the threads rather than for each one. We would need to adjust the constant _THREAD_JOIN_TIMEOUT accordingly though. I don't think the current amount of 600 would as be reasonable for waiting on all of the threads to join. Let me know what you think @1st1 and @asvetlov.

1st1:
> Currently, the timeout parameter for shutdown_default_executor() is specifying the total duration for all of the threads to close, not each one:

I think that's how it's supposed to be.

> In order for it to be for each thread

What's the purpose of that?  Say you have 10000 threads -- do you really want tot wait 10000 * 5 seconds instead of waiting just 5 to try joining them in parallel?

aeros167:
@1st1 
> I think that's how it's supposed to be.

Ah, I see. The documentation I wrote had suggested it was for each thread, rather than all of the threads:

> The *timeout* parameter specifies the amount of time each thread will be
   given to finish joining.

> What's the purpose of that?

I think it might be more straightforward from a design standpoint to have a specific timeout per-thread. Also, any constant we provide could potentially be invalidated if a user wants to spawn a large number of threads than intended, or if those threads happen to be assigned an activity that takes a longer period of time to resolve. If the constant was the amount of time allocated for *each* thread, I think it would scale significantly better.

However, I think that might be too might of a hypothetical justify implementing that. I'd say that if we start having issues from users with the existing constant duration used in `asyncio.run()`, the options would be to up the amount or making it per-thread.

The reason I made the post though was primarily because the existing documentation I created suggested it was for each thread, and that the existing constant duration of 600 seconds might be a bit on the small side for general purpose `asyncio.run()` cases. For now I'll just adjust the documentation, let me know if you think the constant should remain at 600 or be increased.

1st1:
Do we really have it configured to wait for 600 seconds? Wasn't it 5 seconds?

> I think it might be more straightforward from a design standpoint to have a specific timeout per-thread.

How?  It's a threadpool. From the standpoint of the user all threads in the pool are equal.

I also think you're confused here.  When you join a threadpool with an N seconds timeout it will simultaneously join all of its threads, giving all of them N seconds.  The only difference is if we join all threads simultaneously or join them one by one.  Doing the latter isn't a commonly accepted practice, and in reality is just wasteful.

> Each thread within the threadpool is given a timeout duration of 5
> minutes. If the thread hasn't finishing joining within that duration,
> the lock is released and the thread is terminated.

@asvetlov will Python actually terminate a thread if we try to join it and it hits a timeout?  IIRC the thread will simply continue to run, no?

aeros167:
@1st1 
> Do we really have it configured to wait for 600 seconds? Wasn't it 5 seconds?

Oops, 600 was a typo, I mean 300 seconds/5 minutes. In the [original comment](https://github.com/python/cpython/pull/16284#issuecomment-533731019), you had suggested 5 minutes:

> We can have a global asyncio "reasonable" timeout for shutting down threads -- say 5 minutes

> will Python actually terminate a thread if we try to join it and it hits a timeout

No, not from thread.join(timeout), I was just looking over the implementation yesterday. When timeout is not None for `thread.join()`, the operating stops once the duration has released, but the thread remains alive. I had misunderstood this part when I initially opened the PR.

Instead, we would have to do something like this, to shutdown the `ThreadPoolExecutor` when the thread did not join within the timeout duration. I think a warning of some form should be emitted, but that's optional:

```python
    async def shutdown_default_executor(self, timeout=None):
        self._executor_shutdown_called = True
        if self._default_executor is None:
            return
        try:
            await future
        finally:
            thread.join(timeout)
        # thread join timed out
        if (thread.is_alive()):
            # emit warning?
            self._default_executor.shutdown(wait=False)
```

aeros167:
@1st1:
> I also think you're confused here. When you join a threadpool with an N seconds timeout it will simultaneously join all of its threads, giving all of them N seconds.

I was under the impression that the threads were being joined *concurrently* rather than in *parallel*. Meaning that there's cooperative multitasking being done with overlapping time periods from the threads joining, but it is not done simultaneously in true parallelism. Am I misunderstanding something?

1st1:
> I was under the impression that the threads were being joined concurrently rather than in parallel. 

Oh well, in CPython it's grey area :)  Most of the time all threads work cooperatively, i.e. that's how the interpreter executes them.  The GIL does serializes how opcodes are executed forcing threaded code to cooperate.  But thinking about threads in Python in that way isn't productive.  Just like it's not really productive to think about parallel vs concurrent threading in languages without the GIL -- ideally your code should be written in such a way that it doesn't care (i.e. access to shared data is properly guarded and serialized).

Trying to give all threads an equal opportunity to finalize & close isn't the job of asyncio code though.  The GIL is a limitation that imposes certain restrictions on how threads work in CPython, but we shouldn't care.  We should assume that threads in Python work in parallel otherwise it would be insanely hard to reason about them.  Moreover, what if you write code that assumes that threads in CPython run cooperatively and then we ship CPython without GIL? :)  [that won't happen anytime soon]

Therefore I see no need for a loop like `[th.close() for th in executor.threads]`.  Simply waiting on `executor.join()` is enough.

I hope I made myself clear here, feel free to ask more questions.

aeros167:
@1st1 
>  But thinking about threads in Python in that way isn't productive.

Ah, I might have had the difference between concurrency vs parallelism a bit excessively drilled into my head then over the last year. It's been a strong area of focus for me recently and I've tried to ensure that I don't mix up the two areas, but perhaps it's been to a fault.

> We should assume that threads in Python work in parallel otherwise it would be insanely hard to reason about them.

I can definitely understand this point from the perspective of structuring the code, but from the real-time perspective of the actual amount of time it will take to finish joining the threads, doesn't the amount of time scale with each additional thread that needs to be joined? 

I.E. if you have a thread X that takes N amount of time to finishing joining, increasing the count of X to join in the thread pool will increase N. From my understanding, it's not exactly linear, but the total amount of time does increase with each additional thread to join in the current implementation of the GIL. But perhaps this is an unneeded layer of complexity that most users wouldn't consider when specifying a timeout, so from an API perspective it makes more sense for the timeout to be for the thread pool?

I think we should think about it somewhat for creating a reasonable default timeout constant though. I'm not sure if the currently 300 seconds is enough, but if you think it will fit most general use cases for `asyncio.run()` and we emit some form of warning when it times out it should be fine. This would make it easier for easier to realize that they may have to use their own custom implementation for edge cases. 

>  Moreover, what if you write code that assumes that threads in CPython run cooperatively and then we ship CPython without GIL? :) [that won't happen anytime soon]

I certainly hope that will happen some day! It's actually been a goal of mine to be able to contribute towards that effort, believe it or not. I still have a lot to learn about the GIL and the underlying interpreter implementation before I could even *consider* trying to help make the GIL unneeded, but I hope that I'm knowledgeable enough to do so at some point in time. (:

> I hope I made myself clear here, feel free to ask more questions.

Thanks, this is a fairly complex area and there's definitely a lot that I have to learn. I think that I understand it from an abstract perspective, but sometimes it can be difficult to separate details from the categories of:

1) Productive - what developers/users of asyncio are/should be thinking about

2) Underlying implementation - how it actually works, such as GIL limitations

3) Theoretical - abstract ideas, such as concurrency vs parallelism

Optimally while developing asyncio, I should have an understanding of 2 and 3 while providing an API that focuses on 1. But it's certainly easier said than done sometimes. Do you typically have a strategy for not getting lost in the details and focusing on the productive user experience?

Doc/library/asyncio-eventloop.rst
```diff
@@ -167,13 +167,20 @@ Running and stopping the loop
...
    the threads in the :class:`ThreadPoolExecutor`. After calling this method, a
    :exc:`RuntimeError` will be raised if :meth:`loop.run_in_executor` is called
    while using the default executor.
 
+   The *timeout* parameter specifies the amount of time the threadpool will
+   be given to finish joining. The default value is ``None``, which means the
+   threadpool will be given an indefinite amount of time.
+
+   If the timeout duration is reached, a warning is emitted and threadpool is
+   terminated without waiting for its threads to finish joining.
```
aeros, line 182: Working on implementing this next.

1st1:
> Do you typically have a strategy for not getting lost in the details and focusing on the productive user experience?

No strategy, just experience of dealing with code and people. But that's highly subjective and thus I can be wrong. That's why you always want to gather opinions from multiple engineers. On the case is asyncio those multiple engineers is @asvetlov :)

Lib/asyncio/base_events.py
```diff
@@ -560,7 +565,12 @@ def _asyncgen_firstiter_hook(self, agen):
         try:
             await future
         finally:
-            thread.join()
+            thread.join(timeout)
+
+        if (thread.is_alive()):
+            warnings.Warning("The ThreadPoolExecutor did not finishing joining"
```
1st1: It it `Warning` or `warning`?

aeros: @1st1 I believe `Warning` is correct https://docs.python.org/3/library/exceptions.html#Warning.

1st1: Well, yes, you construct a Warning instance, but you don't send it anywhere. Usually we use the https://docs.python.org/3/library/warnings.html#warnings.warn function or similar.

aeros: @1st1

Oops. Thanks for the clarification, I was thinking along the lines of ``raise Warning(msg)`` and somehow ended up with ``warnings.Warning(msg)``. Using `warnings.warn()` makes far more sense. I know I saw it used multiple times while I was reverting the 3.8 API, but I got the syntax mixed up in my head. Warnings are definitely not something that I've utilized in my own code regularly, I typically stick to exceptions or logging.

aeros167: @1st1 Couple of questions regarding the warning:

1) Should it be a `RuntimeWarning`? I think this would be suitable since it occurs during runtime as a result of the thread not being able to close within the specified timeout window. I could imagine an argument for it being a `ResourceWarning` as well though.

2) Should I add a test to `test_asyncio/test_events.py` to make sure the the warning is triggered in an appropriate scenario? I have some familiarity with the existing tests, but it would be my first time writing a new one.

Lib/asyncio/base_events.py
```diff
@@ -560,7 +565,13 @@ def _asyncgen_firstiter_hook(self, agen):
         try:
             await future
         finally:
-            thread.join()
+            thread.join(timeout)
+
+        if (thread.is_alive()):
+            warnings.warn("The ThreadPoolExecutor did not finishing joining"
+                             f"its threads within {timeout} seconds.",
+                             RuntimeWarning)
```
aeros: @1st1 Couple of questions regarding the warning:

1) Should it be a `RuntimeWarning` or `ResourceWarning`?

2) ~Should I add a test to `test_asyncio/test_base_events.py` to make sure the the warning is triggered in an appropriate scenario?~

Edit: Regarding the test, I don't think there's a way to test for this *without* creating a dangling thread due to the nature of the warning. Personally, I don't think it's worth adding an intentional resource leak to the asyncio regression tests just to test the functionality of a single warning. If there's another way to test for the warning without creating a dangling thread, let me know.

aeros:
@1st1 Is there anything that needs to be changed in this PR?

Lib/asyncio/base_events.py
```diff
@@ -560,7 +565,13 @@ async def shutdown_default_executor(self):
         try:
             await future
         finally:
-            thread.join()
+            thread.join(timeout)
+
+        if (thread.is_alive()):
```
gvanrossum:
```suggestion
        if thread.is_alive():
```

Doc/library/asyncio-task.rst
```diff
@@ -235,6 +239,7 @@ Running an asyncio Program
 
         asyncio.run(main())
 
+
```
gvanrossum: I don't think we need this blank line.

```suggestion
```

gvanrossum (approved):
LGTM except for two style nits. I'll try to merge my suggestions.

gvanrossum:
@kumaraditya303 Do you have time to fix the merge conflicts?

kumaraditya303:
> Do you have time to fix the merge conflicts?

I don't have permission to push to this PR so I created https://github.com/python/cpython/pull/97561.

PR #16363, opened 2019-09-24
bpo-13153: Test IDLE Windows no-color astral paste fix

terryjreedy:


terryjreedy:
Something happened that the _tkinter changes disappeared.

PR #16368, opened 2019-09-25
bpo-38255: super() can search attributes as well as methods

rhettinger:
https://bugs.python.org/issue38255


eamanu (approved):
I prefer this PR instead of https://github.com/python/cpython/pull/16331/files. I think is better mention that we can use super() for attribute lookups, and not just change word `attribute` for `methods`. I think that could be confuse.

willingc (approved):

geryogam (approved):
Thanks @rhettinger, I have closed my PR. You can merge.

PR #16373, opened 2019-09-25
bpo-38243: Escape the server_title of DocXMLRPCServer when rendering

corona10:
Escape the server_title of DocXMLRPCServer when rendering,
The title will only be escaped when the server_title should be rendered.

https://bugs.python.org/issue38243

### DIFF
--- a/Lib/test/test_docxmlrpc.py
+++ b/Lib/test/test_docxmlrpc.py
@@ -194,13 +194,14 @@ class DocXMLRPCHTTPGETServer(unittest.TestCase):
             response.read())
 
     def test_server_title_escape(self):
+        # bpo-38243: Ensure that the server title and documentation
+        # are escaped for HTML.
         self.serv.set_server_title('test_title<script>')
         self.serv.set_server_documentation('test_documentation<script>')
         self.assertEqual('test_title<script>', self.serv.server_title)
         self.assertEqual('test_documentation<script>',
                 self.serv.server_documentation)
 
-        # bpo-38243
         generated = self.serv.generate_html_documentation()
         title = re.search(r'<title>(.+?)</title>', generated).group()
         documentation = re.search(r'<p><tt>(.+?)</tt></p>', generated).group()
--- a/Lib/test/test_docxmlrpc.py
+++ b/Lib/test/test_docxmlrpc.py
@@ -193,14 +193,6 @@ class DocXMLRPCHTTPGETServer(unittest.TestCase):
              b'method_annotation</strong></a>(x: bytes)</dt></dl>'),
             response.read())
 
-
-class XMLRPCDocGeneratorTest(unittest.TestCase):
-    def setUp(self):
-        self.serv = DocXMLRPCServer(("localhost", 0), logRequests=False)
-
-    def tearDown(self):
-        self.serv.server_close()
-
     def test_server_title_escape(self):
         self.serv.set_server_title('test_title<script>')
         self.serv.set_server_documentation('test_documentation<script>')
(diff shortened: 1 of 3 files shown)


Lib/xmlrpc/server.py
```diff
@@ -894,7 +895,7 @@ def generate_html_documentation(self):
                                 methods
                             )
 
-        return documenter.page(self.server_title, documentation)
+        return documenter.page(html.escape(self.server_title), documentation)
```
tirkarthi, line 898: Please add a test for this in `Lib/test/test_docxmlrpc.py` . Test could be to set a script/html content using `self.serv.set_server_title` and ensure the documentation returned as html response from the server has escaped content inside the title tag.

corona10: @tirkarthi  Oh thanks I will add it :)

corona10:
@tirkarthi I've updated the PR, thank you for the feedback.

Lib/test/test_docxmlrpc.py
```diff
@@ -193,5 +194,27 @@ def test_annotations(self):
             response.read())
 
 
+class XMLRPCDocGeneratorTest(unittest.TestCase):
```
tirkarthi: IMO you can just add this test to the other class which does very similar `setUp` and `tearDown` since there is only one test for the class.

corona10:
@tirkarthi I've reflected your comment. Thanks for the review.

Lib/test/test_docxmlrpc.py
```diff
@@ -192,6 +193,20 @@ def test_annotations(self):
...
             response.read())
 
+    def test_server_title_escape(self):
+        self.serv.set_server_title('test_title<script>')
+        self.serv.set_server_documentation('test_documentation<script>')
+        self.assertEqual('test_title<script>', self.serv.server_title)
+        self.assertEqual('test_documentation<script>',
+                self.serv.server_documentation)
+
+        # bpo-38243
```
vstinner: Hum, please elaborate the comment. Something like:

"# bpo-38243: Ensure that the server title and documentation are escaped for HTML."

Would you mind to move the comment at the start of the function?

Misc/NEWS.d/next/Security/2019-09-25-13-21-09.bpo-38243.1pfz24.rst
```diff
@@ -0,0 +1,2 @@
+Escape the server_title of :class:`DocXMLRPCServer` when rendering the
+document page. (Contributed by Dong-hee Na in :issue:`38243`.)
```
vstinner: I suggest:

Escape the server title of :class:`xmlrpc.server.DocXMLRPCServer` when rendering the
document page as HTML. (Contributed by Dong-hee Na in :issue:`38243`.)

Misc/NEWS.d/next/Security/2019-09-25-13-21-09.bpo-38243.1pfz24.rst
```diff
@@ -0,0 +1,2 @@
+Escape the server title of :class:`DocXMLRPCServer` when rendering the
```
vstinner: Sphinx cannot find the class: you have to specify the fully qualified name:

:class:`xmlrpc.server.DocXMLRPCServer`

corona10: @vstinner Oh.. got it.

corona10:
@vstinner Updated!

vstinner:
@corona10: I backported manually your change to 3.6. I just had to fix a simple conflict on imports.

Would you mind to try to backport the change to Python 2.7 which also has the bug?

corona10:
@vstinner Sure!

PR #16382, opened 2019-09-25
bpo-38270: Check for hash digest algorithms and avoid MD5

tiran:
Make it easier to run and test Python on systems with restrict crypto policies:

* add requires_hashdigest to test.support to check if a hash digest algorithm is available and working
* avoid MD5 in test_hmac
* replace MD5 with SHA256 in test_tarfile
* mark network tests that require MD5 for MD5-based digest auth or CRAM-MD5

https://bugs.python.org/issue38270

### DIFF
--- a/Misc/NEWS.d/next/Tests/2019-09-25-12-18-31.bpo-38270._x-9uH.rst
+++ b/Misc/NEWS.d/next/Tests/2019-09-25-12-18-31.bpo-38270._x-9uH.rst
@@ -1,0 +1,4 @@
+test.support now has a helper function to check for availibility of a
+hash digest function. Several tests are refactored avoid MD5 and use
+SHA256 instead. Other tests are marked to use MD5 and skipped when MD5 is
+disabled.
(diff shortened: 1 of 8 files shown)


Misc/NEWS.d/next/Tests/2019-09-25-12-18-31.bpo-38270._x-9uH.rst
```diff
@@ -0,0 +1,3 @@
+test.support now has a helper function to check for availibility of a hash
+digest function. Several tests are refactored to either avoid MD5 or to mark
```
vstinner: You may mention that some tests are updated to use SHA256 rather than MD5.

vstinner (approved):
LGTM. I just have a minor suggestion on the NEWS entry.

Misc/NEWS.d/next/Tests/2019-09-25-12-18-31.bpo-38270._x-9uH.rst
tiran: Done

vstinner (approved):
LGTM. The updated NEWS entry is even better ;-)

PR #16384, opened 2019-09-25
bpo-36046: posix_spawn() doesn't support uid/gid

vstinner:
* subprocess.Popen now longer uses posix_spawn() if uid, gid or gids
  are set.
* test_subprocess: add "nobody" and "nfsnobody" group names for
  test_group().
* test_subprocess: test_user() and test_group() are now also tested
  with close_fds=False.

https://bugs.python.org/issue36046

### DIFF
--- a/Lib/subprocess.py
+++ b/Lib/subprocess.py
@@ -1681,7 +1681,10 @@ class Popen(object):
                     and (p2cread == -1 or p2cread > 2)
                     and (c2pwrite == -1 or c2pwrite > 2)
                     and (errwrite == -1 or errwrite > 2)
-                    and not start_new_session):
+                    and not start_new_session
+                    and gid is None
+                    and gids is None
+                    and uid is None):
                 self._posix_spawn(args, executable, env, restore_signals,
                                   p2cread, p2cwrite,
                                   c2pread, c2pwrite,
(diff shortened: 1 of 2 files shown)


vstinner:
cc @pablogsal @nanjekyejoannah

Yhg1s (approved):
This change looks like the correct fix, and with proper tests for it.

I do wonder if the large conditional to check whether posix_spawn should be used should instead be pulled into the _posix_spawn method, with an appropriate comment explaining what is or isn't supported by posix_spawn, and perhaps why... but this fix should probably go in regardless of that.

vstinner:
> I do wonder if the large conditional to check whether posix_spawn should be used should instead be pulled into the _posix_spawn method, with an appropriate comment explaining what is or isn't supported by posix_spawn, and perhaps why... but this fix should probably go in regardless of that.

Maybe, yes. But having tests in the caller avoids to pass the very long list of arguments. Especially, it avoids to pass arguments which are not supposed by posix_spawn() :-)

_execute_child() has 21 arguments... is that really reasonable? Maybe we should put these stuff into a temporary object?

gpshead:
Thanks for catching this!  There is a lot of worthwhile refactoring that could be done in subprocess.py...

Ideally we'd want to make it so that posix_spawn can only ever be used in very specific circumstances _without a need to remember to add new negative conditionals_ when adding new non-spawn-available features.

Using a temporary object rather than argument soup wouldn't change that.

PR #16385, opened 2019-09-25
bpo-38271: encrypt private key test files with AES256

tiran:
The private keys for test_ssl were encrypted with 3DES in traditional
PKCS#5 format. 3DES and the digest algorithm of PKCS#5 are blocked by
some strict crypto policies. Use PKCS#8 format with AES256 encryption
instead.

https://bugs.python.org/issue38271


Lib/test/make_ssl_certs.py
```diff
@@ -206,8 +206,9 @@ def print_cert(path):
     with open('ssl_key.pem', 'w') as f:
         f.write(key)
     print("password protecting ssl_key.pem in ssl_key.passwd.pem")
-    check_call(['openssl','rsa','-in','ssl_key.pem','-out','ssl_key.passwd.pem','-des3','-passout','pass:somepass'])
-    check_call(['openssl','rsa','-in','ssl_key.pem','-out','keycert.passwd.pem','-des3','-passout','pass:somepass'])
+    # pkey uses modern encryption format
```
encukou: This probably won't be true in 5 years. Perhaps leave this comment in the commit message? It makes more sense as only a justification for this change.

encukou (approved):
Aside from that nitpick, this looks legit

Lib/test/make_ssl_certs.py
tiran: SGTM, the comment is gone.

PR #16386, opened 2019-09-25
bpo-38275: Skip ssl tests for disabled versions

tiran:
test_ssl now handles disabled TLS/SSL versions better. OpenSSL's crypto
policy and run-time settings are recognized and tests for disabled versions
are skipped.

https://bugs.python.org/issue38275


vstinner (approved):
LGTM.

This PR basically only adds helper function to factorize the code in test_ssl.py, but it doesn't seem to change the behavior. So it looks safe to me.

PR #16398, opened 2019-09-25
bpo-38279: use of easier to understand syntax 

BMaxV:
The suggested changes mostly deal with my troubles to understand how long a process "lives" in code, how and when they stop.

The first change is putting the processes in a list, after creating and starting them in separate lines. Should be easier to read.

It was unclear to me, why iter(input.get,"stop") and putting "stop" to the task queue would stop a process. Also after reading up on iter(), it should pass "stop" anyway when input.get() wouldn't work. After removing the last "stop" putting loop, it didn't do that though. So I don't get how that works. Since it's an example I don't think it's too much to ask for it to be obvious. Or more "obvious" anway.

So I replaced that for ... iter(...) with a while that just gets stuff basically the same way, and terminates processes explicitly.

This is safe, as far as I understand it, because the for i in range(len(tasks2)) done_queue.get() will always wait until all tasks are done, because of the defaults for queue.get().

https://bugs.python.org/issue38279

### DIFF
--- a/Doc/includes/mp_workers.py
+++ b/Doc/includes/mp_workers.py
@@ -8,7 +8,8 @@ from multiprocessing import Process, Queue, current_process, freeze_support
 #
 
 def worker(input, output):
-    for func, args in iter(input.get, 'STOP'):
+    while True:
+        func,args=input.get()
         result = calculate(func, args)
         output.put(result)
 
@@ -51,8 +52,11 @@ def test():
         task_queue.put(task)
 
     # Start worker processes
+    processes=[]
     for i in range(NUMBER_OF_PROCESSES):
-        Process(target=worker, args=(task_queue, done_queue)).start()
+        p=Process(target=worker, args=(task_queue, done_queue))
+        p.start()
+        processes.append(p)
 
     # Get and print results
     print('Unordered results:')
@@ -68,8 +72,9 @@ def test():
         print('\t', done_queue.get())
 
     # Tell child processes to stop
-    for i in range(NUMBER_OF_PROCESSES):
-        task_queue.put('STOP')
+    for p in processes:
+        p.terminate()
+        #task_queue.put('STOP')
 
 
 if __name__ == '__main__':


the-knights-who-say-ni:

## CLA Missing

@BMaxV

BMaxV:
Ah. I will sign the thing when I have the time. :)

terryjreedy:
Signing takes about 5 minutes.  It is a potential waste of time for us to review until you do.

BMaxV:
Ok I signed the thing and got the confirmation via mail.

BMaxV:
Yo. It's been nearly three weeks. Should I do something else than signing the thing?

terryjreedy:
I manually removed the Unsigned tag and the bot confirmed the signing.  Davin or Antoine can review it now.

pitrou (changes requested):
Some comments here:
* the changes don't follow PEP 8 (whitespace around operators and assignments)
* sending a "STOP" string to gracefully shutdown a worker process is not the same as brutally killing it by calling `terminate()`; I believe that showcasing how to gracefully end a worker process is more useful to the reader (killing processes is easy but can risk losing data by cancelling critical operations)
* while two-argument `iter` is not very common, it's not that difficult to understand and it is expected that people trying to use `multiprocessing` primitives are able to use the documentation to find the meaning of a built-in function (!)

So, while I understand the intent of making an example simpler to approach, in the end it's not obvious that this PR is actually useful.

pitrou:
By the way, sorry for the delay in reviewing this. However, I'm going to close this PR as I don't believe any of the proposed changes are actually desirable. Feel free to open another PR and/or bug entry if you believe other changes may be useful.

BMaxV:
@pitrou Let me get this straight:

You criticse my PR, to improve documentation, based on functionality that's not documented, even though the page that should document it, is the very same I'm trying to improve?

>sending a "STOP" string to gracefully shutdown a worker process

There is no advice on the doc page to do this, or that it works or how it works.

Then completely without any discussion, you just decide the doc page is perfect and changes are not desirable and you close the issue?

PR #16402, opened 2019-09-25
bpo-38234: Add test_init_setpath_config() to test_embed

vstinner:
* Add test_init_setpath_config() to test_embed: test Py_SetPath()
  with PyConfig.
* test_init_setpath() and test_init_setpythonhome() no longer call
  Py_SetProgramName(), but use the default program name.
* _PyPathConfig: only define isolated, site_import
  and base_executable on Windows.
* If executable is set explicitly in the configuration, ignore
  calculated base_executable: _PyConfig_InitPathConfig() will copy
  executable to base_executable.

### DIFF
--- a/Python/pathconfig.c
+++ b/Python/pathconfig.c
@@ -58,7 +58,10 @@ pathconfig_clear(_PyPathConfig *config)
     CLEAR(config->module_search_path);
     CLEAR(config->program_name);
     CLEAR(config->home);
+#ifdef MS_WINDOWS
     CLEAR(config->base_executable);
+#endif
+
 #undef CLEAR
 
     PyMem_SetAllocator(PYMEM_DOMAIN_RAW, &old_alloc);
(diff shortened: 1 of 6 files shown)


Python/pathconfig.c
```diff
@@ -330,18 +337,32 @@ config_calculate_pathconfig(PyConfig *config)
...
+
     COPY_ATTR(program_full_path, executable);
     COPY_ATTR(prefix, prefix);
     COPY_ATTR(exec_prefix, exec_prefix);
-    COPY_ATTR(base_executable, base_executable);
+
 #undef COPY_ATTR
 
+#ifdef MS_WINDOWS
+    /* If a .pth file is found: isolated and site_import are overriden */
```
zooba: Should be `._pth` file.

zooba:
It's nice that you separate out all the `._pth` related stuff under a build flag for Windows, but there have been requests to add support on other platforms as well (I forget the bpo numbers now, or maybe it was in person at PyCon). So maybe make sure you don't make it too inaccessible or we'll have more work to do if/when we make that happen.

Python/pathconfig.c
vstinner: Oops, fixed. Thanks.

vstinner:
> It's nice that you separate out all the ._pth related stuff under a build flag for Windows, but there have been requests to add support on other platforms as well (I forget the bpo numbers now, or maybe it was in person at PyCon). So maybe make sure you don't make it too inaccessible or we'll have more work to do if/when we make that happen.

_PyPathConfig is an internal structure, so it's just an internal cleanup. It will be easy to enable the code on other platforms.

Right now, I'm trying to document the current implementation and write tests for it. But I would be interested to support also base_executable and `__PYVENV_LAUNCHER__` environment variable in Modules/getpath.c.

PR #16403, opened 2019-09-25
Fix and improve `asyncio.run()` docs

aeros:
This PR is based on the suggestions I made in https://github.com/python/cpython/pull/16337#discussion_r328351505, which were approved of by @asvetlov.

### DIFF
--- a/Doc/library/asyncio-task.rst
+++ b/Doc/library/asyncio-task.rst
@@ -225,8 +225,8 @@ Running an asyncio Program
     the end.  It should be used as a main entry point for asyncio
     programs, and should ideally only be called once.
 
-   Return the result of *coro* execution. If *coro* is not a :term:`coroutine`,
-   a :exc:`ValueError` is raised.
+    Return the result of *coro* execution. If *coro* is not a :term:`coroutine`,
+    a :exc:`ValueError` is raised.
 
 
     Example::
--- a/Doc/library/asyncio-task.rst
+++ b/Doc/library/asyncio-task.rst
@@ -212,7 +212,7 @@ Running an asyncio Program
 
 .. function:: run(coro, \*, debug=False)
 
-    Execute the :term:`coroutine` *coro*  and return the result.
+    Execute the :term:`coroutine` *coro* and return the result.
 
     This function runs the passed coroutine, taking care of
     managing the asyncio event loop, *finalizing asynchronous
--- a/Doc/library/asyncio-task.rst
+++ b/Doc/library/asyncio-task.rst
@@ -212,6 +212,8 @@ Running an asyncio Program
 
 .. function:: run(coro, \*, debug=False)
 
+    Execute the :term:`coroutine` *coro*  and return the result.
+
     This function runs the passed coroutine, taking care of
     managing the asyncio event loop, *finalizing asynchronous
     generators*, and closing the threadpool.
@@ -224,10 +226,6 @@ Running an asyncio Program
     This function always creates a new event loop and closes it at
     the end.  It should be used as a main entry point for asyncio
     programs, and should ideally only be called once.
-
-    Return the result of *coro* execution. If *coro* is not a :term:`coroutine`,
-    a :exc:`ValueError` is raised.
-
 
     Example::
 
--- a/Doc/library/asyncio-task.rst
+++ b/Doc/library/asyncio-task.rst
@@ -225,9 +225,10 @@ Running an asyncio Program
     the end.  It should be used as a main entry point for asyncio
     programs, and should ideally only be called once.
 
-    Return a result of *coro* execution, or raise a :exc:`RuntimeError`
-    if ``asyncio.run()`` is called from a running event loop, or a
-    :exc:`ValueError` if *coro* is not a courutine.
+    Within the new event loop, *coro* is executed, returning the result. If
+    ``asyncio.run()`` is called within a running event loop, a
+    :exc:`RuntimeError` is raised. If *coro* is not a :term:`coroutine`, a
+    :exc:`ValueError` is raised.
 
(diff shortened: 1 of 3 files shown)


aeros167:
@1st1 I think it's useful to explicitly mention that a `ValueError` is raised when something other than a coroutine is passed to *coro*, but you are are correct that the `RuntimeError` is already mentioned in a previous section, so that can definitely be removed.

My main concern is just addressing the typo though, so if you want to remove the `ValueError` as well that's fine. Let me know either way.

1st1 (commented):
changes in runners.py appear to be musindented

aeros167:
@1st1 Done. I fixed the indent in the docs earlier, but forgot to do it in runners.py, good catch.

Doc/library/asyncio-task.rst
```diff
@@ -225,9 +225,9 @@ Running an asyncio Program
     the end.  It should be used as a main entry point for asyncio
     programs, and should ideally only be called once.
 
-    Return a result of *coro* execution, or raise a :exc:`RuntimeError`
-    if ``asyncio.run()`` is called from a running event loop, or a
-    :exc:`ValueError` if *coro* is not a courutine.
+    Return the result of *coro* execution. If *coro* is not a :term:`coroutine`,
+    a :exc:`ValueError` is raised.
```
1st1: Can we reformat this yet again? :)

I'd change to:

```
.. function:: run(coro, \*, debug=False)

    Execute the *coro* coroutine and return the result.
```

I don't see any value in documenting the ValueError tbh.

@asvetlov what do you think?

aeros: > Can we reformat this yet again? :)

Np. Even though it's just a couple of lines, this part of the docs for asyncio is especially important. I don't mind revising it a few times to get it right. (:

>  Execute the *coro* coroutine and return the result.

I like this phrasing better than the current one. It also effectively provides the same information that explaining the `ValueError` did, but in less words. Also, this seems more suitable to being the first section, rather the the last. 

Mind if I add the Sphinx role here and slightly reorder this?

```
Execute the :term:`coroutine` *coro* and return the result.
```

I think it's useful to link to glossary so that users who are less familiar with concurrency can obtain a basic understanding. Although one may assume that they've read the rest of the documentation page prior to this point, it's incredibly common for users to only read the documentation for the specific function or method they're interested in.

aeros167:
I also replaced the start of the docstring "Run a coroutine" with "Execute the coroutine *coro* and return the result". I would argue that "Run a coroutine" could be gathered from the name of the function and parameter alone. OTH, "Execute the coroutine *coro* and return the result" provides a bit more information while still remaining succinct.

I can remove this change if needed, but I think it provides a decent improvement and keeps the docstring in sync with the docs.

aeros167:
@1st1 The recommended changes have been applied. Thanks for the in-depth review. I'd much rather see this go through multiple revisions and be as high quality as possible, instead of having the PR merged more quickly. The former provides far more value to the users of asyncio.

aeros167:
On second thought, I'll remove "*coro*" from the docstring. Since name of the argument is technically "main" within runners.py, it wouldn't make sense have "*coro*" there as the arg name.

asvetlov:
Looks good.
Would you consider renaming `coro` to `main` in rst?
I think it makes sense for emphasizing `asyncio.run()` audience (plus reflects the implementation better).

Please double-check how ```:term:`coroutine` ``` is rendered. IIRC it was `Coroutines` or something like this, we used ```:term:`coroutine <coroutine>` ``` notation for some reason.

1st1:
> Would you consider renaming coro to main in rst?

Would `maincoro` work?

aeros167:
@1st1:
> Would maincoro work?

I think this name would make more sense, but it would bit odd if the .rst wasn't consistent with the actual name of the parameter in runners.py. 

Are we able to change the actual parameter name in runners.py without causing backwards compatibility issues? Since we didn't specify the "/" positional only symbol from PEP 570, users could technically call `asyncio.run()` with `asyncio.run(main=main())`, and that would no longer work if it were to be changed to *maincoro*. 

However, that might not be an issue if the parameter name *main* was never officially documented, therefore making it not a part of the public API.

aeros167:
@asvetlov:
> Please double-check how :term:`coroutine` is rendered. IIRC it was Coroutines or something like this, we used :term:`coroutine <coroutine>` notation for some reason.

No problem, I'll build the html locally to make sure the link works properly. 

`asyncio.run()` docs:

image

Clicking the link correctly leads to the glossary entry:

image

I believe the situation you were referring to was probably where the word was "coroutines", so you had to use:
```
:term:`coroutines <coroutine>`
```

Something like this
```
:term:`coroutines`
```
wouldn't work because there is not a literal term in the glossary for "coroutines".

That's a pretty commonly occurring issue with Sphinx, since the link target is very literal.

aeros167:
Looking at the glossary entry made me realize there's a minor typo in it. "Coroutines *is* a more generalized form of subroutines" should be "Coroutines *are* a more generalized form of subroutines". I'll add the correction to this PR, since it's just fixing a single word. That wouldn't be worth it's own PR, but might as well be fixed.

aeros167:
@1st1 Is there anything left to do with this PR? It's currently blocking https://github.com/python/cpython/pull/16360 since I'll need to resolve a conflict. Neither are particularly time sensitive by any means, I just wanted to wrap these up before we start working on the new streaming API for 3.9.

1st1 (approved):

1st1:
> @1st1 Is there anything left to do with this PR?

And... merged.  Thank you, Kyle!

aeros167:
Thanks!

PR #16418, opened 2019-09-26
bpo-38270: More fixes for strict crypto policy

tiran:
test_hmac and test_hashlib test built-in hashing implementations and
OpenSSL-based hashing implementations. Add more checks to skip OpenSSL
implementations when a strict crypto policy is active.

Use EVP_DigestInit_ex() instead of EVP_DigestInit() to initialize the
EVP context. The EVP_DigestInit() function clears alls flags and breaks
usedforsecurity flag again.

https://bugs.python.org/issue38270

### DIFF
--- a/Lib/test/support/__init__.py
+++ b/Lib/test/support/__init__.py
@@ -69,6 +69,11 @@ try:
 except ImportError:
     resource = None
 
+try:
+    import _hashlib
+except ImportError:
+    _hashlib = None
+
 __all__ = [
     # globals
     "PIPE_MAX_SIZE", "verbose", "max_memuse", "use_resources", "failfast",
@@ -86,8 +91,8 @@ __all__ = [
     "create_empty_file", "can_symlink", "fs_is_case_insensitive",
     # unittest
     "is_resource_enabled", "requires", "requires_freebsd_version",
-    "requires_linux_version", "requires_mac_ver", "check_syntax_error",
-    "check_syntax_warning",
+    "requires_linux_version", "requires_mac_ver", "requires_hashdigest",
+    "check_syntax_error", "check_syntax_warning",
     "TransientResource", "time_out", "socket_peer_reset", "ioerror_peer_reset",
     "transient_internet", "BasicTestRunner", "run_unittest", "run_doctest",
     "skip_unless_symlink", "requires_gzip", "requires_bz2", "requires_lzma",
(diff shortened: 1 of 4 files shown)


Lib/test/support/__init__.py
```diff
@@ -649,20 +654,29 @@ def wrapper(*args, **kw):
...
     ValueError: [digital envelope routines: EVP_DigestInit_ex] disabled for FIPS
     ValueError: unsupported hash type md4
     """
     def decorator(func):
         @functools.wraps(func)
         def wrapper(*args, **kwargs):
             try:
-                hashlib.new(digestname)
+                if openssl:
+                    if _hashlib is None:
```
vstinner: I would prefer to avoid having to inspect the _hashlib module to check which implementation picked hashlib. But it's not possible right now, so it's ok to rely on the private _hashlib module. My concern was that other Python implementation may want to reuse Python test suite, but may not have a _hashlib module.

Lib/test/support/__init__.py
```diff
@@ -649,20 +654,29 @@ def wrapper(*args, **kw):
...
     policy.
 
+    If 'openssl' is True, then the decorator checks that OpenSSL provides
+    the algorithm. Otherwise the check falls back to built-in
+    implementations. The usedforsecurity flag is passed to the constructor.
+
     ValueError: [digital envelope routines: EVP_DigestInit_ex] disabled for FIPS
     ValueError: unsupported hash type md4
     """
     def decorator(func):
```
vstinner, line 670: "import test.support" became an import nightmare: it imports way too many modules :-(

Would it be possible to import _hashlib here?

```
try:
    import _hashlib
except ImportError:
    _hashlib = None
```

Lib/test/test_hashlib.py
```diff
@@ -403,6 +415,7 @@ def test_blocksize_name_blake2(self):
         self.check_blocksize_name('blake2b', 128, 64)
         self.check_blocksize_name('blake2s', 64, 32)
 
+
```
vstinner: nitpick: PEP 8 suggests only 1 empty line between functions, you may revert this unrelated change.

encukou (approved):
This looks good to me. Please address Victor's comment (but I don't think the imports are that much of a problem).

Lib/test/support/__init__.py
tiran: It's possible but it doesn't make much sense IMO. The function would have to try/import/except for every decorator call. The support helper package is already a kitchen sink.

vstinner: > It's possible but it doesn't make much sense IMO.

"import test.support" is used by regrtest, our test runner, and almost all tests. Reducing imports should make tests faster and more "isolated".

> The function would have to try/import/except for every decorator call. The support helper package is already a kitchen sink.

Packages are put in sys.modules and so only the first import is slow.

We should split test.support monster into subpackages like test.support.sslutils.

PR #16428, opened 2019-09-26
bpo-38275: Fix test_ssl issue caused by GH-16386

tiran:
Check presence of SSLContext.minimum_version to make tests pass with
old versions of OpenSSL.

https://bugs.python.org/issue38275


vstinner:
>  bpo-38275: Fix for GH-16386 #16428 

Please mention at least test_ssl in the commit title when you merge this PR.

PR #16429, opened 2019-09-26
gh-81322: support multiple separators in Stream.readuntil

bmerry:
Allow Stream.readuntil to take an iterable of separators and match any
of them. The earliest match endpoint wins (which ensures that results
are dependent on the chunking) and on ties shortest separator wins
(which only matters if the user has supplied a redundant set like
[b'\r\n', b'\n'] and the limit is reached).

It's also implemented for the deprecated StreamReader, just because the
code for the two implementations was the same except for one line and it
seemed easier to keep them in sync than leaving two different versions
to maintain.

https://bugs.python.org/issue37141

* Issue: gh-81322


bmerry:
A few questions about this:
- I've implemented it for both StreamReader and Stream, because it seemed like having two identical (apart from `_ensure_can_read`) implementations is better for maintainability than two implementations that have drifted out of sync. But if one would prefer not to make any changes to StreamReader I can revert that. I suggest doing review discussion on the Stream implementation.
- Because the existing functionality is implemented using the general case there might be a very small loss of performance. Are there any benchmarks I should run to test that?
- For matches that end in the same position I arbitrarily decided that shortest should win for the purpose of LimitOverrunError, but it was an arbitrary choice. It could just as easily be longest-wins or first-in-list wins.

Lib/asyncio/streams.py
```diff
@@ -978,26 +992,35 @@ def feed_data(self, data):
...
+                        end = isep + len(sep)
+                        if match_end is None or end < match_end:
+                            match_end = end
+                            match_start = isep
+                if match_end is not None:
                     break
 
                 # see upper comment for explanation.
-                offset = buflen + 1 - seplen
+                offset = buflen + 1 - max_seplen
```
bmerry: Hmm, I see this can make offset negative if max_seplen > min_seplen. I'll add a test to catch it and add a fix.

Lib/asyncio/streams.py
```diff
@@ -952,17 +952,31 @@ def feed_data(self, data):
...
+        the shortest possible separator is considered to be the one that
+        matched.
         """
-        seplen = len(separator)
-        if seplen == 0:
+        if isinstance(separator, bytes):
+            separator = [separator]
+        else:
+            # Makes sure shortest matches wins, and supports arbitrary iterables
+            separator = sorted(separator, key=lambda sep: len(sep))
```
brandtbucher: No need to wrap this in a lambda!

```suggestion
            separator = sorted(separator, key=len)
```

Lib/asyncio/streams.py
```diff
@@ -952,17 +952,31 @@ def feed_data(self, data):
...
+        if not separator:
+            raise ValueError('Separator list should contain at least one element')
```
brandtbucher: Since you support any iterable, maybe leave "list" out of it?

```suggestion
            raise ValueError('Separator should contain at least one element')
```

Lib/asyncio/streams.py
```diff
@@ -978,26 +992,35 @@ def feed_data(self, data):
         #   messages :)
 
         # `offset` is the number of bytes from the beginning of the buffer
-        # where there is no occurrence of `separator`.
+        # where there is no occurrence of any separator.
```
brandtbucher: I'd leave the backticks here. Even though it's a list now, it still makes it clear that we're referring to the variable.

```suggestion
        # where there is no occurrence of any `separator`.
```

Lib/asyncio/streams.py
```diff
@@ -978,26 +992,35 @@ def feed_data(self, data):
...
         offset = 0
 
-        # Loop until we find `separator` in the buffer, exceed the buffer size,
+        # Loop until we find a separator in the buffer, exceed the buffer size,
```
brandtbucher: Here too:

```suggestion
        # Loop until we find a `separator` in the buffer, exceed the buffer size,
```

Lib/asyncio/streams.py
```diff
@@ -1645,18 +1668,32 @@ def _feed_data(self, data):
...
         _ensure_can_read(self._mode)
-        seplen = len(separator)
-        if seplen == 0:
+        if isinstance(separator, bytes):
+            separator = [separator]
+        else:
+            # Makes sure shortest matches wins, and supports arbitrary iterables
+            separator = sorted(separator, key=lambda sep: len(sep))
+        if not separator:
+            raise ValueError('Separator list should contain at least one element')
```
brandtbucher: See above.

```suggestion
            raise ValueError('Separator should contain at least one element')
```

Lib/asyncio/streams.py
```diff
@@ -1672,26 +1709,35 @@ def _feed_data(self, data):
         #   messages :)
 
         # `offset` is the number of bytes from the beginning of the buffer
-        # where there is no occurrence of `separator`.
+        # where there is no occurrence of any separator.
```
brandtbucher: See above.

```suggestion
        # where there is no occurrence of any `separator`.
```

Lib/asyncio/streams.py
```diff
@@ -1672,26 +1709,35 @@ def _feed_data(self, data):
...
         offset = 0
 
-        # Loop until we find `separator` in the buffer, exceed the buffer size,
+        # Loop until we find a separator in the buffer, exceed the buffer size,
```
brandtbucher: See above.

```suggestion
        # Loop until we find a `separator` in the buffer, exceed the buffer size,
```

brandtbucher (changes requested):
Thanks for taking the time to draft this PR @bmerry, and welcome to CPython! :sunglasses:

I'm not *super* familiar with this module, but I found a couple of things that could be tidied up. Other than that, the code looks good! This will definitely need updated docs, though.

Lib/asyncio/streams.py
```diff
@@ -1645,18 +1668,32 @@ def _feed_data(self, data):
...
+        matched.
         """
         _ensure_can_read(self._mode)
-        seplen = len(separator)
-        if seplen == 0:
+        if isinstance(separator, bytes):
+            separator = [separator]
+        else:
+            # Makes sure shortest matches wins, and supports arbitrary iterables
+            separator = sorted(separator, key=lambda sep: len(sep))
```
brandtbucher: Sorry, missed one. Here too!

```suggestion
            separator = sorted(separator, key=len)
```

1st1 (changes requested):
Unfortunately it was decided to revert the current streams implementation from 3.8.  See https://bugs.python.org/issue38242 for more details.  I'm really sorry, but we'll need to rebase this work on the new API we later add to 3.9 :(

bmerry:
@1st1 not a problem - definitely worth getting things right before they go into stdlib. Do you think the revert of the new stream API will conflict much with my changes to the old StreamReader class? If not, perhaps the way forward is for me to remove my changes to `Stream`, address the PR comments on `StreamReader` and then hopefully it shouldn't need much work to rebase after `Stream` is removed?

1st1:
@bmerry I'm not yet sure. I'll be working on a revert later today or tomorrow and will probably know more.  I'd keep everything as is as it's not going to be an easy revert anyways.

bmerry:
@1st1 it looks like you've done the Streams reversion now, and I've merged that into my branch. So hopefully this is ready for review now.

@brandtbucher thanks for the suggestions. I've applied all the ones that Github was still showing after the merge from master - let me know if anything got lost along the way.

Misc/NEWS.d/next/Library/2019-09-26-17-52-52.bpo-37141.onYY2-.rst
```diff
@@ -0,0 +1,2 @@
+Support a list of separators in :meth:`asyncio.Stream.readuntil`, stopping
```
brandtbucher:
```suggestion
Accept an iterable of separators in :meth:`asyncio.Stream.readuntil`, stopping
```

brandtbucher (changes requested):
One more tiny clarification from me, otherwise this looks good. Still needs updated docs though!

Misc/NEWS.d/next/Library/2019-09-26-17-52-52.bpo-37141.onYY2-.rst
bmerry: Done

bmerry:
> Still needs updated docs though!

@brandtbucher which docs need updating?

brandtbucher (changes requested):
One tiny change to the NEWS entry due to the API revert:

Misc/NEWS.d/next/Library/2019-09-26-17-52-52.bpo-37141.onYY2-.rst
```diff
@@ -0,0 +1,2 @@
+Accept an iterable of separators in :meth:`asyncio.Stream.readuntil`, stopping
```
brandtbucher:
```suggestion
Accept an iterable of separators in :meth:`asyncio.StreamReader.readuntil`, stopping
```

bmerry:
> The docs I'm referring to are located in Doc/library/asyncio-stream.rst

Thanks. I'd assumed the library docs were generated from the docstring, which is why I missed this. I'll update it.

brandtbucher (approved):

brandtbucher:
Looks good to me! @1st1?

bmerry:
@1st1 any news on this now that Python 3.9 has reached alpha?

bmerry:
@1st1 any news on what streams will look like in Python 3.9?

bmerry:
@1st1 @asvetlov any news on what streams will look like in Python 3.9?

bmerry:
It looks like Python 3.9 isn't changing the stream API, so I've rebased the PR (although I assume it will only target 3.10 since 3.9 has reached beta). The rebasing turned into a bit of a mess so I ended up squashing the whole thing to make a single new commit.

bmerry:
@brandtbucher any advice on how I can move this forward? There has been no response on it all year.

brandtbucher:
Thanks for your patience @bmerry! I think @1st1 has probably just been busy.

While your work here still looks fine to me in terms of style and documentation, I personally know next to nothing about async stuff. Perhaps @aeros has some time to look at this (and perhaps merge it)?

aeros:
Sorry @bmerry for the delay. There's still a strong long-term intention for an overhaul to the streams API, but due to other work being prioritized (such as exception groups, and by extension, task groups), we decided at the core sprint to delay work on a new streams API for now since it will be less impactful for most users. 

My thoughts are that supporting multiple separators in `readuntil()` is useful enough to consider as an enhancement to the current version of streams, particularly since it will remain functional for several versions before *eventually* being deprecated and removed. 

However, I'd like to get input from @1st1 and @asvetlov to decide if it's worthwhile from their perspective, and in general what our current plans are surrounding enhancements to the existing streams API (in light of the recent decision at the 2020 core sprint to delay work on it for now).

gvanrossum:
Thanks, I will review in the coming week!

Misc/NEWS.d/next/Library/2019-09-26-17-52-52.bpo-37141.onYY2-.rst
bmerry, line 1: Does the filename need to be changed to reflect the Github issue number instead of the BPO number?

gvanrossum: I don't think so -- this is just used to make the files unique, at the next alpha release it will all be merged into a single 3.13.0a6.rst file.

Lib/asyncio/streams.py
```diff
@@ -590,17 +590,31 @@ async def readuntil(self, separator=b'\n'):
...
             raise self._exception
 
         # Consume whole buffer except last bytes, which length is
-        # one less than seplen. Let's check corner cases with
-        # separator='SEPARATOR':
+        # one less than max_seplen. Let's check corner cases with
+        # separator[-1]='SEPARATOR':
         # * we have received almost complete separator (without last
         #   byte). i.e buffer='some textSEPARATO'. In this case we
         #   can safely consume len(separator) - 1 bytes.
```
gvanrossum:
```suggestion
        #   can safely consume max_seplen - 1 bytes.
```

gvanrossum (approved):
One comment nit, I'll merge that myself and then land it. Thanks for your contribution, and on behalf of the asyncio maintainers I apologize for the delay!

PR #16432, opened 2019-09-27
bpo-38290: cleanup ababstractproperty in typing.py

hongweipeng:
https://bugs.python.org/issue38290


ilevkivskyi (approved):
Thanks, LGTM!

PR #16438, opened 2019-09-27
bpo-38293: Allow shallow and deep copying of property objects

GudniNathan:
Copying property objects results in a TypeError. Steps to reproduce:

```
>>> import copy
>>> obj = property()
>>> copy.copy(obj)
````

This affects both shallow and deep copying.  
My idea for a fix is to add property objects to the list of "atomic" objects in the copy module.
These already include types like functions and type objects.

I also added property objects to the unit tests test_copy_atomic and test_deepcopy_atomic. This is my first PR, and it's highly likely I've made some mistake, so please be kind :)

https://bugs.python.org/issue38293

Automerge-Triggered-By: @csabella

### DIFF
--- a/Lib/copy.py
+++ b/Lib/copy.py
@@ -107,7 +107,7 @@ _copy_dispatch = d = {}
 def _copy_immutable(x):
     return x
 for t in (type(None), int, float, bool, complex, str, tuple,
-          bytes, frozenset, type, range, slice,
+          bytes, frozenset, type, range, slice, property,
           types.BuiltinFunctionType, type(Ellipsis), type(NotImplemented),
           types.FunctionType, weakref.ref):
     d[t] = _copy_immutable
@@ -195,6 +195,7 @@ d[type] = _deepcopy_atomic
 d[types.BuiltinFunctionType] = _deepcopy_atomic
 d[types.FunctionType] = _deepcopy_atomic
 d[weakref.ref] = _deepcopy_atomic
+d[property] = _deepcopy_atomic
 
 def _deepcopy_list(x, memo, deepcopy=deepcopy):
     y = []
--- a/Lib/test/test_copy.py
+++ b/Lib/test/test_copy.py
@@ -99,7 +99,7 @@ class TestCopy(unittest.TestCase):
                  42, 2**100, 3.14, True, False, 1j,
                  "hello", "hello\u1234", f.__code__,
                  b"world", bytes(range(256)), range(10), slice(1, 10, 2),
-                 NewStyle, Classic, max, WithMetaclass]
+                 NewStyle, Classic, max, WithMetaclass, property()]
         for x in tests:
             self.assertIs(copy.copy(x), x)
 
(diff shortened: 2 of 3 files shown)

### CODE REVIEW

**Review by eamanu — COMMENTED** [review]
Good catch @GudniNatan i am not copy expert but I test it and run successfully.

I don't know if will need necessary a NEWS entry, IMO yes, but I would like hear some
core dev opinion.

**Review by brandtbucher — CHANGES_REQUESTED** [review]
Thanks @GudniNatan, and welcome to CPython. This looks good!

I think this should have a NEWS entry. It's a new feature (and there's precedent for it when `copy`/`deepcopy` support was added for other objects in the past). [Creating one is super quick!](https://blurb-it.herokuapp.com/add_blurb)

Maybe just something like:

```rst
Add :func:`copy.copy` and :func:`copy.deepcopy` support to :func:`property` objects.
```

Just out of curiosity, what was your use case for wanting to copy these?

GudniNatan:
I was experimenting with using property objects to validate data in dataclasses. Here is some code that should help illustrate:

```
from dataclasses import field, dataclass, asdict, fields


def string_check(x):
    if type(x) != str:
        raise ValueError("Should be a string.")
    return x


def validator_factory(field_name, validator):
    def fget(self):
        return self.__dict__[field_name]

    def fset(self, value):
        self.__dict__[field_name] = validator(value)

    return property(fget, fset)


@dataclass
class MyClass():
    my_value: str = validator_factory("my_value", string_check)


@dataclass
class ClassSchema():
    field_defaults: list


my_class_schema = ClassSchema(
    field_defaults=[f.default for f in fields(MyClass)]
)

print(my_class_schema)
print(asdict(my_class_schema))
```

The MyClass dataclass has an attribute my_value, and the idea is that if you set it to be anything other than a string, it will throw a ValueError.

We were also using a library ([dataclasses-jsonschema](https://pypi.org/project/dataclasses-jsonschema/)) that has some similar functionality to ClassSchema, in that it stores the default values of our dataclasses. It will then try to convert those to dictionaries with the dataclasses asdict function. However, asdict will try to give you a dictionary that is independent of the instance you feed it. To do that it uses the deepcopy, and then you get the `TypeError: can't pickle property objects`

So basically calling asdict() on certain dataclass instances results in this weird error.

brandtbucher (approved):

gvanrossum (commented):
I have a nagging doubt. You can create a property from any callable, e.g. `property(lambda self: None)`. The argument is not type-checked, so you can also create a property from a non-callable, e.g. `property([1, 2, 3])`. When we deep-copy such a property, shouldn't we deep-copy the list (i.e. the `fget` attribute)? After all, when we deep-copy a tuple containing a list, the list gets copied.

@GudniNatan For your use case, is deep-copyability required?

arigo:
@gvanrossum following that argument, ``copy.copy(func)`` should not return its original object either, because ``func`` is mutable (i.e. a following ``func.foo = 42`` should not appear on its copy).  Of course I'm not suggesting that this should be changed; this is only a reminder that ``copy()`` and ``deepcopy()`` are best-effort hacks in the first place.

gvanrossum:
@arigo That's fair. Tuples are special, because they are *meant* to be robust data structures, and a tuple containing a list is a well-known (and used) corner case. I'll approve.

gvanrossum (approved):

brandtbucher:
Congrats on your first CPython contribution @GudniNatan! :champagne:

PR #16443, opened 2019-09-27
bpo-38161: Removes _AwaitEvent from AsyncMock.

lisroach:
This is something we can add back in as a feature in the future, but for now I want to take it out. With the release of 3.8 coming up I don't think this is thought out well enough- I don't like the name, I'm struggling to document it, its not tested, and I'm not sure if it's worth getting stuck with an attribute we may have to make significant changes to in the future. 

If someone really wants it in now that's fine, just make a PR adding all those things in for me :) 

https://bugs.python.org/issue38161


lisroach:
Going to push since we are still in beta and I want it gone before RC tomorrow.

PR #16446, opened 2019-09-28
bpo-38301: In Solaris family, we must be sure to use '-D_REENTRANT'

jcea:
https://bugs.python.org/issue38301


jcea:
Tested under SmartOS "joyent_20190829T000927Z", with pkgsrc "base-64 19.2.0". Tha patch is trivial and it only affects SunOS machines.

PR #16447, opened 2019-09-28
[2.7] bpo-38243: Escape the server title of DocXMLRPCServer

corona10:
https://bugs.python.org/issue38243

### DIFF
--- a/Lib/DocXMLRPCServer.py
+++ b/Lib/DocXMLRPCServer.py
@@ -20,6 +20,17 @@ from SimpleXMLRPCServer import (SimpleXMLRPCServer,
             CGIXMLRPCRequestHandler,
             resolve_dotted_attribute)
 
+
+def _html_escape_quote(s, quote=True):
+    s = s.replace("&", "&amp;") # Must be done first!
+    s = s.replace("<", "&lt;")
+    s = s.replace(">", "&gt;")
+    if quote:
+        s = s.replace('"', "&quot;")
+        s = s.replace('\'', "&#x27;")
+    return s
+
+
 class ServerHTMLDoc(pydoc.HTMLDoc):
     """Class used to generate pydoc HTML document for a server"""
 
@@ -210,14 +221,7 @@ class XMLRPCDocGenerator:
                                 methods
                             )
 
-        escape_table = {
-            "&": "&amp;",
-            '"': "&quot;",
-            "'": "&#x27;",
-            ">": "&gt;",
-            "<": "&lt;",
-        }
-        title = ''.join(escape_table.get(c, c) for c in self.server_title)
+        title = _html_escape_quote(self.server_title)
         return documenter.page(title, documentation)
 
 class DocXMLRPCRequestHandler(SimpleXMLRPCRequestHandler):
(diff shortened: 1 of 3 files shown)


corona10:
@vstinner Please take a look :)

Lib/DocXMLRPCServer.py
```diff
@@ -210,7 +210,15 @@ def generate_html_documentation(self):
...
 
-        return documenter.page(self.server_title, documentation)
+        escape_table = {
+            "&": "&amp;",
+            '"': "&quot;",
+            "'": "&#x27;",
+            ">": "&gt;",
+            "<": "&lt;",
+        }
+        title = ''.join(escape_table.get(c, c) for c in self.server_title)
```
vstinner: Please don't inline html.escape() here, but add a private function at top-level. Function name like "_html_escape_quote()" with code copied from Lib/html/__init__.py. You can hardcode quote=True in the function.

I don't think that your code behaves as html.escape() in master. The replace must be done in a specific order. In case of doubt, I prefer to reuse code from master.

vstinner (changes requested):

Lib/DocXMLRPCServer.py
corona10: I 've updated the PR as you comment!
Thanks for the feedback.

> The replace must be done in a specific order

Yes this should be in master version
But b41cde823d026f2adc21ef14b1c2e92b1006de06 version does not  has to care
about ordering.
Because it does not use replace but iterating and changing.

For example
'<&'
on the master version without ordering can be happen in this below scenario
````
s = '<&'
s = '&lt;&' # if < is replaced first
s = '&amp;lt;&amp;' # what?
````

but on the b41cde823d026f2adc21ef14b1c2e92b1006de06 version
````
s = '<&'
l = [&lt;], c ='<' s = '<&'
l = [&lt; &amp;], c ='&' s = '<&'
''.join(l) # will be '&lt; &amp;' we expacted.
````

~I'd like to optimize html.escape of master version with~ ~b41cde823d026f2adc21ef14b1c2e92b1006de06 ~
~if this showing better performance. (I didn't benchmark yet)~
~What do you think?~

**Comment by corona10** [comment]

File: Lib/DocXMLRPCServer.py (hunk 1 above)
Comment (corona10) [reply Lib/DocXMLRPCServer.py]: I found that it's not worth to do.

**Comment by vstinner** [comment]
> I'd like to optimize html.escape of master version with b41cde8 if this showing better performance. What do you think?

5 str.replace() calls sound inefficient, so there is likely room for enhancement. But I am not sure what is the best approach. If you find something which is faster, please open a separated issue. But a security fix in a super stable branch (2.7) is not the right place to experiment optimization :-) Correctness matters more here ;-)

File: Lib/DocXMLRPCServer.py [hunk 2]
```diff
@@ -20,6 +20,17 @@
             CGIXMLRPCRequestHandler,
             resolve_dotted_attribute)
 
+
+def _html_escape_quote(s, quote=True):
```
vstinner: Oh, my suggestion was to remove the quote parameter since it's always quote in this code. Either rename the function to _html_escape() and keep the parameter, or keep this name but remove the parameter.

corona10:
Updated!

> If you find something which is faster,

Okay, I will take a look at it  :)

vstinner (approved):
LGTM, thanks. It's now better than the first version of the PR ;-)

PR #16448, opened 2019-09-28
bpo-38216, bpo-36274: Allow subclasses to separately override validation and encoding behavior

jaraco:
This PR builds on and supersedes #16321 as an alternative approach for consideration.

### DIFF
--- a/Lib/http/client.py
+++ b/Lib/http/client.py
@@ -1088,13 +1088,12 @@ class HTTPConnection:
         # Save the method for use later in the response phase
         self._method = method
 
-        request = b' '.join((
-            method.encode('ascii'),
-            self._prepare_path(url or '/'),
-            self._http_vsn_str.encode('ascii')
-        ))
+        url = url or '/'
+        self._validate_path(url)
 
-        self._output(request)
+        request = '%s %s %s' % (method, url, self._http_vsn_str)
+
+        self._output(self._encode_request(request))
 
         if self._http_vsn == 11:
             # Issue some standard headers for better HTTP/1.1 compliance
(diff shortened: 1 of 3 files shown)


Lib/http/client.py
```diff
@@ -1085,18 +1085,15 @@ def putrequest(self, method, url, skip_host=False,
...
-                             f"(found at least {match.group()!r})")
+
+        url = url or '/'
+        self._validate_path(url)
+
         request = '%s %s %s' % (method, url, self._http_vsn_str)
 
-        # Non-ASCII characters should have been eliminated earlier
-        self._output(request.encode('ascii'))
+        self._output(self._encode_request(request))
```
gpshead, line 1096: overall I think I do like this PR better.  It seems... cleaner.  The one thing of note different here vs #16321 is effectively this call.  encoding and validation are entirely separate, and the encoding applies to the entire http request line including the method and version.  i like it.

gpshead (approved):

ned-deily (approved):
LGTM, and I also prefer this approach.  Thanks!

Lib/http/client.py
jaraco: This approach is really similar to the first draft I'd proposed, but it had these issues.

- it adds overhead of two function calls to the routine (probably not a big deal)
- it requires the overriding tool to possibly override multiple functions
- it doesn't give an overriding tool the ability to accept something other than a string for the url/path.

It was these reasons that led me toward the approach in the other proposal.

But I agree, this implementation, while providing less freedom of customization, is simpler, easier to read, and likely achieves everything that's needed.

jaraco:
[GH-16475](./16475) is a backport of this pull request to the [3.5 branch](https://github.com/python/cpython/tree/3.5).

jaraco:
[GH-16476](16476) is a backport of this pull request to the [2.7 branch](https://github.com/python/cpython/tree/2.7).

PR #16455, opened 2019-09-28
bpo-38242: Revert new asyncio streaming API

aeros:
Motivation: Comment in bpo issue from @1st1:

> My final opinion on this issue: we must revert the new streaming API from asyncio and take our time to design it properly.  I don't like what we have in the 3.8 branch right now.  I don't feel comfortable rushing decisions and doing last minute API changes.

Reverts the following commits from 3.8:

Merge asyncio streams (primary reversion):
https://github.com/python/cpython/commit/23b4b697e5b6cc897696f9c0288c187d2d24bff2

Documentation:
https://github.com/python/cpython/commit/b9bfe143d151d184615fa3890f78874c5d4ed4c6
https://github.com/python/cpython/commit/7268fd0d212e870103ae052697b32ba54d3f6299
https://github.com/python/cpython/commit/ed9f3562b637a59b9000abbceee5ae369d35444d

Minor test changes (incompatible with `asyncio.StreamReader` and `asyncio.StreamWriter`):
https://github.com/python/cpython/commit/4cdbc452ce3

Manual resolution of several conflicts was also required, the majority were for retaining the deprecation of explicitly passing the loop argument for tasks (https://github.com/python/cpython/commit/345bfc990f5f3e873774051d43136b06bfed82cb). There might be a few deprecation warnings missing, still working on adding some back.

This PR must be carefully reviewed, especially since it is removing multiple tests that no longer function properly without the changes made in https://github.com/python/cpython/commit/23b4b697e5b6cc897696f9c0288c187d2d24bff2. It is important to be 100% certain that no additional tests were removed. I believe @asvetlov would be the most qualified individual to verify the tests are in a correct state, since he wrote the majority of the tests related to the new streaming API.

It is most of the way there, but additional changes may be needed before it is complete. Until we are certain, I am adding the `DO-NOT-MERGE` label.

To @1st1 and @asvetlov:

I have enabled "Allow edits from maintainers" (as I usually do). Please feel free to directly commit changes into the PR. I would appreciate review feedback so that I can understand the reasons behind any changes made, but you certainly don't have to wait on me to make any changes to the PR. Especially since this is a release blocker for 3.8.0.

### DIFF
--- a/Lib/asyncio/streams.py
+++ b/Lib/asyncio/streams.py
@@ -48,11 +48,10 @@ async def open_connection(host=None, port=None, *,
                       "and scheduled for removal in Python 3.10.",
                       DeprecationWarning, stacklevel=2)
     reader = StreamReader(limit=limit, loop=loop)
-    protocol = StreamReaderProtocol(reader, loop=loop, _asyncio_internal=True)
+    protocol = StreamReaderProtocol(reader, loop=loop)
     transport, _ = await loop.create_connection(
         lambda: protocol, host, port, **kwds)
-    writer = StreamWriter(transport, protocol, reader, loop,
-                          _asyncio_internal=True)
+    writer = StreamWriter(transport, protocol, reader, loop)
     return reader, writer
 
 
@@ -87,11 +86,9 @@ async def start_server(client_connected_cb, host=None, port=None, *,
                       DeprecationWarning, stacklevel=2)
 
     def factory():
-        reader = StreamReader(limit=limit, loop=loop,
-                              _asyncio_internal=True)
+        reader = StreamReader(limit=limit, loop=loop)
         protocol = StreamReaderProtocol(reader, client_connected_cb,
-                                        loop=loop,
-                                        _asyncio_internal=True)
+                                        loop=loop)
         return protocol
 
     return await loop.create_server(factory, host, port, **kwds)
@@ -110,12 +107,10 @@ if hasattr(socket, 'AF_UNIX'):
                           "and scheduled for removal in Python 3.10.",
                           DeprecationWarning, stacklevel=2)
         reader = StreamReader(limit=limit, loop=loop)
-        protocol = StreamReaderProtocol(reader, loop=loop,
-                                        _asyncio_internal=True)
+        protocol = StreamReaderProtocol(reader, loop=loop)
         transport, _ = await loop.create_unix_connection(
             lambda: protocol, path, **kwds)
-        writer = StreamWriter(transport, protocol, reader, loop,
-                              _asyncio_internal=True)
+        writer = StreamWriter(transport, protocol, reader, loop)
         return reader, writer
 
     async def start_unix_server(client_connected_cb, path=None, *,
@@ -129,11 +124,9 @@ if hasattr(socket, 'AF_UNIX'):
                           DeprecationWarning, stacklevel=2)
 
         def factory():
-            reader = StreamReader(limit=limit, loop=loop,
-                                  _asyncio_internal=True)
+            reader = StreamReader(limit=limit, loop=loop)
             protocol = StreamReaderProtocol(reader, client_connected_cb,
-                                            loop=loop,
-                                            _asyncio_internal=True)
+                                            loop=loop)
             return protocol
 
         return await loop.create_unix_server(factory, path, **kwds)
@@ -149,20 +142,11 @@ class FlowControlMixin(protocols.Protocol):
     StreamWriter.drain() must wait for _drain_helper() coroutine.
     """
 
-    def __init__(self, loop=None, *, _asyncio_internal=False):
+    def __init__(self, loop=None):
         if loop is None:
             self._loop = events.get_event_loop()
         else:
             self._loop = loop
-        if not _asyncio_internal:
-            # NOTE:
-            # Avoid inheritance from FlowControlMixin
-            # Copy-paste the code to your project
-            # if you need flow control helpers
-            warnings.warn(f"{self.__class__} should be instantiated "
-                          "by asyncio internals only, "
-                          "please avoid its creation from user code",
-                          DeprecationWarning)
         self._paused = False
         self._drain_waiter = None
         self._connection_lost = False
@@ -227,9 +211,8 @@ class StreamReaderProtocol(FlowControlMixin, protocols.Protocol):
 
     _source_traceback = None
 
-    def __init__(self, stream_reader, client_connected_cb=None, loop=None,
-                 *, _asyncio_internal=False):
-        super().__init__(loop=loop, _asyncio_internal=_asyncio_internal)
+    def __init__(self, stream_reader, client_connected_cb=None, loop=None):
+        super().__init__(loop=loop)
         if stream_reader is not None:
             self._stream_reader_wr = weakref.ref(stream_reader,
                                                  self._on_reader_gc)
@@ -290,8 +273,7 @@ class StreamReaderProtocol(FlowControlMixin, protocols.Protocol):
         if self._client_connected_cb is not None:
             self._stream_writer = StreamWriter(transport, self,
                                                reader,
-                                               self._loop,
-                                               _asyncio_internal=True)
+                                               self._loop)
             res = self._client_connected_cb(reader,
                                             self._stream_writer)
             if coroutines.iscoroutine(res):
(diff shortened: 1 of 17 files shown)


aeros167:
There's still a few areas that I need to work on, but this should have the majority of the important changes.

Lib/test/test_asyncio/test_streams.py
```diff
@@ -63,15 +42,13 @@ def tearDown(self):
...
-                                _asyncio_internal=True)
+        stream = asyncio.StreamReader(_asyncio_internal=True)
         self.assertIs(stream._loop, m_events.get_event_loop.return_value)
 
     def _basetest_open_connection(self, open_connection_fut):
         messages = []
         self.loop.set_exception_handler(lambda loop, ctx: messages.append(ctx))
-        with self.assertWarns(DeprecationWarning):
-            reader, writer = self.loop.run_until_complete(open_connection_fut)
+        reader, writer = self.loop.run_until_complete(open_connection_fut)
```
aeros: Should the ``with self.assertWarns(DeprecationWarning):`` be added back to this location? This was present presumably for the deprecation of explicitly passing the loop arg. This also applies the the two similar sections below on lines 79 and 115.

```suggestion
        with self.assertWarns(DeprecationWarning):
            reader, writer = self.loop.run_until_complete(open_connection_fut)
```

Lib/test/test_asyncio/test_streams.py
```diff
@@ -99,9 +76,7 @@ def _basetest_open_connection_no_loop_ssl(self, open_connection_fut):
         messages = []
         self.loop.set_exception_handler(lambda loop, ctx: messages.append(ctx))
         try:
-            with self.assertWarns(DeprecationWarning):
-                reader, writer = self.loop.run_until_complete(
-                    open_connection_fut)
+            reader, writer = self.loop.run_until_complete(open_connection_fut)
```
aeros:
```suggestion
            with self.assertWarns(DeprecationWarning):
                reader, writer = self.loop.run_until_complete(open_connection_fut)
```

Lib/test/test_asyncio/test_streams.py
```diff
@@ -137,8 +112,7 @@ def test_open_unix_connection_no_loop_ssl(self):
     def _basetest_open_connection_error(self, open_connection_fut):
         messages = []
         self.loop.set_exception_handler(lambda loop, ctx: messages.append(ctx))
-        with self.assertWarns(DeprecationWarning):
-            reader, writer = self.loop.run_until_complete(open_connection_fut)
+        reader, writer = self.loop.run_until_complete(open_connection_fut)
```
aeros:
```suggestion
        with self.assertWarns(DeprecationWarning):
            reader, writer = self.loop.run_until_complete(open_connection_fut)
```

aeros167:
Since StreamReader is now the defacto API with Stream being removed, should the it's "internal usage only" warning be removed?
```
DeprecationWarning: <class 'asyncio.streams.StreamReader'> should be instaniated by
asyncio internals only ...
```

Lib/test/test_asyncio/test_streams.py
aeros: Confirmed the purpose of the deprecation warning by running this specific unit test and seeing the warning emitted from this line without it.

aeros167:
Restored `test__all__.py` and `test_sslproto.py` since they did not require reversion. Also, removing tiran from reviewers (I believe he was added from ownership of `test_sslproto`).

Lib/test/test_asyncio/test_windows_events.py
```diff
@@ -118,16 +117,16 @@ def test_pipe(self):
 
         clients = []
         for i in range(5):
-            stream = asyncio.Stream(mode=asyncio.StreamMode.READ,
-                                    loop=self.loop, _asyncio_internal=True)
-            protocol = _StreamProtocol(stream,
-                                       loop=self.loop,
-                                       _asyncio_internal=True)
+            stream_reader = asyncio.StreamReader(loop=self.loop,
+                                                 _asyncio_internal=True)
```
1st1: Probably `_asyncio_internal` is no longer needed.

1st1:
@asvetlov will you have time to glance over today/tomorrow?

Lib/test/test_asyncio/test_windows_events.py
aeros: @1st1 It should probably be removed from the `__init__` of StreamReader (and all corresponding references) since it's no longer internal with Stream removed, but it can remain for StreamReaderProtocol since that class is still intended for internal usage (AFAIK). Is this okay?

1st1:
`StreamReaderProtocol` was technically a public API and a lot of people use it.  So I'd revert this change entirely.  We'll just deprecate all of the current streaming API in 3.9.

aeros167:
> `StreamReaderProtocol` was technically a public API and a lot of people use it.

Ah, I see. I had never used it myself, but based on the documentation summary of StreamReaderProtocol being an adapter class between Protocol and StreamReader, I had assumed it was primarily intended for internal usage. I'll remove it from both of them if it's public API.

aeros167:
Removing it from Process as well. It looks like that is also public API based on the docs.

aeros167:
> As a result, I would assume it should remain for Process.

Never mind, I'm starting to think this wouldn't make much sense. I was initially assuming that I could use something like this within `SubprocessStreamProtocol.__init__()`, but this wouldn't work:
```
def __init__(self, limit, loop):
    super().__init__(loop=loop, _asyncio_internal=False)
```
Instead, we would have to use this:
```
def __init__(self, limit, loop, _asyncio_internal=True):
    super().__init__(loop=loop, _asyncio_internal=_asyncio_internal)
```
IMO, that would be far too counter intuitive for users reading the source code to be useful. Having _asyncio_internal always set to True in this case would imply that users are using the internals anytime they invoke `SubprocessStreamProtocol()`, which doesn't make sense.

I think it would be more straightforward to just remove `_asyncio_internal` from Process as well. Working on that next, let me know what you think @1st1.

1st1:
You should use a combination of `git blame`, `git log`, `git show` to recover all bits and pieces that were checked in when we were working on new Streams.  If that thing wasn't part of Process before it shouldn't be now.

aeros167:
@1st1 
> You should use a combination of git blame, git log, git show to recover all bits and pieces that were checked in when we were working on new Streams.

I have been primarily using `git blame` as much as possible, but I thought that one was a bit more subjective since the constructor args for Process are not documented.

> If that thing wasn't part of Process before it shouldn't be now.

It was added by @asvetlov in https://github.com/python/cpython/commit/ad4ed872415d00fcdfaa52a08108ec752b115000. The `_asyncio_internal` changes were technically done before the new streaming API was implemented in https://github.com/python/cpython/commit/23b4b697e5b6cc897696f9c0288c187d2d24bff2, which is why I was asking.

aeros167:
I finishing removing all instances of `asyncio_internal` and removed a few tests that were lingering which were created for the sole purpose of testing the DeprecationWarnings for usage of the old API. I've also manually added and removed DeprecationWarnings in other parts of the tests. The ones added for the new API were removed, and I added a few `self.assertWarning(DeprecationWarning)` to tests were emitting warnings from explicit passing of the loop arg.

As far as I can tell, that should be everything. All of the tests are passing without any warnings and every component of the new streaming API has been reverted.

The only other thing I can think of would be adding the explicit loop arg deprecation to the old API classes. Currently, this does not emit a deprecation warning:
```
rd = StreamReader(loop=loop)
```
The same applies to StreamReaderProtocol, StreamWriter, SubprocessStreamProtocol, and Process. It was never added to those classes since the deprecation of passing an explicit loop arg was implemented against the new API by @asvetlov in https://github.com/python/cpython/commit/345bfc990f5f3e873774051d43136b06bfed82cb.

However, that might not be necessary if we're adding another new streaming API in 3.9, so I'll leave that decision up to you two. Let me know if there's anything I'm missing. If I don't respond within a short period of time, feel free to directly commit any suggested changes to the PR.

aeros167:
While looking over `test_streams` again I noticed a duplicated test name `test_async_writer_api`. At first, I thought the test itself was duplicated and was considering removing it, but I then realized the tests were different. Based on the current version, it looks like the name of the second one was supposed to be `test_async_writer_api_exception_after_close`. 

I have no idea how that happened in the first place, but it's fixed now.

Lib/asyncio/streams.py
```diff
@@ -766,18 +425,17 @@ def get_extra_info(self, name, default=None):
             if exc is not None:
                 raise exc
         if self._transport.is_closing():
-            # Yield to the event loop so connection_lost() may be
-            # called.  Without this, _drain_helper() would return
-            # immediately, and code that calls
-            #     write(...); await drain()
-            # in a loop would never call connection_lost(), so it
-            # would not see an error when the socket is closed.
-            await tasks.sleep(0, loop=self._loop)
```
aeros, line 775: This was added in commit https://github.com/python/cpython/commit/23b4b697e5b6cc897696f9c0288c187d2d24bff2 with the new streaming API changes, but it looks like it's unrelated. It seems like it would still be quite useful for retaining errors when sockets are closed (see comment on lines 769 - 774 for more info).

1st1:
Closing in favor of https://github.com/python/cpython/pull/16482

PR #16456, opened 2019-09-28
bpo-36670: fixed encoding issue with libregrtest on win systems

LorenzMende:
alternative solution to [PR 15488](https://github.com/python/cpython/pull/15488) - fixing encoding issue with libregrtest on win systems (non english), load of default 0.0 is returned

Was not easily able to write early (instantiation) raise, instead adding of the encoding works fine.
I am unsure about the closing of the write end of the pipe (l. 62) -> tests fail if the pipe is closed, so I uncommented it.


vstinner:
This change only hides the bug, but i doesn't fix it.

I merged PR #16511 which fix this bug, but also multiple other regrtest bugfixes.

PR #16457, opened 2019-09-28
 bpo-37404: Raising value error if an SSLSocket is passed to asyncio functions

idomic:
https://bugs.python.org/issue37404

### DIFF
--- a/Lib/asyncio/selector_events.py
+++ b/Lib/asyncio/selector_events.py
@@ -349,7 +349,7 @@ class BaseSelectorEventLoop(base_events.BaseEventLoop):
         nbytes.
         """
         if isinstance(sock, ssl.SSLSocket):
-            raise TypeError("Socket cannot be of type SSLSocket instance")
+            raise TypeError("Socket cannot be of type SSLSocket")
         if self._debug and sock.gettimeout() != 0:
             raise ValueError("the socket must be non-blocking")
         try:
@@ -389,7 +389,7 @@ class BaseSelectorEventLoop(base_events.BaseEventLoop):
         The return value is the number of bytes written.
         """
         if isinstance(sock, ssl.SSLSocket):
-            raise TypeError("Socket cannot be of type SSLSocket instance")
+            raise TypeError("Socket cannot be of type SSLSocket")
         if self._debug and sock.gettimeout() != 0:
             raise ValueError("the socket must be non-blocking")
         try:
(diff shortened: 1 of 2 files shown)


idomic:
Adding @taleinat as a reviewer

taleinat:
I'm not very familiar with asyncio, but it seems to me that in these cases the code should raise an exception rather than emit a warning. A `TypeError` exception would be appropriate.

1st1:
Agree with @taleinat, a `TypeError` is better than a warning.

asvetlov:
Agree as well.
The code doesn't work with incorrect socket type anyway, using a strict exception is correct in this case.

idomic:
> Agree as well.
> The code doesn't work with incorrect socket type anyway, using a strict exception is correct in this case.

Hi all, would fix it to throw a TypeError exception, I was following the thread in bpo - there was a discussion about sending a warning first instead of just throwing an exception without any notice first.

Will commit a new version soon!

idomic:
Commited the new version: raising TypeError instead of a warning.

idomic:
@taleinat Maybe we can push this one as well? (been pending 2 months)

Lib/asyncio/selector_events.py
```diff
@@ -348,6 +348,8 @@ def remove_writer(self, fd):
         The maximum amount of data to be received at once is specified by
         nbytes.
         """
+        if isinstance(sock, ssl.SSLSocket):
+            raise TypeError("Socket cannot be of type SSLSocket instance")
```
taleinat: "of type SSLSocket instance" is strange. This should say either "of type SSLSocket" or "an SSLSocket instance".

(below as well)

taleinat (changes requested):

taleinat:
@idomic 

> @taleinat Maybe we can push this one as well? (been pending 2 months)

Sorry for letting this languish, that's definitely on me! Thanks for the reminder.

For this, we'll also need to add a NEWS entry. You can use [blurb-it](https://blurb-it.herokuapp.com/) to do so easily in the browser, or install locally with `pip install blurb` and run `blurb add`.

taleinat (approved):
LGTM

taleinat:
@1st1, @asvetlov, care for a quick review before I merge this?

asvetlov (approved):

PR #16459, opened 2019-09-28
bpo-38302. __rpow__ now called when __ipow__ returns NotImplemented

ashkop:
Basically copied the behavior of `binary_iop` for the ternary operator. Also made sure that `op_name` is used in error message.

https://bugs.python.org/issue38302


Objects/abstract.c
```diff
@@ -1216,7 +1235,7 @@ PyNumber_InPlacePower(PyObject *v, PyObject *w, PyObject *z)
 {
     if (v->ob_type->tp_as_number &&
         v->ob_type->tp_as_number->nb_inplace_power != NULL) {
-        return ternary_op(v, w, z, NB_SLOT(nb_inplace_power), "**=");
+        return ternary_iop(v, w, z, NB_SLOT(nb_inplace_power), NB_SLOT(nb_power), "**=");
```
brandtbucher: `PyNumber_InPlacePower` can now just mirror the other `PyNumber_InPlace*` functions, right?

So this is the only line that's needed here.

Lib/test/test_descr.py
```diff
@@ -3900,6 +3900,18 @@ def __ipow__(self, other):
         a = C()
         a **= 2
 
+    def test_ipow_returns_not_implemented(self):
+        class A:
+            def __ipow__(self, other):
```
brandtbucher, line 3908: It looks like we're also lacking a test for when `A` doesn't define `__ipow__`. Do you mind adding one?

brandtbucher (changes requested):
Thanks, this looks good! Just noticed a couple of small things:

Objects/abstract.c
ashkop: Ah yes, thank you.

Lib/test/test_descr.py
ashkop: Sure, added the test

hongweipeng:
Why inplace_pow is a ternary function rather than a binary function? In cevel.c, when processing `**=` or `**`, always set the third parameter to `None`. https://github.com/python/cpython/blob/9a7d9519506ae807ca48ff02e2ea117ebac3450e/Python/ceval.c#L1673-L1676

In the `operator` library, the `__ipow__` passes only two arguments, not three.

Lib/test/test_descr.py
```diff
@@ -3900,6 +3900,31 @@ def __ipow__(self, other):
...
+
+    def test_no_ipow(self):
+        class A:
+            pass
+
+        class B:
+            def __rpow__(self, other):
+                return 1
+
+        a = A()
```
brandtbucher: This can just be:

```suggestion
        a = object()
```

Then you don't need the class definition above!

brandtbucher (changes requested):
@ashkop, one more small thing in the new test. Then it looks good!

Lib/test/test_descr.py
ashkop: Done.

brandtbucher (approved):

serhiy-storchaka (approved):
LGTM. :+1:

Lib/test/test_descr.py
ammaraskar: Could you add a test for the two fallbacks and ensure the same class is returning `NotImplemented` and implementing the regular power methods i.e something like:

```python
class A:
    def __ipow__(self, other):
        return NotImplemented

class B(A):
    def __pow__(self, other):
        return 1
class C(A):
    def __rpow__(self, other):
        return 2
```

The linked bug is only exhibited when a class has an ipow slot but when called it returns `NotImplemented`.

Misc/NEWS.d/next/Core and Builtins/2019-09-28-12-23-23.bpo-38302.hsCNgX.rst
```diff
@@ -0,0 +1 @@
+If :func:`object.__ipow__` returns :const:`NotImplemented` :func:`object.__rpow__` will be called now as expected.
```
ammaraskar: Just fyi, @brettcannon posted on the original bug pointing out that this would not call `__pow__` either. I'd recommend changing this to

```rst
If :func:`object.__ipow__` returns :const:`NotImplemented`, the operator will correctly
fall back to :func:`object.__pow__` and :func:`object.__rpow__` as expected.
```

It might also be worth mentioning this in whatsnew, even though it's obscure it is potentially a change in behavior that people should be informed about.

Objects/abstract.c
```diff
@@ -965,8 +965,9 @@ ternary_op(PyObject *v,
     if (z == Py_None)
         PyErr_Format(
             PyExc_TypeError,
-            "unsupported operand type(s) for ** or pow(): "
+            "unsupported operand type(s) for %.100s: "
```
ammaraskar, line 956: Thanks for fixing this! Really weird that `op_name` went unused so long. Would you mind adding a test for this like:

```python
        x = None
        with self.assertRaises(TypeError) as cm:
            x **= 2
        self.assertIn('unsupported operand type(s) for **=', str(cm.exception))
```

ammaraskar (changes requested):

Objects/abstract.c
ashkop: Thanks, added a new test

Misc/NEWS.d/next/Core and Builtins/2019-09-28-12-23-23.bpo-38302.hsCNgX.rst
ashkop: Updated the news entry, and also added a line to What's New

Lib/test/test_descr.py
ashkop: Not sure I understood correctly what you meant, but I updated the test to also test the `__pow__` fallback.

ammaraskar: Thank you, what you have now looks good. What I meant was that to catch the original bug/properly be a regression test, the class has to look like:

```python
class A:
    def __ipow__(self, _):
        return NotImplemented
    def __pow__(self, other):
        ...
```

if the class looks like

```python
class A:
    def __pow__(self, other):
        ...
```
then the old logic returns a false here and correctly goes to the non-inplace version of power anyway: https://github.com/python/cpython/blob/8e19c8be87017f6bef8e4c936b1e6ddacb558ad2/Objects/abstract.c#L1162-L1163

ammaraskar (approved):

Lib/test/test_descr.py
ashkop: Ah yes, you are right. Thanks!

ashkop:
Merged upstream, added `Py_TEST` usage, removed `TERNARY_OP` definition since `op_name` is now used in the error message.

brettcannon (approved):

brettcannon:
I also tested this PR against my own test suite that uncovered this issue and it passed!

brettcannon:
@ashkop thanks for the PR!

PR #16463, opened 2019-09-28
bpo-34344 Fixed AbstractEventLoopPolicy.get_event_loop docstrings

idomic:
https://bugs.python.org/issue34344

### DIFF
--- a/Lib/asyncio/events.py
+++ b/Lib/asyncio/events.py
@@ -631,6 +631,7 @@ class BaseDefaultEventLoopPolicy(AbstractEventLoopPolicy):
 
     def get_event_loop(self):
         """Get the event loop for the current context.
+
            Returns an instance of EventLoop or raises an exception.
         """
         if (self._local._loop is None and
--- a/Lib/asyncio/events.py
+++ b/Lib/asyncio/events.py
@@ -631,7 +631,7 @@ class BaseDefaultEventLoopPolicy(AbstractEventLoopPolicy):
 
     def get_event_loop(self):
         """Get the event loop for the current context.
-           Returns an instance of EventLoop or raises an exception.
+        Returns an instance of EventLoop or raises an exception.
         """
         if (self._local._loop is None and
                 not self._local._set_called and
--- a/Lib/asyncio/events.py
+++ b/Lib/asyncio/events.py
@@ -577,10 +577,13 @@ class AbstractEventLoopPolicy:
     """Abstract policy for accessing the event loop."""
 
     def get_event_loop(self):
+        """Get the event loop for the current context.
 
-        """ Get the event loop for the current context.
-            Returns an instance of EventLoop or raises an exception.
-       """
+        Returns an event loop object implementing the BaseEventLoop interface,
+        or raises an exception in case no event loop has been set for the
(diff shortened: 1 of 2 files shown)


Lib/asyncio/events.py
```diff
@@ -577,13 +577,10 @@ class AbstractEventLoopPolicy:
...
 
     def get_event_loop(self):
-        """Get the event loop for the current context.
 
-        Returns an event loop object implementing the BaseEventLoop interface,
-        or raises an exception in case no event loop has been set for the
-        current context and the current policy does not specify to create one.
-
-        It should never return None."""
+        """ Get the event loop for the current context.
```
brandtbucher: No leading space here:

```suggestion
        """Get the event loop for the current context.
```

Lib/asyncio/events.py
```diff
@@ -577,13 +577,10 @@ class AbstractEventLoopPolicy:
...
+            Returns an instance of EventLoop or raises an exception.
```
brandtbucher: No hanging indent here:

```suggestion
         Returns an instance of EventLoop or raises an exception.
```

Lib/asyncio/events.py
```diff
@@ -577,13 +577,10 @@ class AbstractEventLoopPolicy:
...
+       """
```
brandtbucher: Bad indent here:

```suggestion
        """
```

brandtbucher (changes requested):
Thanks @idomic, and welcome to CPython!

This change looks good, but I think it was made in the wrong place. This new docstring should be moved to the `get_event_loop` method on **line 632** of this same file. The current docstring for line 580 is perfectly fine as-is.

Besides that, I've got a couple of formatting fixes that should be incorporated. See below:

idomic:
Thanks @brandtbucher for the warm welcome and the detailed comments.
I've removed the past change and fixed the remarks on the docstring at line 632.

Lib/asyncio/events.py
```diff
@@ -630,9 +630,8 @@ def __init__(self):
         self._local = self._Local()
 
     def get_event_loop(self):
-        """Get the event loop.
-
-        This may be None or an instance of EventLoop.
+        """Get the event loop for the current context.
+           Returns an instance of EventLoop or raises an exception.
```
brandtbucher: Looks good @idomic! One more tiny formatting fix (hanging indent):

```suggestion
        Returns an instance of EventLoop or raises an exception.
```

brandtbucher (approved):

Lib/asyncio/events.py
```diff
@@ -630,9 +630,8 @@ def __init__(self):
         self._local = self._Local()
 
     def get_event_loop(self):
-        """Get the event loop.
-
-        This may be None or an instance of EventLoop.
+        """Get the event loop for the current context.
+        Returns an instance of EventLoop or raises an exception.
```
1st1, line 635: Need a newline before "Returns".  See https://www.python.org/dev/peps/pep-0257/ for the reference.

1st1 (approved):

brandtbucher:
Congrats on your first PR @idomic!

PR #16466, opened 2019-09-28
bpo-38307:completes the Stack implementation to yield ending line for each class.

kebab-mai-haddi:
38307: Provide class' end line as well in order to provide the scope of the class in a module.

https://bugs.python.org/issue38307

### DIFF
--- a/Lib/pyclbr.py
+++ b/Lib/pyclbr.py
@@ -43,6 +43,7 @@ import sys
 import importlib.util
 import tokenize
 from token import NAME, DEDENT, OP
+import inspect
 
 __all__ = ["readmodule", "readmodule_ex", "Class", "Function"]
 
@@ -51,7 +52,6 @@ _modules = {}  # Initialize cache of modules we've seen.
 
 class _Object:
     "Information about Python class or function."
-
     def __init__(self, module, name, file, lineno, parent):
         self.module = module
         self.name = name
@@ -84,13 +84,8 @@ class Class(_Object):
 
 
 class Stack(list):
-
-    def __init__(self):
-        import inspect
-        super().__init__()
-
+    
     def __delitem__(self, key):
-        import inspect
         frames = inspect.stack()
         setattr(self[key][0], 'endline',
                 frames[1].frame.f_locals['start'][0] - 1)
@@ -182,8 +177,7 @@ def _readmodule(module, path, inpackage=None):
         search_path = path + sys.path
     spec = importlib.util._find_spec_from_path(fullmodule, search_path)
     if spec is None:
-        raise ModuleNotFoundError(
-            f"no module named {fullmodule!r}", name=fullmodule)
+        raise ModuleNotFoundError(f"no module named {fullmodule!r}", name=fullmodule)
     _modules[fullmodule] = tree
     # Is module a package?
     if spec.submodule_search_locations is not None:
@@ -215,8 +209,7 @@ def _create_tree(fullmodule, path, fname, source, tree, inpackage):
     """
     f = io.StringIO(source)
 
-    # stack = []  # Initialize stack of (class, indent) pairs.
-    stack = Stack()  # changing the source code so as to get the ending line
+    stack = Stack()  # Initialize stack of (class, indent) pairs.
     g = tokenize.generate_tokens(f.readline)
     try:
         for tokentype, token, start, _end, _line in g:
@@ -249,14 +242,14 @@ def _create_tree(fullmodule, path, fname, source, tree, inpackage):
                     del stack[-1]
                 tokentype, class_name, start = next(g)[0:3]
                 if tokentype != NAME:
-                    continue  # Skip class with syntax error.
+                    continue # Skip class with syntax error.
                 # Parse what follows the class name.
                 tokentype, token, start = next(g)[0:3]
                 inherit = None
                 if token == '(':
-                    names = []  # Initialize list of superclasses.
+                    names = [] # Initialize list of superclasses.
                     level = 1
-                    super = []  # Tokens making up current superclass.
+                    super = [] # Tokens making up current superclass.
                     while True:
                         tokentype, token, start = next(g)[0:3]
                         if token in (')', ',') and level == 1:
@@ -293,7 +286,7 @@ def _create_tree(fullmodule, path, fname, source, tree, inpackage):
                 if stack:
                     cur_obj = stack[-1][0]
                     cur_class = _nest_class(
-                        cur_obj, class_name, lineno, inherit)
+                            cur_obj, class_name, lineno, inherit)
                 else:
                     cur_class = Class(fullmodule, class_name, inherit,
                                       fname, lineno)
@@ -399,7 +392,7 @@ def _main():
     else:
         path = []
     tree = readmodule_ex(mod, path)
-    def lineno_key(a): return getattr(a, 'lineno', 0)
+    lineno_key = lambda a: getattr(a, 'lineno', 0)
     objs = sorted(tree.values(), key=lineno_key, reverse=True)
     indent_level = 2
     while objs:
@@ -421,7 +414,6 @@ def _main():
                   .format(' ' * obj.indent, obj.name, obj.super, obj.lineno))
         elif isinstance(obj, Function):
             print("{}def {} {}".format(' ' * obj.indent, obj.name, obj.lineno))
-
 
 if __name__ == "__main__":
     _main()
--- a/Lib/pyclbr.py
+++ b/Lib/pyclbr.py
@@ -66,14 +66,12 @@ class _Object:
 
 class Function(_Object):
     "Information about a Python function, including methods."
-
     def __init__(self, module, name, file, lineno, parent=None):
         _Object.__init__(self, module, name, file, lineno, parent)
 
 
 class Class(_Object):
     "Information about a Python class."
-
     def __init__(self, module, name, super, file, lineno, parent=None):
         _Object.__init__(self, module, name, file, lineno, parent)
         self.super = [] if super is None else super
@@ -100,13 +98,11 @@ def _nest_function(ob, func_name, lineno):
         ob._addmethod(func_name, lineno)
     return newfunc
 
-
 def _nest_class(ob, class_name, lineno, super=None):
     "Return a Class after nesting within ob."
     newclass = Class(ob.module, class_name, super, ob.file, lineno, ob)
     ob._addchild(class_name, newclass)
     return newclass
-
 
 def readmodule(module, path=None):
     """Return Class objects for the top-level classes in module.
@@ -120,7 +116,6 @@ def readmodule(module, path=None):
             res[key] = value
     return res
 
-
 def readmodule_ex(module, path=None):
     """Return a dictionary with all functions and classes in module.
 
@@ -129,7 +124,6 @@ def readmodule_ex(module, path=None):
     Do this by reading source, without importing (and executing) it.
     """
     return _readmodule(module, path or [])
-
 
 def _readmodule(module, path, inpackage=None):
     """Do the hard work for readmodule[_ex].
@@ -210,6 +204,7 @@ def _create_tree(fullmodule, path, fname, source, tree, inpackage):
     f = io.StringIO(source)
 
     stack = Stack()  # Initialize stack of (class, indent) pairs.
+
     g = tokenize.generate_tokens(f.readline)
     try:
         for tokentype, token, start, _end, _line in g:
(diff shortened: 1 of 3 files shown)


the-knights-who-say-ni:

## CLA Missing

@avisrivastava254084

avisrivastava254084:
@the-knights-who-say-ni  I have done the CLA. When will it get updated?

brandtbucher:
Thanks @avisrivastava254084, and welcome to CPython!

Regarding the CLA, the link the bot provided says the following:

> After signing the CLA, please **wait at least one US business day** and then check the status by going to the check-python-cla website. The check will also be run automatically the next time you push changes to your PR.

It's my understanding that these signatures are processed by a human at the PSF, so it probably won't happen on a Saturday night / Sunday morning!

Do you mind updating the PR title to with "bpo-" in front of the issue number, so this can be linked appropriately?

Lib/pyclbr.py
```diff
@@ -51,6 +51,7 @@
 
 class _Object:
     "Information about Python class or function."
+
```
brandtbucher: Please remove all these whitespace lines that you've added.

Lib/pyclbr.py
```diff
@@ -80,6 +83,20 @@ def _addmethod(self, name, lineno):
         self.methods[name] = lineno
 
 
+class Stack(list):
+
+    def __init__(self):
```
brandtbucher: This `__init__` method isn't needed, since all it does is import `inspect` into the function's scope.

Lib/pyclbr.py
```diff
@@ -161,7 +182,8 @@ def _readmodule(module, path, inpackage=None):
         search_path = path + sys.path
     spec = importlib.util._find_spec_from_path(fullmodule, search_path)
     if spec is None:
-        raise ModuleNotFoundError(f"no module named {fullmodule!r}", name=fullmodule)
+        raise ModuleNotFoundError(
```
brandtbucher: Please don't reformat this line. It's fine as-is!

Lib/pyclbr.py
```diff
@@ -227,14 +249,14 @@ def _create_tree(fullmodule, path, fname, source, tree, inpackage):
                     del stack[-1]
                 tokentype, class_name, start = next(g)[0:3]
                 if tokentype != NAME:
-                    continue # Skip class with syntax error.
+                    continue  # Skip class with syntax error.
```
brandtbucher: Please revert this whitespace change as well. Below also...

Lib/pyclbr.py
```diff
@@ -271,7 +293,7 @@ def _create_tree(fullmodule, path, fname, source, tree, inpackage):
                 if stack:
                     cur_obj = stack[-1][0]
                     cur_class = _nest_class(
-                            cur_obj, class_name, lineno, inherit)
```
brandtbucher: Please revert this.

Lib/pyclbr.py
```diff
@@ -377,7 +399,7 @@ def _main():
     else:
         path = []
     tree = readmodule_ex(mod, path)
-    lineno_key = lambda a: getattr(a, 'lineno', 0)
```
brandtbucher: Unidiomatic, sure. But fine as-is, and unrelated to the change. Please revert.

Lib/pyclbr.py
```diff
@@ -193,8 +215,8 @@ def _create_tree(fullmodule, path, fname, source, tree, inpackage):
     """
     f = io.StringIO(source)
 
-    stack = [] # Initialize stack of (class, indent) pairs.
-
+    # stack = []  # Initialize stack of (class, indent) pairs.
```
brandtbucher: Please remove this rather than just commenting it out. It makes for a better diff.

```suggestion

```

Lib/pyclbr.py
```diff
@@ -193,8 +215,8 @@ def _create_tree(fullmodule, path, fname, source, tree, inpackage):
...
+    stack = Stack()  # changing the source code so as to get the ending line
```
brandtbucher: This comment belongs in a commit message, not in the source code. The old comment is fine.

```suggestion
    stack = Stack()  # Initialize stack of (class, indent) pairs.
```

brandtbucher (changes requested):
@avisrivastava254084: this is a very noisy patch. The real changes only affect about 8 lines, but you've changed over 30. Please revert all of the unrelated formatting/whitespace changes (as well as a few others) that I've highlighted in the review below.

Do you mind explaining why this is necessary? In particular, I'm concerned about:

- ...why we need `endline`. Do you have a specific use case?
- ...if `frames[1].frame.f_locals['start'][0] - 1` is the correct way to calculate this.
- ...if doing it in `__delitem__` of a custom `list` subclass is the right way to do this (I'm almost certain it's not).

At a minimum, this will need tests and a [NEWS entry](https://blurb-it.herokuapp.com/).

avisrivastava254084:
@brandtbucher 
Hi! Sorry for a noisy patch, had to work asap.

> .why we need endline. Do you have a specific use case?
So that the class' scope is known. Right now, class' start line is known but not the end line. If we have the end line, we can know the modules imported in b/w start and end. So, we can then deduce that class A used modules: B and C. In this way, we can say that A is dependent on B and C. 

This is useful in drawing and generating UML diagrams, especially in knowing which class is dependent on which classes.

I"ll make the rest of the changes as suggested.

avisrivastava254084:
>...if frames[1].frame.f_locals['start'][0] - 1 is the correct way to calculate this.
I'd want your suggestions or some guiding material that you think I should go through.
>...if doing it in __delitem__ of a custom list subclass is the right way to do this (I'm almost certain it's not).
Same as above.

brandtbucher:
Rather than subclassing `list` and overriding `__delitem__`, just doing `stack[-1].endline = start - 1` before each `del stack[-1]` should suffice. Also, in keeping consistency with other places in the stdlib where this info is used, I would lean towards the spelling `end_lineno`.

A couple of other things: *if* this is accepted, it will need tests, updated docs, and a [NEWS entry](https://blurb-it.herokuapp.com/). All of the classes and comments in this file should also reference the new attribute. And please change the title of this PR to start with "bpo-38307:" so that the bot (and humans) can find your issue!

the-knights-who-say-ni:

## Recognized GitHub username

@kebab-mai-haddi

kebab-mai-haddi:
@brandtbucher 
Apologies for the delay. I have made the changes at the code level as you suggested.

I was wondering if you could help me with the "CLA Not Signed" issue?
I changed my username from avisrivastava254084 to kebab-mai-haddi.

kebab-mai-haddi:
@brandtbucher 
I have made the changes. Is the current implementation alright? If so, can you guide me to the testing phase?

Lib/pyclbr.py
```diff
@@ -202,11 +202,13 @@ def _create_tree(fullmodule, path, fname, source, tree, inpackage):
                 lineno, thisindent = start
                 # Close previous nested classes and defs.
                 while stack and stack[-1][1] >= thisindent:
+                    setattr(stack[-1][0], 'end_lineno', start[0] - 1)
```
brandtbucher: Why did you switch to using `setattr` here? This should be fine, right?

```suggestion
                    stack[-1][0].end_lineno = start[0] - 1
```

Also on the two other lines, below...

brandtbucher (changes requested):
Again, this still needs tests (in `Lib/test/test_pyclbr.py`) and [a NEWS entry](https://blurb-it.herokuapp.com/). The documentation and docstrings will also need to be updated (basically, anywhere the current `lineno` attribute is referenced).

Regarding the CLA, I think you will need to link your new GitHub username to your BPO account and re-sign the CLA. Not sure, though.

Also, just so you don't get discouraged: it's great that you're taking the time to contribute to CPython, but this module has no current experts or maintainers. I am not able to merge this PR once it's ready, and there is a fair chance that no active core developer would feel comfortable adding a new feature like this - especially without significant demand.

csabella:
@kebab-mai-haddi, on the bug tracker, you should be able to edit your GitHub profile name.  Select 'Your details' on the left and it should take you to a screen where you can update your info.

Lib/pyclbr.py
kebab-mai-haddi: I have to set the attribute, right? Otherwise, it results into the error: <> has no attribute end_lineno.

kebab-mai-haddi: >Also, just so you don't get discouraged: it's great that you're taking the time to contribute to CPython, but this module has no current experts or maintainers. I am not able to merge this PR once it's ready, and there is a fair chance that no active core developer would feel comfortable adding a new feature like this - especially without significant demand.

I received some positive responses on Stackoverflow for this. Don't you think there should be a starting and an end line for this?

Even if you think and abide by to go with the popular demand, should I opt for a survey, etc? Is there some platform where I can ask for votes: "Should there be end_lineno?"

Thank you

brandtbucher:
> I received some positive responses on Stackoverflow for this. Don't you think there should be a starting and an end line for this?
>
> Even if you think and abide by to go with the popular demand, should I opt for a survey, etc? Is there some platform where I can ask for votes: "Should there be end_lineno?"

Sorry, I should have been clearer. I myself am not really opposed to adding this feature; I just don't have the necessary permissions to actually merge this PR, and I've never even seen this module before. Once I deem that it's ready, my next step is usually to request the review of a maintainer or owner of this module who does have the ability to merge it. It doesn't look like we have anybody like that, though.

If you wanted to get somebody's attention, resonding to Steven's questions on the BPO issue would be a great place to start (actually, it should probably happen regardless). If that fizzles out, you could also make a post on [the Python-Ideas mailing list](https://mail.python.org/mailman3/lists/python-ideas.python.org) outlining your use case, and gathering feedback. I would also link this PR and the open BPO issue there, as well.

Lib/pyclbr.py
brandtbucher: > I have to set the attribute, right? Otherwise, it results into the error: <> has no attribute end_lineno.

They should both be equivalent. You're certain that applying my suggested change results in an error?

kebab-mai-haddi: Yes. My bad, apologies. I tried the direct approach to `stack[-1]` instead of `stack[-1][0]`.

kebab-mai-haddi:
@brandtbucher 
I have added the NEWS entry. Regarding tests, I tried the following:
```
t1 = type(o1), o1.name, o1.file, o1.module, o1.lineno, o1.end_lineno
t2 = type(o2), o2.name, o2.file, o2.module, o2.lineno, o2.end_lineno
```
the actual code of which is available in the diff of this branch.

The test is failing as the children being compared are of different types. The error is:
```
ERROR: test_nested (test.test_pyclbr.PyclbrTest)
  File "/Users/aviralsrivastava/dev/cpython/Lib/test/test_pyclbr.py", line 219, in test_nested
    compare(None, actual, None, expected)
  File "/Users/aviralsrivastava/dev/cpython/Lib/test/test_pyclbr.py", line 212, in compare
    t2 = type(o2), o2.name, o2.file, o2.module, o2.lineno, o2.end_lineno
AttributeError: 'Function' object has no attribute 'end_lineno'

Ran 6 tests in 5.524s
Merge branch 'endline-in-readmodule-module' of github.com:avisrivastava254084/cpython into endline-in-readmodule-module

FAILED (errors=1)
test test_pyclbr failed
1 test failed again:
    test_pyclbr

1 test failed:
    test_pyclbr

1 re-run test:
    test_pyclbr

Total duration: 11 sec 993 ms
Tests result: FAILURE then FAILURE
make: *** [test] Error 2
```

Now, I am failing to understand what and how should I write the test for the new change of `end_lineno`?
Any direction/lead would be greatly appreciated.

brandtbucher:
@kebab-mai-haddi I'm not too familiar with this code, but a cursory scan tells me this:

`pyclbr` creates two types of objects from modules:
- a `Class` for every `class`
- a `Function` for every `def`

I believe your patch only adds the new `end_lineno` attribute to `Class`. Likely, you also want to add it to `Function` as well (for symmetry and completeness).

kebab-mai-haddi:
> @kebab-mai-haddi I'm not too familiar with this code, but a cursory scan tells me this:
> 
> `pyclbr` creates two types of objects from modules:
> 
> * a `Class` for every `class`
> * a `Function` for every `def`
> 
> I believe your patch only adds the new `end_lineno` attribute to `Class`. Likely, you also want to add it to `Function` as well (for symmetry and completeness).

I have done that. If you check [here](https://github.com/python/cpython/pull/16466/files#diff-11e281561648f5044557f8cc970cae6fR211), you will see that if the object is of type `Function`, I am adding `end_lineno`.

Thank you in advance!

PR #16468, opened 2019-09-28
Update functions.rst

geryogam:
This PR will make the following changes to the [_Built-in Functions_](https://docs.python.org/3/library/functions.html) chapter of the library documentation:

- improve hyperlinks in Sphinx roles (trailing 's' belong to hyperlinks).

Automerge-Triggered-By: @csabella

### DIFF
--- a/Doc/library/functions.rst
+++ b/Doc/library/functions.rst
@@ -1625,7 +1625,7 @@ are always available.  They are listed here in alphabetical order.
    not found in statically compiled languages or languages that only support
    single inheritance.  This makes it possible to implement "diamond diagrams"
    where multiple base classes implement the same method.  Good design dictates
-   that this method has the same calling signature in every case (because the
+   that this method have the same calling signature in every case (because the
    order of calls is determined at runtime, because that order adapts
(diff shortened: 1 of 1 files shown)


Doc/library/functions.rst
```diff
@@ -1625,7 +1625,7 @@ are always available.  They are listed here in alphabetical order.
    not found in statically compiled languages or languages that only support
    single inheritance.  This makes it possible to implement "diamond diagrams"
    where multiple base classes implement the same method.  Good design dictates
-   that this method have the same calling signature in every case (because the
```
ammaraskar: I don't think this is a typo. "have" makes sense here.

Consider:

> It ought to have the same signature

vs

> It ought to has the same signature

geryogam: My fault, it was indeed correct. I have updated the PR. Thanks for the review @ammaraskar.

geryogam: FYI @brandtbucher.

csabella (approved):
lgtm

PR #16469, opened 2019-09-28
bpo-38311: Avoid sqlite3_enable_shared_cache deprecation warning when building with the macOS system SQLite3

devnexen:
sqlite shared cached is disabled since the 10.7 release (2011).

https://bugs.python.org/issue38311

### DIFF
--- a/Modules/_sqlite/module.c
+++ b/Modules/_sqlite/module.c
@@ -131,7 +131,7 @@ PyDoc_STRVAR(module_complete_doc,
 \n\
 Checks if a string contains a complete SQL statement. Non-standard.");
 
-#ifdef HAVE_SHARED_CACHE
+#if defined(HAVE_SHARED_CACHE) && !defined(__APPLE__)
 static PyObject* module_enable_shared_cache(PyObject* self, PyObject* args, PyObject*
         kwargs)
 {


devnexen:
no news

Modules/_sqlite/module.c
```diff
@@ -131,7 +131,7 @@ PyDoc_STRVAR(module_complete_doc,
 \n\
 Checks if a string contains a complete SQL statement. Non-standard.");
 
-#ifdef HAVE_SHARED_CACHE
+#if defined(HAVE_SHARED_CACHE) && !defined(__APPLE__)
```
ned-deily, line 134: This test is not sufficient because it will prevent _sqlite from being built with a local copy of SQLite rather than the Apple-supplied system version and that is something that many distributors of Python on macOS do, including the python.org macOS installers.  So, to avoid the deprecation warning, you want to be able to know you are linking with the system version.  That might be a little tricky to do reliably; it might have to be done in either configure.ac or setup.py and would need to take into account SDKs.  And in most cases it is probably better to avoid using the macOS system SQLite anyway.  So i'm not sure this is worth trying to fix.

ned-deily (changes requested):
Thanks for the PR.  Unfortunately as noted, i'm not sure this is worth trying to fix.  But if you want to take another stab at it, we can review it.

Modules/_sqlite/module.c
ned-deily: To be more precise, it would not prevent _sqlite from being built; it would disallow use of `enable_shared_cache`.  Which, BTW, doesn't seem to be documented in the sqlite3 module doc page?

ned-deily:
Sorry, I just found an earlier open issue, bpo-24464, covering the same topic and with a similar proposed change.  I've closed bpo-38311 as a duplicate of it; let's continue further discussion there.

devnexen:
No problems I understand your reasons, thanks.

PR #16472, opened 2019-09-29
bpo-38019: correctly handle pause/resume reading of closed asyncio unix pipe

asvetlov:
https://bugs.python.org/issue38019


asvetlov:
Now these methods are handled the same way as sockets in `selector_events.py`

PR #16475, opened 2019-09-29
[3.5] bpo-38216, bpo-36274: Allow subclasses to separately override validation and encoding behavior (GH-16448)

jaraco:
* [bpo-38216](https://bugs.python.org/issue38216): Allow bypassing input validation

* [bpo-36274](https://bugs.python.org/issue36274): Also allow the URL encoding to be overridden.

* [bpo-38216](https://bugs.python.org/issue38216), [bpo-36274](https://bugs.python.org/issue36274): Add tests demonstrating a hook for overriding validation, test demonstrating override encoding, and a test to capture expectation of the interface for the URL.

* Call with skip_host to avoid tripping on the host checking in the URL.

* Remove obsolete comment.

* Make _prepare_path_encoding its own attr.

This makes overriding just that simpler.

Also, don't use the := operator to make backporting easier.

* Add a news entry.

* _prepare_path_encoding -> _encode_prepared_path()

* Once again separate the path validation and request encoding, drastically simplifying the behavior. Drop the guarantee that all processing happens in _prepare_path..
(cherry picked from commit 7774d7831e8809795c64ce27f7df52674581d298)

Co-authored-by: Jason R. Coombs <jaraco@jaraco.com>

### DIFF
--- a/Lib/http/client.py
+++ b/Lib/http/client.py
@@ -1079,8 +1079,11 @@ class HTTPConnection:
         # Prevent CVE-2019-9740.
         match = _contains_disallowed_url_pchar_re.search(url)
         if match:
-            raise InvalidURL(f"URL can't contain control characters. {url!r} "
-                             f"(found at least {match.group()!r})")
+            msg = (
+                "URL can't contain control characters. {url!r} "
+                "(found at least {matched!r})"
+            ).format(matched=match.group(), **locals())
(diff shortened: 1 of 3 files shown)


jaraco:
The remaining test failures in CI seem to be spurious or otherwise unrelated to this PR.

larryhastings:
Thanks for the 3.5 love!

PR #16482, opened 2019-09-30
bpo-38242: Revert "bpo-36889: Merge asyncio streams (GH-13251)"

1st1:
See https://bugs.python.org/issue38242 for more details

99% of the work is done by @aeros167; I've just needed to add a few more fixes and couldn't push to his branch; hence the new PR.

### DIFF
--- a/Doc/library/asyncio-stream.rst
+++ b/Doc/library/asyncio-stream.rst
@@ -18,12 +18,19 @@ streams::
     import asyncio
 
     async def tcp_echo_client(message):
-        async with asyncio.connect('127.0.0.1', 8888) as stream:
-            print(f'Send: {message!r}')
-            await stream.write(message.encode())
+        reader, writer = await asyncio.open_connection(
+            '127.0.0.1', 8888)
 
-            data = await stream.read(100)
-            print(f'Received: {data.decode()!r}')
+        print(f'Send: {message!r}')
+        writer.write(message.encode())
+        await writer.drain()
+
+        data = await reader.read(100)
+        print(f'Received: {data.decode()!r}')
+
+        print('Close the connection')
+        writer.close()
+        await writer.wait_closed()
 
     asyncio.run(tcp_echo_client('Hello World!'))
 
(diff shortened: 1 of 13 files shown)


Lib/asyncio/streams.py
```diff
@@ -766,18 +391,17 @@ def get_extra_info(self, name, default=None):
...
-            # called.  Without this, _drain_helper() would return
-            # immediately, and code that calls
-            #     write(...); await drain()
-            # in a loop would never call connection_lost(), so it
-            # would not see an error when the socket is closed.
-            await tasks.sleep(0, loop=self._loop)
+            # Wait for protocol.connection_lost() call
+            # Raise connection closing error if any,
+            # ConnectionResetError otherwise
+            await sleep(0)
```
aeros, line 403: Change made by @asvetlov that was not related to the streaming API. It fixes an issue that prevented the call to `protocol.connection_lost()`. See comment for details.
```suggestion
            # Yield to the event loop so connection_lost() may be
            # called.  Without this, _drain_helper() would return
            # immediately, and code that calls
            #     write(...); await drain()
            # in a loop would never call connection_lost(), so it
            # would not see an error when the socket is closed.
            await tasks.sleep(0, loop=self._loop)
```
Should we restore this? I don't see an issue with doing so or a reason why it should be removed.

1st1: Yes, I'll restore it. Thanks!

aeros167:
> 99% of the work is done by @aeros167; I've just needed to add a few more fixes and couldn't push to his branch; hence the new PR.

Thanks for finishing this up Yury. Out of curiosity, what additional changes needed to be made?

1st1:
> Thanks for finishing this up Yury. Out of curiosity, what additional changes needed to be made?

`.write()` and `.close()` were still awaitable -- which technically was committed before new streams api. Still we want to revert streams to their 3.7 state.

Doc/library/asyncio-stream.rst
```diff
@@ -87,12 +69,8 @@ and work with streams:
 
       The *ssl_handshake_timeout* parameter.
 
-   .. deprecated-removed:: 3.8 3.10
-
-      `open_connection()` is deprecated in favor of :func:`connect`.
-
 .. coroutinefunction:: start_server(client_connected_cb, host=None, \
-                          port=None, \*, loop=None, limit=2**16, \
+                          port=None, \*, loop=None, limit=None, \
```
aeros, line 73: The constant `_DEFAULT_LIMIT` (2**16) is still being used as the default for *limit* in `start_server()`, so this can be restored.
```suggestion
                          port=None, \*, loop=None, limit=2**16, \
```

aeros167:
> .write() and .close() were still awaitable -- which technically was committed before new streams api. Still we want to revert streams to their 3.7 state.

Ah, I wasn't aware we were also removing that change. I hadn't realized we were reverting fully back to the 3.7 API, I thought it was primarily just focused on https://github.com/python/cpython/commit/23b4b697e5b6cc897696f9c0288c187d2d24bff2. Now it makes a lot more sense why you wanted to remove `_asyncio_internal`.

Doc/library/asyncio-stream.rst
aeros: I can do this in a separate PR if needed. This change is rather low priority in comparison to reverting the streaming API, and would not be a release blocker.

1st1: Sure, although right now docs are the same as in 3.7. So I'm not sure it's needed tbh.

aeros: > So I'm not sure it's needed tbh.

Okay, I won't worry about it then, it's fairly trivial.

PR #16483, opened 2019-09-30
bpo-38045: Improve the performance of _decompose() in the enum.py

hongweipeng:
The original code will access each element of `_value2member_map_` and check if the value is the power of 2.

After optimization, access the elements in `_value2member_map_` as needed. In the initial example:
```
from enum import Flag
F = Flag("F", list("abcdefghijklm"))
for idx in range(2**len(F) - 1):
    F(idx)
```
the speedup is more than 100x .(The same for replacing Flag with IntFlag).

https://bugs.python.org/issue38045

### DIFF
--- a/Misc/NEWS.d/next/Library/2019-09-30-12-09-41.bpo-38045.VDRtd3.rst
+++ b/Misc/NEWS.d/next/Library/2019-09-30-12-09-41.bpo-38045.VDRtd3.rst
@@ -1,1 +1,1 @@
-Improve the performance of _decompose() in the enum.py. Patch by hongweipeng.
+Improve the performance of :func:`enum._decompose` in :mod:`enum`. Patch by hongweipeng.
--- a/Misc/NEWS.d/next/Library/2019-09-30-12-09-41.bpo-38045.VDRtd3.rst
+++ b/Misc/NEWS.d/next/Library/2019-09-30-12-09-41.bpo-38045.VDRtd3.rst
@@ -1,0 +1,1 @@
+Improve the performance of _decompose() in the enum.py. Patch by hongweipeng.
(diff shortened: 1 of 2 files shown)


corona10:
@hongweipeng  Can you show me the benchmark result?

hongweipeng:
Use the initial example:
```
from enum import Flag
F = Flag("F", list("abcdefghijklm"))
for idx in range(2**len(F) - 1):
    F(idx)
```
Before patch:
```
# time ./python ts.py

real	0m29.639s
user	0m29.615s
sys	0m0.020s
```
After patch:
```
# time ./python ts.py

real	0m0.135s
user	0m0.122s
sys	0m0.014s
```

corona10:
@ethanfurman 
 
For core-developers,
I didn't check this PR is the correct way.
But seems like there is performance enhancement with this PR.

I left the benchmark for the core-developers

## Baseline
```` bash
 ./python.exe -m pyperf timeit -s 'from enum import Flag' 'F = Flag("F", list("abcdefghijklm"))' 'for idx in range(100):' '   F(idx)'
.....................
Mean +- std dev: 24.5 ms +- 0.3 ms
````

## PR
````
./python.exe -m pyperf timeit -s 'from enum import Flag' 'F = Flag("F", list("abcdefghijklm"))' 'for idx in range(100):' '   F(idx)'
.....................
Mean +- std dev: 7.30 ms +- 0.19 ms
````

File: Misc/NEWS.d/next/Library/2019-09-30-12-09-41.bpo-38045.VDRtd3.rst [hunk 1]
```diff
@@ -0,0 +1 @@
+Improve the performance of _decompose() in the enum.py. Patch by hongweipeng.
```
brandtbucher:
```suggestion
Improve the performance of :func:`enum._decompose` in :mod:`enum`. Patch by hongweipeng.
```

ethanfurman (approved):

ethanfurman:
GitHub seem stuck on bedevere/news -- closing and reopening...

PR #16485, opened 2019-09-30
[3.8] bpo-38242: Revert "bpo-36889: Merge asyncio streams (GH-13251)" (#16482)

1st1:
See https://bugs.python.org/issue38242 for more details

### DIFF
--- a/Lib/asyncio/streams.py
+++ b/Lib/asyncio/streams.py
@@ -1,19 +1,14 @@
 __all__ = (
-    'Stream', 'StreamMode',
-    'open_connection', 'start_server',
-    'connect', 'connect_read_pipe', 'connect_write_pipe',
-    'StreamServer')
+    'StreamReader', 'StreamWriter', 'StreamReaderProtocol',
+    'open_connection', 'start_server')
 
-import enum
 import socket
 import sys
 import warnings
 import weakref
 
 if hasattr(socket, 'AF_UNIX'):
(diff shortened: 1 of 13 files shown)


Lib/asyncio/streams.py
```diff
@@ -766,18 +391,23 @@ def get_extra_info(self, name, default=None):
...
+            # Raise connection closing error if any,
+            # ConnectionResetError otherwise
             # Yield to the event loop so connection_lost() may be
             # called.  Without this, _drain_helper() would return
             # immediately, and code that calls
             #     write(...); await drain()
             # in a loop would never call connection_lost(), so it
             # would not see an error when the socket is closed.
-            await tasks.sleep(0, loop=self._loop)
+            await sleep(0)
```
aeros, line 403: Same as 3.9 PR.
```suggestion
            # Yield to the event loop so connection_lost() may be
            # called.  Without this, _drain_helper() would return
            # immediately, and code that calls
            #     write(...); await drain()
            # in a loop would never call connection_lost(), so it
            # would not see an error when the socket is closed.
            await tasks.sleep(0, loop=self._loop)
```

1st1: I've actually merged the comment, but the `await sleep(0)` line is correct.

aeros: > but the await sleep(0) line is correct

Oh you're right, there's no reason to explicitly pass the loop argument there, that wasn't intentional.

PR #16494, opened 2019-09-30
bpo-38321: Fix _testcapimodule.c warning

vstinner:
Fix the following warning:

    modules\_testcapimodule.c(6409):
    warning C4146: unary minus operator applied to unsigned type,
    result still unsigned

https://bugs.python.org/issue38321

### DIFF
--- a/Modules/_testcapimodule.c
+++ b/Modules/_testcapimodule.c
@@ -1952,7 +1952,7 @@ unicode_asutf8andsize(PyObject *self, PyObject *args)
         return NULL;
     }
 
-    buffer = PyUnicode_AsUTF8AndSize(unicode, &utf8_len); 
+    buffer = PyUnicode_AsUTF8AndSize(unicode, &utf8_len);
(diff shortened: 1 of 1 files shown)


eduardo-elizondo (approved):
Thanks! LGTM

Modules/_testcapimodule.c
```diff
@@ -1952,7 +1952,7 @@ unicode_asutf8andsize(PyObject *self, PyObject *args)
         return NULL;
     }
 
-    buffer = PyUnicode_AsUTF8AndSize(unicode, &utf8_len);
```
eduardo-elizondo, line 1955: Unrelated change?

eduardo-elizondo (approved):

Modules/_testcapimodule.c
vstinner: Yeah, my editor removes trailing spaces. Trailing spaces must die!

PR #16495, opened 2019-09-30
bpo-38006: Clear weakrefs in garbage found by the GC

nascheme:
Fix a bug due to the interaction of weakrefs and the cyclic garbage
collector. We must clear any weakrefs in garbage in order to prevent
their callbacks from executing and causing a crash.

### DIFF
--- a/Modules/gcmodule.c
+++ b/Modules/gcmodule.c
@@ -678,6 +678,21 @@ handle_weakrefs(PyGC_Head *unreachable, PyGC_Head *old)
         op = FROM_GC(gc);
         next = GC_NEXT(gc);
 
+        if (PyWeakref_Check(op)) {
+            /* A weakref inside the unreachable set must be cleared.  If we
+             * allow its callback to execute inside delete_garbage(), it
+             * could expose objects that have tp_clear already called on
+             * them.  Or, it could resurrect unreachable objects.  One way
+             * this can happen is if some container objects do not implement
+             * tp_traverse.  Then, wr_object can be outside the unreachable
+             * set but can be deallocated as a result of breaking the
+             * reference cycle.  If we don't clear the weakref, the callback
+             * will run and potentially cause a crash.  See bpo-38006 for
+             * one example.
+             */
+            _PyWeakref_ClearRef((PyWeakReference *)op);
+        }
+
         if (! PyType_SUPPORTS_WEAKREFS(Py_TYPE(op)))
             continue;
 
@@ -733,6 +748,8 @@ handle_weakrefs(PyGC_Head *unreachable, PyGC_Head *old)
              * is moved to wrcb_to_call in this case.
              */
             if (gc_is_collecting(AS_GC(wr))) {
+                /* it should already have been cleared above */
+                assert(wr->wr_object == Py_None);
                 continue;
             }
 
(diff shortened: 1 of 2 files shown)


tim-one (approved):
Ship it!  Thank you, Neil.

nascheme:
Should we backport to all maintained branches?  I think it applies even to 2.7.

tim-one:
> Should we backport to all maintained branches?

Ya, these holes have been there since the day weakrefs were added to the language.

pitrou:
Uh, I'm a bit anxious about this. This is added to workaround a crash that's caused by a container not implementing the GC protocol correctly.

But in turn it may silence of invocation of many legitimate weakref callbacks. Weakref callbacks are often used to release external resources (for example system resources). This is actually the original motivation for [weakref.finalize](https://docs.python.org/3/library/weakref.html#weakref.finalize), which was originally a `multiprocessing`-internal feature used to release OS handles and the like.

I may be misreading the patch, but if a `weakref` that's collected as part of a cycle doesn't invoke its callback, then this will inevitably produce silent, difficult-to-diagnose bugs in complex systems.

So I would advocate reverting.

Modules/gcmodule.c
```diff
@@ -733,6 +748,8 @@ handle_weakrefs(PyGC_Head *unreachable, PyGC_Head *old)
              * is moved to wrcb_to_call in this case.
              */
             if (gc_is_collecting(AS_GC(wr))) {
+                /* it should already have been cleared above */
```
pitrou, line 751: One question: how do you know it's "already" been cleared? Since the cycle is a cycle, it could have been collected starting from any point. So `wr` could be further down the `unreachable` list, and not yet handled.

tim-one:
To be sure we're on the same page, it was always the case that we suppressed callbacks from trash weakrefs.  If the user wants the callback to be invoked for sure, it's the user's responsibility to ensure the weakref outlives the referent.  `finalize` is great for making that so.    Nothing about that is changed here.

What did change is that a missing `tp_traverse` tricked us into running a callback from a trash weakref, because we didn't realize it _was_ trash.  This occurred during `delete_garbage()`, after the callback had already been `tp_clear`'ed.  Segfault.

I bet we could run more callbacks now, since you added a pass to abort the gc run if trash got resurrected.  But I never heard anyone ask for that, and don't want it myself 😉

Modules/gcmodule.c
tim-one: It doesn't mean `tp_clear`'ed, it means cleared by `_PyWeakref_ClearRef ()`.  That happened on (new) line 715 above.

pitrou: Yes... But this assumes `wr` (which is one of `op`'s weakrefs) is visited before `op` in the unreachable list. I don't think we have any guarantee for that.

tim-one: Sorry, I still don't grasp the concern.  What do you think can go wrong if `wr` appears after `op` in the unreachable list?  `_PyWeakref_ClearRef(wr)` would end up being called twice, but that's harmless.  This loop isn't doing anything _except_ clearing trash weakrefs.

pitrou: I mean the assert may fail. Am I interpreting it wrong?

pitrou: Oh, I see, I had missed the `_PyWeakref_ClearRef(wr)` call above. I thought that additional assert was related to the change in this PR. Apparently not.

tim-one: Right!  This assert may as well always have been there.  I assume Neil added it because it clarified something his brain tripped over.  I appreciate that you're thinking about it too!

PR #16496, opened 2019-09-30
bpo-38326: Use Python API version as config version

ncoghlan:
This allows the config APIs to detect attempts to embed
all potentially ABI incompatible runtime versions, rather
than only being able to detect incompatibilities in the
size of the PyPreConfig and PyConfig structures.

https://bugs.python.org/issue38326

### DIFF
--- a/Include/cpython/initconfig.h
+++ b/Include/cpython/initconfig.h
@@ -45,9 +45,10 @@ PyAPI_FUNC(PyStatus) PyWideStringList_Insert(PyWideStringList *list,
 /* --- PyPreConfig ----------------------------------------------- */
 
 typedef struct {
-    /* Size of the structure in bytes: must be initialized to
-       sizeof(PyPreConfig). Field used for API and ABI compatibility. */
-    size_t struct_size;
+    /* Version of the CPython header files used to compile the embedding
+       application. Expected to be set to PY_VERSION_HEX.
+       Field is used to check for API and ABI compatibility. */
+    uint32_t header_version;
 
     int _config_init;     /* _PyConfigInitEnum value */
 
@@ -131,9 +132,10 @@ PyAPI_FUNC(PyStatus) PyPreConfig_InitIsolatedConfig(PyPreConfig *config);
 /* --- PyConfig ---------------------------------------------- */
 
 typedef struct {
-    /* Size of the structure in bytes: must be initialized to
-       sizeof(PyConfig). Field used for API and ABI compatibility. */
-    size_t struct_size;
+    /* Version of the CPython header files used to compile the embedding
+       application. Expected to be set to PY_VERSION_HEX.
+       Field is used to check for API and ABI compatibility. */
+    uint32_t header_version;
 
     int _config_init;     /* _PyConfigInitEnum value */
 
--- a/Python/initconfig.c
+++ b/Python/initconfig.c
@@ -529,17 +529,6 @@ Py_GetArgcArgv(int *argc, wchar_t ***argv)
      : _PyStatus_NO_MEMORY())
 
 
-static PyStatus
-config_check_struct_size(const PyConfig *config)
-{
-    if (config->struct_size != sizeof(PyConfig)) {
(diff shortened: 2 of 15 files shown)


Python/pystate.c
```diff
@@ -209,7 +209,8 @@ PyInterpreterState_New(void)
     memset(interp, 0, sizeof(*interp));
     interp->id_refcount = -1;
 
-    interp->config.struct_size = sizeof(PyConfig);
+    // TODO: Should this pass the API/ABI version from the parent interpreter?
+    interp->config.header_version = PY_VERSION_HEX;
```
zooba, line 215: Depends whether this field will become a "behave like this version" (then yes) or just "interpreter this struct like this version" (then no).

Python/initconfig.c
```diff
@@ -579,20 +568,75 @@ PyConfig_Clear(PyConfig *config)
...
     if (_PyStatus_EXCEPTION(status)) {
         _PyStatus_UPDATE_FUNC(status);
         return status;
     }
 
+    // Future enhancement: allow old config struct layouts if an old
+    // header_version is specified
+    size_t struct_size = sizeof(*config);
+    memset(config, 0, struct_size);
+    config->header_version = header_version;
```
ncoghlan, line 638: Note that prior to this change, the old code *didn't* pay attention to the `struct_size` setting (it wasn't used in the `memset` call), and if it had been naively adjusted to do so, could have resulted in clearing an arbitrary amount of memory if the `struct_size` field hadn't been initialised correctly by the caller.

With an explicit API/ABI version passed in, we'd never be trusting the supplied value directly - we'd be using it to select amongst a predefined set of known possible structure sizes.

Python/preconfig.c
```diff
@@ -268,31 +268,22 @@ _PyPreCmdline_Read(_PyPreCmdline *cmdline, const PyPreConfig *preconfig)
...
     if (_PyStatus_EXCEPTION(status)) {
         _PyStatus_UPDATE_FUNC(status);
         return status;
     }
 
+    // Future enhancement: allow old config struct layouts if an old
+    // header_version is specified
+    size_t struct_size = sizeof(*config);
+    memset(config, 0, struct_size);
+    config->header_version = header_version;
```
ncoghlan, line 285: As for PyConfig, the initial code didn't pay attention to the `struct_size` setting (it wasn't used in the `memset` call).

Python/pystate.c
ncoghlan: Yeah, I figure it would only matter if it ever became the former, and in that case, this is one of the places that would need to change.

ambv (approved):
I think this is an improvement. I'd like @vstinner to merge so we're clear there's mutual understanding.

Include/cpython/initconfig.h
```diff
@@ -45,9 +45,10 @@ PyAPI_FUNC(PyStatus) PyWideStringList_Insert(PyWideStringList *list,
 /* --- PyPreConfig ----------------------------------------------- */
 
 typedef struct {
-    /* Size of the structure in bytes: must be initialized to
-       sizeof(PyPreConfig). Field used for API and ABI compatibility. */
-    size_t struct_size;
+    /* Version of the CPython header files used to compile the embedding
+       application. Expected to be set to PY_VERSION_HEX.
+       Field is used to check for API and ABI compatibility. */
+    uint32_t header_version;
```
pitrou, line 51: Why `uint32_t`? Your PEP changes say `size_t`.

vstinner:
I proposed PR #16500 to remove struct_size.

ncoghlan:
Closing in favour of #16500

Include/cpython/initconfig.h
ncoghlan: Typo on the PEP PR side (I renamed struct_size without fixing the type).

Python/initconfig.c
ncoghlan: Just noting that the previous code *didn't* naively trust the supplied value: anything that didn't match the expected value was reported as a size mismatch error.

PR #16500, opened 2019-09-30
[3.8] bpo-38304: Remove PyConfig.struct_size

vstinner:
https://bugs.python.org/issue38304

### DIFF
--- a/Doc/c-api/init_config.rst
+++ b/Doc/c-api/init_config.rst
@@ -194,24 +194,17 @@ PyPreConfig
    * Configure the LC_CTYPE locale
    * Set the UTF-8 mode
 
-   The :c:member:`struct_size` field must be explicitly initialized to
-   ``sizeof(PyPreConfig)``.
-
    Function to initialize a preconfiguration:
 
-   .. c:function:: PyStatus PyPreConfig_InitIsolatedConfig(PyPreConfig *preconfig)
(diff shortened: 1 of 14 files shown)


vstinner:
If this change is merged, it should be forward-ported to the master branch.

vstinner:
See https://mail.python.org/archives/list/python-dev@python.org/thread/C7Z2NA2DTM3DLOZCFQAK5A2WFYO3PHHX/ and https://bugs.python.org/issue38326 for the rationale and discussion.

pitrou (approved):
+1 from me. Leaves us time to discuss any further change without hurrying.

vstinner:
> +1 from me. Leaves us time to discuss any further change without hurrying.

I'm a strong believer that a "stable ABI world" is possible... But we cannot implement everything today before 3.8.0rc1 final cut :-)

It seems like we need more time to properly design the overall use story "provide a stable ABI for embedded Python". PyConfig is just a small piece. I didn't notice that vim is directly linked to libpython3.7.so for example.

pitrou:
The "struct size" field didn't define a stable ABI, so it's moot anyway.

vstinner:
See also PR #16496: @ncoghlan proposes to replace struct_size with header_version.

ncoghlan (approved):
Thanks Victor. This update did convince that a stable embedding APi is more plausible than I previously thought it was, but just before an RC isn't the right time for us to be designing the initial pieces of that feature :)

PR #16502, opened 2019-09-30
[3.8] bpo-33418: Restore tp_clear for function object

nascheme:
This is a revert of the revert (GH-15826, ccaea52588).  Having a tp_clear for
functions should be safe (and helpful) now that [bpo-38006](https://bugs.python.org/issue38006) has been fixed.

https://bugs.python.org/issue33418

### DIFF
--- a/Objects/funcobject.c
+++ b/Objects/funcobject.c
@@ -570,23 +570,31 @@ func_new_impl(PyTypeObject *type, PyCodeObject *code, PyObject *globals,
     return (PyObject *)newfunc;
 }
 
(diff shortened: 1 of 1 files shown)


tim-one (approved):
LGTM!

ambv:
For reference:
* this has never been reverted on master (3c452404ae1 is still active there) so it doesn't need a "forward-port" there
* this feature has been there for all published prereleases of 3.8 so it's pretty well covered with tests, it's safe to restore it now

PR #16503, opened 2019-09-30
bpo-38332: catch KeyError from unknown cte in encoded-word

aft90:
Given an encoded-word with an invalid content-transfer-encoding (something other than `b` or `q`, for instance: `=?utf-8?X?somevalue?=`), `_encoded_words.decode()` throws a `KeyError` which propagates all the way down to `email.message.get()`.

Fix is to explicitly catch the `KeyError` in `get_encoded_word` and raise `InvalidEwError` which then causes `get_unstructured` to return the value unchanged.

If accepted, this should be backported to 3.5 through to 3.8

* https://bugs.python.org/issue38332

https://bugs.python.org/issue38332

### DIFF
--- a/Lib/email/_header_value_parser.py
+++ b/Lib/email/_header_value_parser.py
@@ -1057,7 +1057,7 @@ def get_encoded_word(value):
     value = ''.join(remainder)
     try:
         text, charset, lang, defects = _ew.decode('=?' + tok + '?=')
-    except ValueError:
+    except (ValueError, KeyError):
         raise _InvalidEwError(
             "encoded word format invalid: '{}'".format(ew.cte))
     ew.charset = charset
--- a/Lib/test/test_email/test__encoded_words.py
+++ b/Lib/test/test_email/test__encoded_words.py
@@ -58,6 +58,8 @@ class TestDecode(TestEmailBase):
             _ew.decode('=?')
         with self.assertRaises(ValueError):
             _ew.decode('')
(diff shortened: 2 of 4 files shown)


eamanu:
Hi @aft90, thanks for the PR. Please open a BPO  (https://bugs.python.org) to https://bugs.python.org

aft90:
I have: https://bugs.python.org/issue38332. BPO seems to be unresponsive at the moment, so I'll update it when it's back up.

eamanu (commented):
(copy here my bpo comments)

I am not a email expert, but according to RFC 1342 the enconding can be either "B" or "Q". So, I think is reasonable that when a not correct enconding is set, should be raise an exception

I think that we can improve the message raising a more specific Exception

eamanu (approved):
LGTM. I test it and work successfully. IMO the NEW is not necessary. Good catch @aft90

maxking:
I think the changes look good, although, it would be good to have a NEWS item too. This seems 

It is good to have a changelog for bugfixes in case we need to refer to it later.

the-knights-who-say-ni:

## CLA Missing

@andreitroiebbc

maxking (approved):
Thanks @aft90 for working on this.

It looks good to me!

maxking:
Hmm, not sure why it wasn't merged automatically since it seems to have the right labels. Merging manually!

PR #16509, opened 2019-10-01
bpo-38304: PyConfig_InitPythonConfig() cannot fail anymore

vstinner:
PyConfig_InitPythonConfig() and PyConfig_InitIsolatedConfig() no
longer return PyStatus: they cannot fail anymore.

https://bugs.python.org/issue38304

### DIFF
--- a/Doc/c-api/init_config.rst
+++ b/Doc/c-api/init_config.rst
@@ -343,12 +343,12 @@ PyConfig
 
    Structure methods:
 
-   .. c:function:: PyStatus PyConfig_InitPythonConfig(PyConfig *config)
+   .. c:function:: void PyConfig_InitPythonConfig(PyConfig *config)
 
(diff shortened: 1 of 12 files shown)


vstinner:
I forgot to update PyConfig_InitPythonConfig() and PyConfig_InitIsolatedConfig() in my PR #16500. There is no more need to return PyStatus since these functions cannot fail anymore. They don't allocate any memory: strings are initialized to NULL, lists to empty lists (which doesn't allocate any memory).

vstinner:
cc @ncoghlan @ambv That can wait for 3.8.0rc2 or 3.8.0 final if it's too late for rc1.

ambv (approved):
Wait for CI and merge.

ncoghlan:
Good catch!

PR #16519, opened 2019-10-01
gh-82516: simplify the overlap function for IpNetwork

s-sanjay:
https://bugs.python.org/issue38335

* Issue: gh-82516

### DIFF
--- a/Lib/ipaddress.py
+++ b/Lib/ipaddress.py
@@ -669,6 +669,19 @@ class _BaseAddress(_IPAddressBase):
 
         return format(int(self), f'{alternate}0{padlen}{grouping}{fmt_base}')
 
+    @property
+    @functools.lru_cache()
+    def is_private(self):
+        """Test if this address is allocated for private networks.
+
+        Returns:
+            A boolean, True if the address is reserved per
+            iana-ipv6-special-registry for ipv6
+            and iana-ipv6-special-registry for ipv4
+
+        """
+        return any(self in net for net in self._constants._private_networks)
+
 
 @functools.total_ordering
 class _BaseNetwork(_IPAddressBase):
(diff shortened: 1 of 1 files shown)


s-sanjay:
no news entry needed as there is no behavior change

Lib/ipaddress.py
```diff
@@ -1511,7 +1512,6 @@ def __init__(self, address, strict=True):
             self.hosts = self.__iter__
 
     @property
-    @functools.lru_cache()
```
remilapeyre, line 1514: Why did you removed this decorator?

remilapeyre (changes requested):
Hi @s-sanjay, you should not move the method around in the file while changing them as it make reviewing the changes harder. Can you please move them back where they were?

Lib/ipaddress.py
s-sanjay: because self.network_address.is_global which in turn calls is_private already has a cache. There are pros and cons for example if someone removes cache there then suddenly this would loose caching behavior and also this having its own cache frees up the other cache. Should I add it back ?

s-sanjay: because self.network_address.is_global which in turn calls is_private already has a cache. There are pros and cons for example if someone removes cache there then suddenly this would loose caching behavior and also this having its own cache frees up the other cache. Should I add it back ?

s-sanjay:
@remilapeyre yes, agreed I avoid moving methods around but in this case the is_private method was deleted from child classes and moved to the base class and the overlap method was moved closer to the subnet_of and supernet_of functions because overlap and them are related terms and the overlap method simply delegates the call to these two methods. I thought this small refactor will help greatly with future readability of the code. what do you think ? should I put it in the old place ?

PR #16520, opened 2019-10-01
bpo-38336: Remove the __set__ method restriction on data descriptors for attribute lookup precedence

geryogam:
The [language documentation](https://docs.python.org/3/reference/datamodel.html#invoking-descriptors) states:

> Data descriptors with \_\_set\_\_() and \_\_get\_\_() defined always override a redefinition in an instance dictionary. In contrast, non-data descriptors can be overridden by instances.

This override is not limited to data descriptors defining `__set__()` (and `__get__()`). It is also true for data descriptors defining `__delete__()` (and `__get__()`) and data descriptors defining both `__set__()` and `__delete__()` (and `__get__()`).

In other words, _any_ data descriptors (objects defining either `__set__()` or `__delete__()` or both) defining `__get__()` override an attribute redefinition in an instance dictionary.

https://bugs.python.org/issue38336

### DIFF
--- a/Doc/reference/datamodel.rst
+++ b/Doc/reference/datamodel.rst
@@ -1714,9 +1714,9 @@ define :meth:`__get__`, then accessing the attribute will return the descriptor
 object itself unless there is a value in the object's instance dictionary.  If
 the descriptor defines :meth:`__set__` and/or :meth:`__delete__`, it is a data
 descriptor; if it defines neither, it is a non-data descriptor.  Normally, data
-descriptors define both :meth:`__get__` and :meth:`__set__`, while non-data
+descriptors define both :meth:`__get__` and :meth:`__set__` (and/or :meth:`__delete__`), while non-data
 descriptors have just the :meth:`__get__` method.  Data descriptors with
-:meth:`__get__` defined always override a redefinition in an
+:meth:`__get__` and :meth:`__set__` (and/or :meth:`__delete__`) defined always override a redefinition in an
 instance dictionary.  In contrast, non-data descriptors can be overridden by
 instances.
 
--- a/Doc/reference/datamodel.rst
+++ b/Doc/reference/datamodel.rst
@@ -1716,7 +1716,7 @@ the descriptor defines :meth:`__set__` and/or :meth:`__delete__`, it is a data
 descriptor; if it defines neither, it is a non-data descriptor.  Normally, data
 descriptors define both :meth:`__get__` and :meth:`__set__`, while non-data
 descriptors have just the :meth:`__get__` method.  Data descriptors with
-:meth:`__set__` and :meth:`__get__` defined always override a redefinition in an
+:meth:`__get__` defined always override a redefinition in an
 instance dictionary.  In contrast, non-data descriptors can be overridden by
 instances.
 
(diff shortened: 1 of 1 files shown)


brandtbucher (approved):
Thanks @maggyero, looks good to me!

Doc/reference/datamodel.rst
```diff
@@ -1716,7 +1716,7 @@ the descriptor defines :meth:`__set__` and/or :meth:`__delete__`, it is a data
 descriptor; if it defines neither, it is a non-data descriptor.  Normally, data
 descriptors define both :meth:`__get__` and :meth:`__set__`, while non-data
 descriptors have just the :meth:`__get__` method.  Data descriptors with
-:meth:`__set__` and :meth:`__get__` defined always override a redefinition in an
+:meth:`__get__` defined always override a redefinition in an
```
rhettinger: The new wording is confusing it that relies heavily on a reader fully understanding that "data descriptor" means that ``__set__`` or ``__delete__`` is defined; otherwise, the new wording can easily be misread as contradicting the sentence that came before it.

The current wording is clear and covers the common case.

If you really feel like the current wording must change because it doesn't cover the highly usual case of ``__delete__`` being defined but not ``__set__``, then you can just modify the original as to:

'''
    :meth:`__set__` (and/or :meth:`__delete__`) and :meth:`__get__` defined always override a redefinition ...
'''


But do consider that a proposed modification while technically correct will make the overall paragraph harder to understand for someone didn't already having an understanding of data descriptors and what their significance.

rhettinger (changes requested):

Doc/reference/datamodel.rst
geryogam: Thanks @rhettinger, I see your point. I have updated the PR with your suggestion (as for me the paragraph was confusing because `__delete__` was not mentioned).

maggyero:

Doc/reference/datamodel.rst
```diff
@@ -1714,9 +1714,9 @@ define :meth:`__get__`, then accessing the attribute will return the descriptor
 object itself unless there is a value in the object's instance dictionary.  If
 the descriptor defines :meth:`__set__` and/or :meth:`__delete__`, it is a data
 descriptor; if it defines neither, it is a non-data descriptor.  Normally, data
-descriptors define both :meth:`__get__` and :meth:`__set__`, while non-data
+descriptors define both :meth:`__get__` and :meth:`__set__` (and/or :meth:`__delete__`), while non-data
```
rhettinger: This line should be left unchanged.  It talks about the *normal* usage and it is correct.  Also, there is no need to belabor the point about ``__delete__()``.

rhettinger (changes requested):

Doc/reference/datamodel.rst
geryogam: Okay, I thought the normal usage was insisting on the `__get__` addition to a data descriptor (whatever its kind, `__set__`, `__delete__` or both), not on the addition plus the kind of data descriptor. Updated.

maggyero:

rhettinger (approved):

maggyero:
Thanks @rhettinger for reviewing this.

PR #16521, opened 2019-10-01
bpo-38337: Change getattr to inspect.getattr_static

3j14:
When calling `inspect.getmembers` on a class that has a property (`@property`), the property will be called by the `getattr` call in `getmembers`.

Example:

```python
import inspect

class Example:
    def __init__(self, var):
        self._var = var
        print('__init__')

    def foo(self, bar):
        print(bar)
        print('foo')

    @property
    def var(self):
        print('Access property')
        return self._var


if __name__ == '__main__':
    ex = Example('Hello')
    print('--- getmembers from instance ---')
    print(inspect.getmembers(ex))
    print('--- getmembers from class    ---')
    print(inspect.getmembers(Example))
```

## Result:

```
__init__
--- getmembers from instance ---
Access property
[('__class__', <attribute '__class__' of 'object' objects>), ..., ('var', 'Hello')]
--- getmembers from class    ---
[('__class__', <attribute '__class__' of 'object' objects>), ..., ('var', <property object at 0x...>)]
```

## Expected (new result):

```
__init__
--- getmembers from instance ---
[('__class__', <attribute '__class__' of 'object' objects>), ..., ('var', <property object at 0x...>)]
--- getmembers from class    ---
[('__class__', <attribute '__class__' of 'object' objects>), ..., ('var', <property object at 0x...>)]
```

https://bugs.python.org/issue38337

### DIFF
--- a/Lib/inspect.py
+++ b/Lib/inspect.py
@@ -347,7 +347,10 @@ def getmembers(object, predicate=None):
         # like calling their __get__ (see bug #1785), so fall back to
         # looking in the __dict__.
         try:
-            value = getattr_static(object, key)
+            if isinstance(getattr_static(object, key), property):
+                value = getattr_static(object, key)
+            else:
+                value = getattr(object, key)
             # handle the duplicate key
             if key in processed:
                 raise AttributeError
--- a/Lib/inspect.py
+++ b/Lib/inspect.py
@@ -347,7 +347,7 @@ def getmembers(object, predicate=None):
         # like calling their __get__ (see bug #1785), so fall back to
         # looking in the __dict__.
         try:
-            value = getattr(object, key)
+            value = getattr_static(object, key)
             # handle the duplicate key
             if key in processed:
                 raise AttributeError
--- a/Lib/test/test_inspect.py
+++ b/Lib/test/test_inspect.py
@@ -1204,7 +1204,11 @@ class TestClassesAndFunctions(unittest.TestCase):
         foobar = Foo(42)
         try:
             members = inspect.getmembers(foobar)
+            property_member = ('bar', inspect.getattr_static(foobar, 'bar'))
+            self.assertIn(property_member, members)
             class_members = inspect.getmembers(Foo)
+            property_class_member = ('bar', inspect.getattr_static(Foo, 'bar'))
+            self.assertIn(property_member, class_members)
         except NotImplementedError:
             self.fail('getmembers() called property!')
 
--- a/Lib/test/test_inspect.py
+++ b/Lib/test/test_inspect.py
@@ -1191,6 +1191,23 @@ class TestClassesAndFunctions(unittest.TestCase):
         self.assertIn(('f', b.f), inspect.getmembers(b))
         self.assertIn(('f', b.f), inspect.getmembers(b, inspect.ismethod))
 
+    def test_getmembers_static(self):
+        class Foo:
+            def __init__(bar):
+                self._bar = bar
+
+            @property
+            def bar(self):
+                # This property should not be called by getmembers
+                raise NotImplementedError
+            
+        foobar = Foo(42)
+        try:
+            members = inspect.getmembers(foobar)
+            class_members = inspect.getmembers(Foo)
+        except NotImplementedError:
+            self.fail('getmembers() called property!')
+
     def test_getmembers_VirtualAttribute(self):
         class M(type):
             def __getattr__(cls, name):
--- a/Lib/test/test_inspect.py
+++ b/Lib/test/test_inspect.py
@@ -1193,7 +1193,7 @@ class TestClassesAndFunctions(unittest.TestCase):
 
     def test_getmembers_static(self):
         class Foo:
-            def __init__(bar):
+            def __init__(self, bar):
                 self._bar = bar
 
             @property
--- a/Lib/test/test_inspect.py
+++ b/Lib/test/test_inspect.py
@@ -1200,7 +1200,7 @@ class TestClassesAndFunctions(unittest.TestCase):
             def bar(self):
                 # This property should not be called by getmembers
                 raise NotImplementedError
-            
+
         foobar = Foo(42)
         try:
             members = inspect.getmembers(foobar)
--- a/Misc/NEWS.d/next/Library/2019-10-01-12-23-37.bpo-38337.QmLSXW.rst
+++ b/Misc/NEWS.d/next/Library/2019-10-01-12-23-37.bpo-38337.QmLSXW.rst
@@ -1,0 +1,1 @@
+Prevent ``inspect.getmembers`` to call properties. Used ``inspect.getattr_static`` instead.
--- a/Misc/NEWS.d/next/Library/2019-10-01-12-23-37.bpo-38337.QmLSXW.rst
+++ b/Misc/NEWS.d/next/Library/2019-10-01-12-23-37.bpo-38337.QmLSXW.rst
@@ -1,1 +1,1 @@
-Prevent ``inspect.getmembers`` to call properties. Used ``inspect.getattr_static`` instead.
+Prevent :func:`inspect.getmembers` from calling descriptors.


the-knights-who-say-ni:

## CLA Missing

@jnsdrtlf

We also demand... [A SHRUBBERY!](https://www.youtube.com/watch?v=zIV4poUZAQo)


s-sanjay:
should we add a test case ?

Misc/NEWS.d/next/Library/2019-10-01-12-23-37.bpo-38337.QmLSXW.rst
```diff
@@ -0,0 +1 @@
+Prevent ``inspect.getmembers`` to call properties. Used ``inspect.getattr_static`` instead.
```
brandtbucher:
```suggestion
Prevent :func:`inspect.getmembers` from calling descriptors.
```

brandtbucher (changes requested):
Thanks @jnsdrtlf, and welcome to CPython! 😎

I have a minor change to suggest for the NEWS entry, but otherwise this looks like a good improvement.

I also agree that adding regression test is probably a good idea.

brandtbucher:
Actually, from looking at the test failures... I'm not sure if this is the desired behavior (and the docs are ambiguous). Maybe @1st1 can shed some light on the "correct" behavior of this function?

1st1:
Since `inspect.getmembers` is an introspection function it shouldn't run any actual code (i.e. executing getters).  It should compute its result statically.  Using `getattr_static` is the right idea here, for sure.

brandtbucher:
Okay then. So this probably doesn't need a new test, but the ~4 failing ones should be updated!

1st1:
>  So this probably doesn't need a new test, but the ~4 failing ones should be updated!

I think it needs a new test. Basically an object with a property and a test ensuring that the property was discovered by getmembers, but the getter code wasn't run.

jnsdrtlf:
> I think it needs a new test. Basically an object with a property and a test ensuring that the property was discovered by getmembers, but the getter code wasn't run.

Sounds good! I will get back to work and write a test later today.

jnsdrtlf:
The test `test_getmembers_VirtualAttribute` expects this method
```python
@types.DynamicClassAttribute
    def eggs(self):
        return 'spam'
```
to be called. It asserts whether `('eggs', 'spam')` is in `inspect.getmembers`. But this is exactly the behavior that I tried to avoid with my PR.

There is also some wierd behaviour when overwriting `__getattr__` in a metaclass. This probably needs more insight on how `inspect.getmembers` should behave.

**Failing tests:**
 - [ ] `test_get_slot_members`
 - [ ] `test_getmembers_method`
 - [ ] `test_getmembers_VirtualAttribute`

Lib/test/test_inspect.py
```diff
@@ -1191,6 +1191,27 @@ def f(self):
         self.assertIn(('f', b.f), inspect.getmembers(b))
         self.assertIn(('f', b.f), inspect.getmembers(b, inspect.ismethod))
 
+    def test_getmembers_static(self):
+        class Foo:
+            def __init__(bar):
```
hongweipeng: Use `def __init__(self, bar):` please.

3j14: Oups 🤦‍♂  Fixed in commit b415224

hongweipeng:
```
Fixing docs whitespace ... 0 files
Please fix the 1 file(s) with whitespace issues
(on UNIX you can run `make patchcheck` to make the fixes)
```
There seems to be another error, please delete the whitespace of line 1203 in `test_inspect.py`.

https://github.com/jnsdrtlf/cpython/blob/b4152248739073b50c548393161ebc43b4d89b9c/Lib/test/test_inspect.py#L1200-L1204

jnsdrtlf:
Another inconsistent behaviour: `inspect.isdatadescriptor` should return `True` for properties:

> Return true if the object is a data descriptor.
>
> Data descriptors have both a __get__ and a __set__ method. Examples are properties (defined in Python), getsets, and members

([Python Docs](https://docs.python.org/3/library/inspect.html?highlight=isdatadescriptor#inspect.isdatadescriptor))

However, the following code does not return any properties:

```python
class Foo:
    def __init__(self, data):
        self._bar = data
    
    @property
    def bar(self):
        print('--BAR--')
        return self._bar
```
**Result**
```
--BAR--
[]
```
**Expected (new result)**
```python
[('__class__', <attribute '__class__' of 'object' objects>), ('__dict__', <attribute '__dict__' of 'Foo' objects>), ('__weakref__', <attribute '__weakref__' of 'Foo' objects>), ('bar', <property object at 0x...>)]
```


Is this correct so far? I am still confused about how `getmembers` should behave. See `test_getmembers_VirtualAttribute`, this test expects a different behaviour.

jnsdrtlf:
Additionally, 
```python
class Foo:
    pass

inspect.getattr_static(Foo('test'), '__class__')
```
returns `<attribute '__class__' of 'object' objects>` whereas `getattr` would return `<class '__main__.Foo'>`. What should `getmembers` return? If it returns `<attribute '__class__' of 'object' objects>`, the `inspect.isclass` predicate would not work:
```python
class Foo:
    pass

# Will return [] if getattr_static is used
inspect.getmembers(Foo('test'), inspect.isclass)
```

Lib/inspect.py
```diff
@@ -347,7 +347,10 @@ def getmembers(object, predicate=None):
         # like calling their __get__ (see bug #1785), so fall back to
         # looking in the __dict__.
         try:
-            value = getattr_static(object, key)
+            if isinstance(getattr_static(object, key), property):
```
1st1, line 350: Why can't we always use `getattr_static`?  If there's a reason for not using it always, please describe that in a comment.

3j14: This is related to my latest comment. Sorry, I didn't make this clear:

`getattr_static(...,'__class__')` returns `<attribute '__class__' of 'object' objects>` whereas `getattr` would return `<class '__main__.Foo'>`. What should getmembers return? Using `getattr_static` for every attribute doesn't seem to be the solution. It works fine with methods, functions and properties, but some edge cases seem to require `getattr` (such as `__class__`). It is not clear what the expected behaviour is.

1st1: Got it, well, describe this in a comment :)

jnsdrtlf:
I've changed `inspect.getmembers` to only use `inspect.getattr_static` for properties, as other members (e.g. `__class__`) are expected to be returned using `getattr(obj,'__class__')`:

`getattr_static(obj,'__class__')` returns `<attribute '__class__' of 'object' objects>` whereas `getattr` would return `<class '__main__.Foo'>`

This feels more like a temporary fix - could be reverted to the first fix I've committed - although I am just not sure which is better.

At this point, **I am not sure what the actual expected behaviour should be. Someone should review this and the test case that is currently failing**, as it expects a different behaviour (return the actual value of the property).

1st1 (changes requested):
Jonas, I thought about this PR for some time and I think we should reject it.  Modifying `getmembers` in any way would break backwards compatibility in a very non-obvious way for many users.  The ship to fix it has sailed.

We can add a new function `getmembers_static` that would use `getattr_static` instead of `getattr`.  It must be a new distinct function, though.

jnsdrtlf:
@1st1 I share your thoughts on that. It does not seem to be obvious, what `getmembers` should return - and changing it in any way will break existing code. This has already caused me a lot of headaches.

As for introducing a function `getmembers_static`: I still have some questions that should be discussed before doing so:

1. What **should** `getmembers_static` return? Is `getattr_static` appropriate for every member? This would break some predicates (e.g. `isclass`, see comment above). There does not seem to be a "simple" solution.

2. A technical question: The new method should pass most of the same tests as `getmembers`. Would one just copy and paste the old ones? It would also share most of the code - except for maybe 1 line - with `getmembers`.

3. Is there a reason for `getmembers` to return a list of tuples? I would have suggested to use a `dict` instead, which feels more intuitive on first though.

1st1:
> Is there a better place to discuss such questions?

I think it would be better if you could post this to the python-ideas mailing list. I've very limited time and would hate to get this thing staled because of me.

1st1:
Or post it here: https://discuss.python.org/c/ideas

csabella:
I added a link to the discussion on the bug tracker.

jaymegordo:
This would be very helpful, running into the same issue with inspect.getmembers evaluating @property decorated methods.

getmembers_static would solve it.

PR #16527, opened 2019-10-01
bpo-38343: Fixes version handling for nuget packages

zooba:
https://bugs.python.org/issue38343


zooba:
Ping @ambv - this will need cherry-picking for your 3.8 branch (assuming you're going to maintain the 3.8.0 branch separately until release)

PR #16529, opened 2019-10-02
bpo-38334: Fix seeking bug for encrypted zipfiles

danifus:
This resolves an issue where seeking on an encrypted `zipfile` would cause the crc keys to go out of sync resulting in bad data to be read after the seek.

https://bugs.python.org/issue38334


danifus:
Fixed by https://github.com/python/cpython/pull/16937

PR #16530, opened 2019-10-02
bpo-36670, regrtest: Fix TestWorkerProcess.stop()

vstinner:
* Add a RLock to TestWorkerProcess to make stop() reliable when
  regrtest is interrupted by CTRL+c: protect _popen and _killed
  attributes.
* Fix _kill(): don't set _killed to True if _popen is None.
* Fix TestWorkerProcess.__repr__(): start_time is only valid
  if _popen is not None.


vstinner:
I'm not sure if my lock is correct, so I wrote a simpler PR with no lock: PR #16537.

PR #16533, opened 2019-10-02
bpo-38344 else needs to be on the same line as the if

jamesabel:
In activate.bat the else needs to be on the same line as the if

https://bugs.python.org/issue38344


the-knights-who-say-ni:

## CLA Missing

@jamesabel

Lib/venv/scripts/nt/activate.bat
```diff
@@ -21,8 +21,7 @@ set PROMPT=__VENV_PROMPT__%PROMPT%
 if defined PYTHONHOME set _OLD_VIRTUAL_PYTHONHOME=%PYTHONHOME%
 set PYTHONHOME=
 
-if defined _OLD_VIRTUAL_PATH set PATH=%_OLD_VIRTUAL_PATH%
-else set _OLD_VIRTUAL_PATH=%PATH%
+if defined _OLD_VIRTUAL_PATH set PATH=%_OLD_VIRTUAL_PATH% else set _OLD_VIRTUAL_PATH=%PATH%
```
eryksun: This will never execute the `else` clause, and if `_OLD_VIRTUAL_PATH` is defined it will set `PATH` literally to the evaluation of the string `"%_OLD_VIRTUAL_PATH% else set _OLD_VIRTUAL_PATH=%PATH%"`.

Why again were parentheses stripped out of this script? Whatever the reason, there must be a better way. It's dysfunctional to not be able to write this statement as follows:

    if defined _OLD_VIRTUAL_PATH (
        set PATH=%_OLD_VIRTUAL_PATH%
    ) else (
        set _OLD_VIRTUAL_PATH=%PATH%
    )

jamesabel: I didn't mention it previously, but in the current form the script errors out:

C:\Program Files\Python38\Lib\venv\scripts\nt>activate.bat
'else' is not recognized as an internal or external command,
operable program or batch file.

But good point on the parenthesis to make it function properly again.

zooba: The parentheses were stripped out because stray parentheses and quotes were breaking the script for people.

Having `if defined ...` followed by `if not defined ...` on separate (single) lines is probably the best approach.

Lib/venv/scripts/nt/activate.bat
```diff
@@ -21,8 +21,11 @@ set PROMPT=__VENV_PROMPT__%PROMPT%
 if defined PYTHONHOME set _OLD_VIRTUAL_PYTHONHOME=%PYTHONHOME%
 set PYTHONHOME=
 
-if defined _OLD_VIRTUAL_PATH set PATH=%_OLD_VIRTUAL_PATH%
-else set _OLD_VIRTUAL_PATH=%PATH%
```
zooba, line 25: Please restore the old syntax and replace the `else` with `if not defined ...`.

zooba (changes requested):
Also see the issue for ideas about making sure this is tested properly.

jamesabel:

Lib/venv/scripts/nt/activate.bat
```diff
@@ -21,8 +21,8 @@ set PROMPT=__VENV_PROMPT__%PROMPT%
...
+if defined _OLD_VIRTUAL_PATH set "PATH=%_OLD_VIRTUAL_PATH%"
+if not defined _OLD_VIRTUAL_PATH set "_OLD_VIRTUAL_PATH=%PATH%"
```
zooba: Need to lose the quotes as well - they can cause problems if there are mismatched quotes in the variables (and they're not necessary when the statement extends to the end of the line, which is why the original change was necessary).

Misc/NEWS.d/next/Windows/2019-10-05-05-50-58.bpo-38344.scr2LO.rst
```diff
@@ -0,0 +1 @@
+Fix usage of _OLD_VIRTUAL_PATH in activate.bat.  Do not use parenthesis in the if statement.  Instead of if/else, use if defined/if not defined on separate lines.
```
zooba: The NEWS item doesn't need to be this detailed - it'll link back to the bug, which links to the PR, which links to the changes for anyone interested. Simply saying "Fix error message in activate.bat" would be fine.

jamesabel:

zooba:
Thanks! I've merged it as-is, since we need the fix, but we'll leave the bug open for now to consider enhancing the test suite as well.

PR #16535, opened 2019-10-02
Fixes and improvements to 3.8 whatsnew

aeros:
This PR is mostly focused on repairing broken Sphinx roles that weren't correctly generating links, but also adds a few new ones and shortens some of the longer ones using `~`. It also includes a few minor grammar corrections and improvements. 

For comparison, see https://docs.python.org/3.8/whatsnew/3.8.html and build the html locally from this PR. I've verified each of the Sphinx roles that I added locally to ensure the links were properly generated.

Also, is there an easy way to add myself to Misc/ACKS just for 3.8 without opening a separate PR? I added in an entry for myself in 3.9 several months ago, but never did it for 3.8. I've had several merged PRs that were backported to 3.8. I was thinking that it might be possible to add it to the backport PR.

### DIFF
--- a/Doc/whatsnew/3.8.rst
+++ b/Doc/whatsnew/3.8.rst
@@ -1338,8 +1338,8 @@ Build and C API Changes
   :class:`~fractions.Fraction`).  :c:func:`PyNumber_Check` will now return
   ``1`` for objects implementing :meth:`~object.__index__`.
   :c:func:`PyNumber_Long`, :c:func:`PyNumber_Float` and
-  :c:func:`PyFloat_AsDouble` also now use the meth:`object.__index__` method if
-  available.
+  :c:func:`PyFloat_AsDouble` also now use the :meth:`object.__index__` method
+  if available.
   (Contributed by Serhiy Storchaka in :issue:`36048` and :issue:`20092`.)
 
 * Heap-allocated type objects will now increase their reference count
--- a/Doc/whatsnew/3.8.rst
+++ b/Doc/whatsnew/3.8.rst
@@ -748,7 +748,7 @@ The changes above have been backported to 3.7 maintenance releases.
 json.tool
 ---------
 
-Add option ``--json-lines`` to parse every input line as separate JSON objects.
+Add option ``--json-lines`` to parse every input line as a separate JSON object.
 (Contributed by Weipeng Hong in :issue:`31553`.)
 
 
@@ -1574,7 +1574,7 @@ Changes in the Python API
   (Contributed by Eric Snow in :issue:`34651`, modified by Christian Heimes
   in :issue:`37951`.)
 
-* The :meth:`imaplib.IMAP4.logout` method no longer ignores silently arbitrary
+* The :meth:`imaplib.IMAP4.logout` method no longer silently ignores arbitrary
   exceptions.
 
 * The function :func:`platform.popen` has been removed, it was deprecated since
--- a/Doc/whatsnew/3.8.rst
+++ b/Doc/whatsnew/3.8.rst
@@ -405,8 +405,9 @@ Other Language Changes
 
   (Contributed by Jonathan Eunice and Serhiy Storchaka in :issue:`30688`.)
 
-* :class:`dict` and :term:`dictionary view`s are now iterable in reversed insertion
-  order using :func:`reversed`. (Contributed by Rémi Lapeyre in :issue:`33462`.)
+* :class:`dict` and :term:`dictionary view`s are now iterable in reversed
+  insertion order using :func:`reversed`.
+  (Contributed by Rémi Lapeyre in :issue:`33462`.)
 
 * The syntax allowed for keyword names in function calls was further
   restricted. In particular, ``f((keyword)=arg)`` is no longer allowed. It was
--- a/Doc/whatsnew/3.8.rst
+++ b/Doc/whatsnew/3.8.rst
@@ -596,7 +596,8 @@ The :func:`compile` built-in has been improved to accept the
 ``ast.PyCF_ALLOW_TOP_LEVEL_AWAIT`` flag. With this new flag passed,
 :func:`compile` will allow top-level :keyword:`await`, :keyword:`async for` and
 :keyword:`async with` constructs that are usually considered invalid syntax.
-Asynchronous code object marked with the :data:`inspect.CO_COROUTINE` flag may then be returned.
+Asynchronous code object marked with the :data:`inspect.CO_COROUTINE` flag may
+then be returned.
 
 (Contributed by Matthias Bussonnier in :issue:`34616`)
 
--- a/Doc/whatsnew/3.8.rst
+++ b/Doc/whatsnew/3.8.rst
@@ -596,7 +596,7 @@ The :func:`compile` built-in has been improved to accept the
 ``ast.PyCF_ALLOW_TOP_LEVEL_AWAIT`` flag. With this new flag passed,
 :func:`compile` will allow top-level :keyword:`await`, :keyword:`async for` and
 :keyword:`async with` constructs that are usually considered invalid syntax.
-Asynchronous code object marked with the :data:`inspect.CO_COROUTINE` flag may
+Asynchronous code objects marked with the :data:`inspect.CO_COROUTINE` flag may
 then be returned.
 
 (Contributed by Matthias Bussonnier in :issue:`34616`)
--- a/Doc/whatsnew/3.8.rst
+++ b/Doc/whatsnew/3.8.rst
@@ -405,8 +405,8 @@ Other Language Changes
 
   (Contributed by Jonathan Eunice and Serhiy Storchaka in :issue:`30688`.)
 
-* Dict and dictviews are now iterable in reversed insertion order using
-  :func:`reversed`. (Contributed by Rémi Lapeyre in :issue:`33462`.)
+* :class:`dict` and dictionary views are now iterable in reversed insertion
+  order using :func:`reversed`. (Contributed by Rémi Lapeyre in :issue:`33462`.)
 
 * The syntax allowed for keyword names in function calls was further
   restricted. In particular, ``f((keyword)=arg)`` is no longer allowed. It was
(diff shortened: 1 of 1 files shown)


Doc/whatsnew/3.8.rst
```diff
@@ -593,20 +594,22 @@ builtins
 
 The :func:`compile` built-in has been improved to accept the
 ``ast.PyCF_ALLOW_TOP_LEVEL_AWAIT`` flag. With this new flag passed,
-:func:`compile` will allow top-level ``await``, ``async for`` and ``async with``
-constructs that are usually considered invalid syntax. Asynchronous code object
-marked with the ``CO_COROUTINE`` flag may then be returned.
+:func:`compile` will allow top-level :keyword:`await`, :keyword:`async for` and
+:keyword:`async with` constructs that are usually considered invalid syntax.
+Asynchronous code object marked with the :data:`inspect.CO_COROUTINE` flag may
```
JelleZijlstra: This doesn't read well, probably insert "An" as the first word of the sentence.

Doc/whatsnew/3.8.rst
```diff
@@ -745,7 +748,7 @@ The changes above have been backported to 3.7 maintenance releases.
 json.tool
 ---------
 
-Add option ``--json-lines`` to parse every input line as separate JSON object.
+Add option ``--json-lines`` to parse every input line as separate JSON objects.
```
JelleZijlstra: "as a separate JSON object" would be better

Doc/whatsnew/3.8.rst
```diff
@@ -1568,7 +1574,7 @@ Changes in the Python API
   (Contributed by Eric Snow in :issue:`34651`, modified by Christian Heimes
   in :issue:`37951`.)
 
-* The :meth:`imap.IMAP4.logout` method no longer ignores silently arbitrary
+* The :meth:`imaplib.IMAP4.logout` method no longer ignores silently arbitrary
```
JelleZijlstra: Move "silently" either before "ignores" or to the end of the sentence

Doc/whatsnew/3.8.rst
aeros: I agree that the current version doesn't read well, but I would prefer changing "object" to "objects":
```
Asynchronous code objects marked with the :data:`inspect.CO_COROUTINE` flag may
```

Both options would be grammatically correct, it's fairly subjective. The main reason I'm leaning towards the plural here is because there can be multiple asynchronous code objects marked within the code being compiled, not just one.

JelleZijlstra: Yes, that's better

aeros167:
Thanks for feedback @JelleZijlstra.

Doc/whatsnew/3.8.rst
```diff
@@ -405,17 +405,17 @@ Other Language Changes
...
 
 * The syntax allowed for keyword names in function calls was further
   restricted. In particular, ``f((keyword)=arg)`` is no longer allowed. It was
   never intended to permit more than a bare name on the left-hand side of a
   keyword argument assignment term. See :issue:`34641`.
 
 * Generalized iterable unpacking in :keyword:`yield` and
-  :keyword:`return` statements no longer requires enclosing parentheses.
-  This brings the *yield* and *return* syntax into better agreement with
+  :keyword:`return` statements no longer requires enclosing parenthesis.
```
merwok: This looks incorrect: things like `thesis` have plural form `theses`

Doc/whatsnew/3.8.rst
```diff
@@ -405,17 +405,17 @@ Other Language Changes
 
   (Contributed by Jonathan Eunice and Serhiy Storchaka in :issue:`30688`.)
 
-* Dict and dictviews are now iterable in reversed insertion order using
-  :func:`reversed`. (Contributed by Rémi Lapeyre in :issue:`33462`.)
+* :class:`dict` and dictionary views are now iterable in reversed insertion
```
merwok: I think there is a glossary item for dict viewe (can create a reference with the `term` role)

Doc/whatsnew/3.8.rst
aeros: Oh you're correct. I was under the impression that it was `parenthesis` for a single pair of `()` and parentheses referred to multiple pairs of `()`. Apparently, parenthesis refers a single `(` and parentheses refers to a pair `()`. It seems like I was using the wrong form of the word all this time and hadn't noticed, good to know. (:

Doc/whatsnew/3.8.rst
aeros: Yep, looks like there is one https://docs.python.org/3/glossary.html#term-dictionary-view.
```suggestion
* :class:`dict` and :term:`dictionary view`s are now iterable in reversed insertion
```

Doc/whatsnew/3.8.rst
aeros: Reverting this change.
```suggestion
  :keyword:`return` statements no longer requires enclosing parentheses.
```

Doc/whatsnew/3.8.rst
aeros: Looks like that caused it to go over the 80 char limit at 83 chars. I'll adjust the wrapping.

aeros167:
Thanks for the feedback @merwok.

aeros167:
Correcting that sphinx role, it should be
```
:term:`dictionary views <dictionary-view>`
```
instead of
```
:term:`dictionary view`s
```

Doc/whatsnew/3.8.rst
merwok: ReST trick: to avoid having bits dangling out of markup like the plural `s`, there is a way to control the display text separate from the reference:

```rst
and :term:`dict views <dictionary view>`
```

merwok: Ah I see you found that! :slightly_smiling_face:

aeros: Yep, I vaguely remember reading about it in http://docutils.sourceforge.net/rst.html and used it before for explicit links. So, I tried it out on my local copy of `3.8.rst`, did `make html`, and it worked! (:

aeros167:
@rhettinger Will you have time to look this over in the next week or so? Since @ambv has the release date for 3.8 planned on October 14th, we should probably try to wrap this up before October 12th to avoid last minute additions.

rhettinger:
I don't think most of these should be done and they will get in the way of the more substantive edits currently in progress.

aeros:
@rhettinger 
> I don't think most of these should be done and they will get in the way of the more substantive edits currently in progress.

This PR was primarily made because of @ambv's [announcement in python-dev for 3.8.0rc1](https://mail.google.com/mail/u/0/#search/ambv/FMfcgxwDrbrQtzFkCjMbwPRnSNZrSzlh), specifically the request for helping with the "What's New" document:

```
Can you help with the “What’s New” document?
```

As someone who has become highly familiar with using Sphinx roles from other contributions I've made to the documentation and the devguide, I found that I was able to help fix many of the currently broken inline links. Those links can be quite useful for readers to easily navigate to a page for finding more information on a given topic. It's not always straightforward for readers to find the correct location if they're not familiar with the structure of the documentation and the search feature is not always reliable. 

I'm aware that they are not quite as important as the entries themselves, but they certainly have some "substantive" value. I also spent a substantial amount of time reviewing these changes to make sure the ones that were fixed or added were rendering properly and linking to the current location by building the documentation locally, in addition to finding the broken ones in the first place. 

Also, I feel like I've been getting rather mixed and inconsistent messages regarding the usage of Sphinx roles. From the [thread that I created in python-dev regarding Sphinx roles](https://mail.python.org/archives/list/python-dev@python.org/thread/HTAFGTQIZJQUCU6QCVF3KFD3VFGFOBWV/#HTAFGTQIZJQUCU6QCVF3KFD3VFGFOBWV) in addition to other conversations with core developers, my impression has been that the most valuable locations to have them in is "What's New" and the docs. Since then, I haven't emphasized on adding them for individual Misc/NEWS entries, as I was in the past. Recently, I've only been concerned with fixing them in the actual docs and "What's New". Personally, I think that if we're going to use them, we should do so correctly. 

The impression I'm getting here though is that these changes getting in the way of more substantial ones. If this is correct, I am willing to accept that and can focus my efforts on other areas, such as asyncio. 

But, I would highly appreciate it if we could have a clear message on this so that I'm not just spending a portion my available time tidying up "What's New" entries, only for for it to be essentially said that my efforts are getting in the way.

ncoghlan:
@rhettinger There are a number of PRs that have been merged for What's New in the past few days, and you've removed the draft status. Does that mean @aeros can go ahead with resolving the merge conflicts with this PR, and we can discuss any remaining concerns with the specific proposed changes?

rhettinger:
I don't want to go forward with this for a while.  Many of the markup changes are debatable and are mostly non-substantive.   Also, there is excess mark-up that isn't necessary or desirable in Whatsnew (we don't have to link every term to the glossary or every mention of dict to the dict class -- those may be important elsewhere but less so in this document).  FWIW, there are other edits in process that should be finished before creating any more potential merge conflicts.  Also, it takes a lot of time to individually discuss every micro-edit, especially when stylistic changes are at the editors discretion.  The kind of contribution that would be most useful at this point would be better examples for the parts I haven't touched yet (for example, I am meeting soon with Davin to come up with a meaningful and correct example for multiprocessing shared memory).

rhettinger:
FWIW, some of these were helpful.  I moved the useful ones into my latest PR so as to avoid continuing merge conflicts and so that I could see the markup effects one by one.  See https://github.com/python/cpython/pull/16854

Marking this a closed.

aeros:
@rhettinger 

>  Also, there is excess mark-up that isn't necessary or desirable in Whatsnew (we don't have to link every term to the glossary or every mention of dict to the dict class -- those may be important elsewhere but less so in this document).

I primarily focused on doing so for terms that had links to the glossary in other sections and when it seemed relevant. But I can agree with the general sentiment that excessive inline links can be distracting. Was this the primary issue with the PR? If not, could you elaborate on some of the other general issues without having to point to specifics? This would help me with future PRs to What's New.

> FWIW, some of these were helpful. I moved the useful ones into my latest PR so as to avoid continuing merge conflicts and so that I could see the markup effects one by one. See #16854

That's good to hear at least, I'm glad that some of the changes were useful. Would you mind either adding me to the "Co-Authored-By" in the commit or mentioning me in the PR itself?

ncoghlan:
What's New sections are often linked to directly, so a useful policy to follow is "Link a term the first time it is mentioned in a section, but don't link the second and subsequent references".

(That policy is actually generally useful throughout all the documentation, since it's pretty rare for anyone to read entire pages top to bottom - they're more commonly following a link to a specific section or API header)

PR #16536, opened 2019-10-02
bpo-38347: find Python scripts whose name contain a '-'

rpluem:
Allow Python scripts to contain a '-' in their filename when working
recursively.

https://bugs.python.org/issue38347


the-knights-who-say-ni:

## CLA Missing

@rpluem

vstinner (commented):
Would you mind to add a test to Lib/test/test_tools/test_pathfix.py?

rpluem:
> Would you mind to add a test to Lib/test/test_tools/test_pathfix.py?

This requires a bigger amount of work since the ability of pathfix.py to recursively search a directory is not tested at all currently. But the issue only happens in this case. If you specify an explicit filename on command line with '-' in the name things work fine.

vstinner:
> This requires a bigger amount of work since the ability of pathfix.py to recursively search a directory is not tested at all currently. But the issue only happens in this case.

It sounds like a nice enhancement to write a test for this code path.

By the way, I don't see the point of using a regex to match the filename. Why not only checking for the ".py" suffix?

rpluem:
> > This requires a bigger amount of work since the ability of pathfix.py to recursively search a directory is not tested at all currently. But the issue only happens in this case.
> 
> It sounds like a nice enhancement to write a test for this code path.
> 
> By the way, I don't see the point of using a regex to match the filename. Why not only checking for the ".py" suffix?

I tried to keep the changes to a minimum. Of course it can be changed to allow all filenames that just have the .py sujffix. I can adjust this. Whatever is your preference there :-)

vstinner:
The regex is quite old:

```
commit 9af22a037fc961b51651324bdd17a43567a688fd
Author: Guido van Rossum <guido@python.org>
Date:   Fri Aug 19 15:02:57 1994 +0000

    newslist.py: Added search for .newslistrc.py;
    pindent.py: use /usr/local/bin/python;
    pathfix.py: new script to fix #! lines in a group of scripts.

(...)

+ispythonprog = regex.compile('^[a-zA-Z0-9_]+\.py$')
+def ispython(name):
+       return ispythonprog.match(name) >= 0
+

(...)
```

I don't see any rationale for rejecting filenames which don't match '^[a-zA-Z0-9_]+\.py$', I suggest to only check for the suffix.

vstinner:
By the way, please add a NEWS entry using the Python tool "blurb" ("python3 -m pip install --user blurb" then run "blurb").

rpluem:
> By the way, please add a NEWS entry using the Python tool "blurb" ("python3 -m pip install --user blurb" then run "blurb").

I already did: https://github.com/python/cpython/pull/16536/commits/7aa6c0997f40f110bccc180c01b3960737d53aef
Looks like the check status is outdated.

rpluem:
> > This requires a bigger amount of work since the ability of pathfix.py to recursively search a directory is not tested at all currently. But the issue only happens in this case.
> 
> It sounds like a nice enhancement to write a test for this code path.
> 

Test case created (05f15f8). I hope the code is not too weird.

rpluem:
> I don't see any rationale for rejecting filenames which don't match '^[a-zA-Z0-9_]+.py$', I suggest to only check for the suffix.

Done in 0c1e65c

Lib/test/test_tools/test_pathfix.py
```diff
@@ -14,16 +14,24 @@ class TestPathfixFunctional(unittest.TestCase):
     script = os.path.join(scriptsdir, 'pathfix.py')
 
     def setUp(self):
-        self.temp_file = support.TESTFN
-        self.addCleanup(support.unlink, support.TESTFN)
+        self.temp_directory = support.TESTFN + '.d'
+        self.addCleanup(support.rmtree, self.temp_directory)
+        os.mkdir(self.temp_directory)
```
vstinner: Please move this code into test_recursive(). Other tests don't need that.

Lib/test/test_tools/test_pathfix.py
```diff
@@ -14,16 +14,24 @@ class TestPathfixFunctional(unittest.TestCase):
...
+        self.temp_file = self.temp_directory + os.sep + \
+                         os.path.basename(support.TESTFN) + '-t.py'
+        self.addCleanup(support.unlink, self.temp_file)
+
+    def pathfix(self, shebang, pathfix_flags, exitcode=0, stdout='', stderr='',
+                target_file=''):
```
vstinner: I suggest to rename target_file to filename.

Lib/test/test_tools/test_pathfix.py
```diff
@@ -44,55 +52,66 @@ def pathfix(self, shebang, pathfix_flags, exitcode=0, stdout='', stderr=''):
...
+        expected_stderr = f'recursedown(\'{temp_directory_basename}\')\n'
+        for method_name in dir(self):
+            method = getattr(self, method_name)
+            if method_name.startswith('test_') and \
+               method_name != 'test_recursive' and \
+               method_name != 'test_pathfix_adding_errors' and \
+               callable(method):
+                method(target_file=self.temp_directory, stderr=expected_stderr)
+
+    def test_pathfix(self, **kwargs):
```
vstinner: I don't understand the purpose of the new **kwargs argument.

Misc/NEWS.d/next/Tests/2019-10-02-12-27-58.bpo-38347.zdkCBX.rst
```diff
@@ -0,0 +1 @@
+Add a test case for the recursive search including filenames with a '-'.
```
vstinner: Please remove this NEWS entry. The other one is enough.

Misc/NEWS.d/next/Tools/Demos/2019-10-02-09-48-42.bpo-38347.2Tq5D1.rst
```diff
@@ -0,0 +1 @@
+Allow Python scripts to contain a '-' in their filename when working recursively.
```
vstinner: You should mention pathfix somewhere. And this NEWS entry is outdated: it's no longer about "-", but any character.

vstinner (changes requested):

Lib/test/test_tools/test_pathfix.py
rpluem: Fixed in 07f7158

Misc/NEWS.d/next/Tests/2019-10-02-12-27-58.bpo-38347.zdkCBX.rst
rpluem: Done in 14f5aa0d351fddae1e798b5838f498c648bbcbda

Misc/NEWS.d/next/Tools/Demos/2019-10-02-09-48-42.bpo-38347.2Tq5D1.rst
rpluem: Done in a68ad4207100b365d00ffbb745592408d1d55a98 and 05c39a5dd793c45fb6c2f2cff62b947a9576b352

vstinner (changes requested):
I see that you updated your PR, but setUp() still creates a temporary directory (I asked to move that into the test_recursive() method), there is still "**kwargs", and now test_recursive() inspects self to call other functions... I don't understand what you are trying to do.

My expectation is that test_recursive() creates a temporary directory which contains one ".py" file which has a shebang, you can pathfix on the directory, and then you check that the .py file shebang is updated.

Please don't try to magically make other test methods work on a temporary directory. I prefer simple tests.

Misc/NEWS.d/next/Tests/2019-10-02-12-27-58.bpo-38347.zdkCBX.rst
rpluem: Just as a comment: This causes one of the automatic checks to fail again, but I guess this is not a real blocker.

rpluem:
> 
> 
> I see that you updated your PR, but setUp() still creates a temporary directory (I asked to move that into the test_recursive() method), there is still "**kwargs", and now test_recursive() inspects self to call other functions... I don't understand what you are trying to do.
> 
> My expectation is that test_recursive() creates a temporary directory which contains one ".py" file which has a shebang, you can pathfix on the directory, and then you check that the .py file shebang is updated.
> 
> Please don't try to magically make other test methods work on a temporary directory. I prefer simple tests.

The idea was to repeat the same tests we do on simple files for the recursive case as well. Although I don't see it right now it could happen that the other tests as well behave differently when applied to the single path scenario and the recurse scenario. This would catch cases where things work fine for single file scenarios but not for recursive scenarios. 
But I agree and understand that the test approach you propose would just test the specific issue I addressed (script filename not matching regular expression).
Just let me know if this explains my rationale and which way I should move forward.

vstinner:
> The idea was to repeat the same tests we do on simple files for the recursive case as well.

Please don't do that. Keep the tests as simple. Adding a single isolated test enough to check for your bugfix.

rpluem:
After reverting the previous changes to `test_pathfix.py` I hope that 92384a5 now addresses your concerns.

Lib/test/test_tools/test_pathfix.py
```diff
@@ -17,13 +17,17 @@ def setUp(self):
         self.temp_file = support.TESTFN
         self.addCleanup(support.unlink, support.TESTFN)
 
-    def pathfix(self, shebang, pathfix_flags, exitcode=0, stdout='', stderr=''):
+    def pathfix(self, shebang, pathfix_flags, exitcode=0, stdout='', stderr='',
+                filename=''):
+        if filename == '':
+            filename = self.temp_file
+
         with open(self.temp_file, 'w', encoding='utf8') as f:
```
vstinner: You should use filename here and below.

Lib/test/test_tools/test_pathfix.py
```diff
@@ -44,6 +48,22 @@ def pathfix(self, shebang, pathfix_flags, exitcode=0, stdout='', stderr=''):
...
         return new_shebang
 
+    def test_recursive(self):
+        self.temp_directory = support.TESTFN + '.d'
+        self.addCleanup(support.rmtree, self.temp_directory)
+        os.mkdir(self.temp_directory)
+        self.temp_file = self.temp_directory + os.sep + \
+                         os.path.basename(support.TESTFN) + '-t.py'
+        temp_directory_basename = os.path.basename(self.temp_directory)
+        expected_stderr = f'recursedown(\'{temp_directory_basename}\')\n'
```
vstinner: I suggest:

```
tmpdir = support.TESTFN
self.addCleanup(support.rmtree, tmpdir)
os.mkdir(tmpdir)
expected_stderr = f"recursedown('{os.path.basename(tmpdir)}')\n"
... self.pathfix(..., directory=tmpdir, ...) ...
```

By the way, the self.temp_file attribute is useless. You can simply remove the setUp() method. If you replace pathfix() `filename` parameter with a new `directory` parameter, you can write something like:

```
if directory:
  pathfix_arg = directory
  filename = os.path.join(directory, 'script.py')
else:
  filename = support.TESTFN
  self.addCleanup(support.unlink, filename) 
  pathfix_arg = filename

... subprocess.run([sys.executable, self.script, *pathfix_flags, '-n', pathfix_arg], ...) ...
```

rpluem: Wouldn't adding the self.addCleanup call to pathfix cause the cleanup to be added multiple times (each time we call pathfix)? And if yes, is this really desired?
When are the cleanups run? After each single test or after all tests of this class have been executed?

vstinner: Right. One solution is to use try/finally to manually call support.unlink(filename) if directory is not set.

rpluem: Any reason why you don't like the setUp method and the addCleanup approach and prefer a local try / finally solution?

vstinner: Well, you can keep setUp() as it is if you prefer.

Misc/NEWS.d/next/Tools/Demos/2019-10-02-09-48-42.bpo-38347.2Tq5D1.rst
```diff
@@ -0,0 +1 @@
+pathfix.py: Assume all files that end on '.py' are Python scripts when working recursively.
```
vstinner: I'm not sure how you created this file, but blurb uses the directory "Misc/NEWS.d/next/Tools-Demos/", not "Misc/NEWS.d/next/Tools/Demos/".

Lib/test/test_tools/test_pathfix.py
```diff
@@ -44,6 +51,19 @@ def pathfix(self, shebang, pathfix_flags, exitcode=0, stdout='', stderr=''):
 
         return new_shebang
 
+    def test_recursive(self):
+        tmpdir = support.TESTFN + '.d'
+        self.addCleanup(support.rmtree, tmpdir)
+        os.mkdir(tmpdir)
+        expected_stderr = f'recursedown(\'{os.path.basename(tmpdir)}\')\n'
```
vstinner:
```suggestion
        expected_stderr = f"recursedown('{os.path.basename(tmpdir)}')\n"
```

Lib/test/test_tools/test_pathfix.py
```diff
@@ -14,25 +14,32 @@ class TestPathfixFunctional(unittest.TestCase):
...
     def setUp(self):
-        self.temp_file = support.TESTFN
         self.addCleanup(support.unlink, support.TESTFN)
 
-    def pathfix(self, shebang, pathfix_flags, exitcode=0, stdout='', stderr=''):
-        with open(self.temp_file, 'w', encoding='utf8') as f:
+    def pathfix(self, shebang, pathfix_flags, exitcode=0, stdout='', stderr='',
+                directory=''):
+        if directory:
+            filename = os.path.join(directory, 'script-A_1.py')
```
vstinner, line 24:
```suggestion
            filename = os.path.join(directory, 'script.py')
```

vstinner (changes requested):

Misc/NEWS.d/next/Tools/Demos/2019-10-02-09-48-42.bpo-38347.2Tq5D1.rst
rpluem: Weird. It was created by blurb in d8e988efe85bfef307258d06bbeae72e4a21d1fe. Fixed in a17410299b88bb3063daaed18f67efe70c9f2a2c.

Lib/test/test_tools/test_pathfix.py
rpluem: This is on purpose. With 'script.py' the test would not have detected the issue that this pull request fixes. And I think a test case for a fix should fail without the fix and succeed afterwards. We can discuss though how we add a capital letter, a '_', a number and a '-' more nicely to the name. I am also fine if I should add a comment that explains the weird script name.

vstinner: Yeah, please add a comment like "# bpo-38347: Test filename containing uppercase, "-", "_" and digit". Otherwise, someone may remove it because it looks like a typo.

rpluem: Done in 56eaa9441a3f56420817634d910c2575ff288131.

rpluem:
The Azure failure seems to be cause by the inability do download OpenSSL sources for the Ubuntu tests. Maybe a temporary network issue on Azure / Openssl side. I would like to reexecute the Azure pipeline, but I don't know how.

vstinner:
I closed/reopened the PR to schedule a new Azure Pipelines run.

Lib/test/test_tools/test_pathfix.py
```diff
@@ -14,25 +14,34 @@ class TestPathfixFunctional(unittest.TestCase):
...
     def setUp(self):
-        self.temp_file = support.TESTFN
         self.addCleanup(support.unlink, support.TESTFN)
 
-    def pathfix(self, shebang, pathfix_flags, exitcode=0, stdout='', stderr=''):
-        with open(self.temp_file, 'w', encoding='utf8') as f:
+    def pathfix(self, shebang, pathfix_flags, exitcode=0, stdout='', stderr='',
+                directory=''):
+        if directory:
+            # bpo-38347: Test filename should contain lowercase, upercase,
```
vstinner: type: uppercase
```suggestion
            # bpo-38347: Test filename should contain lowercase, uppercase,
```

vstinner (approved):
LGTM, but please fix the typo. It seems like I'm not allowed to push into your branch.

PR #16544, opened 2019-10-02
[WIP] bpo-38350: pydebug now uses -O0 compiler flag

vstinner:
When using ./configure --with-pydebug, -O0 optimization level is now
used instead of -Og. In practice, -O0 provides a better debugging
experience in gdb than -Og.

https://bugs.python.org/issue38350


vstinner:
I'm not sure that it's a good idea: see https://bugs.python.org/issue38350 discussion.

PR #16547, opened 2019-10-02
bpo-38366: dataclasses: generate the _hash_action table more cleanly

ioistired:
currently it's a table, which is more efficient than an if-else chain,
but less efficient. We can have the best of both worlds by converting https://bugs.python.org/issue32929#msg312829
into a loop that generates the table that we currently have. This should make the table easier to maintain and understand.

https://bugs.python.org/issue38366

### DIFF
--- a/Lib/dataclasses.py
+++ b/Lib/dataclasses.py
@@ -6,6 +6,7 @@ import inspect
 import keyword
 import builtins
 import functools
+import itertools
 import _thread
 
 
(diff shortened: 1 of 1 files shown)


the-knights-who-say-ni:

## Recognized GitHub username

@iomintz

ericvsmith:
I'm going to reject this PR. See my rationale in the BPO issue.

PR #16548, opened 2019-10-02
bpo-38353: Cleanup includes in the internal C API

vstinner:
Use forward declaration of types to avoid includes in the internal C
API. Add also comment to justify other includes.

### DIFF
--- a/Include/internal/pycore_ceval.h
+++ b/Include/internal/pycore_ceval.h
@@ -8,9 +8,9 @@ extern "C" {
 #  error "this header requires Py_BUILD_CORE define"
 #endif
 
-#include "pycore_atomic.h"
-#include "pycore_pystate.h"
-#include "pythread.h"
+/* Forward declarations */
+typedef struct pyruntimestate _PyRuntimeState;
+struct _ceval_runtime_state;
 
 PyAPI_FUNC(void) _Py_FinishPendingCalls(_PyRuntimeState *runtime);
 PyAPI_FUNC(void) _PyEval_Initialize(struct _ceval_runtime_state *);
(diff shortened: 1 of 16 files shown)


brettcannon (approved):

vstinner:
@brettcannon: Thanks for the review ;-)

Include/internal/pycore_ceval.h
```diff
@@ -8,9 +8,9 @@ extern "C" {
 #  error "this header requires Py_BUILD_CORE define"
 #endif
 
-#include "pycore_atomic.h"
-#include "pycore_pystate.h"
-#include "pythread.h"
+/* Forward declarations */
+typedef struct pyruntimestate _PyRuntimeState;
```
ncoghlan, line 12: The fact that 3 different headers need this same typedef makes me wonder if it might be worth having a short header called `pycore_state_typedefs.h` that was enough to allow declaration of APIs accepting state pointers, without needing to include the full state struct definitions.

Something like:

```
typedef struct _PyArgv _PyArgv;
typedef struct pyruntimestate _PyRuntimeState;
typedef struct _is PyInterpreterState;
struct _ceval_runtime_state;
```

vstinner:
> The fact that 3 different headers need this same typedef makes me wonder if it might be worth having a short header called pycore_state_typedefs.h that was enough to allow declaration of APIs accepting state pointers, without needing to include the full state struct definitions.

Adding a new header file requires to modify the Unix (Makefile.pre.in) and the Windows (Visual Studio project files, at least 2 files) build systems. IMHO it's too annoying. I'm fine with duplicating these forward definitions.

ncoghlan:
Hmm, I thought it was only new C files that were that annoying to add, but I admit it's been ages since I added a new header file.

PR #16549, opened 2019-10-02
bpo-38353: Fix calculate_argv0_path() for symlinks

vstinner:
calculate_argv0_path() must read argv0_path link, not read
program_full_path link.

### DIFF
--- a/Modules/getpath.c
+++ b/Modules/getpath.c
@@ -959,7 +959,7 @@ calculate_argv0_path(PyCalculatePath *calculate, const wchar_t *program_full_pat
 #if HAVE_READLINK
     wchar_t tmpbuffer[MAXPATHLEN + 1];
     const size_t buflen = Py_ARRAY_LENGTH(tmpbuffer);
-    int linklen = _Py_wreadlink(program_full_path, tmpbuffer, buflen);
+    int linklen = _Py_wreadlink(argv0_path, tmpbuffer, buflen);
     while (linklen != -1) {
         if (_Py_isabs(tmpbuffer)) {
             /* tmpbuffer should never be longer than MAXPATHLEN,


vstinner:
This change only affects ("#ifdef WITH_NEXT_FRAMEWORK").

vstinner:
The bug is that on macOS, readlink() is called on program_full_path (previously known as "progpath"), instead of argv0_path.

The change likely has no effect on practice. I don't expect program_full_path nor argv0_path to be symlinks on macOS.

The bug is quite old, it was introduced by:

```
commit 55070f5d966f09256c603c3a82fab9b034430c6f
Author: Jack Jansen <jack.jansen@cwi.nl>
Date:   Sun Dec 2 23:56:28 2001 +0000

    Changed logic for finding python home in Mac OS X framework Pythons.
    Now sys.executable points to the executable again, in stead of to
    the shared library. The latter is used only for locating the python
    home.
```

Previously, at the first readlink() call, argv0_path was equal to progpath:

```
        strcpy(argv0_path, progpath);
        
#if HAVE_READLINK
        {
                char tmpbuffer[MAXPATHLEN+1];
                int linklen = readlink(progpath, tmpbuffer, MAXPATHLEN);
                if (linklen != -1) {
                        /* It's not null terminated! */
                        tmpbuffer[linklen] = '\0';
                        if (tmpbuffer[0] == SEP)
                                strcpy(argv0_path, tmpbuffer);
                        else {
                                /* Interpret relative to progpath */
                                reduce(argv0_path);
                                join(argv0_path, tmpbuffer);
                        }
                }
        }
#endif /* HAVE_READLINK */
```

Previously = the following commit:

```
commit 305e5d0d4f3a24f199cfea4f9b0e53980de79847
Author: Guido van Rossum <guido@python.org>
Date:   Fri Apr 11 17:18:45 1997 +0000

    Completely revamped the way the default path is constructed.
    
    Idea and first three implementation rounds due to Barry -- after that
    I spent another day on it, hopefully it's enough for now :-)
    
    (Wait for the checkin to Setup.in.)
```

vstinner:
Since Python is working well since 2001 on macOS, I don't think that it's worth it to backport this change to Python 3.8 and older.

PR #16551, opened 2019-10-02
bpo-38355: Fix ntpath.realpath failing on sys.executable

zooba:
https://bugs.python.org/issue38355


zooba:
Turns out, the error check also dealt with the surprise new prefixes that showed up in the tests.

PR #16553, opened 2019-10-03
bpo-36670: Enhance regrtest WindowsLoadTracker

vstinner:
The last line is now passed to the parser even if it does not end
with a newline, but only if it's a valid value.

### DIFF
--- a/Lib/test/libregrtest/win_utils.py
+++ b/Lib/test/libregrtest/win_utils.py
@@ -14,7 +14,7 @@ BUFSIZE = 8192
 LOAD_FACTOR_1 = 0.9200444146293232478931553241
 
 # Seconds per measurement
-SAMPLING_INTERVAL = 5
+SAMPLING_INTERVAL = 1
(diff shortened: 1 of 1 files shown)


Lib/test/libregrtest/win_utils.py
```diff
@@ -14,7 +14,7 @@
 LOAD_FACTOR_1 = 0.9200444146293232478931553241
 
 # Seconds per measurement
-SAMPLING_INTERVAL = 5
+SAMPLING_INTERVAL = 1
```
ammaraskar, line 17: I haven't taken a deeper look at the rest of the code but if the values are actually being sampled at a faster rate, the `LOAD_FACTOR_1` constant needs to be updated. It assumes a sample rate of 1/s right now.

vstinner: Oh. I didn't know that LOAD_FACTOR_1 value depends on SAMPLING_INTERVAL: I wrote PR #16555 to restore SAMPLING_INTERVAL=5.

PR #16555, opened 2019-10-03
bpo-36670, regrtest: Fix WindowsLoadTracker sampling rate

vstinner:
regrtest WindowsLoadTracker: the LOAD_FACTOR_1 constant has been
computed for a sampling rate of 5 seconds.

### DIFF
--- a/Lib/test/libregrtest/win_utils.py
+++ b/Lib/test/libregrtest/win_utils.py
@@ -10,11 +10,11 @@ from test.libregrtest.utils import print_warning
 
 # Max size of asynchronous reads
 BUFSIZE = 8192
-# Exponential damping factor (see below)
+# Seconds per measurement
+SAMPLING_INTERVAL = 5
+# Exponential damping factor for a sampling interval of 5 seconds (see below)
 LOAD_FACTOR_1 = 0.9200444146293232478931553241
 
-# Seconds per measurement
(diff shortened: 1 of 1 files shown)


Lib/test/libregrtest/win_utils.py
```diff
@@ -10,11 +10,11 @@
 
 # Max size of asynchronous reads
 BUFSIZE = 8192
-# Exponential damping factor (see below)
+# Seconds per measurement
+SAMPLING_INTERVAL = 5
+# Exponential damping factor for a sampling interval of 5 seconds (see below)
```
ammaraskar, line 15: Thanks for the comment, I should have had this in here originally. I think it might be worth adding a little explanation like:
```python
# Calculated using 1/exp(5sec/1min) = 1/exp(5/60)
```
the source is https://github.com/torvalds/linux/blob/master/include/linux/sched/loadavg.h#L21

If you feel like 1 second is better, then the new constant would be `0.98347145382161748947374` https://www.wolframalpha.com/input/?i=1%2Fexp%281%2F60%29

ammaraskar (approved):
LGTM with either increasing the sampling interval or using the new 1s constant.

vstinner:
I merged PR #16556 instead.

PR #16557, opened 2019-10-03
bpo-38361: syslog: fixed making default "ident" from sys.argv[0]

vaclavbartos:
The default value of "ident" parameter should be sys.argv[0] with leading path
components stripped, but it contained the last slash, i.e. '/program' instead
of 'program'.

BPO issue: https://bugs.python.org/issue38361

https://bugs.python.org/issue38361

### DIFF
--- a/Misc/NEWS.d/next/Library/2019-10-04-09-49-56.bpo-38361.LM4u4T.rst
+++ b/Misc/NEWS.d/next/Library/2019-10-04-09-49-56.bpo-38361.LM4u4T.rst
@@ -1,0 +1,1 @@
+Fixed an issue where ``ident`` could include a leading path separator when :func:`syslog.openlog` was called without arguments.
--- a/Modules/syslogmodule.c
+++ b/Modules/syslogmodule.c
@@ -99,7 +99,7 @@ syslog_get_argv(void)
     if (slash == -2)
         return NULL;
     if (slash != -1) {
-        return PyUnicode_Substring(scriptobj, slash, scriptlen);
+        return PyUnicode_Substring(scriptobj, slash + 1, scriptlen);
     } else {
         Py_INCREF(scriptobj);
         return(scriptobj);


the-knights-who-say-ni:

## CLA Missing

@vaclavbartos

brandtbucher:
Thanks for your time @vaclavbartos, and welcome to CPython! 😎

This looks like a good change. I assume you’ve already seen the bot’s message about the CLA?

brandtbucher (changes requested):
Even though this is a fairly trivial fix, I still think you should [make a NEWS entry](https://blurb-it.herokuapp.com), since it changes logging output. Just something simple, like:

```rst
Fixed an issue where ``ident`` could include a leading path separator when :func:`syslog.openlog` was called without arguments.
```

And while a regression test or two are always good to include, it looks like this module has almost no test coverage currently due to platform issues. So, *probably* not necessary in this case.

vaclavbartos:
OK, I added the News entry. I signed the CLA yesterday and it's checked now.

Regarding tests - as you noted, it would be quite difficult to implement them, since there is currently no other way to read `ident` than to emit a syslog message, read it from a syslog file and parse `ident` from it - which is highly platfrom dependent.

brandtbucher:
This module doesn't have any experts with GitHub accounts, and hasn't had any real changes in almost a decade. Given that this is a fairly trivial change, perhaps another core dev could look at it?

CC @vstinner

gvanrossum (approved):
Yeah, this looks good. I agree that this looks like a mis-translation of the original code, apparently introduced in revision d63a3b8beb4a0841cb59fb3515347ccaab34b733.

gvanrossum:
Should we backport this to 3.8? Then the behavior would be different in 3.8.1 than in 3.8.0.

brandtbucher:
Even though it's a bug fix, I'm leaning "no". Logging output shouldn't change between minor versions, and I don't know of anyone who sees this as a must-have.

It's just another reason 3.9 will be better than 3.8.

brandtbucher:
Going through some of my old reviews and saw that this was still unmerged. @gvanrossum, is this waiting for anything?

gvanrossum:
Sorry! It should be merged now.

brandtbucher:
Congrats on your first CPython contribution @vaclavbartos! :champagne:

PR #16558, opened 2019-10-03
bpo-38266: Revert bpo-37878: Make PyThreadState_DeleteCurrent() Internal

nanjekyejoannah:
Reverting the removal of  ``PyThreadState_DeleteCurrent()`` with documentation.

https://bugs.python.org/issue38266

### DIFF
--- a/Doc/c-api/init.rst
+++ b/Doc/c-api/init.rst
@@ -1035,6 +1035,14 @@ All of the following functions must be called after :c:func:`Py_Initialize`.
    :c:func:`PyThreadState_Clear`.
 
 
+   .. c:function:: void PyThreadState_DeleteCurrent()
+
+    Destroy the current thread state and release the global interpreter lock.
+    Like :c:func:`PyThreadState_Delete`, the global interpreter lock need not
+    be held. The thread state must have been reset with a previous call
+    to :c:func:`PyThreadState_Clear`.
+
+
 .. c:function:: PY_INT64_T PyInterpreterState_GetID(PyInterpreterState *interp)
 
    Return the interpreter's unique ID.  If there was any error in doing
--- a/Include/cpython/pystate.h
+++ b/Include/cpython/pystate.h
@@ -183,7 +183,7 @@ PyAPI_FUNC(PyInterpreterState *) PyInterpreterState_Head(void);
 PyAPI_FUNC(PyInterpreterState *) PyInterpreterState_Next(PyInterpreterState *);
 PyAPI_FUNC(PyThreadState *) PyInterpreterState_ThreadHead(PyInterpreterState *);
 PyAPI_FUNC(PyThreadState *) PyThreadState_Next(PyThreadState *);
-PyAPI_FUNC(void) PyThreadState_DeleteCurrent();
+PyAPI_FUNC(void) PyThreadState_DeleteCurrent(void);
 
 typedef struct _frame *(*PyThreadFrameGetter)(PyThreadState *self_);
 
--- a/Include/cpython/pystate.h
+++ b/Include/cpython/pystate.h
@@ -183,6 +183,7 @@ PyAPI_FUNC(PyInterpreterState *) PyInterpreterState_Head(void);
 PyAPI_FUNC(PyInterpreterState *) PyInterpreterState_Next(PyInterpreterState *);
 PyAPI_FUNC(PyThreadState *) PyInterpreterState_ThreadHead(PyInterpreterState *);
 PyAPI_FUNC(PyThreadState *) PyThreadState_Next(PyThreadState *);
+PyAPI_FUNC(void) PyThreadState_DeleteCurrent();
 
 typedef struct _frame *(*PyThreadFrameGetter)(PyThreadState *self_);
 
--- a/Include/internal/pycore_pylifecycle.h
+++ b/Include/internal/pycore_pylifecycle.h
@@ -108,7 +108,8 @@ PyAPI_FUNC(PyObject*) _PyErr_WriteUnraisableDefaultHook(PyObject *unraisable);
 
 PyAPI_FUNC(void) _PyErr_Print(PyThreadState *tstate);
 PyAPI_FUNC(void) _PyErr_Display(PyObject *file, PyObject *exception,
-                                PyObject *value, PyObject *tb);
+                                PyObject *value, PyObject *tb);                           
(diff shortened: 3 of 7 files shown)


Doc/c-api/init.rst
```diff
@@ -1035,6 +1035,14 @@ All of the following functions must be called after :c:func:`Py_Initialize`.
    :c:func:`PyThreadState_Clear`.
 
 
+   .. c:function:: void PyThreadState_DeleteCurrent()
+
+    Destroy the current thread state and release the global interpreter lock.
+    Like :c:func:`PyThreadState_Delete`, the global interpreter lock need not
+    be held. The thread state must have been reset with a previous call
+    to :c:func:`PyThreadState_Clear`.
```
vstinner, line 1043: If we start to "expose" this function, would it be possible to be less strict and call PyThreadState_Clear() in PyThreadState_DeleteCurrent()? It sems like calling PyThreadState_Clear() twice is safe.

Include/internal/pycore_pylifecycle.h
```diff
@@ -110,8 +110,6 @@ PyAPI_FUNC(void) _PyErr_Print(PyThreadState *tstate);
 PyAPI_FUNC(void) _PyErr_Display(PyObject *file, PyObject *exception,
                                 PyObject *value, PyObject *tb);
 
-PyAPI_FUNC(void) _PyThreadState_DeleteCurrent(_PyRuntimeState *runtime);
```
vstinner: I don't understand why do you remove this definition: it's still used, no?

Include/cpython/pystate.h
```diff
@@ -183,6 +183,7 @@ PyAPI_FUNC(PyInterpreterState *) PyInterpreterState_Head(void);
 PyAPI_FUNC(PyInterpreterState *) PyInterpreterState_Next(PyInterpreterState *);
 PyAPI_FUNC(PyThreadState *) PyInterpreterState_ThreadHead(PyInterpreterState *);
 PyAPI_FUNC(PyThreadState *) PyThreadState_Next(PyThreadState *);
+PyAPI_FUNC(void) PyThreadState_DeleteCurrent();
```
vstinner: The prototype must be:

```suggestion
PyAPI_FUNC(void) PyThreadState_DeleteCurrent(void);
```

Modules/_tracemalloc.c
```diff
@@ -34,7 +34,7 @@ static struct {
 #if defined(TRACE_RAW_MALLOC)
 /* This lock is needed because tracemalloc_free() is called without
    the GIL held from PyMem_RawFree(). It cannot acquire the lock because it
-   would introduce a deadlock in _PyThreadState_DeleteCurrent(). */
+   would introduce a deadlock in PyThreadState_DeleteCurrent(). */
```
vstinner: I prefer to keep _PyThreadState_DeleteCurrent() here. PyThreadState_DeleteCurrent() is only a public helper function, _PyThreadState_DeleteCurrent() is the real implementation.

Modules/_tracemalloc.c
```diff
@@ -728,7 +728,7 @@ tracemalloc_free(void *ctx, void *ptr)
         return;
 
      /* GIL cannot be locked in PyMem_RawFree() because it would introduce
-        a deadlock in _PyThreadState_DeleteCurrent(). */
+        a deadlock in PyThreadState_DeleteCurrent(). */
```
vstinner: ditto: keep _PyThreadState_DeleteCurrent()

Python/pystate.c
```diff
@@ -874,6 +874,12 @@ _PyThreadState_DeleteCurrent(_PyRuntimeState *runtime)
     PyEval_ReleaseLock();
 }
 
+void
+PyThreadState_DeleteCurrent()
```
vstinner: Please add void to prevent accepting any argument:

```suggestion
PyThreadState_DeleteCurrent(void)
```

vstinner:
Once this change lands, it would be interested to add an unit test, but I'm not sure how to write such test. C code in _testcapi I guess.

Include/internal/pycore_pylifecycle.h
nanjekyejoannah: I am now exposing PyThreadState_DeleteCurrent() instead of the internal  _PyThreadState_DeleteCurrent() I thought. what are you thinking?

nanjekyejoannah:
@vstinner , I have addressed the changes. PTAL

Doc/c-api/init.rst
vstinner: I marked https://bugs.python.org/issue38266 as a release blocker, so the PyThreadState_Clear() question can definitely wait after 3.8.0: we can even revisit this in Python 3.9. Let's stick to adding back the function unmodified.

Include/internal/pycore_pylifecycle.h
vstinner: _PyThreadState_DeleteCurrent() has to stay, it's part of https://pythoncapi.readthedocs.io/runtime.html effort to avoid global variables.

vstinner (approved):
LGTM.

Include/internal/pycore_pylifecycle.h
```diff
@@ -108,8 +108,7 @@ PyAPI_FUNC(PyObject*) _PyErr_WriteUnraisableDefaultHook(PyObject *unraisable);
 
 PyAPI_FUNC(void) _PyErr_Print(PyThreadState *tstate);
 PyAPI_FUNC(void) _PyErr_Display(PyObject *file, PyObject *exception,
-                                PyObject *value, PyObject *tb);
-
+                                PyObject *value, PyObject *tb);
```
vstinner: nitpick: it seems like you removed a newline or something else. you may revert that.

vstinner:
According to https://bugs.python.org/issue38266 discussion, it's a Python 3.8 regression and so this change must be backported to 3.8. I suggest to merge this change into 3.8 and master as soon as possible, to attempt to get it into 3.8.0 final. (But it's ok if we miss the final, it can wait for 3.8.1 otherwise.)

vstinner (approved):
LGTM.

vstinner:
Usually, we let core developers merge their own pull requests. But. Joannah explicitly asked me to merge her PR if I want. Since this change fix a Python 3.8 regression and the next 3.8.0 release (which may be a rc2 or the final) is in 3 days, I prefer to merge it ASAP.

PR #16567, opened 2019-10-03
bpo-34938: mimetypes.init() does not overwrite globals

AWhetter:
When importing mimetypes.init() into a module scope with
any of encodings_map, suffix_map, types_map, or common_types,
subsequent calls to init() would update the maps in
the local module scope and not in the mimetypes module.
This change makes the original maps always get updated.

This will break the undocumented behaviour of any changes to `mimetypes.encodings_map`, `mimetypes.suffix_map`, `mimetypes.types_map`, or `mimetypes.common_types` persisting after calling `mimetypes.init()` because the maps and their equivalent `mimetypes._default<map>` variable are now two separate dictionaries.

https://bugs.python.org/issue34938


the-knights-who-say-ni:

## CLA Missing

@AWhetter

brandtbucher:
Thanks for your time @AWhetter, and welcome to CPython! 😎

I assume you've seen the bot's message about the CLA?

brandtbucher (changes requested):
This looks like a good fix. I think it also needs a regression test in `Lib/test/test_mimetypes.py`.

Looking at the failed test, though, it seems that the current behavior of creating new objects rather than updating the existing ones is expected (though I *think* I agree with you that it's flawed).

Maybe someone from @python/email-team can shed some light on the desired behavior here?

AWhetter:
I have seen the CLA notice. I've already signed up so I'm waiting for the approval to go through  the system at the moment.

Thanks for the fast feedback! I'll get a test added!

zooba:
I posted on the bpo issue, as my comments ended up at a higher level than simple PR feedback.

Also, no need to force push changes. We'll squash merge at the end, and force pushing only breaks the review system.

maxking (approved):
LGTM

Mariatta:
@maxking is this PR ready to merge? 

@zooba can you check the question in https://bugs.python.org/issue34938#msg359070, do you expect those to be worked on this PR, or separate? 

Thanks.

AWhetter:
I've made the suggested documentation changes. The bit about importing `inited` into a local module might require more information. I'm tempted to add a section about it under `/faq/programming.rst` because it could have use elsewhere. What do we think?

YoSTEALTH (changes requested):
There is no need to have line: 388-392 and line: 571 in `Lib/mimetypes.py` is there? Like take out the `_default_mime_types` function and indent back the content. Thanks

Lib/mimetypes.py
```diff
@@ -388,21 +391,23 @@ def _default_mime_types():
     global types_map, _types_map_default
     global common_types, _common_types_default
```
zooba: These `global` statements and the function (and its call) can be removed.

YoSTEALTH (changes requested):
these change must be reverted back, you need to have it as:
``suffix_map = _suffix_map_default.copy()`` not ``suffix_map = _suffix_map_default = {...}`` because they are suppose to be 2 separate variable. On https://github.com/python/cpython/blob/d8fb9e1df51746ab6e26bdb68ab85fd0c0dd042c/Lib/mimetypes.py#L366 ``suffix_map.clear()`` is called and if they are same variable it will be bad.

iritkatriel (changes requested):
This PR has merge conflicts now.

PR #16571, opened 2019-10-03
bpo-38235: Correct some arguments names in logging documentation

AWhetter:
There's some other arguments names that are also incorrect in the documentation. For example `Formatter.formatException` is defined as `def formatException(self, ei):` but documented as `.. method:: formatException(exc_info)`, and `makeLogRecord` is defined as `def makeLogRecord(dict):` but documented as `.. function:: makeLogRecord(attrdict)`. The argument names in the documentation are more clear, but they don't match the code. I'm not sure if the preference is for consistency or readability. So for now I've only changed `lvl` because `level` is a better name, but I'm happy to find and fix the other inconsistencies if that's preferred.

https://bugs.python.org/issue38235

### DIFF
--- a/Doc/library/logging.rst
+++ b/Doc/library/logging.rst
@@ -124,11 +124,11 @@ is the module's name in the Python package namespace.
          :meth:`isEnabledFor` will return/expect to be passed integers.
 
 
-   .. method:: Logger.isEnabledFor(lvl)
+   .. method:: Logger.isEnabledFor(level)
 
-      Indicates if a message of severity *lvl* would be processed by this logger.
+      Indicates if a message of severity *level* would be processed by this logger.
       This method checks first the module-level level set by
-      ``logging.disable(lvl)`` and then the logger's effective level as determined
+      ``logging.disable(level)`` and then the logger's effective level as determined
       by :meth:`getEffectiveLevel`.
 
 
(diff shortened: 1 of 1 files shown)


brandtbucher (approved):

aeros (approved):
Thanks for the PR @AWhetter and welcome!

> I'm not sure if the preference is for consistency or readability.

Typically, that decision is left up to the maintainer for the module, rather than a global rule for all of the documentation. In the case of `logging`, that is @vsajip. I'll add him as a reviewer. 

Personally, I'm in favor of changing *lvl* to *level* for consistency. I'm generally not a fan of abbreviating argument names. 

> I'm happy to find and fix the other inconsistencies if that's preferred.

As for fixing all of the other inconsistencies, let's keep this PR simple and focused just on the *lvl* to *level* conversions. That makes it much easier to review and makes merging the PR a much smaller decision for the maintainer to make. In it's current state, I think it works great as an introductory PR to gain some experience with the CPython workflow. (:

aeros167:
Also @vsajip would you mind being added to the CODEOWNERS file as an owner for anything related to the logging module? This would allow you to be automatically added as a reviewer for any PR that affects logging documentation or code files.

I don't mind opening a PR to add you if you don't have the time to do so, but I wanted to make sure you were okay with it first. Brett was saying that he wants to encourage populating CODEOWNERS with experts instead of manually adding reviewers.

freddrake (approved):

CuriousLearner (approved):
Looks good to me. 🌮

brandtbucher:
Congrats on your first CPython contribution @AWhetter! :champagne:

AWhetter:
Yay! Thank you! :raised_hands:

PR #16575, opened 2019-10-04
gh-81881: Raise a shutil.SpecialFileError when copying a Unix socket

AWhetter:
This used to raise an OSError with a platform dependent message.
This change always raises a SpecialFileError with a consistent message
no matter the platform.

Tackling special devices wasn't part of the original issue, but it is still technically an issue. Although devices are technically copyable (at least they are with `cp`), they will likely hang the same as a named pipe would even though a named pipe is technically copyable as well.

https://bugs.python.org/issue37700

* Issue: gh-81881


brandtbucher:
Thanks for the PR!

brandtbucher (changes requested):
This looks good to me! I would maybe just change the NEWS entry to be less wordy:

```rst
:func:`shutil.copyfile` always raises a :exc:`shutil.SpecialFileError` when trying to copy a Unix socket.
```

brandtbucher:
CC @giampaolo @cjerdonek

AWhetter:
Good idea. I'll get that changed.
When I make changes to a pull request, is it better for me to edit my existing commit, or should I make a new commit for squashing when the branch is merged?

brandtbucher:
I think it helps everyone follow the flow of the PR when things aren't force pushed and history isn't rewritten. It will get squashed in the end, anyways!

I just break up my commits so that the individual diffs are easy to follow. So if I'm editing *and* moving a function, I'll do those in separate commits in order to preserve an easy-to-follow storyline.

brandtbucher (approved):

chrahunt:
Similar to the issue mentioned in [bpo-37701](https://bugs.python.org/issue37701), this raises `SpecialFileError` when src or dest are symlinks to a socket file.

Lib/test/test_shutil.py
```diff
@@ -663,6 +664,29 @@ def test_copytree_named_pipe(self):
...
+            shutil.copytree(TESTFN, TESTFN2)
+        except shutil.Error as e:
+            errors = e.args[0]
+            self.assertEqual(len(errors), 1)
+            src, dst, error_msg = errors[0]
+            self.assertEqual("`%s` is a socket" % sock_path, error_msg)
+        else:
+            self.fail("shutil.Error should have been raised")
+        finally:
+            shutil.rmtree(TESTFN2, ignore_errors=True)
```
giampaolo: This seems unnecessary as per:

```
 self.addCleanup(shutil.rmtree, TESTFN, ignore_errors=True)
```

...some lines above.

Lib/test/test_shutil.py
```diff
@@ -663,6 +664,29 @@ def test_copytree_named_pipe(self):
             shutil.rmtree(TESTFN, ignore_errors=True)
             shutil.rmtree(TESTFN2, ignore_errors=True)
 
+    @unittest.skipUnless(hasattr(socket, 'AF_UNIX'), 'requires socket.AF_UNIX')
+    def test_copytree_socket(self):
```
giampaolo, line 915: IMO testing also `copytree()` is not very useful since internally it uses `copyfile` (which you are testing).

Doc/library/shutil.rst
```diff
@@ -54,6 +54,9 @@ Directory and files operations
    *dst* and return *dst* in the most efficient way possible.
    *src* and *dst* are path-like objects or path names given as strings.
 
+   When *src* is a named pipe or a Unix socket, a :exc:`SpecialFileError`
```
giampaolo:
```diff
-   When *src* is a named pipe or a Unix socket, a :exc:`SpecialFileError`
+   When *src* is a named pipe or a Unix socket, :exc:`SpecialFileError`
```

giampaolo (changes requested):

Lib/test/test_shutil.py
AWhetter: I was copying what was done for named pipes for consistency. But having a test for `copytree()` makes sense to me because if in the future `copytree()` stopped using `copyfile()` for some reason, we would still want `copytree()` to raise an error?

AWhetter:

AWhetter:
Is there anything else that I can do to help progress this review?

oz123:
@giampaolo can this PR be unblocked?

oz123:
@arhadthedev thanks for the quick response!

arhadthedev:
`shutil.SpecialFileError` seems to be undocumented. Would anybody add it into the docmentation as a separate PR?

PR #16579, opened 2019-10-04
bpo-37025: AddRefActCtx() shouldn't be checked for failure

ZackerySpytz:
AddRefActCtx() does not return a value.

https://bugs.python.org/issue37025


zooba:
Closing this - I'd rather just delete all the HAVE_SXS code.

PR #16586, opened 2019-10-04
[2.7] bpo-38235: Correct some arguments names in logging documentation (GH-16571)

AWhetter:
(cherry picked from commit 3142c667b50254daaa28f22c79bdda177136bd03)

Co-authored-by: Ashley Whetter <AWhetter@users.noreply.github.com>

https://bugs.python.org/issue38235

### DIFF
--- a/Doc/library/logging.rst
+++ b/Doc/library/logging.rst
@@ -117,11 +117,11 @@ is the module's name in the Python package namespace.
    See :ref:`levels` for a list of levels.
 
 
-.. method:: Logger.isEnabledFor(lvl)
+.. method:: Logger.isEnabledFor(level)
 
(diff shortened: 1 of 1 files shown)


AWhetter:
@vsajip I just realised that you removed the Python 2.7 version from the issue before merging, which is why @miss-islington didn't backport. Feel free to decline this if you think it doesn't belong in 2.7.

brandtbucher:
What happened here? I don't think #16571 was tagged for backport to 2.7!

vsajip:
> I just realised that you removed the Python 2.7 version from the issue before merging, which is why @miss-islington didn't backport.

No, FYI `miss-islington` only creates backport PRs when a label is explicitly set here - and I didn't set one for 2.7, as 2.7 will be end-of-lifed soon.

PR #3027, opened 2017-08-08
bpo-26510: Allow argparse add_subparsers to take `required` kwarg.

asottile:
I added both a test to demonstrate the current behaviour / workaround as well as the new behaviour.

I'm torn between setting the default `required=True` (which was the python2 behaviour) and setting `required=False` (the current python3 behaviour).

If we're ok with `required=True`, I'll change that up :)

https://bugs.python.org/issue26510

PR #7609, opened 2018-06-11
bpo-33109: Remove now-obsolete What's New entry for bpo-26510.

ned-deily:
https://bugs.python.org/issue33109

PR #6919, opened 2018-05-16
bpo-33109: argparse subparsers are once again not required by default

ned-deily:
bpo-26510 in 3.7.0a2 changed the behavior of argparse to make
subparsers required by default, returning to the behavior of 2.7
and 3.2. The behavior was changed in 3.3 to be no longer required.
While it might make more sense to have the default to required,
compatibility with 3.3 through 3.6 is probably less disruptive
than trying to reintroduce compatibility with 2.7 at this point.
This change restores the 3.6 behavior.

https://bugs.python.org/issue33109

PR #16588, opened 2019-10-04
 bpo-26510: Add versionchanged for required arg of add_subparsers

adamjstewart:
The `required` argument to `argparse.add_subparsers` was added in #3027. This PR specifies the earliest version of Python where it is available.

https://bugs.python.org/issue26510

Automerge-Triggered-By: @merwok

### DIFF
--- a/Doc/library/argparse.rst
+++ b/Doc/library/argparse.rst
@@ -1636,7 +1636,7 @@ Sub-commands
      stored; by default ``None`` and no value is stored
 
    * required_ - Whether or not a subcommand must be provided, by default
-     ``False``.
+     ``False``. (added in 3.7)
 
    * help_ - help for sub-parser group in help output, by default ``None``
 
--- a/Doc/library/argparse.rst
+++ b/Doc/library/argparse.rst
@@ -1792,6 +1792,9 @@ Sub-commands
      >>> parser.parse_args(['2', 'frobble'])
      Namespace(subparser_name='2', y='frobble')
 
+   .. versionadded:: 3.7
+      The *required* keyword argument.
+
 
 FileType objects
 ^^^^^^^^^^^^^^^^
--- a/Doc/library/argparse.rst
+++ b/Doc/library/argparse.rst
@@ -1636,7 +1636,7 @@ Sub-commands
      stored; by default ``None`` and no value is stored
 
    * required_ - Whether or not a subcommand must be provided, by default
-     ``False``. (added in 3.7)
+     ``False`` (added in 3.7)
 
    * help_ - help for sub-parser group in help output, by default ``None``
 
@@ -1792,8 +1792,8 @@ Sub-commands
      >>> parser.parse_args(['2', 'frobble'])
      Namespace(subparser_name='2', y='frobble')
 
-   .. versionadded:: 3.7
-      The *required* keyword argument.
+   .. versionchanged:: 3.7
+      New *required* keyword argument.
 
 
 FileType objects
--- a/Doc/whatsnew/3.7.rst
+++ b/Doc/whatsnew/3.7.rst
@@ -2401,8 +2401,8 @@ Changes in the Python API
   instead of a :class:`bytes` instance.
   (Contributed by Victor Stinner in :issue:`21071`.)
 
-* :meth:`ArgumentParser.add_subparsers() <argparse.ArgumentParser.add_subparsers>`
-  has a new ``required`` parameter that defaults to ``False``.
+* :mod:`argparse` subparsers can now be made mandatory by passing ``required=True``
+  to :meth:`ArgumentParser.add_subparsers() <argparse.ArgumentParser.add_subparsers>`.
   (Contributed by Anthony Sottile in :issue:`26510`.)
 
 * :meth:`ast.literal_eval()` is now stricter.  Addition and subtraction of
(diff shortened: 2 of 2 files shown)


the-knights-who-say-ni:

## CLA Missing

@adamjstewart

Doc/library/argparse.rst
```diff
@@ -1792,6 +1792,9 @@ Sub-commands
      >>> parser.parse_args(['2', 'frobble'])
      Namespace(subparser_name='2', y='frobble')
 
+   .. versionadded:: 3.7
```
merwok:
```suggestion
   .. versionchanged:: 3.7
```

If I recall correctly, the version* directive apply to their parent object block (module, class, function, data), so this is more correct.  We don’t want `add_subparsers` to show up in a list of functions added in 3.7 :slightly_smiling_face:

merwok (changes requested):
Thanks for adding this piece that was missed in the original PR!

Because this part of the docs is super long, could you also add something like `(added in 3.7)` on line 1639?  Would help readers get the version info right when they read about `required`.

Doc/library/argparse.rst
adamjstewart: I just copied this syntax from https://docs.python.org/3/library/argparse.html#argparse.FileType, should I fix that too while I'm at it?

adamjstewart: https://www.sphinx-doc.org/en/1.5/markup/para.html#directive-versionadded seems inconclusive. They use `versionadded` and `versionchanged` for "new parameter" examples. I'll defer to your judgement on this one.

merwok:
The what’s new note was removed by #7609 after the default value was changed in #6919, but now I realize that this was too much: subparsers are not required, but there is a new param!  Would you @adamjstewart mind re-adding a note in what’s new for 3.7?

(also please say if you need any help with the CLA)

adamjstewart:
CLA has been signed 😄

merwok:
Docs are a little vague indeed!  From https://devguide.python.org/documenting/#paragraph-level-markup

> **versionadded**
>    This directive documents the version of Python which added the described feature, or a part of it, to the library

But:

>  **versionchanged**
>     Similar to versionadded, but describes when and what changed in the named feature in some way (new parameters, changed side effects, platform support, etc.)

`new parameters` seems pretty clear here :slightly_smiling_face: 

I wouldn’t change FileType or other files in this PR.

Doc/library/argparse.rst
```diff
@@ -1636,7 +1636,7 @@ Sub-commands
      stored; by default ``None`` and no value is stored
 
    * required_ - Whether or not a subcommand must be provided, by default
-     ``False``.
+     ``False``. (added in 3.7)
```
merwok:
```suggestion
     ``False`` (added in 3.7)
```

Other parameters don’t have final period (and if they had, would be after the parens I think)

Doc/whatsnew/3.7.rst
```diff
@@ -2401,6 +2401,10 @@ Changes in the Python API
   instead of a :class:`bytes` instance.
   (Contributed by Victor Stinner in :issue:`21071`.)
 
+* :meth:`ArgumentParser.add_subparsers() <argparse.ArgumentParser.add_subparsers>`
+  has a new ``required`` parameter that defaults to ``False``.
```
merwok: I think this focuses on the mecanism without explaining the use case.
What do you think of:

```
* :mod:`argparse` subparsers can now be made mandatory by passing ``required=True``
  to :meth:`ArgumentParser.add_subparsers() <argparse.ArgumentParser.add_subparsers>`
```

merwok:
A note about the review: these are all quite minor changes!  I could do them myself by adding commits to the branch directly, but in general I prefer to have a discussion and let the contributor own their PR, unless they are not interested in or don’t have for tiny details, in which case I’m happy to polish and merge :slightly_smiling_face:

merwok (approved):

Doc/whatsnew/3.7.rst
```diff
@@ -2401,6 +2401,10 @@ Changes in the Python API
   instead of a :class:`bytes` instance.
   (Contributed by Victor Stinner in :issue:`21071`.)
 
+* :mod:`argparse` subparsers can now be made mandatory by passing ``required=True``
+  to :meth:`ArgumentParser.add_subparsers() <argparse.ArgumentParser.add_subparsers>`.
```
merwok, line 2405: Small thing I didn’t notice before merge: in func/meth/etc roles, we don’t include the parentheses.  Sphinx has a setting to add them or not automatically to all roles (I think it’s smart enough to not add them if already present, but not 100% sure, will check when the docs rebuild is published).

(I personnally don’t like to have them, especially given Python’s clear difference of referencing a function vs. calling it, but many programmers are used to writing things like `the whatever() function` probably to make it less ambiguous in plain text contexts like email, so Python’s doc is set up to have them :slightly_smiling_face:)

I didn’t pay attention to the fact that the section here (`Changes in the Python API`) was part of `Porting to Python 3.7` (we really need docs previews for PRs!), which made sense in the original PR (changed argparse behaviour) but less so now.  I think it would have been better in `Improved Modules` → `argparse` https://docs.python.org/3.9/whatsnew/3.7.html#argparse

PR #16589, opened 2019-10-04
bpo-38368: Added fix for ctypes crash when handling arrays in structs…

vsajip:
…/unions.

https://bugs.python.org/issue38368

### DIFF
--- a/Modules/_ctypes/stgdict.c
+++ b/Modules/_ctypes/stgdict.c
@@ -644,9 +644,9 @@ PyCStructUnionType_update_stgdict(PyObject *type, PyObject *fields, int isStruct
     stgdict->align = total_align;
     stgdict->length = len;      /* ADD ffi_ofs? */
 
-#define MAX_ELEMENTS 16
+#define MAX_STRUCT_SIZE 16
 
-    if (arrays_seen && (size <= MAX_ELEMENTS)) {
+    if (arrays_seen && (size <= MAX_STRUCT_SIZE)) {
         /*
          * See bpo-22273. Arrays are normally treated as pointers, which is
          * fine when an array name is being passed as parameter, but not when
@@ -722,7 +722,7 @@ PyCStructUnionType_update_stgdict(PyObject *type, PyObject *fields, int isStruct
             if (!PyArg_ParseTuple(pair, "UO|i", &name, &desc, &bitsize)) {
                 PyErr_SetString(PyExc_TypeError,
                     "'_fields_' must be a sequence of (name, C type) pairs");
-                Py_XDECREF(pair);
+                Py_DECREF(pair);
                 return -1;
             }
             dict = PyType_stgdict(desc);
@@ -809,14 +809,21 @@ PyCStructUnionType_update_stgdict(PyObject *type, PyObject *fields, int isStruct
                 PyMem_Free(type_block);
                 return -1;
             }
+            /* In theory, we made this call in the first pass, so it *shouldn't*
+             * fail. However, you never know, and the code above might change
+             * later - keeping the check in here is a tad defensive but it
+             * will affect program size only slightly and performance hardly at
+             * all.
+             */
             if (!PyArg_ParseTuple(pair, "UO|i", &name, &desc, &bitsize)) {
                 PyErr_SetString(PyExc_TypeError,
                                 "'_fields_' must be a sequence of (name, C type) pairs");
-                Py_XDECREF(pair);
+                Py_DECREF(pair);
                 PyMem_Free(type_block);
                 return -1;
             }
             dict = PyType_stgdict(desc);
+            /* Possibly this check could be avoided, but see above comment. */
             if (dict == NULL) {
                 Py_DECREF(pair);
                 PyMem_Free(type_block);
--- a/Modules/_ctypes/stgdict.c
+++ b/Modules/_ctypes/stgdict.c
@@ -825,9 +825,9 @@ PyCStructUnionType_update_stgdict(PyObject *type, PyObject *fields, int isStruct
                              i);
                 return -1;
             }
+            assert(element_index < (ffi_ofs + len)); /* will be used below */
             if (!PyCArrayTypeObject_Check(desc)) {
                 /* Not an array. Just copy over the element ffi_type. */
-                assert(element_index < (len - ffi_ofs));
                 element_types[element_index++] = &dict->ffi_type_pointer;
             }
             else {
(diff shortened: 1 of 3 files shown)


vsajip:
Tests run on Windows x64 but not on Linux. Still investigating.

Modules/_ctypes/stgdict.c
```diff
@@ -695,14 +721,17 @@ PyCStructUnionType_update_stgdict(PyObject *type, PyObject *fields, int isStruct
...
-            else {
+            /*
+             * If it's an array, the corresponding entry in
+             * stgdict->ffi_type_pointer.elements needs to be replaced with
+             * a pointer to a dummy struct.
+             */
+            if (PyCArrayTypeObject_Check(desc)) {
                 Py_ssize_t length = dict->length;
                 StgDictObject *edict;
+                ffi_type * dummy_struct;
```
ZackerySpytz:
```suggestion
                ffi_type *dummy_struct;
```

Modules/_ctypes/stgdict.c
```diff
@@ -695,14 +721,17 @@ PyCStructUnionType_update_stgdict(PyObject *type, PyObject *fields, int isStruct
...
+                ffi_type ** dummy_fields;
```
ZackerySpytz:
```suggestion
                ffi_type **dummy_fields;
```

Modules/_ctypes/stgdict.c
```diff
@@ -712,37 +741,36 @@ PyCStructUnionType_update_stgdict(PyObject *type, PyObject *fields, int isStruct
                                  i);
                     return -1;
                 }
+
+                /* Do separate allocations for now. */
+                dummy_struct = PyMem_New(ffi_type, 1);
+                if (dummy_struct == NULL) {
+                    PyErr_NoMemory();
```
ZackerySpytz: I think there should be a `Py_DECREF(pair);` here.

Modules/_ctypes/stgdict.c
```diff
@@ -712,37 +741,36 @@ PyCStructUnionType_update_stgdict(PyObject *type, PyObject *fields, int isStruct
...
+                    return -1;
+                }
+                dummy_fields = PyMem_New(ffi_type *, length + 1);
+                if (dummy_fields == NULL) {
+                    PyErr_NoMemory();
```
ZackerySpytz: I think `dummy_struct` should be freed here. Also, `pair` should be decrefed.

vsajip: Well, the problem is that we're doing these allocations in a loop, and unless we keep some sort of linked list or dynamically allocated structure to keep track of all our allocations (both `dummy_struct` and `dummy_fields` in earlier iterations of the loop), we will leak this memory in an Out Of Memory (OOM) situation. I'm not aware that we currently do this type of recovery for these internal allocations in an OOM situation, but happy to be proved wrong.

vsajip: I will have a think about an alternative allocation strategy which allows an easier recovery, but it will be potentially a little harder to follow the logic compared to what's there now.

zooba (approved):
I'd like to have one more person go over this line-by-line as well to track the memory usage, but I didn't spot anything that seemed wrong.

vsajip:
> I'd like to have one more person go over this line-by-line as well to track the memory usage

Agreed - please request a review from a core developer who you think might have the bandwidth to do this. I asked various people to review the earlier PR (for issue 22273) but didn't get many takers ...

Modules/_ctypes/stgdict.c
```diff
@@ -681,24 +719,116 @@ PyCStructUnionType_update_stgdict(PyObject *type, PyObject *fields, int isStruct
             if (pair == NULL) {
                 return -1;
             }
+            if (!PyArg_ParseTuple(pair, "UO|i", &name, &desc, &bitsize)) {
+                PyErr_SetString(PyExc_TypeError,
+                    "'_fields_' must be a sequence of (name, C type) pairs");
+                Py_XDECREF(pair);
```
ericsnowcurrently: Why not use `Py_DECREF()`?  Don't we already know that `pair` isn't NULL?

Modules/_ctypes/stgdict.c
```diff
@@ -681,24 +719,116 @@ PyCStructUnionType_update_stgdict(PyObject *type, PyObject *fields, int isStruct
...
+            PyObject *name, *desc;
+            PyObject *pair = PySequence_GetItem(fields, i);
+            StgDictObject *dict;
+            int bitsize = 0;
+
+            if (pair == NULL) {
+                PyMem_Free(type_block);
+                return -1;
+            }
             if (!PyArg_ParseTuple(pair, "UO|i", &name, &desc, &bitsize)) {
```
ericsnowcurrently, line 818: This check (but certainly not the call) might be redundant based on the earlier for loop.

Modules/_ctypes/stgdict.c
```diff
@@ -681,24 +719,116 @@ PyCStructUnionType_update_stgdict(PyObject *type, PyObject *fields, int isStruct
...
                 PyErr_SetString(PyExc_TypeError,
                                 "'_fields_' must be a sequence of (name, C type) pairs");
                 Py_XDECREF(pair);
+                PyMem_Free(type_block);
                 return -1;
             }
             dict = PyType_stgdict(desc);
             if (dict == NULL) {
```
ericsnowcurrently, line 827: This check seems redundant based on the earlier for loop.

ericsnowcurrently (approved):
LGTM.  I've left some minor comments, but they aren't critical.

I did double-check the memory allocations and everything looks good.  The only allocations were the 2 `PySequence_GetItem()` calls and the one `PyMem_Malloc()` call.  They are cleaned up in all the right spots.

Modules/_ctypes/stgdict.c
vsajip: Although it _shouldn't_ fail, I suppose one can't guarantee it? For that reason, I'd leave the check in, it makes the code a wee bit bigger but shouldn't really impact performance.

Modules/_ctypes/stgdict.c
vsajip: As per earlier comment - would prefer to leave the check in for now.

vsajip:
Thanks for the feedback, @zooba and @ericsnowcurrently !

Modules/_ctypes/stgdict.c
vsajip: Good point, I'll change that.

vsajip:
I'll wait for the buildbots to pronounce on this merge before backporting to 3.8/3.7.

doko42:
I checked that at least two packages in Debian and Ubuntu, python-ipaddr and python-pysnmp4 now work again with this patch (only tested on the 3.7 branch).

vsajip:
@doko42 Thanks for the feedback.

PR #16594, opened 2019-10-05
bpo-38376: Fix build error when enabling assertions but not Py_DEBUG.

Yhg1s:
_PyUnicode_CheckConsistency is now defined even without Py_DEBUG set, so get rid of the #define hack that was there to deal with calls in assert statements not explicitly guarded by Py_DEBUG.

https://bugs.python.org/issue38376

### DIFF
--- a/Include/unicodeobject.h
+++ b/Include/unicodeobject.h
@@ -1032,14 +1032,10 @@ PyAPI_FUNC(int) PyUnicode_IsIdentifier(PyObject *s);
 
 /* === Characters Type APIs =============================================== */
 
-#if defined(Py_DEBUG) && !defined(Py_LIMITED_API)
+#if (defined(Py_DEBUG) || !defined(NDEBUG)) && !defined(Py_LIMITED_API)
 PyAPI_FUNC(int) _PyUnicode_CheckConsistency(
     PyObject *op,
     int check_content);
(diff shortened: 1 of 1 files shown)


vsajip (approved):

vstinner (commented):
My intent was to also provide _PyObject_CheckConsistency() in release mode to help debugging GC crash, especially in visit_decref(). I wrote PR #16612 to fix my implementation.

vstinner:
In short, _PyUnicode_CheckConsistency() is now always available in Python 3.8. I wrote PR #16623 to fix its definition in header files.

This change doesn't look correct. I'm not sure that _PyUnicode_CheckConsistency() definition depends on NDEBUG macro.

vstinner:
I merged my PR #16623 instead.

Yhg1s:
FYI, this change was intended to keep visibility parity, because the existing *declaration* (not definition) of _PyUnicode_CheckConsistency in Include/unicodeobject.h was already conditional on NDEBUG or Py_DEBUG. The only other declaration was in Include/internal/pycore_object.h, which wouldn't be visible to user code in the same way as before. Moving the declaration to Include/cpython/unicodeobject.h as you did also solved that same problem, because that header is included by Include/unicodeobject.h when not using the limited API. We should be all good now.

PR #16595, opened 2019-10-05
Ignore .pytest_cache/.

brandtbucher:
Keeps my plugin from sneaking JSON files into commits.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -33,6 +33,7 @@ TAGS
 gmon.out
 .coverage
 .mypy_cache/
+.pytest_cache/
 
 *.exe
 !Lib/distutils/command/*.exe


brandtbucher:
Not sure how/if this should be backported.

ammaraskar:
Sorry if I'm missing something but which plugin are you referring to? Was it added to the workflow recently?

AFAIK we don't use or endorse pytest in cpython development.

brandtbucher:
@ammaraskar The Python plugin for Visual Studio Code generates these caches when the project is opened. I figured that since caches for other “unofficial” tools like `.mypy_cache/` and `.vscode/` are included here, it could be good to add.

Not a huge deal, but could be helpful to others.

ammaraskar:
Sounds good 👍 

Maybe group up the vscode related ones and put a comment there just so people know why?

brandtbucher:
Well, this is a pytest cache, not a VSCode cache. The plugin just happens to trigger it (`.mypy_cache/` is the same).

taleinat:
The Python plugin creates `.pytest_cache` directories for projects not using pytest? Sounds like a bug in the plugin! (And if it's this plugin that caused the addition of `.mypy_cache` to our `.gitignore`, maybe that should be fixed too...)

merwok:
mypy cache is an older manual addition: https://github.com/python/cpython/pull/6614

Is is possible to run (some) CPython tests with pytest?  If not, an IDE plugin dumping things to all your projects uncontrollably does not seem good.

brandtbucher:
It is possible. In fact, I've used pytest in the past to get nice iterative output, with something like `pytest Lib/test/test_glob.py --looponfail`.

Again, this is just a suggestion. I can always add it to my personal ignore file, but I figured if it was useful for me, it might be useful for others!

taleinat (approved):
Looking at our `.gitignore`, it's obvious the current policy is to add everything that is common enough to be useful to many people. So if someone wants this, and there's surely no possible harm, I don't see why not :)

taleinat:
> Not sure how/if this should be backported.

Since the vast majority of development is done based on the master branch, I don't think it is worth backporting this.

brandtbucher:
Thanks @taleinat!

PR #16600, opened 2019-10-05
bpo-38364: unwrap partialmethods just like we unwrap partials

mjpieters:
The inspect.isgeneratorfunction, inspect.iscoroutinefunction and inspect.isasyncgenfunction already unwrap functools.partial objects, this patch adds support for partialmethod objects as well.

https://bugs.python.org/issue38364


Lib/inspect.py
```diff
@@ -170,8 +170,10 @@ def isfunction(object):
 
 def _has_code_flag(f, flag):
     """Return true if ``f`` is a function (or a method or functools.partial
-    wrapper wrapping a function) whose code object has the given ``flag``
+    wrapper wrapping a function or a functools.partialmethod wrapping a
+    function) whose code object has the given ``flag``
```
taleinat, line 387: Shouldn't this be "or a functools.partialmethod wrapping a method"?

Lib/test/test_inspect.py
```diff
@@ -167,12 +167,33 @@ def test_iscoroutine(self):
         gen_coro = gen_coroutine_function_example(1)
         coro = coroutine_function_example(1)
 
+        class PMClass:
```
taleinat: How about a test case for a partialmethod wrapping a normal method?

florimondmanca:
Coincidently, this would also resolve situations when passing the `partial` of a method directly…

```python
import functools, inspect

class Dog:
    async def abark(self): pass

inspect.iscoroutinefunction(dog.abark)  # True
inspect.iscoroutinefunction(functools.partial(dog.abark))  # False
```

This would be because the partial would be unwrapped (via the inner call to `_unwrap_partial()` here), yielding a method object that will be correctly handled by the method unwrapping.

Lib/inspect.py
mjpieters: What matters is the function underneath. The partialmethod object can wrap any descriptor type, but its the function that has the flags.

Lib/test/test_inspect.py
mjpieters: Testing how inspect interacts with `partialmethod()` on regular methods is [covered elsewhere already](https://github.com/python/cpython/blob/9a669d58e8cb586fba38c84d5b631cd8a95d0c0c/Lib/test/test_inspect.py#L859-L871); these are the coroutine-specific tests.

jalaziz:
This is also an issue for unittest mocking. `patch` will return `MagicMock` for partial async methods and not `AsyncMock`.

kumaraditya303 (changes requested):
There are merge conflicts.

mjpieters:

Doc/library/inspect.rst
```diff
@@ -354,6 +357,10 @@ attributes (see :ref:`import-mod-attrs` for module attributes):
       Functions wrapped in :func:`functools.partial` now return ``True`` if the
       wrapped function is a :term:`coroutine function`.
 
+   .. versionchanged:: 3.12
+      Functions wrapped in :func:`functools.partialpartial` now return ``True``
```
kumaraditya303:
```suggestion
      Functions wrapped in :func:`functools.partialmethod` now return ``True``
```

Doc/library/inspect.rst
```diff
@@ -418,6 +425,9 @@ attributes (see :ref:`import-mod-attrs` for module attributes):
       Functions wrapped in :func:`functools.partial` now return ``True`` if the
       wrapped function is a :term:`asynchronous generator` function.
 
+   .. versionchanged:: 3.12
+      Functions wrapped in :func:`functools.partialpartial` now return ``True``
```
kumaraditya303:
```suggestion
      Functions wrapped in :func:`functools.partialmethod` now return ``True``
```

Lib/functools.py
```diff
@@ -423,6 +423,17 @@ def _unwrap_partial(func):
         func = func.func
     return func
 
+def _unwrap_partialmethod(func):
+    prev = None
+    while func is not prev:
+        prev = func
+        while isinstance(getattr(func, "_partialmethod", None), partialmethod):
```
kumaraditya303: Why getattr? What objects have `_partialmethod`?

kumaraditya303 (dismissed):

Doc/library/inspect.rst
mjpieters: 🤦‍♂️ Good catch, that's a really dumb merge error on my part!

Lib/functools.py
mjpieters: `partialmethod(...).__get__()` can return a closure with `_partialmethod` pointing to the instance. This happens if the wrapped callable is not a descriptor, or if the descriptor returned itself again.

See the [`partialmethod._make_unbound_method()` implementation](https://github.com/python/cpython/blob/98308dbeb110198ebe28bdb7720d3671b3e7f57b/Lib/functools.py#L385-L391).

`getattr()` is used because the attribute is optional, and you'd not want this to break even if the supplied object does have a `_partialmethod` attribute that is not actually a `partialmethod` instance.

mjpieters:

encukou:
Hi! Just getting to this, sorry for the delay.
Since we're accessing `_partialmethod` on arbitrary objects, it would be best to rename it to `__partialmethod__` -- make it part of the namespace reserved for Python.

I hope you don't mind me pushing directly to the PR.

encukou (approved):

PR #16607, opened 2019-10-06
bpo-38387: Formally document PyDoc_STRVAR and PyDoc_STR macros

bsolomon1124:
Adds a short description of `PyDoc_STRVAR` and `PyDoc_STR` to "Useful macros" section of C-API docs.

Currently, there is [one lone mention](https://docs.python.org/3/c-api/module.html?highlight=pydoc_strvar#c.PyModuleDef) in the C-API reference, despite the fact that `PyDoc_STRVAR` is ubiquitous to `Modules/`.

Additionally, this properly uses `c:macro` within `c-api/module.html` to link.

https://bugs.python.org/issue38387

### DIFF
--- a/Doc/c-api/intro.rst
+++ b/Doc/c-api/intro.rst
@@ -174,6 +174,17 @@ complete listing.
    .. versionchanged:: 3.8
       MSVC support was added.
 
+.. c:macro:: PyDoc_STRVAR(name, str)
+
+   A shortcut for ``static const char name[] = str``, commonly used
+   to initialize a module, function, class, or method docstring.
+
+   Example::
+
+      PyDoc_STRVAR(module_doc, "This is the module's docstring.");
+
+   .. versionchanged:: 3.8
+      This macro previously inserted ``static char name[]``.
 
 .. _api-objects:
 
--- a/Doc/c-api/intro.rst
+++ b/Doc/c-api/intro.rst
@@ -177,10 +177,10 @@ complete listing.
 .. c:macro:: PyDoc_STRVAR(name, str)
 
    Creates a variable with name ``name`` that can be used in docstrings.
+   If Python is built without docstrings, the value will be empty.
 
-   Use the :c:macro:`PyDoc_STRVAR` or :c:macro:`PyDoc_STR` macros for
-   docstrings to support building Python without docstrings,
-   as specified in :pep:`7`.
+   Use :c:macro:`PyDoc_STRVAR` for docstrings to support building
+   Python without docstrings, as specified in :pep:`7`.
 
    Example::
 
@@ -191,6 +191,22 @@ complete listing.
           {"pop", (PyCFunction)deque_pop, METH_NOARGS, pop_doc},
           // ...
       }
+
+.. c:macro:: PyDoc_STR(str)
+
+   Creates a docstring for the given input string or an empty string
+   if docstrings are disabled.
+
+   Use :c:macro:`PyDoc_STR` in specifying docstrings to support
+   building Python without docstrings, as specified in :pep:`7`.
+
+   Example::
+
+      static PyMethodDef pysqlite_row_methods[] = {
+          {"keys", (PyCFunction)pysqlite_row_keys, METH_NOARGS,
+              PyDoc_STR("Returns the keys of the row.")},
+          {NULL, NULL}
+      };
 
 .. _api-objects:
 
--- a/Doc/c-api/intro.rst
+++ b/Doc/c-api/intro.rst
@@ -176,15 +176,21 @@ complete listing.
 
 .. c:macro:: PyDoc_STRVAR(name, str)
 
-   A shortcut for ``static const char name[] = str``, commonly used
-   to initialize a module, function, class, or method docstring.
+   Creates a variable with name ``name`` that can be used in docstrings.
+
+   Use the :c:macro:`PyDoc_STRVAR` or :c:macro:`PyDoc_STR` macros for
+   docstrings to support building Python without docstrings,
+   as specified in :pep:`7`.
 
    Example::
 
-      PyDoc_STRVAR(module_doc, "This is the module's docstring.");
+      PyDoc_STRVAR(pop_doc, "Remove and return the rightmost element.");
 
-   .. versionchanged:: 3.8
-      This macro previously inserted ``static char name[]``.
+      static PyMethodDef deque_methods[] = {
+          // ...
+          {"pop", (PyCFunction)deque_pop, METH_NOARGS, pop_doc},
+          // ...
+      }
 
 .. _api-objects:
 
--- a/Misc/NEWS.d/next/Documentation/2019-10-06-23-44-15.bpo-38387.fZoq0S.rst
+++ b/Misc/NEWS.d/next/Documentation/2019-10-06-23-44-15.bpo-38387.fZoq0S.rst
@@ -1,1 +1,1 @@
-Document PyDoc_STRVAR macro in the C-API reference.
+Document :c:macro:`PyDoc_STRVAR` macro in the C-API reference.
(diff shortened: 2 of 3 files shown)


Doc/c-api/intro.rst
```diff
@@ -174,6 +174,17 @@ complete listing.
    .. versionchanged:: 3.8
       MSVC support was added.
 
+.. c:macro:: PyDoc_STRVAR(name, str)
+
+   A shortcut for ``static const char name[] = str``, commonly used
```
ammaraskar: I don't think mentioning what this does internally is very useful, makes it somewhat contractual and could get out of date.

Maybe reword it just to be something along the lines of:

> Creates a variable with name `name` that can be used in module, function, classes etc docstrings.

Doc/c-api/intro.rst
```diff
@@ -174,6 +174,17 @@ complete listing.
...
+   to initialize a module, function, class, or method docstring.
+
+   Example::
+
+      PyDoc_STRVAR(module_doc, "This is the module's docstring.");
```
ammaraskar: I think expanding this to:

```c
PyDoc_STRVAR(method_doc, "This is the method's docstring.");
// ...
    {"my_method", my_method, METH_O, method_doc},
```

would be useful to actually show a place where the macro-generated variable is used.

Misc/NEWS.d/next/Documentation/2019-10-06-23-44-15.bpo-38387.fZoq0S.rst
```diff
@@ -0,0 +1 @@
+Document PyDoc_STRVAR macro in the C-API reference.
```
ammaraskar: Please add in a `Patch by <your preferred name>` like this https://github.com/python/cpython/blob/master/Misc/NEWS.d/next/Core%20and%20Builtins/2019-05-28-11-47-44.bpo-37077.S1h0Fc.rst

ammaraskar (changes requested):

Misc/NEWS.d/next/Documentation/2019-10-06-23-44-15.bpo-38387.fZoq0S.rst
bsolomon1124: Thanks for reviewing and hello from Stack Overflow! 👋

Doc/c-api/intro.rst
```diff
@@ -174,6 +174,17 @@ complete listing.
...
+
+   .. versionchanged:: 3.8
+      This macro previously inserted ``static char name[]``.
```
ghost: IMO it's not worthy mentioning this `versionchanged`.

Misc/NEWS.d/next/Documentation/2019-10-06-23-44-15.bpo-38387.fZoq0S.rst
aeros: @bsolomon1124 The `Patch by <name>` is optional, but you can certainly add it if you'd like to.

Doc/c-api/intro.rst
aeros: Agreed, that's also why we typically don't list every possible use case; to avoid creating unnecessary contractual obligations. It adds to the number of potential backwards compatibility concerns.

aeros167:
@bsolomon1124 Thanks for the PR, I think it's certainly worthwhile to document the macro, especially considering how common it's utilized.

But, I'm in agreement with the changes recommended by @ammaraskar: removing the description of the internals and adding a brief example.

bsolomon1124:
All the suggested changes here make sense and I'll add them when I have a chance.

I also think it might be worth mentioning the stated use of these macros:

> Use the PyDoc_STR() or PyDoc_STRVAR() macro for docstrings to support building Python without docstrings (./configure --without-doc-strings).

https://www.python.org/dev/peps/pep-0007/#documentation-strings

aeros167:
I think that might be worth adding, but I would recommend linking to the PEP itself using an inline Sphinx role:
```
Use the ``PyDoc_STR()`` or ``PyDoc_STRVAR()`` macro for docstrings to support
building Python without docstrings, as specified in :pep:`7`.
```
This provides readers a location for further information and means that we don't have to update this section if the syntax ever changes for ``./configure --without-doc-string``.

Doc/c-api/intro.rst
bsolomon1124: I used a method docstring directly from `_collectionsmodule.c`

Doc/c-api/intro.rst
```diff
@@ -174,6 +174,23 @@ complete listing.
    .. versionchanged:: 3.8
       MSVC support was added.
 
+.. c:macro:: PyDoc_STRVAR(name, str)
+
+   Creates a variable with name ``name`` that can be used in docstrings.
```
aeros, line 179: Formatting nit, parameter names should be *italicized* instead of in ``codeblocks``. Also, I think "called name" reads a bit better than "with name name".
```suggestion
   Creates a variable called *name* that can be used in docstrings.
```

aeros (approved):

Doc/c-api/intro.rst
```diff
@@ -174,6 +174,23 @@ complete listing.
    .. versionchanged:: 3.8
       MSVC support was added.
 
+.. c:macro:: PyDoc_STRVAR(name, str)
+
+   Creates a variable called *name* that can be used in docstrings.
+
+   Use the :c:macro:`PyDoc_STRVAR` or :c:macro:`PyDoc_STR` macros for
```
ammaraskar: Since you mentioned `PyDoc_STR` in here, let's add documentation for it as well. Just something simple like, `Returns a docstring for the given input string`

Then the note about pep7 and example can be combined, maybe expanding the example with:
```c
...
    {"append", (PyCFunction)deque_append. METH_NOARGS, PyDoc_STR("some documentation")
...
```

bsolomon1124:
@ammaraskar I've added a section for `PyDoc_STR` also here.  Appreciate if you have time for another review.

Screen Shot 2020-01-26 at 7 05 35 PM

Doc/c-api/intro.rst
```diff
@@ -174,6 +174,39 @@ complete listing.
...
+
+   Creates a docstring for the given input string or an empty string
+   if docstrings are disabled.
+
+   Use :c:macro:`PyDoc_STR` in specifying docstrings to support
+   building Python without docstrings, as specified in :pep:`7`.
+
+   Example::
+
+      static PyMethodDef pysqlite_row_methods[] = {
```
aeros, line 205: Can you modify this example so that it doesn't reference any third party modules (`pysqlite` in this case)? We typically avoid doing that in the documentation unless it's specifically relevant, especially for code examples.

aeros:
While I think it's probably okay in this case, I'd generally recommend against including additional sections in a PR that were outside of the scope of the original issue that it's attached to. The scope of this issue was specifically adding documentation for `PyDoc_STRVAR`.

Going outside of the scope can substantially to the required review time, particularly if you already have received several reviews on the PR. It may not seem like much in this case, but it can really add up when you consider the total volume of open PRs and our limited review time (it's mostly volunteer-driven). 

An exception to the above is if you specifically ask a core developer about it and they approve of including additional related changes in the PR. There's also the option of simply opening a second PR and attaching it to the same issue; assuming they're both related.

Doc/c-api/intro.rst
bsolomon1124: @aeros can you clarify what you mean by 3rd-party modules?  This example is taken from CPython at https://github.com/python/cpython/blob/f669581a9527afb0d2325f9845a86715c0ba365d/Modules/_sqlite/row.c#L223.  Do you mean that we should not include any actual examples from CPython?  In that case we would want to change the `deque_methods` in this PR also.

aeros: > can you clarify what you mean by 3rd-party modules? This example is taken from CPython

Never mind, you can disregard the above comment. There's a package on PyPI titled `pysqlite` which I thought this was in reference to, I hadn't realized this was from a function in a standard library module (which is perfectly okay to use as a reference in the docs). Thanks for clearing that up.

zware (approved):

PR #16611, opened 2019-10-07
bpo-38210: Fix compiler warning in dictobject.c

vstinner:
The "tmp" variable is no longer used.

https://bugs.python.org/issue38210


vstinner:
cc @corona10 @serhiy-storchaka

corona10:
Thanks for the fix

PR #16612, opened 2019-10-07
bpo-36389: _PyObject_CheckConsistency() now also works in release mode

vstinner:
[bpo-36389](https://bugs.python.org/issue36389), [bpo-38376](https://bugs.python.org/issue38376): Modify the following functions to also work in
release mode:

* _PyDict_CheckConsistency()
* _PyObject_CheckConsistency()
* _PyType_CheckConsistency()
* _PyUnicode_CheckConsistency()

The main function _PyObject_CheckConsistency() can be used for
debugging purpose, like crash in visit_decref() of the GC.

_PyBytesWriter_CheckConsistency() now returns 1 and is only used with
assert().

### DIFF
--- a/Objects/dictobject.c
+++ b/Objects/dictobject.c
@@ -459,23 +459,26 @@ static PyObject *empty_values[1] = { NULL };
 int
 _PyDict_CheckConsistency(PyObject *op, int check_content)
 {
-#ifndef NDEBUG
-    _PyObject_ASSERT(op, PyDict_Check(op));
+#define CHECK(expr) \
+    do { if (!(expr)) { _PyObject_ASSERT_FAILED_MSG(op, Py_STRINGIFY(expr)); } } while (0)
+
+    assert(op != NULL);
+    CHECK(PyDict_Check(op));
     PyDictObject *mp = (PyDictObject *)op;
 
     PyDictKeysObject *keys = mp->ma_keys;
     int splitted = _PyDict_HasSplitTable(mp);
     Py_ssize_t usable = USABLE_FRACTION(keys->dk_size);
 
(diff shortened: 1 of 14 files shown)


vstinner:
I would like to push this change into 3.8 since https://bugs.python.org/issue36389 was designed for Python 3.8.

Objects/dictobject.c
```diff
@@ -459,23 +459,26 @@ static PyObject *empty_values[1] = { NULL };
 int
 _PyDict_CheckConsistency(PyObject *op, int check_content)
 {
-#ifndef NDEBUG
-    _PyObject_ASSERT(op, PyDict_Check(op));
+#define ASSERT(expr) \
```
Yhg1s: I'm not particularly fond of 'assert' and 'ASSERT' doing different things, especially since ASSERT does not use assert and does not do nothing in NDEBUG mode. How about 'CHECK' instead?

Yhg1s:
Roping in the 3.8 release manager for the question of whether this should go into 3.8.0.

vstinner:
This change mostly impacts the debugging experience. It doesn't change Python in release mode: modified functions only becomes available in release mode, but are never called.

@Yhg1s: Thanks for your review. I agree that "define ASSERT" is confusing. I renamed my macros to "CHECK" as you suggested.

I also included other fixes for these debug functions.

vstinner:
> @vstinner vstinner removed the needs backport to 3.8 label now 

I wrote a minimum fix for 3.8: PR #16623.

I may backport this change to 3.8, but only after the final 3.8.0 release. We can discuss that later ;-)

PR #16614, opened 2019-10-07
bpo-38353: Simplify calculate_pybuilddir()

vstinner:
Calling _Py_wfopen() is enough to check if filename is an existing
file or not. There is no need to check first isfile().

### DIFF
--- a/Modules/getpath.c
+++ b/Modules/getpath.c
@@ -657,11 +657,6 @@ calculate_pybuilddir(const wchar_t *argv0_path,
         return _PyStatus_NO_MEMORY();
     }
 
-    if (!isfile(filename)) {
-        PyMem_RawFree(filename);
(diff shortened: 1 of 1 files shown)


vstinner:
The pybuilddir.txt configuration file was introduced by https://bugs.python.org/issue9589 with commit e9b428f9977f8733e6b0d2c321c093779f95080f:

```
commit e9b428f9977f8733e6b0d2c321c093779f95080f
Author: Antoine Pitrou <solipsis@pitrou.net>
Date:   Fri Aug 13 22:25:01 2010 +0000

    Reimplement addbuilddir() in C inside getpath.c, so as to execute it
    at interpreter startup before importing any non-builtin modules.
    Should fix #9589.
```

The code evolved a lot since 2010 which made the isfile() call redundant.

cc @pitrou

PR #340, opened 2017-02-27
bpo-32888: Improve exception message in ast.literal_eval

Rosuav:
When a non-literal is given to literal_eval, attempt to be more
helpful with the message, rather than calling it 'malformed'.

https://bugs.python.org/issue32888

PR #16620, opened 2019-10-07
bpo-38396: dump(node) instead of repr(node) on ast.literal_eval

isidentical:
https://bugs.python.org/issue38396

### DIFF
--- a/Lib/ast.py
+++ b/Lib/ast.py
@@ -63,7 +63,7 @@ def literal_eval(node_or_string):
         if isinstance(node, Constant):
             if type(node.value) in (int, float, complex):
                 return node.value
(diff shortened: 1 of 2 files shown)


vstinner (commented):
Can you please try to write a test?

isidentical:
bpo-38396 resolved as duplicate of bpo-32888 and #340 is about fixing it. It has a slightly different approach but resolves the same problem so I think this PR can be closed.

vstinner:
Oh, @ilevkivskyi seems to be against this PR, so I close it: https://bugs.python.org/issue38396#msg354132 I agree that PR #340 should produce shorter and better error message.

PR #16623, opened 2019-10-07
[3.8] bpo-38376: Fix _PyUnicode_CheckConsistency() definition

vstinner:
Always define _PyUnicode_CheckConsistency() in the CPython C API.

https://bugs.python.org/issue38376


vstinner:
I tested manually that Python build correctly with the following commands:

```
./configure --prefix=$HOME/.local --with-assertions
make
```

PR #16629, opened 2019-10-07
bpo-36389: Add newline to _PyObject_AssertFailed()

vstinner:
Add a newline between the verbose object dump and the Py_FatalError()
logs for readability.


vstinner:
Example: https://github.com/python/cpython/pull/16615#issuecomment-539203517

```
(...)
object address  : 0x7feb39ce19b0
object refcount : 1
object type     : 0x7f5ea0
object type name: hamt
object repr     : <hamt object at 0x7feb39ce19b0>
```

becomes with this change:


```
(...)
object address  : 0x7feb39ce19b0
object refcount : 1
object type     : 0x7f5ea0
object type name: hamt
object repr     : <hamt object at 0x7feb39ce19b0>

```

PR #16630, opened 2019-10-07
bpo-38400 Don't check for NULL linked list pointers in _PyObject_IsFreed

pablogsal:
https://bugs.python.org/issue38400


vstinner (approved):
LGTM.

PR #16631, opened 2019-10-07
bpo-38070: Enhance visit_decref() debug trace

vstinner:
subtract_refs() now pass the parent object to visit_decref() which
pass it to _PyObject_ASSERT(). So if the "is freed" assertion fails,
the parent is used in debug trace, rather than the freed object. The
parent object is more likely to contain useful information. Freed
objects cannot be inspected are are displayed as "<object at xxx is
freed>" with no other detail.

### DIFF
--- a/Modules/gcmodule.c
+++ b/Modules/gcmodule.c
@@ -373,9 +373,9 @@ update_refs(PyGC_Head *containers)
 
 /* A traversal callback for subtract_refs. */
 static int
-visit_decref(PyObject *op, void *data)
+visit_decref(PyObject *op, void *parent)
 {
-    _PyObject_ASSERT(op, !_PyObject_IsFreed(op));
+    _PyObject_ASSERT(_PyObject_CAST(parent), !_PyObject_IsFreed(op));
 
     if (PyObject_IS_GC(op)) {
         PyGC_Head *gc = AS_GC(op);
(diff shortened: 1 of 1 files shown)


vstinner:
Example with https://bugs.python.org/issue38400 bug.

Without the change, Python exit with:

```
Modules/gcmodule.c:378: visit_decref: Assertion "!_PyObject_IsFreed(op)" failed
<object at 0x7420a0 is freed>
...
```

With this change, the output is more complete:

```
Modules/gcmodule.c:378: visit_decref: Assertion "!_PyObject_IsFreed(op)" failed

object address  : 0x7ff11b158b40
object refcount : 1
object type     : 0x740be0
object type name: dict
object repr     : {'__repr__': <slot wrapper '__repr__' of 'weakref' objects>, '__hash__': <slot wrapper '__hash__' of 'weakref' objects>, '__call__': <slot wrapper '__call__' of 'weakref' objects>, '__lt__': <slot wrapper '__lt__' of 'weakref' objects>, '__le__': <slot wrapper '__le__' of 'weakref' objects>, '__eq__': <slot wrapper '__eq__' of 'weakref' objects>, '__ne__': <slot wrapper '__ne__' of 'weakref' objects>, '__gt__': <slot wrapper '__gt__' of 'weakref' objects>, '__ge__': <slot wrapper '__ge__' of 'weakref' objects>, '__init__': <slot wrapper '__init__' of 'weakref' objects>, '__new__': <built-in method __new__ of type object at 0x749500>, '__callback__': <member '__callback__' of 'weakref' objects>, '__doc__': None}

...
```

In this case, the bug was that None was seen as freed: the None object can be easily be recovered using a debugger in the visit_decref() frame.

PR #16632, opened 2019-10-08
bpo-38395: Fix ownership in weakref.proxy methods

pablogsal:
The implementation of weakref.proxy's methods call back into the Python
API using a borrowed references of the weakly referenced object
(acquired via PyWeakref_GET_OBJECT). This API call may delete the last
reference to the object (either directly or via GC), leaving a dangling
pointer, which can be subsequently dereferenced.

To fix this, claim temporarily ownership of the referenced object when
calling the appropiate method. Some functions because at the moment they
do not need to access the borrowed referent, but to protect against
future changes to these functions, ownership needs to be fixed in
all potentially affected methids.

https://bugs.python.org/issue38395

### DIFF
--- a/Lib/test/test_weakref.py
+++ b/Lib/test/test_weakref.py
@@ -391,6 +391,23 @@ class ReferencesTestCase(TestBase):
         lyst = List()
         self.assertEqual(bool(weakref.proxy(lyst)), bool(lyst))
 
+    def test_proxy_iter(self):
+        # Test fails with a debug build of the interpreter
+        # (see bpo-38395).
+
+        obj = None
+
+        class MyObj:
+            def __iter__(self):
+                nonlocal obj
+                del obj
+                return NotImplemented
+
+        obj = MyObj()
+        p = weakref.proxy(obj)
+        with self.assertRaises(TypeError):
+            "blech" in p
+
     def test_getweakrefcount(self):
         o = C()
         ref1 = weakref.ref(o)
--- a/Lib/test/test_weakref.py
+++ b/Lib/test/test_weakref.py
@@ -406,6 +406,9 @@ class ReferencesTestCase(TestBase):
         obj = MyObj()
         p = weakref.proxy(obj)
         with self.assertRaises(TypeError):
+            # "blech" in p calls MyObj.__iter__ through the proxy,
+            # without keeping a reference to the real object, so it
+            # can be killed in the middle of the call
             "blech" in p
 
     def test_getweakrefcount(self):
--- a/Objects/weakrefobject.c
+++ b/Objects/weakrefobject.c
@@ -145,11 +145,11 @@ weakref_hash(PyWeakReference *self)
 {
     if (self->hash != -1)
         return self->hash;
-    if (PyWeakref_GET_OBJECT(self) == Py_None) {
+    PyObject* obj = PyWeakref_GET_OBJECT(self);
+    if (obj == Py_None) {
         PyErr_SetString(PyExc_TypeError, "weak object has gone away");
         return -1;
     }
-    PyObject* obj = PyWeakref_GET_OBJECT(self);
     Py_INCREF(obj);
     self->hash = PyObject_Hash(obj);
     Py_DECREF(obj);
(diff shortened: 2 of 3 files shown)


pitrou (approved):
LGTM, thank you

Objects/weakrefobject.c
```diff
@@ -564,21 +572,32 @@ proxy_length(PyWeakReference *proxy)
...
+    int res = 0;
     if (!proxy_checkref(proxy))
         return -1;
 
-    if (value == NULL)
-        return PyObject_DelItem(PyWeakref_GET_OBJECT(proxy), key);
-    else
-        return PyObject_SetItem(PyWeakref_GET_OBJECT(proxy), key, value);
+    PyObject *obj = PyWeakref_GET_OBJECT(proxy);
+    Py_INCREF(obj);
```
vstinner, line 621: nitpick: C99 now allows to move "int res" closer to its usage, like here :-)

Lib/test/test_weakref.py
```diff
@@ -391,6 +391,23 @@ class List(list): pass
...
+        class MyObj:
+            def __iter__(self):
+                nonlocal obj
+                del obj
+                return NotImplemented
+
+        obj = MyObj()
+        p = weakref.proxy(obj)
+        with self.assertRaises(TypeError):
+            "blech" in p
```
vstinner, line 412: The relationship between "a in b" and `__iter__()` is nonobvious. Are you also able to reproduce the crash using something like:

```
# call indirectly __iter__()
for item in p: pass
```

Or maybe just add a comment to explain that in this case, "a in b" calls `__iter__()`.

vstinner (commented):
weakref_hash() doesn't seem to hold a strong reference to call PyObject_Hash() which may execute arbitrary Python code, no?

Same remark for weakref_repr(), weakref_richcompare(), proxy_bool().

Lib/test/test_weakref.py
pablogsal: > Are you also able to reproduce the crash using something like

No, because that keeps a reference to the object in the iterator, not triggering the problem. Will add a comment to explain that

pablogsal:
@vstinner Done, can you review again?

Objects/weakrefobject.c
```diff
@@ -149,7 +149,10 @@ weakref_hash(PyWeakReference *self)
         PyErr_SetString(PyExc_TypeError, "weak object has gone away");
         return -1;
     }
-    self->hash = PyObject_Hash(PyWeakref_GET_OBJECT(self));
+    PyObject* obj = PyWeakref_GET_OBJECT(self);
```
vstinner: Maybe replace PyWeakref_GET_OBJECT(self) with obj in the previous test.

Objects/weakrefobject.c
```diff
@@ -163,24 +166,28 @@ weakref_repr(PyWeakReference *self)
     if (PyWeakref_GET_OBJECT(self) == Py_None)
         return PyUnicode_FromFormat("<weakref at %p; dead>", self);
 
-    if (_PyObject_LookupAttrId(PyWeakref_GET_OBJECT(self), &PyId___name__, &name) < 0) {
+    PyObject* obj = PyWeakref_GET_OBJECT(self);
```
vstinner: Same comment.

vstinner (commented):
Hum, WRAP_UNARY(), WRAP_BINARY(), WRAP_TERNARY() and WRAP_METHOD() still have the same bug, no?

By the way, UNWRAP_I() is no longer used and so can be removed.

Objects/weakrefobject.c
pablogsal: @vstinner Done in d44d9e3

Objects/weakrefobject.c
pablogsal: @vstinner Done in d44d9e3

vstinner (approved):
LGTM. Goodbye borrowed references!

vstinner:
> @pablogsal pablogsal removed the needs backport to 3.6 label 1 hour ago 

I don't think that this change can be qualified as a fix for a security issue, and 3.6 no longer accept bugfixes. So yeah, no need to backport to 3.6 :-)

PR #16640, opened 2019-10-08
Docs: Add asyncio source code links

aeros:
Many of the other stdlib modules link to the source code in the documentation pages. Asyncio currently does not. 

This PR adds source code links to the documentation pages for asyncio. It also fixes the location of a reference point in `asyncio-eventloop.rst` (which fixes a link in `asyncio.rst`), adds a Sphinx term (link to glossary) for "coroutine" in `asyncio-task.rst`, and fixes a minor grammar typo.

### DIFF
--- a/Doc/library/asyncio-eventloop.rst
+++ b/Doc/library/asyncio-eventloop.rst
@@ -1,10 +1,15 @@
 .. currentmodule:: asyncio
 
+.. _asyncio-event-loop:
 
 ==========
 Event Loop
 ==========
 
+**Source code:** :source:`Lib/asyncio/events.py`,
+:source:`Lib/asyncio/base_events.py`
+
+------------------------------------
 
 .. rubric:: Preface
 
@@ -80,8 +85,6 @@ This documentation page contains the following sections:
 * The `Examples`_ section showcases how to work with some event
   loop APIs.
 
-
-.. _asyncio-event-loop:
 
 Event Loop Methods
 ==================
--- a/Doc/library/asyncio-eventloop.rst
+++ b/Doc/library/asyncio-eventloop.rst
@@ -1,6 +1,5 @@
 .. currentmodule:: asyncio
 
-.. _asyncio-event-loop:
 
 ==========
 Event Loop
@@ -85,6 +84,8 @@ This documentation page contains the following sections:
 * The `Examples`_ section showcases how to work with some event
   loop APIs.
 
+
+.. _asyncio-event-loop:
 
 Event Loop Methods
 ==================
--- a/Doc/library/asyncio-task.rst
+++ b/Doc/library/asyncio-task.rst
@@ -210,10 +210,6 @@ is :meth:`loop.run_in_executor`.
 Running an asyncio Program
 ==========================
 
-**Source code:** :source:`Lib/asyncio/runners.py`
-
--------------------------------------------------
-
 .. function:: run(coro, \*, debug=False)
 
     Execute the :term:`coroutine` *coro* and return the result.
@@ -243,6 +239,10 @@ Running an asyncio Program
 
     .. versionchanged:: 3.9
        Updated to use :meth:`loop.shutdown_default_executor`.
+
+.. note::
+   The source code for ``asyncio.run()`` can be found in
+   :source:`Lib/asyncio/runners.py`.
 
 Creating Tasks
 ==============
--- a/Doc/library/asyncio-task.rst
+++ b/Doc/library/asyncio-task.rst
@@ -247,11 +247,6 @@ Running an asyncio Program
 Creating Tasks
 ==============
 
-**Source code:** :source:`Lib/asyncio/tasks.py`,
-:source:`Lib/asyncio/base_tasks.py`
-
------------------------------------
-
 .. function:: create_task(coro, \*, name=None)
 
    Wrap the *coro* :ref:`coroutine <coroutine>` into a :class:`Task`
--- a/Doc/library/asyncio-task.rst
+++ b/Doc/library/asyncio-task.rst
@@ -18,9 +18,9 @@ and Tasks.
 Coroutines
 ==========
 
-Coroutines declared with async/await syntax is the preferred way of
-writing asyncio applications.  For example, the following snippet
-of code (requires Python 3.7+) prints "hello", waits 1 second,
+:term:`Coroutines <coroutine>` declared with the async/await syntax is the
+preferred way of writing asyncio applications.  For example, the following
+snippet of code (requires Python 3.7+) prints "hello", waits 1 second,
 and then prints "world"::
 
     >>> import asyncio
@@ -210,6 +210,10 @@ is :meth:`loop.run_in_executor`.
 Running an asyncio Program
 ==========================
 
+**Source code:** :source:`Lib/asyncio/runners.py`
+
+-------------------------------------------------
+
 .. function:: run(coro, \*, debug=False)
 
     Execute the :term:`coroutine` *coro* and return the result.
@@ -242,6 +246,11 @@ Running an asyncio Program
 
 Creating Tasks
 ==============
+
+**Source code:** :source:`Lib/asyncio/tasks.py`,
+:source:`Lib/asyncio/base_tasks.py`
+
+-----------------------------------
 
 .. function:: create_task(coro, \*, name=None)
 
--- a/Doc/library/asyncio-task.rst
+++ b/Doc/library/asyncio-task.rst
@@ -240,9 +240,9 @@ Running an asyncio Program
     .. versionchanged:: 3.9
        Updated to use :meth:`loop.shutdown_default_executor`.
 
-.. note::
-   The source code for ``asyncio.run()`` can be found in
-   :source:`Lib/asyncio/runners.py`.
+    .. note::
+       The source code for ``asyncio.run()`` can be found in
+       :source:`Lib/asyncio/runners.py`.
 
 Creating Tasks
 ==============
--- a/Doc/library/asyncio.rst
+++ b/Doc/library/asyncio.rst
@@ -4,6 +4,8 @@
 .. module:: asyncio
    :synopsis: Asynchronous I/O.
 
+-------------------------------
+
 .. sidebar:: Hello World!
 
    ::
@@ -90,4 +92,5 @@ Additionally, there are **low-level** APIs for
    asyncio-llapi-index.rst
    asyncio-dev.rst
 
-.. note:: The source code for asyncio can be found in :source:`Lib/asyncio/`.
+.. note::
+   The source code for asyncio can be found in :source:`Lib/asyncio/`.
--- a/Doc/library/asyncio.rst
+++ b/Doc/library/asyncio.rst
@@ -4,7 +4,9 @@
 .. module:: asyncio
    :synopsis: Asynchronous I/O.
 
---------------
+**Source code:** :source:`Lib/asyncio/`
+
+---------------------------------------
 
 .. sidebar:: Hello World!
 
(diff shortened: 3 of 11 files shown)


Doc/library/asyncio-eventloop.rst
```diff
@@ -1,10 +1,15 @@
 .. currentmodule:: asyncio
 
+.. _asyncio-event-loop:
 
 ==========
 Event Loop
 ==========
 
+**Source code:** :source:`Lib/asyncio/events.py`,
+:source:`Lib/asyncio/base_events.py`
```
aeros, line 9: In case it's not clear, both of these will appear on a single line after being rendered in Sphinx:
image

Doc/library/asyncio-task.rst
```diff
@@ -18,9 +18,9 @@ and Tasks.
 Coroutines
 ==========
 
-Coroutines declared with async/await syntax is the preferred way of
-writing asyncio applications.  For example, the following snippet
-of code (requires Python 3.7+) prints "hello", waits 1 second,
+:term:`Coroutines <coroutine>` declared with the async/await syntax is the
+preferred way of writing asyncio applications.  For example, the following
+snippet of code (requires Python 3.7+) prints "hello", waits 1 second,
```
aeros, line 23: I had to adjust the word wrapping here after adding the Sphinx role and grammar fix.

aeros:
Yury had approved of the idea to add source code links to the asyncio docs.

/cc @1st1

Doc/library/asyncio-eventloop.rst
```diff
@@ -81,8 +86,6 @@ This documentation page contains the following sections:
   loop APIs.
 
 
-.. _asyncio-event-loop:
-
```
aeros: I moved this reference to the top of the page primarily because it seemed more appropriate for the "event loops" link in https://docs.python.org/3.9/library/asyncio.html.

Doc/library/asyncio-eventloop.rst
```diff
@@ -1,10 +1,15 @@
 .. currentmodule:: asyncio
 
+.. _asyncio-event-loop:
```
1st1: Please keep the ref where it was.  The page is rather long and the idea was to jump the user right where the event loop methods are defined (skipping the prelude).

Doc/library/asyncio-eventloop.rst
1st1: 👍

Doc/library/asyncio-eventloop.rst
aeros: @1st1 Oh, I see. It seemed a bit inconsistent with the other pages and skipped over the summary, so I thought it was unintentionally misplaced. I'll correct it.

Doc/library/asyncio-eventloop.rst
aeros: Reverted to the original ref location, as requested by Yury.

Doc/library/asyncio-task.rst
```diff
@@ -210,6 +210,10 @@ is :meth:`loop.run_in_executor`.
 Running an asyncio Program
 ==========================
 
+**Source code:** :source:`Lib/asyncio/runners.py`
+
+-------------------------------------------------
```
1st1: Why do we need a line here?  I understand why it's added when a link is top level, but here it seems a bit excessive.

Doc/library/asyncio-task.rst
```diff
@@ -243,6 +247,11 @@ Running an asyncio Program
 Creating Tasks
 ==============
 
+**Source code:** :source:`Lib/asyncio/tasks.py`,
+:source:`Lib/asyncio/base_tasks.py`
+
+-----------------------------------
```
1st1: I think this link is unnecessary -- too much code, most of which is unrelated and/or arcane.

Doc/library/asyncio.rst
```diff
@@ -4,7 +4,9 @@
 .. module:: asyncio
    :synopsis: Asynchronous I/O.
 
---------------
+**Source code:** :source:`Lib/asyncio/`
+
+---------------------------------------
```
1st1: This one is also unnecessary, IMO.  I'm fine to keep it if we move it to the footer of the page.

Doc/library/asyncio-task.rst
aeros: @1st1 It's not needed by any means, I just thought it looked better with the line once I saw it rendered in Sphinx (I originally omitted the line for the non-header ones, as you suggested doing):

With line:
image

Without line:
image


Does seeing the difference in how it renders change anything or would you still prefer to remove the line? Either way is fine for me.

Doc/library/asyncio-task.rst
aeros: @1st1 Are you referring specifically to the to link for `base_tasks.py` or both links? I think we could remove `base_tasks.py`, but `tasks.py` seems like it would at least be useful for a high-level overview.

Doc/library/asyncio.rst
aeros: Okay, I can move it to the footer then.

Doc/library/asyncio-task.rst
1st1: I refer to both.  This documentation page is all about high-level APIs and how to use them correctly.  Bothering the reader with advanced source code isn't helpful -- those who really want to see it can easily look it up themselves.

So I'm fine with adding source links to low-level sections of the documentation, but against doing that in the high-level sections.

Sorry, I'm a bit anal about asyncio docs -- I spent 2 weeks full time rewriting them from scratch so that they are easier to follow. Overloading them with extra details doesn't serve that goal well.

Doc/library/asyncio-task.rst
1st1: You can add a `.. note::` at the end of the docs about asyncio.run:

```
.. note::
   The source code of the function can be found here: :source:`Lib/asyncio/runners.py`
```

Doc/library/asyncio-task.rst
aeros: @1st1 I get it, I understand your desire to keep the docs as succinct as possible and I can see how the source code for both might be arcane in this case. So, I'll remove the links to `tasks.py` and `base_tasks.py`. 

However, I do think `runners.py` has a simple enough code to justify keeping the link, despite it being a source code link in this same page. At least from personal experience, it was really helpful to see how it works and the actual code location of `asyncio.run()` may not be immediately obvious to users.

> So I'm fine with adding source links to low-level sections of the documentation, but against doing that in the high-level sections.

My only concern with this is that the distinction between high-level and low-level API can significantly change over time. Also, the difference is not entirely black and white, despite what the [front page of the asycnio docs](https://docs.python.org/3.9/library/asyncio.html) might imply. 

For example, consider the ["Synchronization Primitives" section](https://docs.python.org/3/library/asyncio-sync.html). I personally don't find `locks.py` to be at all arcane (unlike `tasks.py` and `base_tasks.py`). Also, although it's the highest level of the API for using asynchronous locks, I don't consider it be high-level in terms of having a large degree of abstraction.

So I think that rather than having a general blanket statement of only having source links in the low-level parts, the decision should be on more of a case-by-case basis.

> Sorry, I'm a bit anal about asyncio docs

That's okay. Similar to the code for asyncio, I'd much rather we take the time to strongly consider and discuss any changes made. IMO, it's definitely worthwhile considering the number of users affected; even for minor changes. 

Edit: Regarding `runners.py`, I hadn't yet read your other comment https://github.com/python/cpython/pull/16640#discussion_r333099725 before posting this. Feel free to disregard that part.

Doc/library/asyncio.rst
aeros: Done. It now renders as:

image

Doc/library/asyncio-task.rst
aeros: Done. It now renders as:

image

Doc/library/asyncio-task.rst
aeros: I removed the source links to `tasks.py` and `base_tasks.py` as requested. Leaving this open though so that you can potentially respond to my concern regarding the distinction high-level vs low-level parts of the documentation.

Doc/library/asyncio-task.rst
```diff
@@ -240,6 +240,10 @@ Running an asyncio Program
     .. versionchanged:: 3.9
        Updated to use :meth:`loop.shutdown_default_executor`.
 
+.. note::
+   The source code for ``asyncio.run()`` can be found in
+   :source:`Lib/asyncio/runners.py`.
+
```
1st1, line 246: You should indent this block 4 spaces to the right.

Doc/library/asyncio-task.rst
1st1: > So I think that rather than having a general blanket statement of only having source links in the low-level parts, the decision should be on more of a case-by-case basis.

Yes, and that's what we are trying to do here :)

I agree with you on your points.  In this particular case, the docs page about coroutines is all about high-level asyncio concepts; overloading it with links to tasks.py wouldn't help most readers, hence the decision not to include those links.

Doc/library/asyncio-task.rst
aeros: Ah yes, it would probably look more correct if it was lined up with the existing indentation.

aeros: Done. 
image

Doc/library/asyncio.rst
```diff
@@ -4,7 +4,7 @@
 .. module:: asyncio
    :synopsis: Asynchronous I/O.
 
---------------
+-------------------------------
```
aeros, line 7: The number of chars in the header line is supposed to be equal to the number of characters in the most recent non-empty line. I'm not sure if it makes a difference in rendering, but the doc tests will often have errors if it doesn't line up exactly.

1st1: This is weird, because it's a line separator and not a header.  Could you please double check that a horizontal line is rendered and not something weird is happening instead?

aeros: > This is weird, because it's a line separator and not a header

Oops, I meant line separator. I was thinking of headers because I was just reading over a previous comment.

>  Could you please double check that a horizontal line is rendered and not something weird is happening instead?

Here's what the Sphinx rendering looks like from the PR branch:
image

Which is the same as the current rendering (https://docs.python.org/3/library/asyncio.html):
image

There's no external change made, it's just an reST convention (IIRC). I can remove it if needed though. 

Edit: The slight difference in the vertical height/thickness of the line is due to python.org's CSS and not the Sphinx rendering. Here's the Sphinx rendering built from the cpython master branch:

image

Also, the blue section marker in the first screenshot was from my mouse hovering over it while taking the screenshot.

aeros: From the reST docs (http://docutils.sourceforge.net/docs/ref/rst/restructuredtext.html#sections):

" An underline/overline is a single repeated punctuation character that begins in column 1 and forms a line extending at least as far as the right edge of the title text."

"A blank line after a title is optional."

So the separators and their associated text are considered "Section Titles" regardless of whether or not there's a blank line in between them, and the repeated punctuation character (in this case `-`) is supposed to match (or exceed) the right edge of the title text.

1st1:
Alright, let's merge this! Thank you, Kyle!

1st1 (approved):

1st1:
@aeros You'll need to manually create a PR towards 3.7 (although I suggest not too, as the site will soon switch to rendering 3.8 branch)

aeros:
@1st1 
> although I suggest not too, as the site will soon switch to rendering 3.8 branch

Oh okay, we can just skip the 3.7 backport then. I'm mostly concerned with it being present in the latest stable version of the docs and for 3.9.

aeros:
Also, thanks for the great feedback and discussions as usual Yury. (:

PR #16651, opened 2019-10-08
bpo-38118: Ignore Valgrind false alarm in PyUnicode_Decode()

vstinner:
Valgrind emits "Conditional jump or move depends on uninitialised
value(s)" false alarms on GCC builtin strcmp() function. The GCC code
is correct.

Valgrind bug: https://bugs.kde.org/show_bug.cgi?id=264936

https://bugs.python.org/issue38118


vstinner:
I closed GH-16653 since the Azure Pipelines CI didn't start.

PR #16658, opened 2019-10-08
bpo-38379:  don't claim objects are collected when they aren't

tim-one:
When a finalizer resurrects an object, nothing is actually collected in this run of gc.  Change the stats to reflect that truth.

https://bugs.python.org/issue38379

Automerge-Triggered-By: @tim-one


Lib/test/test_gc.py
```diff
@@ -822,6 +822,68 @@ def test_get_objects_arguments(self):
         self.assertRaises(TypeError, gc.get_objects, "1")
         self.assertRaises(TypeError, gc.get_objects, 1.234)
 
+    def test_38379(self):
```
nascheme, line 825: Maybe add a comment here explaining what the test is about.  "test_38379" will be a bit mysterious a few years from now.

nascheme:
LGTM.  I don't think the performance impact is significant either and the stats are wrong without this fix.

tim-one:
> I don't think the performance impact is significant either

There is no performance hit here - that would come with the second half of the bpo suggestions, if gc were taught how to keep on collecting despite that a finalizer resurrected some trash.  As is (with or without this PR so far), gc gives up after detecting a resurrection.  Which leads to the test case on the bpo issue, where trash keeps on building up forever and gc never make progress.

Lib/test/test_gc.py
tim-one: Absolutely - working with one finger goes slow 😉

tim-one:
I have no idea how to backport this (well - or anything else).  I installed `cherry_picker` and ran the suggested commands, but even in `--dry-run` mode they blow up with stuff like

> subprocess.CalledProcessError: Command '['git', 'log', '-r', 'ecbf35f']' returned non-zero exit status 128.

So I'm giving up before I do real harm ☹️

PR #16659, opened 2019-10-08
bpo-37531: regrtest ignores output on timeout

vstinner:
[bpo-37531](https://bugs.python.org/issue37531), [bpo-38207](https://bugs.python.org/issue38207): On timeout, regrtest no longer attempts to call
`popen.communicate() again: on Windows, it can hang until all child
processes using stdout and stderr pipes completes. Kill the worker
process and ignores its output.

Reenable test_regrtest.test_multiprocessing_timeout().

[bpo-37531](https://bugs.python.org/issue37531): Change also the faulthandler timeout of the main process
from 1 minute to 5 minutes, for Python slowest buildbots.


vstinner:
Oh, I forgot about https://bugs.python.org/issue38207 details: in fact, Linux is also affected. It's not specific to Windows.

PR #16663, opened 2019-10-08
[3.7] bpo-38395: Fix ownership in weakref.proxy methods (GH-16632)

pablogsal:
The implementation of weakref.proxy's methods call back into the Python
API using a borrowed references of the weakly referenced object
(acquired via PyWeakref_GET_OBJECT). This API call may delete the last
reference to the object (either directly or via GC), leaving a dangling
pointer, which can be subsequently dereferenced.

To fix this, claim a temporary ownership of the referenced object when
calling the appropriate method. Some functions because at the moment they
do not need to access the borrowed referent, but to protect against
future changes to these functions, ownership need to be fixed in
all potentially affected methods..
(cherry picked from commit 10cd00a9e3c22af37c748ea5a417f6fb66601e21)

Co-authored-by: Pablo Galindo <Pablogsal@gmail.com>

https://bugs.python.org/issue38395

Automerge-Triggered-By: @ned-deily

### DIFF
--- a/Objects/weakrefobject.c
+++ b/Objects/weakrefobject.c
@@ -145,11 +145,14 @@ weakref_hash(PyWeakReference *self)
 {
     if (self->hash != -1)
         return self->hash;
-    if (PyWeakref_GET_OBJECT(self) == Py_None) {
+    PyObject* obj = PyWeakref_GET_OBJECT(self);
+    if (obj == Py_None) {
         PyErr_SetString(PyExc_TypeError, "weak object has gone away");
         return -1;
     }
-    self->hash = PyObject_Hash(PyWeakref_GET_OBJECT(self));
+    Py_INCREF(obj);
+    self->hash = PyObject_Hash(obj);
+    Py_DECREF(obj);
     return self->hash;
 }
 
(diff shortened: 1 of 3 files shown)


Objects/weakrefobject.c
```diff
@@ -159,28 +162,43 @@ weakref_repr(PyWeakReference *self)
...
     PyObject *name, *repr;
     _Py_IDENTIFIER(__name__);
+    PyObject* obj = PyWeakref_GET_OBJECT(self);
 
-    if (PyWeakref_GET_OBJECT(self) == Py_None)
+    if (obj == Py_None) {
         return PyUnicode_FromFormat("<weakref at %p; dead>", self);
+    }
 
+<<<<<<< HEAD
```
ned-deily: This appears to be an unresolved merge conflict.

ned-deily (changes requested):

Objects/weakrefobject.c
pablogsal: Ooops... Sorry, I missed this one while manually backporting  :( Fixed in b68f6d6

Thanks, Ned!

pablogsal:

ned-deily (approved):

PR #16665, opened 2019-10-08
bpo-38109: Add missing constants to Lib/stat.py

rlamy:
Lib/stat.py was missing the constants S_IFDOOR, S_IFPORT, S_IFWHT as well as the related S_IS... functions. This is a port of https://bitbucket.org/pypy/pypy/commits/f245f35c61eb.

https://bugs.python.org/issue38109

### DIFF
--- a/Lib/stat.py
+++ b/Lib/stat.py
@@ -40,6 +40,10 @@ S_IFREG  = 0o100000  # regular file
 S_IFIFO  = 0o010000  # fifo (named pipe)
 S_IFLNK  = 0o120000  # symbolic link
 S_IFSOCK = 0o140000  # socket file
+# Fallbacks for uncommon platform-specific constants
+S_IFDOOR = 0
+S_IFPORT = 0
+S_IFWHT = 0
 
 # Functions to test for each file type
 
@@ -70,6 +74,18 @@ def S_ISLNK(mode):
 def S_ISSOCK(mode):
     """Return True if mode is from a socket."""
     return S_IFMT(mode) == S_IFSOCK
+
+def S_ISDOOR(mode):
+    """Return True if mode is from a door."""
+    return False
+
+def S_ISPORT(mode):
+    """Return True if mode is from an event port."""
+    return False
+
+def S_ISWHT(mode):
+    """Return True if mode is from a whiteout."""
+    return False
 
 # Names for permission bits
 
--- a/Lib/test/test_stat.py
+++ b/Lib/test/test_stat.py
@@ -16,10 +16,10 @@ class TestFilemode:
                   'UF_IMMUTABLE', 'UF_NODUMP', 'UF_NOUNLINK', 'UF_OPAQUE'}
 
     formats = {'S_IFBLK', 'S_IFCHR', 'S_IFDIR', 'S_IFIFO', 'S_IFLNK',
-               'S_IFREG', 'S_IFSOCK'}
+               'S_IFREG', 'S_IFSOCK', 'S_IFDOOR', 'S_IFPORT', 'S_IFWHT'}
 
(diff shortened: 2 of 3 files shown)


brandtbucher:
Thanks for your time @rlamy, and welcome to CPython! :sunglasses:

This looks like a good change to me! It should also have [a NEWS entry](https://blurb-it.herokuapp.com), though. Just something like:

```rst
Add missing :data:`stat.S_IFDOOR`, :data:`stat.S_IFPORT`, :data:`stat.S_IFWHT`,
:func:`stat.S_ISDOOR`, :func:`stat.S_ISPORT`, and :func:`stat.S_ISWHT` values to
the Python implementation of :mod:`stat`.
```

CC @tiran

brandtbucher (changes requested):
This looks like a good change to me! It should also have [a NEWS entry](https://blurb-it.herokuapp.com), though. Just something like:

```rst
Add missing :data:`stat.S_IFDOOR`, :data:`stat.S_IFPORT`, :data:`stat.S_IFWHT`,
:func:`stat.S_ISDOOR`, :func:`stat.S_ISPORT`, and :func:`stat.S_ISWHT` values to
the Python implementation of :mod:`stat`.
```

vstinner:
I can reproduce https://bugs.python.org/issue38109 with this example:

```
import sys
sys.modules.pop('stat', None)
sys.modules['_stat'] = None
import stat
print(stat.S_ISDOOR(123))
```

Output:

```
AttributeError: module 'stat' has no attribute 'S_ISDOOR'
```

It works as expected with this PR:

```
False
```

vstinner (approved):
LGTM.

Would you mind to add a NEWS entry using the blurb tool? You can use: https://github.com/python/cpython/pull/16665#pullrequestreview-299017207

brandtbucher:
Congrats on your first CPython contribution @rlamy! :champagne:

vstinner:
Ronan already helped CPython in the past :-)

```
commit 569ff4fbbc4cd0589c369519745d8d17bf3094b0
Author: Brett Cannon <brett@python.org>
Date:   Fri Jan 11 18:09:25 2013 -0500

    Issue #15031: Refactor some code in importlib pertaining to validating
    and compiling bytecode.
    
    Thanks to Ronan Lamy for pointing the redundancy and taking an initial
    stab at the refactor (as did Nick Coghlan).

commit 1e331560eea62c78ec189f2b72b59864ee315ddc
Author: Brett Cannon <brett@python.org>
Date:   Mon Jul 2 14:35:34 2012 -0400

    Closes #15030: Make importlib.abc.PyPycLoader respect the new .pyc
    file size header field.
    
    Thanks to Marc Abramowitz and Ronan Lamy for helping out with various
    parts of the patch.

commit ea0b8239401123fa7f41c74f6fc9ded1cf74088a
Author: Brett Cannon <brett@python.org>
Date:   Fri Jun 15 20:00:53 2012 -0400

    Issue #14938: importlib.abc.SourceLoader.is_package() now takes the
    module name into consideration when determining whether a module is a
    package or not. This prevents importing a module's __init__ module
    directly and having it considered a package, which can lead to
    duplicate sub-modules.
    
    Thanks to Ronan Lamy for reporting the bug.
```

brandtbucher:
Ah, my bad. Github’s "First-time Contributor" badge led me to believe otherwise.

Thanks for the clarification @vstinner!

vstinner:
> Ah, my bad. Github’s "First-time Contributor" badge led me to believe otherwise.

Hehe. Python only migrated to Git and GitHub 2 years ago. Previously, we only kept track of the authors in the commit message, not in the author field.

brandtbucher:
*sigh* I guess in the future I’ll have to check `Misc/ACKS` too... Ronan is in there.

PR #16667, opened 2019-10-08
bpo-38415 @asynccontextmanager as decorators like @contextmanager

fried:
I assumed at that an asynccontextmanager would some time be available in the stdlib when we first got async generators.  Before it had been released I cooked up an internal version at facebook whiched mapped all the features of @contextmanager but as an async variant.   This one feature seems to be missing from the stdlib.  

I would like to stop using my version and just use the stdlib version, can we get this feature parity with contextmanager? 

```
@asynccontextmanager
async def our_context(...):
     ...
     yield
     ...

@our_context(...)
async def some_function(...):
   # we are inside the context of our_context now
    ...

@our_context(...)
async def some_other_function(...):
   ...
```

https://bugs.python.org/issue38415

### DIFF
--- a/Lib/contextlib.py
+++ b/Lib/contextlib.py
@@ -191,6 +191,14 @@ class _AsyncGeneratorContextManager(
 ):
     """Helper for @asynccontextmanager decorator."""
 
+    def __call__(self, func):
+        @wraps(func)
+        async def inner(*args, **kwds):
+            async with self.__class__(self.func, self.args, self.kwds):
+                return await func(*args, **kwds)
+
+        return inner
+
     async def __aenter__(self):
         # do not keep args and kwds alive unnecessarily
         # they are only needed for recreation, which is not possible anymore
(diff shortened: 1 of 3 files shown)


1st1:
Awesome! Please create a bpo issue and generate a `NEWS` file using the `blurb` tool.

1st1 (approved):

fried:
alright

1st1:
@ncoghlan Nick, do you want to take a look at this?  Looks good to me.

asvetlov (approved):

ambv:
Rebased on current main (3.11).

Lib/contextlib.py
```diff
@@ -191,6 +191,14 @@ class _AsyncGeneratorContextManager(
...
     """Helper for @asynccontextmanager decorator."""
 
+    def __call__(self, func):
+        @wraps(func)
+        async def inner(*args, **kwds):
+            async with self.__class__(self.func, self.args, self.kwds):
+                return await func(*args, **kwds)
+
+        return inner
+
```
graingert, line 201: this function is now redundant - support for asynccontextmanagers as decorators was added in https://github.com/python/cpython/pull/20516

PR #16679, opened 2019-10-09
bpo-38422: Clarify docstrings of pathlib suffix(es)

cool-RR:
Whenever I use `path.suffix` I have to check again whether it includes the dot or not. I decided to add it to the docstring so I won't have to keep checking. 

https://bugs.python.org/issue38422

### DIFF
--- a/Lib/pathlib.py
+++ b/Lib/pathlib.py
@@ -795,7 +795,11 @@ class PurePath(object):
 
     @property
     def suffix(self):
-        """The final component's last suffix, if any."""
+        """
+        The final component's last suffix, if any.
+
+        This includes the leading period. For example: '.txt'
+        """
         name = self.name
         i = name.rfind('.')
         if 0 < i < len(name) - 1:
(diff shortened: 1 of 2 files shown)


brandtbucher (approved):
Same here, thanks for this improvement @cool-RR!

Lib/pathlib.py
```diff
@@ -795,7 +795,11 @@ def name(self):
 
     @property
     def suffix(self):
-        """The final component's last suffix, if any."""
+        """
+        The final component's last suffix, if any.
+
+        This includes the leading dot. For example: '.txt'
```
taleinat: The `.` character is most often called a "period" rather than a "dot", and I suggest using "period" here as well.

"dot" is ambiguous; there are other dot characters, e.g. ["interpunct" a.k.a. "middle dot"](https://en.wikipedia.org/wiki/Interpunct).

While this page in the docs does refer to periods as "dots" in one other place, in the [os.path.splitext() docs](https://docs.python.org/3/library/os.path.html#os.path.splitext) they are indeed called "periods".

cool-RR: Done.

pitrou (approved):
An obvious improvement. Thanks @cool-RR for doing this.

PR #16681, opened 2019-10-09
bpo-38409: Grammatically correct help documentation of str.strip()

hansrajdas:
* `remove` should be `removed` as done for lstrip and rstrip help docs

strip help doc in python2 was correct but it was updated in python3:
python2.7
```
strip(...)
    S.strip([chars]) -> string or unicode

    Return a copy of the string S with leading and trailing
    whitespace removed.
    If chars is given and not None, remove characters in chars instead.
    If chars is unicode, S will be converted to unicode before stripping
```
python3.7
```
strip(self, chars=None, /)
    Return a copy of the string with leading and trailing whitespace remove.

    If chars is given and not None, remove characters in chars instead.
```

https://bugs.python.org/issue38409

### DIFF
--- a/Misc/NEWS.d/next/Documentation/2019-10-10-01-00-29.bpo-38409.UQgQk1.rst
+++ b/Misc/NEWS.d/next/Documentation/2019-10-10-01-00-29.bpo-38409.UQgQk1.rst
@@ -1,1 +1,0 @@
-Grammatically corrected help(str.strip) documentation. Patch by Hansraj Das
--- a/Misc/NEWS.d/next/Documentation/2019-10-10-01-00-29.bpo-38409.UQgQk1.rst
+++ b/Misc/NEWS.d/next/Documentation/2019-10-10-01-00-29.bpo-38409.UQgQk1.rst
@@ -1,0 +1,1 @@
+Grammatically corrected help(str.strip) documentation. Patch by Hansraj Das
--- a/Objects/clinic/unicodeobject.c.h
+++ b/Objects/clinic/unicodeobject.c.h
@@ -582,7 +582,7 @@ PyDoc_STRVAR(unicode_strip__doc__,
 "strip($self, chars=None, /)\n"
 "--\n"
 "\n"
-"Return a copy of the string with leading and trailing whitespace remove.\n"
+"Return a copy of the string with leading and trailing whitespace removed.\n"
 "\n"
 "If chars is given and not None, remove characters in chars instead.");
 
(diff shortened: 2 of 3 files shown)


Misc/NEWS.d/next/Documentation/2019-10-10-01-00-29.bpo-38409.UQgQk1.rst
```diff
@@ -0,0 +1 @@
+Grammatically corrected help(str.strip) documentation. Patch by Hansraj Das
```
ammaraskar: Trivial fixes like these don't need a news blurb. I'm gonna apply to the `skip-news` label if you wanna take it out.

hansrajdas: Deleted news blurb file. Thank you!

ammaraskar (approved):

Objects/clinic/unicodeobject.c.h
```diff
@@ -582,7 +582,7 @@ PyDoc_STRVAR(unicode_strip__doc__,
 "strip($self, chars=None, /)\n"
 "--\n"
 "\n"
-"Return a copy of the string with leading and trailing whitespace remove.\n"
+"Return a copy of the string with leading and trailing whitespace removed.\n"
```
ammaraskar, line 585: This file is automatically generated using argument clinic, you need to regenerate it using `make regen-all` or `make clinic`

zware:
Considering that this is incorrect and was taken without attribution or approval of the original author, I'm closing this in favor of GH-16682.  Since the original author hasn't signed the CLA, the usability of this patch is murky anyway.

Thanks for contributing, @hansrajdas, but when the reporter has offered a patch on bugs.python.org please give them a chance to either submit their own PR or refuse to do so; this one was submitted only yesterday.

Objects/clinic/unicodeobject.c.h
hansrajdas: executed make clinic

hansrajdas:
> Thanks for contributing, @hansrajdas, but when the reporter has offered a patch on bugs.python.org please give them a chance to either submit their own PR or refuse to do so; this one was submitted only yesterday.

Yes, I saw this bug reported yesterday itself(maked myself in that bug also yesterday) so waited one day if original author wants to send an PR. Sorry, I should have waited more!

zware:
The [documented](https://devguide.python.org/pullrequest/#converting-an-existing-patch-from-b-p-o-to-github) suggested wait time is one week of no response.

hansrajdas:
> The [documented](https://devguide.python.org/pullrequest/#converting-an-existing-patch-from-b-p-o-to-github) suggested wait time is one week of no response.

Thank you @zware for guidance, I was not aware of this. Will always follow this guideline ever in future!

PR #16687, opened 2019-10-10
bpo-38379: Don't block collection of unreachable objects when some objects resurrect

pablogsal:
https://bugs.python.org/issue38379

### DIFF
--- a/Lib/test/test_gc.py
+++ b/Lib/test/test_gc.py
@@ -862,8 +862,8 @@ class GCTests(unittest.TestCase):
         # Nothing is collected - Z() is merely resurrected.
         t = gc.collect()
         c, nc = getstats()
-        self.assertEqual(t, 0)  # before
-        self.assertEqual(c - oldc, 0)   # before
+        self.assertEqual(t, 0)
+        self.assertEqual(c - oldc, 0)
         self.assertEqual(nc - oldnc, 0)
 
         # Z() should not prevent anything else from being collected.
@@ -873,8 +873,8 @@ class GCTests(unittest.TestCase):
         Z()
         t = gc.collect()
         c, nc = getstats()
-        self.assertEqual(t, 2*N)  # before
-        self.assertEqual(c - oldc, 2*N)   # before
+        self.assertEqual(t, 2*N)
+        self.assertEqual(c - oldc, 2*N)
         self.assertEqual(nc - oldnc, 0)
 
         # The A() trash should have been reclaimed already but the
--- a/Lib/test/test_gc.py
+++ b/Lib/test/test_gc.py
@@ -1213,7 +1213,7 @@ class GCTogglingTests(unittest.TestCase):
             # If __del__ resurrected c2, the instance would be damaged, with an
             # empty __dict__.
             self.assertEqual(x, None)
-    
+
 
 def test_main():
     enabled = gc.isenabled()
--- a/Lib/test/test_gc.py
+++ b/Lib/test/test_gc.py
@@ -823,6 +823,7 @@ class GCTests(unittest.TestCase):
         self.assertRaises(TypeError, gc.get_objects, 1.234)
 
     def test_resurrection_does_not_block_cleanup_of_other_objects(self):
+
         # When a finalizer resurrects objects, stats were reporting them as
         # having been collected.  This affected both collect()'s return
         # value and the dicts returned by get_stats().
@@ -861,30 +862,25 @@ class GCTests(unittest.TestCase):
         # Nothing is collected - Z() is merely resurrected.
         t = gc.collect()
         c, nc = getstats()
-        #self.assertEqual(t, 2)  # before
-        self.assertEqual(t, 0)  # after
-        #self.assertEqual(c - oldc, 2)   # before
-        self.assertEqual(c - oldc, 0)   # after
+        self.assertEqual(t, 0)  # before
+        self.assertEqual(c - oldc, 0)   # before
         self.assertEqual(nc - oldnc, 0)
 
-        # Unfortunately, a Z() prevents _anything_ from being collected.
-        # It should be possible to collect the A instances anyway, but
-        # that will require non-trivial code changes.
+        # Z() should not prevents anything from being collected.
         oldc, oldnc = c, nc
         for i in range(N):
             A()
         Z()
-        # Z() prevents anything from being collected.
         t = gc.collect()
         c, nc = getstats()
-        self.assertEqual(t, 2*N)  # after
-        self.assertEqual(c - oldc, 2*N)   # after
+        self.assertEqual(t, 2*N)  # before
+        self.assertEqual(c - oldc, 2*N)   # before
         self.assertEqual(nc - oldnc, 0)
 
-        # The A() trash should have been eliminated and
-        # the Z objects should be removed now
-        zs.clear()
+        # The A() trash should have been reclaimed already but the
+        # 2 copies of Z are still in zs (and the associated dicts).
         oldc, oldnc = c, nc
+        zs.clear()
         t = gc.collect()
         c, nc = getstats()
         self.assertEqual(t, 4)
@@ -1213,7 +1209,6 @@ class GCTogglingTests(unittest.TestCase):
             # If __del__ resurrected c2, the instance would be damaged, with an
             # empty __dict__.
             self.assertEqual(x, None)
-
 
 def test_main():
     enabled = gc.isenabled()
--- a/Lib/test/test_gc.py
+++ b/Lib/test/test_gc.py
@@ -822,7 +822,7 @@ class GCTests(unittest.TestCase):
         self.assertRaises(TypeError, gc.get_objects, "1")
         self.assertRaises(TypeError, gc.get_objects, 1.234)
 
-    def test_38379(self):
+    def test_resurrection_does_not_block_cleanup_of_other_objects(self):
         # When a finalizer resurrects objects, stats were reporting them as
         # having been collected.  This affected both collect()'s return
         # value and the dicts returned by get_stats().
@@ -877,18 +877,18 @@ class GCTests(unittest.TestCase):
         # Z() prevents anything from being collected.
         t = gc.collect()
         c, nc = getstats()
-        #self.assertEqual(t, 2*N + 2)  # before
-        self.assertEqual(t, 0)  # after
-        #self.assertEqual(c - oldc, 2*N + 2)   # before
-        self.assertEqual(c - oldc, 0)   # after
+        self.assertEqual(t, 2*N)  # after
+        self.assertEqual(c - oldc, 2*N)   # after
         self.assertEqual(nc - oldnc, 0)
 
-        # But the A() trash is reclaimed on the next run.
+        # The A() trash should have been eliminated and
+        # the Z objects should be removed now
+        zs.clear()
         oldc, oldnc = c, nc
         t = gc.collect()
         c, nc = getstats()
-        self.assertEqual(t, 2*N)
-        self.assertEqual(c - oldc, 2*N)
+        self.assertEqual(t, 4)
+        self.assertEqual(c - oldc, 4)
         self.assertEqual(nc - oldnc, 0)
 
         gc.enable()
@@ -1213,6 +1213,7 @@ class GCTogglingTests(unittest.TestCase):
             # If __del__ resurrected c2, the instance would be damaged, with an
             # empty __dict__.
             self.assertEqual(x, None)
+    
 
 def test_main():
     enabled = gc.isenabled()
--- a/Lib/test/test_gc.py
+++ b/Lib/test/test_gc.py
@@ -822,6 +822,76 @@ class GCTests(unittest.TestCase):
         self.assertRaises(TypeError, gc.get_objects, "1")
         self.assertRaises(TypeError, gc.get_objects, 1.234)
 
+    def test_resurrection_only_happens_once_per_object(self):
+        class A:  # simple self-loop
+            def __init__(self):
+                self.me = self
+
+        class Lazarus(A):
+            resurrected = 0
+            resurrected_instances = []
+
+            def __del__(self):
+                Lazarus.resurrected += 1
+                Lazarus.resurrected_instances.append(self)
+
+        gc.collect()
+        gc.disable()
+
+        # We start with 0 resurrections
+        laz = Lazarus()
+        self.assertEqual(Lazarus.resurrected, 0)
+
+        # Deleting the instance and triggering a collection
+        # resurrects the object
+        del laz
+        gc.collect()
+        self.assertEqual(Lazarus.resurrected, 1)
+        self.assertEqual(len(Lazarus.resurrected_instances), 1)
+
+        # Clearing the references and forcing a collection
+        # should not resurrect the object again.
+        Lazarus.resurrected_instances.clear()
+        self.assertEqual(Lazarus.resurrected, 1)
+        gc.collect()
+        self.assertEqual(Lazarus.resurrected, 1)
+
+        gc.enable()
+
+    def test_resurrection_is_transitive(self):
+        class Cargo:
+            def __init__(self):
+                self.me = self
+
+        class Lazarus:
+            resurrected_instances = []
+
+            def __del__(self):
+                Lazarus.resurrected_instances.append(self)
+
+        gc.collect()
+        gc.disable()
+
+        laz = Lazarus()
+        cargo = Cargo()
+        cargo_id = id(cargo)
+
+        # Create a cycle between cargo and laz
+        laz.cargo = cargo
+        cargo.laz = laz
+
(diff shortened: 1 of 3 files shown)


Lib/test/test_gc.py
```diff
@@ -877,18 +877,18 @@ def getstats():
         # Z() prevents anything from being collected.
```
tim-one: This, and the comment block above it, are no longer true.  Change to say it _used_ to be true, but that Z() no longer prevents collecting the A()s.

Modules/gcmodule.c
```diff
@@ -624,20 +624,20 @@ visit_move(PyObject *op, PyGC_Head *tolist)
...
 static void
-move_legacy_finalizer_reachable(PyGC_Head *finalizers)
+move_recursive_reachable(PyGC_Head *gc_list)
 {
     traverseproc traverse;
-    PyGC_Head *gc = GC_NEXT(finalizers);
-    for (; gc != finalizers; gc = GC_NEXT(gc)) {
+    PyGC_Head *gc = GC_NEXT(gc_list);
+    for (; gc != gc_list; gc = GC_NEXT(gc)) {
         /* Note that the finalizers list may grow during this. */
```
tim-one: s/gc_list/finalizers/

Modules/gcmodule.c
```diff
@@ -892,7 +892,7 @@ finalize_garbage(PyGC_Head *collectable)
    from the outside (some objects could have been resurrected by a
    finalizer). */
 static int
-check_garbage(PyGC_Head *collectable)
+deduce_unreachable(PyGC_Head *collectable, PyGC_Head *still_reachable)
```
tim-one: `still_reachable` is an odd name for this - maybe `now_reachable`?

Modules/gcmodule.c
```diff
@@ -1111,17 +1128,21 @@ collect(struct _gc_runtime_state *state, int generation,
...
-        delete_garbage(state, &unreachable, old);
-    }
+    // The deduce_unreachable function will move all objects that are still
+    // reachable from the outside into the still_unreachable set.
+    PyGC_Head still_unreachable;
+    gc_list_init(&still_unreachable);
+    if (deduce_unreachable(&unreachable, &still_unreachable)) { // clear PREV_MASK_COLLECTING here
+        // Move every resurrected object (and the ones reachable from them) into
+        // the yoing list for future collection.
+        gc_list_merge(&still_unreachable, young);
```
tim-one: `still_unreachable` throughout the last several lines should be, e.g., `now_reachable` instead, yes?

tim-one: The merge should be into `old` rather than `young`.  A run of gc should leave all generations <= `young` empty (unless `young == old`).

tim-one (commented):
This surprised me - nice!  Left some comments, but am too fried to do a thorough review tonight.  It's certainly on the intended track.

Modules/gcmodule.c
```diff
@@ -299,7 +299,7 @@ gc_list_size(PyGC_Head *list)
 }
 
 /* Append objects in a GC list to a Python list.
- * Return 0 if all OK, < 0 if error (out of memory for list).
+ * Return 0 if all OK, < 0 if error (out of memory for list)deduce_unreachable.
```
tim-one: `deduce_unreachable` looks like a mistake here (has nothing to do with the comment).

Modules/gcmodule.c
```diff
@@ -610,6 +610,19 @@ move_legacy_finalizers(PyGC_Head *unreachable, PyGC_Head *finalizers)
     }
 }
 
+static inline void
+restore_unreachable_mask(PyGC_Head *unreachable)
```
tim-one: Think the name would be clearer as `clear_...` rather than `restore...`.

Modules/gcmodule.c
```diff
@@ -610,6 +610,19 @@ move_legacy_finalizers(PyGC_Head *unreachable, PyGC_Head *finalizers)
...
+{
+    PyGC_Head *gc, *next;
+    unreachable->_gc_next &= ~NEXT_MASK_UNREACHABLE;
```
tim-one, line 633: I believe list heads never have these goofy 😉 flags set.  So better to assert that the flag isn't set here.

Modules/gcmodule.c
```diff
@@ -1000,6 +1013,26 @@ show_stats_each_generations(struct _gc_runtime_state *state)
...
+     * container set are reachable from outside the set (i.e., have a
+     * refcount greater than 0 when all the references within the
+     * set are taken into account).
+     */
+    update_refs(base);  // gc_prev is used for gc_refs
+    subtract_refs(base);
+
+    /* Leave everything reachable from outside young in young, and move
+     * everything else (in young) to unreachable.
+     * NOTE:  This used to move the reachable objects into a reachable
```
tim-one, line 1043: `young` should be `base` throughout.

Modules/gcmodule.c
pablogsal: I added a call to  `validate_list` after `unreachable` returns to be a normal list and an assert at the begging for the list head. Notice that we can not add such a test to `unreachable->_gc_next` because `unreachable->_gc_next` will not have the bit set if the list is empty IIRC.

Modules/gcmodule.c
pablogsal: Good catch, thanks!

Modules/gcmodule.c
pablogsal: Sorry about this, I was coding in a different editor and I clearly don't manage it perfectly :(

Modules/gcmodule.c
tim-one: Not sure I followed that.  `unreachable` (any gc list) is empty if and only if  `unreachable->gc_next == unreachable` .  No flags in the header pointers will ever be set regardless,  empty or not.  Indeed, that's another thing `validate_list` could verify.

Lib/test/test_gc.py
```diff
@@ -861,34 +862,29 @@ def getstats():
         # Nothing is collected - Z() is merely resurrected.
         t = gc.collect()
         c, nc = getstats()
-        #self.assertEqual(t, 2)  # before
-        self.assertEqual(t, 0)  # after
-        #self.assertEqual(c - oldc, 2)   # before
-        self.assertEqual(c - oldc, 0)   # after
+        self.assertEqual(t, 0)  # before
+        self.assertEqual(c - oldc, 0)   # before
```
tim-one: The `before` comments don't make sense anymore.

Modules/gcmodule.c
```diff
@@ -1000,6 +1017,55 @@ show_stats_each_generations(struct _gc_runtime_state *state)
         buf, gc_list_size(&state->permanent_generation.head));
 }
 
+/* Deduce wich objects among "base" are unreachable from outside the list
+   and move them to 'unreachable'. The process consist in the following steps:
+
+1. Copy all references to a different field (gc_prev is used to hold this copy
```
tim-one: We're not copying "references", but refcounts.  Big difference 😉

Modules/gcmodule.c
```diff
@@ -1000,6 +1017,55 @@ show_stats_each_generations(struct _gc_runtime_state *state)
...
 
+/* Deduce wich objects among "base" are unreachable from outside the list
+   and move them to 'unreachable'. The process consist in the following steps:
+
+1. Copy all references to a different field (gc_prev is used to hold this copy 
+   to save memory).
+2. Traverse all objects in "base" and visit all referred objects using
+   "tp_traverse" and for every visited object, substract 1 to the reference
+   count (the one that we copied in the previous step). After this step, all
+   objects that can be reached from outside must have positive reference count,
```
tim-one: change "from outside" by adding "directly" before it.
change "positive" to "strictly positive".

Modules/gcmodule.c
```diff
@@ -1000,6 +1017,55 @@ show_stats_each_generations(struct _gc_runtime_state *state)
...
+   them to the "unreachable" list. This step also needs to move back to "base" all
+   objects that were initially marked as unreachable but are referred transitively
+   by the reachable objects (the ones with positive reference count).
+
+Contracts:
+
+    * The "base" has to be a valid list with no mask set.
+
+    * The "unreachable" list must be uninitialized (this function calls
+      gc_list_init over 'unreachable').
```
tim-one, line 1023: Nice addition!  I mean initializing the list.

Modules/gcmodule.c
```diff
@@ -1000,6 +1017,55 @@ show_stats_each_generations(struct _gc_runtime_state *state)
...
+    
+IMPORTANT: This function leaves 'unreachable' with the NEXT_MASK_UNREACHABLE
+flag set but it does not clear it to skip unnecessary iteration. Before the
+flag is cleared (for example, by using 'clear_unreachable_mask' function or
+by a call to 'move_legacy_finalizers'), the 'unreachable' list is not a normal
+list and we can not use mostgc_list_* functions for it. */
```
tim-one: `mostgc_list` is missing a space.

Modules/gcmodule.c
```diff
@@ -1111,17 +1160,25 @@ collect(struct _gc_runtime_state *state, int generation,
...
-    }
-    else {
-        /* Call tp_clear on objects in the unreachable set.  This will cause
-         * the reference cycles to be broken.  It may also cause some objects
-         * in finalizers to be freed.
-         */
-        m += gc_list_size(&unreachable);
-        delete_garbage(state, &unreachable, old);
-    }
+    PyGC_Head still_unreachable;
```
tim-one: `still_unreachable` is used only inside the following `if` block, so its declaration should be in that block too.

tim-one (commented):
:Left more small comments.  Higher level:  does `check_garbage()` really serve a purpose anymore?  It's doing 2/3rds of the work done by the new function _just_ to see whether at least one thing has become directly reachable.  Which the new function does anyway.  Since the amount of trash is usually small, the relative expense of running the new function all the time is minor (it's not crawling over the entire generation, just the trash objects - which is usually an empty list!).

The fewer moving parts the better 😉

Modules/gcmodule.c
pablogsal: I may be missing something.... If you are referring to add:

```diff
assert((unreachable->_gc_next & NEXT_MASK_UNREACHABLE) == 0); 
```

that will fail as the `_gc_next` field has indeed the flag set:

```
❯ ./python -m test test_gc 
0:00:00 load avg: 1.22 Run tests sequentially
0:00:00 load avg: 1.22 [1/1] test_gc
python: Modules/gcmodule.c:620: clear_unreachable_mask: Assertion `(unreachable->_gc_next & NEXT_MASK_UNREACHABLE) == 0' failed.
Fatal Python error: Aborted
```

Or are you referring to the head itself?

```
    assert(((uintptr_t)unreachable & NEXT_MASK_UNREACHABLE) == 0);
    assert(((uintptr_t)unreachable & PREV_MASK_COLLECTING) == 0)
```

Apologies for the confusion!

pablogsal:
>  Higher level: does `check_garbage()` really serve a purpose anymore? It's doing 2/3rds of the work done by the new function _just_ to see whether at least one thing has become directly reachable. Which the new function does anyway. Since the amount of trash is usually small, the relative expense of running the new function all the time is minor (it's not crawling over the entire generation, just the trash objects - which is usually an empty list!).

I very much agree with this! I did not have removed check_garbage because I wanted to make sure everything else looks fine before discussing about just doing the work every time (calling `deduce_unreachable` over `unreachable`). 

To make sure our assumption is correct, I will run the pyperformance test suite :) (Although I think is  not really necesary, but checking does not hurt),

Modules/gcmodule.c
tim-one: Nope , my apologies - the "rule" that a list head's `next` and `prev` pointers are never polluted with flags appears to be deliberately violated by `move_unreachable` for the `unreachable` list head's `next` pointer.   So your original code was both correct & necessary.

tim-one:
> On the other hand check_garbage is setting clearing the flags (it calls gc_clear_collecting) so someone else needs to do that if we remove the call. Where do you think is the best place to do this if we remove the function?

The bit-fiddling obfuscation in this module is depressing ☹️  There are no principled answers:  the only scheme at work appears to be "delay undoing a trick as long as possible, regardless of how long that leaves things in a variety of semi-broken states, or how obscure they are".

So repair all the remaining damage before `delete_garbage()` is called?  That's the last non-trivial thing collection does.

pablogsal:
> The bit-fiddling obfuscation in this module is depressing ☹️ There are no principled answers: the only scheme at work appears to be "delay undoing a trick as long as possible, regardless of how long that leaves things in a variety of semi-broken states, or how obscure they are".

Yep, modifying this code is surprisingly tricky as there are lots of contracts not precisely well documented and many optimization details make the thing even more obscure.

> So repair all the remaining damage before `delete_garbage()` is called? That's the last non-trivial thing collection does.

Unless we somehow introduce the call to `gc_clear_collecting` somewhere, we will end implementing 90% of the work that `check_garbage` is doing, as the main overhead is the iteration over the list.

tim-one:
I'm not looking to do less work than `check_garbage`, just looking to do it once instead of twice.  More, I think it's a Bad Idea to follow this module's lead of being as "clever" as possible with no regard for how brittle that makes the code.  Do the simplest & most straightforward thing instead?

Looking at older versions of this code is good for the soul 😃   Neil's original was a joy to work with, subtle algorithms but coded in completely straightforward ways.  I dare say I improved some on that, by introducing a concrete notion of "state", with well defined transitions that were trivial to implement and query.  Pointers were always pointers, ints were always ints ...

But the gc header struct had 3 members then (next, prev, and refcount copy).  All the pain was introduced to get rid of the header's refcount member, and almost nothing was straightforward anymore.  It's not a trade-off I would have made, but it's certainly defensible.

In any case, the ubiquitous bit fiddling now also slowed the code.  I don't at all mind making it a bit slower still to preserve _some_ sanity 😜

pablogsal:
@tim-one Ok, after some design iterations, I am more or less happy with the shape of 
40f8cec. This commit removes the `check_garbage` function as per our discussion and encapsulates all the work for handling resurrecting objects. I think is now a bit more clear and makes things a bit more sane :)

Modules/gcmodule.c
```diff
@@ -1003,6 +1000,90 @@ show_stats_each_generations(struct _gc_runtime_state *state)
...
+       * After this function 'unreachable' must not be used anymore and 'still_unreachable'
+         will contain the objects that did not resurrect.
+
+       * The "still_unreachable" list must be uninitialized (this function calls
+         gc_list_init over 'still_unreachable').
+
+IMPORTANT: After a call to this function, the 'still_unreachable' set will have the
+PREV_MARK_COLLECTING set, but the objects in this set are going to be removed (and
+'delete_garbage' handles special cases that will go back to 'old_generation') so
+the removal of the flag can be skipped to avoid extra iteration. */
```
tim-one: Please remove the parenthetical comment:  there are no special cases in `delete_garbage`, and it's a Bad Thing that it ever moves anything that's been `tp_clear`'ed to a generation list.  It should move cleared things to an internal list instead, and complain at the end if that list hasn't been emptied.  But that's for a different PR.

Modules/gcmodule.c
```diff
@@ -1003,6 +1000,90 @@ show_stats_each_generations(struct _gc_runtime_state *state)
...
+{
+    // Remove the PREV_MASK_COLLECTING from unreachable
+    // to prepare it for a new call to 'deduce_unreachable'
+    gc_list_clear_collecting(unreachable);
+
+    // After the call to deduce_unreachable, the 'still_unreachable' set will
+    // have the PREV_MARK_COLLECTING set, but the objects are going to be
+    // removed (and 'delete_garbage' handles special cases that will go
+    // back to 'old_generation') so we can not remove the flag to avoid
+    // extra iteration.
```
tim-one: As above, the parenthetical comment is bogus.

The tail end of the sentence is a bit strained.  How about "so we can skip the expense of clearing the flag"?  Me, I'd skip the explanation and clear the flag regardless.  It takes non-trivial effort to find all the relevant functions and macros to convince yourself it really _is_ safe to skip clearing it.  Brittle.

tim-one (commented):
Looking good!  Left some last minor comments.

Modules/gcmodule.c
pablogsal: You are totally right, I have removed the comment in 07d3f8a.

After this PR we can go for `delete_garbage` :)

Modules/gcmodule.c
pablogsal: Removed the comment in 07d3f8a. I propose to clear the flag regardless in the PR to simplify/fix `delete_garbage` (if it makes sense).

tim-one (approved):
I think it's good to ship.  Nice work!

pablogsal:
Thank you very much for the thorough review! :)

PR #16688, opened 2019-10-10
bpo-38425: Remove ‘res’ may be used uninitialized warning

corona10:
I' ve noticed this warning on this environment.

```
Python/Python-ast.c: In function ‘PyAST_obj2mod’:
Python/Python-ast.c:10253:12: warning: ‘res’ may be used uninitialized in this function [-Wmaybe-uninitialized]
```
CentOS Linux release 7.6.1810
gcc (GCC) 4.8.5 20150623 (Red Hat 4.8.5-36)

https://bugs.python.org/issue38425

### DIFF
--- a/Misc/NEWS.d/next/Core and Builtins/2019-10-10-14-04-26.bpo-38425.xB59iJ.rst
+++ b/Misc/NEWS.d/next/Core and Builtins/2019-10-10-14-04-26.bpo-38425.xB59iJ.rst
@@ -1,2 +1,0 @@
-Remove ‘res’ may be used uninitialized warning. (Contributed by Dong-hee Na
-in :issue:`38425`.)
--- a/Misc/NEWS.d/next/Core and Builtins/2019-10-10-14-04-26.bpo-38425.xB59iJ.rst
+++ b/Misc/NEWS.d/next/Core and Builtins/2019-10-10-14-04-26.bpo-38425.xB59iJ.rst
@@ -1,0 +1,2 @@
+Remove ‘res’ may be used uninitialized warning. (Contributed by Dong-hee Na
+in :issue:`38425`.)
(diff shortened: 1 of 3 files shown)


corona10:
@vstinner 
As far as I know, you are a RedHat expert.
So I request the review of this PR to you

Misc/NEWS.d/next/Core and Builtins/2019-10-10-14-04-26.bpo-38425.xB59iJ.rst
```diff
@@ -0,0 +1,2 @@
+Remove ‘res’ may be used uninitialized warning. (Contributed by Dong-hee Na
+in :issue:`38425`.)
```
vstinner: Please remove the NEWS entry. Fixing a compiler warning doesn't deserve to be notificed to users.

Parser/asdl_c.py
```diff
@@ -1197,7 +1197,7 @@ class PartingShots(StaticVisitor):
 /* mode is 0 for "exec", 1 for "eval" and 2 for "single" input */
 mod_ty PyAST_obj2mod(PyObject* ast, PyArena* arena, int mode)
 {
-    mod_ty res;
+    mod_ty res = NULL;
```
vstinner: You may move the declaration just before it's used, obj2ast_mod() call.

corona10:
@vstinner 
I'd like to add backport label 3.8 / 3.7.
What do you think?

vstinner:
> As far as I know, you are a RedHat expert. So I request the review of this PR to you

This issue is specific to old versions of GCC, not to RHEL/Centos.

> @vstinner I'd like to add backport label 3.8 / 3.7. What do you think?

I think that GCC 4.8.5 is stupid and that GCC 9.2.1 on Fedora 30 is smarter: it doesn't emit this false alarm :-) I'm fine with backporting the change to 3.7 and 3.8.

PR #16695, opened 2019-10-10
bpo-38353: Fix typos in calculate_argv0_path_framework()

vstinner:
[bpo-38353](https://bugs.python.org/issue38353), [bpo-38429](https://bugs.python.org/issue38429): Fix typos introduced by commit
c02b41b1fb115c87693530ea6a480b2e15460424 in
calculate_argv0_path_framework() of getpath.c.

### DIFF
--- a/Modules/getpath.c
+++ b/Modules/getpath.c
@@ -1126,7 +1126,7 @@ calculate_argv0_path_framework(PyCalculatePath *calculate, _PyPathConfig *pathco
     }
 
     reduce(parent);
-    wchar_t *lib_python = joinpath2(path, calculate->lib_python);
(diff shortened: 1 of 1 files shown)


vstinner:
I never tried Python with --enable-framework. It updated my PR to fix more issues. The compilation now pass.

But when I run Python from the source directory, I get this error:

```
$ ./python.exe -m test test_embed  -v
dyld: Library not loaded: /Library/Frameworks/Python.framework/Versions/3.9/Python
  Referenced from: /Users/haypo/python/master/./python.exe
  Reason: image not found
Abort trap: 6
```

I guess that it's normal, Python should be installed. Or maybe I should use --enable-shared as well.

PR #16699, opened 2019-10-10
Remove extra full stop from function header of `get_operator`

hansrajdas:


Python/ast.c
```diff
@@ -958,7 +958,7 @@ PyAST_FromNode(const node *n, PyCompilerFlags *flags, const char *filename_str,
 
 }
 
-/* Return the AST repr. of the operator represented as syntax (|, ^, etc.)
+/* Return the AST repr of the operator represented as syntax (|, ^, etc.)
```
brandtbucher: Pretty sure "repr." here is short for "representation". This function has nothing to do with Python's `repr()`/`__repr__()`...

brandtbucher (changes requested):
Not sure about this one @hansrajdas. See below:

Python/ast.c
hansrajdas: - Agreed, repr expanded to representation
- Extra line between function header and function body also deleted

brandtbucher:
I personally think that the comment (and whitespace) is fine as-is. Sorry.

hansrajdas:
> I personally think that the comment (and whitespace) is fine as-is. Sorry.

Ok, Shall we close this PR?

brandtbucher:
Yes, I would recommend closing it.

hansrajdas:
> Yes, I would recommend closing it.

Sure, Thank you!

PR #16705, opened 2019-10-10
bpo-38436: Improved performance for list addition.

brandtbucher:
This PR adds a fast path for `BINARY_ADD` instructions involving two lists, where the left list has a refcount of exactly 1.

In this case, we instead do a `PySequence_InPlaceConcat` operation. This has the affect of avoiding quadratic complexity for list summations, by keeping only one intermediate result and extending it in-place.

For (potentially large) lists:

```py
a + b + c + d ...
```

Currently executed as:

```py
tmp = a + b
tmp = tmp + c
tmp = tmp + d
...
```

With this change:

```py
tmp = a + b
tmp += c
tmp += d
...
```

~Here are the `pyperformance` results (optimizations enabled):~

**EDIT: there are better measurements in the BPO discussion.**

This is related to my earlier work in [bpo-36229](https://bugs.python.org/issue36229), however this is a much less invasive change that's limited to `ceval.c`, where our knowledge of context is much better. 

https://bugs.python.org/issue38436

### DIFF
--- a/Misc/NEWS.d/next/Core and Builtins/2019-10-10-19-20-24.bpo-38436.LfVbc-.rst
+++ b/Misc/NEWS.d/next/Core and Builtins/2019-10-10-19-20-24.bpo-38436.LfVbc-.rst
@@ -1,0 +1,1 @@
+Improved the performance of repeated/literal :class:`list` addition.
--- a/Python/ceval.c
+++ b/Python/ceval.c
@@ -1551,6 +1551,11 @@ main_loop:
                 sum = unicode_concatenate(tstate, left, right, f, next_instr);
                 /* unicode_concatenate consumed the ref to left */
             }
+            else if (PyList_CheckExact(left) && PyList_CheckExact(right)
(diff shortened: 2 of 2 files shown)


pablogsal:
Where the benchmarks executed with CPU isolation? I am a bit suspicious of the 

```
- pickle_list: 3.63 us +- 0.08 us -> 3.78 us +- 0.10 us: 1.04x slower (+4%)
```

If not, can you repeat them with CPU isolation and affinity? Check [this](https://vstinner.github.io/perf-visualize-system-noise-with-cpu-isolation.html) for more info

brandtbucher:
@pablogsal No isolation (I'm on a MacBook) but I can rerun with affinity on two cores.

pablogsal:
> @pablogsal No isolation (I'm on a MacBook) but I can rerun with affinity on two cores.

I can try to run the benchmarks on my machine or the speed.python.org server.

brandtbucher:
Thanks @pablogsal - that would probably be best!

PR #16707, opened 2019-10-10
bpo-38437: Activate GC_DEBUG when PY_DEBUG is set

pablogsal:
https://bugs.python.org/issue38437

### DIFF
--- a/Modules/gcmodule.c
+++ b/Modules/gcmodule.c
@@ -37,7 +37,10 @@ module gc
 [clinic start generated code]*/
 /*[clinic end generated code: output=da39a3ee5e6b4b0d input=b5c9690ecc842d79]*/
 
-#define GC_DEBUG (0)  /* Enable more asserts */
+
+#ifdef Py_DEBUG
+#  define GC_DEBUG
+#endif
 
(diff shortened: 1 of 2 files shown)


Modules/gcmodule.c
```diff
@@ -37,7 +37,7 @@ module gc
 [clinic start generated code]*/
 /*[clinic end generated code: output=da39a3ee5e6b4b0d input=b5c9690ecc842d79]*/
 
-#define GC_DEBUG (0)  /* Enable more asserts */
+#define GC_DEBUG (Py_DEBUG)  /* Enable more asserts */
```
vstinner: I don't think that such code works. Py_DEBUG should not be defined in release mode.

You should use something like:

```
#ifdef Py_DEBUG
#  define GC_DEBUG
#endif 
```

And later use `#ifdef GC_DEBUG`.

pablogsal: Done in 40d719a

tim-one (approved):
LGTM!

pablogsal:
Thanks for the review!

PR #16714, opened 2019-10-11
Remove strict-aliasing rules errors on gcc 4.8.5

corona10:
Remove below error on 
CentOS Linux release 7.6.1810
gcc (GCC) 4.8.5 20150623 (Red Hat 4.8.5-36)

````bash
Objects/dictobject.c:1136:5: warning: dereferencing type-punned pointer will break strict-aliasing rules [-Wstrict-aliasing]
     PyDictKeyEntry *ep = &DK_ENTRIES(mp->ma_keys)[0];
````

### DIFF
--- a/Objects/dictobject.c
+++ b/Objects/dictobject.c
@@ -1133,7 +1133,7 @@ insert_to_emptydict(PyDictObject *mp, PyObject *key, Py_hash_t hash,
     MAINTAIN_TRACKING(mp, key, value);
 
     size_t hashpos = (size_t)hash & (PyDict_MINSIZE-1);
-    PyDictKeyEntry *ep = &DK_ENTRIES(mp->ma_keys)[0];
+    PyDictKeyEntry *ep = DK_ENTRIES(mp->ma_keys);
     dictkeys_set_index(mp->ma_keys, hashpos, 0);
     ep->me_key = key;
     ep->me_hash = hash;

### CODE REVIEW

**Comment by methane** [comment]
I don't have gcc 4.8.5 environment but your patch looks good to me.

**Comment by corona10** [comment]
@methane  
Thanks for the review!

**Comment by corona10** [comment]
@methane Can we open the backport patch for 3.8 3.7?

**Comment by methane** [comment]
@corona10 I think this is safe to backport.  But I want to wait for buildbots before backporting.

**Comment by corona10** [comment]
@methane Got it!

**Comment by vstinner** [comment]
"But I want to wait for buildbots before backporting."

Thank you!

**Comment by vstinner** [comment]
````bash
Objects/dictobject.c:1136:5: warning: dereferencing type-punned pointer will break strict-aliasing rules [-Wstrict-aliasing]
     PyDictKeyEntry *ep = &DK_ENTRIES(mp->ma_keys)[0];
````

Hum, there are other places with similar code:

```
delitem_common():

    PyDictKeyEntry *ep;
    ...
    ep = &DK_ENTRIES(mp->ma_keys)[ix];

_PyDict_Pop_KnownHash:

    PyDictKeyEntry *ep;
    ...
    ep = &DK_ENTRIES(mp->ma_keys)[ix];

dict_equal:

        PyDictKeyEntry *ep = &DK_ENTRIES(a->ma_keys)[i];
```

Why only insert_to_emptydict() required to be modified?

I'm not sure that the magic DK_ENTRIES() macro respects strict aliasing.

```
#define DK_ENTRIES(dk) \
    ((PyDictKeyEntry*)(&((int8_t*)((dk)->dk_indices))[DK_SIZE(dk) * DK_IXSIZE(dk)]))
```

_dictkeysobject structure ends with:

```
    char dk_indices[];  /* char is required to avoid strict aliasing. */
```

_dictkeysobject ends with the following field, but the field is **not** declared in the structure:

```
PyDictKeyEntry dk_entries[dk_usable];
```

Since it's not declared, I'm not sure that the compiler is able to align it properly.

PR #16717, opened 2019-10-11
bpo-37415: Fix stdatomic.h header check for ICC compiler

vstinner:
Fix stdatomic.h header check for ICC compiler: the ICC implementation
lacks atomic_uintptr_t type which is needed by Python.

https://bugs.python.org/issue37415

### DIFF
--- a/Misc/NEWS.d/next/Build/2019-10-11-15-32-58.bpo-37415.D9RXrq.rst
+++ b/Misc/NEWS.d/next/Build/2019-10-11-15-32-58.bpo-37415.D9RXrq.rst
@@ -1,0 +1,2 @@
+Fix stdatomic.h header check for ICC compiler: the ICC implementation lacks
+atomic_uintptr_t type which is needed by Python.
--- a/configure
+++ b/configure
@@ -782,7 +782,6 @@ infodir
 docdir
 oldincludedir
 includedir
-runstatedir
 localstatedir
 sharedstatedir
 sysconfdir
@@ -896,7 +895,6 @@ datadir='${datarootdir}'
 sysconfdir='${prefix}/etc'
 sharedstatedir='${prefix}/com'
 localstatedir='${prefix}/var'
-runstatedir='${localstatedir}/run'
 includedir='${prefix}/include'
 oldincludedir='/usr/include'
 docdir='${datarootdir}/doc/${PACKAGE_TARNAME}'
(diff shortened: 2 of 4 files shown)


vstinner:
I removed ATOMIC_VAR_INIT() check from configure, since it's not used in Python.

adamjstewart:
This should also fix https://bugs.python.org/issue35473

vstinner:
@adamjstewart: Can you test this PR with ICC?

adamjstewart:
I'm still working on getting access to our cluster, but once I have access I can definitely test this PR!

adamjstewart:
I assume the runstatedir portions of the patch aren't actually needed to fix the intel build?

vstinner:
> I assume the runstatedir portions of the patch aren't actually needed to fix the intel build?

I'm running "autoconf". Depending who run "autoconf" to generate the latest configure in script and depending on my autoconf version, the runstatedir thing may or may not change. I never understood how to get a reliable output using autotools.

adamjstewart:
This patch seems to be working so far. I tried running the unit tests and only saw a few failures:
```
Ran 327 tests in 205.461s

OK (skipped=30)
4 tests failed again:
    test_asyncio test_distutils test_gdb test_imaplib

399 tests OK.

4 tests failed:
    test_asyncio test_distutils test_gdb test_imaplib

13 tests skipped:
    test_curses test_devpoll test_kqueue test_msilib test_ossaudiodev
    test_startfile test_tix test_tk test_ttk_guionly test_winconsoleio
    test_winreg test_winsound test_zipfile64

7 re-run tests:
    test_asyncio test_distutils test_gdb test_imaplib
    test_multiprocessing_fork test_multiprocessing_forkserver
    test_multiprocessing_spawn

Total duration: 20 min 47 sec
Tests result: FAILURE then FAILURE
make: *** [test] Error 2
```
I'm going to try to install anyway.

adamjstewart:
The install succeeded! With the minimal patch found [here](https://github.com/spack/spack/pull/10319/files#diff-2f7bbaa997a6ebefcf86b380392b0d87), I was able to build Python 3.7.4 with Intel 18.0.3.222 on Cray CNL5. The patch successfully applies to Python 3.6.7-3.6.8, 3.7.1-3.7.4, and 3.8.0 release tarballs. We may want to backport this to Python 3.6 if it's still supported. I believe this bug was likely introduced in Python 3.6.7, 3.7.1, and 3.8.0. Thanks for your help @vstinner!

vstinner:
> 4 tests failed: test_asyncio test_distutils test_gdb test_imaplib

That's unrelated to https://bugs.python.org/issue37415

https://bugs.python.org/issue37415 is only about a compilation failure when using ICC.

PR #16724, opened 2019-10-12
bpo-38449: Revert "bpo-22347: Update mimetypes.guess_type to allow oper parsing of URLs (GH-15522)"

maxking:
This reverts commit 87bd2071c756188b6cd577889fb1682831142ceb.

### DIFF
--- a/Lib/mimetypes.py
+++ b/Lib/mimetypes.py
@@ -114,8 +114,7 @@ class MimeTypes:
         but non-standard types.
         """
         url = os.fspath(url)
-        p = urllib.parse.urlparse(url)
-        scheme, url = p.scheme, p.path
+        scheme, url = urllib.parse._splittype(url)
         if scheme == 'data':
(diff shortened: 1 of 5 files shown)


yarikoptic (commented):
Could a unit test been added to verify (current and future) correct operation on a file name with `;` and may be other tricky symbols legit to be in a file name but not a URL?

maxking:
@yarikoptic This is just reverting the previous change which caused the regression so release managers can pick this up for the final releases.

Fixes and more unit tests should be best added in the next PR which should fix the original issue and the issue raised in bpo-38449.

ned-deily (changes requested):
LGTM, except it needs a news (blurb) entry about the revert since it introduces a change between 3.8.0rc1/3.7.5rc1 and 3.8.0final/3.7.5final.

maxking:

ned-deily (approved):

PR #16726, opened 2019-10-12
bpo-38417: Add umask support to subprocess

gpshead:
On POSIX systems, allow the umask to be set in the child process
before we exec.

https://bugs.python.org/issue38417

### DIFF
--- a/Lib/test/test_subprocess.py
+++ b/Lib/test/test_subprocess.py
@@ -1896,9 +1896,6 @@ class POSIXProcessTestCase(BaseTestCase):
 
     @unittest.skipUnless(hasattr(os, 'umask'), 'umask() is not available.')
     def test_umask(self):
-        # we use mkdtemp in the next line to create an empty directory
-        # under our exclusive control; from that, we can invent a pathname
-        # that we _know_ won't exist.  This is guaranteed to fail.
         tmpdir = None
         try:
             tmpdir = tempfile.mkdtemp()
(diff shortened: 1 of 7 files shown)


Lib/test/test_subprocess.py
```diff
@@ -1894,6 +1894,24 @@ def test_extra_groups_error(self):
         with self.assertRaises(ValueError):
             subprocess.check_call([sys.executable, "-c", "pass"], extra_groups=[])
 
+    @unittest.skipUnless(hasattr(os, 'umask'), 'umask() is not available.')
```
eryksun: This test should be skipped in Windows, which implements [`umask`](https://docs.microsoft.com/en-us/cpp/c-runtime-library/reference/umask?view=vs-2019) only to control the default value of the readonly file attribute for files created via `_[w]open` in the current process.

(I know what the Windows CRT does here is nonsense. File attributes are not file permissions. No one can be permitted to write to a readonly file. It's a fundamental attribute of the file that its data cannot be modified. Only `os.access` should take into account the readonly attribute, just as this works for the [i]mmutable file attribute in Linux. We have to live with other people's misinformed design decisions.)

gpshead: Thanks!  Interesting. I didn't realize os.umask was present on Windows but was wondering what would happen there as it'd be a different concept.  Also, this passed the CI regardless... I probably chose a umask that matched what stat returned there?  anyways, skipped on windows now.

eryksun: Currently `os.stat(name).st_mode & 0o666` in Windows will be either 0o666 or 0o444, depending on whether the readonly file attribute is set (see `attributes_to_mode` in "Python/fileutils.c"). For `stat`, the execute bit is also set based on the file extension for file types that are typically executable via `CreateProcessW`: ".exe", ".com", ".bat", and ".cmd". 

This ranges from completely wrong to smelly hack. At some point, real support for file permissions may (should) be implemented, by copying common approaches from other systems such as Cygwin and WSL.

PR #15687, opened 2019-09-05
[3.7] bpo-22347: Update mimetypes.guess_type to allow proper parsing of URLs (GH-15522)

corona10:
https://bugs.python.org/issue22347
(cherry picked from commit 87bd2071c756188b6cd577889fb1682831142ceb)

https://bugs.python.org/issue22347

Automerge-Triggered-By: @maxking

PR #16727, opened 2019-10-12
[3.7] bpo-38449: Revert "bpo-22347: Update mimetypes.guess_type to allow oper parsing of URLs (GH-15685)" (GH-16724)

maxking:
Reverts GH-15687 which caused the issue.

https://bugs.python.org/issue22347


corona10 (approved):
LGTM

ned-deily (approved):

PR #16729, opened 2019-10-12
bpo-38449: Add URL delimiters test cases

corona10:

### DIFF
--- a/Lib/test/test_mimetypes.py
+++ b/Lib/test/test_mimetypes.py
@@ -51,6 +51,12 @@ class MimeTypesTestCase(unittest.TestCase):
         eq(self.db.guess_type('foo.xul', strict=False), ('text/xul', None))
         eq(self.db.guess_extension('image/jpg', strict=False), '.jpg')
 
+    def test_tricky_cases(self):
+        # bpo-38449: Tricky cases should be handled also.
+        eq = self.assertEqual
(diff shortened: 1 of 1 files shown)


Lib/test/test_mimetypes.py
```diff
@@ -51,6 +51,12 @@ def test_non_standard_types(self):
         eq(self.db.guess_type('foo.xul', strict=False), ('text/xul', None))
         eq(self.db.guess_extension('image/jpg', strict=False), '.jpg')
 
+    def test_tricky_cases(self):
```
maxking: Can we instead call it `test_filename_with_url_delimiters` to be more explicit ?

It would be nice to explain a bit why this is tricky, i.e., They would have different mime types if interpreted as URL as compared to when interpreted as filename because of the semicolon.

maxking (changes requested):
I think semicolon is only one of the tricky case, there could be others which would result in a similar error case, especially the delimiters for various parts of a URL:

      scheme://netloc/path;parameters?query#fragment

So, I *think*, `;`, `?` and `#` would have similar behaviour when parsing using `urllib.parse.urlparse` for determining the mimetype.

corona10:
Thanks for the review.

corona10:
@maxking gentle ping~

PR #16730, opened 2019-10-12
[docs] Announce the change in the CancelledError inheritance

pgjones:
This is a fairly noticeable change that requires adjustments in existing asyncio code. I think it should be announced.

@1st1 Probably best for you to decide if this is the correct way to announce this.


1st1 (approved):

1st1:
Thank you!

PR #16736, opened 2019-10-12
bpo-38456: Use /bin/true in test_subprocess

gpshead:
Instead of sys.executable, "-c", "pass" or "import sys; sys.exit(0)"
use /bin/true when it is available.  On a reasonable machine this
shaves up to two seconds wall time off the otherwise ~40sec execution
on a --with-pydebug build.  It should be more notable on many
buildbots or overloaded slower I/O systems (CI, etc).

### DIFF
--- a/Lib/test/test_subprocess.py
+++ b/Lib/test/test_subprocess.py
@@ -54,6 +54,14 @@ NONEXISTING_CMD = ('nonexisting_i_hope',)
 # Ignore errors that indicate the command was not found
 NONEXISTING_ERRORS = (FileNotFoundError, NotADirectoryError, PermissionError)
 
+ZERO_RETURN_CMD = (sys.executable, '-c', 'pass')
+
+
+def setUpModule():
+    if os.access('/bin/true', os.X_OK):
+        global ZERO_RETURN_CMD
+        ZERO_RETURN_CMD = ('/bin/true',)  # Faster than Python startup.
+
 
 class BaseTestCase(unittest.TestCase):
     def setUp(self):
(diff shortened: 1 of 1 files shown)


Lib/test/test_subprocess.py
```diff
@@ -54,6 +54,14 @@
 # Ignore errors that indicate the command was not found
 NONEXISTING_ERRORS = (FileNotFoundError, NotADirectoryError, PermissionError)
 
+ZERO_RETURN_CMD = (sys.executable, '-c', 'pass')
+
+
+def setUpModule():
+    if os.access('/bin/true', os.X_OK):
```
pablogsal: Should we try to use `shutil.which("true")` instead or in case `/bin/true` is not present? In my system I only have `/usr/bin/true` for example.

pablogsal (approved):
One comment, otherwise LGTM

Thanks for working on this :tada:

Lib/test/test_subprocess.py
gpshead: sure, why not.  done.

pablogsal:
We have a failure in FreeBSD:

```exception
ERROR: setUpModule (test.test_subprocess)
  File "C:\buildbot.python.org\3.x.kloth-win64\build\lib\test\test_subprocess.py", line 62, in setUpModule
    if (os.access(shell_true, os.X_OK) and
TypeError: access: path should be string, bytes or os.PathLike, not NoneType
Ran 0 tests in 0.014s
```

I created https://github.com/python/cpython/pull/16739 to fix the error

PR #16737, opened 2019-10-12
[3.8] bpo-38456: Use /bin/true in test_subprocess

gpshead:
Instead of sys.executable, "-c", "pass" or "import sys; sys.exit(0)"
use /bin/true when it is available.  On a reasonable machine this
shaves up to two seconds wall time off the otherwise ~40sec execution
on a --with-pydebug build.  It should be more notable on many
buildbots or overloaded slower I/O systems (CI, etc).

(cherry picked from commit 67b93f80c764bca01c81c989d74a99df208bea4d)


gpshead:
Applying this improvement to test_subprocess in 3.8 makes sense, but only after 3.8.0 is out the door.

PR #16739, opened 2019-10-13
bpo-38456: Handle the case when there is no 'true' command

pablogsal:


pablogsal:
I'm going to go ahead and merge this one as is very straightforward change to avoid a buildbot storm

gpshead:
haha, thanks!  I'll just include this in my 3.8 PR.

PR #16621, opened 2019-10-07
Grammar fixes to whatsnew/3.8.rst

akuchling:
I read through the 3.8 document and fixed various grammar issues.  Also, the library modules weren't quite in alphabetical order, and there was a duplicated idlelib/IDLE section (maybe a merging mishap?), so I rearranged them without making any other textual change in that commit.  This means the PR is likely easier to read if you go commit-by-commit instead of reading the entire diff.

PR #16745, opened 2019-10-13
Rebased version of what's new PR

akuchling:
Reworked version of PR https://github.com/python/cpython/pull/16621, taken against master and created afresh.  I changed it to use the correct accent for Giampaolo Rodolà's name.


pablogsal (approved):

terryjreedy (approved):
I checked that this is the same diff as before.

PR #16752, opened 2019-10-13
Misc gc code & comment cleanups.

tim-one:
validate_list:  there are two temp flags polluting pointers, but this checked only one.  Now it checks both, and verifies that the list head's pointers are not polluted.

move_unreachable: repaired incoherent comments.  Added new comments.  Cleared the pollution of the unreachable list head's 'next' pointer (it was expedient while the function was running, but there's no excuse for letting this damage survive the function's end).

### DIFF
--- a/Modules/gcmodule.c
+++ b/Modules/gcmodule.c
@@ -331,23 +331,37 @@ append_objects(PyObject *py_list, PyGC_Head *gc_list)
 
 #ifdef GC_DEBUG
 // validate_list checks list consistency.  And it works as document
-// describing when expected_mask is set / unset.
+// describing when flags are expected to be set / unset.
+// `head` must be a doubly-linked gc list, although it's fine (expected!) if
+// the prev and next pointers are "polluted" with flags.
+// What's checked:
+// - The `head` pointers are not polluted.
+// - The objects' prev pointers' PREV_MASK_COLLECTING flags are all
+//   `prev_value` (PREV_MASK_COLLECTING for set, 0 for clear).
+// - The objects' next pointers' NEXT_MASK_UNREACHABLE flags are all
+//   `next_value` (NEXT_MASK_UNREACHABLE for set, 0 for clear).
+// - The prev and next pointers are mutually consistent.
 static void
-validate_list(PyGC_Head *head, uintptr_t expected_mask)
+validate_list(PyGC_Head *head, uintptr_t prev_value, uintptr_t next_value)
 {
+    assert((head->_gc_prev & PREV_MASK_COLLECTING) == 0);
+    assert((head->_gc_next & NEXT_MASK_UNREACHABLE) == 0);
     PyGC_Head *prev = head;
     PyGC_Head *gc = GC_NEXT(head);
     while (gc != head) {
-        assert(GC_NEXT(gc) != NULL);
-        assert(GC_PREV(gc) == prev);
-        assert((gc->_gc_prev & PREV_MASK_COLLECTING) == expected_mask);
+        PyGC_Head *trueprev = GC_PREV(gc);
+        PyGC_Head *truenext = (PyGC_Head *)(gc->_gc_next  & ~NEXT_MASK_UNREACHABLE);
+        assert(truenext != NULL);
+        assert(trueprev == prev);
+        assert((gc->_gc_prev & PREV_MASK_COLLECTING) == prev_value);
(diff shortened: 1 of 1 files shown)


Modules/gcmodule.c
```diff
@@ -1021,15 +1042,15 @@ show_stats_each_generations(struct _gc_runtime_state *state)
...
+
 IMPORTANT: This function leaves 'unreachable' with the NEXT_MASK_UNREACHABLE
 flag set but it does not clear it to skip unnecessary iteration. Before the
 flag is cleared (for example, by using 'clear_unreachable_mask' function or
 by a call to 'move_legacy_finalizers'), the 'unreachable' list is not a normal
 list and we can not use most gc_list_* functions for it. */
 static inline void
 deduce_unreachable(PyGC_Head *base, PyGC_Head *unreachable) {
-    validate_list(base, 0);
+    validate_list(base, 0, 0);
```
pablogsal, line 1053: Although I think is clear already, I wonder if we can create a macro for NO flags so these calls look like:

```
validate_list(base, GC_NO_FLAG, GC_NO_FLAG)
```

or similar. I was thinking about this because now that there are two numbers, having two 0's on the call can be confusing at first. Is a very minor thing, so feel free to ignore the suggestion :)

pablogsal (approved):
This looks great! I just left some minor comments, but feel free to go ahead with the merge :)

Modules/gcmodule.c
```diff
@@ -331,23 +331,37 @@ append_objects(PyObject *py_list, PyGC_Head *gc_list)
...
+// What's checked:
+// - The `head` pointers are not polluted.
+// - The objects' prev pointers' PREV_MASK_COLLECTING flags are all
+//   `prev_value` (PREV_MASK_COLLECTING for set, 0 for clear).
+// - The objects' next pointers' NEXT_MASK_UNREACHABLE flags are all
+//   `next_value` (NEXT_MASK_UNREACHABLE for set, 0 for clear).
+// - The prev and next pointers are mutually consistent.
 static void
-validate_list(PyGC_Head *head, uintptr_t expected_mask)
+validate_list(PyGC_Head *head, uintptr_t prev_value, uintptr_t next_value)
```
pablogsal, line 345: I would suggest that instead of `prev_value` and `next_value` we use something that makes clear that this is the value of the flags/mask/...

Modules/gcmodule.c
```diff
@@ -434,6 +448,9 @@ visit_reachable(PyObject *op, PyGC_Head *reachable)
     const Py_ssize_t gc_refs = gc_get_refs(gc);
 
     // Ignore untracked objects and objects in other generation.
+    // This also skips objects "to the left" of the current position in
+    // move_unreachable's scan of the 'young' list - they've already been
+    // traversed, and no longer have the PREV_MASK_COLLECTING flag.
```
pablogsal, line 453: Great addition! This will make life much easier to people learning the algorithm by just reading the code!

Modules/gcmodule.c
```diff
@@ -546,8 +564,9 @@ move_unreachable(PyGC_Head *young, PyGC_Head *unreachable)
             PyGC_Head *last = GC_PREV(unreachable);
             // NOTE: Since all objects in unreachable set has
             // NEXT_MASK_UNREACHABLE flag, we set it unconditionally.
-            // But this may set the flat to unreachable too.
-            // move_legacy_finalizers() should care about it.
+            // But this may pollute the unreachable list head's 'next' pointer
+            // too.  That's semantically senseless but expedient here - the
```
pablogsal:
```suggestion
            // too. That's semantically senseless but expedient here - the
```

Modules/gcmodule.c
tim-one: I actually find that less clear, but won't object if you do it 😉 .  Clearer still would be, e.g.,
```c
#define VALIDATE_NO_FLAGS(list) validate_list(list, 0, 0)
```
plus 3 more for the other possibilities   But in this file especially, I'm sick to death of having to track down what macros expand to 😉

Modules/gcmodule.c
tim-one: That was explained in the new comments before the function, which exhaustively spelled out the meaningful possible values.  If I could have thought of better names, I already would have :wink: - nothing substitutes for reading the docs.

PR #16755, opened 2019-10-13
bpo-38469: Handle named expression scope whith global/nonlocal keywords

pablogsal:
https://bugs.python.org/issue38469

### DIFF
--- a/Lib/test/test_named_expressions.py
+++ b/Lib/test/test_named_expressions.py
@@ -476,7 +476,6 @@ spam()"""
         global GLOBAL_VAR
         def f():
             global GLOBAL_VAR
-            GLOBAL_VAR = None
             [GLOBAL_VAR := sentinel for _ in range(1)]
             self.assertEqual(GLOBAL_VAR, sentinel)
         try:
@@ -500,7 +499,6 @@ spam()"""
             nonlocal_var = None
             def g():
                 nonlocal nonlocal_var
-                nonlocal_var = None
                 [nonlocal_var := sentinel for _ in range(1)]
             g()
             self.assertEqual(nonlocal_var, sentinel)
@@ -511,7 +509,6 @@ spam()"""
         def f():
             nonlocal_var = None
             def g():
-                nonlocal_var = None
                 [nonlocal_var := sentinel for _ in range(1)]
             g()
             self.assertEqual(nonlocal_var, None)
--- a/Lib/test/test_named_expressions.py
+++ b/Lib/test/test_named_expressions.py
@@ -1,6 +1,7 @@
 import os
 import unittest
 
+GLOBAL_VAR = None
 
 class NamedExpressionInvalidTest(unittest.TestCase):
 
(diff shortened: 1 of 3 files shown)


Lib/test/test_named_expressions.py
```diff
@@ -470,5 +471,52 @@ def test_named_expression_variable_reuse_in_comprehensions(self):
                 self.assertEqual(ns["x"], 2)
                 self.assertEqual(ns["result"], [0, 1, 2])
 
+    def test_named_expression_global_scope(self):
+        sentinel = object()
+        global GLOBAL_VAR
+        def f():
+            global GLOBAL_VAR
+            GLOBAL_VAR = None
```
gvanrossum: What happens without this initialization? The example in the [bpo issue](https://bugs.python.org/issue38469) doesn't have this. Do we need another set of tests where there's no other local assignment to the same (global or nonlocal) variable?

Python/symtable.c
```diff
@@ -1469,8 +1469,14 @@ symtable_extend_namedexpr_scope(struct symtable *st, expr_ty e)
 
         /* If we find a FunctionBlock entry, add as NONLOCAL/LOCAL */
         if (ste->ste_type == FunctionBlock) {
-            if (!symtable_add_def(st, target_name, DEF_NONLOCAL))
-                VISIT_QUIT(st, 0);
+            long target_in_scope = _PyST_GetSymbol(ste, target_name);
+            if (target_in_scope & DEF_GLOBAL){
```
gvanrossum: Style nit: space before `{`.

gvanrossum (commented):
Did you force-push anything? GitHub told me to refresh the PR but there was no new commit.

Also, alas, I don't feel comfortable approving this, I don't know the symtable code well enough any more. Sorry!

Lib/test/test_named_expressions.py
pablogsal: This was to avoid weird results when some other test modifies the global to some other thing but is really not needed. I will remove them to make it less confusing.

pablogsal:
> Did you force-push anything? GitHub told me to refresh the PR but there was no new commit.

Yeah, I force pushed a version just some minutes ago. Sorry for the confusion.

pablogsal:
> Also, alas, I don't feel comfortable approving this, I don't know the symtable code well enough any more. Sorry!

Not a problem, thanks for the review! :)

ncoghlan:
There should be existing test cases for this, so the question is "What's wrong with them"?

The testing style in the module is also to rely on exec(), rather than on the test module itself (otherwise a compile error will keep the entire test file from running)

ncoghlan:
Sorry, I retract that "exec" comment - it only applies to the tests that are explicitly checking for compilation errors

ncoghlan:
And the missing test case is also genuinely missing rather than existing in a flawed form (the global declaration test doesn't use a comprehension, the comprehension tests don't use a global)

Misc/NEWS.d/next/Core and Builtins/2019-10-13-23-41-38.bpo-38469.9kmuQj.rst
```diff
@@ -0,0 +1,2 @@
+Fixed a bug where the scope of named expressions was not being resolved
+correctly in the present of the *global* keyword. Patch by Pablo Galindo.
```
ncoghlan:
```suggestion
correctly in the presence of the *global* keyword. Patch by Pablo Galindo.
```

ncoghlan (approved):

pablogsal:
Thanks @ncoghlan for the thorough review!

PR #16780, opened 2019-10-14
bpo-36338: urllib.urlparse rejects invalid IPv6 addresses

vstinner:
The urllib.urlparse module now rejects invalid IPv6 addresses and
invalid port numbers when parsing an URL.

https://bugs.python.org/issue36338

### DIFF
--- a/Lib/test/test_urlparse.py
+++ b/Lib/test/test_urlparse.py
@@ -450,6 +450,7 @@ class UrlParseTestCase(unittest.TestCase):
         self.checkJoin('a', 'b', 'b')
 
     def test_RFC2732(self):
+        # IPv6 address
         str_cases = [
             ('http://Test.python.org:5432/foo/', 'test.python.org', 5432),
             ('http://12.34.56.78:5432/foo/', '12.34.56.78', 5432),
@@ -481,7 +482,8 @@ class UrlParseTestCase(unittest.TestCase):
             ('http://[::12.34.56.78]:/foo/', '::12.34.56.78', None),
             ('http://[::ffff:12.34.56.78]:/foo/',
              '::ffff:12.34.56.78', None),
-            ]
+            ('http://[::]/', '::', None),
+        ]
         def _encode(t):
             return t[0].encode('ascii'), t[1].encode('ascii'), t[2]
         bytes_cases = [_encode(x) for x in str_cases]
@@ -489,15 +491,52 @@ class UrlParseTestCase(unittest.TestCase):
             urlparsed = urllib.parse.urlparse(url)
             self.assertEqual((urlparsed.hostname, urlparsed.port) , (hostname, port))
 
+        # bpo-36338: reject invalid IPv6 addresses
         str_cases = [
-                'http://::12.34.56.78]/',
-                'http://[::1/foo/',
-                'ftp://[::1/foo/bad]/bad',
-                'http://[::1/foo/bad]/bad',
-                'http://[::ffff:12.34.56.78']
+            # invalid IPv6 address
+            'http://[abcd:x::]/',
+            'http://[::1]example.com/',
+            # missing "[" or "]"
+            'http://::12.34.56.78]/',
+            'http://[::1/foo/',
+            'ftp://[::1/foo/bad]/bad',
+            'http://[::1/foo/bad]/bad',
+            'http://[::ffff:12.34.56.78',
+            # double ":"
+            'http://[::1]::80/',
+            # invalid "[" or "]" characters
+            'http://[[::1]]/',
+            'http://[::1][]/',
+            'http://[::1]:[]/',
+            'http://good.com[bad.com]',
+            'http://benign.com\\[attacker.com]',
+        ]
         bytes_cases = [x.encode('ascii') for x in str_cases]
         for invalid_url in str_cases + bytes_cases:
             self.assertRaises(ValueError, urllib.parse.urlparse, invalid_url)
+
+    def test_user_passwd(self):
+        # bpo-33342: username and password with "[" and "]" characters
+        for username in ("", "user", "[", "user[", "us]er", "us%er"):
+            for password in ("", "password", "p[ass]word", "]", "pa%%word"):
+                for url_host, hostname, port in (
+                    ("localhost", "localhost", None),
+                    ("127.0.0.1", "127.0.0.1", None),
+                    ("[::1]", '::1', None),
+                    ("[::1]:", '::1', None),
+                    ("[::1%scope]", '::1%scope', None),
+                    ("[::1]:443", '::1', 443),
+                ):
+                    netloc = f"{username}:{password}@{url_host}"
+                    url = f"//{netloc}/path"
+                    with self.subTest(url=url):
+                        urlparsed = urllib.parse.urlparse(url)
+                        self.assertEqual(urlparsed.netloc, netloc)
+                        self.assertEqual(urlparsed.username, username)
+                        self.assertEqual(urlparsed.password, password)
+                        self.assertEqual(urlparsed.hostname, hostname)
+                        self.assertEqual(urlparsed.port, port)
+                        self.assertEqual(urlparsed.path, "/path")
 
     def test_urldefrag(self):
         str_cases = [
@@ -530,6 +569,12 @@ class UrlParseTestCase(unittest.TestCase):
         p = urllib.parse.urlsplit(b'http://[FE80::822a:a8ff:fe49:470c%tESt]:1234')
         self.assertEqual(p.hostname, b"fe80::822a:a8ff:fe49:470c%tESt")
         self.assertEqual(p.netloc, b'[FE80::822a:a8ff:fe49:470c%tESt]:1234')
+
+        for invalid_char in ('%', '[', ']'):
+            invalid_url = f'http://[::1%sco{invalid_char}pe]'
+            with self.subTest(invalid_url=invalid_url):
+                with self.assertRaises(ValueError):
+                    urllib.parse.urlparse(invalid_url)
 
     def test_urlsplit_attributes(self):
         url = "HTTP://WWW.PYTHON.ORG/doc/#frag"
@@ -620,9 +665,8 @@ class UrlParseTestCase(unittest.TestCase):
 
         # Verify an illegal port raises ValueError
         url = b"HTTP://WWW.PYTHON.ORG:65536/doc/#frag"
-        p = urllib.parse.urlsplit(url)
-        with self.assertRaisesRegex(ValueError, "out of range"):
-            p.port
+        with self.assertRaisesRegex(ValueError, "Port out of range 0-65535"):
+            urllib.parse.urlsplit(url)
 
     def test_attributes_bad_port(self):
         """Check handling of invalid ports."""
--- a/Lib/urllib/parse.py
+++ b/Lib/urllib/parse.py
@@ -167,13 +167,7 @@ class _NetlocResultMixinBase(object):
     def port(self):
         port = self._hostinfo[1]
         if port is not None:
-            try:
-                port = int(port, 10)
-            except ValueError:
-                message = f'Port could not be cast to integer value as {port!r}'
-                raise ValueError(message) from None
-            if not ( 0 <= port <= 65535):
-                raise ValueError("Port out of range 0-65535")
+            port = _validate_port(port)
         return port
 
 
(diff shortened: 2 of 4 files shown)


Lib/urllib/parse.py
```diff
@@ -388,17 +382,49 @@ def _splitparams(url):
...
+    scope = (parts[1] if len(parts) > 1 else None)
+
+    import ipaddress
+    try:
+        ipaddress.IPv6Address(ipv6)
+    except ipaddress.AddressValueError:
+        return False
+
+    if scope is not None:
+        if '%' in scope or '[' in scope or ']' in scope:
```
corona10: How about this?
```python
if scope is not None:
    for s in scope:
        if s in ('%', '[', ']'):
            return False
```

vstinner:
@corona10: I wrote the scope test differently for make it more readable.

corona10 (approved):
> @corona10: I wrote the scope test differently for make it more readable.

@vstinner Awesome!

vstinner:
@vadmium, @zooba, @tirkarthi: Would you mind to review this change? What do you think of my approach using ipaddress and excluding some characters from the IPv6 scope ("zone identifier")?

vstinner:
I modified my PR to also fix https://bugs.python.org/issue33342 : "urllib IPv6 parsing fails with special characters in passwords".

Lib/test/test_urlparse.py
```diff
@@ -620,9 +665,8 @@ def test_urlsplit_attributes(self):
 
         # Verify an illegal port raises ValueError
         url = b"HTTP://WWW.PYTHON.ORG:65536/doc/#frag"
-        p = urllib.parse.urlsplit(url)
-        with self.assertRaisesRegex(ValueError, "out of range"):
-            p.port
+        with self.assertRaisesRegex(ValueError, "Port out of range 0-65535"):
```
tirkarthi, line 668: This slightly backwards incompatible but I am okay with the intention of the PR in validating port during parsing instead of accessing the port attribute since it just means the URL is invalid and is known earlier.

Lib/urllib/parse.py
```diff
@@ -388,17 +382,48 @@ def _splitparams(url):
...
+        raise ValueError("Port out of range 0-65535: %r" % port_str)
+    return port
+
+def _check_ipv6_host(host):
+    # "[ipv6]" host: ensure that it starts with "[", ends with "]"
+    # and that ipv6 doesn't contain "[" nor "]".
+    if not(host.startswith('[') and host.endswith(']')):
+        return False
+
+    parts = host[1:-1].split('%', 1)
```
tirkarthi, line 401: Slightly offtopic but this just reminded me that ipaddress module doesn't support scope id in IPV6 address yet. Maybe once https://github.com/python/cpython/pull/13772/ is merged we can just catch ValidationError and return False since the same validation would already be done in the parser to check for % in scope and remove validation here.

Lib/test/test_urlparse.py
vstinner: I'm checking the port to reject [ and ] in the port number. Reject port number outside the [0; 65535] is a side effect. IMHO it's a good thing to reject an invalid URL, no?

tirkarthi: Yes, to be more clear I am fine with the change. It's that previously port validation is done while accessing port attribute allowing invalid URL to be parsed but now it's done in parsing itself which is better as per this PR.

Lib/urllib/parse.py
vstinner: The allowed characters in the scope part is not well defined. I read https://tools.ietf.org/html/rfc4007 and https://tools.ietf.org/html/rfc6874 RFC 6874:

   A <zone_id> SHOULD contain only ASCII characters classified as
   "unreserved" for use in URIs [RFC3986].  This excludes characters
   such as "]" or even "%" that would complicate parsing.  However, the
   syntax described below does allow such characters to be percent-
   encoded, for compatibility with existing devices that use them.

   If an operating system uses any other characters in zone or interface
   identifiers that are not in the "unreserved" character set, they MUST
   be represented using percent encoding [RFC3986].

So... is % allowed in an **unquoted** URL?

vstinner:
@orsenthil: Would you mind to review this change? Any idea for the allowed characters in an IPv6 scope?

orsenthil:
@vstinner  - Sure, I will review. I will refer to the RFC for the valid characters for the IPv6 scope.

vstinner:
I tried to allow [ and ] in the user:password part, but then the URL parser is confused by the URL:

```
http://[::1%sc[o]pe]
```

It reads it as IPv6 `::1%sc[o`.

vstinner:
I rebased my PR to fix the merge conflict.

@orsenthil: Ping for review.

vstinner:
Firefox doesn't seem to accept % in the IPv6 part of an URL. When I type the following URL, it opens Google with the URL as a search...

```
http://[::1%1]:8000/
```

vstinner:
> Firefox doesn't seem to accept % in the IPv6 part of an URL. When I type the following URL, it opens Google with the URL as a search...

Same behavior in Chromium.

Lib/urllib/parse.py
orsenthil: >  However, the syntax described below does allow such characters to be percent-
encoded, for compatibility with existing devices that use them.

@vstinner  - I read the rfc6874 and the part that you quoted. If we want strict adherence to the RFC, I think, allowing  % in unquoted URL is correct.

I looked at https://github.com/python/cpython/pull/13772 , which is trying to bring in the scope id to ipaddress module. There is an agreement and a documented statement that states: "If present, the scope ID must be non-empty,       and may not contain ``%``."

Let's keep the current change, and not allow '%', assuming that it wont be common for all practical purposes. It will be consistent within the standard library. If the allowance of '%' in scope-id is desired, that could be changed in ipaddress module, and once the ipaddress module scope-id support is merged, we could use the facility here.

orsenthil (approved):
LGTM. Thank you.

orsenthil:
> Ping for review.

Done. Thank you, Victor.

Lib/urllib/parse.py
vstinner: > I think, allowing % in unquoted URL is correct.

Sorry but I'm not used to the urllib module. Is urllib.parse.urlsplit() supposed to get a "quoted" or "unquoted" URL?

orsenthil: >  Is urllib.parse.urlsplit() supposed to get a "quoted" or "unquoted" URL?

It is supposed to get the unquoted URL. I relied on tests of urlparse to state this.

---

My reading of RFC 6874 and especially, the part quoted makes me think that 'percent-encoded' character like `%40`  or `%25` _could be_ present in the zone-id component beyond the first '%'.  - If it is the case, the current implementation will False for it a valid IPv6 URL, *but* this is consistent with what *ipaddress* module return and it is documented by the ipaddress module.

I lack the experience to say something confidently about 'percent-encoded characters' in zone-id, and i think being consistent within modules and with documentation is the most appropriate thing to do.

I am +1 with committing this change. Please let me know you have hesitation or do you want (me/) us to research further.

orsenthil:
> Firefox doesn't seem to accept % in the IPv6 part of an URL.

> Same behavior in Chromium.

I expected these browsers to have percent-encode these and work.

https://bugzilla.mozilla.org/show_bug.cgi?id=700999

This is the most relevant information I found 

https://en.wikipedia.org/wiki/IPv6_address#Use_of_zone_indices_in_URIs

```
When used in uniform resource identifiers (URI), the use of the percent sign causes a syntax conflict, therefore it must be escaped via percent-encoding,[11] e.g.:

http://[fe80::1ff:fe23:4567:890a%25eth0]/
```

Lib/urllib/parse.py
orsenthil: 1. urllsplit accepts unquoted URl and returns the components of the URI. The change proposed in this patch and tests is doing this accurately.
2. urlopen and open interface will quote the URL and percent-encode them for the special characters. (So, the examples of firefox and chromium presented in the discussion, I will expected those URIs to work in those browsers if they are percent-encoded.)

vstinner:
The [living URL Standard](https://url.spec.whatwg.org/) [doesn't implement IPv6 scope](https://url.spec.whatwg.org/#host-representation) on purpose:

> Support for <zone_id> is intentionally omitted.

This comment points to https://www.w3.org/Bugs/Public/show_bug.cgi?id=27234#c2 which is a comment written by Ryan Sleevi at 2015-08-14:

> Yes, we're especially not keen to support these in Chrome and have repeatedly decided not to. The platform-specific nature of <zone_id> makes it difficult to impossible to validate the well-formedness of the URL (see https://tools.ietf.org/html/rfc4007#section-11.2 , as referenced in 6874, to fully appreciate this special hell). Even if we could reliably parse these (from a URL spec standpoint), it then has to be handed 'somewhere', and that opens a new can of worms.

> Even 6874 notes how unlikely it is to encounter these in practice

```
   Thus, URIs including a
   ZoneID are unlikely to be encountered in HTML documents.  However, if
   they do (for example, in a diagnostic script coded in HTML), it would
   be appropriate to treat them exactly as above.
```

> Note that a 'dumb' parser may not be sufficient, as the Security Considerations of 6874 note:

```
   To limit this risk, implementations MUST NOT allow use of this format
   except for well-defined usages, such as sending to link-local
   addresses under prefix fe80::/10.  At the time of writing, this is
   the only well-defined usage known.
```

> And also

```
   An HTTP client, proxy, or other intermediary MUST remove any ZoneID
   attached to an outgoing URI, as it has only local significance at the
   sending host.
```

> This requires a transformative rewrite of any URLs going out the wire. That's pretty substantial. Anne, do you recall the bug talking about IP canonicalization (e.g. http://127.0.0.1 vs http://[::127.0.0.1] vs http://012345 and friends?) This is conceptually a similar issue - except it's explicitly required in the context of <zone_id> that the <zone_id> not be emitted.

> There's also the issue that zone_id precludes/requires the use of APIs that user agents would otherwise prefer to avoid, in order to 'properly' handle the zone_id interpretation. For example, Chromium on some platforms uses a built in DNS resolver, and so our address lookup functions would need to define and support <zone_id>'s and map them to system concepts. In doing so, you could end up with weird situations where a URL works in Firefox but not Chrome, even though both 'hypothetically' supported <zone_id>'s, because FF may use an OS routine and Chrome may use a built-in routine and they diverge.

> Overall, our internal consensus is that <zone_id>'s are bonkers on many grounds - the technical ambiguity (and RFC 6874 doesn't really resolve the ambiguity as much as it fully owns it and just says #YOLOSWAG) - and supporting them would add a lot of complexity for what is explicitly and admittedly a limited value use case.

Firefox feature request https://bugzilla.mozilla.org/show_bug.cgi?id=700999 has been rejected using this comment as well at 2015-08-14.

---

Currently, only Microsoft Edge supports IPv6 scope: Firefox and Chromium don't.

I suggest to follow Firefox, Chromium and living URL Standard example: **don't support IPv6 scope**.

My current implementation doesn't implement the RFC 6874 which suggests to use ``%25`` between the IPv6 and the scope. For example address ``::1`` with scope ``eth0`` should be written ``::1%25eth0``. This syntax is hard to read if you use numeric scopes which are common: ``::1`` with scope ``2`` should be written ``::1%252`` :-(

vstinner:
> It is supposed to get the unquoted URL. I relied on tests of urlparse to state this.

This makes me even more uncomfortable to support IPv6 scope: it is not well defined if urlsplit() is expected to be used on a quote or unquoted URL. This is a major difference for RFC 6874 which is tied to quoted characters. Not well defined means: we should not have to dig into tests to reverse engineer the "expected" function behavior. It should be well documented and well tested.

I mean that **if** someone wants to support IPv6 scope in URL, I suggest to first clarify what urlsplit() expects. IMHO fixing this is out of the scope of fixing https://bugs.python.org/issue36338 security vulnerability.

Lib/urllib/parse.py
opavlyuk: @orsenthil , @vstinner In https://github.com/python/cpython/pull/13772 I come to the decision, that `%` character shall not be allowed in <zone_id> part according to paragraph 5 of [Section 11.2 RFC 4007](https://tools.ietf.org/html/rfc4007#section-11.2):

>    An implementation MAY support other kinds of non-null strings as
>    <zone_id>.  However, **the strings must not conflict with the delimiter
>    character**.

At the same time, [Section 2 RFC 6874](https://tools.ietf.org/html/rfc6874#section-2), especially paragraph 4, presumes, that <zone_id> part may contain `%` represented using percent encoding.

That is confusing and seems to conflict with [RFC 4007](https://tools.ietf.org/html/rfc4007).

That's why, I decided to assume, that [RFC 6874](https://tools.ietf.org/html/rfc6874) conserns only representing IPv6 Zone Identifiers in URI`s.

vstinner:
I failed finding time to finish the PR. I prefer to abandon it.

PR #16782, opened 2019-10-14
bpo-38488: Upgrade bundled versions of pip & setuptools

xavfernandez:
Following the release of pip 19.3.

It's unclear to me if a bpo ticket is needed ?

https://bugs.python.org/issue38488


the-knights-who-say-ni:

## Recognized GitHub username

@xavfernandez

tirkarthi:
This would need a bpo and a news entry similar to https://bugs.python.org/issue35807 .

Lib/ensurepip/__init__.py
```diff
@@ -27,8 +27,8 @@ def _run_pip(args, additional_paths=None):
         sys.path = additional_paths + sys.path
 
     # Install the bundled software
-    import pip._internal
-    return pip._internal.main(args)
+    import pip._internal.main
+    return pip._internal.main.main(args)
```
pradyunsg: Are we OK to change this to use runpy? That way we'd basically be doing python -m pip, which is directly supported.

See https://github.com/pradyunsg/pip-cli/blob/master/src/_pip_cli.py#L75 for how we can do that. You can skip setting the env there, since that's for passing information that's not relevant here.

chrahunt: I wouldn't say that it's directly supported (as in, part of the public API) since running in the same process space is still not considered supported, but it would avoid any issues with changes to the location of the entrypoint function.

It may also be helpful to consider backporting that change to bugfix Python versions, which can help people that build and patch in new versions of pip (as mentioned in 09fd200).

taleinat: Why is this not simply run in a sub-process, then?

encukou:
Also see this Discourse discussion: https://discuss.python.org/t/can-we-finally-add-a-minimal-api-to-pip/2833

encukou:
The best (though still not supported) way to run pip in-process is with `runpy`, as mentioned in a new comment in pip: https://github.com/pypa/pip/pull/7471/files

Could you update the pull request to use that?
Note that it will `exit()` the current process, so it will need a [documentation change](https://docs.python.org/3.5/library/ensurepip.html#ensurepip.bootstrap) and a [what's new entry](https://github.com/python/cpython/blob/master/Doc/whatsnew/3.9.rst).

Please comment if you have any questions, or if you'd prefer to leave this to someone else.

pradyunsg:
A subprocess.run might be a more robust solution here.

taleinat:
> A subprocess.run might be a more robust solution here.

As I mentioned in the inline comment, I wonder why this isn't run in a sub-process, and suggest doing so.

Regardless, that could be a separate issue. This simple update should likely go in.

taleinat:
@xavfernandez, have you any intention to continue working on this?

pradyunsg:
Sounds fair to me -- I filed a bpo issue for this a while back -- maybe someone/me can find it and link to it from here.

None of my concerns here are blocking this version update, which should definitely go in.

pradyunsg (approved):

tirkarthi:
@pradyunsg Is it this https://bugs.python.org/issue38662?

xavfernandez:
> @xavfernandez, have you any intention to continue working on this?

Yes, however it's unclear what needs to be done in this PR ?
Should I switch to runpy and follow encukou suggestion ?
Or simply launch a subprocess running `python -m pip` ?

pradyunsg:
@tirkarthi Yes, indeed. Many thanks for finding that. :)

@xavfernandez IMO nothing else is needed here -- we can tackle those in a follow up. /cc @ncoghlan @dstufft @pfmoore in case one of them wants to review + click merge here. :)

taleinat:
By now the latest versions of setuptools and pip have moved along:

* [setuptools](https://pypi.org/project/setuptools/) is at 42.0.2
* [pip](https://pypi.org/project/pip/) is at 19.3.1

So we should bump those and then IMO this would be ready to go.

xavfernandez:
@taleinat I've bumped setuptools to 42.0.2 :+1:

taleinat (approved):
LGTM

Lib/ensurepip/__init__.py
ncoghlan: We don't use runpy (or a subprocess) here because of the path manipulation just above to import pip from the bundled wheel archives.

It's useful to think of ensurepip as being part of pip, as we're free to make whatever changes are needed to sync up when updating to new versions of pip.

pradyunsg: Let's discuss in https://bugs.python.org/issue38662 or #17029

Misc/NEWS.d/next/Library/2019-10-15-23-28-11.bpo-38488.hFQNgA.rst
```diff
@@ -0,0 +1 @@
+Update ensurepip to install pip 19.3 and setuptools 42.0.2.
```
The-Compiler: Nitpick:

```suggestion
Update ensurepip to install pip 19.3.1 and setuptools 42.0.2.
```

The-Compiler (commented):
Looks like setuptools v43.0.0 and v44.0.0 were released in the meantime.

Also, note that pip 19.3 is needed to install manylinux2014 wheels ([PEP-599](https://www.python.org/dev/peps/pep-0599/)) via `python -m venv` virtualenvs - since those wheels seem to get more common, I guess it'd be good to have this in Python 3.8.2.

xavfernandez:
Setuptools version updated to 44.0.0 and news file updated.

The-Compiler (approved):
Seems fine to me, though I'm not a core developer or anything like that!

Lib/ensurepip/__init__.py
vstinner: https://bugs.python.org/issue38662 is now fixed in master. You can rebased your PR on master.

xavfernandez: Done.

pradyunsg:
Gentle nudge on this.

/cc @pfmoore @dstufft @ncoghlan @zooba (sorry for being noisy!)

ncoghlan (approved):
Sorry about the delay, I suspect this was a case of all of us assuming one of the others would approve & merge it.

stratakis:
Various buildbots which run the tests with an installed python, fail with:

chmod: target/lib/python3.10/site-packages/pip-20.1.dist-info/RECORD: new permissions are r-xrw-rwx, not r-xr--r-x
chmod: target/lib/python3.10/site-packages/pip-20.1.dist-info/INSTALLER: new permissions are r-xrw-rwx, not r-xr--r-x
chmod: target/lib/python3.10/site-packages/setuptools-46.1.3.dist-info/RECORD: new permissions are r-xrw-rwx, not r-xr--r-x
chmod: target/lib/python3.10/site-packages/setuptools-46.1.3.dist-info/INSTALLER: new permissions are r-xrw-rwx, not r-xr--r-x

vstinner:
> Various buildbots which run the tests with an installed python, fail with: (...)

It's the "chmod" step which fails. I guess that only "Install" buildbot workers are affected. I failed to reproduce the issue with commands:

```
./configure --prefix $PWD/target --with-pydebug
make && make install
chmod -R -w target/
```

I get these permissions:

```
$ ls -l target/lib/python3.10/site-packages/setuptools-46.1.3.dist-info/RECORD
-r--r--r--. 1 vstinner vstinner 14560 28 mai   03:40 target/lib/python3.10/site-packages/setuptools-46.1.3.dist-info/RECORD
```

vstinner:
I reverted the change: see https://bugs.python.org/issue38488

xavfernandez:
The permission issue is likely due to https://github.com/pypa/pip/issues/8164 that was fixed in 20.1.1.
Sorry I did not keep this PR up-to-date with the latest pip, #20491 should hopefully fix this.

PR #16784, opened 2019-10-14
bpo-38473: Handle autospecced functions and methods used with attach_mock

tirkarthi:
Extract signature from the inner mock object for methods/functions autospecced and attached to a mock object using `attach_mock`.

https://bugs.python.org/issue38473

### DIFF
--- a/Lib/unittest/mock.py
+++ b/Lib/unittest/mock.py
@@ -824,10 +824,7 @@ class NonCallableMock(Base):
                 # not used.
                 child = _extract_mock(child)
                 children = child._mock_children
-                try:
-                    sig = child._spec_signature
-                except AttributeError:
-                    sig = None
+                sig = child._spec_signature
 
         return sig
 
--- a/Lib/unittest/mock.py
+++ b/Lib/unittest/mock.py
@@ -819,9 +819,7 @@ class NonCallableMock(Base):
             else:
                 # If an autospecced object is attached using attach_mock the
                 # child would be a function with mock object as attribute from
-                # which signature has to be derived. If there is no signature
-                # attribute then fallback to None to ensure old signature is
-                # not used.
+                # which signature has to be derived.
                 child = _extract_mock(child)
                 children = child._mock_children
                 sig = child._spec_signature
--- a/Lib/unittest/mock.py
+++ b/Lib/unittest/mock.py
@@ -817,13 +817,13 @@ class NonCallableMock(Base):
             if child is None or isinstance(child, _SpecState):
                 break
             else:
-                children = child._mock_children
                 # If an autospecced object is attached using attach_mock the
                 # child would be a function with mock object as attribute from
                 # which signature has to be derived. If there is no signature
                 # attribute then fallback to None to ensure old signature is
                 # not used.
                 child = _extract_mock(child)
+                children = child._mock_children
                 try:
                     sig = child._spec_signature
                 except AttributeError:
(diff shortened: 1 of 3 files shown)


Lib/unittest/mock.py
```diff
@@ -820,7 +820,16 @@ def _get_call_signature_from_name(self, name):
                 break
             else:
                 children = child._mock_children
```
tirkarthi: Note to self, reviewer : Do we need to default to empty dictionary too guarding against `AttributeError` for `_mock_children`?

tirkarthi:
Sorry, I somehow forgot this issue in my list. Would appreciate a review of this since it caused a regression. cc: @mariocj89 

Thanks

Lib/unittest/mock.py
```diff
@@ -817,8 +817,17 @@ def _get_call_signature_from_name(self, name):
...
+                # child would be a function with mock object as attribute from
+                # which signature has to be derived. If there is no signature
+                # attribute then fallback to None to ensure old signature is
+                # not used.
+                child = _extract_mock(child)
                 children = child._mock_children
-                sig = child._spec_signature
+                try:
+                    sig = child._spec_signature
+                except AttributeError:
```
mariocj89: In which situation will we fall here? Can you add a tests that exercises this?

tirkarthi: I couldn't think of a situation that would get to this as autospecced ones should have this. But given that this was a regression of the previous change without tests catching it I added it as a fallback to have signature as None in case there are scenarios not tested here or try to use `getattr` with None as default.

mariocj89 (approved):

Lib/unittest/mock.py
mariocj89: I personally don't like the idea of having code "just if", but you wrote this function initially, so... 🤷‍♂  😛

Lib/unittest/mock.py
cjw296: What test cases could we add to verify that?

Lib/unittest/mock.py
cjw296: If we can't prove it's needed, let's drop it until we can. The backport caught up now, just want to get as many upstream PRs merged first before releasing, so we should get good feedback if this code can be reached.

cjw296:
I'm afraid I don't have a lot to offer here, since I've never used `attach_mock`.

tirkarthi:
I have removed the `try/catch` block. From https://github.com/python/cpython/blob/161e7b36b1ea871a1352ccfc1d4f4c1eda76830f/Lib/unittest/mock.py#L434 `_spec_signature` seems to be always initialized with None on mock object. The attribute error was due to inner mock object not being extracted which is fixed. I was slightly concerned since the tests didn't catch it and caused a regression. I understand your points on having an untested block as a fallback. I will revisit this if there is a report of it in future. Thanks for your inputs.

mariocj89 (approved):

Lib/unittest/mock.py
```diff
@@ -817,6 +817,12 @@ def _get_call_signature_from_name(self, name):
             if child is None or isinstance(child, _SpecState):
                 break
             else:
+                # If an autospecced object is attached using attach_mock the
+                # child would be a function with mock object as attribute from
+                # which signature has to be derived. If there is no signature
```
mariocj89: You might want to review this comment (maybe remove it as well and include it in the commit message).

tirkarthi: Thanks for catching this. I prefer to keep it here with the part about fallback updated since it will be good to know why extract_mock is needed. I will add this in the comment over this change.

cjw296 (approved):

PR #16787, opened 2019-10-14
bpo-38474: Replace digit check logic into Py_ISDIGIT

corona10:
https://bugs.python.org/issue38474


corona10:
@serhiy-storchaka 
Can you please take look at this PR?

PR #16788, opened 2019-10-14
bpo-38006: Add unit test for weakref clear bug

nascheme:

### DIFF
--- a/Lib/test/test_gc.py
+++ b/Lib/test/test_gc.py
@@ -22,6 +22,11 @@ except ImportError:
                 raise TypeError('requires _testcapi.with_tp_del')
         return C
 
+try:
+    from _testcapi import ContainerNoGC
+except ImportError:
+    ContainerNoGC = None
+
 ### Support code
 ###############################################################################
 
@@ -958,6 +963,69 @@ class GCTests(unittest.TestCase):
         self.assertEqual(nc - oldnc, 0)
 
         gc.enable()
+
+    @unittest.skipIf(ContainerNoGC is None,
+                     'requires ContainerNoGC extension type')
+    def test_trash_weakref_clear(self):
+        # Test that trash weakrefs are properly cleared (bpo-38006).
+        #
+        # Structure we are creating:
+        #
+        #   Z <- Y <- A--+--> WZ -> C
+        #             ^  |
+        #             +--+
+        # where:
+        #   WZ is a weakref to Z with callback C
+        #   Y doesn't implement tp_traverse
+        #   A contains a reference to itself, Y and WZ
+        #
+        # A, Y, Z, WZ are all trash.  The GC doesn't know that Z is trash
+        # because Y does not implement tp_traverse.  To show the bug, WZ needs
+        # to live long enough so that Z is deallocated before it.  Then, if
+        # gcmodule is buggy, when Z is being deallocated, C will run.
+        #
+        # To ensure WZ lives long enough, we put it in a second reference
+        # cycle.  That trick only works due to the ordering of the GC prev/next
+        # linked lists.  So, this test is a bit fragile.
+        #
+        # The bug reported in bpo-38006 is caused because the GC did not
+        # clear WZ before starting the process of calling tp_clear on the
+        # trash.  Normally, handle_weakrefs() would find the weakref via Z and
+        # clear it.  However, since the GC cannot find Z, WR is not cleared and
+        # it can execute during delete_garbage().  That can lead to disaster
+        # since the callback might tinker with objects that have already had
+        # tp_clear called on them (leaving them in possibly invalid states).
+
+        wr_callback_was_run = False
+        def callback(wr):
+            # We cannot allow this callback to run.  It could access objects
+            # with invalid states due to tp_clear already being run.  Or, it
+            # could resurrecte objects that were previously considered trash.
+            nonlocal wr_callback_was_run
+            wr_callback_was_run = True
+
+        class A:
+            __slots__ = ['a', 'y', 'wz']
+
+        class Z:
+            pass
+
+        # setup required object graph, as described above
+        a = A()
+        a.a = a
+        a.y = ContainerNoGC(Z())
+        a.wz = weakref.ref(a.y.value, callback)
+        # create second cycle to keep WZ alive longer
+        wr_cycle = [a.wz]
+        wr_cycle.append(wr_cycle)
+        # ensure trash unrelated to this test is gone
+        gc.collect()
+        # release references and create trash
+        del a, wr_cycle
+        gc.collect()
+        # if the callback runs, we have a bug
+        self.assertFalse(wr_callback_was_run)
+
 
 class GCCallbackTests(unittest.TestCase):
     def setUp(self):
--- a/Lib/test/test_gc.py
+++ b/Lib/test/test_gc.py
@@ -1015,12 +1015,14 @@ class GCTests(unittest.TestCase):
         wr_cycle.append(wr_cycle)
         # ensure trash unrelated to this test is gone
         gc.collect()
+        gc.disable()
         # release references and create trash
         del a, wr_cycle
         gc.collect()
         # if called, it means there is a bug in the GC.  The weakref should be
         # cleared before Z dies.
         callback.assert_not_called()
+        gc.enable()
 
 
 class GCCallbackTests(unittest.TestCase):
--- a/Lib/test/test_gc.py
+++ b/Lib/test/test_gc.py
@@ -1,4 +1,5 @@
 import unittest
+import unittest.mock
 from test.support import (verbose, refcount_test, run_unittest,
                           strip_python_stderr, cpython_only, start_threads,
                           temp_dir, requires_type_collecting, TESTFN, unlink,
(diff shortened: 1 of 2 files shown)


Modules/_testcapimodule.c
```diff
@@ -6538,6 +6538,51 @@ static PyTypeObject MethStatic_Type = {
...
+    PyObject_HEAD
+    PyObject *value;
+} ContainerNoGCobject;
+
+static PyObject *
+ContainerNoGC_new(PyTypeObject *type, PyObject *args, PyObject *kwargs)
+{
+    PyObject *value;
+    char *names[] = {"value", NULL};
+    if (!PyArg_ParseTupleAndKeywords(args, kwargs, "O", names, &value))
```
pablogsal: Nit: Can you add braces per PEP 7?

Modules/_testcapimodule.c
```diff
@@ -6538,6 +6538,51 @@ static PyTypeObject MethStatic_Type = {
...
+        return NULL;
+    PyObject *self = type->tp_alloc(type, 0);
+    if (self == NULL)
```
pablogsal: Same as before, can you add braces?

Modules/_testcapimodule.c
```diff
@@ -6735,6 +6780,13 @@ PyInit__testcapi(void)
     Py_DECREF(subclass_with_finalizer_bases);
     PyModule_AddObject(m, "HeapCTypeSubclassWithFinalizer", HeapCTypeSubclassWithFinalizer);
 
+    if (PyType_Ready(&ContainerNoGC_type) < 0)
+        return NULL;
+    Py_INCREF(&ContainerNoGC_type);
+    if (PyModule_AddObject(m, "ContainerNoGC",
+                           (PyObject *) &ContainerNoGC_type) < 0)
```
pablogsal, line 6790: Same as before, can you add braces?

Modules/_testcapimodule.c
```diff
@@ -6735,6 +6780,13 @@ PyInit__testcapi(void)
     Py_DECREF(subclass_with_finalizer_bases);
     PyModule_AddObject(m, "HeapCTypeSubclassWithFinalizer", HeapCTypeSubclassWithFinalizer);
 
+    if (PyType_Ready(&ContainerNoGC_type) < 0)
```
pablogsal: Same as before, can you add braces?

Lib/test/test_gc.py
```diff
@@ -959,6 +964,69 @@ def getstats():
...
+        # The bug reported in bpo-38006 is caused because the GC did not
+        # clear WZ before starting the process of calling tp_clear on the
+        # trash.  Normally, handle_weakrefs() would find the weakref via Z and
+        # clear it.  However, since the GC cannot find Z, WR is not cleared and
+        # it can execute during delete_garbage().  That can lead to disaster
+        # since the callback might tinker with objects that have already had
+        # tp_clear called on them (leaving them in possibly invalid states).
+
+        wr_callback_was_run = False
+        def callback(wr):
```
pablogsal: We can use a `unittest.mock` object here to simplify this logic:

```diff
diff --git a/Lib/test/test_gc.py b/Lib/test/test_gc.py
index 803191fd69..f37f1e62d1 100644
--- a/Lib/test/test_gc.py
+++ b/Lib/test/test_gc.py
@@ -12,6 +12,7 @@ import textwrap
 import threading
 import time
 import weakref
+import unittest.mock
 
 try:
     from _testcapi import with_tp_del
@@ -996,13 +997,7 @@ class GCTests(unittest.TestCase):
         # since the callback might tinker with objects that have already had
         # tp_clear called on them (leaving them in possibly invalid states).
 
-        wr_callback_was_run = False
-        def callback(wr):
-            # We cannot allow this callback to run.  It could access objects
-            # with invalid states due to tp_clear already being run.  Or, it
-            # could resurrecte objects that were previously considered trash.
-            nonlocal wr_callback_was_run
-            wr_callback_was_run = True
+        callback = unittest.mock.Mock()
 
         class A:
             __slots__ = ['a', 'y', 'wz']
@@ -1024,7 +1019,7 @@ class GCTests(unittest.TestCase):
         del a, wr_cycle
         gc.collect()
         # if the callback runs, we have a bug
-        self.assertFalse(wr_callback_was_run)
+        callback.assert_not_called()
 
 
 class GCCallbackTests(unittest.TestCase):
```

Lib/test/test_gc.py
```diff
@@ -959,6 +964,69 @@ def getstats():
 
         gc.enable()
 
+    @unittest.skipIf(ContainerNoGC is None,
+                     'requires ContainerNoGC extension type')
+    def test_trash_weakref_clear(self):
```
pablogsal, line 970: This test is great! Thanks for the time working on this. I find the comments very clear and I know is hard to control all these moving pieces. Great work!

Lib/test/test_gc.py
```diff
@@ -959,6 +964,69 @@ def getstats():
...
+            # could resurrecte objects that were previously considered trash.
+            nonlocal wr_callback_was_run
+            wr_callback_was_run = True
+
+        class A:
+            __slots__ = ['a', 'y', 'wz']
+
+        class Z:
+            pass
+
```
pablogsal, line 1007: I will suggest starting the test disabling the collector, and end the test enabling it again so there are no surprising calls in the middle because hidden allocations when we don't expect (this tends to happen if the test runs with the `--huntrleaks`  in my experience).

pablogsal (commented):
Thanks @nascheme for working on this! Looks great, I left some suggestions :)

Lib/test/test_gc.py
```diff
@@ -959,6 +964,69 @@ def getstats():
...
+        #
+        #   Z <- Y <- A--+--> WZ -> C
+        #             ^  |
+        #             +--+
+        # where:
+        #   WZ is a weakref to Z with callback C
+        #   Y doesn't implement tp_traverse
+        #   A contains a reference to itself, Y and WZ
+        #
+        # A, Y, Z, WZ are all trash.  The GC doesn't know that Z is trash
```
pablogsal:
```suggestion
        # A, Y, Z, WZ are all trash. The GC doesn't know that Z is trash
```

Lib/test/test_gc.py
```diff
@@ -959,6 +964,69 @@ def getstats():
...
+        # because Y does not implement tp_traverse.  To show the bug, WZ needs
+        # to live long enough so that Z is deallocated before it.  Then, if
+        # gcmodule is buggy, when Z is being deallocated, C will run.
+        #
+        # To ensure WZ lives long enough, we put it in a second reference
```
pablogsal, line 988: Idea: After the latest changes that we made in previous PRs, another thing we can do to make sure WZ lives long enough is having some object in the same cycle as A that resurrects, resurrecting WZ with it. In this way, we know that WZ will be alive when `delete_garbage` is executed. Please, double-check my logic :)

ambv:
Neil, this is great work.

Lib/test/test_gc.py
tim-one: Except it's not enough that WZ merely be alive, it must also be that WZ hasn't been cleared (in the weakref sense, not the `tp_clear` sense).  But Neil fixed this problem by ensuring WZ _is_ cleared, long before we get anywhere near `delete_garbage`.  So deallocating Z will no longer trigger C.

So the problem can't be triggered in code since Neil's fix.  Code before that is also before the more-recent finalizer changes, so in those running a finalizer that resurrects something causes the gc run to abort, before  `delete_garbage` is run.  So Z, WZ, and C would all survive then, but undamaged.

pablogsal: > Except it's not enough that WZ merely be alive, it must also be that WZ hasn't been cleared (in the weakref sense, not the `tp_clear` sense)

But if I deactivate Neil's fix for this problem, with the current state of the gc, the scenario that I described should trigger the error (the callback is invoked), no?

tim-one: Probably, but we really want a test that fails under some version of Python that was actually released.

pablogsal: > Probably, but we really want a test that fails under some version of Python that was actually released.

Oh, my bad then. I thought this test was not going to be backported.

tim-one: I don't think backporting is relevant.  To show that a bug has been fixed, we need a test case that fails under an earlier version.  There's no need to backport it.  It's more:  "here:  run this under 3.7.2 (whatever).  See?  It fails.  But under current Python it doesn't.".  Indeed, we never backport tests to versions under which they fail 😉

pablogsal: > Indeed, we never backport tests to versions under which they fail

Gotcha, everything clear. For one second I was thinking that we were going to backport the test *and* the fix. :)

tim-one: We can backport the test and the fix (and should!), but we certainly can't include the recent gc finalizer enhancements in old-version bugfix releases.

nascheme: If we had something like ContainerNoGC in old Python versions, the test would be easy to backport and should fail in versions going back to 2.7 (and earlier, up to when weakrefs were added).  If you want to make Python crash, find an object where tp_clear leaves it in an invalid state.   types.SimpleNamespace is one example.  Probably nicer if the unit tests don't SEGV on failure. ;-P

Lib/test/test_gc.py
nascheme: Running "del x, y" should not trigger GC but I agree it is better to use gc.disable() and be very sure.  Randomly failing tests are not nice.

pablogsal (approved):

tim-one (approved):

pablogsal:
Don't worry about the buildbot failure: is unrelated.

PR #16790, opened 2019-10-14
Doc: 3.8 is now stable.

JulienPalard:


ned-deily (approved):
LGTM

ned-deily:
@JulienPalard, I pushed this before I remembered that Doc/tools/static/switchers.js also needs to be updated :(

PR #16800, opened 2019-10-15
bpo-38478: Correctly handle keyword argument with same name as positional-only parameter

pablogsal:
https://bugs.python.org/issue38478

### DIFF
--- a/Lib/inspect.py
+++ b/Lib/inspect.py
@@ -2960,7 +2960,7 @@ class Signature:
                         arguments[param.name] = tuple(values)
                         break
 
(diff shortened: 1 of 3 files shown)


pablogsal:
CC: @tirkarthi

tirkarthi (approved):
LGTM. I too had the same solution in mind since it's simple . Thanks Pablo :)

pablogsal:
Thanks @tirkarthi :)

pablogsal:
The buildbot failure is unrelated: `test_start_tls_server_1`

PR #16802, opened 2019-10-15
bpo-38483: Add support for venv.ini

cooperlees:
- Add argparse.ArgumentDefaultsHelpFormatter so user can see overloads
- Add new CliDefaults NamedTuple with defaults
- Check for ~/.venvrc + parse if exists
- Check args, if true or false bool otherwise string
- Overload by creating NamedTuple passing kwargs
- Update docs

Test:
- Run `./python.exe -m venv --help`
- Add unit tests for new function

https://bugs.python.org/issue38483

### DIFF
--- a/Doc/using/venv-create.inc
+++ b/Doc/using/venv-create.inc
@@ -133,17 +133,23 @@ detail (typically a script or shell function will be used).
    PowerShell activation scripts installed under POSIX for PowerShell Core
    support.
 
-The venv module supports a Running Command file (commandparser ini format) within
-a users home directory (``~/.venvrc``) to overload CLI defaults. This allows
-any CLI argument mention above to me overloaded by default.
-- Please replace '-' chars with '_' to match cli args.
-Example Usage:
+The venv module supports an ini config file in the following locations:
 
-[venv]
-upgrade_deps = true
+* ``$HOME/.venv.ini`` on Linux/Unix (+ Darwin/Mac OSX) platforms
+* ``%APPDATA%\Python\venv\venv.ini'`` on Windows platforms
+
+to overload CLI defaults. This allows any CLI argument mentioned above
+to be overloaded by default.
+
+* Please replace '-' chars with '_' to match cli args.
+
+Example ``venv.ini``:
+
+    [venv]
+    upgrade_deps = true
 
 This would result in ``--upgrade-deps`` to defualt to True.
-Please use ``--help`` to test ``~/.venvrc``
+Please use ``--help`` to test ``venv.ini`` applied defaults
 
 .. versionadded:: 3.9
-   Add support for ~/.venvrc to overload CLI argument defaults
+   Add support for venv.ini to overload CLI argument defaults
--- a/Doc/using/venv-create.inc
+++ b/Doc/using/venv-create.inc
@@ -135,21 +135,30 @@ detail (typically a script or shell function will be used).
 
 The venv module supports an ini config file in the following locations:
 
-* ``$HOME/.venv.ini`` on Linux/Unix (+ Darwin/Mac OSX) platforms
-* ``%APPDATA%\Python\venv\venv.ini'`` on Windows platforms
+* ``$HOME/.config/python/venv.ini`` on Linux/Unix (+ Darwin/Mac OSX) platforms
+* ``%APPDATA%\Python\venv.ini'`` on Windows platforms
 
-to overload CLI defaults. This allows any CLI argument mentioned above
-to be overloaded by default.
+This config allows a user to alter the CLI option defaults set in the venv module.
+The config only alters the default and if specified on the command line it
+will obtain that value.
 
 * Please replace '-' chars with '_' to match cli args.
 
 Example ``venv.ini``:
 
     [venv]
-    upgrade_deps = true
+    upgrade_deps = True
 
-This would result in ``--upgrade-deps`` to defualt to True.
+This would result in ``--upgrade-deps`` to default to True.
 Please use ``--help`` to test ``venv.ini`` applied defaults
 
+If a non valid key is set in ``venv.ini`` it will be silently ignored.
+Example:
+
+    [venv]
+    guido_retired = True
+
+will result in no changes to defaults to occur
+
 .. versionadded:: 3.9
-   Add support for venv.ini to overload CLI argument defaults
+   Add support for venv.ini to supply CLI argument defaults
--- a/Doc/using/venv-create.inc
+++ b/Doc/using/venv-create.inc
@@ -132,3 +132,18 @@ detail (typically a script or shell function will be used).
 .. versionadded:: 3.8
    PowerShell activation scripts installed under POSIX for PowerShell Core
    support.
+
+The venv module supports a Running Command file (commandparser ini format) within
+a users home directory (``~/.venvrc``) to overload CLI defaults. This allows
+any CLI argument mention above to me overloaded by default.
+- Please replace '-' chars with '_' to match cli args.
+Example Usage:
+
+[venv]
+upgrade_deps = true
+
+This would result in ``--upgrade-deps`` to defualt to True.
+Please use ``--help`` to test ``~/.venvrc``
+
+.. versionadded:: 3.9
+   Add support for ~/.venvrc to overload CLI argument defaults
--- a/Lib/test/test_venv.py
+++ b/Lib/test/test_venv.py
@@ -5,6 +5,7 @@ Copyright (C) 2011-2012 Vinay Sajip.
 Licensed to the PSF under a contributor agreement.
 """
 
+import configparser
 import ensurepip
 import os
 import os.path
@@ -378,6 +379,34 @@ class BasicTest(BaseTest):
         out, err = check_output([bash, test_script])
         self.assertEqual(out, "".encode())
         self.assertEqual(err, "".encode())
+
+
+class CliTest(BaseTest):
+    """ Test various CLI related functions """
+
+    BAD_VENVRC = "|not_venv_or_ini|\nkey = value\n"
+    GOOD_VENVRC = "[venv]\nupgrade_deps = TrUe"  # case on purpose for test
+
+    def test_get_default_args(self):
+        # Use env_dir as it gets cleaned up
+        defaults = venv.CliDefaults()
+        venvrc_path = os.path.join(self.env_dir, "venvrc")
+
+        # Test we handle no config existing
+        # - Not using homedir incase user has a .venvrc
+        self.assertEqual(venv.get_default_args(venvrc_path), defaults)
+
+        # Write out bad RC file + ensure we throw exception
+        with open(venvrc_path, "w") as vfp:
+            vfp.write(self.BAD_VENVRC)
+        with self.assertRaises(configparser.MissingSectionHeaderError):
+            venv.get_default_args(venvrc_path)
+
+        # Write out good RC file and expect upgrade_deps == True
+        expected = venv.CliDefaults(upgrade_deps=True)
+        with open(venvrc_path, "w") as vfp:
+            vfp.write(self.GOOD_VENVRC)
+        self.assertEqual(venv.get_default_args(venvrc_path), expected)
 
 
 @requireVenvCreate
--- a/Lib/test/test_venv.py
+++ b/Lib/test/test_venv.py
@@ -384,29 +384,29 @@ class BasicTest(BaseTest):
 class CliTest(BaseTest):
     """ Test various CLI related functions """
 
-    BAD_VENVRC = "|not_venv_or_ini|\nkey = value\n"
-    GOOD_VENVRC = "[venv]\nupgrade_deps = TrUe"  # case on purpose for test
+    BAD_VENV_INI = "|not_venv_or_ini|\nkey = value\n"
+    GOOD_VENV_INI = "[venv]\nupgrade_deps = TrUe"  # case on purpose for test
 
     def test_get_default_args(self):
         # Use env_dir as it gets cleaned up
         defaults = venv.CliDefaults()
-        venvrc_path = os.path.join(self.env_dir, "venvrc")
+        venv_ini_path = os.path.join(self.env_dir, "venv.ini")
 
         # Test we handle no config existing
-        # - Not using homedir incase user has a .venvrc
-        self.assertEqual(venv.get_default_args(venvrc_path), defaults)
+        # - Not using homedir incase user has a venv.ini
+        self.assertEqual(venv.get_default_args(venv_ini_path), defaults)
 
         # Write out bad RC file + ensure we throw exception
-        with open(venvrc_path, "w") as vfp:
-            vfp.write(self.BAD_VENVRC)
+        with open(venv_ini_path, "w") as vfp:
+            vfp.write(self.BAD_VENV_INI)
         with self.assertRaises(configparser.MissingSectionHeaderError):
-            venv.get_default_args(venvrc_path)
+            venv.get_default_args(venv_ini_path)
 
         # Write out good RC file and expect upgrade_deps == True
         expected = venv.CliDefaults(upgrade_deps=True)
-        with open(venvrc_path, "w") as vfp:
-            vfp.write(self.GOOD_VENVRC)
-        self.assertEqual(venv.get_default_args(venvrc_path), expected)
+        with open(venv_ini_path, "w") as vfp:
+            vfp.write(self.GOOD_VENV_INI)
+        self.assertEqual(venv.get_default_args(venv_ini_path), expected)
 
 
 @requireVenvCreate
--- a/Lib/test/test_venv.py
+++ b/Lib/test/test_venv.py
@@ -381,10 +381,12 @@ class BasicTest(BaseTest):
         self.assertEqual(err, "".encode())
 
     def test_get_default_args_errors(self):
-        bad_bool = "[venv]\nupgrade_deps = Try\n"
+        bad_bool = "[DEFAULT]\nupgrade_deps = Try\n"
         bad_format_venv_ini = "|not_venv_or_ini|\nkey = value\n"
-        bad_duplicate_keys = "[venv]\nupgrade_deps = True\nupgrade_deps = True\n"
-        no_valid_options = "[venv]\nguido_retired = True\n"
+        bad_duplicate_keys = (
+            "[DEFAULT]\nupgrade_deps = True\nupgrade_deps = True\n"
+        )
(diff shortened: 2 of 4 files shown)


Doc/using/venv-create.inc
```diff
@@ -132,3 +132,24 @@ detail (typically a script or shell function will be used).
 .. versionadded:: 3.8
    PowerShell activation scripts installed under POSIX for PowerShell Core
    support.
+
+The venv module supports an ini config file in the following locations:
+
+* ``$HOME/.venv.ini`` on Linux/Unix (+ Darwin/Mac OSX) platforms
```
vsajip: This location should be `$HOME/.config/python/venv.ini` to match the [XDG Base Directory Specification](https://specifications.freedesktop.org/basedir-spec/basedir-spec-latest.html).

Doc/using/venv-create.inc
```diff
@@ -132,3 +132,24 @@ detail (typically a script or shell function will be used).
...
+* ``%APPDATA%\Python\venv\venv.ini'`` on Windows platforms
```
vsajip: Perhaps `%APPDATA%\Python\venv.ini` - there should be no need for an extra `venv` subdirectory for just one file.

Doc/using/venv-create.inc
```diff
@@ -132,3 +132,24 @@ detail (typically a script or shell function will be used).
...
+
+to overload CLI defaults. This allows any CLI argument mentioned above
```
vsajip: This usage of 'overload' might be confusing - perhaps 'supply' might be used instead. Anything provided on the command line should override a corresponding setting in `venv.ini`. The values in `venv.ini` only override those that would be otherwise set in the code in the absence of appropriate CLI arguments. So, I suggest you update the comment to make it clearer that:

* If arguments are provided on the command line, they will be used to set parameter values.
* Otherwise, if a setting is specified in `venv.ini`, it will be used.
* In the absence of both of the above, a suitable default will be applied in the code.

Doc/using/venv-create.inc
```diff
@@ -132,3 +132,24 @@ detail (typically a script or shell function will be used).
...
+to be overloaded by default.
+
+* Please replace '-' chars with '_' to match cli args.
+
+Example ``venv.ini``:
+
+    [venv]
+    upgrade_deps = true
+
+This would result in ``--upgrade-deps`` to defualt to True.
```
vsajip: defualt -> default

Doc/using/venv-create.inc
```diff
@@ -132,3 +132,24 @@ detail (typically a script or shell function will be used).
...
+Please use ``--help`` to test ``venv.ini`` applied defaults
+
+.. versionadded:: 3.9
+   Add support for venv.ini to overload CLI argument defaults
```
vsajip: overload -> supply

Lib/test/test_venv.py
```diff
@@ -380,6 +381,34 @@ def test_deactivate_with_strict_bash_opts(self):
         self.assertEqual(err, "".encode())
 
 
+class CliTest(BaseTest):
+    """ Test various CLI related functions """
+
+    BAD_VENV_INI = "|not_venv_or_ini|\nkey = value\n"
+    GOOD_VENV_INI = "[venv]\nupgrade_deps = TrUe"  # case on purpose for test
```
vsajip: The section should be `DEFAULT` rather than `venv`, as it's the only one.

Lib/venv/__init__.py
```diff
@@ -4,19 +4,34 @@
...
 import sys
 import sysconfig
 import types
-
+from typing import NamedTuple, Optional
 
 CORE_VENV_DEPS = ('pip', 'setuptools')
+VENV_INI_DEFAULT = '~/.venv.ini'
+if sys.platform == "win32":
+    VENV_INI_DEFAULT = os.path.expandvars(r'%APPDATA%\Python\venv\venv.ini')
```
vsajip: Would redo this section as
```
if os.name == 'nt':
    _dir = os.path.expanduser('~/.config')
else:
    _dir = os.path.expandvars(r'%APPDATA%\Python')
os.makedirs(_dir, exist_ok=True)
VENV_INI_DEFAULT_PATH = os.path.join(_dir, 'venv.ini')
del _dir
```

Lib/venv/__init__.py
```diff
@@ -409,8 +424,34 @@ def create(env_dir, system_site_packages=False, clear=False,
                          prompt=prompt, upgrade_deps=upgrade_deps)
     builder.create(env_dir)
 
+
+def get_default_args(venv_ini_default=VENV_INI_DEFAULT):
```
vsajip: venv_ini_default -> venv_ini_path
VENV_INI_DEFAULT -> VENV_INI_DEFAULT_PATH.

Might be an idea to just default to `None` here, and in the body do
```
if venv_ini_path is None:
    venv_ini_path = DEFAULT_VENV_INI_PATH
```
I generally do this whenever the default value is not a simple constant (it looks on the surface like a constant, but involves some platform-dependent computation to derive).

Lib/venv/__init__.py
```diff
@@ -409,8 +424,34 @@ def create(env_dir, system_site_packages=False, clear=False,
...
+    """ Check if a venv.ini exists in the platform location
+        + override defaults if it exists. Use `--help` to test ini config """
+    venvini_path = os.path.expanduser(venv_ini_default)
```
vsajip: No, compute it completely above, as I've suggested in an earlier comment.

Lib/venv/__init__.py
```diff
@@ -4,19 +4,34 @@
...
+from typing import NamedTuple, Optional
 
 CORE_VENV_DEPS = ('pip', 'setuptools')
+VENV_INI_DEFAULT = '~/.venv.ini'
+if sys.platform == "win32":
+    VENV_INI_DEFAULT = os.path.expandvars(r'%APPDATA%\Python\venv\venv.ini')
 logger = logging.getLogger(__name__)
 
 
+class CliDefaults(NamedTuple):
```
vsajip: Suggest renaming CliDefaults -> Defaults. They're just defaults to use for the underlying API, and may or may not come from a command line.

Lib/venv/__init__.py
```diff
@@ -409,8 +424,34 @@ def create(env_dir, system_site_packages=False, clear=False,
...
+def get_default_args(venv_ini_default=VENV_INI_DEFAULT):
+    """ Check if a venv.ini exists in the platform location
+        + override defaults if it exists. Use `--help` to test ini config """
+    venvini_path = os.path.expanduser(venv_ini_default)
+    if not os.path.exists(venvini_path):
+        return CliDefaults()
+
+    cp = configparser.ConfigParser()
+    cp.read(venvini_path)
+    if not cp.has_section("venv"):
```
vsajip: venv -> DEFAULT

Lib/venv/__init__.py
```diff
@@ -409,8 +424,34 @@ def create(env_dir, system_site_packages=False, clear=False,
...
+    arg_overloads = {}
+    for name, value in cp.items("venv"):
+        if value.lower() == "true":
+            arg_overloads[name] = True
+        elif value.lower() == "false":
+            arg_overloads[name] = False
+        else:
+            arg_overloads[name] = value
+
+    return CliDefaults(**arg_overloads)
```
vsajip: This logic will fail with a `TypeError` if the ini contains a variable which is not a valid tuple field name, or a typo in the value. As this is user-specified, untrusted input, checking needs to be bullet-proof.

Would redo as a loop over the known/expected field names, with checking for expected types, doing the 'true'/ True etc. conversions, checking for duplicates (raise `ValueError`) and so on.

Misc/NEWS.d/next/Library/2019-10-15-13-03-59.bpo-34556.K2XyWH.rst
```diff
@@ -0,0 +1 @@
+Add ``venv.ini`` support to venv module to overload CLI argument defaults. Patch by Cooper Ry Lees
```
vsajip: overload -> supply

Lib/test/test_venv.py
```diff
@@ -380,6 +381,34 @@ def test_deactivate_with_strict_bash_opts(self):
...
+        with open(venv_ini_path, "w") as vfp:
+            vfp.write(self.BAD_VENV_INI)
+        with self.assertRaises(configparser.MissingSectionHeaderError):
+            venv.get_default_args(venv_ini_path)
+
+        # Write out good RC file and expect upgrade_deps == True
+        expected = venv.CliDefaults(upgrade_deps=True)
+        with open(venv_ini_path, "w") as vfp:
+            vfp.write(self.GOOD_VENV_INI)
+        self.assertEqual(venv.get_default_args(venv_ini_path), expected)
```
vsajip: I would add more failure cases than just a bad section. Cases to include (both successes and failures):

* Typos in bool values (e.g. `upgrade_deps = flse` - `TypeError` should be raised)
* Duplicated values (`ValueError` should be raised)
* Spurious key values (ignored)
* Multiple values set at once

vsajip (changes requested):

vsajip:
> I do want to state, I think running os.mkdirs on import is bad form. What do you think about moving the dir creation into get_default_args or somewhere else non global?

Agreed - sorry I missed that. That's kind of the reason why I just use `None` in kwarg defaults and do the computation where needed, and not at import time. So in `get_default_args` would be better.

> - I didn't understand what "Multiple values set at once" means ... I feel like configparser covers this

I meant have a test configuration where you e.g. set more than just `upgrade_deps`. That may be your motivating use case but the tests need to cover other scenarios too. Include boolean _and_ string values in the configuration and test for the expected values.

> - I couldn't get the DEFAULT section to work
>  - Everywhere I looked everyone uses a different section with DEFAULT section to provide default values.
>  - I couldn't get it to work how you've requested - Do you have some example code anywhere using it the way you'd like me to use it here?

This configuration:
```
# test.ini
[DEFAULT]
upgrade_deps = True
guido_retired = True
foo = bar
```
with this code:
```
# test.py
import configparser

cp = configparser.ConfigParser()
cp.read('test.ini')
for k, v in cp.items('DEFAULT'):
    print(k, v)
```
prints
```
upgrade_deps True
guido_retired True
foo bar
```
So I'm not sure why you couldn't get the values from the DEFAULT section. Let me know if further clarification is needed.

I'll take another look once you've made the changes discussed above. Thanks!

cooperlees:
Ok, cool - I now do test with multiple values coming from the .ini file.

Here is more info on how DEFAULT fails (I don't get it either as I wrote a very similiar test.py):
Code change to use DEFAULT + test .ini: https://pastebin.com/US2h9tHL
Test Failure Output from `./python.exe -m test -v test_venv` (on my mac): https://pastebin.com/nxpkS0q8

The patch (git diff output) also has the move of mkdirs.

Thoughts here?

vsajip:
> Thoughts here?

Please push your changes so they can be reviewed here rather than on a pastebin. You should be able to get it working with the DEFAULT section - I haven't time now to look into it, sorry!

ambv:
@vsajip: why do you want the configuration section to be called "DEFAULT"? This is not intended usage of this capability of configparser. DEFAULT is to provide reusable keys and values for interpolation or to be reused across many sections. DEFAULT would look ugly as a documented name for the section in venv.ini.

It's not very popular to use the DEFAULT section like this. Notably, .pypirc, .flake8, and setup.cfg don't.

Lib/venv/__init__.py
```diff
@@ -409,8 +429,52 @@ def create(env_dir, system_site_packages=False, clear=False,
...
+    cp.read(venv_ini_path)
+    if not cp.has_section("venv"):
+        return defaults
+
+    read_kwargs = {}
+    for name, value in cp.items("venv"):
+        try:
+            getattr(defaults, name)
+        except AttributeError:
+            continue
```
ambv: At this point `if not hasattr(defaults, name): continue` is clearer. Better yet, to ensure you're only allowing overriding things that are proper fields on the Defaults namedtuple, do:
```py3
if name not in Defaults._fields:
    continue
```

Lib/venv/__init__.py
```diff
@@ -409,8 +429,52 @@ def create(env_dir, system_site_packages=False, clear=False,
...
+        if name in bool_keys:
+            if value.lower() == "true":
+                read_kwargs[name] = True
+            elif value.lower() == "false":
+                read_kwargs[name] = False
+            else:
+                raise ValueError(
+                    f"{name} is a bool. Config can only be set to "
+                    + f"False or True only. Not '{value}'"
+                )
```
ambv: Instead of this entire custom logic, you can just say:
```py3
read_kwargs[name] = cp["venv"].getboolean(key)
```
It will also raise ValueError on invalid values.

Lib/venv/__init__.py
```diff
@@ -409,8 +429,52 @@ def create(env_dir, system_site_packages=False, clear=False,
...
     builder.create(env_dir)
 
+
+def get_default_args(venv_ini_path=VENV_INI_DEFAULT_PATH):
+    """ Check if a venv.ini exists in the platform config location
+        + set defaults if conf exists. Use `--help` to test ini config """
+    bool_keys = {
+        "clear", "copies", "system_site_packages", "upgrade_deps",
+        "use_symlinks", "without_pip"
+    }
```
ambv: To make sure this stays current:
```py3
bool_keys = {
    key
    for key, value_type in Defaults._field_types.items()
    if value_type is bool
}
```

Lib/venv/__init__.py
```diff
@@ -4,18 +4,38 @@
...
 import sysconfig
 import types
-
+from typing import NamedTuple, Optional
 
 CORE_VENV_DEPS = ('pip', 'setuptools')
 logger = logging.getLogger(__name__)
 
+if os.name == "nt":
+    _dir = os.path.expandvars(r'%APPDATA%\Python')
```
ambv: Pro-tip: if your raw-string is not a regular expression and you write R'...' instead of r'...' then GitHub will highlight it better.

ambv (commented):
This looks good to me overall. My only concerns are:

1. Are the configuration paths okay? For example on macOS `appdirs.user_config_dir('python')` will give you `~/Library/Application Support/python` and not `~/.config/python`.
2. Should there be a way to specify a custom `--config=PATH`?

Lib/venv/__init__.py
cooperlees: TIL

Lib/venv/__init__.py
cooperlees: I did discuss above this option above. Cool, if you think it's not crazy I'll do it too.

Lib/venv/__init__.py
cooperlees: Much cleaner. Thanks.

Lib/venv/__init__.py
cooperlees: It seems in `vscode` too in case you didn't know :D

cooperlees:
> This looks good to me overall. My only concerns are:
> 
> 1. Are the configuration paths okay? For example on macOS `appdirs.user_config_dir('python')` will give you `~/Library/Application Support/python` and not `~/.config/python`.

What would be the suggestion here? I am happy to hardcode that for Mac OS X if we feel better or open to a better option. I couldn't see a stdlib or API to use to get this and I wanted to have as little hard coded as possible.

> 2. Should there be a way to specify a custom `--config=PATH`?

My only goal was to allow for different default arguments. This gave (in my head) clear distinction that the config changes defaults and specific options are always preferred.

If people feel having `--config` is useful, we could add a `--config=PATH`, but:
- would it then only alter defaults OR override passed in other options too? 
- Do we just ignore other passed in options is `--config` exists?

I wanted/chose to ignore these edge cases and start simpler by only changing defaults.

vsajip:
> It's not very popular to use the DEFAULT section like this. Notably, .pypirc, .flake8, and setup.cfg don't.

But those configurations have multiple sections, where it might make sense to e.g. use the DEFAULT section for interpolated values. That doesn't apply here, does it? There's only the one section. Are there any use cases where multiple sections might be required?

cooperlees:
Where are these multiple sections you see? Please link me, so I can then go look at their code.

Flake8 - 1 Section: http://flake8.pycqa.org/en/2.6.0/config.html
mypy - 1 Section: https://mypy.readthedocs.io/en/latest/config_file.html
etc.

What do you feel using DEFAULT achieves? Not using DEFAULT (i.e. unique) allows us to potentially allow support from reading setup.cfg or something in the future. Running the test suite shows it causes issues as I linked to above, I can't work out why. If someone can, I'll change it and use it. But I am not for DEFAULT personally, but can live with it if others think it's the correct thing to do.

vsajip:
> Where are these multiple sections you see? Please link me, so I can then go look at their code.

* For `.pypirc`, [this documentation page](https://docs.python.org/3.2/distutils/packageindex.html#the-pypirc-file) shows at least two sections, `distutils` and `pypi`.
* For `.flake8`, [this documentation page](https://flake8.pycqa.org/en/3.7.9/user/configuration.html#using-local-plugins) shows at least two sections, `flake8` and `flake8:local-plugins`.
* For `setup.cfg`, [this documentation page](https://docs.python.org/3/distutils/configfile.html#writing-the-setup-configuration-file) shows at least two sections, `build_ext` and `bdist_rpm`, and makes clear that you can have sections for every `distutils` command such as `build_py`, `install` etc.

Those were the original examples cited. Not sure why you'd need to look at their code.

> What do you feel using DEFAULT achieves?

Obviously it's just an opinion, but if there aren't multiple logical sections in a configuration, I just use the `DEFAULT` section as it's the only one that is expected to be there (sort of, as it's explicitly documented - all others are application-specific). When I write a configuration file that has multiple sections, I name them appropriately and don't use `DEFAULT` in such cases other than for placeholder values or common values, as documented for `ConfigParser` in the Python documentation.

> Running the test suite shows it causes issues as I linked to above, I can't work out why.

Well, the [code snippet I posted above](#issuecomment-549263645) shows that one can read from the `DEFAULT` section just as for any other section.

Doc/using/venv-create.inc
merwok: `$HOME/.config` is the default value for `$XDG_CONFIG_HOME`; if a Python tool follows the spec it should follow it fully IMO.

Could the file be named `venv.cfg` rather than `.ini`?  Windows ini format is an ill-defined format with many variations, whereas `cfg` is more evocative of `configparser` module.

cooperlees:
Just going to switch all my workflows over the virtualenv and automate install of it since it does all this. Thanks.

PR #16805, opened 2019-10-15
bpo-33604: Raise TypeError on missing hmac arg.

gpshead:
Also updates the documentation to clarify the situation surrounding
the digestmod parameter that is required despite its position in the
argument list as of 3.8.0 as well as removing old python2 era
references to "binary strings".

We indavertently had this raise ValueError in 3.8.0 for the missing
arg.  This is not considered an API change as no reasonable code would
be catching this missing argument error in order to handle it.

https://bugs.python.org/issue33604

### DIFF
--- a/Doc/library/hmac.rst
+++ b/Doc/library/hmac.rst
@@ -14,12 +14,13 @@
 This module implements the HMAC algorithm as described by :rfc:`2104`.
 
 
-.. function:: new(key, msg=None, digestmod=None)
+.. function:: new(key, msg=None, digestmod=b'')
 
    Return a new hmac object.  *key* is a bytes or bytearray object giving the
    secret key.  If *msg* is present, the method call ``update(msg)`` is made.
    *digestmod* is the digest name, digest constructor or module for the HMAC
-   object to use. It supports any name suitable to :func:`hashlib.new`.
+   object to use.  It may be any name suitable to :func:`hashlib.new`.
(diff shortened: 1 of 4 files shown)


Doc/library/hmac.rst
```diff
@@ -14,12 +14,13 @@
 This module implements the HMAC algorithm as described by :rfc:`2104`.
 
 
-.. function:: new(key, msg=None, digestmod=None)
+.. function:: new(key, msg=None, digestmod=b'')
```
tiran: digestmod is either a string or a callable, not bytes.

Carreau (approved):

tiran (changes requested):
The digestmod name must be a string or callable.

gpshead:

tirkarthi:
Does changing exception type between 3.8.0 and 3.8.1 needs a versionchanged directive note in doc or is NEWS entry sufficient for these cases?

tiran (approved):

gpshead:
I'm not adding a versionchanged because the exception type when an argument was missing doesn't really count as a meaningful logic part of the API.  Nobody would be likely to ever call this, looking for that exception, and doing something else.  arguments can be made in favor of both TypeError and ValueError, but TypeError is a bit more consistent with the _intent_ so moving it to that seems better.

PR #16813, opened 2019-10-15
bpo-38490: statistics: Add covariance, Pearson's correlation, and simple linear regression

twolodzko:
This PR adds functions for calculating bivariate covariance and Pearson's correlation.

https://bugs.python.org/issue38490


the-knights-who-say-ni:

## Recognized GitHub username

@twolodzko

twolodzko:
I've signed CLA & confirmed mail at least three times now, it does not seem to work.

corona10:
I'd like to recommend you to open the PR after the discussion on https://bugs.python.org/issue38490 is finalized. :)

Doc/library/statistics.rst
```diff
@@ -68,6 +68,17 @@ tends to deviate from the typical or average values.
...
 
+Statistics for relations between two variables
+----------------------------------------------
+
+These functions calculate statistics regarding relations between two random variables.
+
+=========================  =====================================================
+:func:`covariance`         Sample covariance for two variables.
+:func:`correlation`        Pearson's correlation coefficient for two variables.
+:func:`linear_regression`  Intercept and slope fot simple linear regression.
```
taleinat: Typo here.
```suggestion
:func:`linear_regression`  Intercept and slope for simple linear regression.
```

Doc/library/statistics.rst
```diff
@@ -559,6 +570,93 @@ However, for reading convenience, most of the examples show sorted sequences.
 
    .. versionadded:: 3.8
 
+.. function:: covariance(x, y, /)
+
+   Calculates covariance of two variables *x* and *y*. Covariance is
+   a measure of the joint variability of two variables.
+
+   Raises :exc:`StatisticsError` if both variables have same number of data
```
taleinat: This seems to be reversed: An exception would  be raised if the number of data points were different, right?

(The same goes fro the docs of the following function as well.)

taleinat (changes requested):

twolodzko:

Doc/library/statistics.rst
```diff
@@ -559,6 +570,93 @@ However, for reading convenience, most of the examples show sorted sequences.
 
    .. versionadded:: 3.8
 
+.. function:: covariance(x, y, /)
+
+   Calculates covariance of two variables *x* and *y*. Covariance is
+   a measure of the joint variability of two variables.
+
+   Raises :exc:`StatisticsError` if both variables have different number of data
```
taleinat: This can be reworded to be shorter and simpler, in keeping with the style of this document. For example:

"Both inputs must be of the same length (no less than two), otherwise StatisticsError is raised."

The same goes for the other two functions.

Doc/library/statistics.rst
```diff
@@ -559,6 +570,93 @@ However, for reading convenience, most of the examples show sorted sequences.
 
    .. versionadded:: 3.8
 
+.. function:: covariance(x, y, /)
+
+   Calculates covariance of two variables *x* and *y*. Covariance is
```
taleinat: Use "Return the ..." rather than "Calculates", as elsewhere here.

Misc/NEWS.d/next/Library/2019-10-16-08-08-14.bpo-38490.QbDXEF.rst
```diff
@@ -0,0 +1 @@
+Covariance, Pearson's correlation, and simple linear regression functionality was added to statistics module
```
taleinat: Please add a period at the end of the sentence.

Also, add "Patch by Tymoteusz Wołodźko" at the end! :)

taleinat (changes requested):
Thanks for making the effort to add this and for making changes so quickly!

I've gone through the docs thoroughly and have a few requests (see inline comments.)

Additionally, in general, the use of "variables" in the suggested changes isn't currently present in these docs, and is a significant change in tone. Specifically, for many not coming from a Math or Statistics background, it will likely be confused with the common meaning of "variable" in programming, which is slightly but significantly different. Please reword to use "data" or "input data" as done elsewhere in these docs.

taleinat:
Additionally, it seems prudent to state that the covariance computed is what is often called the "sample covariance". This is an estimate for the covariance of the two originating random variables. 

This would be similar to the existing docs for `mean()`: "Return the sample arithmetic mean of data."

Lib/statistics.py
```diff
@@ -817,6 +844,110 @@ def pstdev(data, mu=None):
...
+    (1.75684970486..., 3.09078914170...)
+
+    """
+    n = len(regressor)
+    if len(dependent_variable) != n:
+        raise StatisticsError('linear regression requires that both variables have same number of data points')
+    if n < 2:
+        raise StatisticsError('linear regression requires at least two data points')
+    try:
+        cor = correlation(regressor, dependent_variable)
```
taleinat: This algorithm unnecessarily computes the means and stddev of both data sets twice. It appears that one can simply use `slope = cov(regressor, dependent_variable) / (stdx * stdx)` instead! (This also means there's no need to compute `stdy` at all!)

Or slightly simpler:
```python
slope = cov(regressor, dependent_variable) / variance(regressor)
```

Lib/statistics.py
```diff
@@ -817,6 +844,110 @@ def pstdev(data, mu=None):
...
+        stdx = stdev(regressor)
+        stdy = stdev(dependent_variable)
+        slope = cor * (stdy / stdx)
+    except ZeroDivisionError:
```
taleinat, line 948: This exception handling is unnecessary, since the call to `correlation()` will make the same check.

(If changing this as suggested above to use `covariance()` but not `correlation()`, only the final division operation would this exception handling.)

taleinat (changes requested):
Please also add tests; the doc-tests are not enough.

See: `Lib/test/test_statistics.py`.

Doc/library/statistics.rst
```diff
@@ -596,14 +596,14 @@ However, for reading convenience, most of the examples show sorted sequences.
...
+   for two inputs. Pearson's correlation coefficient *r* takes values
    between -1 and +1. It measures the strength and direction of the linear
    relationship, where +1 means very strong, positive linear relationship,
    -1 very strong, negative linear relationship, and 0 no linear relationship.
 
-   Raises :exc:`StatisticsError` if both variables have different number of data
-   points, or if any of the variables has less then two data points, or if
-   :func:`stdev` of any of the two variables is equal to zero (it is constant).
+   Both inputs must be of the same length (no less than two), and :func:`stdev`
+   of both inputs needs to be greater then zero, otherwise :exc:`StatisticsError`
```
taleinat: Perhaps just "and not be constant" rather than "and stdev of both inputs needs to be greater than zero"? (Below for linear regression as well.)

Lib/statistics.py
```diff
@@ -93,8 +93,8 @@
 0.75
 >>> correlation(x, y)  #doctest: +ELLIPSIS
 0.31622776601...
->>> linear_regression(x, y)  #doctest: +ELLIPSIS
-(1.5, 0.099999999999...)
+>>> linear_regression(x, y)  #doctest:
```
taleinat, line 96: Maybe use the following to avoid decimal representation issues? Or use different values for the example so that "+ELLIPSIS" is enough.

```python3
>>> '%.2f, %.2f' % linear_regression(x, y)
'1.50, 0.10'
```

Lib/statistics.py
```diff
@@ -854,8 +854,8 @@ def pstdev(data, mu=None):
 def covariance(x, y, /):
     """Covariance
 
-    Calculates covariance of two variables *x* and *y*. Covariance is
-    a measure of the joint variability of two variables.
+    Return the covariance of two inputs *x* and *y*. Covariance is
```
taleinat: Forgot to use "sample covariance" here too :)

Lib/test/test_statistics.py
```diff
@@ -2312,6 +2312,72 @@ def test_error_cases(self):
...
+    
+    def test_unequal_size(self):
+        for x, y in [
+            ([1, 2, 3], [1, 2]),
+            ([1, 2], [1, 2, 3]),
+        ]:
+            with self.assertRaises(self.module.StatisticsError):
+                statistics.covariance(x, y)
+            with self.assertRaises(self.module.StatisticsError):
+                statistics.covariance(x, y)
```
taleinat: Copy/paste error here, testing `covariance()` twice but not `correlation()`. Below as well.

Lib/test/test_statistics.py
```diff
@@ -2312,6 +2312,72 @@ def test_error_cases(self):
...
+            ([1, 2,], [1,]),
+        ]:
+            with self.assertRaises(self.module.StatisticsError):
+                statistics.covariance(x, y)
+            with self.assertRaises(self.module.StatisticsError):
+                statistics.covariance(x, y)
+            with self.assertRaises(self.module.StatisticsError):
+                statistics.linear_regression(x, y)
+                
+
```
taleinat: Add test cases for covariance as well?

Doc/library/statistics.rst
```diff
@@ -559,6 +570,93 @@ However, for reading convenience, most of the examples show sorted sequences.
...
+      >>> x = [1, 2, 3, 4, 5, 6, 7, 8, 9]
+      >>> y = [9, 8, 7, 6, 5, 4, 3, 2, 1]
+      >>> correlation(x, x)
+      1.0
+      >>> correlation(x, y)
+      -1.0
+
+.. function:: linear_regression(regressor, dependent_variable)
+
+   Return the ``(intercept, slope)`` tuple of the `simple linear regression
```
taleinat: ISTM this should also mention that this is for simple linear regression using the "ordinary least squares" method. (In the doc-string too.)

Doc/library/statistics.rst
```diff
@@ -559,6 +570,93 @@ However, for reading convenience, most of the examples show sorted sequences.
...
+
+   .. doctest::
+
+      >>> year = [1971, 1975, 1979, 1982, 1983]
+      >>> films_total = [1, 2, 3, 4, 5]
+      >>> intercept, slope = linear_regression(year, films_total)
+      >>> round(intercept + slope * 2019)
+      16
+
+   We could also use it to predict how many Monty Python films existed when
```
taleinat: This is hilarious! (Perhaps `use it to "predict" how many...`, since this isn't an actual prediction?)

taleinat (changes requested):
Excellent! Just a few more small things.

taleinat:
Also note that there's a whitespace issue in one of the files (that's causing the test failures here.) To find the issue quicly, you can run `make patchcheck`, or without make (e.g. on Windows), `python Tools/scripts/patchcheck.py`.

twolodzko:

Lib/statistics.py
twolodzko: It seems to work now.

taleinat: Yes, but tests that sometimes fail can require a lot of effort down the road. We should make every effort to make tests robust.

taleinat:
@rhettinger, I think this is now ready for your review.

taleinat (approved):
LGTM

Through possibly some doctests could use more robust handling of float representation issues, e.g. `0.1` vs. `0.099999999...`.

twolodzko:
Is there anything I should do about the PR, or it just got lost into the void?

taleinat:
@twolodzko, this PR isn't lost in the void. I regret the long delay.

This is awaiting review from @rhettinger and/or those from whom he requested a review, @tim-one and @stevendaprano.

Lib/statistics.py
```diff
@@ -817,6 +844,108 @@ def pstdev(data, mu=None):
...
+    cov = covariance(x, y)
+    stdx = stdev(x)
+    stdy = stdev(y)
+    try:
+        return cov / (stdx * stdy)
+    except ZeroDivisionError:
+        raise StatisticsError('at least one of the inputs is constant')
+
+
+def linear_regression(regressor, dependent_variable):
```
rhettinger: Let's make these positional-only arguments.  We may well want to change the name of these parameters later.

Lib/statistics.py
```diff
@@ -817,6 +844,108 @@ def pstdev(data, mu=None):
...
+    if len(dependent_variable) != n:
+        raise StatisticsError('linear regression requires that both inputs have same number of data points')
+    if n < 2:
+        raise StatisticsError('linear regression requires at least two data points')
+    try:
+        slope = covariance(regressor, dependent_variable) / variance(regressor)
+    except ZeroDivisionError:
+        raise StatisticsError('regressor is constant')
+    intercept = mean(dependent_variable) - slope * mean(regressor)
+    return intercept, slope
```
rhettinger: I recommend using a named tuple here.

twolodzko: Done. I used `LinearRegression` namedtuple, if you have suggestion for better name, feel free to suggest something. I considered `RegressionParameters` or `LinearRegressionParameters`, but first is slightly ambiguous (there are different "regressions"), and second one rather lengthy. The only problem with `LinearRegression` is that it conflicts with scikit-learn class with same name, but I hardly see scenario where this would be a problem for anyone.

Lib/statistics.py
```diff
@@ -910,10 +910,13 @@ def correlation(x, y, /):
         raise StatisticsError('at least one of the inputs is constant')
 
 
-def linear_regression(regressor, dependent_variable):
+LinearRegression = namedtuple('LinearRegression', ['intercept', 'slope'])
```
rhettinger, line 913: The parameter order seems backwards to me, here and through out the PR.  Traditionally, we say an equation for a line in is slope-intercept form as `y = mx + b` with the *m* preceding the *b*.

twolodzko: It depends on your background, [in statistics](https://en.wikipedia.org/wiki/Linear_regression#Introduction) the model is often written as y = β_0 + β_1 * x + ε , so the intercept is written before. Also, if now it is a named tuple, the order is of less importance. Of course, if you insist, I may change it.

stevendaprano:
Thank you Tymoteusz for your work.

I agree with Raymond that using a namedtuple is a good idea here.

On Fri, Oct 09, 2020 at 12:52:16AM -0700, Raymond Hettinger wrote:
> The parameter order seems backwards to me, here and through out the 
> PR.  Traditionally, we say an equation for a line in is 
> slope-intercept form as `y = mx + b` with the *m* preceding the *b*.

This is true for most areas of mathematics except for statistics and 
finance. In those, the usual form of the linear regression equation is:

    y = a + bx

which as Raymond points out is the reverse of the usual form:

    y = mx + c

(or sometimes `mx + b` for the constant term).


See, for example:

https://www.statisticssolutions.com/conduct-interpret-linear-regression/

https://www.investopedia.com/terms/r/regression.asp



-- 
Steve

stevendaprano:
On Fri, Oct 09, 2020 at 01:15:52AM -0700, Tymoteusz Wołodźko wrote:



> Also, if now it is a named tuple, the order is of 

> less importance. Of course, if you insist, I may change it.



I would prefer you keep the (intercept, gradient) order which is more 

conventional in statistics.

twolodzko:
@stevendaprano @rhettinger should I apply any more changes, or is it ok as-is?

taleinat:
The only current outstanding issue is the order of returned values from `linear_regression()`.

ISTM given the short discussion here that the order of `(intercept, gradient)` could go either way, and that in the context of statistics the current order seems fine. And since this will be a named tuple, the order is not as important as it would be otherwise. As long as the documentation and doc-string are consistent and clear on this, which they are, IMO this is fine.

I'd like to get @rhettinger's approval, or otherwise, for this before moving forward.

stevendaprano:
Given that this is the statistics module, not linear algebra, I think we 
ought to stick to the common convention taught in stats classes, which 
is

    y = a + b x

i.e. (intercept, gradient) in that order. Unless Raymond strongly 
objects, I say go with this order.

Do we have time to add a new feature to this before feature-freeze? I 
would like to add two methods to the tuple returned:


    def predict_y(self, x):
        """Return the predicted y value for the given x.

        Returns the predicted response or dependent variable.
        """
        return self.intercept + self.gradient*x


    def predict_x(self, y):
        """Return the predicted x value for the given y.

        Returns the predicted explanatory or independent variable.
        """
        return (y - self.intercept)/self.gradient


Predicting the x or y values from the linear regression line are very 
common operations, I think it would be very useful to offer them without 
expecting users to create the function themselves.


-- 
Steve

twolodzko:
@stevendaprano I'd be _against_ having `predict_x` as this is incorrect if this is a statistical model y = a + bx + noise, because it ignores the noise.

twolodzko:
It's been two weeks and the discussion seems to got stuck. Is there anything more I should change about the PR?

ZackerySpytz:
Please add an entry to `Doc/whatsnew/3.10.rst` and add `versionadded` directives to the documentation for the new functions.

taleinat:
> @stevendaprano I'd be _against_ having `predict_x` as this is incorrect if this is a statistical model y = a + bx + noise, because it ignores the noise.

I'm also -1 on adding `.predict_{x,y}` methods. I feel that they would be beyond the scope of what this module sets out to achieve, and they'd also be surprising for many to find on a named tuple.

Doc/whatsnew/3.10.rst
```diff
@@ -217,6 +217,13 @@ The :mod:`shelve` module now uses :data:`pickle.DEFAULT_PROTOCOL` by default
 instead of :mod:`pickle` protocol ``3`` when creating shelves.
 (Contributed by Zackery Spytz in :issue:`34204`.)
 
+statistics
+----------
+
+Added :func:`covariance`, Pearson's :func:`correlation`, and simple :func:`linear_regression`
+functionalities to the :mod:`statistics` module.
```
taleinat, line 224:
```suggestion
functions to the :mod:`statistics` module.
```

taleinat:
```suggestion
Added :func:`~statistics.covariance`, Pearson's
:func:`~statistics.correlation`, and simple
:func:`~statistics.linear_regression` functions.
```

taleinat:
Apart from my minor suggestion for the What's New entry, this looks good to me.

@rhettinger, would you like to take another look? Any other comments?

Doc/library/statistics.rst
```diff
@@ -564,6 +575,98 @@ However, for reading convenience, most of the examples show sorted sequences.
...
+
+.. function:: linear_regression(regressor, dependent_variable)
+
+   Return the intercept and slope of `simple linear regression
+   <https://en.wikipedia.org/wiki/Simple_linear_regression>`_
+   parameters estimated using ordinary least squares. Simple linear
+   regression describes relationship between *regressor* and
+   *dependent variable* in terms of linear function:
+
+      *dependent_variable = intercept + slope * regressor + noise*
```
stevendaprano: The markup here seems to be broken.

    *dependent_variable = intercept + slope * regressor + noise*

Should be:

    *dependent_variable = intercept + slope \* regressor + noise*

Untested; if the backslash doesn't work, try verbatim mode

    ```*```

to escape the multiplication sign.

Misc/NEWS.d/next/Library/2019-10-16-08-08-14.bpo-38490.QbDXEF.rst
```diff
@@ -0,0 +1 @@
+Covariance, Pearson's correlation, and simple linear regression functionality was added to statistics module. Patch by Tymoteusz Wołodźko.
```
stevendaprano, line 1: Did you mean "functionality" or "functions"? I am good with either choice, but "functions" might be slightly less unusual.

stevendaprano (commented):
@twolodzko I am very impressed in this, thank you, and I look forward to using it. I just commented on what looks like a broken piece of ReST and a slightly unusual choice of wording, but apart from this I am very happy with your patch and I think it should be merged.

Misc/NEWS.d/next/Library/2019-10-16-08-08-14.bpo-38490.QbDXEF.rst
taleinat: @twolodzko has already applied my suggested changes to that, which include replacing "functionality" with "functions".

Doc/library/statistics.rst
taleinat: Actually, the existing markup works as intended, because asterisks surrounded by whitespace are apparently ignored. See [the docs](https://www.sphinx-doc.org/en/master/usage/restructuredtext/basics.html#inline-markup) for some more details.

I still think that explicitly escaping the middle asterisk, as suggested by @stevendaprano, would be a good idea to make it's intention clear and for future-proofing.

twolodzko:
If I'm not missing something, all issues seem to be resolved right now.

taleinat:
> If I'm not missing something, all issues seem to be resolved right now.

Indeed.

twolodzko:
It's been over 20 days since last discussion on this PR. I don't want to be pushy, but this PR was posted over a year ago and at some point I would forget about it as well, and all the time we all spend on it would be wasted. If there's anything I can do about it, I'm open to it, but right now it is not exactly clear for me if we are waiting for something?

ZackerySpytz:
Unfortunately, it seems like there are some minor grammatical errors in the new docstrings and documentation.

twolodzko:
@ZackerySpytz could you be more specific?

twolodzko:
I don't know what is to be fixed. Hope this will get merged at some point. If there's anything I need to do, please tag me.

twolodzko:
@taleinat is there anything blocking? Sorry for pinging you, but you seem to be the only one who cares about this PR.

twolodzko:
This PR is ready. It waits for reviews and merge. If there is anything I can do about it, please let me know.

taleinat:
Merged!

This has indeed reached a very good state a while ago and was reviewed favorably by several core devs. Definitely good to have this for 3.10.

PR #16815, opened 2019-10-15
bpo-35998: Fix test_asyncio.test_start_tls_server_1()

vstinner:
main() is now responsible to send the ANSWER, rather than
ServerProto. main() now waits until it got the HELLO before sending
the ANSWER over the new transport.

Previously, there was a race condition between main() replacing the
protocol and the protocol sending the ANSWER once it gets the HELLO.

TLSv1.3 was disabled for the test: reenable it.

https://bugs.python.org/issue35998

### DIFF
--- a/Lib/test/test_asyncio/test_sslproto.py
+++ b/Lib/test/test_asyncio/test_sslproto.py
@@ -497,14 +497,6 @@ class BaseStartTLS(func_tests.FunctionalTestCaseMixin):
 
         server_context = test_utils.simple_server_sslcontext()
         client_context = test_utils.simple_client_sslcontext()
-        if (sys.platform.startswith('freebsd')
-                or sys.platform.startswith('win')
-                or sys.platform.startswith('darwin')):
-            # bpo-35031: Some FreeBSD and Windows buildbots fail to run this test
-            # as the eof was not being received by the server if the payload
-            # size is not big enough. This behaviour only appears if the
-            # client is using TLS1.3.  Also seen on macOS.
-            client_context.options |= ssl.OP_NO_TLSv1_3
         answer = None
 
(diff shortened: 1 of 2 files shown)


vstinner:
I tested manually this fix on my laptop using `./python -u -m test test_asyncio -m test.test_asyncio.test_sslproto.SelectorStartTLSTests.test_start_tls_server_1 -F -j100` command. Without my fix, the command fails in less than 100 attempts in less than 30 seconds. With the fix, the test runs successfuly more than 1000 times for longer than 3 minutes:

```
...
0:03:53 load avg: 99.51 [1315] test_asyncio passed
0:03:53 load avg: 99.51 [1316] test_asyncio passed
0:03:53 load avg: 99.51 [1317] test_asyncio passed
0:03:53 load avg: 99.51 [1318] test_asyncio passed
```

I did a similar check on Gentoo's buildbot worker using using `./python -u -m test test_asyncio -m test.test_asyncio.test_sslproto.SelectorStartTLSTests.test_start_tls_server_1 -F -j2` command. Without the fix, it fails in less than 3 attempts in a few seconds. With the fix, the test runs successfully more than 250 tests for longer than 5 minutes:

```
...
0:05:13 load avg: 11.74 [278] test_asyncio passed
0:05:13 load avg: 11.74 [279] test_asyncio passed
0:05:13 load avg: 11.74 [280] test_asyncio passed
```

pablogsal (approved):
Thanks for fixing this tricky bug, @vstinner.

Great work!

vstinner:
This change needs to be backported to 3.7 and 3.8, but I prefer to wait to see if buildbots like this new change, before doing the backport.

PR #16819, opened 2019-10-16
bpo-38470: Add hostname support to ssl.get_server_certificate.

juhovh:
Many servers in the cloud environment require SNI to be used during the
SSL/TLS handshake, therefore it is not possible to fetch their certificates
using the ssl.get_server_certificate interface.

This change adds an additional optional hostname argument that can be used to
set the SNI. Note that it is intentionally a separate argument instead of
using the host part of the addr tuple, because one might want to explicitly
fetch the default certificate or fetch a certificate from a specific IP
address with the specified SNI hostname. A separate argument also works better
for backwards compatibility.

https://bugs.python.org/issue38470

### DIFF
--- a/Doc/library/ssl.rst
+++ b/Doc/library/ssl.rst
@@ -425,14 +425,15 @@ Certificate handling
       previously. Return an integer (no fractions of a second in the
       input format)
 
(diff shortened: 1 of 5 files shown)


juhovh:
Accidentally had invalid bpo issue name. 😓 Will close this and re-open a new one.

PR #16822, opened 2019-10-16
Improve the Distributing Python Modules and Installing Python Modules documentation

geryogam:
Changes to the [Distributing Python Modules](https://docs.python.org/3/distributing/index.html) and [Installing Python Modules](https://docs.python.org/3/installing/index.html) documentation:

- correct some redirected URIs (because of wrong path component) with dead fragment components;
- correct a proper noun (Python Packaging Index -> Python Package Index);
- correct some "module" terms to "package" (= _distribution_ package, not _import_ package/module);
- improve punctuation;
- improve case.


Doc/distributing/index.rst
```diff
@@ -1,12 +1,11 @@
 .. _distributing-index:
 
-###############################
-  Distributing Python Modules
-###############################
+****************************
+Distributing Python Packages
+****************************
```
merwok, line 5: Hello!  Thanks for making a PR to improve Python docs!

Distutils is deprecated: its code and docs have been adopted by the setuptools project.  Distutils and setuptools are also less important now that pyproject.toml exists so that people can use different build tools and keep compatibility with PyPI and pip / tox / etc.

For these reasons, I would only accept clear fixes, not formatting or style changes.  The two changes here for example (header style and word choice) are out of scope.

Doc/distributing/index.rst
```diff
@@ -31,10 +30,10 @@ installing other Python projects, refer to the
 Key terms
 =========
 
-* the `Python Packaging Index <https://pypi.org>`__ is a public
+* The `Python Package Index <https://pypi.org>`__ is a public
```
merwok, line 33: fixed in 0be338199fd663f020d833a4db185d0c5a0e0078

Doc/distributing/index.rst
```diff
@@ -119,47 +119,47 @@ recommended tools`_.
...
+   https://packaging.python.org/tutorials/packaging-projects/#generating-distribution-archives
+.. _Uploading the project to the Python Package Index: \
+   https://packaging.python.org/tutorials/packaging-projects/#uploading-the-distribution-archives
 .. _The .pypirc file: \
    https://packaging.python.org/specifications/pypirc/
 
 
-How do I...?
-============
+How do I…
```
merwok, line 144: I rarely see English speakers use this character; I use it in personal writing but not in the Python docs.

Doc/installing/index.rst
```diff
@@ -196,8 +196,8 @@ Common installation issues
 Installing into the system Python on Linux
 ------------------------------------------
 
-On Linux systems, a Python installation will typically be included as part
-of the distribution. Installing into this Python installation requires
+On Linux, a Python installation will typically be included as part
+of the Linux distribution. Installing into this Python installation requires
```
merwok, line 200: Disagree with these

Doc/installing/index.rst
```diff
@@ -220,14 +220,14 @@ There are also additional resources for `installing pip.
 Installing binary extensions
 ----------------------------
 
-Python has typically relied heavily on source based distribution, with end
-users being expected to compile extension modules from source as part of
+Python has typically relied heavily on source based distributions, with end
+users being expected to compile binary extensions from source as part of
```
merwok, line 224: _extension modules_ is the way Python calls these (native code shared libraries built from other languages)

merwok:
One wrong word is already fixed, the other things are too late or incorrect.  Thanks but I’m rejected these changes.

maggyero:
Fair enough. Thanks for the review.

PR #16826, opened 2019-10-16
bpo-17393: do not convert import of non-package

romuald:
In lib2to3, when a patched code is importing a module that has the same
name as a local directory, check that it contains an `__init__.py` file
before converting it to a local import

Quick example:

```
fixtures/__init__.py
fixtures/json/things.json
```

Previously `import json` was converted to `from . import json`

https://bugs.python.org/issue17393

### DIFF
--- a/Lib/lib2to3/fixes/fix_import.py
+++ b/Lib/lib2to3/fixes/fix_import.py
@@ -12,7 +12,7 @@ Becomes:
 
 # Local imports
 from .. import fixer_base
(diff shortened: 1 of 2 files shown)


encukou:
I don't feel comfortable reviewing changes here. The other suggested reviewers might also not know too much about lib2to3; they were autoselected by the `import` keyword.
Adding Benjamin as the lib2to3 expert.

PR #16829, opened 2019-10-16
bpo-38502: regrtest uses process groups if available

vstinner:
regrtest now uses process groups in the multiprocessing mode (-jN
command line option) if process groups are available: if os.setsid()
and os.killpg() functions are available.

https://bugs.python.org/issue38502

### DIFF
--- a/Lib/test/libregrtest/runtest_mp.py
+++ b/Lib/test/libregrtest/runtest_mp.py
@@ -3,6 +3,7 @@ import faulthandler
 import json
 import os
 import queue
+import signal
 import subprocess
 import sys
 import threading
@@ -30,6 +31,8 @@ assert MAIN_PROCESS_TIMEOUT >= PROGRESS_UPDATE
 
 # Time to wait until a worker completes: should be immediate
 JOIN_TIMEOUT = 30.0   # seconds
+
+USE_PROCESS_GROUP = (hasattr(os, "setsid") and hasattr(os, "killpg"))
 
 
 def must_stop(result, ns):
@@ -59,12 +62,16 @@ def run_test_in_subprocess(testname, ns):
     # Running the child from the same working directory as regrtest's original
     # invocation ensures that TEMPDIR for the child is the same when
     # sysconfig.is_python_build() is true. See issue 15300.
+    kw = {}
+    if USE_PROCESS_GROUP:
+        kw['start_new_session'] = True
     return subprocess.Popen(cmd,
                             stdout=subprocess.PIPE,
                             stderr=subprocess.PIPE,
                             universal_newlines=True,
                             close_fds=(os.name != 'nt'),
-                            cwd=support.SAVEDCWD)
+                            cwd=support.SAVEDCWD,
+                            **kw)
 
 
 def run_tests_worker(ns, test_name):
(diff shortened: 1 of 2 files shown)


vstinner:
@nanjekyejoannah @pitrou @pablogsal @gpshead: Do you any drawback in this approach?

Lib/test/libregrtest/runtest_mp.py
```diff
@@ -149,16 +156,21 @@ def _kill(self):
...
+        if USE_PROCESS_GROUP:
+            what = f"{self} process group"
+        else:
+            what = f"{self}"
+
+        print(f"Kill {what}", file=sys.stderr, flush=True)
         try:
-            popen.kill()
+            if USE_PROCESS_GROUP:
+                os.killpg(popen.pid, signal.SIGKILL)
```
pitrou, line 167: Shouldn't you call `popen.wait()` after this?

Lib/test/libregrtest/runtest_mp.py
```diff
@@ -149,16 +156,21 @@ def _kill(self):
...
+            else:
+                popen.kill()
         except ProcessLookupError:
-            # Process completed, the TestWorkerProcess thread read its exit
```
pitrou, line 156: Why do you remove this comment?

pitrou (commented):
Looks good on the principle, but a couple comments.

Misc/NEWS.d/next/Tests/2019-10-17-00-49-38.bpo-38502.vUEic7.rst
```diff
@@ -0,0 +1,3 @@
+regrtest now uses process groups in the multiprocessing mode (-jN command line
```
gpshead: test.regrtest

Lib/test/libregrtest/runtest_mp.py
gpshead: fwiw, the old/existing code didn't.  (`popen.kill()` does not call `.wait()`)

gpshead (commented):
Overall I think trying this to see how it pans out.  I can't claim to understand all of the specifics of process groups and/or sessions.

Lib/test/libregrtest/runtest_mp.py
vstinner: popen.wait() is always called.

The main thread calls stop() which calls _kill().

wait() is called in the TestWorkerProcess thread: _run_process() ends with "finally: self._wait_completed()" to ensure that we also read the exit status to prevent leaking a zombie process.

In my experience, calling wait() from 2 different threads is causing too many troubles :-)

Misc/NEWS.d/next/Tests/2019-10-17-00-49-38.bpo-38502.vUEic7.rst
vstinner: Fixed.

Lib/test/libregrtest/runtest_mp.py
vstinner: I added the comment back.

PR #16832, opened 2019-10-17
bpo-38580: Document that select() accepts iterables, not just sequences

jstasiak:
~~I think this one can go through without an issue number.~~

https://bugs.python.org/issue38580

### DIFF
--- a/Modules/selectmodule.c
+++ b/Modules/selectmodule.c
@@ -232,11 +232,11 @@ select.select
 
 Wait until one or more file descriptors are ready for some kind of I/O.
 
-The first three arguments are sequences of file descriptors to be waited for:
+The first three arguments are iterables of file descriptors to be waited for:
 rlist -- wait until ready for reading
 wlist -- wait until ready for writing
 xlist -- wait for an "exceptional condition"
-If only one kind of condition is required, pass [] for the other lists.
+If only one kind of condition is required, pass empty iterables for the other lists.
 
 A file descriptor is either a socket or file object, or a small integer
 gotten from a fileno() method call on one of those.
@@ -313,7 +313,7 @@ select_select_impl(PyObject *module, PyObject *rlist, PyObject *wlist,
     }
 #endif /* SELECT_USES_HEAP */
 
-    /* Convert sequences to fd_sets, and get maximum fd number
+    /* Convert iterables to fd_sets, and get maximum fd number
      * propagates the Python exception set in seq2set()
      */
     rfd2obj[0].sentinel = -1;
--- a/Modules/selectmodule.c
+++ b/Modules/selectmodule.c
@@ -257,7 +257,7 @@ descriptors can be used.
 static PyObject *
 select_select_impl(PyObject *module, PyObject *rlist, PyObject *wlist,
                    PyObject *xlist, PyObject *timeout_obj)
-/*[clinic end generated code: output=2b3cfa824f7ae4cf input=177e72184352df25]*/
+/*[clinic end generated code: output=2b3cfa824f7ae4cf input=fc219cac9803bb5f]*/
 {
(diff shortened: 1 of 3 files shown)


brandtbucher (approved):

brandtbucher:
The argument clinic docstring for `select_select_impl` in `Modules/selectmodule.c` can probably be updated too... but I think this PR is fine without it.

corona10 (commented):
As @brandtbucher review,
Please update the docstring of https://github.com/python/cpython/blob/303475e87372aacb17990802538e440735a69525/Modules/selectmodule.c#L235

jstasiak:
> As @brandtbucher review,
> Please update the docstring of
> 
> https://github.com/python/cpython/blob/303475e87372aacb17990802538e440735a69525/Modules/selectmodule.c#L235

Done.

corona10 (commented):
@jstasiak cc @brandtbucher 

IMHO, I don't know this is the right patch.
I am not a select module expert, but if this patch is accepted.
Docstring means that it will accept the dictionary for the  `rlist, wlist, xlist`.
(I am not meaning select.select({}, {}, {}, 1) rasing TypeError).
But is this working properly?

brandtbucher:
@corona10 Yes, it should be fine. This function calls `seq2set` on each iterable, which uses `PySequence_Fast` to convert them to lists (or similar) before doing any work.

Any finite iterable should be okay here.

jstasiak:
> Docstring means that it will accept the dictionary for the `rlist, wlist, xlist`.

@corona10 Yeah it accepts dictionaries now, that's how I discovered that the documentation is being overly strict here.

corona10 (changes requested):
@jstasiak 
> Yeah it accepts dictionaries now, that's how I discovered that the documentation is being overly strict here.

I think that we should discuss with core developers which part should be changed.

- Updating implementation to only accept sequence type and raising TypeError
or
- Updating docs.

I 'd like to recommend it to open a new issue on bug.python.org
After the decision is finalized we can finish this PR.

So I'm left request change, not to this PR but request to opening the issue.

Thanks!

brandtbucher:
Okay then. Since this more controversial than I anticipated, I think a BPO issue should be opened for this.

jstasiak:
That's fair, just opened https://bugs.python.org/issue38580

jstasiak:
There's no interest in [the issue](https://bugs.python.org/issue38580) so far, so I let myself mention two recent select module contributors in hope I receive some valuable input here: @vstinner @taleinat

Modules/selectmodule.c
```diff
@@ -232,11 +232,11 @@ select.select
 
 Wait until one or more file descriptors are ready for some kind of I/O.
 
-The first three arguments are sequences of file descriptors to be waited for:
+The first three arguments are iterables of file descriptors to be waited for:
 rlist -- wait until ready for reading
 wlist -- wait until ready for writing
 xlist -- wait for an "exceptional condition"
-If only one kind of condition is required, pass [] for the other lists.
```
taleinat: In this line, the original "pass []" is clearer and simpler than "pass empty iterables", and is still correct.

taleinat (changes requested):
Thanks for following this up, @jstasiak. This looks very good to me, I have just one minor comment.

Modules/selectmodule.c
jstasiak: That's fair, I brought back the original.

jstasiak:
@taleinat Should I write a news entry or is it fine as it is?

taleinat:
> @taleinat Should I write a news entry or is it fine as it is?

Since this is just updating the documentation to better describe the existing behavior, IMO there's no need for a NEWS entry, as for other normal doc-only changes.

PR #16838, opened 2019-10-18
bpo-38517: Support partial functions & partialmethod's in functools.cached_property

ricardobranco777:
Currently, the cached_property decorator does not support partial functions and partialmethod's.

This is needed so we can dynamically create cached attributes in a class like this:

This PR adds a `name` parameter to specify the name of the function.

Use case:

```
import os
from functools import cached_property, partialmethod

class ProcNet:
    def __new__(cls, *args, **kwargs):
        for proto in ('icmp', 'icmp6', 'raw', 'raw6', 'tcp', 'tcp6', 'udp', 'udp6', 'udplite', 'udplite6'):
            setattr(cls, proto, cached_property(partialmethod(cls._proto, proto), name=proto))
        return super().__new__(cls, *args, **kwargs)

    def _proto(self, proto):
        """
        Get /proc/net/{icmp,icmp6,raw,raw6,tcp,tcp6,udp,udp6,udplite,udplite6}
        """
        with open(os.path.join("/proc/net", proto)) as file:
            return file.read()

net = ProcNet()
print(net.tcp)
```

https://bugs.python.org/issue38517

### DIFF
--- a/Lib/functools.py
+++ b/Lib/functools.py
@@ -911,9 +911,9 @@ _NOT_FOUND = object()
 
 
 class cached_property:
-    def __init__(self, func):
+    def __init__(self, func, name=None):
         self.func = func
-        self.attrname = None
+        self.attrname = name
         self.__doc__ = func.__doc__
         self.lock = RLock()
 
@@ -946,7 +946,10 @@ class cached_property:
                 # check if another thread filled cache while we awaited lock
                 val = cache.get(self.attrname, _NOT_FOUND)
                 if val is _NOT_FOUND:
-                    val = self.func(instance)
+                    if callable(self.func):
+                        val = self.func(instance)
+                    else:
+                        val = self.func.func(instance, *self.func.args, **self.func.keywords)
                     try:
                         cache[self.attrname] = val
                     except TypeError:
--- a/Misc/NEWS.d/next/Library/2019-10-18-16-57-52.bpo-38517.PBwbZj.rst
+++ b/Misc/NEWS.d/next/Library/2019-10-18-16-57-52.bpo-38517.PBwbZj.rst
@@ -1,0 +1,2 @@
+Add support for partial functions and partialmethod's to
+functools.cached_property


the-knights-who-say-ni:

## Recognized GitHub username

@ricardobranco777

taleinat:
According to @ncoghlan's message on the issue, this should be closed, since will be confusingly inconsistent with other kinds of descriptors.

taleinat:
Sorry for the previous comment being so negative and terse...

Thanks for this PR, @ricardobranco777!

Note that for the supplied example, you can already achieve this today, slightly awkwardly, using normal `functools.partial()`:
```python3
class ProcNet:
    def __new__(cls, *args, **kwargs):
        for proto in ('icmp', 'icmp6', 'raw', 'raw6', 'tcp', 'tcp6', 'udp', 'udp6', 'udplite', 'udplite6'):
            prop = cached_property(partial(cls._proto, proto=proto))
            setattr(cls, proto, prop)
            prop.__set_name__(cls, proto)
        return super().__new__(cls, *args, **kwargs)
```

I'll also have to mention that the above is still a strange way to achieve what the example is doing, since it will re-create the cached property handlers for each instance. The following, for example, will properly set them on the class:

```python3
class ProcNet:
    pass

for proto in ('icmp', 'icmp6', 'raw', 'raw6', 'tcp', 'tcp6', 'udp', 'udp6', 'udplite', 'udplite6'):
    @cached_property
    def prop(self, *, proto=proto):
        with open(os.path.join("/proc/net", proto)) as file:
            return file.read()
    setattr(ProcNet, proto, prop)
    prop.__set_name__(ProcNet, proto)
```

taleinat:
Or to completely avoid the extra parameter in the properties:

```python3
class ProcNet:
    pass

def make_prop(proto):
    @cached_property
    def prop(self):
        with open(os.path.join("/proc/net", proto)) as file:
            return file.read()
    setattr(ProcNet, proto, prop)
    prop.__set_name__(ProcNet, proto)
for proto in ('icmp', 'icmp6', 'raw', 'raw6', 'tcp', 'tcp6', 'udp', 'udp6', 'udplite', 'udplite6'):
    make_prop(proto)
```

taleinat:
I'm closing this given the current state of discussion. If further discussion shows that this is actually desired, I'd be happy to re-open this.

ricardobranco777:
True.  In this case I think this is cleaner:

```
class ProcNet:
    def __new__(cls, *args, **kwargs):
        if not hasattr(cls, 'icmp'):
            for proto in ('icmp', 'icmp6', 'raw', 'raw6', 'tcp', 'tcp6', 'udp', 'udp6', 'udplite', 'udplite6'):
                prop = cached_property(partial(cls._proto, proto=proto))
                setattr(cls, proto, prop)
                prop.__set_name__(cls, proto)
        return super().__new__(cls, *args, **kwargs)
```

taleinat:
@ricardobranco777, yep, that looks like it would work too.

It does have its downsides:
* It is not thread-safe, though in this case that's likely not a significant issue.
* It incurs an avoidable, if small, overhead during the creation of every instance.

Ultimately it's mostly a matter of taste :)

